PR #7, opened 2024-11-07
[misc] fix: unknown keyword arg model for single gpu and file name math conflict

goriri:


PeterSH6:
Hi, it seems that there are some bugs when world_size == 1. Thanks for your contribution!

However, I found that your commit cannot solve this bug as `GPUExecutor` from vllm doesn't receive a model `state_dict` from outside (i.e., actor). I think we can simply use `SPMDGPUExecutor` when world_size == 1.

I will fix it later.

PeterSH6:
Hi @goriri , we fix it here: https://github.com/volcengine/verl/pull/9

PR #18, opened 2024-11-20
[chore] Break the loop after obtaining the register_center actor

kevin85421:
Currently, the loop will always iterate 120 times.


PeterSH6:
Nice catch!

PR #19, opened 2024-11-21
[chore] remove unnecessary updating of `_worker_names`

kevin85421:
It looks like that `self._worker_names` doesn't need to be updated.

https://github.com/volcengine/verl/blob/27b43eba2b8905fdf18237548e596819e1831fdb/single_controller/ray/base.py#L261

### DIFF
--- a/single_controller/ray/base.py
+++ b/single_controller/ray/base.py
@@ -205,7 +205,6 @@ class RayWorkerGroup(WorkerGroup):
     def _init_with_detached_workers(self, worker_names):
         workers = [ray.get_actor(name=name) for name in worker_names]
         self._workers = workers
-        self._worker_names = worker_names
         self._world_size = len(worker_names)
 
     def _init_with_resource_pool(self, resource_pool, ray_cls_with_init, bin_pack, detached):
--- a/single_controller/ray/base.py
+++ b/single_controller/ray/base.py
@@ -188,6 +188,7 @@ class RayWorkerGroup(WorkerGroup):
         self.name_prefix = get_random_string(length=6) if name_prefix is None else name_prefix
 
         if worker_names is not None:
+            assert self._is_init_with_detached_workers
             self._worker_names = worker_names
 
         if self._is_init_with_detached_workers:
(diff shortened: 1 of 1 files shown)


kevin85421:
trace the code more. It may be required because the constructor is called inside the same class's method

PeterSH6:
I will check if this will bring some conflicts tomorrow.

Initialize the `_woker_names` in the constructor may cause some confliction to L261 you listed here

kevin85421:
Thanks! @PeterSH6 I will also try to run some examples from my side. Would you mind giving me a pointer about the tests?

PeterSH6:
@kevin85421 Thanks! We didn't open-source the tests as we haven't set up the CI in GitHub recently. I will release some relevant tests later.

single_controller/ray/base.py
```diff
@@ -187,6 +187,9 @@ def __init__(self,
         self.ray_cls_with_init = ray_cls_with_init
         self.name_prefix = get_random_string(length=6) if name_prefix is None else name_prefix
 
+        if worker_names is not None:
+            self._worker_names = worker_names
```
PeterSH6, line 192: I wonder if adding `assert self._is_init_with_detached_workers` between L190-191 would be better?

Your modification will not cause errors in current usage but it may be better to add this line.

PeterSH6 (changes requested):

PeterSH6:
@kevin85421 I have added some tests for Ray single-controller in #23 

These test files would be useful for this PR: 
- [detached worker](https://github.com/volcengine/verl/tree/main/tests/ray/detached_worker)
- [colocate workers](https://github.com/volcengine/verl/blob/main/tests/ray/test_colocated_workers.py)

single_controller/ray/base.py
kevin85421: updated!

kevin85421:
Thank you for the review! I will spend more time on this repo after Thanksgiving. We are currently working on a PoC for veRL with Ray Compiled Graphs.

In addition, we have limited access to A100s. Are there any examples that can be run on a smaller GPU machine so that I can use for dev?

PR #34, opened 2024-12-03
[model] feat: support models without pad_token

eric-haibin-lin:
handle the case when tokenizer.pad_token is None. Also fix an error that BaseShardingManager cannot be found.

### DIFF
--- a/verl/third_party/vllm/vllm_v_0_5_4/tokenizer.py
+++ b/verl/third_party/vllm/vllm_v_0_5_4/tokenizer.py
@@ -70,7 +70,7 @@ class TokenizerGroup:
     # FIXME(sgm): for simplicity, we assign the special token here
     @property
     def pad_token_id(self):
-        return self.tokenizer.pad_token_id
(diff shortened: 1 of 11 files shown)


verl/third_party/vllm/vllm_v_0_5_4/tokenizer.py
```diff
@@ -70,7 +70,7 @@ def get_lora_tokenizer(self, lora_request: Optional[LoRARequest]) -> "PreTrained
     # FIXME(sgm): for simplicity, we assign the special token here
     @property
     def pad_token_id(self):
-        return self.tokenizer.pad_token_id
+        return self.tokenizer.pad_token_id if self.tokenizer.pad_token_id is not None else self.tokenizer.eos_token_id
```
vermouth1992, line 73: Actually, we can simply set pad_token_id to eos_token_id after tokenizer initializes in the trainer. This makes code changes significantly fewer.

eric-haibin-lin:
superceded by https://github.com/volcengine/verl/pull/36

PR #38, opened 2024-12-06
[ci] feat: add more CI workflow

PeterSH6:
- As titled


verl/protocol.py
```diff
@@ -498,6 +498,47 @@ def chunk(self, chunks: int) -> List['DataProto']:
 
         return output
 
+    def unfold_column_chunks(self, n_split, split_keys=None):
```
vermouth1992: Remove this

PeterSH6:
Nice! Finally...

Some TODOs:
- Update and fix the colocated worker name in the log and relevant CI
- Install cupy in L20 using docker.
- Check the connection error in the ignore ray task in L20.

eric-haibin-lin (approved):

eric-haibin-lin:
so this PR run e2e test with the digit completion trainer. We do not have e2e tests covering the verl ppo trainer yet?

vermouth1992:
> so this PR run e2e test with the digit completion trainer. We do not have e2e tests covering the verl ppo trainer yet?

It does cover the verl ppo trainer, but not the main entry. It's better to add QWen 0.5b as a nightly ci later

PR #52, opened 2024-12-16
[BREAKING][refact]: move actor/critic/hybrid_engine/reward_model/rollout/workers out of ppo directory for reuse

PeterSH6:
- A BREAKING PR, need to fix the import
- In current architecture:
  - `verl.workers`: store all the files related to SPMD distributed computation
  - `verl.trainer`: store different RL algorithms, SFT, and generation scripts


verl/trainer/workers/actor/megatron_actor.py
```diff
@@ -38,7 +38,7 @@
 from verl.utils.megatron.pipeline_parallel import (compute_transformers_input_shapes, make_batch_generator)
 from verl import DataProto
 from verl.trainer.ppo import core_algos
-from verl.trainer.ppo.actor import BasePPOActor
+from verl.trainer.workers.actor import BasePPOActor
```
PeterSH6: We may also need to rename 
- BasePPOActor -> BaseActor
- DataParallelPPOActor -> DataParallelActor

Other classes are similar

vermouth1992: Maybe we can simply delete base class

PeterSH6:
```python
from verl.trainer.workers.critic import DataParallelPPOCritic
from verl.trainer.workers.actor.megatron_actor import MegatronPPOActor
```
DDP & Megatron class have the same issue, I think we still need to delete `PPO`. Shall we just make the `PPO` implementation a base class? So that other algorithms can simply override the functions.

eric-haibin-lin (approved):

PR #55, opened 2024-12-18
[ppo] chore: remove unused flash_attn dependency, and add docs for GSM8k reward

eric-haibin-lin:
as title


eric-haibin-lin:
also added citation for @lqtrung1998 @allanj 's work

PR #57, opened 2024-12-19
Add pyproject.toml

pcmoritz:
This makes the project PEP 517 compliant (see https://packaging.python.org/en/latest/guides/modernize-setup-py-project/) which is generally seen as good practice (e.g. it adds build isolation for pip install and makes it more compatible with other package managers).


eric-haibin-lin:
Thanks for the contribution!

PR #58, opened 2024-12-19
rollout: FIRE sampling added.

laonahongchen:
Add FIRE sampling from paper _Flaming-hot Initiation with Regular Execution Sampling for Large Language Models_ (https://arxiv.org/abs/2410.21236) to the PPO trainer loop as an additional option to turn on.


eric-haibin-lin (commented):
Sorry for the late review and thanks for the contribution! Would you mind introducing a new class inheriting vllmRollout, add the custom sampling logics? This way it'd be easier to maintain the codebase

laonahongchen:
Hi, Sorry for the delay as I find it extremely challenging to update the code with the latest version of the verl as many things have changed and I am not able to access the original test environment again.

I have updated the code logic following the suggestion from @eric-haibin-lin above, and also merge the latest commits. Please take another round of review and let me know if there is any further suggestions. Thanks!

verl/workers/fsdp_workers.py
```diff
@@ -298,7 +298,11 @@ def _build_rollout(self):
             rollout_sharding_manager = BaseShardingManager()
             # TODO: a sharding manager that do nothing?
         elif self.config.rollout.name == 'vllm':
-            from verl.workers.rollout.vllm_rollout import vLLMRollout, vllm_mode
+            if self.config.rollout.use_fire_sampling:
```
eric-haibin-lin, line 301: Thank you for the update! I'd suggest making minimal changes and have a test that demonstrate the basic usage. I recommend: 
- add a test with use_fire_sampling=True with fsdp + vllm v0.6.3
- remove the changes to megatron workers 
- remove the fire vllm rollout spmd implementation 

The test can be added similar to https://github.com/volcengine/verl/blob/main/.github/workflows/e2e_digit_completion.yml#L46

laonahongchen: Thank you for the feedback. I am not so familiar with the test system here, but I have made a new commit. Do you mind take a review letting me know if this is the desired format? Thanks!

eric-haibin-lin (commented):
thank u. could you fix the errors reported from CI?

laonahongchen:
Hi, 

Thanks for the info. Again, I am not so familiar with the CI system, so I am wondering is the current status ready to go after my last commit? Or is there anything I still need work on as the only failing check I see seems to be caused by random shutdown from the server.

eric-haibin-lin (approved):
Re-triggered CI. Thanks for the contribution!

PR #74, opened 2025-01-02
Support Mlflow, allow forward to have different batch size, compute more metrics

fzyzcjy:
Hi, this is a tiny PR to

* support mlflow in addition to existing wandb and console
* allow critic forward to have different batch size than critic update (if I understand correctly, usually forward can have larger one)
* expose more metrics, mainly the "timing per token"

Example outputs of mlflow:

image

Example outputs of the new metrics:

image

### DIFF
--- a/verl/trainer/ppo/ray_trainer.py
+++ b/verl/trainer/ppo/ray_trainer.py
@@ -225,6 +225,13 @@ def compute_timing_metrics(batch, timing_raw):
     }
 
 
+@contextmanager
+def _timer(name: str, timing_raw: Dict[str, float]):
+    with Timer(name=name, logger=None) as timer:
+        yield
+    timing_raw[name] = timer.last
+
+
 class RayPPOTrainer(object):
     """
     Note that this trainer runs on the driver process on a single CPU/GPU node.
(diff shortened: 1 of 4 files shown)


verl/workers/fsdp_workers.py
```diff
@@ -577,7 +575,7 @@ def compute_values(self, data: DataProto):
             load_fsdp_param_and_grad(module=self.critic_module,
                                      device_id=torch.cuda.current_device(),
                                      load_grad=self._is_offload_grad)
-        micro_batch_size = self.config.ppo_micro_batch_size
+        micro_batch_size = self.config.forward_micro_batch_size
```
PeterSH6, line 578: I wonder why we need an extra forward_micro_batch_size config?

verl/trainer/ppo/ray_trainer.py
```diff
@@ -536,3 +567,10 @@ def fit(self):
             val_metrics = self._validate()
             pprint(f'Final validation metrics: {val_metrics}')
             logger.log(data=val_metrics, step=global_steps)
+
+
+@contextmanager
+def _timer(name: str, timing_raw: Dict[str, float]):
```
PeterSH6: It would be better to move this function above the `RayPPOTrainer`

verl/workers/fsdp_workers.py
PeterSH6: Sorry I miss your comments above. I think we can adopt larger micro batch size in fwd only computation as it doesn't require activation.

fzyzcjy: Yes that's what I was thinking - fwd only computation can benefit from larger bs.

verl/trainer/ppo/ray_trainer.py
fzyzcjy: Sure

fzyzcjy: Done

PeterSH6:
LGTM!

PR #75, opened 2025-01-02
Tiny refactor and cleanup

fzyzcjy:
This is a very tiny code cleanup containing:

1. Refactor `update_sampling_params`. It originally uses a global variable and update states back and forth, and now uses immutable local variables.
2. Try to avoid mutate config in-place, which may cause a bit of confusion and bug in logic
3. Add `val_before_train: true` in the default config. This is because, otherwise users has to write `+trainer.val_before_train=...`, while all other configs has the style of `trainer.val_before_train=...`

~~Note: This PR starts from https://github.com/volcengine/verl/pull/74, thus the GitHub diff page will be clearer after #74 is merged.~~ (Ignore this line - that PR is already merged)

### DIFF
--- a/verl/trainer/config/ppo_megatron_trainer.yaml
+++ b/verl/trainer/config/ppo_megatron_trainer.yaml
@@ -143,3 +143,4 @@ trainer:
   critic_warmup: 0
   default_hdfs_dir: ~/experiments/gsm8k/ppo/${trainer.experiment_name}
   default_local_dir: checkpoints/${trainer.project_name}/${trainer.experiment_name}
+  val_before_train: True
--- a/verl/trainer/fsdp_sft_trainer.py
+++ b/verl/trainer/fsdp_sft_trainer.py
@@ -20,6 +20,8 @@ TODO(zhangchi.usc1992)
 
 import os
 
+from verl.utils.config import config_normalize_batch_size
+
 os.environ['NCCL_DEBUG'] = 'WARN'
 os.environ['TOKENIZERS_PARALLELISM'] = 'true'
 
@@ -29,7 +31,7 @@ import torch
 import torch.distributed
 from torch import nn, optim
 from torch.distributed.fsdp import FullyShardedDataParallel as FSDP, MixedPrecision, ShardingStrategy, CPUOffload
-from transformers import AutoTokenizer, AutoModelForCausalLM, PreTrainedModel, AutoConfig
+from transformers import AutoModelForCausalLM, PreTrainedModel, AutoConfig
 from verl.utils.torch_functional import get_cosine_schedule_with_warmup
 from tensordict import TensorDict
 from torch.utils.data import DataLoader, DistributedSampler
@@ -83,11 +85,8 @@ class FSDPSFTTrainer(object):
         if self.device_mesh.get_rank() == 0:
             print(f'Normalize batch size by dp {dp_size}')
 
-        assert self.config.data.train_batch_size % dp_size == 0
-        assert self.config.data.micro_batch_size % dp_size == 0
-
-        self.config.data.train_batch_size //= dp_size
-        self.config.data.micro_batch_size //= dp_size
+        config_normalize_batch_size(self.config.data, 'train_batch_size', dp_size)
+        config_normalize_batch_size(self.config.data, 'micro_batch_size', dp_size)
 
     def _build_dataloader(self):
         config = self.config
@@ -118,7 +117,7 @@ class FSDPSFTTrainer(object):
                                                 rank=rank,
                                                 drop_last=True)
         self.train_dataloader = DataLoader(dataset=self.train_dataset,
-                                           batch_size=config.data.train_batch_size,
+                                           batch_size=config.data.train_batch_size_normalized,
                                            sampler=self.train_sampler,
                                            drop_last=True)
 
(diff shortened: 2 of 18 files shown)


verl/trainer/config/ppo_megatron_trainer.yaml
```diff
@@ -143,3 +143,4 @@ trainer:
   critic_warmup: 0
   default_hdfs_dir: ~/experiments/gsm8k/ppo/${trainer.experiment_name}
   default_local_dir: checkpoints/${trainer.project_name}/${trainer.experiment_name}
+  val_before_train: True
```
eric-haibin-lin, line 146: Would you mind also updating the docs in https://github.com/volcengine/verl/blob/main/docs/examples/config.rst for val_before_train, and the forward batch size option added last time?

eric-haibin-lin (commented):
thx for the contribution!

verl/trainer/config/ppo_megatron_trainer.yaml
fzyzcjy: Sure!

fzyzcjy: Done

PeterSH6:
Nice work! I'll review it later today.

fzyzcjy:
> thx for the contribution!

You are welcome!

> Nice work!

Thanks!

> I'll review it later today.

Take your time :)

verl/trainer/fsdp_sft_trainer.py
```diff
@@ -20,6 +20,8 @@
 
 import os
 
+from verl.utils.config import config_normalize_batch_size
```
PeterSH6, line 23: Can you move this import above L35, which would be tidier

verl/utils/config.py
```diff
@@ -14,10 +14,21 @@
...
+from omegaconf import DictConfig, open_dict
 
 
 def update_dict_with_config(dictionary: Dict, config: DictConfig):
     for key in dictionary:
         if hasattr(config, key):
             dictionary[key] = getattr(config, key)
+
+
+def config_normalize_batch_size(config, key: str, divider: int):
```
PeterSH6, line 26: I'm interested to know why in-place normalization might lead to confusion and how you feel when reading and using veRL's batch size configuration.

Additionally, I'd like to hear about any bugs you may have encountered while using in-place modification.

verl/workers/rollout/vllm_rollout/vllm_rollout.py
```diff
@@ -118,26 +118,10 @@ def __init__(self, actor_module: nn.Module, config: DictConfig, tokenizer, model
...
         print(f"kwargs: {kwargs}")
-        self.sampling_params = SamplingParams(**kwargs)
+        self.partial_sampling_params = SamplingParams(**kwargs)
 
         self.pad_token_id = tokenizer.pad_token_id
 
-    @contextmanager
-    def update_sampling_params(self, **kwargs):
-        # update sampling params
-        old_sampling_params_args = {}
```
PeterSH6, line 128: Our original implementation also used a local variable while only maintaining one `SamplingParams`.

It seems that your refactor introduced deepcopy.
Could you clarify the benefit of your refactor more? Thanks!

PeterSH6 (changes requested):

verl/workers/rollout/vllm_rollout/vllm_rollout.py
fzyzcjy: Sure, I personally think the refactor is a little bit better, because we now avoid mutable global state, which seems to be generally not preferable in software engineering (e.g. a bit less readable and error-prone). The deepcopy is slower, but it seems the time cost is negligible, especially compared with the very heavy generate sequences.

Anyway, as the title suggests, this is just a tiny refactor, and feel free to revert it if you do not want :)

verl/utils/config.py
fzyzcjy: As a simple example:

```
def __init__():
  f()
  ...
  normalize_batch_size()
  ...
  g()

def f():
  ...
  f2()

def f2():
  ...
  f3()

def f3():
  utilize config.batch_size

def g():
  utilize config.batch_size
```

Then, inside f3 we are using the un-normalized batch size, while in g we use the normalized one.
 
Surely this is just a made-up artificial case (and is very easy to realize something goes wrong), but in the real world I am worried whether, in the future, we will read batch size somewhere deeply nested, and accidentally it is *before* the normalize_batch_size, and then we are in trouble.

P.S. it is a bit about mutability in software engineering, which can introduce bugs when things unintentionally mutates. For example, in Rust, things are immutable by default.

Again, as is mentioned in the next comment, this is just a tiny refactor and feel free to accept or revert it :)

verl/trainer/fsdp_sft_trainer.py
fzyzcjy: (will do that in the next batch after discussions of other things)

verl/utils/config.py
PeterSH6: Got it. I think your refactor regarding the in-place normalization is great.

However, I wonder how we deal with more complicated config mutate scenarios (such as $\times$ n in PR https://github.com/volcengine/verl/pull/80) if we follow the no-in-place mutate principle.

It would be better to keep only one additional config for each \*_batch_size config and we may need to make the normalize config function more general

fzyzcjy: > we may need to make the normalize config function more general

Totally agree! What about `config_normalize_batch_size(config, 'batch_size', self.config.rollout.n / self.device_mesh.shape[0])`, and at the same time change "divider" to "multiplier"?
(We may not face rounding issues since divider is always power of 2)

eric-haibin-lin:
thanks for the previous efforts! Closing it due to multiple conflicts with the latest version. Feel free to open a new PR for new refactor/cleanup fixes. Thanks!!!

fzyzcjy:
Sure, you are welcome! I will do that later when using verl again

PR #79, opened 2025-01-05
[Pending SGLang] Support torch.compile and allow disabling FSDP

fzyzcjy:
NOTE: This PR is pending SGLang integration, because I observe weird vllm errors like `CUDA error: an illegal memory access was encountered`, and want to debug *after* SGLang is added to easily check whether it is vllm-related bug or not.

Compile is compatible with FSDP but has not done that yet. So this PR firstly does two things:

1. Allow users to do torch.compile (but need to disable FSDP currently)
2. Allow users to disable FSDP 

Btw I see a comment `# TODO: add transformer policy` near FSDP creation, and I delete it because I guess that line means "support using Transformers without FSDP", which is done in this PR.

A quick tests shows memory is saved and time is a bit faster on a 0.5B model. IIRC torch.compile is known to be speed up different models to different extents, so other models may have bigger/smaller speedup accordingly. In addition, for different batch sizes and model sizes, the amount of memory saved can be different. Screenshot below is GPU memory consumption. As can be seen, the highest bar (update policy) becomes lower.

Before:
image

After:
image

Btw, power consumption becomes a bit lower in this experiment as well (left is old code, right is new code):

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ veRL is fast with:
 
 
 <p align="center">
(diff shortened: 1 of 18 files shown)


eric-haibin-lin:
Closing due to lack of activity. Thanks for the PR and please reopen if it's rebased

fzyzcjy:
Sure!

PR #81, opened 2025-01-06
Fix repeat

caoshiyi:
DataProto misses the repeat function.


PeterSH6:
Thanks for your investigation! Sorry that I missed this feature.
Your impl looks good but I'll upload our implementation to keep consistency.
I'll close it first.

PR #83, opened 2025-01-06
[misc] feat: add several useful functions in protocol

PeterSH6:
- Fixed the missing repeat function reported in https://github.com/volcengine/verl/pull/81
- Add several data protocol functions that will be beneficial for future features

### DIFF
--- a/verl/protocol.py
+++ b/verl/protocol.py
@@ -36,6 +36,32 @@ except:
     pass
 
 
+def pad_dataproto_to_divisor(data: 'DataProto', size_divisor: int):
+    """Pad a DataProto to size divisible by size_divisor
+
+    Args:
+        size_divisor (int): size divisor
+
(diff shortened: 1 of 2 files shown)


verl/protocol.py
```diff
@@ -403,6 +470,47 @@ def chunk(self, chunks: int) -> List['DataProto']:
 
         return output
 
+    def unfold_column_chunks(self, n_split, split_keys=None):
```
vermouth1992: Remove this function as it is not very general

vermouth1992:
Add save_to_disk and load_from_disk. Also, add the CI.

PeterSH6:
@vermouth1992 Done, I'll setup the CI when the machine is ready.

We can merge it first to fix the bug in ray_trainer.py

PR #85, opened 2025-01-07
Support megatron 0.6 in veRL

Chendong98:
## Description
I am opening this PR with the hope of adding veRL support to Megatron 0.6 (although I noticed that the veRL paper seems to have already used Megatron 0.6 as the test version). From my naive perspective, I envision two possible approaches:

1. Communication at the parameter level.
2. Creating a MemoryBuffer in veRL that is fully aligned with the ParamAndGradBuffer in Megatron 0.6, and then performing broadcast and other communication operations based on this buffer.

communication-demo

In the current draft, when `self._pp_rank == pp_rank`, it directly uses the buffer defined in Megatron 0.6 (without even checking if use_distributed_optimizer is set), and communicates at the parameter level during parameter synchronization—this, of course, incurs some performance overhead.

At the very least, this approach seems feasible.

#### PPO training deepseek-llm-7b-chat on GSM8K

- convergence

2D14E950-E982-45FA-C7E8-8DDBBE088F10

- performance

724101DD-D581-4D27-95FD-9BEEC5F8E772

actor_gsm

#### PPO training deepseek-coder-6.7b-instruct on GSM8K + MATH

- convergence

D7D5550C-5B66-4B3D-B391-083D8FC19B71

- performance

86A488BF-3D4F-4C0B-E99C-923903FAEC78

actor_math

The curve demonstrates that this pr does not negatively impact the training convergence. 
The performance showes different trends: improvement in one case and degradation in the other. The reasons for these differences require further analysis.

### DIFF
--- a/verl/workers/hybrid_engine/megatron_vllm.py
+++ b/verl/workers/hybrid_engine/megatron_vllm.py
@@ -143,7 +143,7 @@ class AllGatherPPModel:
                     dist.broadcast(tensor=memory_buffer.data, src=global_src, group=self.pp_group, async_op=False)
             else:
                 # a workaround for megatron >= 0.6.0
-                for _, param in sorted(self.pp_models[cur_pp_rank].named_paramters()):
+                for _, param in sorted(self.pp_models[cur_pp_rank].named_parameters()):
                     dist.broadcast(tensor=param.data, src=global_src, group=self.pp_group, async_op=False)
 
     def forward(self, *inputs, **kwargs):
--- a/verl/workers/hybrid_engine/megatron_vllm.py
+++ b/verl/workers/hybrid_engine/megatron_vllm.py
@@ -15,6 +15,8 @@
 This file contains a Megatron style Hybrid Engine that shares the weights of the actor with the inference engine.
 """
 
+import importlib
+from packaging.version import Version
 import torch
 import torch.distributed as dist
 
@@ -30,6 +32,8 @@ from verl.utils.memory_buffer import (
     build_memory_reference_from_module,
     get_weight_buffer_meta_from_module,
 )
+
+megatron_version = megatron_version = Version(importlib.metadata.version('megatron-core'))
 
 
 class AllGatherPPModel:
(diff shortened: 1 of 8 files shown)


PeterSH6:
Another question is whether there are any issues in MCore 0.6? 
If not, we may not need to patch the upstream megatron anymore.

verl/workers/hybrid_engine/megatron_vllm.py
```diff
@@ -121,8 +138,13 @@ def allgather_params(self):
             global_src = dist.get_global_rank(group=self.pp_group, group_rank=cur_pp_rank)
 
             # NOTE(sgm): the async op may cause memory leakage of the memory_buffer/pp_models
-            for memory_buffer in self.memory_buffers[cur_pp_rank].values():
-                dist.broadcast(tensor=memory_buffer.data, src=global_src, group=self.pp_group, async_op=False)
+            if megatron_version < Version('0.6.0'):
+                for memory_buffer in self.memory_buffers[cur_pp_rank].values():
+                    dist.broadcast(tensor=memory_buffer.data, src=global_src, group=self.pp_group, async_op=False)
+            else:
+                # a workaround for megatron >= 0.6.0
```
vermouth1992, line 145: What's the performance if we broadcast tensor one by one? Actually, I feel that it's actually not a big deal since the computation time would dominate the data transfer time.

Chendong98: I tested the performance using the Llama-3.2-1B-Instruct model with a dp=1, tp=1, pp=2 configuration on two setups: two A800 PCIe cards and two H20 NVLink cards. I performed the broadcast 50 times and took the average time for each method.
	•	On the A800 PCIe cards, broadcasting tensors one by one took an average of 0.0769s, while broadcasting by buffer took 0.0666s. 
	•	On the H20 NVLink cards, broadcasting tensors one by one took an average of 0.0049s, while broadcasting by buffer took 0.0041s.

Cppowboy:
When will this feature be merged into the main branch?

verl/workers/hybrid_engine/megatron_vllm.py
```diff
@@ -81,11 +85,24 @@ def __init__(self, model_provider) -> None:
...
-        model = self.pp_models[pp_rank]
-        weight_buffer_meta = get_weight_buffer_meta_from_module(model)
-        self.memory_buffers[pp_rank] = build_memory_buffer(weight_buffer_meta)
+        if megatron_version >= Version('0.6.0') and pp_rank == self._pp_rank:
+            from verl.utils.memory_buffer import MemoryBuffer
+            # The code here is very hard-coded, based on the following assumptions:
+            # 1. `len(_this_rank_models) == 1`
+            # 2. `_this_rank_models[0]` is a instance of `DistributedDataParallel` and `use_distributed_optimizer=True`
+            # 3. Only bfloat16 data type is used in parameters
+            source = self._this_rank_models[0].buffers[0].param_data
```
ShareLer, line 94: If only buffers[0] are fetched here, does that mean that the overlap_grad_reduce parameters of DDP cannot be used? Because when overlap_grad_reduce=True, there may be multiple _ParamAndGradBuffer.

vermouth1992:
Hi @Chendong98, mcore has been upgrade to v0.11 in this PR. https://github.com/volcengine/verl/pull/392. Your contribution is acknowledged. Feel free to contact us if there is anything wrong. Thanks!

PR #88, opened 2025-01-08
[docker] megatron: add TE to ngc dockerfile

eric-haibin-lin:
Image:
`verlai/verl:ngc-th2.4.0-cu124-vllm0.6.3-te1.7-v0.0.4`


PeterSH6:
We can also upgrade ray to the latest version without depending on 2.10

eric-haibin-lin:
> We can also upgrade ray to the latest version without depending on 2.10

This will install ray 2.40

Issue #53, opened 2024-12-16
Do we have plans for data packing?


YixinSong-e:
Thank you for creating this excellent project. The design is elegant and the implementation is very efficient.

I'm wondering if there are plans to support sequence packing? 
OpenRLHF has demonstrated significant speedup after supporting sequence packing. Given VERL's already impressive performance, I believe adding packing support could bring even greater efficiency gains.

PeterSH6:
Hi @YixinSong-e, do you mean data packing in the FSDP backend?
We already implemented data packing (i.e. remove padding?) in the Megatron-LM backend.
For the FSDP backend, we do have some monkey-patch to implement it upon transformers but it's not general for all models in transformers. So we haven't released them yet.
And we also noticed that only some important models (e.g., llama, qwen...) support data packing naturally in transformers.

We're thinking about some elegant way to support data packing for all models in our fsdp workers. Do you have any suggestions or references?

SalmanMohammadi:
Hey @PeterSH6! 

+1 for such an awesome project. My favourite in the new wave of open-source RLHF frameworks.

> For the FSDP backend, we do have some monkey-patch to implement it upon transformers but it's not general for all models in transformers. So we haven't released them yet.

I'm curious what kind of monkey patch you mean here? I'd be interested in contributing if you'd like some help, I'm working on something similar in another repo.

fzyzcjy:
+1 This seems to be widely used and speed up training, and I am happy to PR as well.

PeterSH6:
Hi @SalmanMohammadi and @fzyzcjy , sorry for the late response.

The monkey patch we apply in transformers==4.44 is something like that:

```python
def flash_attn2_rmpad_forward(
        self: LlamaFlashAttention2,
        hidden_states: torch.Tensor,
        attention_mask: Optional[torch.Tensor] = None,
        position_ids: Optional[torch.LongTensor] = None,
        past_key_value: Optional[Cache] = None,
        output_attentions: bool = False,
        use_cache: bool = False,
        cache_position: Optional[torch.LongTensor] = None,
        **kwargs  # for compatibility
):
    hidden_states = hidden_states.squeeze(0)  # (total_nnz, hidden_size)
    # this matches with hidden_states
    indices, cu_seqlens, max_seqlen_in_batch = _get_unpad_data(attention_mask)

    hidden_states = pad_input(hidden_states,
                              indices=indices,
                              batch=attention_mask.shape[0],
[...]

PR #91, opened 2025-01-10
[misc] feat: spport rmpad/data-packing in FSDP with transformers

PeterSH6:
- Use `actor_rollout_ref.model.use_rmpad=True` + `critic.model.use_rmpad=True \` + `reward_model.model.use_rmpad=True ` to enable rmpad for different models. Default set to False
- Using `AutoModelForTokenClassification` for Value and Reward Model. Instead of using SeqenceClassification
- Compute logprob convert to `log_probs_from_logits_response_rmpad`

Resolve: https://github.com/volcengine/verl/issues/53

**Comparison using DeepSeek7b and GSM8k:**
About 1.7x speedup compare to no rmpad (original cases)

### DIFF
--- a/tests/e2e/run_ray_trainer_rmpad.sh
+++ b/tests/e2e/run_ray_trainer_rmpad.sh
@@ -1,0 +1,12 @@
+#!/usr/bin/env bash
+
+set -e -x
+
+python3 tests/e2e/arithmetic_sequence/rl/main_trainer.py \
+    data.train_files=tests/e2e/arithmetic_sequence/data/train.parquet \
+    data.val_files=tests/e2e/arithmetic_sequence/data/test.parquet \
+    actor_rollout_ref.model.path=Qwen/Qwen2.5-0.5B \
+    actor_rollout_ref.model.use_remove_padding=True \
+    critic.model.path=Qwen/Qwen2.5-0.5B \
+    critic.model.use_remove_padding=True \
+    trainer.total_epochs=1 \
--- a/tests/e2e/run_ray_trainer_rmpad.sh
+++ b/tests/e2e/run_ray_trainer_rmpad.sh
@@ -5,7 +5,9 @@ set -e -x
 python3 tests/e2e/arithmetic_sequence/rl/main_trainer.py \
     data.train_files=tests/e2e/arithmetic_sequence/data/train.parquet \
     data.val_files=tests/e2e/arithmetic_sequence/data/test.parquet \
-    actor_rollout_ref.model.path=Qwen/Qwen2.5-0.5B \
+    actor_rollout_ref.model.path=tests/e2e/arithmetic_sequence/model \
+    actor_rollout_ref.rollout.name=vllm \
+    actor_rollout_ref.rollout.tensor_model_parallel_size=1 \
     actor_rollout_ref.model.tokenizer_path=tests/e2e/arithmetic_sequence/model \
     critic.model.path=Qwen/Qwen2.5-0.5B \
     critic.model.use_remove_padding=True \
--- a/tests/e2e/run_ray_trainer_rmpad.sh
+++ b/tests/e2e/run_ray_trainer_rmpad.sh
@@ -6,7 +6,7 @@ python3 tests/e2e/arithmetic_sequence/rl/main_trainer.py \
     data.train_files=tests/e2e/arithmetic_sequence/data/train.parquet \
     data.val_files=tests/e2e/arithmetic_sequence/data/test.parquet \
     actor_rollout_ref.model.path=Qwen/Qwen2.5-0.5B \
-    actor_rollout_ref.model.use_remove_padding=True \
+    actor_rollout_ref.model.tokenizer_path=tests/e2e/arithmetic_sequence/model \
     critic.model.path=Qwen/Qwen2.5-0.5B \
     critic.model.use_remove_padding=True \
-    trainer.total_epochs=1 \
+    trainer.total_epochs=1
--- a/tests/rollout/run_fsdp_vllm.py
+++ b/tests/rollout/run_fsdp_vllm.py
@@ -127,9 +127,8 @@ def main():
     for i in range(batch_size):
         idx_list.append(_pre_process_inputs(pad_token_id, input_ids[i]))
     print('start generation')
-    while True:
-        outputs = llm.generate(prompt_token_ids=idx_list, sampling_params=sampling_params, use_tqdm=False)
-        vllm_output = outputs[0].cuda()
+    outputs = llm.generate(prompt_token_ids=idx_list, sampling_params=sampling_params, use_tqdm=False)
+    vllm_output = outputs[0].cuda()
     if torch.distributed.get_rank() == 0:
         print(f'hf response: {tokenizer.batch_decode(response)}')
         print(f'vllm response: {tokenizer.batch_decode(vllm_output)}')
--- a/tests/rollout/run_fsdp_vllm.py
+++ b/tests/rollout/run_fsdp_vllm.py
@@ -127,8 +127,9 @@ def main():
     for i in range(batch_size):
         idx_list.append(_pre_process_inputs(pad_token_id, input_ids[i]))
     print('start generation')
-    outputs = llm.generate(prompt_token_ids=idx_list, sampling_params=sampling_params, use_tqdm=False)
-    vllm_output = outputs[0].cuda()
+    while True:
+        outputs = llm.generate(prompt_token_ids=idx_list, sampling_params=sampling_params, use_tqdm=False)
+        vllm_output = outputs[0].cuda()
     if torch.distributed.get_rank() == 0:
         print(f'hf response: {tokenizer.batch_decode(response)}')
         print(f'vllm response: {tokenizer.batch_decode(vllm_output)}')
--- a/verl/models/registry.py
+++ b/verl/models/registry.py
@@ -17,6 +17,32 @@ from typing import List, Optional, Type
 
 import torch.nn as nn
 
+# Supported models using HF Rmpad
+# TODO(sgm): HF may supported more than listed here, we should add more after testing
+from transformers import LlamaConfig, MistralConfig, GemmaConfig, Qwen2Config
+
+_RMPAD_MODELS = {
+    'LlamaForCausalLM': LlamaConfig,
+    'MistralForCausalLM': MistralConfig,
+    'GemmaForCausalLM': GemmaConfig,
+    'Qwen2ForCausalLM': Qwen2Config
+}
+
+
+def if_support_rmpad(model_archs: List[str]):
+    if isinstance(model_archs, list):
+        for model_arch in model_archs:
+            if model_arch in _RMPAD_MODELS.keys():
+                return
+        raise ValueError(f"Model architectures {model_archs} are not supported for now. "
+                         f"RMPad supported architectures: {_RMPAD_MODELS.keys()}")
+    elif isinstance(model_archs, str):
+        if not model_archs in _RMPAD_MODELS.keys():
+            raise ValueError(f"Model architecture {model_archs} is not supported for now. "
+                             f"RMPad supported architectures: {_RMPAD_MODELS.keys()}")
+
+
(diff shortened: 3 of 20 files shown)


verl/workers/actor/dp_actor.py
```diff
@@ -41,17 +43,44 @@ def __init__(
...
+                # unpad the position_ids to align the rotary
+                position_ids_rmpad = index_first_axis(rearrange(position_ids.unsqueeze(-1), "b s ... -> (b s) ..."),
+                                                      indices).transpose(0, 1)
+                # only pass input_ids and position_ids to enable flash_attn_varlen
+                output = self.actor_module(input_ids=input_ids_rmpad,
+                                           attention_mask=None,
+                                           position_ids=position_ids_rmpad,
+                                           use_cache=False)  # prevent model thinks we are generating
+                logits_rmpad = output.logits.squeeze(0)  # (total_nnz, vocab_size)
+                logits_rmpad = logits_rmpad / temperature
```
vermouth1992: Change this line to in-place op

vermouth1992:
Shall we add a supported model list and raise error if the model is not in the list?

vermouth1992:
Try to avoid using log_probs_from_logits_response_rmpad because there is an unpad op inside. unpad is a cuda-blocking op. Instead, we can directly use unpad input_ids from the input

PeterSH6:
I think this list depends on transformers lib. No sure where to get this list. I didn't find any doc about the feature in transformers.

vermouth1992:
> I think this list depends on transformers lib. No sure where to get this list. I didn't find any doc about the feature in transformers.

Simply add potential models in the CI. If the model passes CI, then add to the supported list. I guess we can target

PeterSH6:
> Try to avoid using log_probs_from_logits_response_rmpad because there is an unpad op inside. unpad is a cuda-blocking op. Instead, we can directly use unpad input_ids from the input

Sure, I will write a new API for unpad input_ids

PeterSH6:
> Simply add potential models in the CI. If the model passes CI, then add to the supported list. I guess we can target

Shall we add the test_transformers.py to CI? I didn't do it as I think it only depends on the transformers version and flash_attn version.

So, I guess the goal for the CI is to test whether the latest transformers + flash_attn would break our implementation

vermouth1992:
> > Simply add potential models in the CI. If the model passes CI, then add to the supported list. I guess we can target
> 
> Shall we add the test_transformers.py to CI? I didn't do it as I think it only depends on the transformers version and flash_attn version.
> 
> So, I guess the goal for the CI is to test whether the latest transformers + flash_attn would break our implementation

After this PR, we should set a minimum version of transformers

verl/utils/torch_functional.py
```diff
@@ -325,6 +325,29 @@ def log_probs_from_logits_response_rmpad(input_ids, attention_mask, logits_rmpad
...
     return output
 
+def log_probs_from_logits_all_rmpad(input_ids_rmpad, logits_rmpad, indices, batch_size, seqlen, response_length):
+    """Compute the log_probs from logits with rmpad input_ids and logits. Note that
+    logits_rmpad = model(input_ids_rmpad). For each sentences, there is a shift between
+    logits and input_ids.
+    The reason for this function to is to compute logprobs_from_logits in rmpad mode because it is memory-intensive
+    for large vocab_size
+    
+    Args:
```
vermouth1992, line 336: The args here is incorrect

PeterSH6: solved

verl/workers/fsdp_workers.py
```diff
@@ -119,6 +120,10 @@ def _build_model_optimizer(self,
         # override model kwargs
         actor_model_config = AutoConfig.from_pretrained(local_path, trust_remote_code=trust_remote_code)
 
+        if use_rmpad:
+            from verl.models.registry import if_support_rmpad
```
vermouth1992: The name if_support_rmpad is mislead, how about check_model_support_rmpad?

verl/workers/fsdp_workers.py
```diff
@@ -707,12 +729,39 @@ def init_model(self):
...
+                reward_rmpad = output.logits
+                reward_rmpad = reward_rmpad.squeeze(0)  # (total_nnz)
+
+                # pad it back
+                rm_score = pad_input(reward_rmpad, indices=indices, batch=batch, seqlen=seqlen).squeeze(-1)
+            else:
+                output = self.reward_module(input_ids=input_ids,
+                                            attention_mask=attention_mask,
+                                            position_ids=position_ids)
+                rm_score = output.logits  # (batch_size, seq_len, 2)
```
vermouth1992: The shape annotation should be # (batch_size, seq_len, 1)?

verl/workers/fsdp_workers.py
PeterSH6: Sure

verl/workers/fsdp_workers.py
PeterSH6: solved

verl/models/registry.py
```diff
@@ -17,6 +17,32 @@
 
 import torch.nn as nn
 
+# Supported models using HF Rmpad
+# TODO(sgm): HF may supported more than listed here, we should add more after testing
+from transformers import LlamaConfig, MistralConfig, GemmaConfig, Qwen2Config
+
+_RMPAD_MODELS = {
```
vermouth1992: This may cause issue for reward model and value model because they are xxxForTokenClassification?

vermouth1992: Shall we add a e2e CI that guarantees runnable? Just use QWen2.5-0.5b to run one step

vermouth1992: Make sure all the features are toggled

eric-haibin-lin:
shall we rename the option to sequence_packing or remove_padding, the name rmpad is not straightforward for others

verl/models/registry.py
PeterSH6: > This may cause issue for reward model and value model because they are xxxForTokenClassification?

This would be fine as the `config.architectures` field from the AutoConfig is always xxxForCausal.
But it's true that using architectures is not that straightforward. I'll use model_type instead

PeterSH6:
> shall we rename the option to sequence_packing or remove_padding, the name rmpad is not straightforward for others

Nice suggestion

verl/models/registry.py
PeterSH6: > Shall we add a e2e CI that guarantees runnable? Just use QWen2.5-0.5b to run one step

The ray_trainer may not have an option to run one step. One epoch is feasible.

tests/rollout/run_fsdp_vllm.py
```diff
@@ -127,8 +127,9 @@ def main():
     for i in range(batch_size):
         idx_list.append(_pre_process_inputs(pad_token_id, input_ids[i]))
     print('start generation')
-    outputs = llm.generate(prompt_token_ids=idx_list, sampling_params=sampling_params, use_tqdm=False)
-    vllm_output = outputs[0].cuda()
+    while True:
```
eric-haibin-lin: why put it in a forever loop?

verl/workers/actor/dp_actor.py
```diff
@@ -41,17 +43,47 @@ def __init__(
...
-                                       use_cache=False)  # prevent model thinks we are generating
-            logits = output.logits / temperature
-            logits = logits[:, -response_length - 1:-1]
-            log_probs = logprobs_from_logits(logits, micro_batch['responses'])
+            input_ids = micro_batch['input_ids']
+            batch_size, seqlen = input_ids.shape
+            attention_mask = micro_batch['attention_mask']
+            position_ids = micro_batch['position_ids']
+
+            if self.use_remove_padding:
```
eric-haibin-lin, line 57: @PeterSH6 @vermouth1992 I am thinking if it's cleaner if the remove padding logit is also done in the monkey patch way, where the actor_module forward is patched and users always pass in the original input ids? 

Something like 
```
llama_forward_fn = LlamaModelForCausalLM.forward

def forward_remove_padding(...):
    input_rmpad = fn(input_ids)
    llama_forward_fn(xx_rmpad)

LlamaModelForCausalLM.forward = forward_remove_padding
```

tests/rollout/run_fsdp_vllm.py
PeterSH6: Sorry, a wrong commit

verl/workers/actor/dp_actor.py
PeterSH6: In my opinion, currently users can only need to pass the original input_ids to the data-proto. 
And we just need to unpad it before the model fwd.

If we use monkey-patch, we may need to maintain patches for all models.

vermouth1992: I agree that we should explicitly expose the logics of data packing so that users can easily modify it.

vermouth1992 (approved):

YixinSong-e:
input_ids_rmpad, indices, cu_seqlens, max_seqlen_in_batch = unpad_input(
ValueError: too many values to unpack (expected 4)
When using hh-rlhf with rmpadding=True

vermouth1992:
Could you try flash-attn < 2.7? I guess flash-attn 2.7 changes the API. @PeterSH6 We should either enforce the version in setup.py or adapt the API for flash-attn >= 2.7

vermouth1992:
Or we can move the unpad logic into verl in case future breaks.

YixinSong-e:
> Or we can move the unpad logic into verl in case future breaks.

Thanks for your help! :) 
My flash_attn version is 2.7.2. I will try to downgrade it.

PeterSH6:
I think we should deal with both cases: greater than 2.7 and lower.

A utility function would be enough for the two cases

YixinSong-e:
from verl.utils.torch_functional import prepare_input_for_rmpad
ImportError: cannot import name 'prepare_input_for_rmpad' from 'verl.utils.torch_functional' (/verl_0111/verl/utils/torch_functional.py)

I find that thie function is not used. And not in the codebase. Maybe should just delete this line?

vermouth1992:
Sure, simply delete it. Sorry for the inconvenience

YixinSong-e:
It's OK. And there is another error I need help.
  File "/home/verl_0111/verl/single_controller/ray/base.py", line 399, in func
    return getattr(self.worker_dict[key], name)(*args, **kwargs)
  File "/home/verl_0111/verl/single_controller/base/decorator.py", line 404, in inner
    return func(*args, **kwargs)
  File "/home/verl_0111/verl/workers/fsdp_workers.py", line 853, in compute_rm_score
    token_level_scores = self._expand_to_token_level(data, scores)
  File "/home/verl_0111/verl/workers/fsdp_workers.py", line 775, in _expand_to_token_level
    token_level_scores[torch.arange(batch_size), eos_mask_idx] = scores
RuntimeError: shape mismatch: value tensor of shape [128, 2048] cannot be broadcast to indexing result of shape [128]

tests/e2e/run_ray_trainer_rmpad.sh
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env bash
+
+set -e -x
+
+python3 tests/e2e/arithmetic_sequence/rl/main_trainer.py \
```
vermouth1992, line 5: Why this CI only contains critic rmpad but not the actor?

vermouth1992:
You are using an actual reward model?

vermouth1992:
Could you toggle off reward model remove-padding for now? I guess we lack CI for model-based reward function.

YixinSong-e:
OK. Thanks~~~

vermouth1992:
Also, I guess you have to change AutoModelForTokenClassification to AutoModelForSequenceClassification in order to make non-rmpad version work. We will fix this as soon as possible and add a CI for this case.

tests/e2e/run_ray_trainer_rmpad.sh
PeterSH6: This is due to the misalignment between digit completion task  and the Qwen model. It may have some bug when using CharTokenizer

PR #99, opened 2025-01-12
[misc] fix reward model issue with TokenClassification model and support running particular steps instead of epochs

PeterSH6:
This PR solves:

- Add a CI with model-based reward to ensure it's always runnable
- Support running for particular steps instead of epochs
- Fix reward model chooses reward on EOS token after switching from AutoModelForSequenceClassification to AutoModelForTokenClassification

The flash_attention compatible issue will be fixed in the next PR.
I may need some time to check how to install different versions of flash_attn more efficiently in the CI machines.


verl/trainer/ppo/ray_trainer.py
```diff
@@ -558,26 +564,35 @@ def fit(self):
...
+                if self.global_steps >= self.total_training_steps:
+
+                    # perform validation after training
+                    if self.val_reward_fn is not None:
+                        val_metrics = self._validate()
+                        pprint(f'Final validation metrics: {val_metrics}')
+                        logger.log(data=val_metrics, step=self.global_steps)
+                    return
 
         # perform validation after training
```
vermouth1992: Do we need this after we use total_training_steps?

PeterSH6: This can be deleted

vermouth1992 (approved):

PR #121, opened 2025-01-21
[perf] feat: support ref/rm offload

vermouth1992:
- Force ref/rm to use CPUOffload. Fix root FSDP unit not reshard weights after forward
- HSDP support is on hold and assert False right now.


verl/workers/fsdp_workers.py
```diff
@@ -838,8 +869,9 @@ def _build_model(self, config):
             device_id=torch.cuda.current_device(),
             sharding_strategy=ShardingStrategy.FULL_SHARD,  # zero3
```
PeterSH6: We should use HybridShard when the device mesh is 2D

PeterSH6 (approved):

PR #123, opened 2025-01-23
[perf] feat: support meta device init and parallel load for fsdp 

zhiqi-0:
This PR supports:
- meta device init (which keeps the shared parameters)
- parallel pre-trained weight init for FSDP from huggingface checkpoint

### DIFF
--- a/verl/utils/fsdp_utils.py
+++ b/verl/utils/fsdp_utils.py
@@ -12,10 +12,18 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from typing import Dict
 import functools
+import json
+import math
+import itertools
+import os
+from contextlib import contextmanager
 from torch.distributed.fsdp.wrap import size_based_auto_wrap_policy, transformer_auto_wrap_policy
 from transformers.trainer_pt_utils import get_module_class_from_name
 import torch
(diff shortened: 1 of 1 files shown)


verl/utils/fsdp_utils.py
```diff
@@ -120,3 +128,158 @@ def load_fsdp_optimizer(optimizer, device_id):
...
+    Huggingface checkpoint contains:
+    - config.json: a json file for model configuration
+    - model.safetensor.index.json: a json file for safetensors (parameters & buffers) index
+    - model-000x-of-ooxx.safetensors: a binary file for safetensors (parameters & buffers) chunks
+
+    Each rank will own a part of model chunks and load them directly into GPU memory.
+    """
+    from safetensors.torch import load_file
+
+    index_file = os.path.join(filepath, "model.safetensors.index.json")
```
vermouth1992: How to handle the case, where the model is very small and there is no model.safetensors.index.json such as https://huggingface.co/Qwen/Qwen2.5-0.5B-Instruct/tree/main?

zhiqi-0: updated now to load the whole model state on rank0 and broadcast to all other GPUs

PR #124, opened 2025-01-23
[algo] feat: support GRPO algorithm

PeterSH6:
- Implement KL loss, GRPO outcome adv, and utilize bon rollouts
- Provide scripts for deepseek and qwen on GSM8k. Can provide more for other datasets.
- Support seq balance


verl/trainer/ppo/ray_trainer.py
```diff
@@ -112,23 +112,36 @@ def apply_kl_penalty(data: DataProto, kl_ctrl: core_algos.AdaptiveKLController,
...
+        token_level_rewards = data.batch['token_level_rewards']
         advantages, returns = core_algos.compute_gae_advantage_return(token_level_rewards=token_level_rewards,
                                                                       values=values,
                                                                       eos_mask=response_mask,
                                                                       gamma=gamma,
                                                                       lam=lam)
         data.batch['advantages'] = advantages
         data.batch['returns'] = returns
+    elif adv_estimator == 'grpo':
+        assert data.batch.batch_size[0] % num_repeat == 0, 'batch_size must be divisible by num_repeat'
```
vermouth1992: Why do we need to have this constraints? Can we support scenario where the bon of each sample is different?

PeterSH6: I found that we can use different sampling_params for different prompts. So we can support different bon for each prompt by modifying the vllm_rollout.py. Reference: https://github.com/vllm-project/vllm/pull/3570/files

I'll delete this line to support such usage.

verl/trainer/ppo/core_algos.py
```diff
@@ -106,6 +107,54 @@ def compute_gae_advantage_return(token_level_rewards: torch.Tensor, values: torc
     return advantages, returns
 
 
+# NOTE(sgm): this implementation only consider outcome supervision, where the reward is a scalar.
+def compute_grpo_outcome_advantage(token_level_rewards: torch.Tensor,
```
vermouth1992, line 111: We should give each prompt a uuid and use the uuid for grouping

PeterSH6: Yes, index is the uuid. I pass the uuid through the `index` param.

vermouth1992 (approved):

Issue #97, opened 2025-01-12
Fused CE loss integration


eric-haibin-lin:
Integrate it with main stream models: https://github.com/apple/ml-cross-entropy so that model with large vocab size uses much less memory

hongpeng-guo:
I can take a look of this one if it's still available.

Kirrito-k423:
Thank you for this feature request! The Apple ML-CrossEntropy library (https://github.com/apple/ml-cross-entropy) provides a memory-efficient fused cross-entropy loss implementation that is particularly beneficial for models with large vocabulary sizes.

Integrating this would involve:

1. **Dependency**: Add `ml-cross-entropy` as an optional dependency
2. **Loss backend abstraction**: Create an abstraction layer in the GRPO/PPO trainer to switch between the standard PyTorch cross-entropy and the fused implementation
3. **Memory profiling**: Benchmark the memory savings, especially for models with vocab sizes > 100k
4. **Gradient compatibility**: Ensure the fused loss is compatible with the backward pass used in the RL training loop

A good approach would be to make this a configurable option (e.g., `actor.loss.fused_ce=True`) so users can opt in based on their vocabulary size and memory constraints.

Would you be interested in contributing a PR for this? If so, the existing loss computation in `verl/trainer/ppo/core_algos.py` would be the starting point.

Kirrito-k423:
English:
Thanks for the feature request. A related large-vocabulary memory path now exists in Verl, but it is not the Apple `ml-cross-entropy` integration requested in this issue.

Evidence:
- #462 added a merged "Memory efficient cross entropy with a linear layer fused" implementation: https://github.com/verl-project/verl/pull/462
- The current fused implementation is exposed as `linear_cross_entropy` in `verl/utils/kernel/linear_cross_entropy.py`: https://github.com/verl-project/verl/blob/main/verl/utils/kernel/linear_cross_entropy.py#L38-L119
- The Triton model backend calls that function when fused kernels are enabled: https://github.com/verl-project/verl/blob/main/verl/models/transformers/dense_common.py#L157-L195
- The performance docs state that Liger's `fused_linear_cross_entropy` is disabled because Verl computes log-probabilities through its own path, and recommend combining `use_liger` with `use_fused_kernels` for the speed/memory tradeoff: https://github.com/verl-project/verl/blob/main/docs/perf/perf_tuning.rst#L170-L184

[...]

Issue #96, opened 2025-01-12
Liger kernel integration


eric-haibin-lin:
Integrate liger kernel for main stream models such as Qwen, llama.
Help wanted

PeterSH6:
A quick comment: I think we should be really careful about adding features that require patches and more package dependencies.

As their optimizations may not be applicable for all transformers models and would increase maintaining workload.

hongpeng-guo:
cc @ByronHsu

deter3:
> A quick comment: I think we should be really careful about adding features that require patches and more package dependencies.
> 
> As their optimizations may not be applicable for all transformers models and would increase maintaining workload.

There is big wave of using RL for LLMs after deepseek R1 . If there is no Liger kernel integration , veRL will be facing huge GPU memory requirements and training cost . You can either catch this wave or play safe .

PeterSH6:
> > A quick comment: I think we should be really careful about adding features that require patches and more package dependencies.
> > 
> > As their optimizations may not be applicable for all transformers models and would increase maintaining workload.
> 
> There is big wave of using RL for LLMs after deepseek R1 . If there is no Liger kernel integration , veRL will be facing huge GPU memory requirements and training cost . You can either catch this wave or play safe . 

Cannot agree more. We'll support it soon.

vermouth1992:
Already implemented in PPOTrainer and SFTTrainer

PR #133, opened 2025-01-26
[Liger-kernel] Add an option to use `_apply_liger_kernel_to_instance()` to load model

hongpeng-guo:
## Summary

This PR enables to use Liger Kernel's `_apply_liger_kernel_to_instance` to init a fsdp worker model.

## Main Changes

1. Adding an option of using `liger_kernel.transformers.AutoLigerKernelForCausalLM` to load a model from pretained, instead of the default `transformers.AutoModelForCausalLM`
2. Added a test case using configuration file `tests/e2e/run_qwen_gsm8k_model_rm_liger_kernel.sh`

## Related Issue

#96 

## TODO

#97 optimize the memory usage when computing entropy & log_probs
https://github.com/volcengine/verl/blob/6d96fda3d47f057caaa8f494ca7804181903e911/verl/workers/actor/dp_actor.py#L94-L106

### DIFF
--- a/tests/e2e/run_qwen_gsm8k_model_rm_liger_kernel.sh
+++ b/tests/e2e/run_qwen_gsm8k_model_rm_liger_kernel.sh
@@ -1,0 +1,52 @@
+set -x
+
+export VLLM_ATTENTION_BACKEND=XFORMERS
+
+python3 -m verl.trainer.main_ppo \
+    data.train_files=$HOME/data/gsm8k/train.parquet \
+    data.val_files=$HOME/data/gsm8k/test.parquet \
+    data.train_batch_size=1024 \
+    data.val_batch_size=1312 \
+    data.max_prompt_length=512 \
+    data.max_response_length=512 \
+    data.return_raw_chat=True \
+    actor_rollout_ref.model.path=Qwen/Qwen2.5-0.5B \
+    actor_rollout_ref.actor.optim.lr=1e-6 \
+    actor_rollout_ref.model.use_remove_padding=True \
+    actor_rollout_ref.model.use_liger=True \
+    actor_rollout_ref.actor.optim.lr_warmup_steps_ratio=0.1 \
+    actor_rollout_ref.actor.ppo_mini_batch_size=256 \
+    actor_rollout_ref.actor.ppo_micro_batch_size=32 \
+    actor_rollout_ref.actor.fsdp_config.param_offload=False \
+    actor_rollout_ref.actor.fsdp_config.grad_offload=False \
+    actor_rollout_ref.actor.fsdp_config.optimizer_offload=False \
+    actor_rollout_ref.rollout.log_prob_micro_batch_size=128 \
+    actor_rollout_ref.rollout.tensor_model_parallel_size=2 \
+    actor_rollout_ref.rollout.name=vllm \
+    actor_rollout_ref.rollout.gpu_memory_utilization=0.4 \
+    actor_rollout_ref.ref.log_prob_micro_batch_size=128 \
+    actor_rollout_ref.ref.fsdp_config.param_offload=True \
+    critic.optim.lr=1e-5 \
+    critic.model.use_remove_padding=True \
+    critic.optim.lr_warmup_steps_ratio=0.05 \
+    critic.model.path=Qwen/Qwen2.5-0.5B \
+    critic.model.enable_gradient_checkpointing=False \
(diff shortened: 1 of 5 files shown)


verl/workers/fsdp_workers.py
```diff
@@ -193,6 +194,17 @@ def _build_model_optimizer(self,
...
                                                                 trust_remote_code=trust_remote_code)
+            # Apply Liger kernel to the model if use_liger is set to True
+            if use_liger:
+                try:
+                    # Import Liger kernel module and use it to load the model
+                    from liger_kernel.transformers.monkey_patch import _apply_liger_kernel_to_instance
+                    _apply_liger_kernel_to_instance(model=actor_module)
+                except ImportError:
+                    # Fallback to use AutoModelForCausalLM and print warning message
+                    logger.warning("Liger kernel was requested but not installed - falling back to AutoModelForCausalLM")
```
corbt: Personally I would prefer for the job to fail outright if liger is requested but not installed. Just printing a warning is too easy to miss in all the job outputs.

hongpeng-guo: I like your proposal. I think we can remove the `try ... catch ...` logic and add `liger_kernel` a required dependency of verl. There are still a few extra PRs in Liger kernel targeting for the full integration. Once those are done, I think it makes sense to add `liger-kernel` to the `requirements.txt` cc @vermouth1992 @eric-haibin-lin

vermouth1992: I guess it's a good idea to make liger-kernel a required dependency and remove try except.

tests/e2e/run_qwen_gsm8k_model_rm_liger_kernel.sh
```diff
@@ -0,0 +1,52 @@
+set -x
```
vermouth1992, line 1: Could you add this test to CI and make sure liger-kernel is toggled?

tests/e2e/run_qwen_gsm8k_model_rm_liger_kernel.sh
```diff
@@ -0,0 +1,52 @@
...
+    data.val_files=$HOME/data/gsm8k/test.parquet \
+    data.train_batch_size=1024 \
+    data.val_batch_size=1312 \
+    data.max_prompt_length=512 \
+    data.max_response_length=512 \
+    data.return_raw_chat=True \
+    actor_rollout_ref.model.path=Qwen/Qwen2.5-0.5B \
+    actor_rollout_ref.actor.optim.lr=1e-6 \
+    actor_rollout_ref.model.use_remove_padding=True \
+    +actor_rollout_ref.model.use_liger=True \
```
hongpeng-guo, line 16: The flag for `use_liger` is here.

tests/e2e/run_qwen_gsm8k_model_rm_liger_kernel.sh
hongpeng-guo: Sure, added this test to `.github/workflows/e2e_gsm8k.yml`

hongpeng-guo (commented):
A relevant PR for the fused kernel of `ce_loss` and `entropy` as described in TODO can be referenced here: https://github.com/linkedin/Liger-Kernel/pull/551

verl/workers/fsdp_workers.py
vermouth1992: Could you remove try except and run the CI again?

tests/e2e/run_qwen_gsm8k_model_rm_liger_kernel.sh
vermouth1992: I guess liger-kernel has to be added to setup.py and project.toml

vermouth1992:
Please run the formatter `bash script/format.sh`

hongpeng-guo:
> Please run the formatter `bash script/format.sh`

Got it! Will handle the comments soon. One suggestion is to have a `Contributor's Guid` section on the readme, or in the doc, i.e., https://docs.sglang.ai/references/contribution_guide.html The contributors would know the checks to run before pushing a PR. Setting up some `pre-commit` hooks will be better :)

verl/workers/fsdp_workers.py
hongpeng-guo: Sure, done.

vermouth1992 (approved):

PR #135, opened 2025-01-26
[misc] fix nan in non_tensor_batch union

PeterSH6:
- In some cases, there would be a nan value in the np.array.
- We fix it using pandas as np fail to assert array when the type is object


vermouth1992:
Could you add a test case?

vermouth1992 (approved):

PR #136, opened 2025-01-26
[BREAKING][misc] feat: change micro_batch_size to micro_batch_size_per_gpu

PeterSH6:
## Summary

This PR changes all the micro_batch_size to micro_batch_size_per_gpu.

**The Core logic of setting batch size:**
- **All algorithmic metrics** (train batch size, ppo mini batch size): are global (from the perspective of single-controller), which will be normalized in each Worker.
- **All performance-related parameters** (micro batch size, max token length in dynamic batch size) are local parameters, which represent the data sizes per GPU (i.e., each Worker).

## Main Changes

1. Change the scripts and config and delete the normalization for micro_bsz
2. Fix CI for SFT


vermouth1992 (commented):
It's not a good idea to break user's API. We should introduce per gpu bsz in addition to the original parameter. When both of them present, we should assert false. When the old parameter is used, we should print a warning.

PeterSH6:
Sure. We should set the default value of micro_batch_size_per_gpu to null to implement that.

Moreover, for the examples I provide, I think we should simply use micro_batch_size_per_gpu but not the original micro_batch_size.

vermouth1992 (approved):

PR #141, opened 2025-01-27
[misc] fix: gradient accumulation in seq balance and modify default vllm log level

PeterSH6:
- Previous gradient accumulation value is computed by micro_batch_size, which is wrong when using dynamic_bsz
- Fix ci script to avoid overlooking this issue
- Change vLLM state log default value to True to disable log.
- We will check the `self.config.actor.ppo_mini_batch_size % self.config.actor.ppo_micro_batch_size_per_gpu == 0` after normalization in fsdp_workers instead of in dp_actor and dp_critic.

### DIFF
--- a/verl/workers/actor/dp_actor.py
+++ b/verl/workers/actor/dp_actor.py
@@ -268,7 +268,11 @@ class DataParallelPPOActor(BasePPOActor):
                     metrics['actor/kl_loss'] = kl_loss.detach().item()
                     metrics['actor/kl_coef'] = self.config.kl_loss_coef
 
-                loss = policy_loss / self.gradient_accumulation
+                if self.config.use_dynamic_bsz:
+                    # relative to the dynamic bsz
+                    loss = policy_loss * (len(data) / self.config.ppo_mini_batch_size)
+                else:
+                    loss = policy_loss / self.gradient_accumulation
                 loss.backward()
 
                 data = {
--- a/verl/workers/actor/dp_actor.py
+++ b/verl/workers/actor/dp_actor.py
@@ -222,7 +222,6 @@ class DataParallelPPOActor(BasePPOActor):
             if self.config.use_dynamic_bsz:
                 max_token_len = self.config.ppo_max_token_len_per_gpu * self.ulysses_sequence_parallel_size
                 micro_batches, _ = rearrange_micro_batches(batch=mini_batch, max_token_len=max_token_len)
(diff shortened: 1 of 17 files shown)


verl/workers/actor/dp_actor.py
```diff
@@ -224,7 +222,9 @@ def update_policy(self, data: DataProto):
             if self.config.use_dynamic_bsz:
                 max_token_len = self.config.ppo_max_token_len_per_gpu * self.ulysses_sequence_parallel_size
                 micro_batches, _ = rearrange_micro_batches(batch=mini_batch, max_token_len=max_token_len)
+                self.gradient_accumulation = len(micro_batches)
```
vermouth1992: Shall we delete this line?

verl/workers/critic/dp_critic.py
```diff
@@ -160,8 +157,10 @@ def update_critic(self, data: DataProto):
             if self.config.use_dynamic_bsz:
                 max_token_len = self.config.ppo_max_token_len_per_gpu * self.ulysses_sequence_parallel_size
                 micro_batches, _ = rearrange_micro_batches(batch=mini_batch, max_token_len=max_token_len)
+                self.gradient_accumulation = len(micro_batches)
```
vermouth1992: Delete this line

verl/workers/fsdp_workers.py
```diff
@@ -125,6 +125,7 @@ def __init__(self, config: DictConfig, role: str):
                 self.config.actor.ppo_micro_batch_size //= (self.device_mesh.shape[0] //
                                                             self.ulysses_sequence_parallel_size)
                 self.config.actor.ppo_micro_batch_size_per_gpu = self.config.actor.ppo_micro_batch_size
+                assert self.config.actor.ppo_mini_batch_size % self.config.actor.ppo_micro_batch_size_per_gpu == 0
```
vermouth1992, line 128: We also have to encore self.config.actor.ppo_mini_batch_size >= n_gpus * self.config.actor.ppo_micro_batch_size_per_gpu

PeterSH6: Is it necessary? The mini_batch_size here is already normalized, then if self.config.actor.ppo_mini_batch_size < n_gpus * self.config.actor.ppo_micro_batch_size_per_gpu, the above line will not get 0

vermouth1992 (approved):

Issue #149, opened 2025-01-28
Test data padding does not match world_size?


chujiezheng:
I encountered a problem that the padded test data here

https://github.com/volcengine/verl/blob/9fca71d2ca55a83e05f33662846e0ac99269be7e/verl/trainer/ppo/ray_trainer.py#L418-L419

cannot be dispatched properly (`AssertionError: only support equal chunk.`)

https://github.com/volcengine/verl/blob/9fca71d2ca55a83e05f33662846e0ac99269be7e/verl/single_controller/base/decorator.py#L275-L276

So I wonder whether the divisor `self.actor_rollout_wg.world_size` is the same as the chunk number `worker_group.world_size`?

PR #150, opened 2025-01-28
[misc]fix: pad dataproto when pad size is larger than len(dataproto)

PeterSH6:
- As titled
- Solved: #149 

Waiting for testing from @chujiezheng


verl/protocol.py
```diff
@@ -51,7 +51,13 @@ def pad_dataproto_to_divisor(data: 'DataProto', size_divisor: int):
     assert isinstance(data, DataProto), 'data must be a DataProto'
```
vermouth1992, line 51: Shall we create a test case?

chujiezheng:
This fix works for me

PR #152, opened 2025-01-28
fix: redundant non_zero_mask

emergenz:
`token_level_rewards == (token_level_rewards * non_zero_mask)`

### DIFF
--- a/verl/trainer/ppo/core_algos.py
+++ b/verl/trainer/ppo/core_algos.py
@@ -128,8 +128,7 @@ def compute_grpo_outcome_advantage(token_level_rewards: torch.Tensor,
             shape: (bs, response_length)
     """
     response_length = token_level_rewards.shape[-1]
-    non_zero_mask = (token_level_rewards != 0)
-    scores = (token_level_rewards * non_zero_mask).sum(dim=-1)
+    scores = token_level_rewards.sum(dim=-1)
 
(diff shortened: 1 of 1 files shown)


verl/trainer/ppo/core_algos.py
```diff
@@ -128,8 +128,7 @@ def compute_grpo_outcome_advantage(token_level_rewards: torch.Tensor,
             shape: (bs, response_length)
     """
     response_length = token_level_rewards.shape[-1]
-    non_zero_mask = (token_level_rewards != 0)
-    scores = (token_level_rewards * non_zero_mask).sum(dim=-1)
+    scores = token_level_rewards.sum(dim=-1)
```
vermouth1992, line 131: @PeterSH6 Could you help check this?

PeterSH6: Sure.

PeterSH6: I think in our codebase, the non_zero_mask is redundant as our verifier will only return a scalar output at the position of eos_id.

But if a verifier would produce more than one reward value on multiple postitions of the sequence, even after the position of eos_id. Then it's necessary to apply the eos_id. Adding this eos_mask could make the computation more robust but I'm not sure if this scenario really exists.

PR #153, opened 2025-01-28
validate `use_remove_padding` when applying sequence parallelism

chujiezheng:
This is because the ulysses_sp is activated only when `use_remove_padding` is enabled:

https://github.com/volcengine/verl/blob/ab525bce267e4b3805ceabac8441afc4385bba96/verl/workers/actor/dp_actor.py#L71-L128

Without this check, users may encounter OOM issues when the set sp_size > 1 but `use_remove_padding` is mistakenly disabled.

### DIFF
--- a/verl/utils/config.py
+++ b/verl/utils/config.py
@@ -86,4 +86,13 @@ def validate_config(config):
             assert config.critic.ppo_mini_batch_size % config.critic.ppo_micro_batch_size == 0
             assert config.critic.ppo_micro_batch_size * sp_size >= n_gpus
 
+    # Check if use_remove_padding is enabled when using sequence parallelism
+    if config.actor_rollout_ref.actor.ulysses_sequence_parallel_size > 1:
+        assert config.actor_rollout_ref.actor.use_remove_padding, \
+            "When using sequence parallelism for actor, you must enable `use_remove_padding`."
+    
+    if config.actor_rollout_ref.ref.ulysses_sequence_parallel_size > 1:
+        assert config.actor_rollout_ref.ref.use_remove_padding, \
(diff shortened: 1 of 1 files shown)


vermouth1992 (approved):

vermouth1992:
Could you format the code by running `bash script/format.sh`

chujiezheng:
Done!

verl/utils/config.py
```diff
@@ -86,4 +86,13 @@ def check_mutually_exclusive(mbs, mbs_per_gpu, name: str):
             assert config.critic.ppo_mini_batch_size % config.critic.ppo_micro_batch_size == 0
             assert config.critic.ppo_micro_batch_size * sp_size >= n_gpus
 
+    # Check if use_remove_padding is enabled when using sequence parallelism
+    if config.actor_rollout_ref.actor.ulysses_sequence_parallel_size > 1:
```
vermouth1992: It seems that the correct key is config. actor_rollout_ref.model.use_remove_padding, critic.model.use_remove_padding and reward_model.model.use_remove_padding

chujiezheng: Fixed now!

vermouth1992: Unfortunately, the key is still not correct :(

PeterSH6: @chujiezheng The key should be `actor_rollout_ref.model.use_remove_padding`. The actor and ref are not necessary as they have the same model type.

chujiezheng: @vermouth1992 @PeterSH6 Fixed now 🥹

vermouth1992 (approved):

Issue #154, opened 2025-01-28
Python import errors after pip install


zhe-thoughts:
To reproduce:

```
pip install git+https://github.com/volcengine/verl.git

from verl.workers.fsdp_workers import ActorRolloutRefWorker, CriticWorker
```
Error: 
```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/zhezh/anaconda3/envs/rl/lib/python3.12/site-packages/verl/workers/fsdp_workers.py", line 29, in <module>
    from verl.single_controller.base import Worker
  File "/Users/zhezh/anaconda3/envs/rl/lib/python3.12/site-packages/verl/single_controller/__init__.py", line 19, in <module>
    with open(os.path.join(version_folder, 'version/version')) as f:
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
FileNotFoundError: [Errno 2] No such file or directory: '/Users/zhezh/anaconda3/envs/rl/lib/python3.12/site-packages/verl/single_controller/version/version'
```

eric-haibin-lin:
Should be fixed now. Feel free to reopen if the issue is still there

zhe-thoughts:
Thanks @eric-haibin-lin . I think there is still a problem

After `pip install -U git+https://github.com/volcengine/verl.git`

I'm getting:

```
(rl) zhezh@zhezh-mlt ~ % python
Python 3.12.8 | packaged by Anaconda, Inc. | (main, Dec 11 2024, 10:37:40) [Clang 14.0.6 ] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> from verl.workers.fsdp_workers import ActorRolloutRefWorker, CriticWorker
WARNING 01-29 10:32:29 _custom_ops.py:19] Failed to import from vllm._C with ModuleNotFoundError("No module named 'vllm._C'")
INFO 01-29 10:32:29 importing.py:10] Triton not installed; certain GPU-related functions will not be available.
/Users/zhezh/anaconda3/envs/rl/lib/python3.12/site-packages/vllm/connections.py:8: RuntimeWarning: Failed to read commit hash:
No module named 'vllm._version'
  from vllm.version import __version__ as VLLM_VERSION
>>> 
```

eric-haibin-lin:
this is a warning from vllm and they fixed recently. Does not affect performance or correctness

PR #156, opened 2025-01-29
fix: fix missing version file dependency when installing without -e

eric-haibin-lin:
There're two version files after the single_controller refactoring. Let's only keep one source of truth.

fixes https://github.com/volcengine/verl/issues/154


vermouth1992:
Shall we make one of our tests not installed with -e?

eric-haibin-lin:
It can only be reproduced if installing from git+https://github or future pypi, because the version file is not in setup.py

Issue #151, opened 2025-01-28
Pluggable reward function?


corbt:
Hi there! I've been trying out veRL on some toy math datasets and it's working well. However, my issue is that since the reward functions appear to be hard-coded in `main_ppo`, in order to experiment with different reward functions it is necessary to fork the library and edit this file. I just want to double-check that this is correct, and there's no way to import veRL and call it from another Python wrapper script while passing in my own custom reward function implementation? If that's indeed correct, then you can also consider this issue a feature request. 🙂

PeterSH6:
Hi @corbt, thank you for raising this important point!

You're correct - the reward functions are currently hard-coded in the `_select_rm_score_fn` function. To experiment with different reward functions, you would need to either modify this function directly or use monkey patching to override `_select_rm_score_fn`.

Making the reward function more modular and pluggable is definitely a valuable suggestion. We'll add this feature request to our development roadmap. If you're interested in contributing to this improvement, we'd be happy to review the PR.

Additionally, we're always looking to expand our collection of reward functions for different datasets. Contributions in this area would be greatly appreciated!

Please let us know if you have any other questions or suggestions.

accupham:
@PeterSH6 Thanks for adding in the plugin feature. What should the return value of the custom compute_score function be? A float between -1.0 and 1.0?

PeterSH6:
@accupham In most cases, a float in [-1.0, 1.0] works fine. But it can be other ranges based on your algorithm design

corbt:
Added in https://github.com/volcengine/verl/pull/162

PR #162, opened 2025-01-30
Allow users to pass in custom compute_score function

corbt:
This is a follow-up to https://github.com/volcengine/verl/issues/151

## Motivation

Currently, in order to add a custom score function you need to fork verl and update the `_select_rm_score_fn` to define your logic. This makes it harder to use verl as part of a larger application while staying up to date with upstream improvements in verl.

It would be convenient to allow end users to directly pass in a reward function they wish to use, without requiring them to clone/fork verl to do so.

## Design

In this PR I slightly modify `main_ppo.py` to allow users to import a new function `run_ppo`. `run_ppo` behaves very similarly to the existing `main`, with the important addition of a new `compute_score` argument. This argument, if passed in, is used to compute the score of every generation. This is the change that allows 

The `compute_score` function is similar in shape to the existing `compute_score` on gsm8k and math. However, I have added a new `data_source` parameter so that the user can compute the score differently if desired depending on the task shape.

## Example Usage

This is a sample script showing how you can use the new functionality. I have tested that this works.

```python
from verl.trainer.main_ppo import run_ppo
from omegaconf import OmegaConf


def custom_compute_score(data_source, solution_str, ground_truth):
    """Dummy compute_score function that reward the model for generations of exactly 20 characters :)
    """
    return abs(len(solution_str) - 20)


config = OmegaConf.load("vendor/verl/verl/trainer/config/ppo_trainer.yaml")

# Update config as needed
config.data.train_files = "path/to/train.parquet"
config.data.val_files = "path/to/test.parquet"
# ...

run_ppo(config, custom_compute_score)
```

## Breaking changes

There are no breaking changes in this PR. It is still possible to call `python -m verl.trainer.main_ppo ...` as before (although if you want to pass in a custom compute_score you will need to use the new method described above).

## Possible future work

It would be great to move to [structured configs](https://omegaconf.readthedocs.io/en/2.1_branch/structured_config.html) as well since they'd allow us to have typesafe, autocompletable configurations from Python. I thought about adding those changes here as well but they would be much more extensive and I'm not sure whether there's interest from the project.

### DIFF
--- a/verl/trainer/main_ppo.py
+++ b/verl/trainer/main_ppo.py
@@ -21,11 +21,11 @@ from verl.utils.reward_score import gsm8k, math
 from verl.trainer.ppo.ray_trainer import RayPPOTrainer
 
 
-def _select_rm_score_fn(data_source):
+def _default_compute_score(data_source, solution_str, ground_truth):
     if data_source == 'openai/gsm8k':
-        return gsm8k.compute_score
+        return gsm8k.compute_score(solution_str, ground_truth)
     elif data_source == 'lighteval/MATH':
-        return math.compute_score
+        return math.compute_score(solution_str, ground_truth)
     else:
         raise NotImplementedError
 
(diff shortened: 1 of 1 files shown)


corbt:
(cc @PeterSH6 for review)

PeterSH6:
Thanks for your PR!

Will review later today :)

PeterSH6:
@corbt This PR LGTM! Nice work!

I wonder what the APIs will be like when you move them to structured configs?

PeterSH6 (approved):

PR #173, opened 2025-01-30
[feat, SFT] Support LigerKernel for SFT

xingyaoww:
This PR adds support for LigerKernel for SFT and add a test to CI for this feature.

### DIFF
--- a/.github/workflows/e2e_sft.yml
+++ b/.github/workflows/e2e_sft.yml
@@ -57,4 +57,4 @@ jobs:
       - name: Running gsm8k e2e training tests on 8 L20 GPUs with sequence parallism and liger
         run: |
           ray stop --force
-          bash examples/sft/gsm8k/run_qwen_05_sp2_liger.sh 8 $HOME/ckpts/
+          bash tests/sft/run_sft_qwen05_sp2_liger.sh 8 $HOME/ckpts/
--- a/.github/workflows/e2e_sft.yml
+++ b/.github/workflows/e2e_sft.yml
@@ -54,3 +54,7 @@ jobs:
         run: |
           ray stop --force
           bash tests/sft/run_sft_sp_loss_match.sh
+      - name: Running gsm8k e2e training tests on 8 L20 GPUs with sequence parallism and liger
+        run: |
+          ray stop --force
+          bash examples/sft/gsm8k/run_qwen_05_sp2_liger.sh 8 $HOME/ckpts/
--- a/examples/sft/gsm8k/run_qwen_05_sp2_liger.sh
+++ b/examples/sft/gsm8k/run_qwen_05_sp2_liger.sh
@@ -1,0 +1,33 @@
+set -x
+
+if [ "$#" -lt 2 ]; then
+    echo "Usage: run_qwen_05_sp2.sh <nproc_per_node> <save_path> [other_configs...]"
+    exit 1
+fi
+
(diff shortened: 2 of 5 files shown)


verl/trainer/config/sft_trainer.yaml
```diff
@@ -23,6 +23,7 @@ model:
   lora_rank: 0  # Set to positive value to enable LoRA (e.g., 32)
   lora_alpha: 16  # LoRA scaling factor
   target_modules: all-linear  # Target modules for LoRA adaptation
+  use_liger: True
```
eric-haibin-lin: not sure if we want to enable it by default. thoughts? @PeterSH6 @vermouth1992

.github/workflows/e2e_sft.yml
```diff
@@ -54,3 +54,7 @@ jobs:
         run: |
           ray stop --force
           bash tests/sft/run_sft_sp_loss_match.sh
+      - name: Running gsm8k e2e training tests on 8 L20 GPUs with sequence parallism and liger
+        run: |
+          ray stop --force
+          bash examples/sft/gsm8k/run_qwen_05_sp2_liger.sh 8 $HOME/ckpts/
```
PeterSH6: In the CI, can you upload a script that's only for the CI purpose in tests/sft?

The script should be identical to the example but only run one single step

examples/sft/gsm8k/run_qwen_05_sp2_liger.sh
```diff
@@ -0,0 +1,33 @@
...
+    +data.prompt_dict_keys=['question'] \
+    +data.response_dict_keys=['answer'] \
+    data.micro_batch_size=4 \
+    model.partial_pretrain=Qwen/Qwen2.5-0.5B-Instruct \
+    model.use_liger=True \
+    trainer.default_local_dir=$save_path \
+    trainer.project_name=gsm8k-sft \
+    trainer.experiment_name=gsm8k-sft-qwen-2.5-0.5b-instruct-sp2-liger \
+    trainer.logger=['console'] \
+    trainer.total_training_steps=1 \
```
PeterSH6: In the example, please delete the `trainer.total_training_steps=1`. Otherwise, users have to modify your script before training.

verl/trainer/config/sft_trainer.yaml
xingyaoww: Reverted to default false!

.github/workflows/e2e_sft.yml
xingyaoww: yep! fixed!

examples/sft/gsm8k/run_qwen_05_sp2_liger.sh
xingyaoww: fixed!

vermouth1992 (approved):

verl/trainer/config/sft_trainer.yaml
eric-haibin-lin: Thanks! It would be great to mention this option in https://github.com/volcengine/verl/blob/main/docs/perf/perf_tuning.rst as well~

xingyaoww: OpenHands helped me did it here: https://github.com/volcengine/verl/pull/178 :)

PR #175, opened 2025-01-30
data: fix the math dataset source

eric-haibin-lin:
since 'lighteval/MATH' is no longer available on huggingface.

### DIFF
--- a/verl/workers/megatron_workers.py
+++ b/verl/workers/megatron_workers.py
@@ -397,10 +397,6 @@ class ActorRolloutRefWorker(MegatronWorker):
         torch.cuda.empty_cache()
         return output
 
-    @register(dispatch_mode=Dispatch.MEGATRON_COMPUTE_PROTO)
-    def compute_log_prob(self, data: DataProto):
-        raise NotImplementedError()
-
     @register(dispatch_mode=Dispatch.ONE_TO_ALL)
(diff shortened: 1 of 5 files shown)


verl/workers/megatron_workers.py
```diff
@@ -397,6 +397,10 @@ def compute_ref_log_prob(self, data: DataProto):
         torch.cuda.empty_cache()
         return output
 
+    @register(dispatch_mode=Dispatch.MEGATRON_COMPUTE_PROTO)
+    def compute_log_prob(self, data: DataProto):
+        raise NotImplementedError()
```
PeterSH6: To fix the missing `compute_log_prob` in megatron-lm backend, I think we can simply return an empty dataProto and print some warning.

As the generate sequences func in megatron_workers already recompute the log_prob.
See: https://github.com/volcengine/verl/blob/main/verl/workers/megatron_workers.py#L368-L373

eric-haibin-lin: I'll fix it soon

vermouth1992: Shall we merge this PR first and fix compute log prob in a separate PR?

PeterSH6: Agree

eric-haibin-lin: reverted the change and leave it for future PR

PR #177, opened 2025-01-31
Add option to log validation generations to wandb

corbt:
## Motivation

Often the summary of average/max/min reward is not enough information, and it's helpful to look at some real-world generations to see how the model's actual behavior is changing over time. This can be particularly helpful for debugging issues like the generation being cut off before reasoning finishes.

## Change

This PR introduces a new `trainer.val_generations_to_log_to_wandb` config value, with a default of 0. If set to a number larger than 0, it logs that number of inputs/outputs/scores each time the validation set is generated and scored. It uses a [wandb Table](https://docs.wandb.ai/guides/track/log/log-tables/) to do so, adding a single row for each validation set run.

I choose to log the data in this format because it allows a user to easily see how the outputs for a given input change over time by looking down a column vertically.

## Screenshot

Screenshot 2025-01-31 at 8 02 47 AM

Note: if there's already another way to accomplish this easily let me know! I was surprised not to find a way to see sample generations because I find that quite useful, so let me know if I'm missing something.

### DIFF
--- a/verl/trainer/config/ppo_trainer.yaml
+++ b/verl/trainer/config/ppo_trainer.yaml
@@ -158,6 +158,7 @@ trainer:
   project_name: verl_examples
   experiment_name: gsm8k
   logger: [ 'console', 'wandb' ]
+  val_generations_to_log_to_wandb: 0
   nnodes: 1
   n_gpus_per_node: 8
   save_freq: -1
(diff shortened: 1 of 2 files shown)


vermouth1992:
@PeterSH6 Shall we unify the config in e2e ci?

PeterSH6:
> @PeterSH6 Shall we unify the config in e2e ci?

Yes, I think it's necessary

vermouth1992:
Hi @corbt,

Could you rebase main and this should fix the CI. This feature is important for case study!

corbt:
Sure thing, rebased!

PeterSH6:
Hi @corbt,

Could you add the `val_generations_to_log_to_wandb: 0` config to the ppo_megatron_trainer.yaml to support the megatron backend?

PeterSH6:
Merged first. Will fix the Megatron ci in the next PR

PR #182, opened 2025-02-01
fix a typo in logprobs_from_logits_flash_attn

bebetterest:
It seems that the tensor index should be removed. 🤯

### DIFF
--- a/docs/examples/gsm8k_example.rst
+++ b/docs/examples/gsm8k_example.rst
@@ -74,7 +74,7 @@ We provide a SFT Trainer using PyTorch FSDP in
 Users can customize their own SFT
 script using our FSDP SFT Trainer.
 
-We also provide various training scripts for SFT on GSM8K dataset in `gsm8k sft directory <https://github.com/volcengine/verl/blob/main/examples/gsm8k/sft/>`_.
+We also provide various training scripts for SFT on GSM8K dataset in `gsm8k sft directory <https://github.com/volcengine/verl/blob/main/examples/sft/gsm8k/>`_.
 
 .. code:: shell
 


eric-haibin-lin:
thansk for the PR. what was the error you were getting initially?

vermouth1992:
It's actually not tensor index. See here https://github.com/Dao-AILab/flash-attention/blob/main/flash_attn/ops/triton/cross_entropy.py#L255 We only need the ce loss, but not the z_losses

bebetterest:
I met this error when training

```python
  File "/home/aiscuser/rrl_verl/verl/verl/single_controller/ray/base.py", line 399, in func
    return getattr(self.worker_dict[key], name)(*args, **kwargs)
  File "/home/aiscuser/rrl_verl/verl/verl/single_controller/base/decorator.py", line 404, in inner
    return func(*args, **kwargs)
  File "/home/aiscuser/rrl_verl/verl/verl/workers/fsdp_workers.py", line 471, in compute_log_prob
    old_log_probs = self.actor.compute_log_prob(data=data)
  File "/home/aiscuser/rrl_verl/verl/verl/workers/actor/dp_actor.py", line 191, in compute_log_prob
    _, log_probs = self._forward_micro_batch(micro_batch, temperature=temperature)
  File "/home/aiscuser/rrl_verl/verl/verl/workers/actor/dp_actor.py", line 138, in _forward_micro_batch
    log_probs = logprobs_from_logits(logits, micro_batch['responses'])
  File "/home/aiscuser/rrl_verl/verl/verl/utils/torch_functional.py", line 59, in logprobs_from_logits
    output = output.view(*batch_dim)
RuntimeError: shape '[4, 4096]' is invalid for input of size 1
```
https://github.com/volcengine/verl/blob/13762f43d3fdb6ed1b66f5ec70749759d8001891/verl/utils/torch_functional.py#L49-L62

I would check again. thx🤯

bebetterest:
Oh, I find that it's because I did not apply the right flash-attn version. `cross_entropy_loss` returns losses, z_losses after https://github.com/Dao-AILab/flash-attention/commit/d8aacc510c20cc89ecf1afba1df234d239c0e37a. Thx🤯

vermouth1992:
I guess the later is better

PR #196, opened 2025-02-04
Add push to hub functionality

### LINKED ISSUE #190: Integrate Verl with 🤗 hub
Hi veRL authors!

Congrats on this amazing open-source framework.

I'm Niels from the open-source team at 🤗 , and was wondering whether you would be up for providing a `--push_to_hub` flag to the training scripts, which will automatically push the trained model to the hub, along with various metadata, like a `verl` tag.

We do the same for other libraries, e.g. for TRL which can be observed here: https://huggingface.co/models?other=trl, as well as for the RL algorithms being used, e.g. orpo: https://huggingface.co/models?other=orpo.

See here on how this is implemented for TRL: https://github.com/huggingface/trl/blob/bbdd6db17c49db813695d0a8bc0da7bf6b1bb88e/trl/trainer/grpo_trainer.py#L611.

We can do the same for your library :)

Let me know if you're interested, happy to send a PR.

Kind regards,

Niels

NielsRogge:
This PR adds a `push_to_hub` option to the YAML configuration. This enables people to push their trained model to the 🤗 hub, allowing easy inference afterwards. Moreover, each pushed model gets an automated model card along with a `verl` tag, enabling you to see which models people train using VeRL.

For PPO training:

```bash
    ...
    trainer.push_to_hub=True \
    trainer.hub_model_id="username/model-name" \
    trainer.hub_private=False \
    trainer.hub_token="your_token" # Optional
```

For SFT training:

```bash
python -m verl.trainer.main_sft \
    ...
    trainer.push_to_hub=True \
    trainer.hub_model_id="username/model-name" \
    trainer.hub_private=False \
    trainer.hub_token="your_token" # Optional
```

Fixes #190


verl/single_controller/base/worker.py
```diff
@@ -183,4 +183,28 @@ def execute_with_func_generator(self, func, *args, **kwargs):
     @register(dispatch_mode=Dispatch.ALL_TO_ALL, execute_mode=Execute.RANK_ZERO)
     def execute_func_rank_zero(self, func, *args, **kwargs):
         result = func(*args, **kwargs)
-        return result
\ No newline at end of file
+        return result
+
+    def push_to_hub(self, repo_id: str, local_path: str, private: bool = False, token: str = None):
```
vermouth1992: This function should not be defined here. Instead, it should be defined as a class method of ActorRolloutRefWorker here just like `save_checkpoint` https://github.com/volcengine/verl/blob/main/verl/workers/fsdp_workers.py#L798. After the definition, give it a decorator (@register(dispatch_mode=Dispatch.ONE_TO_ALL)), which indicates that the workergroup can invoke the `push_to_hub` method as a class method of itself. Under the hood, it calls `push_to_hub` of every worker inside the workergroup, and the data is dispatched as ONE_TO_ALL (the same input data is distributed to every worker). Inside this function, we can add 
```python
if self.rank == 0:
    # do something
torch.distributed.barrier()  # wait for rank zero to finish
```
This is the programming paradigm of single controller + multi-controller (HybridFlow) and we will have detail tutorials of this later.

vermouth1992:
Hi @NielsRogge,
Nice work! Just wonder is there any CI that we can run to protect this functionality?

verl/single_controller/ray/base.py
```diff
@@ -366,6 +366,23 @@ def workers(self):
     def world_size(self):
         return self._world_size
 
+    def push_to_hub(self, repo_id: str, local_path: str, private: bool = False, token: str = None):
```
PeterSH6: The push_to_hub api should also be added to fsdp_workers.py and megatron_worker.py with a decorator as suggested by @vermouth1992 in https://github.com/volcengine/verl/pull/196/files#r1941126564

PeterSH6: We already have a execute mode for execute_rank_zero, namely: 
`@register(dispatch_mode=Dispatch.ONE_TO_ALL, execute_mode=Execute.RANK_ZERO)`

PeterSH6 (changes requested):

NielsRogge:
Thanks for your comments, addressed them.

verl/trainer/fsdp_sft_trainer.py
```diff
@@ -439,6 +439,64 @@ def save_checkpoint(self, step):
             if self.config.trainer.default_hdfs_dir:
                 hdfs_io.makedirs(self.config.trainer.default_hdfs_dir, exist_ok=True)
                 hdfs_io.copy(src=path, dst=self.config.trainer.default_hdfs_dir, dirs_exist_ok=True)
+
+            # Push to hub if enabled
+            if self.config.trainer.push_to_hub:
```
eric-haibin-lin, line 444: thanks for the contribution! 
could you consolidate these in a util function so that the trainer code looks less bulky? thanks!

NielsRogge: Sure, added a utility function

NielsRogge:
Hi @eric-haibin-lin does this PR look good to you?

verl/workers/fsdp_workers.py
```diff
@@ -573,6 +573,24 @@ def load_checkpoint(self, path, del_local_after_load=True):
         if self._is_offload_param:
             offload_fsdp_param_and_grad(module=self.actor_module_fsdp, offload_grad=self._is_offload_grad)
 
+    @register(dispatch_mode=Dispatch.ONE_TO_ALL, execute_mode=Execute.RANK_ZERO)
+    def push_to_hub(self, repo_id: str, local_path: str, private: bool = False, token: str = None):
```
eric-haibin-lin, line 577: the change here is not necessary right? @vermouth1992 ?

vermouth1992: Hi @NielsRogge,

We have changed all the checkpoint saver in the training code to only save distributed checkpoints. So, there is no way for us to push to huggingface hub inside the code. We require the user to merge checkpoint after training. Would you please add your code to here https://github.com/volcengine/verl/blob/main/scripts/model_merger.py? This also applies to megatron backend later when we refactor the megatron code. Thanks!

vermouth1992: In fact, the functionality already exists. https://github.com/volcengine/verl/blob/main/scripts/model_merger.py#L154

NielsRogge: Ok thanks for the pointer @vermouth1992. However pushing that doesn't add a model card and `verl` tags.

Would it be useful to add this functionality?

eric-haibin-lin: I'd agree with @vermouth1992 that the merger script is already sufficient. I personally prefer a minimal trainer loop. @NielsRogge your efforts are much appreciated. I'll close this PR for now but future contributions are always welcome!

NielsRogge:
Hi @eric-haibin-lin just want to point out that we currently have 66k models on 🤗 tagged with TRL: https://huggingface.co/models?other=trl, and none for "verl".

Would be great to achieve the same for your library, if you're interested.

eric-haibin-lin:
@NielsRogge we'd love to work and contribute to the huggingface community! huggingface invented great library and platform and is great to the community. But would the model_merger script be a better place since the safetensor conversion is done there? I'd suggest moving the hf tag / hf token changes to `scripts/model_merger.py`

NielsRogge:
Yes, that could work. It would be nice if people can pass a boolean push_to_hf in the config so that the model gets pushed to the hub after training (and have "verl" in its metadata). Is this something you could open a PR for?

PR #201, opened 2025-02-05
fix: fix typo

kinman0224:


verl/utils/seqlen_balancing.py
```diff
@@ -222,7 +222,7 @@ def ceildiv(a, b):
 
 
 def rearrange_micro_batches(batch: TensorDict, max_token_len, dp_group=None):
-    """Split the batch into a list of micro_batches, where the max_token_len is smaller than max_token_len
+    """Split the batch into a list of micro_batches, where the max_token_len is greater than max_seq_len
```
vermouth1992: I guess the original meaning would be where the length of each micro_batch is smaller than max_token_len

kinman0224: Oh, it’s still a little unclear! updated

vermouth1992: You can search for the tests and it will be clear.

PR #207, opened 2025-02-05
Add stronger reward verification sandbox

ZefanW:
Add stronger verification support as is used in https://github.com/PRIME-RL/PRIME

- [x] Batched verification
- [x] Python interpreter
- [x] Stronger math verifier
- [x] Continuous score for code test

### DIFF
--- a/verl/trainer/config/ppo_trainer.yaml
+++ b/verl/trainer/config/ppo_trainer.yaml
@@ -142,6 +142,7 @@ reward_model:
   ulysses_sequence_parallel_size: 1 # sp size
   use_dynamic_bsz: ${critic.use_dynamic_bsz}
   forward_max_token_len_per_gpu: ${critic.forward_max_token_len_per_gpu}
+  reward_manager: naive
 
 algorithm:
   gamma: 1.0
--- a/verl/trainer/main_ppo.py
+++ b/verl/trainer/main_ppo.py
@@ -104,7 +104,11 @@ class RewardManager():
         assert len(sequences_str) == len(ground_truth) == len(data_sources)
         try:
             scores = asyncio.run(
-                parallel_compute_score_async(self.compute_score, sequences_str, ground_truth, data_sources, num_processes=64))
+                parallel_compute_score_async(self.compute_score,
+                                             sequences_str,
+                                             ground_truth,
+                                             data_sources,
+                                             num_processes=64))
         except asyncio.TimeoutError as e:
             print('Global timeout in reward computing! Setting all as 0.')
             scores = [0. for _ in range(len(sequences_str))]
--- a/verl/trainer/main_ppo.py
+++ b/verl/trainer/main_ppo.py
@@ -23,20 +23,8 @@ from tqdm.asyncio import tqdm
 from verl import DataProto
 import torch
 from verl.utils.reward_score import gsm8k, math, prime_math, prime_code
+from verl.utils.reward_score import _default_compute_score
 from verl.trainer.ppo.ray_trainer import RayPPOTrainer
-
-
-def _default_compute_score(data_source, solution_str, ground_truth):
-    if data_source == 'openai/gsm8k':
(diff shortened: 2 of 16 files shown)


verl/trainer/main_ppo.py
```diff
@@ -14,21 +14,58 @@
...
+        if isinstance(result, Exception) or result is None:
+            # Handle failed or timed-out tasks
+            scores.append(0.0)
+            continue
+        else:
+            scores.append(float(int(result[0][0])))
+        # TODO: implement continual code scoring in sandboxes
+    return scores
 
 class RewardManager():
```
eric-haibin-lin: shall we introduce a new reward manager class that overrides the `__call__` method?

ZefanW: I guess the new implementation of `__call__` does not affect the original functionality of `__call__`, as long as the evaluation function itself can run properly in a multi-process environment (i.e., the objects used are all picklable and there are no side effects). I'll test the GSM8K and MATH examples.

verl/utils/reward_score/prime_code/utils.py
```diff
@@ -0,0 +1,45 @@
...
+            result.append(res)
+            metadata_list.append(metadata)
+        except Exception as e:
+            # print(e) # some tracebacks are extremely long.
+            traceback.print_exc(10)
+            result.append([-1 for i in range(len(sample['inputs']))])
+            metadata_list.append({})
+
+
+def check_correctness(in_outs: Optional[dict], generation, timeout=10, debug=True):
```
vermouth1992, line 26: Could you add a test case to show the usage of this function?

verl/trainer/main_ppo.py
```diff
@@ -14,20 +14,61 @@
...
-        raise NotImplementedError
+async def single_compute_score(completion, reference, task, executor, timeout=300.):
+    loop = asyncio.get_running_loop()
+    try:
+        # Ensure process_completion is called properly
+        tasks = [
+            asyncio.wait_for(
+                loop.run_in_executor(
+                    executor,
+                    partial(_default_compute_score, task, completion, reference)  # Ensure synchronous
```
vermouth1992: We will need a function that can pass in compute_score from main_task so that user can pass in their own compute_score function

verl/utils/reward_score/prime_code/utils.py
ZefanW: done in https://github.com/volcengine/verl/pull/207/commits/7f031b9c9e4dc28fe509e32b46912214e1f7cc42

verl/trainer/main_ppo.py
ZefanW: done in https://github.com/volcengine/verl/pull/207/commits/7f031b9c9e4dc28fe509e32b46912214e1f7cc42

verl/trainer/main_ppo.py
```diff
@@ -14,81 +14,19 @@
 """
 Note that we don't combine the main with ray_trainer as ray_trainer is used by other main.
 """
+from concurrent.futures import ProcessPoolExecutor
+import asyncio
+from functools import partial
+
+from tqdm.asyncio import tqdm
```
eric-haibin-lin: the ProcessPoolExecutor / asyncio libs are no longer needed in this file

verl/trainer/config/ppo_trainer.yaml
```diff
@@ -142,6 +142,7 @@ reward_model:
   ulysses_sequence_parallel_size: 1 # sp size
   use_dynamic_bsz: ${critic.use_dynamic_bsz}
   forward_max_token_len_per_gpu: ${critic.forward_max_token_len_per_gpu}
+  reward_manager: naive
```
eric-haibin-lin, line 145: could you add this option to https://github.com/volcengine/verl/blob/main/docs/examples/config.rst as well? thanks!

eric-haibin-lin (approved):
No major concern from my side

Issue #212, opened 2025-02-06
Incorrect response_attention_mask observed


kinman0224:
When running instruct serial model, like `examples/ppo_trainer/run_qwen2-7b.sh`, I observed that certain requests would stop with `<|endoftext|>`. This occurs because vLLM retrieves the eos_token_id from `generation_config.json`, where the eos_token_id is specified as `[151645, 151643]`.

As a result, the response_attention_mask in `vllm_rollout.py` becomes incorrect, as the response lacks `<|im_end|>`. 
```
response_attention_mask = get_eos_mask(response_id=response, eos_token=eos_token_id, dtype=attention_mask.dtype)
```

This issue arises because the eos_token_id is passed in e.g. `fsdp_worker.py` through the following code:
```
meta_info = {'eos_token_id': self.tokenizer.eos_token_id, 'pad_token_id': self.tokenizer.pad_token_id}
```

Consequently, the response_attention_mask is set to all 1 (including pad_token).

PR #213, opened 2025-02-06
[rollout]: fix incorrect response_attention_mask in vLLM rollout

kinman0224:
This PR addresses issue https://github.com/volcengine/verl/issues/212.

The changes include:
- read eos_token_id from generation_config to ensure alignment with vLLM
- modified the get_eos_mask function to accept both int and list types for the eos_token parameter.

### DIFF
--- a/verl/workers/fsdp_workers.py
+++ b/verl/workers/fsdp_workers.py
@@ -149,7 +149,7 @@ class ActorRolloutRefWorker(Worker):
                                role='actor'):
         from verl.utils.model import print_model_size, update_model_config
         from verl.utils.torch_dtypes import PrecisionType
-        from transformers import AutoModelForCausalLM, AutoConfig
+        from transformers import AutoModelForCausalLM, AutoConfig, GenerationConfig
         from torch.distributed.fsdp import FullyShardedDataParallel as FSDP, ShardingStrategy, MixedPrecision, CPUOffload
         from torch import optim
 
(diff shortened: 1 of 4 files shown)


vermouth1992:
This PR should also remove the need to define a custom hf_tokenizer https://github.com/volcengine/verl/blob/main/verl/utils/tokenizer.py#L35 as gemma-2b-it also contains 2 eos token

verl/workers/megatron_workers.py
```diff
@@ -146,6 +146,8 @@ def _build_model_optimizer(self,
         # Step 2: get the actor_model_config
         actor_model_config = AutoConfig.from_pretrained(local_path)
 
+        self.generation_config = GenerationConfig.from_pretrained(local_path)
```
vermouth1992: Shall we add a fallback, for some models, there is no generation config?

kinman0224: Yes! That would be better. I will update the code.

kinman0224: have updated

verl/workers/fsdp_workers.py
```diff
@@ -445,7 +447,14 @@ def generate_sequences(self, prompts: DataProto):
                                      load_grad=self._is_offload_grad)
 
         prompts.batch = prompts.batch.cuda()
-        meta_info = {'eos_token_id': self.tokenizer.eos_token_id, 'pad_token_id': self.tokenizer.pad_token_id}
+        meta_info = {
+            'eos_token_id':
+                self.generation_config.eos_token_id
+                if self.generation_config.eos_token_id is not None else self.tokenizer.eos_token_id,
```
vermouth1992: If self.generation_config is None, will this raise exception?

kinman0224: oh! there is mistake

PeterSH6:
Do we still need to update generation config in vllm SamplingParams?

vermouth1992:
CI machine is running out of disk...

vermouth1992 (approved):

PR #219, opened 2025-02-07
megatron: drop the need for megatron patches

eric-haibin-lin:
Docs to be updated in a separate PR.


verl/workers/actor/megatron_actor.py
```diff
@@ -302,20 +302,24 @@ def forward_step(batch_iter, model):
...
         batch_generator = make_batch_generator(batches, vpp_size=len(self.actor_module))
+        from verl.utils.megatron.pipeline_parallel import require_extra_schedule_kwargs
+        schedule_kwargs = {}
+        if require_extra_schedule_kwargs():
+            schedule_kwargs = {'hidden_size': self.model_config.hidden_size}
 
         # TODO: we may use the new schedule instead
         # for flash-attn: (seq_len, batch_size, hidden_size) = (mbs*seq_len, 1, hidden_size)
         if mpu.get_pipeline_model_parallel_world_size() > 1:
+            schedule_kwargs['input_shapes'] = input_shapes
```
PeterSH6, line 313: The input_shapes param is used for fixing the vpp hanging problem. If we enable vpp, I guess we still need a patch in v0.4.

eric-haibin-lin: Alright... maybe we should target v0.7 then

PR #225, opened 2025-02-08
[ckpt] feat: support saving and loading FSDP full state dict in ckpt manager

PeterSH6:
- As titled
- Currently, the FSDPCkptManger can save/load either sharded state dict and full state dict.
- Recommend using sharded state dict as full state dict is too slow.

**TODO**
- But we should provide a merging script for the sharded state dict to form a huggingface ckpt.


vermouth1992:
Do we have a merge ckpt script?

PeterSH6:
> Do we have a merge ckpt script?

Will add one in a separate PR

PR #240, opened 2025-02-10
Typo in the rank zero code

ExtremeViscent:
The original Ray controller method `execute_rank_zero_sync()` is not functional. Fixed.


eric-haibin-lin:
could you also add a test to tests/ray? thanks!

PR #244, opened 2025-02-10
[misc] feat: give main_task num_cpus=1

vermouth1992:
give main_task num_cpus=1 and make sure that main_task should not be scheduled on head


Ber666:
Hi, thanks for the great work. Could you explain why we don't want the `main_task` to be scheduled on head, and how `num_cpus=1` achieve that?

Thanks again!

jgreer013:
@vermouth1992 Please clarify your code comment

PR #247, opened 2025-02-11
[bug]fix bug on self.prompt_key/self.response_key of SFTDataset

XiongcaiLuo:
- change self.prompt_key/self.response_key from list type to str type.
- Otherwise, self.prompts/self.responses will be a Dataframe, and will raise an Exception when calling tolist() : `self.prompts = self.prompts.tolist()`


cauyxy (approved):
good catch

PR #258, opened 2025-02-12
clean 'WG_BACKEND' unused code.

zhanluxianshen:
1  i see `WG_BACKEND` will == 'ray'
2 i can not found the pkg about ' from verl.single_controller.torchrpc.k8s_client import get_ip_addr'

would these changes is ok for that?

### DIFF
--- a/verl/single_controller/base/worker.py
+++ b/verl/single_controller/base/worker.py
@@ -42,10 +42,8 @@ class WorkerHelper:
             if os.getenv("WG_BACKEND", None) == "ray":
                 import ray
                 return ray._private.services.get_node_ip_address()
-            elif os.getenv("WG_BACKEND", None) == "torch_rpc":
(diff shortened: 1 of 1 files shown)


vermouth1992:
Yes, please delete the torchrpc part

verl/single_controller/base/worker.py
```diff
@@ -45,6 +45,7 @@ def get_node_ip_by_sdk():
             elif os.getenv("WG_BACKEND", None) == "torch_rpc":
                 from verl.single_controller.torchrpc.k8s_client import get_ip_addr
```
vermouth1992: Could you delete this part and assert NotImplementedError for WG_BACKEND != Ray？

zhanluxianshen: yep, i delete that.
But ci trigger the error.
So i test the code.

Issue #189, opened 2025-02-03
Encounter issue when perform RL rollout


HillZhang1999:
I tried to use verl to perform GRPO with 8k response length for 32B models. Here is my training script (I used vllm 0.6.3):

```
export CUDA_VISIBLE_DEVICES=0,1,2,3,4,5,6,7
export N_GPUS=8
export BASE_MODEL=/mnt/bn/hillzhang/TinyZero/DeepSeek-R1-Distill-Qwen-32B
export DATA_DIR=/mnt/bn/hillzhang/TinyZero/data/Math-RL-Data-41k-250201-filtered_processed
export ROLLOUT_TP_SIZE=8
export EXPERIMENT_NAME=Math-RL-Data-41k-250201-filtered_processed-32b-grpo-$(date +%Y%m%d_%H%M%S)
export VLLM_ATTENTION_BACKEND=XFORMERS

python3 -m verl.trainer.main_ppo \
algorithm.adv_estimator=grpo \
data.train_files=${DATA_DIR}/train.parquet \
data.val_files=${DATA_DIR}/test.parquet \
data.train_batch_size=32 \
data.val_batch_size=300 \
data.max_prompt_length=1024 \
data.max_response_length=8192 \
actor_rollout_ref.model.use_remove_padding=True \
actor_rollout_ref.model.enable_gradient_checkpointing=True \
actor_rollout_ref.actor.use_dynamic_bsz=True \
actor_rollout_ref.actor.ppo_max_token_len_per_gpu=16384 \
actor_rollout_ref.model.path=${BASE_MODEL} \
actor_rollout_ref.actor.optim.lr=1e-6 \
actor_rollout_ref.actor.ppo_mini_batch_size=32 \
actor_rollout_ref.actor.fsdp_config.param_offload=False \
actor_rollout_ref.actor.fsdp_config.grad_offload=False \
actor_rollout_ref.actor.fsdp_config.optimizer_offload=False \
actor_rollout_ref.rollout.tensor_model_parallel_size=${ROLLOUT_TP_SIZE} \
actor_rollout_ref.rollout.gpu_memory_utilization=0.6 \
algorithm.kl_ctrl.kl_coef=0.0 \
actor_rollout_ref.actor.use_kl_loss=True \
actor_rollout_ref.actor.kl_loss_coef=0.0 \
actor_rollout_ref.actor.kl_loss_type=low_var_kl \
actor_rollout_ref.ref.fsdp_config.param_offload=True \
[...]

PR #259, opened 2025-02-12
[misc] fix: disable chunked-prefill by default

PeterSH6:
Thanks: @HillZhang1999

- Related issue: https://github.com/volcengine/verl/issues/189

`[36m(main_task pid=3523385)[0m ValueError: max_num_batched_tokens (8192) is smaller than max_model_len (9216). This effectively limits the maximum sequence length to max_num_batched_tokens and makes vLLM reject longer sequences. Please increase max_num_batched_tokens or decrease max_model_len.`

When enable_chunked_prefill is activated, the aforementioned issue will be concealed. Please increase `max_num_batched_tokens` or `decrease max_model_len`.

### DIFF
--- a/verl/trainer/config/ppo_trainer.yaml
+++ b/verl/trainer/config/ppo_trainer.yaml
@@ -82,7 +82,7 @@ actor_rollout_ref:
     log_prob_use_dynamic_bsz: ${actor_rollout_ref.actor.use_dynamic_bsz}
     log_prob_max_token_len_per_gpu: ${actor_rollout_ref.actor.ppo_max_token_len_per_gpu}
     disable_log_stats: True
-    enable_chunked_prefill: True # could get higher throughput
+    enable_chunked_prefill: False # may get higher throughput when set to True
     # for hf rollout
     do_sample: True
     # number of responses (i.e. num sample times)


verl/trainer/config/ppo_trainer.yaml
```diff
@@ -82,7 +82,7 @@ actor_rollout_ref:
     log_prob_use_dynamic_bsz: ${actor_rollout_ref.actor.use_dynamic_bsz}
     log_prob_max_token_len_per_gpu: ${actor_rollout_ref.actor.ppo_max_token_len_per_gpu}
     disable_log_stats: True
-    enable_chunked_prefill: True # could get higher throughput
+    enable_chunked_prefill: False # may get higher throughput when set to True. When activated, Please increase max_num_batched_tokens or decrease max_model_len.
```
vermouth1992: Instead of setting this to False, is it better to add assertion in code to check max_num_batched_tokens and max_model_len

vermouth1992:
```python
if max_num_batched_tokens < max_model_len and enable_chunked_prefill:
    raise ValueError(xxx)
```

PeterSH6: Sure

vermouth1992 (approved):

Issue #273, opened 2025-02-14
Expose `remove_previous_ckpt` option to training entry point and user configuration


zwhe99:
#### Description
The `remove_previous_ckpt` option is implemented in the framework but is not exposed to the training entry point or user configuration. This limits its usability, as users cannot enable or disable this feature during training.

#### Current Behavior
The `FSDPCheckpointManager` has `remove_previous_ckpt` and `del_local_after_load ` options:
https://github.com/volcengine/verl/blob/62e23aee0b4e2c36f04a5b95fcd9f0a4eb724ee2/verl/utils/checkpoint/fsdp_checkpoint_manager.py#L96
https://github.com/volcengine/verl/blob/62e23aee0b4e2c36f04a5b95fcd9f0a4eb724ee2/verl/utils/checkpoint/fsdp_checkpoint_manager.py#L51
which are defaulted to `True` and can not be configure by users.

#### Expected Behavior

The `remove_previous_ckpt` option should be exposed in the training configuration.

Users should be able to enable or disable it via the training entry point.

eric-haibin-lin:
Resolved in https://github.com/volcengine/verl/pull/274. Thanks  for the contribution!

PR #274, opened 2025-02-14
feat: Expose `remove_previous_ckpt` option to training entry point an…

zwhe99:
Related issue: https://github.com/volcengine/verl/issues/273

- Add `remove_previous_ckpt_in_save` and `del_local_ckpt_after_load` configuration option in `ppo_trainer.yaml`
- Update `RayPPOTrainer` to support optional checkpoint deletion during loading
- Modify `ActorRolloutRefWorker` and `CriticWorker` to pass checkpoint removal flag

### DIFF
--- a/verl/workers/fsdp_workers.py
+++ b/verl/workers/fsdp_workers.py
@@ -545,7 +545,7 @@ class ActorRolloutRefWorker(Worker):
         return output
 
     @register(dispatch_mode=Dispatch.ONE_TO_ALL)
-    def save_checkpoint(self, local_path, hdfs_path=None, global_step=0):
+    def save_checkpoint(self, local_path, hdfs_path=None, global_step=0, remove_previous_ckpt=False):
         # only support save and load ckpt for actor
         assert self._is_actor
         import torch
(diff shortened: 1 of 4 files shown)


verl/workers/fsdp_workers.py
```diff
@@ -554,14 +554,17 @@ def save_checkpoint(self, local_path, hdfs_path=None, global_step=0):
                                      device_id=torch.cuda.current_device(),
                                      load_grad=self._is_offload_grad)
 
-        self.checkpoint_manager.save_checkpoint(local_path=local_path, hdfs_path=hdfs_path, global_step=global_step)
+        self.checkpoint_manager.save_checkpoint(local_path=local_path,
+                                                hdfs_path=hdfs_path,
+                                                global_step=global_step,
+                                                remove_previous_ckpt=self.config.trainer.remove_previous_ckpt_in_save)
```
PeterSH6: You cannot use `self.config.trainer.remove_previous_ckpt_in_save` here.
Instead, you should add a new param `remove_previous_ckpt` in L548 and use it here.

The `self.config.trainer.remove_previous_ckpt_in_save` will only be passed in the ray_trainer.py

verl/workers/fsdp_workers.py
```diff
@@ -828,7 +831,10 @@ def save_checkpoint(self, local_path, hdfs_path=None, global_step=0):
...
```
PeterSH6: Similar here. Your load_checkpoint is right

PeterSH6 (approved):

PR #281, opened 2025-02-15
fix the split placement example

xffxff:
The split placement example is outdated, I tried it and encountered some errors. To address this, the following changes were made in this PR
1. Copied the content from `verl/trainer/config/ppo_trainer.yaml` to `examples/split_placement/config/ppo_trainer_split.yaml`
2. Copied `RayPPOTrainer.fit` method into the `fit` func in `examples/split_placement/split_monkey_patch.py` and modified it to get the futures of `critic_output` and `actor_output`

### DIFF
--- a/examples/split_placement/split_monkey_patch.py
+++ b/examples/split_placement/split_monkey_patch.py
@@ -14,12 +14,13 @@
 """
 An naive implementation of split placment example
 """
-import os
 from pprint import pprint
-from verl.single_controller.ray import RayResourcePool, RayWorkerGroup, RayClassWithInitArgs
 from verl import DataProto
-from verl.trainer.ppo.ray_trainer import compute_advantage, apply_kl_penalty, reduce_metrics, compute_data_metrics, Role, create_colocated_worker_cls
-from codetiming import Timer
+from verl.trainer.ppo.ray_trainer import compute_advantage, apply_kl_penalty, reduce_metrics, compute_data_metrics, _timer, compute_timing_metrics
+from copy import deepcopy
+import numpy as np
+import torch
+import uuid
 
 
 def fit(self):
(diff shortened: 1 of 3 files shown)


examples/split_placement/split_monkey_patch.py
```diff
@@ -36,126 +37,161 @@ def fit(self):
...
+
+                # NOTE: make sure you set blocking=False in update_actor and update_crtic in the worker class
+                with _timer('update_actor_critic', timing_raw):
+                    critic_output = critic_output.get()
+                    critic_output_metrics = reduce_metrics(critic_output.meta_info['metrics'])
+                    metrics.update(critic_output_metrics)
+
+                    actor_output = actor_output.get()
+                    actor_output_metrics = reduce_metrics(actor_output.meta_info['metrics'])
+                    metrics.update(actor_output_metrics)
```
xffxff, line 167: Modified to get futures of critic_output and actor_output

PeterSH6: Nice job! What do you modify here?

xffxff: > Modified to get futures of critic_output and actor_output

The comments here are just to highlight the changes compared to `RayTrainer.fit`, because most of the code is copied from `RayTrainer.fit`.
image

These changes are copied from [examples/split_placement/split_monkey_patch.py](https://github.com/volcengine/verl/blob/e47718f6d582b9741f621bc7b06c6581d303bc0e/examples/split_placement/split_monkey_patch.py#L116-L126),  because the `update_actor` and `update_critic` would be non-blocking as described in https://github.com/volcengine/verl/tree/main/examples/split_placement#step-2-make-the-models-executed-asynchronously. I didn't add any new things here, I just copied bits from both `RayTrainer.fit` and `examples/split_placement/split_monkey_patch.py` to make it consistent with the original logic.

PeterSH6 (approved):

PR #284, opened 2025-02-15
[misc] feat: support offload parameter and optimizer during rollout

PeterSH6:
- Fixed FSDP1 model offload
- With `actor_rollout_ref.actor.fsdp_config.param_offload=True \` and `actor_rollout_ref.actor.fsdp_config.optimizer_offload=True \ `. The GPU memory utilization can increase to 0.9
- With actor, critic and reference offload all enabled, there will only be one model copy at a time in the GPU memory. Therefore, we can further increase the `micro_batch_size_per_gpu` or `max_token_per_gpu`

**Specifically:**
- During rollout, only rollout model and KVCache are in the GPU memory.
- During critic compute values, only the critic model will stay in the GPU memory while its optimizer and other model states are in CPU main memory
- During actor update, the actor model, optimizer are stored on GPU while the reference model and critic model, critic optimizer are offloaded to CPU.

### DIFF
--- a/verl/utils/fsdp_utils.py
+++ b/verl/utils/fsdp_utils.py
@@ -20,6 +20,8 @@ import itertools
 import os
 from contextlib import contextmanager
 from torch.distributed.fsdp.wrap import size_based_auto_wrap_policy, transformer_auto_wrap_policy
+from torch.distributed.fsdp import FullyShardedDataParallel as FSDP
+from torch.distributed.fsdp._runtime_utils import _lazy_init
 from transformers.trainer_pt_utils import get_module_class_from_name
 import torch
 import torch.nn as nn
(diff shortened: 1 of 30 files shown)


verl/utils/fsdp_utils.py
```diff
@@ -105,58 +107,61 @@ def lambda_policy_fn(module):
...
-def offload_fsdp_param_and_grad(module, offload_grad=False):
-    for _, param in module.named_parameters():
-        if hasattr(param, "_local_shard"):
-            param._local_shard = param._local_shard.to("cpu", non_blocking=True)
-        param.data = param.data.to('cpu', non_blocking=True)
-        if offload_grad and param.grad is not None:
-            param.grad = param.grad.to("cpu", non_blocking=True)
-    torch.cuda.empty_cache()
+@torch.no_grad()
+def offload_fsdp_model_to_cpu(model: FSDP, empty_cache: bool = True):
```
vermouth1992, line 111: Shall we write a unit test for these two functions?

vermouth1992 (approved):

verl/utils/fsdp_utils.py
PeterSH6: Will add a CI in the future

PR #286, opened 2025-02-16
distro: make liger-kernel optional. do not rely on requirement.txt during setup

eric-haibin-lin:


pyproject.toml
```diff
@@ -36,22 +36,25 @@ dependencies = [
...
-    "tensordict",
-    "transformers<4.48",
-    "vllm<=0.6.3",
+    "pandas",
     "peft",
-    "liger-kernel",
+    "pyarrow>=15.0.0",
+    "pybind11",
     "pylatexenc",
+    "ray>=2.10",
```
vermouth1992, line 44: This should be ray>= 2.38

pyproject.toml
```diff
@@ -36,22 +36,25 @@ dependencies = [
...
+    "tensordict<0.6",
+    "transformers",
```
vermouth1992, line 46: transformers has to be < 4.38

eric-haibin-lin: used to be < 4.38 i think

pyproject.toml
eric-haibin-lin: since setup.py and toml specifies the minimal requirements, i think 2.10 is also fine? qwen team has some issues using ray 2.38 so i want to be less restrictive here

vermouth1992: It looks fine to me

pyproject.toml
vermouth1992: Yes, there are some issues when transformers >= 4.38 for verl main

eric-haibin-lin:
```
# transformers 4.47.0 contains the following bug:
# AttributeError: 'Gemma2Attention' object has no attribute '_flash_attn_uses_top_left_mask'
RUN pip3 install --no-cache-dir \
    torch==2.4.0 \
    accelerate \
    codetiming \
    dill \
    hydra-core \
    numpy \
    pybind11 \
    tensordict \
    "transformers==4.46.3"
```
Just tested again the latest transformers. gemma2 works fine now

PeterSH6: Should be >= 4.48 (not 4.38?)

vermouth1992: The requirement is removed

PR #301, opened 2025-02-18
[megatron] fix: critic and reward model load tokenizer from config

HollowMan6:
Currently, the worker will fail if the critic or reward model path doesn't contain a tokenizer. This PR tries to fix this by loading tokenizer from the config for the previously mentioned case.

- For the critic model, we fall back to load from `critic.model.tokenizer_path`.
- For the reward model, we first fall back to load from `reward_model.model.rm_tokenizer`, and then `reward_model.model.input_tokenizer` if that is not set.


HollowMan6:
Just fixed the yapf formatting issue

verl/workers/megatron_workers.py
```diff
@@ -639,15 +644,31 @@ def __init__(self, config):
...
         local_path = copy_local_path_from_hdfs(model_path)
-        self.tokenizer = hf_tokenizer(local_path)
+        try:
+            self.tokenizer = hf_tokenizer(local_path)
+        except OSError:
+            # If the model path doesn't contain a tokenizer, we use rm_tokenizer or sft_tokenizer
+            if rm_tokenizer:
+                self.tokenizer = rm_tokenizer
+            else:
+                self.tokenizer = sft_tokenizer
```
eric-haibin-lin: thanks for the contribution! I'd suggest raise the exception on oserror if rm_tokenizer is None, instead of fallback to the sft tokenizer

HollowMan6: Thank you for your suggestion! I've done it accordingly and raised a `ValueError` instead to make it more accurate. When no tokenizer is found in the model path, `rm_tokenizer` should be provided by the user, and it's not a system error.

verl/single_controller/base/megatron/worker.py
```diff
@@ -39,7 +39,7 @@ def get_megatron_rank_info(self):
         info = DistRankInfo(tp_rank=tp_rank, dp_rank=dp_rank, pp_rank=pp_rank, cp_rank=cp_rank)
         return info
 
-    def _init_hf_config_and_tf_config(self, model_path, dtype, override_model_config, override_transformer_config):
+    def _init_hf_config_and_tf_config(self, model_path, dtype, override_model_config, override_transformer_config, tokenizer=None):
```
ETOgaosion: I believe that you can put the tokenizer in the second place after `model_path`, and the tokenizer is not so clear, maybe `tokenizer_or_path` is better?

Maybe we can force the caller to set tokenizer, in the raw logic, actually `tokenizer_path == model_path`

verl/single_controller/base/megatron/worker.py
```diff
@@ -49,7 +49,16 @@ def _init_hf_config_and_tf_config(self, model_path, dtype, override_model_config
 
         # Step 1: initialize the tokenizer
         self.local_path = copy_to_local(model_path)
-        self.tokenizer = hf_tokenizer(self.local_path)
+        try:
```
ETOgaosion: Maybe no need to try catch in the first place? Can you try to use the type direction to judge first?

Issue #331, opened 2025-02-21
Tried to run main_generation.py, but it raised KeyError: ConfigAttributeError: Key 'actor' is not in struct.


Di-viner:
Thanks for the great work! 
Tried to run `main_generation.py`, but got a KeyError: ConfigAttributeError: Key 'actor' is not in struct.

similar to `examples/generation/run_deepseek_v2_lite_math.sh`:
```
python3 -m verl.trainer.main_generation \
    trainer.nnodes=1 \
    trainer.n_gpus_per_node=2 \
    data.path=... \
    data.prompt_key=prompt \
    data.n_samples=1 \
    data.output_path=... \
    model.path=... \
    +model.trust_remote_code=True \
    rollout.temperature=1.0 \
    rollout.top_k=50 \
    rollout.top_p=0.7 \
    rollout.prompt_length=2048 \
    rollout.response_length=2048 \
    rollout.tensor_model_parallel_size=2 \
    rollout.gpu_memory_utilization=0.8
```

but got a KeyError:
```
Traceback (most recent call last):
  File "/verl/verl/trainer/main_generation.py", line 68, in main
    wg.init_model()
  File "/verl/verl/single_controller/ray/base.py", line 42, in func
    output = ray.get(output)
  File "/verl/lib/python3.9/site-packages/ray/_private/auto_init_hook.py", line 21, in auto_init_wrapper
    return fn(*args, **kwargs)
  File "/verl/lib/python3.9/site-packages/ray/_private/client_mode_hook.py", line 103, in wrapper
    return func(*args, **kwargs)
  File "/verl/lib/python3.9/site-packages/ray/_private/worker.py", line 2755, in get
    values, debugger_breakpoint = worker.get_objects(object_refs, timeout=timeout)
  File "/verl/lib/python3.9/site-packages/ray/_private/worker.py", line 908, in get_objects
    raise value
ray.exceptions.ActorDiedError: The actor died because of an error raised in its creation task, ray::5ZAuQSActorRolloutRefWorker_0:1:ActorRolloutRefWorker.__init__() (pid=328776, ip=10.17.176.67, actor_id=08452910c3c019b8e00ed93801000000, repr=<verl.workers.fsdp_workers.ActorRolloutRefWorker object at 0x7ff00c3f2850>)
  File "/verl/verl/workers/fsdp_workers.py", line 88, in __init__
[...]

PR #351, opened 2025-02-23
[fix] Improve the params template for generation

BearBiscuit05:
fix the issue[#331](https://github.com/volcengine/verl/issues/331)

### DIFF
--- a/.github/workflows/vllm.yml
+++ b/.github/workflows/vllm.yml
@@ -54,5 +54,5 @@ jobs:
       - name: Run QWen 0.5B generation test
         run: |
           cd tests/generation
-          bash ./run_gen_qwen05.sh 1 $HOME/data/gen/qwen_05_gen_test.parquet
+          bash ./run_gen_qwen05.sh 4 $HOME/data/gen/qwen_05_gen_test.parquet
           rm -rf $HOME/data/gen/qwen_05_gen_test.parquet
--- a/.github/workflows/vllm.yml
+++ b/.github/workflows/vllm.yml
@@ -51,3 +51,8 @@ jobs:
           pip3 install --upgrade vllm
           cd tests/rollout
           torchrun --standalone --nnodes=1 --nproc_per_node=4 $(which pytest) -s test_vllm_spmd.py
+      - name: Run QWen 0.5B generation test
+        run: |
+          cd tests/generation
+          bash ./run_gen_qwen05.sh 1 $HOME/data/gen/qwen_05_gen_test.parquet
+          rm -rf $HOME/data/gen/qwen_05_gen_test.parquet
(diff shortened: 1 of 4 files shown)


vermouth1992:
Could you help add a test of QWen 0.5b generation to protect this functionality?

BearBiscuit05:
Sure, I used Qwen0.5B for testing on a single machine. But in which directory under the "test" directory should I add the test?

vermouth1992:
Could you create a new folder under test with name "generation". Under the folder, create a new bash script that runs QWen 0.5b for generation. And call the generation script here https://github.com/volcengine/verl/blob/main/.github/workflows/vllm.yml#L49 by creating a new test item. Thanks!

BearBiscuit05:
Running with 1 GPU works normally, but when setting nproc_per_node > 1, it produces the error `Duplicate GPU detected: rank 0 and rank 1 both on CUDA device 31000`. I'm unsure whether this is caused by parameter configuration issues or a hardware-related problem. Could you help me identify the root cause?

vermouth1992:
Could you check the version of ray?

BearBiscuit05:
Ray version is 2.10, and I ran PPO on 2 * A100 successfully. So I think it may be a parameter problem. I will check it tomorrow.

vermouth1992:
You can either set max_colocate_count to 1 https://github.com/volcengine/verl/blob/main/verl/single_controller/ray/base.py#L55 or upgrade ray to the latest to resolve this problem

vermouth1992:
Yes, please set tp=2

BearBiscuit05:
done, the script successfully ran on 4 GPUs with TP=2.

vermouth1992 (approved):

BearBiscuit05:
I found that when num_gpus == TP, due to dp == 1, the filling of the dummy won't be triggered, which causes an error when calling `wg.generate_sequences(data)` for dispatch. I'm not sure whether the dummy is still needed or if dispatch is not required when dp == 1. I'm not very familiar with Ray for now.
errer happens when gpus=2,tp=2
```bash
Traceback (most recent call last):
  File "/verl/verl/trainer/main_generation.py", line 110, in main
    output = wg.generate_sequences(data)
  File "/verl/verl/single_controller/ray/base.py", line 39, in func
    args, kwargs = dispatch_fn(self, *args, **kwargs)
  File "/verl/verl/single_controller/base/decorator.py", line 276, in dispatch_dp_compute_data_proto
    splitted_args, splitted_kwargs = _split_args_kwargs_data_proto(worker_group.world_size, *args, **kwargs)
  File "/verl/verl/single_controller/base/decorator.py", line 50, in _split_args_kwargs_data_proto
    splitted_args.append(arg.chunk(chunks=chunks))
  File "/verl/verl/protocol.py", line 499, in chunk
    assert len(
AssertionError: only support equal chunk. Got size of DataProto 39 and chunk 2.
```

asirgogogo:
same here

PR #353, opened 2025-02-23
[Fix] Deprecate `val_batch_size`

tongyx361:
Validation datasets are sent to inference engines as a whole batch, which will schedule the memory themselves.

- [x] Remove `val_batch_size` from examples
- [x] Set default values of `val_batch_size` in configs as `null` and add DEPRECATED comments
- [x] Add deprecation warnings about `val_batch_size` in `_validate_config`

### DIFF
--- a/verl/trainer/ppo/ray_trainer.py
+++ b/verl/trainer/ppo/ray_trainer.py
@@ -465,6 +465,11 @@ class RayPPOTrainer(object):
                 assert config.critic.model.use_remove_padding, \
                     "When using sequence parallelism for critic, you must enable `use_remove_padding`."
 
+        if config.get('val_batch_size', None) is not None:
+            print(
+                f"WARNING: val_batch_size is deprecated. Validation datasets are sent to inference engines as a whole batch, which will schedule the memory themselves."
+            )
+
         print("[validate_config] All configuration checks passed successfully!")
 
     def _create_dataloader(self):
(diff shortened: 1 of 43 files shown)


verl/trainer/ppo/ray_trainer.py
```diff
@@ -465,6 +465,11 @@ def check_mutually_exclusive(mbs, mbs_per_gpu, name: str):
                 assert config.critic.model.use_remove_padding, \
                     "When using sequence parallelism for critic, you must enable `use_remove_padding`."
 
+        if config.get('val_batch_size', None) is not None:
```
PeterSH6: This should be `config.data`?

verl/workers/actor/megatron_actor.py
```diff
@@ -136,6 +136,11 @@ def _validate_config(self, config) -> None:
         """Validate config options not implemented for Megatron backend"""
         assert config.get('ulysses_sequence_parallel_size', 1) == 1
 
+        if config.get('val_batch_size', None) is not None:
```
PeterSH6: There are no val_batch_size in the actor field, but only in the data field

verl/workers/critic/megatron_critic.py
```diff
@@ -78,6 +78,11 @@ def _validate_config(self, config) -> None:
         """Validate config options not implemented for Megatron backend"""
         assert config.get('ulysses_sequence_parallel_size', 1) == 1
 
+        if config.get('val_batch_size', None) is not None:
+            print(
+                f"WARNING: val_batch_size is deprecated. Validation datasets are sent to inference engines as a whole batch, which will schedule the memory themselves."
+            )
+
```
PeterSH6: Same?

vermouth1992 (approved):

Issue #298, opened 2025-02-18
How to load trained ckpt and eval with it on specific dataset?


jankinf:
Great repo! Could you please provide some guidances on how to load a trained checkpoint and perform evaluation on a specific dataset. The documentation seems unclear about this workflow.

jankinf:
Additional Information: 
The save_checkpoint function in https://github.com/Jiayi-Pan/TinyZero/blob/main/verl/workers/fsdp_workers.py#L478 outputs checkpoints in HuggingFace format (I'm not sure if the code in this repository corresponds to a specific historical version of verl). Therefore, it can be loaded directly using `AutoModelForCausalLM.from_pretrained`, which seems convenient and works well. The output model directory structure is as follows:

```
.
└── actor
    └── global_step_100
        ├── added_tokens.json
        ├── config.json
        ├── generation_config.json
        ├── merges.txt
        ├── model-00001-of-00003.safetensors
        ├── model-00002-of-00003.safetensors
        ├── model-00003-of-00003.safetensors
        ├── model.safetensors.index.json
        ├── special_tokens_map.json
        ├── tokenizer_config.json
        ├── tokenizer.json
        └── vocab.json
```

However, the save_checkpoint used in this repository generates distributed rank replicas (https://github.com/volcengine/verl/blob/main/verl/utils/checkpoint/fsdp_checkpoint_manager.py#L134), which can be loaded using provided `load_checkpoint`. If I try to execute `AutoModelForCausalLM.from_pretrained` first and then load a model from one of the ranks(for example `model_world_size_2_rank_0.pt`), the code will throw an error. The output model directory structure is as follows:

```
.
├── global_step_100
│   ├── actor
│   │   ├── extra_state_world_size_2_rank_0.pt
│   │   ├── extra_state_world_size_2_rank_1.pt
│   │   ├── huggingface
│   │   │   ├── added_tokens.json
│   │   │   ├── config.json
│   │   │   ├── merges.txt
│   │   │   ├── special_tokens_map.json
│   │   │   ├── tokenizer_config.json
│   │   │   ├── tokenizer.json
│   │   │   └── vocab.json
│   │   ├── model_world_size_2_rank_0.pt
│   │   ├── model_world_size_2_rank_1.pt
│   │   ├── optim_world_size_2_rank_0.pt
│   │   └── optim_world_size_2_rank_1.pt
│   └── data.pt
└── latest_checkpointed_iteration.txt
```

[...]

PR #359, opened 2025-02-24
Support FSDP model ckpt loading and do evaluation on specific dataset: Issue #298

jankinf:


jankinf:
Sorry, I'm not quite clear about the core purpose of main_generation. If it's only for inference and evaluation, it shouldn't need an actor (for training?), right? Am I misunderstanding something?

jankinf:
main_generation seems to have logic that is not properly decoupled. If it's only for generation purposes, the config should only need to provide rollout parameters. If actor parameters are provided, it automatically loads an optimizer, which doesn't seem necessary.

jankinf:
incorporate fsdp ckpt loading func into main_generation. fixes #298

jankinf:
any feedback?

scripts/eval_on_data.sh
```diff
@@ -0,0 +1,15 @@
+export HYDRA_FULL_ERROR=1
```
eric-haibin-lin: I suggest we move this to folder `example/evaluation/`

scripts/main_generation.sh
```diff
@@ -0,0 +1,19 @@
+export HYDRA_FULL_ERROR=1
```
eric-haibin-lin, line 1: `example/generation/hf_generate_qwen_from_fsdp_ckpt.sh`

verl/trainer/eval_on_data.py
```diff
@@ -0,0 +1,198 @@
+import hydra
```
eric-haibin-lin, line 1: In principle modules under verl namespace are expected to be modular and minimal, for others to extend (still not there yet but that's goal). Currently we have:
- sft 
- generation 
- eval based on generated results (from https://github.com/volcengine/verl/pull/777, i just recently merged it without noticing your PR, sorry)
- on policy RL

There are other use cases that can simply combine these basic modules, such as distillation (first gen then SFT). And i think the eval on data script is similar to gen + eval. 

In that case I suggest we put this trainer implementation (eval_on_data.py & yaml) in examples/evaluation as examples, instead of inside the verl namespace. Do you think that make sense?

verl/trainer/main_generation.py
```diff
@@ -14,25 +14,17 @@
 """
 Generate responses given a dataset of prompts
 """
+import os
 import ray
-import numpy as np
 import hydra
-import os
-
-os.environ['NCCL_DEBUG'] = 'WARN'
```
eric-haibin-lin: Is the removal intentional?

verl/workers/fsdp_workers.py
```diff
@@ -141,6 +141,7 @@ def __init__(self, config: DictConfig, role: str):
 
     def _build_model_optimizer(self,
                                model_path,
+                               fsdp_model_path,
```
eric-haibin-lin: please add this as the last optional argument

eric-haibin-lin: since this is for actor only, let's rename the arg and config to actor_fsdp_model_path

eric-haibin-lin (changes requested):
Hi @jankinf thank you for your contribution! Really appreciate it and sorry that I missed your PR last month. Please see my comments below

scripts/main_generation.sh
```diff
@@ -0,0 +1,19 @@
...
+export NCCL_DEBUG='WARN'
+export TOKENIZERS_PARALLELISM='true'
```
jankinf, line 3: export NCCL_DEBUG='WARN'
export TOKENIZERS_PARALLELISM='true'

I added these two lines here and remove the env setting commands in the `main_generation.py`.

scripts/eval_on_data.sh
jankinf: LGTM

scripts/main_generation.sh
jankinf: LGTM

verl/trainer/eval_on_data.py
jankinf: It makes sense, I will move eval_on_data.py and yaml to examples/evaluations

verl/trainer/main_generation.py
jankinf: I added the following two lines in main_generation.sh, so I can remove env settings here
export NCCL_DEBUG='WARN'
export TOKENIZERS_PARALLELISM='true'

verl/workers/fsdp_workers.py
jankinf: done

jankinf:
fixed

w568w:
May I kindly ask whether this pull request is anywhere closer to get merged? 

It is really weird that FSDP checkpoints, in which format _every_ official documentation and recipe recommend users to save their training artifacts, is **incompatible** with the `verl.trainer.main_generation` script. 🤦

jankinf:
@w568w I tested this when submitting the PR and it worked at the time. However, there hasn't been any feedback from the official team for a while. You can try this code, but I can't guarantee it will work since it's been some time since my last test.

PR #386, opened 2025-02-26
[feat] Initial support for VLMs, add Qwen2.5VL GRPO example

hiyouga:
## What does this PR do?

This PR migrates the feature of RL on VLMs in our implementation in [EasyR1](https://github.com/hiyouga/EasyR1) fork back to veRL. We have validated this feature using Qwen2.5-VL 7B model on 8*H100 GPUs. The configuration and data processing script are provided along this PR for easy reproducing.

## How to reproduce?

1. Download and preprocess the dataset

```bash
python3 examples/data_preprocess/geo3k.py --local_dir ~/data/geo3k
```

2. Start GRPO training

```bash
bash examples/grpo_trainer/run_qwen2_5_vl-7b.sh
```

## Dependencies

- vllm>=0.7.3
- transformers>=4.49.0
- [qwen-vl-utils](https://pypi.org/project/qwen-vl-utils/)
- [mathruler](https://pypi.org/project/mathruler/)

### New dataflow for multimodal RL

In this PR, we introduce two new concepts in the dataflow, `multi_modal_data` and `multi_modal_inputs`. The former means the multi-modal features required by the **rollout** worker (such as vLLM), while the latter means the multi-modal features required by the **actor/critic** worker (such as an HF model). They are different because the rollout and actor workers have their own data format requirements.

Taking Qwen2-VL + huggingface + vLLM as an example, the data structure should be:

- **multi_modal_data**: {"image": [PIL.Image, PIL.Image, ...]}
- **multi_modal_inputs**: {"pixel_values": torch.Tensor, "image_grid_thw": torch.Tensor}

Both of them are converted to numpy objects and placed in the non-tensor batch in DataProto.

This design can be extended to other modalities/VLMs easily due to the agnostic of models.

### Other changes

- Data
  - Support pre-processing the [Geometry3k](https://huggingface.co/datasets/hiyouga/geometry3k) dataset.
  - Support `config.data.image_key`, which should be **a list of Pillow images**.

- Actor/Ref/Critic
  - Support `multi_modal_inputs`.
  - Process position ids to adapt to the m-rope .

- Rollout
  - Update dtensor weight loader to adapt to the Qwen2-VL architecture in vLLM 0.7+.
  - Support `multi_modal_data`.
  - Use `raw_prompt_ids` as the vLLM inputs to **avoid unpadding** the input ids.

- Reward Manager
  - Add **mathruler** for more accurate math scores on the Geometry 3k dataset

- Models
  - Support calculating the position ids for the m-rope in Qwen2-VL.
  - Support removing padding in flash attention2 for m-rope (transformers itself **does not support it**).

- Sharding Manager
  - Support all-gathering the non-tensor batch.

- FSDP Workers / Checkpoint Merger
  - Support `AutoModelForVision2Seq` at model initialization.

Note: The Ulysses parallelism is not completed yet. We will support it in the next update.

## Performance

We provide the estimated MFU of the language model part for H100 GPUs. These values are lower than the actual ones because **we did not compute the FLOPs of the vision tower part**.

- `remove_padding=False`: MFU ~7%
- `remove_padding=True`: MFU ~20%

The training and test reward score curves are presented as follows.

image

## Who can review?

@vermouth1992 @PeterSH6

### DIFF
--- a/verl/trainer/main_ppo.py
+++ b/verl/trainer/main_ppo.py
@@ -46,8 +46,9 @@ def main_task(config, compute_score=None):
     local_path = copy_to_local(config.actor_rollout_ref.model.path)
 
     # instantiate tokenizer
-    from verl.utils import hf_tokenizer
+    from verl.utils import hf_tokenizer, hf_processor
     tokenizer = hf_tokenizer(local_path)
+    processor = hf_processor(local_path, use_fast=True)  # used for multimodal LLM, could be none
 
     # define worker classes
     if config.actor_rollout_ref.actor.strategy == 'fsdp':
@@ -117,6 +118,7 @@ def main_task(config, compute_score=None):
 
     trainer = RayPPOTrainer(config=config,
                             tokenizer=tokenizer,
+                            processor=processor,
                             role_worker_mapping=role_worker_mapping,
                             resource_pool_manager=resource_pool_manager,
                             ray_worker_group_cls=ray_worker_group_cls,
--- a/verl/utils/dataset/rl_dataset.py
+++ b/verl/utils/dataset/rl_dataset.py
@@ -14,32 +14,29 @@
 
 from omegaconf import ListConfig
 import os
-from typing import List, Union
+from typing import List, Union, Optional
 import copy
 import pandas as pd
+from collections import defaultdict
 
 import torch
 import numpy as np
 from torch.utils.data import Dataset
-from transformers import PreTrainedTokenizer
+from transformers import PreTrainedTokenizer, ProcessorMixin
 
 from verl.utils.model import compute_position_id_with_mask
 import verl.utils.torch_functional as verl_F
 
 
 def collate_fn(data_list: list[dict]) -> dict:
-    tensors = {}
-    non_tensors = {}
+    tensors = defaultdict(list)
+    non_tensors = defaultdict(list)
 
     for data in data_list:
         for key, val in data.items():
             if isinstance(val, torch.Tensor):
-                if key not in tensors:
-                    tensors[key] = []
                 tensors[key].append(val)
             else:
-                if key not in non_tensors:
-                    non_tensors[key] = []
                 non_tensors[key].append(val)
 
     for key, val in tensors.items():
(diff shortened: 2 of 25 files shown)


khazic:
good job

vermouth1992:
Item TODO: I guess it's better to provide a documentation on how to extend to other VLMs like ulysses/rmpad, etc,. so that other contributors can add their interested VLMs.

hiyouga:
@vermouth1992 Sure, but it requires non-trivial effort to build a unified guideline for all the VLMs, so we are considering improving the document in the next PR.

verl/trainer/main_ppo.py
```diff
@@ -46,8 +46,9 @@ def main_task(config, compute_score=None):
     local_path = copy_local_path_from_hdfs(config.actor_rollout_ref.model.path)
 
     # instantiate tokenizer
-    from verl.utils import hf_tokenizer
+    from verl.utils import hf_tokenizer, hf_processor
     tokenizer = hf_tokenizer(local_path)
+    processor = hf_processor(local_path, use_fast=True)
```
PeterSH6: It would be better to add a comment indicating that hf_processor is for multi-modal scenario

verl/utils/dataset/rl_dataset.py
```diff
@@ -138,18 +156,53 @@ def __getitem__(self, item):
 
         prompt_with_chat_template = self.tokenizer.apply_chat_template(chat, add_generation_prompt=True, tokenize=False)
 
+        if 'images' in row_dict:  # expand image token
+            raw_prompt = prompt_with_chat_template.replace('<image>', '<|vision_start|><|image_pad|><|vision_end|>')
```
PeterSH6, line 164: [DO NOT CONSIDER IN THIS PR]
A note for future refactoring based on our discussion:
- This line may model specific and inference engine related

We will utilize an input processor to deal with different inputs from different multi-modal LLMs.

verl/utils/dataset/rl_dataset.py
```diff
@@ -138,18 +156,53 @@ def __getitem__(self, item):
 
         prompt_with_chat_template = self.tokenizer.apply_chat_template(chat, add_generation_prompt=True, tokenize=False)
 
+        if 'images' in row_dict:  # expand image token
```
PeterSH6: image_key

verl/utils/dataset/rl_dataset.py
```diff
@@ -138,18 +156,53 @@ def __getitem__(self, item):
...
+
         input_ids, attention_mask = verl_F.tokenize_and_postprocess_data(prompt=prompt_with_chat_template,
                                                                          tokenizer=self.tokenizer,
                                                                          max_length=self.max_prompt_length,
                                                                          pad_token_id=self.tokenizer.pad_token_id,
                                                                          left_pad=True,
                                                                          truncation=self.truncation)
 
-        position_ids = compute_position_id_with_mask(attention_mask)
+        if "images" in row_dict:
```
PeterSH6: image_key

PeterSH6 (commented):
Brilliant work!

hiyouga:
@PeterSH6 Thanks for your review. I have resolve these comments, could you check the latest PR?

PeterSH6:
LGTM! 
The current PR could support integrating most MLLMs by matching the `hf_processors` output keys to the model forward function signature. So elegant.

One last thing, could you add an E2E CI similar to [E2E GSM8K CI](https://github.com/volcengine/verl/blob/main/.github/workflows/e2e_gsm8k.yml)? 
The CI can be run on `l20-0` using your example data and only run 1 step by specifying `trainer.total_training_steps=1` similar to this [script](https://github.com/volcengine/verl/blob/main/tests/e2e/run_qwen_gsm8k_function_rm.sh)

eric-haibin-lin:
@hiyouga could you help resolve the conflict? thanks!

hiyouga:
@eric-haibin-lin Sure! The conflict is solved

PeterSH6 (dismissed):

eric-haibin-lin:
the model_rmpad test hang again

vermouth1992 (approved):

PeterSH6 (approved):

PR #392, opened 2025-02-26
megatron：Update megatron-lm to `core_r0.11.0`

ISEEKYAN:
# Support Megatron mcore 0.11

## Description
This PR introduces official support for Megatron mcore 0.11 with the following updates:
- Upgraded Megatron to version `core_r0.11.0`
- Applied compatibility patch `patches/mcore_r0.11.patch`
- Removed legacy version support for cleaner implementation

Special thanks to @chendong-1998 for:
- Original Megatron upgrade from 0.4 to 0.6 (#93f6a7e)

## Compatibility Notes
Current implementation requires careful handling due to dependency conflicts:
- `megatron-core==0.11.0` requires torch>=2.6
- `vllm==0.6.3` requires torch==2.4

Installation constraints:
1. Must use vllm's torch dependency (2.4) as baseline
2. Do NOT run `pip install -e .` in mcore directory (will upgrade torch to 2.6)
3. Apply compatibility patch manually after installation

### test with `verl/examples/ppo_trainer/run_deepseek_megatron.sh`
image


verl/models/llama/megatron/checkpoint_utils/llama_loader.py
```diff
@@ -12,6 +12,8 @@
 # See the License for the specific language governing permissions and
```
eric-haibin-lin, line 12: could u keep the v0.4 patch file for now in case others are want to run v0.4 for comparison. thanks!

eric-haibin-lin: we can remove the v0.4 patch after the next stable release of verl

verl/utils/megatron_utils.py
```diff
@@ -23,12 +25,18 @@
...
-# from megatron.core.distributed import DistributedDataParallelConfig
+from megatron.core.distributed import DistributedDataParallelConfig
 from megatron.core.distributed import DistributedDataParallel as DDP
 from megatron.core.enums import ModelType
+from megatron.core import ModelParallelConfig
+from megatron.core.optimizer import OptimizerConfig
+
+
+def get_model_config(model):
+    return get_attr_wrapped_model(model, 'megatron_config', allow_none=False)
```
PeterSH6, line 39: I think `get_model_config()` is no longer necessary if we change the customized model class element.
Currently in `ParallelLlamaForCausalLMRmPadPP`
- config -> Huggingface config
- megatron_config -> `megatron.core.ModelParallelConfig`

We could rename `megatron_config` to `config` and rename `config` to `hf_config`

verl/utils/megatron_utils.py
```diff
@@ -95,22 +103,30 @@ def get_model(model_provider_func, model_type=ModelType.encoder_or_decoder, wrap
...
-                module=model_chunk,
-                data_parallel_group=mpu.get_data_parallel_group(with_context_parallel=True),
-                accumulate_allreduce_grads_in_fp32=True,
-                overlap_grad_reduce=False,
-                use_distributed_optimizer=True,
-                disable_bucketing=(model_chunk_idx > 0)) for (model_chunk_idx, model_chunk) in enumerate(model)
-        ]
+        ddp_models = []
+        for model_chunk_idx, model_chunk in enumerate(model):
+            # __import__("ipdb").set_trace()
```
PeterSH6: A redundant debug line

verl/utils/megatron_utils.py
```diff
@@ -216,7 +225,7 @@ class FakeTimers:
     """Disable All Megatron Timing with FakeTimers"""
 
     def __init__(self):
-        from megatron.timers import DummyTimer
+        from megatron.core.timers import DummyTimer
```
PeterSH6: Do we still need Timers in MCore v0.11?
This FakeTimer is mainly for `optimizer.step()` in MCore v0.4.
As there's no need to use timer in `optimizer.step()` in MCore v0.11, I suggest simply delete this class.
(Also delete its usage in L212)

verl/workers/sharding_manager/megatron_vllm.py
```diff
@@ -81,11 +83,24 @@ def __init__(self, model_provider) -> None:
 
     def _build_param_buffer(self, pp_rank):
         """Build the parameter buffer in each pp rank"""
-        model = self.pp_models[pp_rank]
-        weight_buffer_meta = get_weight_buffer_meta_from_module(model)
-        self.memory_buffers[pp_rank] = build_memory_buffer(weight_buffer_meta)
+        if pp_rank == self._pp_rank:
+            from verl.utils.memory_buffer import MemoryBuffer
+            # The code here is very hard-coded, based on the following assumptions:
```
PeterSH6, line 88: That's true. Just wondering if the `MemoryBuffer` will complicate the weight synchronization process when we enable EP?
If so, we can abandon the `MemoryBuffer` and change to per-parameter synchronization

verl/models/llama/megatron/checkpoint_utils/llama_loader.py
ISEEKYAN: OK，I will add that back

verl/utils/megatron_utils.py
```diff
@@ -139,7 +155,7 @@ def unwrap_model(model, module_instances=ALL_MODULE_WRAPPER_CLASSNAMES):
 
 def convert_config(hf_config: PretrainedConfig, megatron_config) -> TransformerConfig:
     print(f'megatron config {megatron_config}')
-    dt = PrecisionType.to_dtype(megatron_config['param_dtype'])
+    dt = torch.bfloat16
```
PeterSH6: How to pass the parameter dtype of Megatron in the current implementation?

verl/models/llama/megatron/checkpoint_utils/llama_saver.py
```diff
@@ -12,17 +12,21 @@
...
 from typing import Optional
 import torch.distributed as dist
+
+import megatron
 from megatron import get_args
+from megatron.core import mpu
+from megatron.model import Float16Module
+from megatron.model import DistributedDataParallel as LocalDDP
+
+from megatron.training.utils import print_rank_0, unwrap_model
```
PeterSH6, line 29: We copied several module and util functions from the Megatron-LM package into the `megatron_utils.py`.
It would be better if we can remove importing from outside `megatron.core`

PeterSH6 (commented):
Brilliant work!

verl/utils/megatron_utils.py
ISEEKYAN: if modifying L22-L24 in `verl/utils/torch_dtypes.py`, like 
```
HALF_LIST = [16, "16", "fp16", "float16", torch.float16]
FLOAT_LIST = [32, "32", "fp32", "float32", torch.float32]
BFLOAT_LIST = ["bf16", "bfloat16", torch.bfloat16]
```
then the args could be delivered by `dt = PrecisionType.to_dtype(megatron_config.params_dtype) `

verl/utils/megatron_utils.py
ISEEKYAN: I tried, it works without the timer, I will delete the whole timer

verl/utils/megatron_utils.py
ISEEKYAN: ok, delete it

verl/models/llama/megatron/checkpoint_utils/llama_saver.py
ISEEKYAN: sure, fixing this

verl/utils/megatron_utils.py
ISEEKYAN: I agree the multiple config types in `ParallelLlamaForCausalLMRmPadPP` can be confusing. However, directly renaming `megatron_config` to `config` and `config` to `hf_config` might introduce breaking changes to existing code that references these properties.
I suggest merging the current PR first, then handling the renaming separately with proper migration planning

verl/workers/sharding_manager/megatron_vllm.py
vermouth1992: I agree to abandon MemoryBuffer to change to per-parameter synchronization

vermouth1992: Essentially, weight binding can be generalized to per parameter all-gather and redistribute

verl/utils/megatron_utils.py
PeterSH6: I agree

vermouth1992:
The issue of vllm can be ignored as it is fixed on main

PeterSH6 (approved):

vermouth1992 (approved):

Cppowboy:
Where the patches/mcore_r0.11.patch file is located?

PR #403, opened 2025-02-27
[fix] Communicating for non_tensor_batch in fsdp_vllm.py

UbeCc:
Current verl only perform communication for non_tensor_batch at `fsdp_ulysses.py`. I support the communication for `fsdp_vllm.py`

### DIFF
--- a/verl/protocol.py
+++ b/verl/protocol.py
@@ -27,6 +27,7 @@ import torch
 import tensordict
 from tensordict import TensorDict
 from torch.utils.data import DataLoader, Dataset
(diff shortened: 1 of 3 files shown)


vermouth1992:
Actually, could you move this functionality to a class method of DataProto? Thanks.

UbeCc:
> Actually, could you move this functionality as a class method of DataProto? Thanks.

Fine, it's already done

hiyouga:
I think this feature already exists in https://github.com/volcengine/verl/pull/386

vermouth1992:
This feature is already implemented in #497

PR #427, opened 2025-02-28
docs: add hf ckpt to faq, and include verl apis in the website

eric-haibin-lin:
Now APIs can be displayed: 

image

### DIFF
--- a/verl/protocol.py
+++ b/verl/protocol.py
@@ -84,7 +84,7 @@ def union_tensor_dict(tensor_dict1: TensorDict, tensor_dict2: TensorDict) -> Ten
     return tensor_dict1
 
 
-def union_numpy_dict(tensor_dict1: dict[np.ndarray], tensor_dict2: dict[np.ndarray]) -> dict[np.ndarray]:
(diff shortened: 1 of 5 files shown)


verl/protocol.py
```diff
@@ -84,7 +84,7 @@ def union_tensor_dict(tensor_dict1: TensorDict, tensor_dict2: TensorDict) -> Ten
     return tensor_dict1
 
 
-def union_numpy_dict(tensor_dict1: dict[np.ndarray], tensor_dict2: dict[np.ndarray]) -> dict[np.ndarray]:
```
vermouth1992, line 87: The current standard of type hint actually prefers dict over typing.Dict, list over typing.List

vermouth1992: This starts from python 3.10.  https://docs.python.org/3/library/typing.html

eric-haibin-lin: ok. The doc server is not configured with py3.10 and threw errors. I'll update to 310 and see if we can build the doc

eric-haibin-lin: updated. rendering works as expected

vermouth1992 (approved):

PR #436, opened 2025-03-01
Fix a bug for grad norm computation in SFT. Get loss 10% lower after fixing this bug

ZihanWang314:
Thanks for your great work on veRL! However, the current implementation in SFT for grad accumulation seems wrong. The loss should be divided first, then do backward. Experiments show that fixing this bug can get loss 10% lower (from 2.7 -> 2.4) on some tasks

### DIFF
--- a/verl/trainer/fsdp_sft_trainer.py
+++ b/verl/trainer/fsdp_sft_trainer.py
@@ -394,8 +394,10 @@ class FSDPSFTTrainer(object):
         n_micro_batches = len(micro_batches)
         step_loss = 0
         for micro_batch in micro_batches:
-            loss = self._compute_loss_and_backward(batch=micro_batch) / n_micro_batches
+            loss = self._compute_loss_and_backward(batch=micro_batch, do_backward=False) / n_micro_batches
+            loss.backward()
             step_loss += loss.item()
+            grad_norm = self.fsdp_model.clip_grad_norm_(max_norm=self.config.optim.clip_grad)
 
         self.fsdp_model.clip_grad_norm_(max_norm=self.config.optim.clip_grad)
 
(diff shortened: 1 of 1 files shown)


verl/trainer/fsdp_sft_trainer.py
```diff
@@ -394,8 +394,10 @@ def training_step(self, batch: TensorDict):
         n_micro_batches = len(micro_batches)
         step_loss = 0
         for micro_batch in micro_batches:
-            loss = self._compute_loss_and_backward(batch=micro_batch) / n_micro_batches
+            loss = self._compute_loss_and_backward(batch=micro_batch, do_backward=False) / n_micro_batches
+            loss.backward()
             step_loss += loss.item()
+            grad_norm = self.fsdp_model.clip_grad_norm_(max_norm=self.config.optim.clip_grad)
```
vermouth1992, line 400: Is it correct to perform a global grad clip after all the micro batches forward and backward?

ZihanWang314: Of course yes if the grad norm is divided first. Note that when every micro batch comes in, it only contributes to its fraction in the final batch, both for loss and for grad norm, so I'm very sure to implement correct micro-batching, each micro batch should be forward and backward first (with the loss divided before doing backward), then finally do the norm clip.

ZihanWang314:
image
image
Please also see the loss curve and grad norm curve here. With current implementation, grad norm is very high when micro batch size is smaller than batch size

verl/trainer/fsdp_sft_trainer.py
```diff
@@ -394,8 +394,10 @@ def training_step(self, batch: TensorDict):
         n_micro_batches = len(micro_batches)
         step_loss = 0
         for micro_batch in micro_batches:
-            loss = self._compute_loss_and_backward(batch=micro_batch) / n_micro_batches
+            loss = self._compute_loss_and_backward(batch=micro_batch, do_backward=False) / n_micro_batches
+            loss.backward()
```
xingyaoww, line 398: I think we need to move the .backward() with in the context (https://github.com/volcengine/verl/pull/436/files#diff-07c65794985259af3f0f8e156fcf70119ad136cd42e06b4c6584799b1ebc8f2fR300)

otherwise SP will likely cause OOM issue

verl/trainer/fsdp_sft_trainer.py
hiyouga: It's ok to perform backprop for each microbatch data, but I think the `clip_grad_norm` operation can be done after computing gradients for all micro-batches.

Suggest change:
```python
        for micro_batch in micro_batches:
            loss = self._compute_loss_and_backward(batch=micro_batch, do_backward=False) / n_micro_batches
            loss.backward()
            step_loss += loss.item()

        grad_norm = self.fsdp_model.clip_grad_norm_(max_norm=self.config.optim.clip_grad)
```

Issue #338, opened 2025-02-21
Add Ascend NPU support for veRL


Yikun:
This RFC proposed to provide initial support for veRL on Ascend NPU.

### PR1：Support FSDP worker and vLLM Ascend https://github.com/volcengine/verl/pull/332

A basic PR based on FSDP and Ascend NPU to make GRPO workflow work on Ascend NPU
(based on https://github.com/volcengine/verl/pull/198 which is proposed by @Chendong98 )

Include Single controller, FSDP worker Ascend NPU support, vLLM Ascend support (>=0.7.1)

Co-author: 
  - @as12138 (FSDP worker / vLLM Ascend) 
  - @Chendong98(Single controller) 
  - @celestialli (vLLM Ascend)

### PR2：Support Megatron 0.6 in veRL https://github.com/volcengine/verl/pull/85

Based on PR1
Owner: @Chendong98

### PR3: Support MindSpeed worker in veRL  TBD

Based on PR2
Owner: @Chendong98

Yikun:
We had a internal meeting about Ascend NPU support for veRL, above is the initial info to make things clear, hopes no more misunderstanding.

takagi97:
@Yikun Thanks for your efforts! Can PR https://github.com/volcengine/verl/pull/332 be directly implemented on 910b2c 64GB?

Yikun:
Yes, the initial support will based on community CI which are running on Altlas A2 series, for more info pls see CI info: https://github.com/volcengine/verl/actions/runs/13649493801/job/38154761144

takagi97:
> Yes, the initial support will based on community CI which are running on Altlas A2 series, for more info pls see CI info: https://github.com/volcengine/verl/actions/runs/13649493801/job/38154761144

Thank you for your quick response! I will try it.

PR #448, opened 2025-03-03
[CI] Add e2e_ascend CI

Yikun:
related: https://github.com/volcengine/verl/issues/338


Yikun:
> Access to the path '/data00/tiger/actions-runner/_work/_temp/_github_workflow/event.json' is denied.

Emm, it seems there are something wrong on self host runner setup

PR #451, opened 2025-03-03
modify the use_kl_loss parameter

qyccc:
From the definetion of apply_kl_penalty, it should be if use_kl_loss==True, then apply_kl_penalty.


qyccc:
ok, I figured out. Thank you

Issue #446, opened 2025-03-03
How to set customized reward function?


Jerry-hyl:
How to set customized reward function in RL algorithms like ppo, grpo etc.? I found that the `compute_score` parameter in the file `main_ppo.py` is always set to None and cannot be revised other than editing the source code.

eric-haibin-lin:
https://github.com/volcengine/verl/pull/452 would be helpful, before that, i recommend try to make a copy of main_ppo.py and rely on verl pip install

Kirrito-k423:
中文：

当前 `main` 已经支持自定义 reward function，不需要再直接修改 `main_ppo.py`。推荐使用当前配置路径：

```bash
python3 -m verl.trainer.main_ppo   ...   reward.custom_reward_function.path=/abs/path/to/my_reward.py   reward.custom_reward_function.name=my_reward_function
```

`my_reward.py` 中的函数可按文档使用类似签名：

```python
def my_reward_function(data_source, solution_str, ground_truth, extra_info=None):
    return 0.1
```

如果函数名就是 `compute_score`，`reward.custom_reward_function.name` 可以保持默认。当前代码会从 `config.reward.custom_reward_function.path/name` 动态加载函数，并优先把它作为 reward manager 的 `compute_score`；未配置时才回退到内置 reward functions。旧的顶层 `custom_reward_function` 配置也有迁移逻辑，但新配置建议使用 `reward.custom_reward_function.*`。

English:

Current `main` already supports custom reward functions, so editing `main_ppo.py` directly is no longer required. The recommended override is:

```bash
python3 -m verl.trainer.main_ppo   ...   reward.custom_reward_function.path=/abs/path/to/my_reward.py   reward.custom_reward_function.name=my_reward_function
```

The custom function can use a signature like:

```python
def my_reward_function(data_source, solution_str, ground_truth, extra_info=None):
    return 0.1
```

If the function is named `compute_score`, `reward.custom_reward_function.name` can stay at its default. The current code dynamically loads `config.reward.custom_reward_function.path/name` and uses it as the reward manager's `compute_score`; without that config, it falls back to the built-in reward functions. The legacy top-level `custom_reward_function` config is also migrated, but new configs should prefer `reward.custom_reward_function.*`.

This response was generated by a scheduled Codex GPT5.5 issue processing task.

PR #452, opened 2025-03-03
feat: support loading reward function from an external file

dirtyDan0:
This PR is inspired by https://github.com/volcengine/verl/issues/446 and https://github.com/volcengine/verl/pull/162.

Though https://github.com/volcengine/verl/pull/162 tried to enable users to directly pass in a customized reward function, https://github.com/volcengine/verl/issues/446 thought *the `compute_score` parameter in the file `main_ppo.py` is always set to None and cannot be revised other than editing the source code*. 

https://github.com/volcengine/verl/blob/65cceb3c9dbe6e18230bc3dc045af6e3280c9752/verl/trainer/main_ppo.py#L23-L39

I found that in `main` we cannot specify our customized reward function, so the parameter `compute_score` in both `run_ppo` and `main_task` cannot take effect because `main` doesn't pass `compute_score` to `run_ppo`.

In order to enable users to try their customized reward functions without editing the source code inside `verl`, I propose a method that allow users to dynamically load customized reward fuction by specifying a file containing the reward function through parameters `custom_reward_function.path` and `custom_reward_function.name`(optional, by default compute_score)

Here's an example.
```bash
FILE="$(pwd)/my_reward_function.py"
FUNCTION_NAME="my_reward_function"

python3 -m verl.trainer.main_ppo --config-path=./config --config-name='ppo_trainer'\
    ...
    custom_reward_function.path=$FILE\
    custom_reward_function.name=$FUNCTION_NAME\
    ...
```

In `my_reward.py` there should be a function `my_reward_function` with parameters `data_source`, `solution_str`, `ground_truth` and `extra_info` ( the same as `_default_compute_score`) https://github.com/volcengine/verl/blob/65cceb3c9dbe6e18230bc3dc045af6e3280c9752/verl/utils/reward_score/__init__.py#L17

In this way, if we want to test multiple customized reward functions, we can implement them in one or mutiple files and change  `custom_reward_function.path` and `custom_reward_function.name` for different trials.

If you don't specify `custom_reward_function.path`, there would be no change to the use of the builtin datasets and their corresponding reward functions defined in [verl/utils/reward_score/__init__.py.](https://github.com/volcengine/verl/blob/65cceb3c9dbe6e18230bc3dc045af6e3280c9752/verl/utils/reward_score/__init__.py).

### DIFF
--- a/.github/workflows/e2e_gsm8k.yml
+++ b/.github/workflows/e2e_gsm8k.yml
@@ -84,3 +84,7 @@ jobs:
         run: |
           ray stop --force
           bash tests/e2e/run_qwen_gsm8k_model_rm_liger_kernel.sh
+      - name: Running gsm8k e2e training tests on 8 L20 GPUs with rmpad using customized reward function
+        run: |
+          ray stop --force
+          bash tests/e2e/run_qwen_gsm8k_custom_function_rm.sh
--- a/docs/examples/config.rst
+++ b/docs/examples/config.rst
@@ -309,6 +309,18 @@ Reward Model
   if ``naive``. If all verification functions are multiprocessing-safe, the reward
   manager can be set to ``prime`` for parallel verification.
 
+Customized Reward Function
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. code:: yaml
+  
+   custom_reward_function:
+     path: null
+     name: compute_score
+
+- ``custom_reward_function.path``: The path to the file containing your customized reward function. If not specified, pre-implemented reward functions will be used.
+- ``custom_reward_function.name`` (Optional) : The name of the reward function within the specified file. Default is 'compute_score'.
+
 Algorithm
 ~~~~~~~~~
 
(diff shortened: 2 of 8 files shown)


vermouth1992:
Could you add a test case so that this function won't be broken in the future.

eric-haibin-lin:
This looks awesome, thanks for the contribution. Maybe we should add a section in https://verl.readthedocs.io/en/latest/preparation/reward_function.html (https://github.com/volcengine/verl/blob/main/docs/preparation/reward_function.rst) as well to document how to train with a custom reward function

dirtyDan0:
> Could you add a test case so that this function won't be broken in the future.

 @vermouth1992 
A test case is added. See .github/workflows/e2e_gsm8k.yml, tests/e2e/run_qwen_gsm8k_custom_function_rm.sh and tests/e2e/check_custom_rwd_fn.py

> This looks awesome, thanks for the contribution. Maybe we should add a section in 
https://verl.readthedocs.io/en/latest/preparation/reward_function.html (https://github.com/volcengine/verl/blob/main/docs/preparation/reward_function.rst) as well to document how to train with a custom reward function

@eric-haibin-lin 
Docs updated.  See docs/examples/config.rst and docs/preparation/reward_function.rst

dirtyDan0:
@vermouth1992 Could you help me approve the workflows? Thank U :)

vermouth1992:
Please test the cases locally and fix it. Thanks!

dirtyDan0:
> It seems that the gsm8k tests failed. Please test the cases locally and fix it. Thanks!

I changed the paths to absolute paths and the local test passed. 😀

vermouth1992 (approved):

PR #459, opened 2025-03-04
[fix] separate prompt and response in reward manager

hiyouga:
## What does this PR do?

1. Separate the prompt part and the response part in reward manager to avoid the reward leakage of format reward.
2. Update the reward score function for Geometry3k dataset.
3. Update the content in the readme file.

## Who can review?

@vermouth1992 @PeterSH6

### DIFF
--- a/verl/workers/reward_manager/naive.py
+++ b/verl/workers/reward_manager/naive.py
@@ -52,8 +52,8 @@ class NaiveRewardManager:
             valid_response_ids = response_ids[:valid_response_length]
 
             # decode
-            sequences = torch.cat((valid_prompt_ids, valid_response_ids))
-            sequences_str = self.tokenizer.decode(sequences)
+            prompt_str = self.tokenizer.decode(valid_prompt_ids)
+            response_str = self.tokenizer.decode(valid_response_ids)
 
             ground_truth = data_item.non_tensor_batch['reward_model']['ground_truth']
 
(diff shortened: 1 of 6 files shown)


verl/workers/reward_manager/naive.py
```diff
@@ -63,7 +63,8 @@ def __call__(self, data: DataProto):
 
             score = self.compute_score(
                 data_source=data_source,
-                solution_str=sequences_str,
+                prompt_str=prompt_str,
+                response_str=response_str,
```
PeterSH6: Shall we only pass the response_str? I guess this will break the PRIME reward manager

verl/utils/reward_score/__init__.py
```diff
@@ -14,25 +14,25 @@
 # from . import gsm8k, math, prime_math, prime_code
 
 
-def _default_compute_score(data_source, solution_str, ground_truth, extra_info=None):
+def _default_compute_score(data_source, prompt_str, response_str, ground_truth, extra_info=None):
```
PeterSH6: Same here. Shall we only pass the response_str? I guess this will break the PRIME reward manager.

verl/workers/reward_manager/naive.py
hiyouga: Sure, I have updated accordingly

PeterSH6 (approved):

PR #463, opened 2025-03-04
support speed up downloading model from modelscope

mi804:
Add support for downloading models from modelscope by setting `VERL_USE_MODELSCOPE=True`


requirements.txt
```diff
@@ -18,3 +18,4 @@ torchdata
 transformers
 vllm<=0.6.3
 wandb
+modelscope
```
hiyouga: It would be better to make the modelscope dependency optional

mi804: have move dependency check before import modelscope

hiyouga (commented):
LGTM

PR #470, opened 2025-03-04
ascend: update e2e_ascend.yml

celestialli:


vermouth1992:
Do we want to merge this?

willem-bd:
@celestialli Just a quick question why should we list all these devices ?

celestialli:
> @celestialli Just a quick question why should we list all these devices ?

Sorry, this PR is to trigger ascend CI for test, and is no longer needed. I'll close this.

PR #480, opened 2025-03-05
[feat] add val_generations_to_log_to_swanlab

Zeyi-Lin:
In this PR, a `val_generations_to_log_to_swanlab` parameter has been added. When this parameter is set to 1, it supports logging the generated text from eval in SwanLab.

@hiyouga 

---

This pull request introduces logging of validation generations to Swanlab in addition to Wandb. The changes include updates to several configuration files and the addition of a new logging method in the `ray_trainer.py` file.

Key changes include:

### Configuration Updates:
* Added `val_generations_to_log_to_swanlab` parameter to the `trainer` section in the following configuration files:
  * `examples/split_placement/config/ppo_trainer_split.yaml`
  * `verl/trainer/config/ppo_megatron_trainer.yaml`
  * `verl/trainer/config/ppo_trainer.yaml`

### Code Updates:
* Added a new method `_maybe_log_val_generations_to_swanlab` to log validation samples to Swanlab in `verl/trainer/ppo/ray_trainer.py`
* Updated the `_validate` method to call the new Swanlab logging method in `verl/trainer/ppo/ray_trainer.py`

---

### DIFF
--- a/verl/trainer/config/ppo_megatron_trainer.yaml
+++ b/verl/trainer/config/ppo_megatron_trainer.yaml
@@ -152,6 +152,7 @@ trainer:
   experiment_name: gsm8k
   logger: ['console', 'wandb']
   val_generations_to_log_to_wandb: 0
+  val_generations_to_log_to_swanlab: 0
   nnodes: 1
   n_gpus_per_node: 8
   save_freq: -1
--- a/verl/trainer/config/ppo_megatron_trainer.yaml
+++ b/verl/trainer/config/ppo_megatron_trainer.yaml
@@ -152,7 +152,6 @@ trainer:
   experiment_name: gsm8k
   logger: ['console', 'wandb']
   val_generations_to_log_to_wandb: 0
-  val_generations_to_log_to_swanlab: 0
   nnodes: 1
   n_gpus_per_node: 8
   save_freq: -1
--- a/verl/trainer/ppo/ray_trainer.py
+++ b/verl/trainer/ppo/ray_trainer.py
@@ -605,6 +605,52 @@ class RayPPOTrainer(object):
         # Update reference and log
         wandb.log({"val/generations": new_table}, step=self.global_steps)
         self.validation_table = new_table
+ 
+    def _maybe_log_val_generations_to_swanlab(self, inputs, outputs, scores):
+        """Log a table of validation samples to swanlab"""
+
+        generations_to_log = self.config.trainer.val_generations_to_log_to_swanlab
+
+        if generations_to_log == 0:
+            return
+
+        if generations_to_log > 0 and 'swanlab' not in self.config.trainer.logger:
+            print(
+                'WARNING: `val_generations_to_log_to_swanlab` is set to a positive value, but no swanlab logger is found. ')
+            return
+
+        import swanlab
+        import numpy as np
+
+        # Create tuples of (input, output, score) and sort by input text
+        samples = list(zip(inputs, outputs, scores))
+        samples.sort(key=lambda x: x[0])  # Sort by input text
+
+        # Use fixed random seed for deterministic shuffling
+        rng = np.random.RandomState(42)
+        rng.shuffle(samples)
+
+        # Take first N samples after shuffling
+        samples = samples[:generations_to_log]
+
+        # Add new row with all data
+        swanlab_text_list = []
+        for i, sample in enumerate(samples):
+            row_text = f"""
+            input: {sample[0]}
+            
+            ---
+            
+            output: {sample[1]}
+            
+            ---
+            
+            score: {sample[2]}
+            """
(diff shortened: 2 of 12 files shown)


hiyouga (approved):

verl/trainer/config/ppo_megatron_trainer.yaml
```diff
@@ -152,6 +152,7 @@ trainer:
   experiment_name: gsm8k
   logger: ['console', 'wandb']
   val_generations_to_log_to_wandb: 0
+  val_generations_to_log_to_swanlab: 0
```
vermouth1992: We should not create a separate config for this. Instead, it should be combined with val_generations_to_log_to_wandb and the backend is chosen according to trainer.logger

hiyouga:
Please run `bash scripts/format.sh`
Also see the above comments

verl/trainer/config/ppo_megatron_trainer.yaml
Zeyi-Lin: Great suggestion, we have already completed the automatic selection of the logger.

verl/trainer/ppo/ray_trainer.py
```diff
@@ -606,6 +606,53 @@ def _maybe_log_val_generations_to_wandb(self, inputs, outputs, scores):
         wandb.log({"val/generations": new_table}, step=self.global_steps)
         self.validation_table = new_table
 
+    def _maybe_log_val_generations_to_swanlab(self, inputs, outputs, scores):
+        """Log a table of validation samples to swanlab"""
```
PeterSH6: Same here, this func is largely similar to wandb.
Please merge them into one function and avoid making the `ray_trainer.py` too tedious

PeterSH6 (changes requested):

verl/trainer/ppo/ray_trainer.py
Zeyi-Lin: I have split the functions with the consideration of supporting more loggers more flexibly in the future, such as mlflow and tensorboard. 
Please take a look and see if this is feasible.

hiyouga: How about move these functions `_log_generations_to_wandb` and `_log_generations_to_swanlab` to https://github.com/volcengine/verl/blob/main/verl/utils/tracking.py

hiyouga: Actually they are static method because `self` is not used

Zeyi-Lin: Maybe it's a good structure. Let me try it

Zeyi-Lin: Actually `_log_generations_to_wandb` uses `self.validation_table`.

hiyouga: Yeah, we might need a new class to maintain this attribute

PeterSH6: Yes, I agree with this design: move all the log_generation_to_xx into a class that manage these operations outside the `ray_trainer.py`

Zeyi-Lin: OK, I have made a move in the latest commit.

Zeyi-Lin: Additionally, should we consider renaming the parameter `val_generations_to_log_to_wandb` to match more loggers?
such as, `val_generations_to_log`?

verl/utils/tracking.py
```diff
@@ -162,3 +162,61 @@ def _flatten_dict(raw: Dict[str, Any], *, sep: str) -> Dict[str, Any]:
     ans = pd.json_normalize(raw, sep=sep).to_dict(orient='records')[0]
     assert isinstance(ans, dict)
     return ans
+
+
+class ValidationGenerationsLogger:
```
hiyouga, line 168: Should we use `dataclass` here?

hiyouga (commented):
The new design looks extremely good to me, nice work @Zeyi-Lin

verl/utils/tracking.py
Zeyi-Lin: yeah, I added it

vermouth1992 (dismissed):

Zeyi-Lin:
It seems that the CI errors are caused by network factors:
```
RuntimeError: An error occurred while downloading using `hf_transfer`. Consider disabling HF_HUB_ENABLE_HF_TRANSFER for better error handling.
```

vermouth1992:
Please fix formatting

hiyouga (approved):

vermouth1992 (approved):

vermouth1992:
@PeterSH6 Please see whether the changes make sense. I guess you need to resolve the requested change before this PR can be merged.

PeterSH6 (approved):
LGTM

PR #409, opened 2025-02-27
fix: (1) skipped last step (2) redundant validation and logging

dirtyDan0:
This PR solves these 2 following problems.

1. Last step skipped

   `self.global_steps += 1` before if `self.global_steps >= self.total_training_steps` makes the last step skipped.

   We start from step 1, and we expect `self.total_training_steps` in total. 

   https://github.com/volcengine/verl/blob/82b38e25c72e1b6de7d7d2092af6e1ed5dd2a400/verl/trainer/ppo/ray_trainer.py#L999-L1001 

   When `self.global_steps == self.total_training_steps-1`:

   * we have only executed `self.total_training_steps-1` steps

   * `self.global_steps` is updated to `self.total_training_steps`
   * `self.global_steps >= self.total_training_steps` is satisfied, and the training ends.

   Therefore, we should put `self.global_steps += 1` at last

2. redundant validation and logging

   If  `self.total_training_steps % self.config.trainer.test_freq == 0` :

   * `self._validate()` will be executed twice 

     1. https://github.com/volcengine/verl/blob/82b38e25c72e1b6de7d7d2092af6e1ed5dd2a400/verl/trainer/ppo/ray_trainer.py#L984

     2. https://github.com/volcengine/verl/blob/82b38e25c72e1b6de7d7d2092af6e1ed5dd2a400/verl/trainer/ppo/ray_trainer.py#L1005

   * logging will also be executed twice

     1. https://github.com/volcengine/verl/blob/82b38e25c72e1b6de7d7d2092af6e1ed5dd2a400/verl/trainer/ppo/ray_trainer.py#L985 and https://github.com/volcengine/verl/blob/82b38e25c72e1b6de7d7d2092af6e1ed5dd2a400/verl/trainer/ppo/ray_trainer.py#L997
     2. https://github.com/volcengine/verl/blob/82b38e25c72e1b6de7d7d2092af6e1ed5dd2a400/verl/trainer/ppo/ray_trainer.py#L1007

PR #483, opened 2025-03-05
[fix] ppo trainer exit too early

### LINKED ISSUE #481: Training mIss the last batch of last epoch.
Current code add the global_steps before checking wether the it finished all the training steps. This will result missing training the last batch of last epoch. For example, if there are in total 2 steps, the code starts from step 1, and here it will adds to step 2 and therefore directly stop training.

```
                self.global_steps += 1

                if self.global_steps >= self.total_training_steps:

                    # perform validation after training
                    if self.val_reward_fn is not None:
                        val_metrics = self._validate()
                        pprint(f'Final validation metrics: {val_metrics}')
                        logger.log(data=val_metrics, step=self.global_steps)
                    if self.config.trainer.save_freq > 0 and \
                            (self.global_steps - 1) % self.config.trainer.save_freq != 0:
                        with _timer('save_checkpoint', timing_raw):
                            self._save_checkpoint()
                    return
```

hiyouga:
The global step should be `1` at the end of first step, not `2`.

Fixes #481

### DIFF
--- a/verl/trainer/ppo/ray_trainer.py
+++ b/verl/trainer/ppo/ray_trainer.py
@@ -887,11 +887,9 @@ class RayPPOTrainer(object):
             if self.config.trainer.get('val_only', False):
                 return
 
-        # we start from step 1
-        self.global_steps += 1
-
         for epoch in range(self.config.trainer.total_epochs):
             for batch_dict in self.train_dataloader:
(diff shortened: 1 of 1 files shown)


hiyouga:
Closed in favor of https://github.com/volcengine/verl/pull/409

PR #499, opened 2025-03-06
[CI] feat: auto cancel previous CI in the same PR

tongyx361:
- [x] Add concurrency to workflows to cancel previous workflows when new commit is pushed to the same branch.
- [ ] Cancel all workflows/jobs from the same commit if any fails? (Not sure whether we really need it)

Note: we leave out `secrets_scan.yml` and `scorecard.yml` to avoid any possible leakage or security risk, which also cost little.

### DIFF
--- a/.github/workflows/dataset.yml
+++ b/.github/workflows/dataset.yml
@@ -18,6 +18,11 @@ on:
       - "**/*.py"
       - .github/workflows/dataset.yml
 
+# Cancel jobs on the same ref if a new one is triggered
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}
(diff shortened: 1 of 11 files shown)


willem-bd:
> * [x]  Add concurrency to workflows to cancel previous workflows when new commit is pushed to the same branch.
> * [ ]  Cancel all workflows/jobs from the same commit if any fails? (Not sure whether we really need it)
> 
> Note: we leave out `secrets_scan.yml` and `scorecard.yml` to avoid any possible leakage or security risk, which also cost little.

I don't think we need to Cancle all workflows/jobs if any fails, as we need to check all the error before fixing them.

tongyx361:
According to the records above, the old workflows are successfully canceled when a new commit is pushed.

PeterSH6 (approved):

PR #500, opened 2025-03-06
[ckpt] sort pgs by node ip to make RANK consistent across nodes

wuxibin89:


vermouth1992 (approved):

eric-haibin-lin:
What happens without this fix?

PeterSH6:
@eric-haibin-lin Those who don't have shared cloud storage within the cluster may fail to resume training, although they save the sharded checkpoints due to the random assignment from Ray.
For example, in the first run, node 0 is assigned with rank[0 - 7]. Then, in the second run, node 0 will be assigned with rank [15 - 23]. 
I think sorting with IP will make our Ray WorkerGroup more similar to torchrun's behavior

PeterSH6 (approved):

vermouth1992:
> @eric-haibin-lin Those who don't have shared cloud storage within the cluster may fail to resume training, although they save the sharded checkpoints due to the random assignment from Ray. For example, in the first run, node 0 is assigned with rank[0 - 7]. Then, in the second run, node 0 will be assigned with rank [15 - 23]. I think sorting with IP will make our Ray WorkerGroup more similar to torchrun's behavior

I don't quite get why this matters as the model checkpoint is loaded according to the rank in Pytorch

PR #505, opened 2025-03-07
misc: precheck resource pool available to prevent pg hang

### LINKED ISSUE #503: Add available resource checking in ray cluster before launching jobs
If the available resources in ray cluster is not enough to schedule the job, the program will hang indefinitely. We should add a checker and only launch training when there are enough resources.

wuxibin89:
close #503

### DIFF
--- a/verl/trainer/ppo/ray_trainer.py
+++ b/verl/trainer/ppo/ray_trainer.py
@@ -99,7 +99,7 @@ class ResourcePoolManager:
     def _check_resource_available(self):
         """Check if the resource pool can be satisfied in this ray cluster."""
         node_available_resources = ray.state.available_resources_per_node()
-        node_available_gpus = {node: node_info.get(['GPU'], 0) for node, node_info in node_available_resources.items()}
+        node_available_gpus = {node: node_info.get('GPU', 0) for node, node_info in node_available_resources.items()}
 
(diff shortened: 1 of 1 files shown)


verl/trainer/ppo/ray_trainer.py
```diff
@@ -89,10 +90,39 @@ def create_resource_pool(self):
...
+        for resource_pool_name, process_on_nodes in self.resource_pool_spec.items():
+            num_gpus, num_nodes = process_on_nodes[0], len(process_on_nodes)
+            for node, available_gpus in node_available_gpus.items():
+                if available_gpus >= num_gpus:
+                    node_available_gpus[node] -= num_gpus
+                    num_nodes -= 1
+                    if num_nodes == 0:
+                        break
+            if num_nodes > 0:
+                raise ValueError(
```
vermouth1992, line 122: Shall we add a while loop here to wait for some time instead of directly raise?

wuxibin89: For most common case, we assume user have already setup multi-node ray cluster, and then `ray job submit ...`, so the wait loop may have no meaning. I prefer to raise exception immediately and let user retry.

vermouth1992 (approved):

PR #527, opened 2025-03-10
[bugfix] Fix position embedding processing for Qwen2.5-VL

Yifan-Song793:
[bugfix] Fix position embedding processing for Qwen2.5-VL

In the `RLHFDataset.__getitem__` method, a bug was identified in how multimodal position IDs (3D in Qwen2.5-VL) are determined. Previously, the code checked for `self.image_key in row_dict` to decide whether to use multimodal position IDs. However, since `self.image_key` is popped from `row_dict` during image token expansion, this check incorrectly fails for subsequent operations.

This causes the VL model to use incorrect position IDs, resulting in significant performance degradation. 

image

The fix introduces an explicit `is_multi_modal` flag to properly track multimodal content throughout the processing pipeline.


hiyouga (approved):
LGTM! thanks for fixing

PR #542, opened 2025-03-11
[bugfix] fix: generation script

nwiad:
# Description
- Corrected dummy size to avoid faulty communication.
- Fixed batch number calculation.
- Adjusted worker group role to alleviate memory overhead.
- Add ray.init() to prevent failing to register worker.

### DIFF
--- a/verl/trainer/main_generation.py
+++ b/verl/trainer/main_generation.py
@@ -40,6 +40,7 @@ from verl.single_controller.ray import RayClassWithInitArgs, RayResourcePool, Ra
 def main(config):
     run_generation(config)
 
+
 def run_generation(config) -> None:
 
     if not ray.is_initialized():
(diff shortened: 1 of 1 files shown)


verl/trainer/main_generation.py
```diff
@@ -38,6 +38,25 @@
...
 @hydra.main(config_path='config', config_name='generation', version_base=None)
 def main(config):
+    run_generation(config)
+
+def run_generation(config) -> None:
+    runtime_env = {
+        'env_vars': {
+            'TOKENIZERS_PARALLELISM': 'true',
+            'NCCL_DEBUG': 'WARN',
+            'CUDA_VISIBLE_DEVICES': '0,1,2,3,4,5,6,7'
```
vermouth1992: Why do we need CUDA_VISIBLE_DEVICES?

nwiad: Yes I believe this part is over-elaborate. I'll just keep it simple.

vermouth1992 (approved):

vermouth1992 (approved):

PR #543, opened 2025-03-11
refactor: remove custom vllm weight loader and use model.load_weights directly

wuxibin89:
As we're moving to vllm>=0.7.3, we should remove `verl/third_party` complelely in the future.


PeterSH6:
vllm_spmd is for vllm > 0.7.3. We should keep it rather than delete it.

After the vllm > 0.7.3 becomes robust (both in system and algorithm side), we can delete the previous version. Otherwise, we should keep it as a robust version.

vermouth1992:
Could you rebase main and trigger CI again? There is a new test using a single GPU

vermouth1992:
We can delete vllm_spmd file and see whether the CI will break

vermouth1992:
There seems to be an import problem.

PeterSH6:
It seems that the `hf_weight_loader` can be deleted too.

wuxibin89:
@PeterSH6 Can we cleanup code related to `actor_rollout_ref.rollout.load_format=hf`? It seems there's no need anymore.

PeterSH6:
The hf_weight_loader is mainly for `world_size == 1`. Can we use dtensor when `world_size == 1`?

wuxibin89:
> The hf_weight_loader is mainly for `world_size == 1`. Can we use dtensor when `world_size == 1`?

No, if world_size == 1, model.state_dict() just contain origin torch.Tensor.

PeterSH6 (approved):

PR #547, opened 2025-03-11
Regularise Prime RM

cemiu:
1. Fixes prime reward manager passing different solution_strs compared to naive reward manager
2. adds slicing support to DataProto, rather than gracefully misbehaving (and returning DataProtoItem containing slices)

### DIFF
--- a/verl/protocol.py
+++ b/verl/protocol.py
@@ -195,6 +195,15 @@ class DataProto:
             return 0
 
     def __getitem__(self, item):
+        if isinstance(item, slice):
+            if self.batch is not None:
+                tensor_data = self.batch[item]
+            else:
+                tensor_data = None
+
+            non_tensor_data = {key: val[item] for key, val in self.non_tensor_batch.items()}
+            return DataProto(batch=tensor_data, non_tensor_batch=non_tensor_data, meta_info=self.meta_info)
+
         tensor_data = self.batch[item]
         non_tensor_data = {key: val[item] for key, val in self.non_tensor_batch.items()}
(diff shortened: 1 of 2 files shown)


verl/protocol.py
```diff
@@ -195,6 +195,15 @@ def __len__(self):
...
+                tensor_data = self.batch[item]
+            else:
+                tensor_data = None
+
+            non_tensor_data = {key: val[item] for key, val in self.non_tensor_batch.items()}
+            return DataProto(batch=tensor_data, non_tensor_batch=non_tensor_data, meta_info=self.meta_info)
+
         tensor_data = self.batch[item]
         non_tensor_data = {key: val[item] for key, val in self.non_tensor_batch.items()}
         return DataProtoItem(batch=tensor_data, non_tensor_batch=non_tensor_data, meta_info=self.meta_info)
```
vermouth1992, line 209: I also think DataProtoItem is not a good design and should be removed. However, when item is a single index, the batch dim vanishes. Is there any way to handle both cases?

cemiu: We could always maintain the dimensionality for both batch and non_tensor_batch, and then add a squeeze function to DataProto. However, int indices removing a dimension while slices maintain it is fairly common in Python (lists, numpy, torch, etc.) so it's not necessarily a bad thing.

I am far too new to the code base to know what the appropriate call is though

cemiu: @vermouth1992 I have another RP outstanding that requires support for slicing behaviour. Will it be better to undo the commit from here and submit them together?

PR #553, opened 2025-03-12
[rollout] feat: support sampling in validation stage

PeterSH6:
Currently, eager mode is applied in the validation stage. However, in some reasoning tasks, we may need to generate n times and average the scores.

In this PR, we support using non-eager sampling parameters during validation by specifying the `val_kwargs` in `actor_rollout_ref.rollout` config field.

**Future work**
- [ ] Merge `vllm_rollout_spmd.py` and `vllm_rollout.py` into one file.

### DIFF
--- a/verl/trainer/config/ppo_megatron_trainer.yaml
+++ b/verl/trainer/config/ppo_megatron_trainer.yaml
@@ -84,6 +84,13 @@ actor_rollout_ref:
       gate_proj_layer_name: gate_up
     # number of responses (i.e. num sample times)
     n: 1
+    val_kwargs:
+      # sampling parameters for validation
+      top_k: -1 # 0 for hf rollout, -1 for vllm rollout
+      top_p: 1.0
+      temperature: 0
+      n: 1
+      do_sample: False # default eager for validation
 
 critic:
   strategy: megatron
--- a/verl/workers/rollout/vllm_rollout/vllm_rollout.py
+++ b/verl/workers/rollout/vllm_rollout/vllm_rollout.py
@@ -193,6 +193,7 @@ class vLLMRollout(BaseRollout):
                 'temperature': self.config.val_kwargs.temperature,
                 'n': self.config.val_kwargs.n,
             }
+
         # users can customize different sampling_params at different run
         with self.update_sampling_params(**kwargs):
             output = self.inference_engine.generate(
--- a/verl/workers/rollout/vllm_rollout/vllm_rollout.py
+++ b/verl/workers/rollout/vllm_rollout/vllm_rollout.py
@@ -185,7 +185,7 @@ class vLLMRollout(BaseRollout):
                 'temperature': 0,
                 'n': 1  # if greedy, only 1 response
             }
-        elif is_validate and do_sample:
+        elif is_validate:
             # TODO: try **
             kwargs = {
                 'top_k': self.config.val_kwargs.top_k,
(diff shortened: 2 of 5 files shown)


vermouth1992:
When eval using n samples, shall we also record pass@n metrics?

PeterSH6:
@vermouth1992 Yes, we have an implementation of pass@n logging. Will push them in the upcoming PRs

vermouth1992 (dismissed):

verl/trainer/config/ppo_megatron_trainer.yaml
```diff
@@ -84,6 +84,13 @@ actor_rollout_ref:
       gate_proj_layer_name: gate_up
     # number of responses (i.e. num sample times)
     n: 1
+    val_kwargs:
```
tongyx361, line 87: yaml items are naturally kwargs. Is `val` a better name?

```suggestion
    val:
```

verl/trainer/config/ppo_megatron_trainer.yaml
```diff
@@ -84,6 +84,13 @@ actor_rollout_ref:
...
+      # sampling parameters for validation
+      top_k: -1 # 0 for hf rollout, -1 for vllm rollout
+      top_p: 1.0
+      temperature: 0
+      n: 1
```
tongyx361, line 92: It might be helpful to support different `n` per validation dataset.

This might be simply implemented by duplicating the data and setting `n` in `SamplingParams` as `1`.

verl/workers/rollout/vllm_rollout/vllm_rollout.py
```diff
@@ -184,6 +185,14 @@ def generate_sequences(self, prompts: DataProto, **kwargs) -> DataProto:
                 'temperature': 0,
                 'n': 1  # if greedy, only 1 response
             }
+        elif is_validate and do_sample:
```
tongyx361: `and do_sample` is redundant. But leaving it as comment might making this branch more readable.


```suggestion
        elif is_validate: # and do_sample
```

tongyx361 (approved):
This PR looks correct to me.
But some details can be improved. Could you / Should I make a new commit or PR?

verl/workers/rollout/vllm_rollout/vllm_rollout.py
PeterSH6: This `do_sample` is necessary. If the user accidentally set do_sample=False while setting `temperature>0`. Then, sampling parameters will be override

verl/trainer/config/ppo_megatron_trainer.yaml
PeterSH6: val_kwargs is used to specify that this field relates to the kwargs variable in vllm_rollout.py

verl/trainer/config/ppo_megatron_trainer.yaml
PeterSH6: That's a good suggestion for the next PR.

verl/workers/rollout/vllm_rollout/vllm_rollout.py
tongyx361: I mean there is a `if not do_sample` above, so here `elif` implicitly implies `and do_sample`.

PeterSH6: Yes, that's what I refer to. When do_sample=False and validate=True. Without a do_sample check, the sampling parameters would be wrong if someone accidentally set temperature>0

PeterSH6: I see, we should use if not elif...

vermouth1992 (approved):

PR #557, opened 2025-03-12
[feat] add auto padding for DataProto

zw0610:
Instead of requiring user to manually repeat `DataProto` to meat the dp size of a `WorkerGroup`, auto_padding options pad the DataProto when chunk is called.
This feat is turned off by default and can be activate globally as well as dataproto-specifically

### DIFF
--- a/verl/protocol.py
+++ b/verl/protocol.py
@@ -43,17 +43,17 @@ except:
 class _DataProtoConfigMeta(type):
     _config = {}
 
-    _verl_auto_padding = "_verl_auto_padding"
+    auto_padding_key = "_verl_auto_padding"
 
     @property
     def auto_padding(cls):
         enabled_by_env = os.getenv("VERL_AUTO_PADDING", "FALSE").upper() in ["TRUE", "1"]
-        return enabled_by_env or cls._config.get(cls._verl_auto_padding, False)
+        return enabled_by_env or cls._config.get(cls.auto_padding_key, False)
 
     @auto_padding.setter
     def auto_padding(cls, enabled: bool):
         assert isinstance(enabled, bool), f"enabled must be a boolean, got {enabled} as {type(enabled)}"
-        cls._config[cls._verl_auto_padding] = enabled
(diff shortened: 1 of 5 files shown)


verl/single_controller/base/decorator.py
```diff
@@ -42,16 +44,41 @@ class Execute(Enum):
     RANK_ZERO = 1
 
 
-def _split_args_kwargs_data_proto(chunks, *args, **kwargs):
+def _split_args_kwargs_data_proto(chunks, *args, __padding_127xcd7=False, **kwargs):
```
hongpeng-guo, line 47: Maybe we can use a more meaningful name for this kwarg, i.e., `enable_auto_padding=False`. I understand  this "wired" name is for backward compatibility, i.e., avoid conflicting with potential user defined args. But it still makes sense to choose a good arg name and make it readable. 
 
We are planing to add API stability to the key APIs. Let's choose a better name and then mark it as stable. if we have to make some changes with potential break, it's better to break now than later.

hongpeng-guo (commented):
@eric-haibin-lin Could you please take a look at this one?

verl/protocol.py
```diff
@@ -301,7 +328,7 @@ def check_consistency(self):
                     0] == batch_size, f'key {key} length {len(val)} is not equal to batch size {batch_size}'
 
     @classmethod
-    def from_single_dict(cls, data: Dict[str, Union[torch.Tensor, np.ndarray]], meta_info=None):
+    def from_single_dict(cls, data: Dict[str, Union[torch.Tensor, np.ndarray]], meta_info=None, auto_padding=False):
```
eric-haibin-lin, line 331: please add docs for this API

verl/single_controller/base/decorator.py
zw0610: @vermouth1992 What do you think, since the *weirdness* of the name came from your idea?

hongpeng-guo:
close this PR and pivot for #1356

Issue #501, opened 2025-03-06
[RFC] ActorEnvironment implementation to support tool calling


casper-hansen:
This is an issue to request for comments (RFC) from veRL maintainers @eric-haibin-lin @PeterSH6 @vermouth1992 on an implementation that would fundamentally support tool calling in the roadmap #354 and close #344. veRL has seen great adoption, but a lot of interesting work is present in forks like [RAGEN](https://github.com/ZihanWang314/ragen) or [Search-R1](https://github.com/PeterGriffinJin/Search-R1). This would help bring more use-cases directly into veRL instead of un-maintained modified implementations elsewhere that cannot benefit from the new features that veRL keeps adding.

Specifically, I am undecided on which implementation is the best when it comes to the actual use-case: reusing RAGEN/Search-R1 code or redesigning the vLLM SPMD for async streamed outputs.

## Background

The actor in GRPO generates many completions. During the process of generating completions, the model can learn interleaved tool calling. However, this requires stopping on certain tokens like `</tool_call>` to execute and return results back into the model.

## Which use-cases could interleaved tool-calling enable?

1. A simple and common use-case is search/retrieval. Given a search tool, the model can learn to query and retrieve additional information into the context.
2. Code execution and feedback. You can implement an environment with execution and feedback. This extends to the possibility of generating multiple parallel solutions for an advanced use-case.
3. Evaluate mathematical expressions and return feedback. Advanced use-cases could be calling Wolframalpha API or using Lean 4 for theorem proving.

## Rough idea of implementation

The main idea is to create a class `ActorEnvironment` that will be the default in `RayPPOTrainer`. In the initial implementation, `ActorEnvironment` will only include refactor code from the `fit` method, but allow for extensibility through defining a custom `ActorEnvironment` as an input to `RayPPOTrainer`.

## Class outline

```python
class ActorEnvironment:
    def __init__(self): pass
    def step(self): pass
    def update(self): pass
```

### step method

https://github.com/volcengine/verl/blob/4a291fa7604afacc81b902234dd58d1fbb1fc936/verl/trainer/ppo/ray_trainer.py#L901-L926

### update method

https://github.com/volcengine/verl/blob/4a291fa7604afacc81b902234dd58d1fbb1fc936/verl/trainer/ppo/ray_trainer.py#L989-L995

## Use-case: DuckDuckGo ActorEnvironment

[...]

PR #560, opened 2025-03-12
[Part 1/2] Refactor actor training logic into ActorEnvironment

casper-hansen:
This is part 1/2 for the issue #501. This PR addresses provides the following features:
- This is a refactor of the actor logic found in `RayPPOTrainer.fit()`, no new functionality added.
- This creates extensibility, allowing the user to define their own ActorEnvironment and pass it through their own `main_ppo.py` script.
- When a user passes a custom environment, it creates the possibility to create a custom handling of rollouts for a use-case, which will enable part 2 of #501.

Part 2 will come in the future after a lot more experimentation on the specific use-case as outlined in #501.

### DIFF
--- a/verl/trainer/ppo/actor_env.py
+++ b/verl/trainer/ppo/actor_env.py
@@ -18,7 +18,6 @@ This is the default ActorEnvironment that performs the rollout and actor update.
 import uuid
 import numpy as np
 from copy import deepcopy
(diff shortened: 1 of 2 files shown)


casper-hansen:
Looks like the commits are with an invalid user which means I cannot sign CLA. I will have to redo the commits.

PR #561, opened 2025-03-12
[Part 1/2] Refactor actor training logic into ActorEnvironment

casper-hansen:
This is part 1/2 for the issue #501. This PR addresses provides the following features:
- This is a refactor of the actor logic found in `RayPPOTrainer.fit()`, no new functionality added.
- This creates extensibility, allowing the user to define their own ActorEnvironment and pass it through their own `main_ppo.py` script.
- When a user passes a custom environment, it creates the possibility to create a custom handling of rollouts for a use-case, which will enable part 2 of #501.

Part 2 will come in the future after a lot more experimentation on the specific use-case as outlined in #501.

### DIFF
--- a/verl/trainer/ppo/actor_env.py
+++ b/verl/trainer/ppo/actor_env.py
@@ -21,10 +21,29 @@ from copy import deepcopy
 
 
 class ActorEnvironment:
-    def __init__(self, config, actor_rollout_wg, reward_fn):
+
+    def __init__(self):
+        pass
+
+    def post_init(self, config, actor_rollout_wg, reward_fn, tokenizer, processor):
+        """A post init function that adds the essentials for the training loop
+        after initializing the ActorEnvironment class.
+
+        Args:
+            config: The veRL config.
+            actor_rollout_wg: The actor rollout worker group.
+            reward_fn: The reward function.
+            tokenizer: The tokenizer.
+            processor: The processor (multimodal).
+            
+        Returns:
+            Batch: Enriched batch containing generated sequences, rewards, and metadata
+        """
         self.config = config
         self.actor_rollout_wg = actor_rollout_wg
         self.reward_fn = reward_fn
+        self.tokenizer = tokenizer
+        self.processor = processor
 
     def step(self, batch, gen_batch, timing_raw):
         """Performs a rollout step to generate sequences and compute rewards.
--- a/verl/trainer/ppo/actor_env.py
+++ b/verl/trainer/ppo/actor_env.py
@@ -1,0 +1,68 @@
+# Copyright 2024 Bytedance Ltd. and/or its affiliates
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""
+This is the default ActorEnvironment that performs the rollout and actor update.
+"""
+
+import uuid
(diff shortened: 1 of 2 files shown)


verl/trainer/ppo/actor_env.py
```diff
@@ -0,0 +1,68 @@
...
+
+
+class ActorEnvironment:
+
+    def __init__(self, config, actor_rollout_wg, reward_fn):
+        self.config = config
+        self.actor_rollout_wg = actor_rollout_wg
+        self.reward_fn = reward_fn
+
+    def step(self, batch, gen_batch, timing_raw):
```
vermouth1992, line 48: Could you provide a high-level abstraction of what step and update should do and how to extend them to new scenarios?

vermouth1992: Just add some docstring will have great help

casper-hansen: I just added docstrings for step and update, making the functionality more obvious to users. Extension will come later in a later PR, I hope the current docstrings will be helpful in guiding users.

casper-hansen: @vermouth1992 Please let me know if we can resolve this conversation and get this part merged. Then I will start the work on part 2.

vermouth1992: LGTM. @PeterSH6 Could you please comment also?

vermouth1992: Hi @casper-hansen,

After some discussions, I guess it's better to demonstrate an example that utilizes this API to accomplish some tasks such as a fake search or something like that before it is merged. Thanks!

PeterSH6:
Cool feature! I have two questions regarding this refactor:

1. How to implement vLLM SPMD async with the ActorEnvironment? Correct me if I'm wrong: It seems that the `step` function assumes that all the `responses` in this batch will be returned before moving to the next stage. Therefore, if we want to "pipeline" the generation with later experience preparation stage, we have to break this ActorEnvironment.

2. If some researchers would like to design a dataflow where the actor will interact with some environment like OpenAI GYMs (not tools), it seems that they still need to modify the whole ray_trainer.py to pass the environments into the `actor_environement`.

casper-hansen:
@PeterSH6 All great questions.

1. Keep in mind this is just a refactor of the existing code where we wait for all parts of the batch to finish before moving on. We can introduce a form of pipelining where we only process active experiences, but I think you still have to wait for the batch to finish. The main functionality of this PR is allowing users to create separate ActorEnvironment's.
2. The goal is to make sure the user does not have to modify the ray_trainer.py The assumption is that the configuration file could contain additional information since that is passed in, however, that might not be a sufficient solution.
    - Would you like me to work on making it easier to pass additional information for custom dataflows into the ActorEnvironment class? I could implement a `actor_env_kwargs` or let the user create a partial ActorEnvironment and fill in the missing arguments from `init_workers()`.
3. I will attempt a vLLM SPMD using offline async as I think this is the cleanest implementation. You can stream partial outputs in offline mode if you use the `AsyncLLMEngine` from vLLM, enabling interleaved tool calling as described in the RFC.
    - I expect this to be hard to implement. If I am unsuccessful, I will most likely reuse RAGEN/Search-R1.
    - I have created an unrelated example in my framework AutoAWQ if you want to understand how this works https://casper-hansen.github.io/AutoAWQ/examples/#vllm

verl/trainer/ppo/actor_env.py
casper-hansen: Hi @vermouth1992. It will be hard to develop the example before this PR is merged. If I develop both things at the same time, there will be a lot of merge conflicts to resolve as it will be developed on an old commit, making it less feasible.

This is why part 1 is about allowing extensibility in veRL because then part 2 can work with the latest changes on the main branch.

What do you think about this?

casper-hansen: @vermouth1992 If it helps, I can import the Search-R1 and adapt it to DuckDuckGo in an example that uses a custom `ActorEnvironment`. This will be an easier feature because I don't need to develop anything that's new.

PR #564, opened 2025-03-12
[config] feat: lr_warmup_steps

tongyx361:
This PR adds the `lr_warmup_steps` configuration.

Note the `num_warmup_steps` is prior to `lr_warmup_steps_ratio`.

### DIFF
--- a/docs/examples/config.rst
+++ b/docs/examples/config.rst
@@ -91,6 +91,7 @@ Actor/Rollout/Reference Policy
       ulysses_sequence_parallel_size: 1 # sp size
       optim:
         lr: 1e-6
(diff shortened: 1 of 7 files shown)


PeterSH6 (approved):

BearBiscuit05:
It seems that the `lr_warmup_steps` also need to be added to the `generation.yaml`.

tongyx361:
> It seems that the `lr_warmup_steps` also need to be added to the `generation.yaml`.

Could you help explain why? There is no `actor_rollout_ref.actor.optim` configuration in `generation.yaml` nor need to schedule learning rate in generation.

BearBiscuit05:
You are right, I have fixed this on #591

PR #565, opened 2025-03-12
misc: remove redundant .to(device)

dirtyDan0:
As a `DataProto` instance, calling `to(device)` already moves data.batch to the specified device.

https://github.com/volcengine/verl/blob/329dcfe1dd60f2d736ee55914e2a49e1887718eb/verl/protocol.py#L324-L336


hiyouga:
Should we also remove them in `compute_ref_log_prob` and `compute_log_prob`?

hiyouga (approved):

PR #575, opened 2025-03-13
fix: remove redundant torch.cuda.empty_cache()

wuxibin89:
#556 take effort to remove remove unnecessary empty_cache, but will cause CUDA oom at vllm wake_up.
```text
  File "/opt/tiger/ray/session_2025-03-13_12-11-30_408315_2895/runtime_resources/working_dir_files/_ray_pkg_a64b690733067c5c/verl/workers/fsdp_workers.py", line 481, in generate_sequences
    with self.rollout_sharding_manager:
  File "/opt/tiger/ray/session_2025-03-13_12-11-30_408315_2895/runtime_resources/working_dir_files/_ray_pkg_a64b690733067c5c/verl/workers/sharding_manager/fsdp_vllm.py", line 82, in __enter__
    self.inference_engine.wake_up()
  File "/usr/local/lib/python3.11/dist-packages/vllm/entrypoints/llm.py", line 1244, in wake_up
    self.llm_engine.wake_up()
  File "/usr/local/lib/python3.11/dist-packages/vllm/engine/llm_engine.py", line 1859, in wake_up
    self.model_executor.wake_up()
  File "/usr/local/lib/python3.11/dist-packages/vllm/executor/executor_base.py", line 216, in wake_up
    self.collective_rpc("wake_up")
  File "/usr/local/lib/python3.11/dist-packages/vllm/executor/uniproc_executor.py", line 56, in collective_rpc
    answer = run_method(self.driver_worker, method, args, kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/dist-packages/vllm/utils.py", line 2196, in run_method
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/dist-packages/vllm/worker/worker.py", line 140, in wake_up
    allocator.wake_up()
  File "/usr/local/lib/python3.11/dist-packages/vllm/device_allocator/cumem.py", line 207, in wake_up
    create_and_map(handle)
  File "/usr/local/lib/python3.11/dist-packages/vllm/device_allocator/cumem.py", line 75, in create_and_map
    python_create_and_map(*allocation_handle)
RuntimeError: CUDA Error: out of memory at /workspace/csrc/cumem_allocator.cpp:62
```
[...]

PR #586, opened 2025-03-13
fix: remove a empty_cache #575 missed

dirtyDan0:
Pending...
***
~~https://github.com/volcengine/verl/pull/575 removed all `torch.cuda.empty_cache()` in `verl/workers/fsdp_workers.py`, but fsdp_workers calls `offload_fsdp_model_to_cpu`, which has `torch.cuda.empty_cache()` that was missed.~~

https://github.com/volcengine/verl/blob/e7c40b3531f82d4502a0bf8b74f0d3796f9dac82/verl/workers/fsdp_workers.py#L531

https://github.com/volcengine/verl/blob/e7c40b3531f82d4502a0bf8b74f0d3796f9dac82/verl/utils/fsdp_utils.py#L115-L132

### DIFF
--- a/verl/utils/fsdp_utils.py
+++ b/verl/utils/fsdp_utils.py
@@ -112,7 +112,7 @@ def get_fsdp_wrap_policy(module, config=None, is_lora=False):
 
 
 @torch.no_grad()
-def offload_fsdp_model_to_cpu(model: FSDP, empty_cache: bool = True):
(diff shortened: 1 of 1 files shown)


hiyouga (changes requested):
I think this modification do have side-effect to the VRAM usage. Here is my experimental result.
image

PeterSH6:
Not all empty_cache could be deleted.
We should keep it first based on @hiyouga experiment

PR #587, opened 2025-03-13
fix: use weight iterator instead of list

hiyouga:
## What does this PR do?

Follow #543 

We should avoid materializing a list of full tensors because it may result in OOM. Use weight iterator to replace it.

## Who can review?

@vermouth1992 @PeterSH6

### DIFF
--- a/verl/workers/sharding_manager/fsdp_vllm.py
+++ b/verl/workers/sharding_manager/fsdp_vllm.py
@@ -15,7 +15,9 @@
 import os
 import logging
 import torch
+from typing import Dict, Iterable, Union, Tuple
 import numpy as np
+from torch.distributed._tensor import DTensor
 from torch.distributed.fsdp.fully_sharded_data_parallel import FullyShardedDataParallel as FSDP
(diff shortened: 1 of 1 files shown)


verl/workers/sharding_manager/fsdp_vllm.py
```diff
@@ -92,10 +105,8 @@ def __enter__(self):
             self.inference_engine.sync_model_weights(params, load_format=load_format)
         else:
             self.inference_engine.wake_up()
-            world_size = torch.distributed.get_world_size()
             model = self.inference_engine.llm_engine.model_executor.driver_worker.worker.model_runner.model
-            loaded_params = model.load_weights(
-                ((name, param.full_tensor() if world_size != 1 else param) for name, param in params.items()))
+            loaded_params = model.load_weights(self._make_weight_iterator(params))
```
wuxibin89, line 109: `(name, param.full_tensor() if world_size != 1 else param) for name, param in params.items())` here is a generator expression not list, it's not materialize when calling load_weights.

hiyouga: oh  i get it

PR #309, opened 2025-02-19
docs: add an example for Ray on Slurm

danielz02:
A working Slurm example adapted from https://docs.ray.io/en/latest/ray-core/starting-ray.html

PR #585, opened 2025-03-13
doc: add multinode training and debug tutorial

wuxibin89:
#354

PR #588, opened 2025-03-13
docs: fix hardcoded parameters in the Slurm example

danielz02:
Follow-up to https://github.com/volcengine/verl/pull/309


vermouth1992 (approved):

wuxibin89:
@danielz02 Hi, I wrote a multinode training tutorial in #585 , can you help supplementing the slurm part, cause we don't slurm cluster.

Issue #589, opened 2025-03-14
Setting CUDA_VISIBLE_DEVICES for Ray Actors


jinqinn:
I am encountering an issue with Ray's distributed task execution where the worker nodes do not seem to recognize the CUDA_VISIBLE_DEVICES environment variable. Despite setting the CUDA_VISIBLE_DEVICES variable correctly on the environment before launching the Ray workers, they do not use the specified GPUs for computation.

kevin85421:
Hi @jinqinn, I am Kai-Hsun from the Ray team. Would you mind providing a simple Ray example to explain what's the actual behavior and what's your expected behavior? Thanks!

jinqinn:
> Hi [@jinqinn](https://github.com/jinqinn), I am Kai-Hsun from the Ray team. Would you mind providing a simple Ray example to explain what's the actual behavior and what's your expected behavior? Thanks!

```
import os

os.environ["CUDA_VISIBLE_DEVICES"] = "3"

import ray
import torch

ray.init(ignore_reinit_error=True)

@ray.remote(num_gpus=1)
class GpuActor:
    def __init__(self):
        self.device = torch.device("cuda" if torch.cuda.is_available() else "cpu")
        print(f"Actor initialized on device: {self.device}")
        print(f"CUDA_VISIBLE_DEVICES in the actor: {os.environ.get('CUDA_VISIBLE_DEVICES')}")

    def get_device_info(self):
        if self.device.type == 'cuda':
            print(f"Using GPU: {torch.cuda.current_device()} - {torch.cuda.get_device_name(self.device)}")
        else:
            print("No GPU detected.")
        return os.environ.get("CUDA_VISIBLE_DEVICES")

actor = GpuActor.remote()
result = ray.get(actor.get_device_info.remote())
print(f"CUDA_VISIBLE_DEVICES in the head node: {os.environ.get('CUDA_VISIBLE_DEVICES')}, CUDA_VISIBLE_DEVICES returned by actor: {result}")
```

I am using Ray to distribute tasks across multiple workers, and I want to ensure that the Ray actor is able to correctly identify and use the CUDA_VISIBLE_DEVICES environment variable. This variable controls which GPUs are visible to the program, and it is set in the head node. However, I need to ensure that when the actor is launched, it is aware of this environment variable and correctly selects the available GPU.

wuxibin89:
[...]

PR #590, opened 2025-03-14
[fix] support setting CUDA_VISIBLE_DEVICES environment variable for Ray tasks

jinqinn:
https://github.com/volcengine/verl/issues/589
This update introduces support for setting the CUDA_VISIBLE_DEVICES environment variable for tasks running in Ray. By configuring this environment variable, users can control which GPUs are visible to Ray workers, providing better management of resources, especially in multi-GPU environments.

### DIFF
--- a/verl/trainer/main_ppo.py
+++ b/verl/trainer/main_ppo.py
@@ -16,6 +16,7 @@ Note that we don't combine the main with ray_trainer as ray_trainer is used by o
 """
 from verl.trainer.ppo.ray_trainer import RayPPOTrainer
 
+import os
 import ray
 import hydra
 
@@ -59,6 +60,7 @@ def run_ppo(config) -> None:
         # this is for local ray cluster
         ray.init(runtime_env={
             'env_vars': {
+                'CUDA_VISIBLE_DEVICES': os.environ.get('CUDA_VISIBLE_DEVICES'),
                 'TOKENIZERS_PARALLELISM': 'true',
                 'NCCL_DEBUG': 'WARN',
                 'VLLM_LOGGING_LEVEL': 'WARN'


verl/trainer/main_ppo.py
```diff
@@ -59,6 +60,7 @@ def run_ppo(config) -> None:
         # this is for local ray cluster
         ray.init(runtime_env={
             'env_vars': {
+                'CUDA_VISIBLE_DEVICES': os.environ.get('CUDA_VISIBLE_DEVICES'),
```
wuxibin89, line 63: Ray will automatically set `CUDA_VISIBLE_DEVICES ` for actor/task according to `num_gpus` in @ray.remote. For example, 
- @ray.remote(num_gpus=1): actor/task will be set CUDA_VISIBLE_DEVICES=0
- @ray.remote(num_gpus=2): actor/task will be set CUDA_VISIBLE_DEVICES=0,1
- @ray.remote(num_cpus=1): actor/task will be set CUDA_VISIBLE_DEVICES='', meaning that it can't see any GPU
The main_task can't see GPU is because it's only assigned cpu: https://github.com/volcengine/verl/blob/main/verl/trainer/main_ppo.py#L71

jinqinn: > Ray will automatically set `CUDA_VISIBLE_DEVICES ` for actor/task according to `num_gpus` in @ray.remote. For example,
> 
> * @ray.remote(num_gpus=1): actor/task will be set CUDA_VISIBLE_DEVICES=0
> * @ray.remote(num_gpus=2): actor/task will be set CUDA_VISIBLE_DEVICES=0,1
> * @ray.remote(num_cpus=1): actor/task will be set CUDA_VISIBLE_DEVICES='', meaning that it can't see any GPU
>   The main_task can't see GPU is because it's only assigned cpu: https://github.com/volcengine/verl/blob/main/verl/trainer/main_ppo.py#L71

Thank you for the clarification! However, I would like to mention that when setting @ray.remote(num_gpus=2), it doesn't seem to allow the GPUs to be scheduled on specific devices such as CUDA_VISIBLE_DEVICES=3,4. Instead, it always assigns the GPUs starting from device 0.
It would be much more convenient if the environment variable could be automatically set according to the selected GPUs, so that we could directly specify which GPUs to use without having to manage this manually.

wuxibin89: That's unexpected, can you print `CUDA_VISIBLE_DEVICES` of each actor/task and post a reproduce script?

jinqinn: > That's unexpected, can you print `CUDA_VISIBLE_DEVICES` of each actor/task and post a reproduce script?

```
import os

os.environ["CUDA_VISIBLE_DEVICES"] = "3"

import ray
import torch

ray.init(ignore_reinit_error=True)

@ray.remote(num_gpus=1)
class GpuActor:
    def __init__(self):
        self.device = torch.device("cuda" if torch.cuda.is_available() else "cpu")
        print(f"Actor initialized on device: {self.device}")
        print(f"CUDA_VISIBLE_DEVICES in the actor: {os.environ.get('CUDA_VISIBLE_DEVICES')}")

    def get_device_info(self):
        if self.device.type == 'cuda':
            print(f"Using GPU: {torch.cuda.current_device()} - {torch.cuda.get_device_name(self.device)}")
        else:
            print("No GPU detected.")
        return os.environ.get("CUDA_VISIBLE_DEVICES")

actor = GpuActor.remote()
result = ray.get(actor.get_device_info.remote())
print(f"CUDA_VISIBLE_DEVICES in the head node: {os.environ.get('CUDA_VISIBLE_DEVICES')}, CUDA_VISIBLE_DEVICES returned by actor: {result}")
```

pip list | grep ray
ray                               2.42.1

output:
2025-03-14 02:36:27,884 INFO worker.py:1832 -- Connected to Ray cluster. View the dashboard at 127.0.0.1:8265 
(GpuActor pid=9721) Actor initialized on device: cuda
(GpuActor pid=9721) CUDA_VISIBLE_DEVICES in the actor: 4
(GpuActor pid=9721) Using GPU: 0 - NVIDIA A100-PCIE-40GB
CUDA_VISIBLE_DEVICES in the head node: **3**, CUDA_VISIBLE_DEVICES returned by actor: **4**

wuxibin89: ~~Unlike torchrun, ray do not respect CUDA_VISIBLE_DEVICES when you `ray start` or set in main.py(like you do above), and it will use all GPUs in this node, and automatically scheduler actor/task to these GPUs(by setting CUDA_VISIBLE_DEVICES).~~
- ~~CUDA_VISIBLE_DEVICES=0,1,2,3 torchrun train_script.py: torchrun will only use GPU 0~4~~
- ~~CUDA_VISIBLE_DEVICES=0,1,2,3 ray start --head: ray will ignore CUDA_VISIBLE_DEVICES and still use whole 8 GPUs~~

jinqinn: I retested it, and when I set the environment variable for the Ray actor, 
```
os.environ["CUDA_VISIBLE_DEVICES"] = "3"
ray.init(runtime_env={
    'env_vars': {
        'CUDA_VISIBLE_DEVICES': os.environ.get('CUDA_VISIBLE_DEVICES'),
    }
})
```
the logs were as follows:
(GpuActor pid=452176) Actor initialized on device: cuda
(GpuActor pid=452176) CUDA_VISIBLE_DEVICES in the actor: 3
(GpuActor pid=452176) Using GPU: 0 - NVIDIA A100-PCIE-40GB
CUDA_VISIBLE_DEVICES in the head node: 3, CUDA_VISIBLE_DEVICES returned by actor: 3

When I didn't set the environment variable for the Ray actor, 
```
os.environ["CUDA_VISIBLE_DEVICES"] = "3"
ray.init()
```
the logs were as follows:
(GpuActor pid=9722) Actor initialized on device: cuda
(GpuActor pid=9722) CUDA_VISIBLE_DEVICES in the actor: 0
(GpuActor pid=9722) Using GPU: 0 - NVIDIA A100-PCIE-40GB
CUDA_VISIBLE_DEVICES in the head node: 3, CUDA_VISIBLE_DEVICES returned by actor: 0

wuxibin89: From ray-2.43.0, if user set CUDA_VISIBLE_DEVICES, then ray will respect it:
https://github.com/ray-project/ray/blob/ray-2.43.0/python/ray/_private/worker.py#L1016-L1021

wuxibin89: But it's still not reasonable to set CUDA_VISIBLE_DEVICES in `ray.init`, since all actor/task will use this value.
```python
os.environ["CUDA_VISIBLE_DEVICES"] = "3"
ray.init(runtime_env={
    'env_vars': {
        'CUDA_VISIBLE_DEVICES': os.environ.get('CUDA_VISIBLE_DEVICES'),
    }
})
```

jinqinn: > From ray-2.43.0, if user set CUDA_VISIBLE_DEVICES, then ray will respect it: https://github.com/ray-project/ray/blob/ray-2.43.0/python/ray/_private/worker.py#L1016-L1021

👍

PR #601, opened 2025-03-14
Fix a bug in loss mask calculation in sft_dataset.py

DaoD:
It seems that there is a bug in the SFTDataset.

While it is uncommon to truncate sequence from the left side, but it is often used for SFT the base model (no chat template used). When the truncation is set as "left", the current loss_mask calculation seems incorrect. It does not consider the prompt_length changing by the truncation operation. This PR is used to fix this.

### DIFF
--- a/verl/utils/dataset/sft_dataset.py
+++ b/verl/utils/dataset/sft_dataset.py
@@ -148,7 +148,7 @@ class SFTDataset(Dataset):
             if self.truncation == 'left':
                 # actually, left truncation may not be reasonable
                 truncated_prompt_length = sequence_length - max_length
-                prompt_length = max(0, prompt_length - tokens_to_remove)
+                prompt_length = max(0, prompt_length - truncated_prompt_length)
                 input_ids = input_ids[-self.max_length:]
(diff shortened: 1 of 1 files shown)


verl/utils/dataset/sft_dataset.py
```diff
@@ -147,6 +147,8 @@ def __getitem__(self, item):
         elif sequence_length > self.max_length:
             if self.truncation == 'left':
                 # actually, left truncation may not be reasonable
+                tokens_to_remove = sequence_length - max_length
+                prompt_length = max(0, prompt_length - tokens_to_remove)
```
hiyouga: can we directly set prompt_length to `len(input_ids)`?

hiyouga: tokens_to_remove is somehow confusing

DaoD: Not really, `len(input_ids)` is the length of the whole sequence, while prompt_length is only the length of the prompt part. I have change the parameter name into truncated_prompt_length, which may be more readable. Please check it.

hiyouga (approved):

PR #602, opened 2025-03-14
fix: concat metrics in DataProto.concat

wuxibin89:
Metrics should be in `non_tensor_batch` instead of `meta_info`, as DataProto not concat meta_info. If set in meta_info, PPOTrainer only get metrics of DP rank 0.

### DIFF
--- a/verl/workers/megatron_workers.py
+++ b/verl/workers/megatron_workers.py
@@ -367,7 +367,9 @@ class ActorRolloutRefWorker(MegatronWorker):
 
         log_gpu_memory_usage('After update policy', logger=logger)
 
-        output = DataProto(non_tensor_batch={metric: np.array([value] if np.isscalar(value) else value) for metric, value in metrics.items()})
+        output = DataProto(non_tensor_batch={
+            metric: np.array([value] if np.isscalar(value) else value) for metric, value in metrics.items()
+        })
         torch.cuda.empty_cache()
         return output
 
@@ -645,7 +647,9 @@ class CriticWorker(MegatronWorker):
         global_num_tokens = data.meta_info['global_token_num']
         estimated_flops, promised_flops = self.flops_counter.estimate_flops(global_num_tokens, delta_time)
         metrics['mfu/critic'] = estimated_flops * self.config.ppo_epochs / promised_flops / self.world_size
-        output = DataProto(non_tensor_batch={metric: np.array([value] if np.isscalar(value) else value) for metric, value in metrics.items()})
+        output = DataProto(non_tensor_batch={
+            metric: np.array([value] if np.isscalar(value) else value) for metric, value in metrics.items()
+        })
         return output
 
     @register(dispatch_mode=Dispatch.ONE_TO_ALL)
--- a/verl/workers/megatron_workers.py
+++ b/verl/workers/megatron_workers.py
@@ -21,7 +21,6 @@ import ray
 import torch
 import torch.distributed
 import torch.nn as nn
-import numpy as np
 from omegaconf import DictConfig
 
 from verl.single_controller.base.megatron.worker import MegatronWorker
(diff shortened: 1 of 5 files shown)


verl/workers/megatron_workers.py
```diff
@@ -366,9 +367,7 @@ def update_actor(self, data: DataProto):
 
         log_gpu_memory_usage('After update policy', logger=logger)
 
-        # TODO: here, we should return all metrics
-        output = DataProto(meta_info={'metrics': metrics})
-        output = output.to('cpu')
+        output = DataProto(non_tensor_batch={metric: np.array(value) for metric, value in metrics.items()})
```
vermouth1992: I guess it's better to give metric under a unified namespace in case we would like to add non-metric data into the non_tensor

vermouth1992: Also, I guess it's problematic because the non tensor requires to have the same batch size as batch

wuxibin89: Yes, actually there's 2 places where tensor and metrics return together, so metrics can't put into non_tensor_batch.
- https://github.com/volcengine/verl/blob/main/recipe/prime/prime_fsdp_workers.py#L278
- https://github.com/volcengine/verl/blob/main/recipe/prime/prime_fsdp_workers.py#L316

wuxibin89: Metric is so general that I suggest to add a new `.metrics` filed to DataProto, the reason are:
- metrics can't be set to `.non_tensor_batch` along with `.batch`, since these two fields need same batch size
- metrics can't be set to `.meta_info`, since this field should not be concat

New `.metrics` field should be much like `.meta_info`, except that it will be concat in `DataProto.concat`.

wuxibin89: Furthermore, should we remove `.non_tensor_batch`? Since it's designed as a dictionary contains numpy arrays with same batch size to `.batch`. For now it's mainly used to store multi_modal_inputs, can we also store them in `.batch`? cc @hiyouga

vermouth1992: Currently, verl relies on vllm 0.6.3 that requires torch 2.4 and tensordict 0.5. The latest tensordict supports non tensor data, and we can completely remove DataProto data structure when we upgrade.

vermouth1992: I guess we can either create a new field in DataProto that is concat during collection, or we can modify dispatch function to aggregate meta_info. However, it's hard to generalize what should we aggregate in meta_info.

vermouth1992: Another approach to handle metrics is to perform allgather inside workers

hiyouga: @wuxibin89 The multi modal features have different lengths (e.g. [[P, P, P], [P, P]]) so I put them in non tensor batch

hiyouga: We cannot pad the multi modal features because the pad tensor will consume much VRAM/RAM.

wuxibin89: > I guess we can either create a new field in DataProto that is concat during collection, or we can modify dispatch function to aggregate meta_info. However, it's hard to generalize what should we aggregate in meta_info.

It's not a good idea to aggregate meta_info,  it may contains any python objects.

hiyouga: Yeah. We expect the input and output of dataproto dispatch are identical, simply concatnating the meta_info breaks such rule

vermouth1992: It seems that the best approach might be to perform allgather/allreduce inside workers as if there is no single controller :)

wuxibin89: I don't tend to allreduce metrics inside workers, cause it introduce a strong barrier across all workers and make async impossible to driver.

hiyouga:
@wuxibin89 is this pr ready for merge?

PR #603, opened 2025-03-14
[bugfix] remove redundant and harmful metric appending

gameofdimension:
this line is unnecessary can cause error to metric reduce


verl/workers/actor/dp_actor.py
```diff
@@ -325,6 +325,5 @@ def update_policy(self, data: DataProto):
 
                 grad_norm = self._optimizer_step()
                 data = {'actor/grad_norm': grad_norm.detach().item()}
-            append_to_dict(metrics, data)
```
hiyouga, line 328: I think we can add indent to this line. The grad norm should be appended to the metrics

PR #618, opened 2025-03-16
[misc] feat: support token_level_loss and support different clip range for low and high

PeterSH6:
- As titled

### DIFF
--- a/verl/trainer/config/ppo_megatron_trainer.yaml
+++ b/verl/trainer/config/ppo_megatron_trainer.yaml
@@ -28,6 +28,9 @@ actor_rollout_ref:
     use_dynamic_bsz: False
     use_torch_compile: True # False to disable torch compile
     clip_ratio: 0.2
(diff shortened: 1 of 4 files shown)


verl/trainer/config/ppo_megatron_trainer.yaml
```diff
@@ -28,6 +28,9 @@ actor_rollout_ref:
     use_dynamic_bsz: False
     use_torch_compile: True # False to disable torch compile
     clip_ratio: 0.2
+    clip_ratio_low: 0.2
+    clip_ratio_high: 0.2
+    use_token_level_loss: False
```
tongyx361, line 33: Is `token_mean_loss` a better name?

```suggestion
    use_token_mean_loss: False
```

tongyx361 (approved):
This PR looks good to me.

The only minor issue is that `token_mean_loss` might be a better name.

Issue #598, opened 2025-03-14
install error


xiaoooobai:
when use   pip3 install -e . 
error happened:
ERROR: Could not find a version that satisfies the requirement math-verify (from verl) (from versions: none)
ERROR: No matching distribution found for math-verify

yuezhao238:
me too😣

PeterSH6:
Which torch version are you using? I can installed verl using torch 2.4 + cu12.4.

yuezhao238:
I resolved it after using an environment with python 3.10

HaotianYang03:
> I resolved it after using an environment with python 3.10. [huggingface/Math-Verify#38](https://github.com/huggingface/Math-Verify/issues/38)

Thank you so much,  it works for me

shizhediao:
Saved my day, thanks!

Kirrito-k423:
中文：这个安装问题看起来已经解决，可以关闭。原始报错是 `math-verify` 无法被 pip 找到；后续用户反馈使用 Python 3.10 环境后解决（https://github.com/verl-project/verl/issues/598#issuecomment-2726244953），另一名用户也确认该方案有效（https://github.com/verl-project/verl/issues/598#issuecomment-2726473073）。同时，PR #619 已合并并专门修正了 #598 相关的安装说明（https://github.com/verl-project/verl/pull/619），当前安装文档要求 Python >=3.10，`pyproject.toml` 也声明 `requires-python = ">=3.10"`。

English: This installation issue appears to be resolved and can be closed. The original failure was that pip could not find `math-verify`; later comments confirmed that using a Python 3.10 environment fixed the problem (https://github.com/verl-project/verl/issues/598#issuecomment-2726244953), and another user confirmed the same workaround worked for them (https://github.com/verl-project/verl/issues/598#issuecomment-2726473073). PR #619 was merged specifically to fix the installation instructions for #598 (https://github.com/verl-project/verl/pull/619), and the current install docs require Python >=3.10 while `pyproject.toml` declares `requires-python = ">=3.10"`.

This response was generated by a scheduled Codex GPT5.5 issue processing task.

PR #619, opened 2025-03-16
[fix] fix python env issue in install

Django-Jiang:
Fix issue mentioned #598 by resetting python to `3.10`


eric-haibin-lin (approved):
thx!

PR #623, opened 2025-03-16
[recipe] feat: integrate DAPO and provide reproduction script

tongyx361:
> [!WARNING]
> As mentioned in https://github.com/volcengine/verl/pull/623#issuecomment-2733688593, the implementation of gradient accumulation in verl has been only compatible with the sequence-mean loss, but all the DAPO experiments with the token-mean loss were run with the incompatible implementation.
> **We keep it as is for reproducibility in this branch** and will fix it in another PR for the main branch.


verl/protocol.py
```diff
@@ -370,6 +370,41 @@ def select(self, batch_keys=None, non_tensor_batch_keys=None, meta_info_keys=Non
 
         return DataProto(batch=sub_batch, non_tensor_batch=non_tensor_batch, meta_info=sub_meta_info)
 
+    def sel_idxs(self, idxs):
```
PeterSH6: use `select`

verl/trainer/config/ppo_trainer.yaml
```diff
@@ -170,6 +181,10 @@ algorithm:
   kl_ctrl:
     type: fixed
     kl_coef: 0.001
+  filter_groups:
```
PeterSH6: Add them in megatron to avoid breaking CI

verl/trainer/config/ppo_trainer.yaml
```diff
@@ -170,6 +181,10 @@ algorithm:
...
+    enable: False
+    fill_train_batch: True
```
PeterSH6:
```suggestion
    fill_to_train_batch_batch: True
```

verl/trainer/ppo/metric_utils.py
```diff
@@ -166,3 +167,38 @@ def compute_throughout_metrics(batch: DataProto, timing_raw: Dict[str, float], n
...
+
+
+def bootstrap_metric(data: list[dict[str, Any]],
+                     subset_size: int,
+                     reduce_fns: list[Callable[[np.ndarray], float]],
+                     n_bootstrap: int = 1000,
+                     seed: int = 42) -> list[tuple[float, float]]:
+    """
+    Bootstrap the metric to get the confidence interval
+    """
```
PeterSH6:
```suggestion
```

verl/workers/reward_manager/naive.py
```diff
@@ -15,17 +15,31 @@
...
+                 max_resp_len=None,
+                 overlong_buffer_cfg=None) -> None:
         self.tokenizer = tokenizer
         self.num_examine = num_examine  # the number of batches of decoded responses to print to the console
         self.compute_score = compute_score or _default_compute_score
+        self.reward_fn_key = reward_fn_key
+        self.overlong_buffer_cfg = overlong_buffer_cfg
+        self.max_resp_len = max_resp_len
 
+        if self.overlong_buffer_cfg is not None:
```
PeterSH6:
```suggestion
        if self.overlong_buffer_cfg is not None and self.overlong_buffer_cfg.enable:
```

verl/workers/reward_manager/naive.py
```diff
@@ -90,20 +108,48 @@ def __call__(self, data: DataProto):
...
+            for key, value in result.items():
+                reward_extra_info[key].append(value)
+
+            final_reward += reward
+
+            overlong_buffer_len = self.overlong_buffer_cfg.len
+            if overlong_buffer_len > 0:
+                overlong_penalty_factor = self.overlong_buffer_cfg.penalty_factor
+                exceed_len = valid_response_length - (self.max_resp_len - overlong_buffer_len)
+                overlong_reward = max(-exceed_len / overlong_buffer_len * overlong_penalty_factor, 0)
```
PeterSH6:
```suggestion
                overlong_reward = min(-exceed_len / overlong_buffer_len * overlong_penalty_factor, 0)
```

verl/trainer/ppo/ray_trainer.py
```diff
@@ -827,6 +888,76 @@ def fit(self):
...
+                        for uid, tok_rewards in zip(batch.non_tensor_batch['uid'], batch.batch['token_level_rewards']):
+                            seq_reward = torch.sum(tok_rewards).item()
+                            uid2seq_rewards[uid].append(seq_reward)
+
+                        uid2seq_reward_std = {}
+                        for uid, seq_rewards in uid2seq_rewards.items():
+                            uid2seq_reward_std[uid] = np.std(seq_rewards)
+
+                        kept_uids = [uid for uid, std in uid2seq_reward_std.items() if std > 0]
+                        filter_metric_dict["non_uni_rew_prompt_ratio"] = len(kept_uids) / len(uid2seq_rewards)
```
PeterSH6: make it clearly: 
non_uniform_reward_prompt_ratio

verl/trainer/ppo/ray_trainer.py
```diff
@@ -827,6 +888,76 @@ def fit(self):
...
+                        filter_metric_dict["non_uni_rew_prompt_bsz"] = len(kept_uids)
+
+                        kept_idxs = []
+
+                        train_prompt_bsz = len(batch.batch)
```
PeterSH6: need to read from the config

verl/trainer/ppo/ray_trainer.py
```diff
@@ -827,6 +888,76 @@ def fit(self):
...
+                                if len(kept_uids) == train_prompt_bsz:
+                                    break
+
+                        for idx, uid in enumerate(batch.non_tensor_batch['uid']):
+                            if uid in kept_uids:
+                                kept_idxs.append(idx)
+                        filter_metric_dict["non_uni_rew_traj_bsz"] = len(kept_idxs)
+
+                        world_size = self.actor_rollout_wg.world_size
+                        kept_idxs = kept_idxs[:len(kept_idxs) // world_size * world_size]
```
PeterSH6: NOT FIXED: 
```python3
if len(kept_idx) == 0:
    continue
```

verl/trainer/config/ppo_trainer.yaml
```diff
@@ -31,6 +33,9 @@ actor_rollout_ref:
     ppo_max_token_len_per_gpu: 16384 # n * ${data.max_prompt_length} + ${data.max_response_length}
     grad_clip: 1.0
     clip_ratio: 0.2
+    clip_ratio_low: 0.2
+    clip_ratio_high: 0.2
+    use_token_level_loss: False
```
YuMS: It appears that the behavior for computing pg loss prior to add the `use_token_level_loss` flag was already at token level. Should we set this to `True` for sake of backward compatibility?

YuMS:
According to your calculation, `use_token_level_loss` should set to True. However when mini_batch_size != micro_batch_size. Verl still calculate the sample level averaged loss between micro batches.

Should the pg_loss, as long as `entropy_loss` and `kl_loss`(included in policy_loss), be reweighed by sum of sample length of current micro batch? The current implement is as follows:
```
  if self.config.use_dynamic_bsz:
      # relative to the dynamic bsz
      loss = policy_loss * (len(data) / self.config.ppo_mini_batch_size)
  else:
      loss = policy_loss / self.gradient_accumulation
```
https://github.com/volcengine/verl/pull/623/files#diff-e06e73d3a7775a502b7aea91103e7911f6597eb48e4b898db558766cdd41daf9R318

yqy2001:
Great catch! We didn't account for this in our run. We'll experiment with it to understand its effect and see whether this affects our previous conclusion.

> According to your calculation, `use_token_level_loss` should set to True. However when mini_batch_size != micro_batch_size. Verl still calculate the sample level averaged loss between micro batches.
> 
> Should the pg_loss, as long as `entropy_loss` and `kl_loss`(included in policy_loss), be reweighed by sum of sample length of current micro batch? The current implement is as follows:
> 
> ```
>   if self.config.use_dynamic_bsz:
>       # relative to the dynamic bsz
>       loss = policy_loss * (len(data) / self.config.ppo_mini_batch_size)
>   else:
>       loss = policy_loss / self.gradient_accumulation
> ```
> 
> https://github.com/volcengine/verl/pull/623/files#diff-e06e73d3a7775a502b7aea91103e7911f6597eb48e4b898db558766cdd41daf9R318

verl/trainer/main_ppo.py
```diff
@@ -146,10 +146,20 @@ def main_task(config):
         raise NotImplementedError
 
     compute_score = get_custom_reward_fn(config)
-    reward_fn = reward_manager_cls(tokenizer=tokenizer, num_examine=0, compute_score=compute_score)
+    reward_fn = reward_manager_cls(tokenizer=tokenizer,
+                                   num_examine=0,
+                                   compute_score=compute_score,
+                                   reward_fn_key=config.data.reward_fn_key,
+                                   max_resp_len=config.data.max_response_length,
+                                   overlong_buffer_cfg=config.custom_reward_function.overlong_buffer)
```
vermouth1992: I'm afraid that this makes the reward manager too specific to DAPO

vermouth1992: Reward Manager should be able to be pluggable and DAPO can implement its own reward in the recipe folder

verl/trainer/ppo/ray_trainer.py
```diff
@@ -314,6 +316,23 @@ def _validate_config(self):
         # number of GPUs total
         n_gpus = config.trainer.n_gpus_per_node * config.trainer.nnodes
 
+        filter_cfg = config.algorithm.filter_groups
```
vermouth1992: This also makes the main function too DAPO specific

vermouth1992:
We can upstream compatible changes such as clip high/low, token-level loss first

vermouth1992:
In the long run, we will have to make worker to be able to used as API and every algorithm construct their own ray_trainer.py

recipe/dapo/src/dapo_ray_trainer.py
saidineshpola, line 1: Please add missing AdvantageEstimator import from verl.trainer.ppo.ray_trainer

verl/trainer/ppo/ray_trainer.py
```diff
@@ -544,33 +550,87 @@ def _validate(self):
...
         data_sources = np.concatenate(data_source_lst, axis=0)
 
-        # evaluate test_score based on data source
-        data_source_reward = {}
-        for i in range(reward_tensor.shape[0]):
-            data_source = data_sources[i]
-            if data_source not in data_source_reward:
-                data_source_reward[data_source] = []
-            data_source_reward[data_source].append(reward_tensor[i].item())
+        data_src2prompt2var2vals = defaultdict(lambda: defaultdict(lambda: defaultdict(list)))
```
PeterSH6: Would it be possible to wrap it as a util function instead of implementing the reduce logic in ray_trainer?

verl/trainer/ppo/ray_trainer.py
```diff
@@ -544,33 +550,87 @@ def _validate(self):
...
+        for sample_idx, data_source in enumerate(data_sources):
+            prompt = sample_inputs[sample_idx]
+
+            var2vals = data_src2prompt2var2vals[data_source][prompt]
```
PeterSH6: What's the meaning of data_src2prompt2var2vals

PeterSH6:
LGTM!

PeterSH6 (approved):

PeterSH6:
Really nice job!!!

verl/trainer/ppo/ray_trainer.py
tongyx361: 1. `2` means `to`, constructing the multi-level dict
2. `var` means a kind of "variable" such as `acc`, `final_reward`
3. `vals` means the "values" of the `var`iables of the trajectories after some `prompt` from some `data_src` 

I will add comments in a future PR to make it clearer.

PR #629, opened 2025-03-17
Fix computer_score in fsdp_workers.py

Django-Jiang:
Fix issue when not switching the chat template, `rm_data` is undefined


vermouth1992 (approved):

PeterSH6:
@Django-Jiang Can you run the format script? `bash scripts/format.sh`

eric-haibin-lin (approved):
Just ran `bash scripts/format.sh` to fix the lint issue

PR #195, opened 2025-02-04
feat: Add multi-turn SFT support

xingyaoww:
# Multi-turn Conversation Fine-tuning Support

## Overview
This PR adds support for fine-tuning models on multi-turn conversations, including proper chat template handling and loss masking for assistant responses. The implementation includes support for the OpenHands SFT dataset and handles conversations up to 32k tokens.

## Key Features
- New `MultiTurnSFTDataset` class for handling multi-turn conversations
- Proper chat template integration using HuggingFace's `apply_chat_template`
- Smart loss masking that targets only assistant responses
- Support for both single-turn and multi-turn training in the same trainer
- Token length limiting and conversation filtering

## Implementation Details
1. Dataset:
   - Uses proper chat templates from the model's tokenizer
   - Handles system, user, and assistant messages
   - Supports conversations with multiple turns
   - Filters conversations exceeding token limits

2. Training:
   - Added `use_multiturn` flag in config
   - Added `messages_key` for multi-turn data format
   - Maintains backward compatibility with single-turn training
   - Works with existing features (FSDP, sequence parallel, etc.)

3. Examples and Tests:
   - Added OpenHands SFT dataset preprocessing script
   - Added multi-turn training example
   - Added comprehensive unit tests
   - Moved tests to appropriate locations

## Usage Example
```yaml
# Config for multi-turn training
data:
  use_multiturn: true
  messages_key: messages
  max_length: 32000
  truncation: right
```

##Testing
- Unit tests for dataset functionality
- Integration with existing training pipeline
- Example scripts tested with OpenHands dataset
- Coverage for both single-turn and multi-turn modes

# Documentation
- Added comments explaining multi-turn specific features
- Updated config defaults with multi-turn options
- Added example scripts with documentation
- Added preprocessing script with dataset-specific handling

---

ok.. this is another PR mostly done by OpenHands with me messaging it about 10 times..

It is still WIP as I'm testing it on a training job now, will report back if it works

PR #635, opened 2025-03-17
support multiturn prompt for SFT

seanexp:
# What is this PR about

Current implementation of [SFTDataset](https://github.com/volcengine/verl/blob/main/verl/utils/dataset/sft_dataset.py#L110-L121) only supports single turn prompt. This PR adds support for multi-turn prompt which is in messages format.

### DIFF
--- a/verl/utils/dataset/sft_dataset.py
+++ b/verl/utils/dataset/sft_dataset.py
@@ -18,7 +18,7 @@ SFT dataset
 Each parquet file contains
 """
 
-from typing import List, Union
+from typing import List, Union, Literal, Iterable
 
 import pandas as pd
 
@@ -43,8 +43,8 @@ class SFTDataset(Dataset):
                  prompt_dict_keys=None,
                  response_key='response',
                  response_dict_keys=None,
-                 max_length=1024,
-                 truncation='error'):
+                 max_length: int = 1024,
+                 truncation: Literal['error', 'left', 'right'] = 'error'):
         assert truncation in ['error', 'left', 'right']
         self.truncation = truncation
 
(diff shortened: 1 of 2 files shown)


verl/utils/dataset/sft_dataset.py
```diff
@@ -94,15 +94,15 @@ def series_to_item(ls):
             except Exception:
                 print(f'self.prompts={self.prompts}')
                 raise
-        self.prompts = self.prompts.tolist()
+        self.prompts = self.prompts.squeeze().tolist()
```
seanexp, line 97: When using `pandas==2.2.2`, this line raises an error complaining DataFrame does not have `.tolist()` method. So I added `.squeeze()` method.

verl/utils/dataset/sft_dataset.py
```diff
@@ -164,7 +171,7 @@ def __getitem__(self, item):
             # mask out prompt for SFT.
             loss_mask[:min(prompt_length, loss_mask.size(0)) - 1] = 0
         # mask out the last token in response
-        loss_mask[min(prompt_length + response_length, loss_mask.size(0)) - 1] = 0
+        # loss_mask[min(prompt_length + response_length, loss_mask.size(0)) - 1] = 0
```
seanexp, line 174: I have no idea why eos_token masking is needed so I erased the line. I'll revert this if the masking is needed.

seanexp:
I am aware of this [PR](https://github.com/volcengine/verl/pull/195), but I add this PR because the PR is WIP.

verl/utils/dataset/sft_dataset.py
eric-haibin-lin: @vermouth1992 could u help take a look?

PR #644, opened 2025-03-17
feat: support no reference model; fix KL issues

dirtyDan0:
### Before get started

Difference between KL penalty in reward and KL loss

>  [!TIP]
>
>  1. In-reward KL penalty
>
>
>  $$
>  r_t = r_{\varphi}(q, o_{\leq t}) - \beta\ \boxed{\log \frac{\pi_{\theta}(o_t | q, o_{<t})}{\pi_{\text{ref}}(o_t | q, o_{<t})}}
>  $$
>
>  2. KL Loss
>
>  $$
>  L^{\text{PPO}}(\theta) = \mathbb{E}_t [ \min(ratio_t A_t, \text{clip}(ratio_t, 1 - \epsilon, 1 + \epsilon) A_t) ]
>  $$
>
>  $$
>  \- \beta\ \boxed{D_{\text{KL}}(\pi_{\theta} || \pi_{\text{ref}})}
>  $$

### Problems

1. The current code doesn't support not using reference model

   This feature is half-implemented since the very first commit but never completed, e.g., `RayPPOTrainer` has an attribute `use_reference_policy` but it's always True since role_worker_mapping always has `Role.RefPolicy`.

2. Restriction of `use_kl_loss` 

   Currently, `use_kl_loss` determines whether to use in-reward kl penalty or kl loss. So we can not use **both or neither**.

   https://github.com/volcengine/verl/blob/87a813658fa24d3f3677c505a3d0f9680b177274/verl/trainer/ppo/ray_trainer.py#L875-L879

   https://github.com/volcengine/verl/blob/87a813658fa24d3f3677c505a3d0f9680b177274/verl/workers/actor/dp_actor.py#L299-L307

>  [!CAUTION]  
>
>  ### You may have unintentionally adopted in-reward KL penalty
>
>  For the experiments you've conducted, if you set `actor.use_kl_loss`=False or didn't set it (Default is False),***You unintentionally used in-reward KL penalty.*** If you don't want any KL, you should set `actor_rollout_ref.actor.use_kl_loss=False` and `algorithm.use_kl_in_reward=False` (or not to set them because they are the default config) after this commit.

3. Deprecated config

   After investigation, I guess Critic may used to be responsible for in-reward KL. But this feature seems paralyzed.

   1. Line 290, there may used to be `config.algorithm.kl_ctrl.target_kl` and `config.critic.kl_ctrl.horizon` , which are not supported currently.

      https://github.com/volcengine/verl/blob/3ec83117c3f3f10f858ff6cb4ca65e92a93b941b/verl/trainer/ppo/ray_trainer.py#L289-L293

   2. In `verl/workers/critic/megatron_critic.py` : redundant set of `self.kl_ctrl`

      https://github.com/volcengine/verl/blob/3b18b0eb740bdd289a1b9b776e1b0d1d9291c791/verl/workers/critic/megatron_critic.py#L69-L73

### What’s Changed?

1. Add support for not using reference model
2. Fixed the incomplete code of the KL controller.
3. A test case for using both kl terms
4. Some other misc issues in the code.

### How to disable reference model

* set `actor_rollout_ref.actor.use_kl_loss=False` and `algorithm.use_kl_in_reward=False` (They are by default False, so you can simply not set them)


examples/grpo_trainer/run_deepseek7b_llm.sh
```diff
@@ -14,9 +14,7 @@ python3 -m verl.trainer.main_ppo \
     actor_rollout_ref.model.use_remove_padding=True \
     actor_rollout_ref.actor.ppo_mini_batch_size=256 \
     actor_rollout_ref.actor.ppo_micro_batch_size_per_gpu=80 \
-    actor_rollout_ref.actor.use_kl_loss=True \
-    actor_rollout_ref.actor.kl_loss_coef=0.001 \
```
hiyouga: leave kl_loss_coef in the bash file

docs/examples/gsm8k_example.rst
```diff
@@ -156,7 +157,6 @@ The script of run_deepseek7b_llm.sh
       critic.ppo_micro_batch_size_per_gpu=32 \
       critic.model.fsdp_config.param_offload=False \
       critic.model.fsdp_config.optimizer_offload=False \
-      algorithm.kl_ctrl.kl_coef=0.001 \
```
hiyouga: leave kl coef in the file

verl/trainer/config/ppo_megatron_trainer.yaml
```diff
@@ -29,9 +29,10 @@ actor_rollout_ref:
     use_torch_compile: True # False to disable torch compile
     clip_ratio: 0.2
     entropy_coeff: 0.001
-    use_kl_loss: False # True for GRPO
-    kl_loss_coef: 0.001 # for grpo
-    kl_loss_type: low_var_kl # for grpo
+    global_kl:
```
hiyouga: why not use `kl_loss`? how can we explain `global_kl`?

hiyouga (commented):
Generally it looks good to me. But we should keep the configuration lines about kl in the bash files to exposure them to user

verl/trainer/config/ppo_megatron_trainer.yaml
dirtyDan0: 👌

docs/amd_tutorial/amd_build_dockerfile.md
```diff
@@ -97,14 +97,21 @@ PYTHONUNBUFFERED=1 python3 -m verl.trainer.main_ppo \
...
+ actor_rollout_ref.ref.enable=True \
  actor_rollout_ref.ref.log_prob_micro_batch_size_per_gpu=4 \
  critic.optim.lr=1e-5 \
  critic.model.path=$MODEL_PATH \
  critic.ppo_micro_batch_size_per_gpu=4 \
- algorithm.kl_ctrl.kl_coef=0.001 \
+ algorithm.in_reward_kl.enable=True \
+ algorithm.in_reward_kl.type=kl \
+ algorithm.in_reward_kl.kl_ctrl.type=fixed \
+ algorithm.in_reward_kl.kl_ctrl.coef=0.001 \
```
hiyouga: how about simply use coef=0 to indicate the kl term is disabled? we can check if it is enabled in code by `if coef > 1e-6`

hiyouga: however, this pr is already pretty enough

dirtyDan0: Good suggestion. I think maybe I should use coef==0 (coef<=1e-6) instead of enable==False to make the config more concise.  I will push a commit very soon.

docs/amd_tutorial/amd_build_dockerfile.md
```diff
@@ -97,14 +97,19 @@ PYTHONUNBUFFERED=1 python3 -m verl.trainer.main_ppo \
  actor_rollout_ref.actor.optim.lr=1e-6 \
  actor_rollout_ref.actor.ppo_mini_batch_size=64 \
  actor_rollout_ref.actor.ppo_micro_batch_size_per_gpu=4 \
+ actor_rollout_ref.actor.kl_loss.type=low_var_kl \
+ actor_rollout_ref.actor.kl_loss.coef=0.001 \
```
hiyouga: if there any trouble if both two kl terms are enabled?

dirtyDan0: Theoretically, the update of parameters will be more conservative. I think I can add a notice to remind the users.
Let users themselves decide whether to use both/either/neither.

dirtyDan0:
review needed🤗

hiyouga:
@vermouth1992 how about your opinion?

verl/trainer/ppo/ray_trainer.py
```diff
@@ -880,9 +884,9 @@ def fit(self):
                         batch.batch['token_level_scores'] = reward_tensor
 
                         # compute rewards. apply_kl_penalty if available
-                        if not self.config.actor_rollout_ref.actor.get('use_kl_loss', False):
+                        if self.config.algorithm.kl_ctrl.kl_coef > 1e-6:
```
vermouth1992: Why is the threshold 1e-6?

docs/examples/config.rst
```diff
@@ -206,6 +206,13 @@ Actor/Rollout/Reference Policy
 
     - Trading speed for GPU memory.
 
+- ``actor_rollout_ref.actor.use_kl_loss``: Whether to enable kl loss. Default is False.
+
+- ``actor_rollout_ref.actor.kl_loss_coef``: The coefficient of kl loss. Default is 0.001. Will be set to 0 when ``actor_rollout_ref.actor.use_kl_loss``=False.  If \> 1e-6, kl loss will be enabled.
+
+- ``actor_rollout_ref.actor.kl_loss_type``: Support ``kl``, ``abs``, ``mse`` and ``full``. How to calculate the kl divergence between actor and reference policy. For
```
hiyouga: add `low_var_kl`

verl/trainer/ppo/ray_trainer.py
hiyouga: @dirtyDan0 can you add a config key like `use_kl_in_reward` to indicate the effectiveness of kl_in_reward instead of using `1e-6`...

hiyouga: we can skip the computation of ref log probs if both `use_kl_loss` and `use_kl_in_reward` are disabled

dirtyDan0: we compute ref log probs when `self.use_reference_policy=True` <- `Role.RefPolicy in role_worker_mapping` <- `config.algorithm.use_kl_in_reward or config.actor_rollout_ref.actor.use_kl_loss`

verl/trainer/config/ppo_trainer.yaml
```diff
@@ -170,10 +170,13 @@ algorithm:
   gamma: 1.0
   lam: 1.0
   adv_estimator: gae
+  use_kl_in_reward: False
```
hiyouga, line 173: default value should be true because PPO use kl in reward

hiyouga, line 173: if `use_kl_in_reward` is false by default, we should update all the ppo scripts and unittests

hiyouga: however, we recommend to set it to true because use kl loss is false by default

dirtyDan0: updated

verl/trainer/config/ppo_trainer.yaml
```diff
@@ -170,7 +170,7 @@ algorithm:
   gamma: 1.0
   lam: 1.0
   adv_estimator: gae
-  use_kl_in_reward: False
+  use_kl_in_reward: True
```
hiyouga: so should we turn off `use_kl_in_reward` in the scripts whose `use_kl_loss` is true?

dirtyDan0: It depends on the adv estimator, e.g. for grpo we should use_kl_loss=True and use_kl_in_reward=False. I'll try to explicitly specifying this in all concerned scripts.

dirtyDan0: We should always explicitly set them and don't reply on the default config for scripts. Btw i think by default setting use_kl_loss=False and use_kl_in_reward=True may cause misunderstanding for new users like "why i should use kl penalty in reward and not use kl loss by default???". If we set both of them as False by default it will cause no problem if all scripts explictly specifying the corresponding kl config w.r.t **the adv estimator**.

hiyouga: sure you are right. basically we should update all the scripts and e2e ut scripts in this PR

dirtyDan0:
After investigation, I concluded the kl config for difference adv estimator as following. I will update the scripts under the following criteria:
* every rl script must specify its adv_estimator
* if not specified before, i will set `gae` as default
* for gae, I will delete all redundant config concerning `actor_rollout_ref.ref`
* if use kl, the corresponding coefficient and type will be explicitly set

| adv_estimator | kl_in_reward | kl_loss |
| ------------- | ------------ | ------- |
| gae           | ❌            | ❌       |
| grpo          | ❌            | ✅       |
| reinforce++   | ✅            | ❌       |
| remax         | ✅            | ❌       |
| rloo          | ✅            | ❌       |

hiyouga:
wait, shouldn't gae use kl_in_reward?

dirtyDan0:
https://arxiv.org/abs/1506.02438 gae doesn't use any kl term.

hiyouga:
@dirtyDan0 but Openai's PPO used https://arxiv.org/abs/1707.06347  gae in verl means ppo

dirtyDan0:
In https://arxiv.org/abs/1707.06347:
1. the kl is not in reward but in loss
image

2. It recommends PPO-Clip, while other methods serve as baselines.
image
image
KL serves as a poor baseline.
image

dirtyDan0:
https://spinningup.openai.com/en/latest/algorithms/ppo.html
image

hiyouga:
get it

dirtyDan0:
review needed🥲

hiyouga (approved):

vermouth1992 (approved):

PR #648, opened 2025-03-18
misc: change main_task to TaskRunner actor

### LINKED ISSUE #539: [BUG] _check_resource_available of verl/trainer/ppo/ray_trainer.py causes error
I was running grpo with 8 gpus in a node, and set `trainer.n_gpus_per_node=8`.
However, during the first training step after the first validation phase, I got an error: `ValueError: Total available GPUs 0 is less than total desired GPUs 8`.

I don't know the exact cause of the error though.

wuxibin89:
Use ray actor instead of task to run main_task
- Ray task is retried in system error(oom/segmentfault), which may cause unexpectedly behavior
- Actor is more trackable in ray dashboard, e.g logging/stacktrace/profile

close #539

### DIFF
--- a/verl/trainer/main_ppo.py
+++ b/verl/trainer/main_ppo.py
@@ -51,10 +51,6 @@ def get_custom_reward_fn(config):
 
 @hydra.main(config_path='config', config_name='ppo_trainer', version_base=None)
 def main(config):
-    run_ppo(config)
-
-
-def run_ppo(config) -> None:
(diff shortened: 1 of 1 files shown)


vermouth1992 (approved):

zhouxie000:
hi, any chance we could keep the run_ppo api? We are currently taking dependency on the function as entrypoint in our custom entryscript with a slightly different config setup. Thanks!

PR #650, opened 2025-03-18
Use vLLM V1 Engine without hotfix

fyqqyf:
Install the latest vLLM (before compatibility with torch2.6) to bypass the previous hotfixes mentioned in the documentation. @xihuai18


rbao2018:
问一下V1引擎还有memory leak和assert len(seq)==1 这样的问题吗？

hiyouga:
vllm 0.8.0 is posted and it is compatible with verl through my validation

rbao2018:
> vllm 0.8.0 is posted and it is compatible with verl through my validation

Thx, we are looking forward to verl & sglang to updating their torch requirement

eric-haibin-lin:
thank you for the PR but currently using vllm 0.8.2 or v0.8.3 works just fine.

PR #663, opened 2025-03-18
[tracking] swanlab add `verl` config

Zeyi-Lin:
Add `verl` as the `framework` parameter to the SwanLab config table, so more developers can see that this training comes from `verl`.


eric-haibin-lin (approved):

Zeyi-Lin:
I executed `script/format.sh`, and now the format check should pass

vermouth1992 (approved):

PR #669, opened 2025-03-19
optimize assertion logic

dingyuan-shi:
In main_ppo.py, the assertion will ensure critic and actor having the same strategy.
However, when using advatage estimator like gae, we do not need critic, making the assertion being redundant.
This PR optimize the assertion to avoid being to strict.

### DIFF
--- a/verl/trainer/main_ppo.py
+++ b/verl/trainer/main_ppo.py
@@ -87,15 +87,21 @@ class TaskRunner:
         tokenizer = hf_tokenizer(local_path)
         processor = hf_processor(local_path, use_fast=True)  # used for multimodal LLM, could be none
 
+        # judge whether use critic
+        if config.algorithm.adv_estimator in ['gae']:
+            use_critic = True
+        else:
+            use_critic = False
+            
         # define worker classes
         if config.actor_rollout_ref.actor.strategy == 'fsdp':
-            assert config.actor_rollout_ref.actor.strategy == config.critic.strategy
+            assert (not use_critic) or (use_critic and config.actor_rollout_ref.actor.strategy == config.critic.strategy)
             from verl.workers.fsdp_workers import ActorRolloutRefWorker, CriticWorker
             from verl.single_controller.ray import RayWorkerGroup
             ray_worker_group_cls = RayWorkerGroup
 
         elif config.actor_rollout_ref.actor.strategy == 'megatron':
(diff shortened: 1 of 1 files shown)


dirtyDan0:
use_critic has already been set when RayPPOTrainer inits, and the GAE case has been taken into account.
https://github.com/volcengine/verl/blob/468adf22c43b744348051fccd7a5d830c6c3c36a/verl/trainer/ppo/ray_trainer.py#L299-L307

If you don't modify the source code and follow the default config in ppo_(megatron)_trainer.yaml, this assertion will always be true because by default.

https://github.com/volcengine/verl/blob/468adf22c43b744348051fccd7a5d830c6c3c36a/verl/trainer/config/ppo_trainer.yaml#L25-L26
https://github.com/volcengine/verl/blob/468adf22c43b744348051fccd7a5d830c6c3c36a/verl/trainer/config/ppo_trainer.yaml#L104-L105
https://github.com/volcengine/verl/blob/468adf22c43b744348051fccd7a5d830c6c3c36a/verl/trainer/config/ppo_megatron_trainer.yaml#L23-L24
https://github.com/volcengine/verl/blob/468adf22c43b744348051fccd7a5d830c6c3c36a/verl/trainer/config/ppo_megatron_trainer.yaml#L100-L101

dingyuan-shi:
This come up with the case when I try to train actor using megatron with reinforce++.
I only need to modify the actor strategy to megatron, but this assertion error emeged.
Without modifying main_ppo, I have to change a setting (ie. critic strategy) that I have never used, which is a little bit confusing.

dirtyDan0:
> This come up with the case when I try to train actor using megatron with reinforce++. I only need to modify the actor strategy to megatron, but this assertion error emeged. Without modifying main_ppo, I have to change a setting (ie. critic strategy) that I have never used, which is a little bit confusing.

python3 -m verl.trainer.main_ppo --config-path=config \
    **--config-name='ppo_megatron_trainer.yaml'**\

eric-haibin-lin:
@dingyuan-shi is your issue still there with --config-name='ppo_megatron_trainer.yaml' ?

dingyuan-shi:
> @dingyuan-shi is your issue still there with --config-name='ppo_megatron_trainer.yaml' ?

It's ok, the issue can be fixed!

PR #674, opened 2025-03-19
[misc] Add Ulysses parallel config precheck

uygnef:
Prevents training hangs by validating `num_key_value_heads % ulysses_sequence_parallel_size == 0` before training.

### DIFF
--- a/verl/models/transformers/llama.py
+++ b/verl/models/transformers/llama.py
@@ -24,7 +24,8 @@ from transformers.models.llama.modeling_llama import apply_rotary_pos_emb
 from transformers.cache_utils import Cache
 from transformers.utils import logging
 from transformers.modeling_flash_attention_utils import _flash_attention_forward
-from verl.utils.ulysses import gather_heads_scatter_seq, gather_seq_scatter_heads, get_ulysses_sequence_parallel_world_size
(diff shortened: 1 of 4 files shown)


vermouth1992:
I guess this check is unnecessary as we can always repeat kv before performing a2a. It also enables setting large ulysses_sequence_parallel_size

uygnef:
> I guess this check is unnecessary as we can always repeat kv before performing a2a. It also enables setting large ulysses_sequence_parallel_size

oh, it should be num_attention_heads rather than kv head.

vermouth1992:
Actually, not every model config may have `num_attention_heads` and this potentially leads to bug. I guess it's better to add the check to the monkey patch script per model where the config and ulysses are determined: https://github.com/volcengine/verl/blob/main/verl/models/transformers/llama.py#L31

uygnef:
@vermouth1992 Hi, I've made the changes. Could you please review them? Thanks!

vermouth1992 (approved):

vermouth1992:
Some tests failed

uygnef:
> Some tests failed

Sorry for not testing thoroughly. By the way, could you tell me how to run these test workflows before I submit my own changes?

vermouth1992 (approved):

Issue #673, opened 2025-03-19
[bug] vllm_rollout_spmd different behavior for setting max_model_len compared with vllm_rollout


twangnyc:
Hi,

I encountered the issue training with verl with vLLM complaining `Input prompt (6768 tokens) is too long and exceeds limit of 4096`. My max_prompt_length is 3072 and max_response_length is 1024. After looking into the verl code, in vllm_rollout_spmd.py, max_model_length is hard coded to be prompt_length + response length (https://github.com/volcengine/verl/blame/main/verl/workers/rollout/vllm_rollout/vllm_rollout_spmd.py#L110), and some of my data might exceed that length. However, in vllm_rollout.py file, max_model_length is a configurable number https://github.com/volcengine/verl/blame/main/verl/workers/rollout/vllm_rollout/vllm_rollout.py#L110, and increasing max_model_length in config file should be able to avoid this problem.

PR #676, opened 2025-03-19
[Bug fix] change max_model_len to be configurable in vllm_rollout_spmd

twangnyc:
Related to this issue https://github.com/volcengine/verl/issues/673 

Make vllm_rollout_spmd.py (https://github.com/volcengine/verl/blob/main/verl/workers/rollout/vllm_rollout/vllm_rollout_spmd.py#L110) behavior on max_model_len configuration to be consistent with vllm_rollout.py (https://github.com/volcengine/verl/blob/main/verl/workers/rollout/vllm_rollout/vllm_rollout.py#L110).


twangnyc:
Reopening after setting up right author email

PR #677, opened 2025-03-19
[Bug fix] change max_model_len to be configurable in vllm_rollout_spmd

twangnyc:
Related to this issue https://github.com/volcengine/verl/issues/673

Make vllm_rollout_spmd.py (https://github.com/volcengine/verl/blob/main/verl/workers/rollout/vllm_rollout/vllm_rollout_spmd.py#L110) behavior on max_model_len configuration to be consistent with vllm_rollout.py (https://github.com/volcengine/verl/blob/main/verl/workers/rollout/vllm_rollout/vllm_rollout.py#L110).


PeterSH6 (commented):
Nice catch. We should merge the two vllm files into one in the future.

twangnyc:
Checking if we would like to merge this fix?

PeterSH6 (approved):

Issue #682, opened 2025-03-20
[BUG] Megatron cannot save checkpoint


haolin-nju:
Verl expects 2 arguments and a kwargs in the save_checkpoint function of megatron_workers.py https://github.com/volcengine/verl/blob/5367156a0ea6ff154e5ea7d449cddbddcda9b2fc/verl/workers/megatron_workers.py#L449

However, in the place of calling it, verl passes 4 arguments  https://github.com/volcengine/verl/blob/5367156a0ea6ff154e5ea7d449cddbddcda9b2fc/verl/trainer/ppo/ray_trainer.py#L666-L670

It seems that global_steps is missing.
@PeterSH6 PTAL

haolin-nju:
> Hello [@haolin-nju](https://github.com/haolin-nju) , we are releasing checkpoint function in PR [#687](https://github.com/volcengine/verl/pull/687) , you can help to review and test, feel free to open issue/PR if you encounter new bugs.

Very good job! I tried quick-fix the issue, though another issue came up: variable `architecture` mismatches `architectures` in config.json. I find the PR fixes this issue too. I'll give it a try asap. 👍

haolin-nju:
> Hello [@haolin-nju](https://github.com/haolin-nju) , we are releasing checkpoint function in PR [#687](https://github.com/volcengine/verl/pull/687) , you can help to review and test, feel free to open issue/PR if you encounter new bugs.

Hi,
The PR funtions well in my environment! However, I've noticed that the load/save checkpoint logic does not utilize the `load_checkpoint` and `save_checkpoint` from Megatron-LM, which support distributed checkpoints for very large models. Is this because the Megatron model defined in verl does not inherit from `GPTModel`? I believe that delegating checkpoint responsibilities to Megatron could relieve verl from handling these complex tasks. What are your thoughts on this? Do you have any plans to support `GPTModel` in the near future?

haolin-nju:
> Thanks for sharing your concerns~
> 
> 1. Currently, our implementation works the same as Megatron's sharded checkpointing, where we split model weights and optimizer states according to model parallel rank. The checkpointing logic itself is simple, and Megatron's implementation is very complex to debug : ).
[...]

PR #634, opened 2025-03-17
[ckpt] Add Optimizer Save/Load Support and Llama 3.1 Integration for Megatron

uygnef:
This PR introduces the following changes:

## Optimizer Save/Load Support for Megatron
 Added functionality to save and load the optimizer state, enabling seamless auto-resume for Megatron training.
## Llama 3.1 Support for Megatron
 Integrated support for Llama 3.1 within Megatron. Special thanks to @ShareLer, who developed this feature.
## Minor Bug Fixes:
Addressed a few small issues.

Issue #605, opened 2025-03-15
checkpoint save error using megatron as backend


wedu-nvidia:
I am using the lastest container
FROM whatcanyousee/verl:vemlp-th2.4.0-cu124-vllm0.6.3-ray2.10-te2.0-megatron0.11.0-v0.0.5
I am trying to simple test and actually it can run successfully but suffers from error for checking point save.
this is the command:
```
python3 -m verl.trainer.main_ppo --config-path=config \
    --config-name='ppo_megatron_trainer.yaml'\
    data.train_files=/home/wedu/data/gsm8k/train.parquet \
    data.val_files=/home/wedu/data/gsm8k/test.parquet \
    data.train_batch_size=32 \
    data.max_prompt_length=512 \
    data.max_response_length=512 \
    actor_rollout_ref.model.path=/home/wedu/Qwen2.5-0.5B \
    actor_rollout_ref.actor.optim.lr=2e-6 \
    actor_rollout_ref.actor.ppo_mini_batch_size=32 \
    actor_rollout_ref.actor.ppo_micro_batch_size_per_gpu=4 \
    actor_rollout_ref.actor.megatron.tensor_model_parallel_size=2 \
    actor_rollout_ref.actor.megatron.pipeline_model_parallel_size=1 \
    actor_rollout_ref.actor.megatron.sequence_parallel=True \
    actor_rollout_ref.rollout.log_prob_micro_batch_size_per_gpu=8 \
    actor_rollout_ref.rollout.tensor_model_parallel_size=2 \
    actor_rollout_ref.rollout.name=vllm \
    actor_rollout_ref.rollout.gpu_memory_utilization=0.5 \
    actor_rollout_ref.ref.log_prob_micro_batch_size_per_gpu=16 \
    actor_rollout_ref.ref.megatron.tensor_model_parallel_size=1 \
    algorithm.kl_ctrl.kl_coef=0.001 \
    algorithm.adv_estimator=grpo \
    +actor_rollout_ref.actor.use_kl_loss=True \
    +actor_rollout_ref.rollout.n_val=1 \
    trainer.critic_warmup=0 \
    trainer.logger=['console'] \
    trainer.project_name='verl_megatron_gsm8k_examples' \
    trainer.experiment_name='deepseek_llm_1b3_function_rm' \
    trainer.n_gpus_per_node=2 \
    trainer.nnodes=1 \
    trainer.save_freq=1 \
    trainer.test_freq=1 \
    trainer.total_epochs=15 \
[...]

PR #510, opened 2025-03-07
[misc] fix typo and add llama_saver

caaatch22:

PR #687, opened 2025-03-20
[feat] Megatron checkpoint support for current Llama and Qwen models

ETOgaosion:
# Intro

Support Megatron checkpoint for Model, Optimizer States and RNG states, with a new layer of abstraction: `MegatronCheckpointManager` like FSDP. Also add checkpoint tests.

# Involved Issues and PRs

This solved issue #682 #605 , including PR #510 #634 #368 #330 . Thanks for the great efforts of @uygnef, @ShareLer and @caaatch22 in these contributions.

# TODOs

- [ ] Support Megatron dist checkpointing mechanism, now use torch.save/load to store/restore model weights.
- [x] Quick: Also store hf format model.


examples/checkpoint/run_deepseek_megatron_ckpt.sh
PeterSH6, line 1: We may not need these examples in the checkpoint directory. As similar scripts are already posted in tests

verl/trainer/config/ppo_trainer.yaml
```diff
@@ -183,7 +185,7 @@ trainer:
   n_gpus_per_node: 8
   save_freq: -1
   # auto: find the last ckpt to resume. If can't find, start from scratch
-  resume_mode: auto # or auto or resume_path if 
+  resume_mode: disable # or auto or resume_path if
```
PeterSH6: I think the default value should be `auto`. When there's no ckpt, the auto mode will just train the model from scratch.

verl/trainer/config/ppo_megatron_trainer.yaml
```diff
@@ -175,8 +183,10 @@ trainer:
   n_gpus_per_node: 8
   save_freq: -1
   # auto: find the last ckpt to resume. If can't find, start from scratch
-  resume_mode: disable # or auto or resume_path if 
+  resume_mode: disable # or auto or resume_path if
```
PeterSH6: I think the default value should be `auto`. When there's no ckpt, the auto mode will just train the model from scratch.

verl/utils/checkpoint/fsdp_checkpoint_manager.py
```diff
@@ -56,16 +57,22 @@ def __init__(self,
...
-        super().__init__(model, optimizer, lr_scheduler, processing_class)
+        super().__init__(model, checkpoint_contents=checkpoint_contents)
+        self.optimizer = optimizer
+        self.lr_scheduler = lr_scheduler
+        self.processing_class = processing_class
 
-    def load_checkpoint(self, path=None, del_local_after_load=False, *args, **kwargs):
-        if path is None:
+    def load_checkpoint(self, local_path: str, hdfs_path: str = None, del_local_after_load=False):
+        local, ckpt_path = self.checkpath(local_path, hdfs_path)
```
PeterSH6: Although in current codebase, hdfs_path is not implemented, we still need this feature in the future.
Therefore, for the ckpt_path, I think we should distinguish `local_path` and `hdfs_path`.
`ckpt_path` is a bit ambiguous.

verl/utils/checkpoint/fsdp_checkpoint_manager.py
```diff
@@ -149,11 +159,11 @@ def save_checkpoint(self, local_path: str, global_step: int, remove_previous_ckp
...
-            hf_local_path = os.path.join(local_path, 'huggingface')
+            hf_local_path = os.path.join(ckpt_path, 'huggingface')
             os.makedirs(hf_local_path, exist_ok=True)
             self.model._fsdp_wrapped_module.config.save_pretrained(hf_local_path)
             self.processing_class.save_pretrained(hf_local_path)
 
         torch.distributed.barrier()
 
-        self.previous_save_local_path = local_path
+        self.previous_save_path = ckpt_path
```
PeterSH6: Same here. Local may be necessary.

PeterSH6 (commented):
Nice job!

verl/utils/checkpoint/megatron_checkpoint_manager.py
```diff
@@ -0,0 +1,269 @@
...
+        # TODO: shall we remove previous ckpt every save?
+        if remove_previous_ckpt:
+            self.remove_previous_save_local_path()
+        if local:
+            local_path = self.local_mkdir(local_path)
+        torch.distributed.barrier()
+
+        # Save Model
+        if 'model' in self.checkpoint_contents:
+            state_dict = self.weight_saver(self.model,
```
PeterSH6, line 248: It seems that we store the whole weight. We should save the sharded weight here for better resume

verl/utils/checkpoint/checkpoint_manager.py
```diff
@@ -39,32 +38,36 @@ class BaseCheckpointManager:
...
-                                                                                             ProcessorMixin]):
+    def __init__(self, model=None, checkpoint_contents: list = ['model', 'optimizer', 'extra']):
         self.previous_global_step = None
-        self.previous_save_local_path = None
+        self.previous_saved_path = None
 
         self.model = model
-        self.optimizer = optimizer
-        self.lr_scheduler = lr_scheduler
-        self.processing_class = processing_class
```
PeterSH6, line 51: processing_class (tokenizer) should be saved

examples/checkpoint/run_deepseek_megatron_ckpt.sh
```diff
@@ -0,0 +1,91 @@
+set -x
```
vermouth1992, line 1: I guess we can turn on the checkpoint functionality to old examples to avoid adding too many examples

ETOgaosion: We default save at some frequency and auto resume?

vermouth1992: I guess we can keep the current changes

vermouth1992 (approved):

PR #688, opened 2025-03-20
Adding Openmanus-RL to the Awesome work

Kunlun-Zhu:
Adding Openmanus-RL: a llm agent rl tunning repo with verl


eric-haibin-lin (approved):
Nice!

PR #695, opened 2025-03-21
fix typo

Cppowboy:


eric-haibin-lin (approved):
while there may be other memory related issues in sglang, i think it's ok to merge this PR first and recognize the contribution from @Cppowboy

eric-haibin-lin:
@Cppowboy i guess the test failed due to transformers version, which is fixed by verl main now.

eric-haibin-lin:
Already fixed in https://github.com/volcengine/verl/pull/756. Thank you for the attempt to fix sglang! @Cppowboy

Issue #703, opened 2025-03-21
CI: not to use the **latest** version


dirtyDan0:
Hi, I noticed that some CI tests rely on the **latest** version without specifying a specific tested version of some packages. Often, after a new version is released (e.g., vLLM 0.8.0 few days ago and Transformers 4.50.0 a few hours ago), related CI tests may fail due to a lack of timely adaptation. I believe we should always specify the exact version in the tests and only add new tests after ensuring compatibility with that version.

![Image](https://github.com/user-attachments/assets/5c016c9e-5859-43d7-aa86-cc1ebd5e8596)

PR #704, opened 2025-03-21
fix: support transformers==4.50.0

dirtyDan0:
https://github.com/volcengine/verl/issues/703

### DIFF
--- a/.github/workflows/model.yml
+++ b/.github/workflows/model.yml
@@ -41,7 +41,7 @@ jobs:
       - name: Install the current repository and upgrade to latest transformers/flash_attn
         run: |
           pip3 install -e .[test]
-          pip3 install --upgrade transformers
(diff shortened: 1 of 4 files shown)


eric-haibin-lin (approved):
Would you mind also sharing the original error message you got with latest transformers?

vermouth1992:
Instead of pinning the version, I guess we should fix what transformers 4.50 breaks.

dirtyDan0:
> Would you mind also sharing the original error message you got with latest transformers?

https://github.com/volcengine/verl/actions/runs/13983522247/job/39191777347

dirtyDan0:
> Instead of pinning the version, I guess we should fix what transformers 4.50 breaks.

updated👌I found apply_monkey_patch_to_{qwen2/llama} didn't take effect. So was ulysses.

vermouth1992 (approved):

PR #729, opened 2025-03-24
[feat] add wake_up level to reduce vllm GPU memory usage peak

yqyao:
# Reduce GPU Memory Peak Usage with wake_up Levels
This PR introduces the wake-up function in vLLM to reduce peak GPU memory usage during model initialization, especially for large models like Qwen2.5-7B.

**Key Features**
* Parameters:
level=1: Load both model weights and KV cache in a staged manner. (default version)
level=2: Load only model weights gradually.
level=3: Load only the KV cache gradually.
* Helps delay and distribute GPU memory allocation during startup to prevent OOM issues.
* Tested with GRPO on 8 and 32 GPUs: peak GPU memory usage reduced by ~10 GB for Qwen2.5-7B, enabling inference on more constrained hardware.

VLLM code https://github.com/yqyao/vllm/tree/wake_up_v0.8.1

### DIFF
--- a/verl/workers/sharding_manager/fsdp_vllm.py
+++ b/verl/workers/sharding_manager/fsdp_vllm.py
@@ -91,12 +91,23 @@ class FSDPVLLMShardingManager(BaseShardingManager):
         if vllm_version in ('0.4.2', '0.5.4', '0.6.3'):
             self.inference_engine.sync_model_weights(params, load_format=load_format)
         else:
-            self.inference_engine.wake_up()
+            if os.environ.get("WAKE_UP_VLLM", "false") == 'true':
+                # to reduce GPU memory usage peak
+                # level=1, mean load weights and kv cache, level=2 mean load weights, level=3 mean load kv cache
+                self.inference_engine.wake_up(level=2)
+            else:
+                self.inference_engine.wake_up()
+
             world_size = torch.distributed.get_world_size()
             model = self.inference_engine.llm_engine.model_executor.driver_worker.worker.model_runner.model
             loaded_params = model.load_weights(
                 ((name, param.full_tensor() if world_size != 1 else param) for name, param in params.items()))
             logger.info(f"vLLM load wegiths, loaded_params: {len(loaded_params)}")
+            if os.environ.get("WAKE_UP_VLLM", "false") == 'true':
+                del params
+                torch.cuda.empty_cache()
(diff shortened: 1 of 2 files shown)


verl/workers/sharding_manager/fsdp_vllm.py
```diff
@@ -91,12 +91,23 @@ def __enter__(self):
         if vllm_version in ('0.4.2', '0.5.4', '0.6.3'):
             self.inference_engine.sync_model_weights(params, load_format=load_format)
         else:
-            self.inference_engine.wake_up()
+            if os.environ.get("WAKE_UP_VLLM", "false") == 'true':
+                # to reduce GPU memory usage peak
+                # level=1, mean load weights and kv cache, level=2 mean load weights, level=3 mean load kv cache
+                self.inference_engine.wake_up(level=2)
```
wuxibin89, line 97: Should wake_up level be same with sleep level?
https://github.com/volcengine/verl/blob/main/verl/workers/rollout/vllm_rollout/vllm_rollout_spmd.py#L118

hiyouga:
Will this feature be merged into vllm's main branch? What is the relation between this feature and https://github.com/vllm-project/vllm/issues/15254 ?

verl/workers/sharding_manager/fsdp_vllm.py
yqyao: > Should wake_up level be same with sleep level? https://github.com/volcengine/verl/blob/main/verl/workers/rollout/vllm_rollout/vllm_rollout_spmd.py#L118

Theoretically, this is the best approach, but it seems that vLLM currently only supports `sleep` at level 1 (weights) and level 2 (weights + kv_cache). We may need to wait for official support from vLLM. For now, I’ve provided a temporary version on my end.

yqyao:
> Will this feature be merged into vllm's main branch? What is the relation between this feature and [vllm-project/vllm#15254](https://github.com/vllm-project/vllm/issues/15254) ?

It seems to be the same question. We may need to wait for official support from vLLM. For now, I’ve provided a temporary version.

hiyouga:
Got it. I am happy to see this feature in verl if vllm's officially supports it.

hiyouga:
Fixed by https://github.com/volcengine/verl/pull/987 , thanks for your initial idea and contribution

PR #357, opened 2025-02-24
feat: add support for ulysses sequence parallel for transformers >= 0.48

xffxff:
close #312 

Add support for ulysses sp for transformers >= 0.48

I've tested transformers 0.45.0, 0.46.0, 0.47.0, 0.48.0 and 0.49.0, using sp=2 with the following script in my local env
```bash
#!/bin/bash

set -ex
VERSIONS=("4.45.0" "4.46.0" "4.47.0" "4.48.0" "4.49.0")

for version in "${VERSIONS[@]}"; do
    echo "Testing with Transformers version ${version}"
    echo "----------------------------------------"
    
    pip install "transformers==${version}"
    
    PYTHONPATH=./ torchrun --nproc_per_node=2 tests/model/test_transformers_ulysses.py
    
    echo "----------------------------------------"
    echo "Completed testing for version ${version}"
    echo ""
done
```

PR #735, opened 2025-03-24
refactor: unify ulysses flash attention patch to avoid single model patches

wuxibin89:
**This is an effort to unify transformers monkey patch to support ulyssess sequence parallellism for more models.**

### Basic idea
In transformer architecture, all operations except attention are token-wise, include RoPE, LayerNorm, MLP, etc, so we just need to patch the attention function. 
For now, ulyssess sequence relies on sequence packing and flash attention,  and transformers widely use `_flash_attention_forward` in each model's Attention module, e.g LlamaAttention, Qwen2Attention. So we just need to add 2 all-to-all operations before and after `_flash_attention_forward`.
image

- We introduce an additional all_gather in each layer for position_ids because `prepare_fa2_from_position_ids` needs it. The all_gather communication cost is `O(nnz)`, which should be negligible compare to QKV, meanwhile we also reduce RoPE computation to 1/sp_size of the original.

### Correctness Verification
[run_qwen2-7b_seq_balance.sh](https://github.com/volcengine/verl/blob/main/examples/ppo_trainer/run_qwen2-7b_seq_balance.sh) with `ulysses_sequence_parallel_size=2`
- red(baseline): main branch transformers==4.47.1
- purple: dev branch transformers==4.47.1
- green: dev branch transformers==4.49.0
image

By unifying monkey patch, we can avoid individual model patches and achieve better forward compatibility with transformers, avoid issue like #357 #704.

Also remove `check_model_support_rmpad` since we enforce `attn_implementation="flash_attention_2"`, every model which supports FlashAttention2 should support sequence packing.

- [x] unify LLM model patch
- [ ] clean llama/qwen attention patch
- [ ] support qwen2vl ulyssess sp
- [ ] unify VLM model patch with LLM model

### DIFF
--- a/tests/model/test_transformers_ulysses.py
+++ b/tests/model/test_transformers_ulysses.py
@@ -25,13 +25,13 @@ from verl.protocol import DataProto
 from flash_attn.bert_padding import unpad_input, index_first_axis, rearrange
 from transformers import LlamaConfig, Qwen2Config
 from transformers import AutoModelForCausalLM
-from verl.models.transformers.monkey_patch import apply_monkey_patch_to_llama, apply_monkey_patch_to_qwen2
+from verl.models.transformers.monkey_patch import apply_monkey_patch
 
 # TODO(sgm): add more models for test
 # we only need one scale for each model
 test_configs = {
-    'llama': (LlamaConfig(num_hidden_layers=2), apply_monkey_patch_to_llama),
-    'qwen2': (Qwen2Config(num_hidden_layers=2), apply_monkey_patch_to_qwen2)
+    'llama': (LlamaConfig(num_hidden_layers=2), apply_monkey_patch),
+    'qwen2': (Qwen2Config(num_hidden_layers=2), apply_monkey_patch)
 }
 
 
@@ -57,11 +57,11 @@ def test_hf_casual_fwd():
 
     for model_name, (config, apply_monkey_patch) in test_configs.items():
         # patch before load
-        apply_monkey_patch()
         with torch.device('cuda'):
             model = AutoModelForCausalLM.from_config(config=config,
                                                      torch_dtype=torch.bfloat16,
                                                      attn_implementation='flash_attention_2')
+            apply_monkey_patch(model)
             model = model.to(device='cuda')
             sync_model_parameters_global(model)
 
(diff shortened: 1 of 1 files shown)


hiyouga:
```
__________ ERROR collecting tests/model/test_transformers_ulysses.py ___________
ImportError while importing test module '/data00/tiger/huggingface/verl/verl/tests/model/test_transformers_ulysses.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
/root/miniconda3/lib/python3.10/importlib/__init__.py:126: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
tests/model/test_transformers_ulysses.py:28: in <module>
    from verl.models.transformers.monkey_patch import apply_monkey_patch_to_llama, apply_monkey_patch_to_qwen2
E   ImportError: cannot import name 'apply_monkey_patch_to_llama' from 'verl.models.transformers.monkey_patch' (/data00/tiger/huggingface/verl/verl/verl/models/transformers/monkey_patch.py)
```

verl/models/registry.py
```diff
@@ -17,28 +17,6 @@
...
-
-
-def check_model_support_rmpad(model_type: str):
-    assert isinstance(model_type, str)
-    if not model_type in _MODELS_SUPPORT_RMPAD:
-        raise ValueError(f"Model architecture {model_type} is not supported for now. "
-                         f"RMPad supported architectures: {_MODELS_SUPPORT_RMPAD}."
-                         f"Please set `use_remove_padding=False` in the model config.")
-
-    if model_type in ("qwen2_vl", "qwen2_5_vl"):  # patch remove padding for qwen2vl mrope
```
hiyouga, line 32: These patches are essential to the rmpad functionality of qwen2vl models, we should enable them even if ulysses is not enanbled

hiyouga (changes requested):

verl/models/registry.py
wuxibin89: Yes, I move it to `apply_monkey_patch`, which is called when use_remove_padding or ulysses enabled

hiyouga (approved):

tongyx361:
I tested on the main branch (https://github.com/volcengine/verl/commit/797f9994b73f6335aa2349cdb2e9bb2a42f13f5c) but got stuck with `transformers==4.49.0` but went well with `4.47.1` using [run_qwen2-7b_seq_balance.sh](https://github.com/volcengine/verl/blob/main/examples/ppo_trainer/run_qwen2-7b_seq_balance.sh) with `ulysses_sequence_parallel_size=2`. Could you try reproducing this problem?

vermouth1992:
Does CI test with sp=2？

PR #743, opened 2025-03-25
Support InternLM3 in Verl

guox18:
vLLM (version 0.8.1) now provides native support for InternLM3. With a minor and safe modification to model/register.py, Verl can also be made compatible with InternLM3. I have successfully trained my InternLM3 model using Verl.

### DIFF
--- a/verl/models/registry.py
+++ b/verl/models/registry.py
@@ -19,7 +19,7 @@ import torch.nn as nn
 
 # Supported models using HF Rmpad
 # TODO(sgm): HF may supported more than listed here, we should add more after testing
-_MODELS_SUPPORT_RMPAD = {'llama', 'mistral', 'gemma', 'qwen2', 'qwen2_vl', 'qwen2_5_vl'}
+_MODELS_SUPPORT_RMPAD = {'llama', 'mistral', 'gemma', 'qwen2', 'qwen2_vl', 'qwen2_5_vl', 'internlm3'}
 
 
 def check_model_support_rmpad(model_type: str):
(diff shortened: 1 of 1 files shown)


eric-haibin-lin:
Thanks! We are looking for more reference model training baselines. If your dataset is open resource, would you be willing to add new results in this page? https://verl.readthedocs.io/en/latest/experiment/ppo.html  (in docs/experiment/ppo.rst)

verl/models/registry.py
```diff
@@ -19,7 +19,7 @@
 
 # Supported models using HF Rmpad
 # TODO(sgm): HF may supported more than listed here, we should add more after testing
-_MODELS_SUPPORT_RMPAD = {'llama', 'mistral', 'gemma', 'qwen2', 'qwen2_vl', 'qwen2_5_vl'}
+_MODELS_SUPPORT_RMPAD = {'llama', 'mistral', 'gemma', 'qwen2', 'qwen2_vl', 'qwen2_5_vl', 'internlm3'}
```
wuxibin89, line 22: This is not needed any more, any model with `_supports_flash_attn_2=True` should support sequence packing.

verl/models/registry.py
```diff
@@ -41,13 +41,16 @@ def check_model_support_rmpad(model_type: str):
...
 _MODELS = {
     "LlamaForCausalLM":
         ("llama", ("ParallelLlamaForCausalLMRmPadPP", "ParallelLlamaForValueRmPadPP", "ParallelLlamaForCausalLMRmPad")),
     "Qwen2ForCausalLM":
         ("qwen2", ("ParallelQwen2ForCausalLMRmPadPP", "ParallelQwen2ForValueRmPadPP", "ParallelQwen2ForCausalLMRmPad")),
     "MistralForCausalLM": ("mistral", ("ParallelMistralForCausalLMRmPadPP", "ParallelMistralForValueRmPadPP",
-                                       "ParallelMistralForCausalLMRmPad"))
+                                       "ParallelMistralForCausalLMRmPad")),
+    "InternLM3ForCausalLM":
+        ("internlm3", ("ParallelLlamaForCausalLMRmPadPP", "ParallelLlamaForValueRmPadPP", "ParallelLlamaForCausalLMRmPad")),
```
wuxibin89, line 53: Does InternLM3 use same architecture as LLama?

verl/models/registry.py
guox18: Does that mean,  all models supported by vllm be compatible with verl in the future?

verl/models/registry.py
guox18: yes, I remember I saw this in a huggingface transformer update

guox18:
> Thanks! We are looking for more reference model training baselines. If your dataset is open resource, would you be willing to add new results in this page? https://verl.readthedocs.io/en/latest/experiment/ppo.html (in docs/experiment/ppo.rst)

Will internlm3 be supported in verl? if so , i can train an internlm3 model on gsm8k and post my results in ppo.rst !

PR #745, opened 2025-03-25
Wedu/fix log

wedu-nvidia:
I don't think we need to accumulate the log data.
Instead, we should log data at each step. I've noticed that accumulating logs makes the data files larger, and Weights & Biases has limitations on table size.


eric-haibin-lin (commented):
thanks for creating the PR. The change log seems to include multiple changes. Could you clean it up and state the main goal of this PR?

eric-haibin-lin:
Closing due to inactivity

PR #749, opened 2025-03-25
fix: set torch dtype to auto

ccclyu:
### Issue

As transformers `from_pretrained()`  automatically load ckpt in the default torch.dtype (here, fp32) and not follow the  dtype of the weights themselve, it often causes high CPU memory usage even OOM, which crushes the training. 

### Fix 
Follow by  https://github.com/huggingface/transformers/issues/34743 and tested on loading Llama-3.1-405B-Instruct. 

```
- model = AutoModelForCausalLM.from_pretrained(model_file)
+ model = AutoModelForCausalLM.from_pretrained(model_file, torch_dtype="auto")
```

### Potential problem:

Some CPUs can not support `bfloat16` and it might take some time to convert back to `float32`.

### DIFF
--- a/verl/utils/checkpoint/megatron_checkpoint_manager.py
+++ b/verl/utils/checkpoint/megatron_checkpoint_manager.py
@@ -264,7 +264,7 @@ class MegatronCheckpointManager(BaseCheckpointManager):
                         state_dict['score.weight'] = state_dict['score.weight']
                     else:
                         from transformers import AutoModelForCausalLM
-                        model = AutoModelForCausalLM.from_pretrained(self.config.model.path)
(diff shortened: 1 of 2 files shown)


vermouth1992:
Setting dtype to auto is not desirable. The reason we force dtype to be fp32 is that this will makes the optimizer states to be fp32. This is crucial for convergence. In fact, the computation is still performed in bf16 when autocast is enabled.

ccclyu:
Yes. I tested on the megatron as backend and it needs to load the checkpoints to cpu regardless of training framework. https://github.com/volcengine/verl/blob/main/verl/utils/model.py#L319. Please correct me if I am wrong.

vermouth1992 (approved):

PR #753, opened 2025-03-25
PRIME baseline

ZefanW:
add example PRIME script and wandb log to doc


eric-haibin-lin (approved):
great! do you plan to add other datasets?

ZefanW:
> great! do you plan to add other datasets?

Yes! We didn’t include additional datasets in this baseline because the training and validation sets should share a similar range and format for convenient baseline runs. Yet, datasets like Minerva Math and Olympiads only offer test splits. 
Fortunately, we now have Eurus-2-RL-Data and DAPO-Math-17k training sets for AIME2024 validation. We will add this baseline soon.

PR #631, opened 2025-03-17
feat: add dynamic sampler for curriculum learning in RL

zpqiu:
# Dynamic Sampler for Curriculum Learning in RL

> [!NOTE]
> I have completed the main code development and validation experiments for the sampler. I haven't yet added the calling logic to RayPPOTrainer or included the corresponding configuration items. If the maintainers believe this feature is useful and suitable for integration into verl, I will continue to refine this PR. 

## Overview

This PR introduces a `DynamicSampler` class that implements curriculum learning based on sample difficulty. The sampler dynamically adjusts the difficulty distribution of training data based on model performance.

Key features:
- Stratified sampling based on sample difficulty levels
- Automatic difficulty adjustment based on model accuracy
- Support for mixed sampling (from current and harder difficulty levels)
- Compatible with StatefulDataLoader for checkpoint and resume

## Implementation Details

`DynamicSampler` extends `torch.utils.data.Sampler` with the following capabilities:
1. Grouping dataset samples by difficulty level
2. Sampling based on current difficulty level and mixing ratio
3. Dynamically adjusting difficulty level based on model performance
4. Supporting state saving and loading for training resumption

## Integration with RLHF Training

In `ray_trainer.py`, we can integrate `DynamicSampler` by:
1. In the `_create_dataloader` method, deciding whether to use `DynamicSampler` based on configuration
2. Calling `update_sampling_policy` at the end of each training step to update the sampling strategy

```python
# ... existing code ...
def _create_dataloader(self):
    # ... existing code ...
    
    # Use dynamic sampler for curriculum learning
    if self.config.data.get('use_dynamic_sampler', False) and self.train_dataset.difficulties is not None:
        sampler = DynamicSampler(
            data_source=np.arange(len(self.train_dataset)),
            difficulties=self.train_dataset.difficulties,
            total_steps=len(self.train_dataset) * self.config.trainer.total_epochs // self.config.data.train_batch_size,
            batch_size=self.config.data.train_batch_size,
            init_level=self.config.data.get('init_level', 0),
            mix_harder_samples=self.config.data.get('mix_harder_samples', 0.2),
            smooth_window=self.config.data.get('smooth_window', 3),
            threshold_for_next_level=self.config.data.get('threshold_for_next_level', 0.75),
            threshold_for_prev_level=self.config.data.get('threshold_for_prev_level', 0.3),
            seed=self.config.data.get('seed', 42)
        )
    else:
        # Use existing random or sequential sampler
        if self.config.data.shuffle:
[...]

PR #759, opened 2025-03-25
feat: add dynamic sampler for curriculum learning in RL

zpqiu:
Re-submit #631 

# Dynamic Sampler for Curriculum Learning in RL

## Overview

This PR introduces a `DynamicSampler` class that implements curriculum learning based on sample difficulty. The sampler dynamically adjusts the difficulty distribution of training data based on model performance.

## Implementation Details

`DynamicSampler` extends `torch.utils.data.Sampler` with the following capabilities:
1. Grouping dataset samples by difficulty level
2. Sampling based on current difficulty level and mixing ratio
3. Dynamically adjusting difficulty level based on model performance
4. Supporting state saving and loading for training resumption

## Usage

1. Build the dataset with difficulty level

```bash
python3 examples/data_preprocess/orz_with_difficulty_dataset.py
```

2. Train

```bash
bash examples/grpo_trainer/run_qwen2-7b_orz_curriculum_learning.sh
```

The training curves of the example dataset and training script will look like:
截屏2025-03-27 21 56 06

### DIFF
--- a/verl/trainer/ppo/ray_trainer.py
+++ b/verl/trainer/ppo/ray_trainer.py
@@ -412,7 +412,8 @@ class RayPPOTrainer(object):
             if self.config.trainer.total_training_steps is not None:
                 total_training_steps = self.config.trainer.total_training_steps
             else:
-                total_training_steps = (len(self.train_dataset) // self.config.data.train_batch_size) * self.config.trainer.total_epochs
+                total_training_steps = (len(self.train_dataset) //
+                                        self.config.data.train_batch_size) * self.config.trainer.total_epochs
             sampler = DynamicSampler(data_source=self.train_dataset,
                                      difficulties=self.train_dataset.difficulties,
                                      total_steps=total_training_steps,
--- a/examples/data_preprocess/orz_with_difficulty_dataset.py
+++ b/examples/data_preprocess/orz_with_difficulty_dataset.py
@@ -20,12 +20,13 @@ import datasets
 from verl.utils.hdfs_io import copy, makedirs
 import argparse
 
-
-PROMPT_TEMPLATE = ("Solve the following math problem step by step. "
-                    "The reasoning process and answer are enclosed within <think> </think> and <answer> </answer> tags, respectively, "
-                    "i.e., <think> reasoning process here </think> <answer> answer here </answer>. \n\n"
-                    "{question}\n\n"
-                    "Remember to put your answer inside <answer> </answer> tags, and your final answer will be extracted automatically by the \\boxed{{}} tag.")
+PROMPT_TEMPLATE = (
+    "Solve the following math problem step by step. "
+    "The reasoning process and answer are enclosed within <think> </think> and <answer> </answer> tags, respectively, "
+    "i.e., <think> reasoning process here </think> <answer> answer here </answer>. \n\n"
+    "{question}\n\n"
+    "Remember to put your answer inside <answer> </answer> tags, and your final answer will be extracted automatically by the \\boxed{{}} tag."
+)
 
 if __name__ == '__main__':
     parser = argparse.ArgumentParser()
@@ -53,12 +54,12 @@ if __name__ == '__main__':
             question = PROMPT_TEMPLATE.format(question=question)
             solution = example.pop('answer')
             data = {
-                "data_source": "lighteval/MATH", # reuse the existing reward function
+                "data_source": "lighteval/MATH",  # reuse the existing reward function
                 "prompt": [{
                     "role": "user",
                     "content": question
                 }],
-                "difficulty": max(min(int((1.0-example.pop('pass_at_n')) * 10) + 1, 10), 1),
+                "difficulty": max(min(int((1.0 - example.pop('pass_at_n')) * 10) + 1, 10), 1),
                 "ability": "math",
                 "reward_model": {
                     "style": "rule",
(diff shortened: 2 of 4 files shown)


eric-haibin-lin:
Thanks! Do you have some reference performance result of the implemented learning method?

zpqiu:
Yes, but I haven't performed a complete evaluation on the example code provided above - it's just a sample. 

The experimental results below require customized prompts and reward functions.

- Model：Qwen2.5-7B
- Data：pe-nlp/ORZ-MATH-57k-Filter-difficulty-decontaminated
- Hyperparameters
    ```markdown
    data.train_batch_size=1024 \
    data.val_batch_size=512 \
    data.max_prompt_length=1024 \
    data.max_response_length=8192 \
    data.shuffle=True \
    actor_rollout_ref.model.path=Qwen/Qwen2.5-7B \
    actor_rollout_ref.actor.optim.lr=3e-6 \
    actor_rollout_ref.actor.shuffle=True \
    actor_rollout_ref.actor.ppo_mini_batch_size=4096 \
    actor_rollout_ref.actor.kl_loss_coef=0.001 \
    actor_rollout_ref.rollout.temperature=1.0 \
    actor_rollout_ref.rollout.n=8 \
    algorithm.kl_ctrl.kl_coef=0.0
    ```
- Max-steps: 200
- Evaluation
    - temprature=0.6
    - Metric: Pass@1 [8]

  |  | AIME 2024 | MATH 500 | GPQA |
  | --- | --- | --- | --- |
  | random sampler | 16.25 | 79.37 | 35.75 |
  | dynamic sampler | 14.58 | 82.13 | 40.12 |

### Other settings

- Prompt
  ```markdown
  (
      "A conversation between User and Assistant. The User asks a question, and the Assistant solves it. "
      "The Assistant first thinks about the reasoning process in the mind and then provides the User with the answer. "
      "The reasoning process and answer are enclosed within <think> </think> and <answer> </answer> tags, respectively, i.e., <think> reasoning process here </think> <answer> answer here </answer>. "
      "User: You must put your answer inside <answer> </answer> tags, i.e., <answer> answer here </answer>. "
      "{prompt}\n"
      "Assistant: <think>"
  )
  ```
- [Reward function](https://github.com/zpqiu/verl/blob/cl/verl/utils/reward_score/r1.py)

mertunsall:
I think it would be nice to let people write their own sampling strategy as well, just like the custom reward function. This PR can create a nice interface and give people a class structure that they can replicate.

verl/trainer/ppo/ray_trainer.py
```diff
@@ -406,16 +407,32 @@ def _create_dataloader(self):
             'truncation', 'error'
         ), f'dataset truncation {self.train_dataset.truncation} must be the same as config {self.config.data.get("truncation", "error")}'
         # use sampler for better ckpt resume
-        if self.config.data.shuffle:
+        if self.config.data.get('use_dynamic_sampler', False):
```
eric-haibin-lin, line 410: do you think it's possible to rewrite the interface similar to https://github.com/volcengine/verl/pull/832/files, where we can load a custom sampler class, with 2 arugments "data_source" and "data.config", so users have maximum flexibility in extending their own sampler?

verl/trainer/ppo/ray_trainer.py
```diff
@@ -406,16 +407,32 @@ def _create_dataloader(self):
...
             train_dataloader_generator = torch.Generator()
             train_dataloader_generator.manual_seed(self.config.data.get('seed', 1))
             sampler = RandomSampler(data_source=self.train_dataset, generator=train_dataloader_generator)
         else:
             sampler = SequentialSampler(data_source=self.train_dataset)
 
+        num_workers = 8
+        if self.config.data.get('use_dynamic_sampler', False):
+            # disable multiprocessing for dynamic sampler
+            num_workers = 0
```
eric-haibin-lin, line 432: why does it not supporting multi-processing?

eric-haibin-lin: i guess num_workers can be a field in data config as well, default to 8

verl/trainer/ppo/ray_trainer.py
```diff
@@ -943,5 +964,8 @@ def fit(self):
                     progress_bar.close()
                     return
 
+                if self.config.data.get('use_dynamic_sampler', False):
+                    self.train_dataloader.sampler.update_sampling_policy(metrics["critic/score/mean"])
```
eric-haibin-lin, line 968: for future extensions, i am not sure what other sampling strategies would take as the policy update arguments. so it's kind of hard to design the interface. any thoughts? 
If it's hard to decide for now, i'd suggest we move the implementation as a custom trainer, like prime trainer, in the example folder, for experimental usage, instead of adding it to the common ray trainer. what do you think?

verl/trainer/ppo/ray_trainer.py
zpqiu: Multiprocessing will initiate prefetching, which prevents the DynamicSampler from adjusting sampling strategies in a timely manner based on latest metrics such as accuracy

verl/trainer/ppo/ray_trainer.py
zpqiu: Thanks for your suggestion. Do you mean creating a new folder to consolidate all the code in this PR (including dataset, sampler, and trainer) similar to `recipe/prime`? I think this is reasonable, since its effectiveness still needs validation on a larger scale.

If so, do I still need to look into how to load a custom sampler class as an external file? I understand that this is probably not necessary for now.

maksimstw:
We now have some experiments supporting that curriculum learning is highly effective. #1020 
https://arxiv.org/abs/2504.05520

eric-haibin-lin:
could you review this RFC for new recipes? Thanks https://github.com/volcengine/verl/issues/1354

PR #769, opened 2025-03-26
doc: upgrade to vllm 0.8.2

hiyouga:


vermouth1992:
Could you help upgrade all the CI to use vllm==0.8.2 and make setup.py uses 0.8.2 as well? Thanks.

docker/Dockerfile.ngc.vllm0.8
```diff
@@ -42,8 +42,8 @@ RUN pip config set global.index-url "${PIP_INDEX}" && \
     pip config set global.extra-index-url "${PIP_INDEX}" && \
     python -m pip install --upgrade pip
 
-# Install torch-2.6.0 + vllm-0.8.1
-RUN pip install --no-cache-dir vllm==0.8.1 torch==2.6.0 torchvision==0.21.0 torchaudio==2.6.0 tensordict torchdata \
+# Install torch-2.6.0 + vllm-0.8.2
+RUN pip install --no-cache-dir vllm==0.8.2 torch==2.6.0 torchvision==0.21.0 torchaudio==2.6.0 tensordict torchdata \
     transformers>=4.49.0 accelerate datasets peft hf-transfer \
     ray codetiming hydra-core pandas pyarrow>=15.0.0 pylatexenc qwen-vl-utils wandb dill pybind11 liger-kernel mathruler \
```
eric-haibin-lin: could u also change it to `ray[default]`, and flash-attn==2.7.4.post1 below?

eric-haibin-lin: i guess we can build an image and then update all images used in CI

hiyouga:
Let me build a 0.8.2 docker tonight

.github/workflows/sanity.yml
```diff
@@ -52,7 +52,3 @@ jobs:
       - name: Run license test
         run: |
           python3 tests/sanity/check_license.py --directory .
-      - name: Run dependency test
```
hiyouga, line 55: remove this ut because we directly use the dependencies of setup.py in the pyproject toml

Issue #708, opened 2025-03-21
[Project] deepseek R1 infrastructure


eric-haibin-lin:
The verl team like to invite the community to help contribute to build infrastructure to efficiently scale up to models of deepseek scale and develop recipes to reproduce deepseek r1 results for the broader open source and research community. You're encouraged to join the slack channel for discussions on sub-topics.

# Evals

Tasks:
- add evaluation script to reproduce ds-r1 results on several benchmarks from ds-r1 checkpoint
  - [ ] gpqa diamond (english)
  - [ ] LiveCodeBench (code)
  - [ ] SWE-bench Verified (code)
  - [ ] CNMO 2024 (math)

Notes:
- refer to `examples/data_preprocess` for data preprocessing examples
- refer to trainer/main_generation.py for generation (but evaluation code is missing)
- if possible, upload preprocessed dataset to huggingface
- start with ds distilled models for quick verification before scaling up to 671b

# Training engine 
Tasks: 
- verify `GPTModel` with mcore v0.11. There’s experimental integration of critic and actor models with GPTModel class
  - [ ] verify checkpoint manager with GPTModel 
  - [ ] Investigate the issue of convergence issue of seq packing when micro bsz > 1 @GAIR-NLP 
Mcore ds-v3 perf tuning 
  - [ ] Run ds-v3 with mcore on a range of seqlen and GPU count, tune nd parallel / memory configs 
- Mcore ds-v3 convergence verification: to ensure the correctness of mcore ds-v3 implementation, we shall run a smaller version of ds with same model architecture. 
  - [ ] obtain a ds-v3-medium pretrained ckpt with fewer params 
  - [ ] Compare mcore convergence with FSDP + ds-v3-medium baseline
  - [ ] Optimize ds-v3 long context throughout with FSDP backend for faster FSDP experiment iterations (e.g. liger alignment loss, recompute strategy, ring attn). 

Notes:
- To make the perf tuning more faithful, it’s better to develop a benchmark script that include alignment loss (CE loss, CE and entropy loss) 
- GAIR NLP team is looking GPTModel with sequence packing support when micro_bsz > 1

# data & recipe 

Tasks:
Math and scientific sources:  
- Further improve the dataset used for math RL, maybe based on DAPO 17k and other open source ones
Code:
- Build a baseline code recipe in verl main repo, using small models such as llama or qwen-7b
- Curate dataset for code RL training (start with existing open source ones) 

Notes:
[...]

PR #777, opened 2025-03-26
Support R1 Benchmark Evaluation

yyDing1:
https://github.com/volcengine/verl/issues/708

Support Evaluaton:
- [x] GPQA Diamond (english)
- [x] LiveCodeBench (code)
- [x] AIME 2024 (math)
- [x] CNMO 2024 (math)

Test
- [x] DS-R1-Distill-Qwen2.5-1.5B
- [x] DS-R1

---

Example Eval Scripts in `recipes/r1/run_r1_distill_qwen.sh`

---

Eval Results of DS-R1-Distill-Qwen2.5-1.5B (k=8)

Dataset | Test Results | Reported
-- | -- | --
GPQA Diamond | 35.3 | 33.8
LiveCodeBench | 16.9 | 16.9
AIME 2024 | 30.4 | 28.9
CNMO 2024 (en) | 45.1 | -
CNMO 2024 (zh) | 41.0 | -

---

Eval Results (DS-R1)

Dataset | Test Results (k=1) | Test Results (k=4) | Reported
-- | -- | -- | --
GPQA Diamond | 67.7 | 69.6 | 71.5
LiveCodeBench | 64.7 | 63.1 | 65.9
AIME 2024 | 86.7 | 79.2 | 79.8
CNMO 2024 | 75.0 | 78.5 | 78.8

The final eval results will be placed [here](https://huggingface.co/datasets/dyyyyyyyy/r1-benchmark-eval).


examples/data_preprocess/r1_bench.py
```diff
@@ -0,0 +1,210 @@
+# Copyright 2024 Bytedance Ltd. and/or its affiliates
```
eric-haibin-lin: Thanks for your great contribution! 

we suggest we start a new directory in `recipe/r1/`, and then add custom data_processing script, eval script in it. Ideally main_eval should also allow customization of reward_fn like https://github.com/volcengine/verl/blob/main/verl/trainer/main_ppo.py#L24, so that livecodebench and gpqa.py can also be moved to the recipe folder, and one can pass the corresponding function name in examples/r1_evaluation/run_r1_distill_qwen.sh. Would you be willing to do a small restructure of your PR? Thanks!

yyDing1: Got it, I'll make these updates soon.

yyDing1:
@eric-haibin-lin I've restructured the code and updated the DS-R1 results. I noticed that the variance between the k=1 and k=4 results is still quite large. Perhaps increasing k further could stabilize it (k=64 in R1 Paper), but I can't do this due to limited budget for API calls.

Dataset (Questions with Generated Responses) [here](https://huggingface.co/datasets/dyyyyyyyy/r1-benchmark-eval).

eric-haibin-lin (dismissed):
Thank you, great work!

eric-haibin-lin:
It's best to create some tests such that the scripts won't be broken in the future. shall we add a test for a 1b model on the smallest dataset, similar to https://github.com/volcengine/verl/blob/main/.github/workflows/e2e_gsm8k_prime.yml for eval?

yyDing1:
Thank you for the suggestion! I've now added evaluation support for AIME2024 (30 questions), which has been successfully tested on my self-hosted runner.

eric-haibin-lin (approved):
LGTM, pending test result

verl/trainer/config/evaluation.yaml
```diff
@@ -3,4 +3,8 @@ data:
   prompt_key: prompt
   response_key: responses
   data_source_key: data_source
-  reward_model_key: reward_model
\ No newline at end of file
+  reward_model_key: reward_model
+
+custom_reward_function:
```
eric-haibin-lin, line 8: if the test needs more code updates, would u mind updating https://github.com/volcengine/verl/blob/main/docs/examples/config.rst as well? 
if the test passes, i guess we can merge this PR for now and submit a separate PR for docs (docs do not trigger CI tests)

PR #780, opened 2025-03-26
Megatron checkpoint default not save hf_models, and provide model merge tool.

ETOgaosion:
Because CI is too slow, combine the features and functions of checkpoint here in 1 PR.

# Add Layer idx to decode layers

But it seems to be hard to attach a "correct" layer number to each layer, now verl implemented megatron each pp and vpp rank's layers start from index 0, leading to some inconvenience for merging tool.

The difficulty mainly comes from `torch.nn.ModuleList` implementation, [it suggests and forces to directly use index rather than custom layer number](https://github.com/pytorch/pytorch/blob/8a40fca9a130e9c4032d6f53c746d21e1124c536/torch/nn/modules/container.py#L302C5-L324C66).

Current solution is that we modify the layer number to actual number starts from pp and vpp offset when saving megatron checkpoint, and recover when loading. When use merging tool, there is no need for extra scans.

# Huggingface Model loader logic simplified

Since every rank can have access to state_dict, there is actually no need to broadcast the weights among mp and dp groups at all, and all from rank 0. The implementation before is too costly and may cause OOM issue because each rank can take up whole model space in GPU.

And the loader logic is not straight-forward, since everyone only need to load its vpp_size number of layers, why iterate over whole num_layers.

So current solution is every rank load itself's sharded weights from `state_dict`.

But this requires users having storage nodes available to connect with every calculation nodes. For those who can only use rank 0 to store huggingface model, we move original implementation to deperacated besides new version of file.

# Modify test scripts to reuse downloaded huggingface model

Avoid errors when connecting with huggingface to access metadata.

# Modify CI workflows to enable load-balance of CI machines

Currently L20-0 takes up 6 more jobs than L20-1, try reduce the pipeline bubble of each task.


vermouth1992:
Agreed. We should modify the layer number to its actual layer number irrespective of its pp/vpp.

vermouth1992:
We can set an attribute to each layer so that we don't have to worry about torch.nn.ModuleList. Just like https://github.com/huggingface/transformers/blob/main/src/transformers/models/llama/modeling_llama.py#L509

vermouth1992:
> > We can set an attribute to each layer so that we don't have to worry about torch.nn.ModuleList. Just like https://github.com/huggingface/transformers/blob/main/src/transformers/models/llama/modeling_llama.py#L509
> 
> I'll check the `state_dict` generated by this method.

We can modify the keys in state_dict according to the actual layer index.

vermouth1992 (approved):

verl/models/qwen2/megatron/checkpoint_utils/qwen2_loader.py
```diff
@@ -66,11 +66,11 @@ def load_state_dict_to_megatron_qwen2(state_dict,
     def _get_gpt_model(model):
         return model
 
-    def broadcast_params(module):
+    def fetch_params(module):
         for param in module.parameters():
-            torch.distributed.broadcast(param.data,
-                                        src=mpu.get_data_parallel_src_rank(),
-                                        group=mpu.get_data_parallel_group())
+            torch.distributed.fetch(param.data,
```
haolin-nju, line 71: Is it a typo? I cannot find torch.distributed.fetch until pytorch2.6. Do you mean send/recv?

ETOgaosion: We currently support pytorch 2.4 like the document says. A good mention to save us from a bug when migrating to the releasing container.

Issue #751, opened 2025-03-25
在某一个step模型的输出突然全变为!，kl变为Nan


yushuiwx:
same with 

- https://github.com/volcengine/verl/issues/747
- https://github.com/volcengine/verl/issues/721

注意到我们的模型使用的deepseek-ai/DeepSeek-R1-Distill-Qwen-1.5B，!对应的token id是0.

![Image](https://github.com/user-attachments/assets/65b746e5-a24f-4db9-ab3b-6ec28eb09c11)

ftgreat:
请问跑起来至少需要多少卡，谢谢

AIBionics:
可以提供对应的训练脚本和数据源，这样方便项目人员复现和排查

yushuiwx:
> Could you please try the following steps:
> 
> * Resume the last checkpoint and make sure the nan persists
> * Resume from the last checkpoint and step save frequency to 1
> * Resume from the last checkpoint and make sure nan immediately appears in the following step
> * Add a breakpoint in the ppo_actor loss function and print old_log_probs, log_probs and ratio to find out which part in the loss computation causes nan.

Hello, after careful analysis, I found that the issue lies in the calculation of `low_var_kl`. In the problematic step, the loss has normal values (without NaN or Inf), but during backpropagation, the `exp` operation in `ratio = torch.exp(kl)` within `low_var_kl` （https://github.com/volcengine/verl/blob/8cae42dc29736d0802ded43c5ecf67a809d56bd8/verl/trainer/ppo/core_algos.py#L386）causes extreme values, leading to NaN in `kl`. My temporary solution is to use `torch.clamp(kl, min=-5, max=5)` to constrain `kl` before computing `ratio`. This approach works, but I feel uncertain about it. Could you kindly help confirm whether this is an appropriate solution?

yushuiwx:
> Could you please try the following steps:
> 
> * Resume the last checkpoint and make sure the nan persists
> * Resume from the last checkpoint and step save frequency to 1
> * Resume from the last checkpoint and make sure nan immediately appears in the following step
> * Add a breakpoint in the ppo_actor loss function and print old_log_probs, log_probs and ratio to find out which part in the loss computation causes nan.

I found several similar issues [#747,](https://github.com/volcengine/verl/issues/747), [727](https://github.com/volcengine/verl/issues/727), etc . I suspect that they might be caused by the same underlying problem.

LYMDLUT:
[...]

Issue #747, opened 2025-03-25
训练会突然在某个步骤失败，模型某个指标会变成 nan，奖励变成0，然后开始不停重复生成感叹号 !!!!!!!!!


JinliangLu96:
Hi，使用 PPO/GRPO/VERL 训练 PRIME 的数学 RL 数据的时候会出现失败的情况，模型会随机在某一个步骤的 KL/grad norm 变成 nan，然后后面的奖励都是0，生成的内容都是感叹号!!!!!

我尝试通过在之前没出问题的 checkpoint 处重新加载，似乎无效，模型到了某一个特定的步骤处还是会出现 nan 的问题，即使我删除了 data.pt，修改了脚本中的 kl_loss_type, train_batch_size 以及 lr，nan仍然会非常固定在那个步骤出现😂。有趣的是如果我重新跑一个模型，这个错误出现的步骤就不太一样了hhh

下面是我的参数

`set -x

export VLLM_ATTENTION_BACKEND=XFORMERS

math_train_path=data/prime/train.parquet
math_test_path=data/prime/test.parquet

train_files="['$math_train_path']"
test_files="['$math_test_path']"

python3.11 -m verl.trainer.main_ppo \
    algorithm.adv_estimator=reinforce_plus_plus \
    data.train_files="$train_files" \
    data.val_files="$test_files" \
    data.train_batch_size=1024 \
    data.max_prompt_length=1024 \
    data.max_response_length=3072 \
    actor_rollout_ref.model.path=Qwen2.5-7B_niumina_math \
    actor_rollout_ref.actor.optim.lr=1e-6 \
    actor_rollout_ref.model.use_remove_padding=True \
    actor_rollout_ref.actor.ppo_mini_batch_size=256 \
    actor_rollout_ref.actor.ppo_micro_batch_size_per_gpu=8 \
    actor_rollout_ref.actor.use_kl_loss=True \
    actor_rollout_ref.actor.kl_loss_coef=0.001 \
    actor_rollout_ref.actor.kl_loss_type=low_var_kl \
    actor_rollout_ref.model.enable_gradient_checkpointing=True \
    actor_rollout_ref.actor.fsdp_config.param_offload=False \
    actor_rollout_ref.actor.fsdp_config.grad_offload=False \
    actor_rollout_ref.actor.fsdp_config.optimizer_offload=False \
    actor_rollout_ref.rollout.log_prob_micro_batch_size_per_gpu=8 \
[...]

Issue #637, opened 2025-03-17
add loss, grad, vllm weight nan check


vermouth1992:

Kirrito-k423:
Thanks for raising this. I checked the current `main` snapshot (`1e3b6ffcd1815dab5dc46e0cea275657a1823441`, 2026-06-30).

中文：目前看起来 #789 已经覆盖了“grad”这一部分：当前 FSDP / AutoModel / TorchTitan / VeOmni engine 都会在 `grad_norm` 非有限时跳过 `optimizer.step()`。不过我没有找到明确的 loss NaN/Inf 运行时检查，也没有找到 vLLM weight/buffer 更新路径在 `model.load_weights(...)` 或 buffer copy 前做 `torch.isfinite` / `torch.isnan` 校验。vLLM 当前路径主要是接收 bucket、拆分 parameter/buffer、加载参数和复制 buffer。

English: The gradient part appears partially implemented via #789 and retained in the current engine code: several optimizer paths skip the update when `grad_norm` is non-finite. I could not find an explicit runtime guard for non-finite loss values, nor an explicit NaN/Inf validation for transferred vLLM weights/buffers before loading/copying them.

Because the issue body is empty, I would keep this open for clarification rather than close it as completed. Could maintainers confirm the desired remaining behavior: fail fast on non-finite loss, skip update on non-finite loss, and/or validate vLLM transferred weights before loading?

This response was generated by a scheduled Codex GPT5.5 issue processing task.

PR #789, opened 2025-03-27
fix: skip the update step when encountering gradient overflow

zhao9797:
due to issues such as mixed precision updates or corrupted data, model training may crash. to prevent abnormal updates, you can check grad_norm when updating the model, which might be a temporarily effective solution. however, if similar issues occur frequently, it is necessary to further investigate the data and loss design for a more thorough troubleshooting
cover:  #637 #747 #751


vermouth1992 (approved):

vermouth1992:
Currently, the CI machine is overloaded. Let's trigger the CI some time later.

Issue #779, opened 2025-03-26
how to support sync reference model


felixdae:
a feature like [this](https://github.com/huggingface/trl/pull/2700)

lihaoling:
+1

gameofdimension:
a simple and fast implementation is here https://github.com/volcengine/verl/pull/794

Kirrito-k423:
定时清理检查结果：这个 issue 最近一次由当前运行账号以外账号产生的公开回复或公开引用是 `2025-03-28T06:16:18Z`。本轮检查日期为 2026-07-04，该时间已经超过 6 个月。检查来源包括 issue comments API `https://api.github.com/repos/verl-project/verl/issues/779/comments?per_page=100` 和 timeline API `https://api.github.com/repos/verl-project/verl/issues/779/timeline?per_page=100`。

按照本仓库的定时 issue 清理标准，我会先留下这条说明并关闭该 issue。若作者或社区成员仍能提供新的复现、证据或需求背景，请继续评论请求重开，或创建新的 issue 并引用这里的上下文。

---

Scheduled cleanup check: the latest public reply or public reference from an account other than the current automation account was `2025-03-28T06:16:18Z`. The check date for this run is 2026-07-04, so that activity is more than 6 months old. The checked public sources were the issue comments API `https://api.github.com/repos/verl-project/verl/issues/779/comments?per_page=100` and the timeline API `https://api.github.com/repos/verl-project/verl/issues/779/timeline?per_page=100`.

Following the scheduled issue cleanup policy, I am leaving this note first and then closing the issue. If the author or community can still provide a fresh reproduction, evidence, or requirement context, please comment to request reopening or open a new issue that links back to this one.

This response was generated by a scheduled Codex GPT5.5 issue processing task.

PR #794, opened 2025-03-28
tentative approach to sync ref model

gameofdimension:
Try to address issue https://github.com/volcengine/verl/issues/779. That is implementing periodic synchronization between the reference model and the actor.

We've observed that the reference model and actor with the same rank exist in the same process, despite belonging to different worker groups. Here's our proposed solution:

1. Have the driver collect actor model IDs/addresses from the actor worker group
2. Pass this information to the reference worker group
3. Reference workers restore actors from the corresponding IDs

### DIFF
--- a/examples/grpo_trainer/run_qwen2-7b_sync_ref_model.sh
+++ b/examples/grpo_trainer/run_qwen2-7b_sync_ref_model.sh
@@ -1,0 +1,45 @@
+set -x
+
+export VLLM_ATTENTION_BACKEND=XFORMERS
+
+python3 -m verl.trainer.main_ppo \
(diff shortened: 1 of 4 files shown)


gameofdimension:
@vermouth1992 @hiyouga
Hi, how do you think feature https://github.com/volcengine/verl/issues/779 and this implementation

eric-haibin-lin (commented):
I think this is part of a paradigm of various ways to "asynchronously" update the weights of xxx roles in RL and require careful thoughts instead of quick patches. @PeterSH6 have you thought of this pattern (sync ref model & actor)?

PR #800, opened 2025-03-28
Update checkpoint doc

ETOgaosion:
Also fix some APIs.


examples/checkpoint/run_deepseek_megatron_ckpt.sh
```diff
@@ -60,22 +57,19 @@ python3 -m verl.trainer.main_ppo --config-path=config \
     actor_rollout_ref.actor.ppo_mini_batch_size=256 \
     actor_rollout_ref.actor.ppo_micro_batch_size_per_gpu=4 \
     actor_rollout_ref.actor.megatron.pipeline_model_parallel_size=2 \
-    actor_rollout_ref.actor.megatron.virtual_pipeline_model_parallel_size=2 \
```
eric-haibin-lin, line 63: is this change intentional?

eric-haibin-lin (commented):
nice doc!

eric-haibin-lin (approved):
lgtm. checked rendering, it works fine

examples/checkpoint/run_deepseek_megatron_ckpt.sh
ETOgaosion: Yeah, deepseek 7b has 30 layers, so not supports pp * vpp = 4, as 30 % 4 != 0

PR #809, opened 2025-03-28
docs: add triton compile err to faq

eric-haibin-lin:


vermouth1992:
I guess this is probably not enough as flash-attn cross_entropy_loss uses triton as well :( The root error comes from triton, not from torch.compile

tongyx361 (approved):

PR #810, opened 2025-03-28
feat: fix last step save strategy and check for folder creation

BaiqingL:
The old logic would not save at the last step, as it requires save_freq to be a positive number, so negative numbers are filtered. In addition, adding in an `os.makedirs` call to attempt to create the folder, keeping in mind not to overwrite if a folder already exists via the exist_ok parameter to resolve a low hanging fruit todo.


dirtyDan0:
the setting of save_freq allows for only save the last step(by setting it to be a very large num e.g. 99999) or not to save it(when<=0). it's on purpose. However, the new logic no longer support not saving.

BaiqingL:
> the setting of save_freq allows for only save the last step(by setting it to be a very large num e.g. 99999) or not to save it(when<=0). it's on purpose. However, the new logic no longer supports not saving.

The old logic required that self.config.trainer.save_freq > 0 as one of the conditions, so if the save_freq was set to -1, there is no save logic possible, unless I missed it somewhere? The reason being when I was using save_freq = -1 I observed there was no checkpoints being saved at the last step

BaiqingL:
I thought that the purpose of having save_freq == -1 is to trigger the save on last step behavior, so if this is not the case, what should the user configure to only save on last step?

PR #811, opened 2025-03-28
add pylint to CI

frederrx:
* add a workflow to run pylint
* add a section to `pyproject.toml` that blacklists all rules which would trigger given the current code
* pin a version of pylint in `requirements.txt` for reproducability

In a followup PR I will remove some rules from the blacklist and fix some bugs.


eric-haibin-lin (approved):
thanks!

Issue #799, opened 2025-03-28
Typo of Megatron installation guide


dingyuan-shi:
There is a typo in the installation command of Megatron-LM.

<img width="743" alt="Image" src="https://github.com/user-attachments/assets/61eabf17-236c-4ebb-87bd-aba4fc78ab14" />

The branch name here should be ``core_r0.11.0`` rather than ``core_v0.11.0``

<img width="1164" alt="Image" src="https://github.com/user-attachments/assets/bfd63477-11e0-40e9-88da-f21073aa8f0c" />

eric-haibin-lin:
Thanks for reporting. PR is welcome! https://github.com/volcengine/verl/blob/main/docs/start/install.rst

dingyuan-shi:
> Thanks for reporting. PR is welcome! https://github.com/volcengine/verl/blob/main/docs/start/install.rst

Ok, I have create the PR: https://github.com/volcengine/verl/pull/813

PR #813, opened 2025-03-29
Fix typo on installation guide

dingyuan-shi:
Modify the version number of Megatron-llm from ``core_v0.11.0`` to ``core_r0.11.0``

### DIFF
--- a/docs/start/install.rst
+++ b/docs/start/install.rst
@@ -112,7 +112,7 @@ Megatron is optional. It's dependencies can be setup as below:
    # transformer engine
    pip3 install git+https://github.com/NVIDIA/TransformerEngine.git@stable
 
-   git clone -b core_v0.11.0 https://github.com/NVIDIA/Megatron-LM.git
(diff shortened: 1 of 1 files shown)


vermouth1992:
Maybe we should use this tag? https://github.com/NVIDIA/Megatron-LM/commits/v0.11.0/ @ETOgaosion

dingyuan-shi:
> Maybe we should use this tag? https://github.com/NVIDIA/Megatron-LM/commits/v0.11.0/ @ETOgaosion

based on the discussion of this issue (https://github.com/volcengine/verl/issues/799), I think the branch name should be core_r0.11.0

eric-haibin-lin:
i agree using the release tag would be more precise since the branch commit may be updated in the future https://github.com/NVIDIA/Megatron-LM/commits/v0.11.0/

dingyuan-shi:
Ok, I have changed to use the tag

vermouth1992 (approved):

PR #818, opened 2025-03-29
[log] feat: timer log

tongyx361:
This PR adds logs like this:

```stdout
step start at 2025-03-30 01:42:51.773944
gen start at 2025-03-30 01:42:51.774983
```


PeterSH6:
Would it be better if we make it to debug level?

PR #820, opened 2025-03-29
[BREAKING config] fix: move val_before_train to config yaml. Using trainer.val_before_train instead of +trainer.val_before_train going forward

tongyx361:


eric-haibin-lin (commented):
could u modify all example script, and the verl getting started doc?

```
Could not append to config. An item is already at 'trainer.val_before_train'.
Either remove + prefix: 'trainer.val_before_train=False'
Or add a second + to add or override 'trainer.val_before_train': '++trainer.val_before_train=False'
```

eric-haibin-lin:
Looks like it will likely break all existing users' script if val_before_fit is set. But i guess put it in the config is the right thing

eric-haibin-lin (approved):

eric-haibin-lin:
@tongyx361 could you update config.rst in a new PR to explain the meaning of this flag? thanks

Issue #608, opened 2025-03-15
[Bug] Assertion Errors introduced by Math Verifier, while not influencing the correctness of Math dataset testing.


ETOgaosion:
Commit d4a00ef introduced bugs with math-verify, including others complaint of installation error, and verification assertion errors. It seems that solution and prediction strings format mismatch with the math-verify requests. Below is the error, and there are also NaN errors.

![Image](https://github.com/user-attachments/assets/da11f58b-cced-48f1-a561-c80a446a5ff5)

It seems that both Qwen-2-7b-Instruct and Qwen2.5-7b-Instruct have such error.

But it causes not that much harm to training process, because of the usage of asyncio.

![Image](https://github.com/user-attachments/assets/031474aa-fb73-4428-9f3f-45455b621024)

Maybe there are 2 solutions:

1. Force the alignment between models outputs and math-verify inputs.
2. Add options to choose whether to use math-verify.

yyDing1:
Thank you for pointing out these issues.

These error messages do not affect the calculation of rewards or the training process, as math-verify will return `false` for any answers that cannot be correctly parsed or compared. However, math-verify still outputs these intermediate error messages.

Additionally, it is not the usage of `asyncio` that avoids this issue. I have tried sequential execution, and it still does not interrupt due to these errors.

In fact, these intermediate error messages can indeed be distracting for developers. I will attempt to remove these error messages

Kirrito-k423:
中文：我建议关闭这个 issue。原始问题的核心影响面已经被后续改动移除：#683 已合并并将 Math-Verify 改为可选依赖，同时恢复 MATH 数据集默认使用原 verifier；当前 `verl/utils/reward_score/__init__.py` 中 MATH 数据集默认调用 `math_reward.compute_score`，`math_verify` 只作为注释中的可选集成保留。后续 #5839 也进一步为 `verl/utils/reward_score/math_verify.py` 增加了子进程执行和超时保护。结合维护者在本 issue 中已确认这些中间错误信息不会影响 reward/training 计算，我认为该 issue 描述的默认路径问题已经解决。

[...]

PR #829, opened 2025-03-30
WIP [reward] fix: enable math-verify if available

eric-haibin-lin:


eric-haibin-lin:
Unfortunately i come across issues like
https://github.com/huggingface/Math-Verify/issues/30#issuecomment-2764744494

@yyDing1 did you figure out how to work around it?

yyDing1:
These error messages do not affect the calculation of rewards or the training process, as math-verify will return false for any answers that cannot be correctly parsed or compared. (https://github.com/volcengine/verl/issues/608)

However, the current implementation of `math-verify` may not handle these intermediate error messages gracefully.

To resolve this, you can make a simple modification to the math-verify package's source code:

Replace `logger.exception` with `logger.error` in the following section:

https://github.com/huggingface/Math-Verify/blob/7923db0ecebde0b58b83b1df8faaccc85e69541e/src/math_verify/grader.py#L807-L817

After this change, the system will only output a concise `"Error during comparison"` message instead of verbose stack traces.

PR #851, opened 2025-03-31
Support Megatron 0.11.0 and vLLM 0.8.2, update images to use latest vllm and Megatron

ETOgaosion:
# Codes are ready, devided into 3 PR

This PR includes:

# 1. Support of Megatron 0.11.0 and vLLM 0.8.2

## Per-tensor weights loading to reduce peak memory cost

image

## Mitigate the concept of Micro-DP

There are many cases that micro-dp cannot handle well and may need complicated logics, Now we directly use tensor parallel of Megatron and vLLM. Since per-tensor all-gather has a minimal memory cost, so the communication group itself does not matter most.

## Import Megatron modules on GPU nodes

This is because now `import megatron.core` will end up call triton, it will detect current GPU drivers and cause error if we initialize these on CPU node. So lots of codes are refactored.

# 2. Update docker image and doc

- [ ] TODO: vLLM test `test_vllm_hf_loader.py` still use old vLLM API, so currently stay unchanged.

# 3. Fix rng checkpoints

Only 1 node in a machine save rng_states to avoid conflicts.

# TODO

- [ ] Not Support Context Parallel currently

### DIFF
--- a/verl/utils/checkpoint/megatron_checkpoint_manager.py
+++ b/verl/utils/checkpoint/megatron_checkpoint_manager.py
@@ -302,9 +302,9 @@ class MegatronCheckpointManager(BaseCheckpointManager):
             torch.distributed.barrier()
 
             rng_state_path = get_rng_states_checkpoint_path(local_path)
-            rng_state = self.get_rng_state()
-            torch.save(rng_state, rng_state_path)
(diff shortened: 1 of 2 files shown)


eric-haibin-lin:
why was this PR closed?

PR #858, opened 2025-04-01
Add `mathruler` to requirement

JustinTong0323:
Which is required here:
https://github.com/volcengine/verl/blob/776b0a9ddc5a5a8c7d9adf07172cac774ab2afc7/verl/utils/reward_score/geo3k.py#L16

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -19,3 +19,4 @@ torchdata
 transformers
 # vllm==0.6.3.post1
 wandb
+mathruler


hiyouga:
Use `pip install -e .[geo]` to install mathruler https://github.com/volcengine/verl/blob/9dcac14f1ba7c9a729ded6e4807b2fe89af321ec/setup.py#L45

PR #861, opened 2025-04-01
[test]: update vllm_spmd test for > 0.7.3

BearBiscuit05:
I tested the `Deepseek-7B-chat` and `Qwen2-7B` these two models, the former showed a 0% difference in output, while the latter exhibited a 10.25% difference in output, with no significant issues in the output. So I manually adjusted the error tolerance to 15%. I’m not sure if this will work.

### DIFF
--- a/tests/rollout/test_vllm_spmd.py
+++ b/tests/rollout/test_vllm_spmd.py
@@ -15,13 +15,15 @@
 import os
 import torch
 import transformers
+from torch.distributed.fsdp import FullyShardedDataParallel as FSDP, ShardingStrategy, MixedPrecision, CPUOffload
+from torch.distributed.fsdp.api import ShardingStrategy, ShardedStateDictConfig, StateDictType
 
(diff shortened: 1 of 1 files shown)


tests/rollout/test_vllm_spmd.py
```diff
@@ -152,11 +183,14 @@ def test_vllm_spmd():
         generated_text = output.outputs[0].text
         vllm_response_tokens.append(generated_text)
 
-    print(f'vllm response: {vllm_response_tokens}')
+    if torch.distributed.get_rank() == 0:
+        print(f'hf response: {hf_response_tokens}')
+        print(f'vllm response: {vllm_response_tokens}')
     assert are_lists_similar(hf_response_tokens, vllm_response_tokens), \
         f'Strings differ more than 10%:\n'
```
hiyouga, line 190: It should be 15% now, right?

hiyouga (approved):

PR #868, opened 2025-04-01
[algo] misc: remove redundant tile([1, response_length]), efficient broadcast instead

dirtyDan0:
as titled

### DIFF
--- a/verl/trainer/ppo/core_algos.py
+++ b/verl/trainer/ppo/core_algos.py
@@ -106,7 +106,7 @@ def compute_gae_advantage_return(token_level_rewards: torch.Tensor, values: torc
 # NOTE(sgm): this implementation only consider outcome supervision, where the reward is a scalar.
 def compute_grpo_outcome_advantage(token_level_rewards: torch.Tensor,
                                    response_mask: torch.Tensor,
-                                   index: torch.Tensor,
+                                   index: np.ndarray,
                                    epsilon: float = 1e-6):
     """
     Compute advantage for GRPO, operating only on Outcome reward 
@@ -123,7 +123,6 @@ def compute_grpo_outcome_advantage(token_level_rewards: torch.Tensor,
         Returns: `(torch.Tensor)`
             shape: (bs, response_length)
     """
-    response_length = token_level_rewards.shape[-1]
     scores = token_level_rewards.sum(dim=-1)
 
     id2score = defaultdict(list)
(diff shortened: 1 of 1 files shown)


verl/trainer/ppo/core_algos.py
```diff
@@ -145,14 +160,15 @@ def compute_grpo_outcome_advantage(token_level_rewards: torch.Tensor,
                 raise ValueError(f"no score in prompt index: {idx}")
         for i in range(bsz):
             scores[i] = (scores[i] - id2mean[index[i]]) / (id2std[index[i]] + epsilon)
-        scores = scores.unsqueeze(-1).tile([1, response_length]) * eos_mask
 
-    return scores, scores
+        returns = scores.unsqueeze(-1) * eos_mask
+
+    return returns * compute_last_mask(eos_mask), returns
```
hiyouga: the valid number of elements in returns should be the same as the number of actions  https://github.com/volcengine/verl/blob/4fec38c5b3cf4ec45d27b3148e822c2d1881282b/verl/trainer/ppo/core_algos.py#L297

WDYT? @vermouth1992

dirtyDan0: eos mask == response mask (all valid response tokens are 1 otherwise 0), so `returns = scores.unsqueeze(-1) * eos_mask` gives all valid tokens the same final score. 
https://github.com/volcengine/verl/blob/4fec38c5b3cf4ec45d27b3148e822c2d1881282b/verl/trainer/ppo/ray_trainer.py#L192

For ppo_kl, **the mask is done by advantages**. If we want it to be outcome reward model, advantages have already masked all other tokens as 0 in the first place. Only EOS have advs.

https://github.com/volcengine/verl/blob/4fec38c5b3cf4ec45d27b3148e822c2d1881282b/verl/trainer/ppo/core_algos.py#L299-L305

dirtyDan0: The name of EOS mask is quiet misleading. It's reponse mask in fact.
https://github.com/volcengine/verl/blob/4fec38c5b3cf4ec45d27b3148e822c2d1881282b/verl/trainer/ppo/ray_trainer.py#L166-L176

https://github.com/volcengine/verl/blob/4fec38c5b3cf4ec45d27b3148e822c2d1881282b/verl/trainer/ppo/ray_trainer.py#L846

vermouth1992: Yes... It's actually response_mask...Could you please submit a PR to change it? Thanks!

hiyouga: @dirtyDan0 But if we use `last_mask`, all other response tokens except eos token will not receive any gradients, right?

dirtyDan0: I checked trl and found that for rloo both EOS only and sequence-level exist.  By default sequence-level.
This boils down to how you treat rlhf, as modeling the entire generation as a single action, or to each token.
I think we can add a config like trl.

https://github.com/huggingface/trl/blob/657f9ce6ee10b29187351c34e645bcaac0a522a8/trl/trainer/rloo_trainer.py#L409-L435

https://github.com/huggingface/trl/blob/657f9ce6ee10b29187351c34e645bcaac0a522a8/trl/trainer/rloo_config.py#L99-L102

hiyouga: If we use outcome reward model, just use the sequence-level loss

vermouth1992 (approved):

PR #871, opened 2025-04-01
Batch Rewards

mertunsall:


hiyouga (approved):

hiyouga:
remember to run `bash scripts/format.sh`

mertunsall:
Finished reformatting.

mertunsall:
I synced with main as well @hiyouga

mertunsall:
There was a test not passing due to missing license, I added it.

mertunsall:
@hiyouga can you explain why the checks are not passing?

hiyouga:
@mertunsall the license should start with https://github.com/volcengine/verl/blob/8cae42dc29736d0802ded43c5ecf67a809d56bd8/verl/__init__.py#L1

mertunsall:
are you sure? this one is different: https://github.com/volcengine/verl/blob/8cae42dc29736d0802ded43c5ecf67a809d56bd8/verl/workers/reward_manager/prime.py

hiyouga:
@mertunsall you can check this script https://github.com/volcengine/verl/blob/8cae42dc29736d0802ded43c5ecf67a809d56bd8/tests/sanity/check_license.py

mertunsall:
I made a fix for this, license check seems to pass right now. I added an option for an individual contributor license.

Do you have any insights on the other test that doesn't pass?

hiyouga (dismissed):

mertunsall:
resolved merge conflicts.

hiyouga (dismissed):

mertunsall:
I realized some more changes were brought by DAPO release - I fixed them and also added an example.

verl/trainer/main_ppo.py
```diff
@@ -35,6 +35,7 @@ def get_custom_reward_fn(config):
     spec = importlib.util.spec_from_file_location("custom_module", file_path)
     module = importlib.util.module_from_spec(spec)
     try:
+        sys.modules["custom_module"] = module
```
patrik-bartak, line 37: This is a bugfix right? Without this line, I get errors when using the Prime reward manager + a custom reward function. I wanted to submit an issue and fix it but I see you ran into the same problem. Correct me if I am wrong.

Maybe then this line should also be added to the get_custom_reward_fn inside main_eval.py?

mertunsall: yes, thanks for pointing this out. I am adding some new functionalities for adding kwargs to the reward function as well, I will add this once everything is tested

mertunsall:
One check was cancelled for some reason - but I think this should be ready to merge @hiyouga

hiyouga:
Yeah let's merge it

hiyouga (approved):

verl/workers/reward_manager/batch.py
```diff
@@ -0,0 +1,107 @@
+# Copyright 2025 Individual Contributor: Mert Unsal
```
eric-haibin-lin, line 1: Thanks for the PR. Could you explain when & how this reward manager can be used?

hiyouga: Using batch rewards, we can use multiprocessing to accelerate the computation of reward scores

Django-Jiang: hi, I just wonder what is the different of batch reward manager compared to prime

patrik-bartak: There can be situations where computing the reward for a batch using one call is much faster than performing a call for each sample. Especially if your reward is implemented as a service.

patrik-bartak:
@mertunsall 
Hey I just wanted to ask here in case you already know the answer before I start digging into the code, but the data_sources within a batch can be different right? For example, if we configure multiple evaluation datasets, we have to make sure that the reward function is able to handle with a batch where data comes from different data sources in that batch.

mertunsall:
Yes, that sounds right to me!

xin-ran-w:
How to use this reward manager in config.

PR #889, opened 2025-04-03
refactor: remove unused filter_prompts parameter from RLHFDataset

0x404:
`filter_prompts` has never been used, I think this parameter has been replaced by `filter_overlong_prompts` so we can simply remove this.

### DIFF
--- a/docs/workers/ray_trainer.rst
+++ b/docs/workers/ray_trainer.rst
@@ -25,7 +25,6 @@ then tokenize.
                                        tokenizer=self.tokenizer,
                                        prompt_key=self.config.data.prompt_key,
                                        max_prompt_length=self.config.data.max_prompt_length,
-                                       filter_prompts=True,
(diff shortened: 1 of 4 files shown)


hiyouga (approved):

0x404:
just merge main and resolve conflicts

eric-haibin-lin (approved):
thanks for the PR!

eric-haibin-lin:
I think the RLHF Dataset needs customization as well. @0x404 would u be interested to enable external classes following this PR? https://github.com/volcengine/verl/commit/77babf195669e4bc5691ef41ff4d4338ba443097

0x404:
> I think the RLHF Dataset needs customization as well. @0x404 would u be interested to enable external classes following this PR?

Yeah I'm interested, sounds like a good implementation. I can implement this in next few days.

PR #890, opened 2025-04-03
refactor: use hf Dataset instead of pandas DataFrame in RLHFDataset

0x404:
HF Dataset provides better memory management and can handle larger datasets. It also supports multi-process acceleration during map/filter operations (while pandas requires version >2.0).

Now we can specify `filter_overlong_prompts` on large-scale datasets when set `filter_overlong_prompts_workers` to a appreciate num.

### DIFF
--- a/docs/amd_tutorial/amd_build_dockerfile_page.rst
+++ b/docs/amd_tutorial/amd_build_dockerfile_page.rst
@@ -48,6 +48,7 @@ docker/Dockerfile.rocm
         liger-kernel \
         numpy \
         pandas \
+        datasets \
         peft \
         "pyarrow>=15.0.0" \
         pylatexenc \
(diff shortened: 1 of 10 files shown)


hiyouga (dismissed):

eric-haibin-lin (approved):

twangnyc:
One of the breaking behavior change on this would be we can no longer point a directory as the `config.data.val_files` for `config.data.train_files` as previously `pd.read_parquet` support directly reading from a directory... @0x404

0x404:
HI @twangnyc, I think you can use wildcard if you want to use the entire directory as your dataset. for example, you can set your `config.data.val_files` to `your_dir/*.parquet`, and this should work.

twangnyc:
@0x404 Thanks for your reply! Yes we are using wildcard to temporary fix our failing code running previously. Noticed you raised the new PR to make it compatible with pd.Dataframe behavior. Appreciate it!

PR #902, opened 2025-04-03
[data] feat: last_user_msg_template via Jinja

tongyx361:
This PR allows users to flexibly customize the content of the last user message via Jinja template.


verl/utils/dataset/rl_dataset.py
```diff
@@ -75,13 +76,18 @@ def process_image(image: dict, max_pixels: int = 2048 * 2048, min_pixels: int =
...
     """
     We assume the dataset contains a column that contains prompts and other information
+
+    Args:
+        last_user_msg_template: str
+            Jinja template with the message dict as input, e.g., "{{ content | trim }}\n\nLet's think step by step."
     """
 
     def __init__(self,
                  parquet_files: Union[str, List[str]],
```
eric-haibin-lin, line 86: Just FYI we shall change the dataset API to take data.config as the input so that it's easier for others to extend verl with custom dataset
https://github.com/volcengine/verl/blob/main/verl/trainer/fsdp_sft_trainer.py#L128-L131

eric-haibin-lin (dismissed):

eric-haibin-lin:
Dataset tests failed..
image

verl/utils/dataset/rl_dataset.py
tongyx361: Indeed, allowing to customize the dataset class is the most flexible way, especially for a framework! However, configuration via files are still meaningful for conducting many comparative experiments.

PR #912, opened 2025-04-04
Add seed to vllm_rollout_spmd.py

sunjin-k:
See https://github.com/vllm-project/vllm/commit/8b664706aa72315e436097d19257536228e01792 
In summary, now when using external launcher in vLLM, a Seed must be set.

### DIFF
--- a/tests/rollout/test_vllm_spmd.py
+++ b/tests/rollout/test_vllm_spmd.py
@@ -168,6 +168,7 @@ def test_vllm_spmd():
         skip_tokenizer_init=False,
         enable_prefix_caching=True,
         trust_remote_code=True,
+        seed=0,
     )
 
     # Warmup iterations
--- a/verl/workers/rollout/vllm_rollout/vllm_rollout_spmd.py
+++ b/verl/workers/rollout/vllm_rollout/vllm_rollout_spmd.py
@@ -124,6 +124,7 @@ class vLLMRollout(BaseRollout):
             enable_chunked_prefill=config.enable_chunked_prefill,
             enable_prefix_caching=True,
             trust_remote_code=trust_remote_code,
+            seed=0,
         )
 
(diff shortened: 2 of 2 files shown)


sunjin-k:
Added 'seed' to one more file where 'external_launcher' is used

PeterSH6:
Does the seed affect the responses' quality/distribution?

sunjin-k:
I don't know..

vermouth1992:
I assume the seed should depend on the model parallel rank.

vermouth1992:
@sunjin-k Could you set the seed based on the data parallel rank. e.g. seed = seed + external_dp_rank? And make seed configurable with config. Thanks!

verl/workers/rollout/vllm_rollout/vllm_rollout_spmd.py
```diff
@@ -124,6 +124,7 @@ def __init__(self, model_path: str, config: DictConfig, tokenizer, model_hf_conf
             enable_chunked_prefill=config.enable_chunked_prefill,
             enable_prefix_caching=True,
             trust_remote_code=trust_remote_code,
+            seed=0,
```
hiyouga: Let us use different seed among different dp rank

```python
import os

seed = int(os.getenv("RANK", "0")) // tensor_parallel_size
```

tests/rollout/test_vllm_spmd.py
```diff
@@ -168,6 +168,7 @@ def test_vllm_spmd():
         skip_tokenizer_init=False,
         enable_prefix_caching=True,
         trust_remote_code=True,
+        seed=0,
```
hiyouga: same as below

hiyouga (changes requested):

hiyouga (approved):

PR #916, opened 2025-04-04
Add Rollout and Validation dumps to file

MantasBaksys:


eric-haibin-lin (commented):
thanks for the contribution! would u mind edit your commit with your github id, and also resolve conflicts?

MantasBaksys:
@eric-haibin-lin done!

mertunsall:
@eric-haibin-lin Can we merge this? We tested on our end and it looks good

verl/trainer/config/ppo_trainer.yaml
```diff
@@ -194,8 +194,11 @@ trainer:
   total_training_steps: null
   project_name: verl_examples
   experiment_name: gsm8k
-  logger: [ 'console', 'wandb' ]
-  log_val_generations: 0
+  logger: [ 'console', 'database' ]
+  validation_data_dir: ~/validation_data # directory for logging the validation data
```
eric-haibin-lin: let's reorganize it this way:
```
logger: ['console', ['database']
logger_kwargs:
    val_data_dir: 
    log_val_generations: true/false
    rollout_data_dir: 
    log_rollout_generations: true/false
```

verl/trainer/ppo/ray_trainer.py
```diff
@@ -274,6 +274,9 @@ def __init__(self,
         self.use_rm = Role.RewardModel in role_worker_mapping
         self.ray_worker_group_cls = ray_worker_group_cls
         self.validation_generations_logger = ValidationGenerationsLogger()
```
eric-haibin-lin: can we pass the dir directly to ValidationGenerationsLogger & RolloutLogger? no need to store the dir in trainer

verl/utils/tracking.py
```diff
@@ -179,13 +180,17 @@ def _flatten_dict(raw: Dict[str, Any], *, sep: str) -> Dict[str, Any]:
 @dataclasses.dataclass
```
eric-haibin-lin: could u add a unit test for these two loggers?

verl/trainer/ppo/ray_trainer.py
```diff
@@ -481,7 +484,20 @@ def _create_dataloader(self):
             self.config.actor_rollout_ref.actor.optim.total_training_steps = total_training_steps
             self.config.critic.optim.total_training_steps = total_training_steps
 
-    def _maybe_log_val_generations(self, inputs, outputs, scores):
+    def _log_all_generations(self, inputs, outputs, scores, epoch=None):
+        """Log a table of rollout samples to the configured logger (upload to Hub preffered)"""
```
eric-haibin-lin: preffered -> preferred? 
what Hub?

verl/trainer/ppo/ray_trainer.py
```diff
@@ -919,6 +939,18 @@ def fit(self):
 
                         batch.batch['token_level_scores'] = reward_tensor
 
+                        if self.config.trainer.log_rollout_generations > 0:
+                            # Extract inputs, outputs and scores from the batch
+                            input_ids = batch.batch['input_ids']
```
eric-haibin-lin: can we move the preprocessing logic for logging to the _log_all_generations function instead. also rename `self._log_all_generations` to `self._maybe_log_all_generations` and move the condition check `if self.config.trainer.log_rollout_generations` inside the function. 

In principle we want to keep the training loop succinct

verl/utils/tracking.py
```diff
@@ -255,3 +260,60 @@ def log_generations_to_mlflow(self, samples, step):
...
+    def log(self, loggers, samples, step, epoch=None, data_dir=None):
+        if 'database' in loggers:
+            if data_dir is None:
+                raise ValueError("Data directory must be provided for database logging")
+            if epoch is None:
+                raise ValueError("Epoch number be provided for database logging")
+            self.log_generations_to_database(samples, step, epoch, data_dir)
+
+    def log_generations_to_database(self, samples, step, epoch, data_dir):
+        """Log the rollout samples to a parquet file at data_dir"""
```
eric-haibin-lin: in the docstring please specify the expected input format for samples. maybe as a logger it should just take a dict as the input, and let user prepare the `data` dict outside of the logger? i'm not sure how logging requirements will evolve overtime, since there's on going efforts for vlm, browser-use, etc

eric-haibin-lin (changes requested):
thanks for the PR! left a few comments

MantasBaksys:
Thanks for the comments! I have gone through each of them and adressed them in the new commit. Feel free to have another look and let's get this merged 🚀 @eric-haibin-lin

verl/utils/tracking.py
```diff
@@ -178,14 +179,23 @@ def _flatten_dict(raw: Dict[str, Any], *, sep: str) -> Dict[str, Any]:
...
         if 'swanlab' in loggers:
             self.log_generations_to_swanlab(samples, step)
         if 'mlflow' in loggers:
             self.log_generations_to_mlflow(samples, step)
+        if 'database' in loggers:
+            if self.data_dir is None:
+                raise ValueError("data_dir must be provided during initialization for database logging")
+            self.log_generations_to_database(samples, step, epoch)
 
     def log_generations_to_wandb(self, samples, step):
```
mertunsall: if we log all generations to wandb, I assume it's going to be super slow. but this is automatically called to log all generations to wandb when we include `'wandb'` in logger. Maybe I am wrong here, but I think there should be an option to dump all rollout locally and none to wandb. Can you please test this with `wandb` included in the logger?

mertunsall:
We severely simplified implementation and I tested it extensively. the idea was to decouple logging to wandb, swanml, etc. from the local dump since local dump is very cheap and doesn't rely on the network connection. 

Here is an example rollout data saved locally and then pushed to HF: https://huggingface.co/datasets/mertunsal/test_rollout

@eric-haibin-lin please let me know what you think!

mertunsall:
Hi @hiyouga , would it be possible for you to review this PR please?

verl/trainer/ppo/ray_trainer.py
```diff
@@ -632,6 +663,19 @@ def _validate(self):
 
         self._maybe_log_val_generations(inputs=sample_inputs, outputs=sample_outputs, scores=sample_scores)
 
+        # dump generations
+        from omegaconf import OmegaConf
+
+        val_data_dir = OmegaConf.select(self.config.trainer, "validation_data_dir", default=None)
```
hiyouga: use self.config.trainer.get("validation_data_dir")

verl/trainer/ppo/ray_trainer.py
```diff
@@ -1049,6 +1093,22 @@ def fit(self):
                         actor_output_metrics = reduce_metrics(actor_output.meta_info["metrics"])
                         metrics.update(actor_output_metrics)
 
+                    # Log rollout generations if enabled
+                    rollout_data_dir = OmegaConf.select(self.config.trainer, "rollout_data_dir", default=None)
```
hiyouga: same

verl/trainer/ppo/ray_trainer.py
```diff
@@ -531,6 +531,37 @@ def _create_dataloader(self):
...
+        for key, values in reward_extra_infos_dict.items():
+            if len(values) == len(inputs):
+                data_dict[key] = values
+
+        data_dict["step"] = [self.global_steps] * len(inputs)
+
+        from datasets import Dataset
+
+        ds = Dataset.from_dict(data_dict)
+        ds.to_parquet(dump_filename)
```
hiyouga: why not use jsonl for visualization?

hiyouga:
@mertunsall done

verl/trainer/ppo/ray_trainer.py
mertunsall: it was easier to push to HF - changed it to JSONL now.

verl/trainer/ppo/ray_trainer.py
mertunsall: done

verl/trainer/ppo/ray_trainer.py
mertunsall: done

mertunsall:
changed based on your review @hiyouga - should be good to merge!

hiyouga (approved):

eric-haibin-lin (approved):

PR #923, opened 2025-04-05
[feat] enable FSDP and vLLM 0.8.2 to support DPSK v3 training.

BearBiscuit05:
This is a solution to an error that occurs when vLLM 0.8.2 loads models in the ds_v3 MoE format. For MoE models, directly calling vLLM's `load_weights· function will result in an error, primarily because the model params lose the related methods of the `FusedMoe` class. Therefore, it is necessary to identify and call the specific parameter loading method based on the model parameter names at runtime. However, the current version is a dirty implementation because, in reality, if invasive changes were made to vLLM, it would only require modifying 2 lines of code and adding a new function to identify the layer count (I have already marked comments in the code). I’m not sure if there’s a better implementation and would like some suggestions.

### DIFF
--- a/tests/rollout/test_vllm_spmd.py
+++ b/tests/rollout/test_vllm_spmd.py
@@ -103,25 +103,7 @@ def test_vllm_spmd():
     print('start generation')
     input_ids = input_ids.cuda()
     attention_mask = attention_mask.cuda()
-    batch_size = input_ids.size(0)
-
-    generation_config = GenerationConfig(do_sample=False)
-    actor_model.cuda()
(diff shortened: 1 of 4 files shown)


BearBiscuit05:
And I think the best way to compare whether vLLM resharding is accurate should be to directly compare the inference results of vLLM with the inference results after resharding the parameters exported by FSDP to vLLM.

eric-haibin-lin (approved):

gxy-gxy:
请问可以分享启动v3的脚本吗？

感谢！

PR #924, opened 2025-04-05
feat: support customized datasets for RayPPOTrainer

0x404:
This PR enable user to specify their customized dataset for RayPPOTrainer.

NOTE: the RLHFDataset interface has been broken into:
```
RLHFDataset(
    data_files: Union[str, List[str]],
    tokenizer: PreTrainedTokenizer,
    config: DictConfig,
    processor: Optional[ProcessorMixin] = None
)
```

and the custom dataset class MUST also use this interface.

cc @eric-haibin-lin


verl/trainer/ppo/ray_trainer.py
```diff
@@ -406,21 +407,71 @@ def check_mutually_exclusive(mbs, mbs_per_gpu, name: str):
...
-                                         image_key=self.config.data.get('image_key', 'images'),
-                                         max_prompt_length=self.config.data.max_prompt_length,
-                                         return_raw_chat=self.config.data.get('return_raw_chat', False),
-                                         truncation=self.config.data.get('truncation', 'error'),
-                                         filter_overlong_prompts=self.config.data.filter_overlong_prompts,
-                                         num_workers=self.config.data.get('filter_overlong_prompts_workers', None))
-        assert self.train_dataset.truncation == self.config.data.get(
-            'truncation', 'error'
-        ), f'dataset truncation {self.train_dataset.truncation} must be the same as config {self.config.data.get("truncation", "error")}'
+        from verl.utils.import_utils import load_extern_type
```
eric-haibin-lin, line 412: thanks for the PR. I think for extensions it's important to make the interface flexible. So in principle we should not have required argument (or only have one or two required arguments), while the rest of the args can be defined by users themselves. I'd recommend passing the data config directly. verl uses hydra, so at the config level users can already easily extend it with +my_own_config.xxx

For the truncation check, i suggest we move it to the "else" branch and apply the check to RLHFDataset only.

eric-haibin-lin (commented):
thank you! a few comments

verl/trainer/ppo/ray_trainer.py
0x404: Okay, actually we only have one required arguments here, i.e., `data_files`. in the `_initialize_customized_dataset` function, we will automatically check the user's dataset `__init__` function, and extract the user defined required arguments.

For example:
```python
class CustomizedDataset(Dataset):
    def __init__(self, data_files):
        ...
# this will only receive the config.data.train_files/val_files
```

and if user want's the entire config:
```python
class CustomizedDataset(Dataset):
    def __init__(self, data_files, config):
        ...
# here config will receive config.data where user can set the config they want
```

and if user want some filed of this config, let's say `prompt_key` and `filter_overlong_prompts`:
```python
class CustomizedDataset(Dataset):
    def __init__(self, data_files, prompt_key, filter_overlong_prompts):
        ...
# here, prompt_key will receive config.data.prompt_key and 
# filter_overlong_prompts will receive config.data.filter_overlong_prompts
```

> For the truncation check, i suggest we move it to the "else" branch and apply the check to RLHFDataset only.

great idea, I'll move this check latter

0x404: Hi @eric-haibin-lin, I add a document to explain the usage of the customized dataset. Any advice would be appreciated.

eric-haibin-lin: Thanks for the update, much appreciated. I think it's ok to break the API of RLHFDataset with 
```
RLHFDataset(data_files or parquet_files, config)
```
and we do not need to implement `_initialize_customized_dataset`. what do you think?

0x404: I'm OK with both and tend to prefer the current implementation, since this implementation gives users more flexibility and is compatible with:
```
RLHFDataset(data_files or parquet_files, config)
```

I have one question: the customized dataset may need a tokenizer and processor (for multimodal) and if we only allow using config, then the dataset itself should create a tokenizer, and we also maintain a tokenizer in the trainer. This feels somewhat weird and may sometimes cause inconsistencies, since our trainer's tokenizer may have [been corrected](https://github.com/volcengine/verl/blob/6efa0181faf5a7149a3950038f5c7c157d5c1016/verl/utils/tokenizer.py#L35).

Maybe I can refactor and move `_initialize_customized_dataset` into dataset utils to be trainer-independent. What do you think?

eric-haibin-lin: you're right, tokenizer should be a required argument, too. if we do `RLHFDataset(data_files, tokenizer, config)` then there's no need for _initialize_customized_dataset except for the function call with `load_extern_type`?

0x404: Yes, if we mandate that `data_files`, `tokenizer`, and `config` are required and only required in the customized dataset, then we can simply initialize the custom dataset without `_initialize_customized_dataset`.

In this way, similar to tokenizer, the processor should also be required, so we need to modify custom dataset to have a processor argument. However, the custom dataset may not need the processor (if not multi modal), and having an unused argument makes it a bit weird and not very elegant.

Therefore, a simple solution is that we have some args that we are going to use to initialize the custom dataset: `data_files`, `tokenizer`, `processor`, and `config`. We then inspect the custom dataset's `__init__` function to see what arguments it wants, and we pass only the wanted args to initialize the custom dataset. This is the motivation for why I made this `_initialize_customized_dataset` func. What do you think?

eric-haibin-lin: i do not think inspecting `__init__` is necessary. passing the entire config object sounds more flexible. 
and yes, processor could be included as well.

0x404:
HI @eric-haibin-lin, as discussed, I have break RLHFDataset interface into:
```
RLHFDataset(
    data_files: Union[str, List[str]],
    tokenizer: PreTrainedTokenizer,
    config: DictConfig,
    processor: Optional[ProcessorMixin] = None
)
```
and made the corresponding changes. please have a check when you have time.

verl/trainer/ppo/ray_trainer.py
```diff
@@ -437,19 +441,12 @@ def _create_dataloader(self):
...
-                                       prompt_key=self.config.data.prompt_key,
-                                       image_key=self.config.data.get('image_key', 'images'),
-                                       max_prompt_length=self.config.data.max_prompt_length,
-                                       return_raw_chat=self.config.data.get('return_raw_chat', False),
-                                       truncation=self.config.data.get('truncation', 'error'),
-                                       filter_overlong_prompts=self.config.data.filter_overlong_prompts,
-                                       num_workers=self.config.data.get('filter_overlong_prompts_workers', None))
-        assert self.val_dataset.truncation == self.config.data.get(
-            'truncation', 'error'
-        ), f'dataset truncation {self.val_dataset.truncation} must be the same as config {self.config.data.get("truncation", "error")}'
```
0x404, line 452: we don't need this truncation check anymore since the dataset is initialized from config, the `dataset.truncation` should always equal to `cofnig.data.truncation`.

docs/workers/ray_trainer.rst
```diff
@@ -21,7 +21,7 @@ then tokenize.
 
 .. code:: python
 
-   self.train_dataset = RLHFDataset(parquet_files=self.config.data.train_files,
+   self.train_dataset = RLHFDataset(data_files=self.config.data.train_files,
                                        tokenizer=self.tokenizer,
```
eric-haibin-lin, line 25: could you update the example in this doc as well?

eric-haibin-lin (approved):
Thanks for your patience and the contribution!

docs/workers/ray_trainer.rst
0x404: Sure, I have updated the example and fixed ci falied due to the breaking change of RLHFDataset. please have a check :)

eric-haibin-lin (approved):

Issue #903, opened 2025-04-03
model_merger.py fails due to type mismatch


XiaofengLin7:
Thanks author for the great work. I have tried to convert the checkpoint to huggingface format to further evaluate the model's performance. However when I tried to run the following commands

```
checkpoint_dir="./checkpoints/REIL/sokoban-rl-exp-0.5b-0.005beta-logic-with-kl/global_step_1500/actor"
BASE_MODEL="./models/rlft/models--Qwen--Qwen2.5-0.5B-Instruct/snapshots/7ae557604adf67be50417f59c2c2f167def9a775"
target_dir="./checkpoints/REIL/sokoban-rl-exp-0.5b-0.005beta-logic-with-kl/"

python3 reil/scripts/model_merger.py --backend fsdp \
    --hf_model_path $BASE_MODEL \
    --local_dir $checkpoint_dir \
    --target_dir $target_dir
```

model_merger.py is the same as the one in the current repo.
However, it encountered
```
    assert isinstance(weight, torch.distributed._tensor.DTensor)
AssertionError
```
on line 81, the variable weight is torch.Tensor instead of torch.distributed._tensor.DTensor.
What could be wrong? Would love to hear from your advice.

XiaofengLin7:
With further experiments, I found if when saving checkpoint, the number of working gpus is one, weight will be therefore saved as torch.Tensor. If there were multiple working gpus, the type would be torch.distributed._tensor.DTensor. This should necessitate a few more lines on checking data type in the model_merger.py

Altair-Alpha:
+1 on this, same issue when merging a 1.5B model trained with FSDP tp=1. Needs little modification to the script to make it work.

Altair-Alpha:
https://github.com/volcengine/verl/pull/925 made a PR here

Kirrito-k423:
中文：这个问题已经由 #925 修复并合入。原始失败点是 FSDP checkpoint 在 `trainer.n_gpus_per_node=1` 时可能加载为普通 `torch.Tensor`，而旧的 `model_merger.py` 只接受 `DTensor`。当前代码已经在 FSDP merger 中同时处理 `DTensor` 和 non-DTensor：non-DTensor 情况会推断简单的 FSDP mesh，并在合并阶段按普通 tensor shard 拼接。因此我认为这个 issue 可以按已修复关闭。

[...]

PR #925, opened 2025-04-05
fix: support non-DTensor when converting fsdp checkpoints to hf model

Altair-Alpha:
As mentioned in https://github.com/volcengine/verl/issues/903, the model_merger script has some problem when dealing with saved fsdp checkpoint trained with `trainer.n_gpus_per_node=1`.  The loaded `weight` is of type `Tensor` instead of `DTensor`. This PR supported this situation.

### DIFF
--- a/scripts/model_merger.py
+++ b/scripts/model_merger.py
@@ -17,6 +17,7 @@ import re
 import os
 import torch
 import argparse
+import numpy as np
 from transformers import AutoConfig, AutoModelForCausalLM, AutoModelForTokenClassification, AutoModelForVision2Seq
 from concurrent.futures import ThreadPoolExecutor
 from torch.distributed._tensor import DTensor, Shard, Placement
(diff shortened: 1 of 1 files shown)


scripts/model_merger.py
```diff
@@ -212,7 +221,7 @@ def check_megatron_checkpoint_path(model_path):
     return sharded_dirs, tp_size, pp_size
 
 
-def convert_megatron_checkpoints_to_hfmodes():
```
Altair-Alpha, line 215: Also fixing a typo in function name

scripts/model_merger.py
```diff
@@ -85,11 +86,16 @@ def convert_fsdp_checkpoints_to_hfmodels():
                             map_location='cpu')
     pivot_key = sorted(list(state_dict.keys()))[0]
     weight = state_dict[pivot_key]
-    assert isinstance(weight, torch.distributed._tensor.DTensor)
-    # get sharding info
-    device_mesh = weight.device_mesh
-    mesh = device_mesh.mesh
-    mesh_dim_names = device_mesh.mesh_dim_names
+
+    if isinstance(weight, torch.distributed._tensor.DTensor):
```
vermouth1992: In Pytorch 2.6, DTensor is moved to torch.distributed.tensor.DTensor. Would you mind also fix this as well? Thanks!

Altair-Alpha:
@vermouth1992 Thanks for the hint. Seems that torch has moved `DTensor` to `torch.distributed.tensor` instead of `_tensor` starting from 2.5 (ref: https://github.com/pytorch/pytorch/blob/v2.5.0-rc1/torch/distributed/_tensor/__init__.py#L2).

Since the official image of verl still uses torch 2.4, I made a compatible change for both 2.4- and 2.5+ in the latest commit.

vermouth1992 (approved):

PR #927, opened 2025-04-05
[log] fix: validation metrics

tongyx361:
This PR:

1. calculate metrics for "reward" by default if "acc"s are not avaiblable
2. don't calculate majority voting metrics if "pred"s are not available

### DIFF
--- a/verl/trainer/ppo/metric_utils.py
+++ b/verl/trainer/ppo/metric_utils.py
@@ -170,7 +170,7 @@ def compute_throughout_metrics(batch: DataProto, timing_raw: Dict[str, float], n
     }
 
 
-def bootstrap_metric(data: list[dict[str, Any]],
+def bootstrap_metric(data: list[Any],
                      subset_size: int,
                      reduce_fns: list[Callable[[np.ndarray], float]],
                      n_bootstrap: int = 1000,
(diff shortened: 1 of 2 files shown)


hiyouga (approved):

hiyouga:
Let us re-trigger the failed test case

mertunsall:
@tongyx361 would it be possible to pass what metrics to calculate in the logs as a string maybe? in verifiable domains pass@N makes the most sense and you don't necessarily want to see standard deviation all the time.

eric-haibin-lin:
> @tongyx361 would it be possible to pass what metrics to calculate in the logs as a string maybe? in verifiable domains pass@N makes the most sense and you don't necessarily want to see standard deviation all the time.

do you mean that you want to make the existing log more succinct and apply some filters? Or do you mean `maj@{n}` is not enough?

mertunsall:
The first one. For example, users can pass something in the config to disable reporting std or maj or pass. Logging everything by default makes the table cluttered and hard to read.

tongyx361:
> The first one. For example, users can pass something in the config to disable reporting std or maj or pass. Logging everything by default makes the table cluttered and hard to read.

Thanks for your suggestion! This is reasonable. I will open an issue for this first and implement it later.

PR #938, opened 2025-04-06
[fix] eos_mask to response_mask for reinforce++-baseline method

none0663:
[fix] misleading eos_mask->response_mask for reinforce_plus_plus_baseline.


eric-haibin-lin (approved):

eric-haibin-lin:
thanks!

PR #948, opened 2025-04-07
feat: add code-r1

ganler:
More info at: https://github.com/ganler/code-r1/tree/main

This PR is a slightly optimized version compared to Code-R1's 443f8da8058d07051793eff875da117534fd1ba1.

### DIFF
--- a/.github/workflows/e2e_coder1_grpo.yml
+++ b/.github/workflows/e2e_coder1_grpo.yml
@@ -1,0 +1,62 @@
+name: e2e_coder1_grpo
+
+on:
+  # Trigger the workflow on push or pull request,
+  # but only for the main branch
+  push:
+    branches:
+      - main
+      - v0.2.x
+    paths:
+      - "**/*.py"
+      - .github/workflows/e2e_coder1_grpo.yml
+  pull_request:
+    branches:
+      - main
+      - v0.2.x
+    paths:
+      - "**/*.py"
+      - .github/workflows/e2e_coder1_grpo.yml
+      - "tests/e2e/*.sh"
+
+# Cancel jobs on the same ref if a new one is triggered
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}
+
+# Declare permissions just read content.
+permissions:
+  contents: read
+
+jobs:
+  e2e_coder1_grpo:
+    runs-on: [self-hosted, l20-1]
+    timeout-minutes: 40 # Increase this timeout value as needed
+    env:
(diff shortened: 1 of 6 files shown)


.github/workflows/e2e_coder1_grpo.yml
```diff
@@ -0,0 +1,62 @@
...
+      - "**/*.py"
+      - .github/workflows/e2e_coder1_grpo.yml
+  pull_request:
+    branches:
+      - main
+      - v0.2.x
+    paths:
+      - "**/*.py"
+      - .github/workflows/e2e_coder1_grpo.yml
+      - "tests/e2e/*.sh"
```
eric-haibin-lin, line 20: could you limit the test trigger path to include the new script added by you, instead of including all scripts under `e2e`?

eric-haibin-lin: please remove the push triggers and keep the pull_request triggers. Our CI is at capacity at the moment. 
Do you mind also temporarily removing all other yml files so that we can test your new yml for a few rounds.

examples/data_preprocess/coder1.py
```diff
@@ -0,0 +1,314 @@
+# SPDX-FileCopyrightText: (c) iSE UIUC Research Group
```
eric-haibin-lin, line 1: please update license according to `tests/sanity/check_license.py` (or add custom headers)

tests/e2e/run_qwen_coder1_grpo.sh
```diff
@@ -0,0 +1,45 @@
+set -x
+
+export VLLM_ATTENTION_BACKEND=XFORMERS
+
+huggingface-cli download Qwen/Qwen2.5-0.5B --local-dir $HOME/models/Qwen/Qwen2.5-0.5B
+
+python3 -m verl.trainer.main_ppo \
```
eric-haibin-lin, line 7: what's the expected performance with this script? If it's reasonably good we should put a reference example under `example/` folder?

verl/utils/reward_score/coder1/__init__.py
```diff
@@ -0,0 +1,155 @@
...
+    if not validate_response_structure(response):
+        return False, reward_log
+
+    if len(extract_code_from_string(try_extract_answer(response)).strip()) == 0:
+        return False, reward_log
+
+    return True, ""
+
+
+def _check_correctness(solution_str, ground_truth, extra_info) -> Tuple[bool, str]:
```
eric-haibin-lin, line 61: could you add some comments on the high level approach of how correctness is checked?

eric-haibin-lin (changes requested):
Thanks for the contribution! Would you mind sharing what's the comparison between firejail and sandbox fusion if you're aware?

PR #958, opened 2025-04-07
fix: move megatron import into megatron related branch

0x404:
users using fsdp backend may no have megatron installed, directly running this script will lead to an import error.


eric-haibin-lin (approved):
thanks!

Issue #418, opened 2025-02-28
optim.warmup_style do not take effect


wenxueru:
In `verl/workers/fsdp_workers.py`, there are only:
```python 
from verl.utils.torch_functional import get_constant_schedule_with_warmup
critic_lr_scheduler = get_constant_schedule_with_warmup(optimizer=critic_optimizer,
                                                        num_warmup_steps=num_warmup_steps)
```
and 
```python 
from verl.utils.torch_functional import get_constant_schedule_with_warmup
actor_optimizer = optim.AdamW(actor_module_fsdp.parameters(),
                              lr=optim_config.lr,
                              betas=optim_config.get('betas', (0.9, 0.999)),
                              weight_decay=optim_config.get('weight_decay', 1e-2))

total_steps = optim_config.get('total_training_steps', 0)
num_warmup_steps_ratio = optim_config.get('lr_warmup_steps_ratio', 0.)
num_warmup_steps = int(num_warmup_steps_ratio * total_steps)

print(f'Total steps: {total_steps}, num_warmup_steps: {num_warmup_steps}')

actor_lr_scheduler = get_constant_schedule_with_warmup(optimizer=actor_optimizer,
                                                       num_warmup_steps=num_warmup_steps)
```

Even though `ppo_trainer.yaml` have relevant setting:
```yaml
optim:
  lr: 1e-6
  lr_warmup_steps_ratio: 0. # the total steps will be injected during runtime
  min_lr_ratio: null # only useful for warmup with cosine
  warmup_style: constant # select from constant/cosine
  total_training_steps: -1 # must be override by program
```

eric-haibin-lin:
Looks like a bug in our code. Would you mind submitting a bug fix? Thanks!

PR #959, opened 2025-04-07
fix: optim.warmup_style do not take effect (#418)

czczup:
Support to set warmup_style=='cosine'.


czczup:
A simple verification was done, and setting `actor_rollout_ref.actor.optim.warmup_style=cosine` started to take effect.

image

vermouth1992 (approved):

PR #960, opened 2025-04-07
add rollout config swap_space for vllm_rollout

pengsun:
When training big model and/or super-long seq len using vLLM rollout, you may encounter the error
```
...
in _swap_out raise RuntimeError( RuntimeError: Aborted due to the lack of CPU swap space. Please increase the swap space to avoid this error
``` 

(updated)This can be fixed by setting bigger `swap_space` for vLLM.  E.g., in your training bash you can do the
```
...
actor_rollout_ref.rollout.engine_kwargs.swap_space=32 \
...
```
which sets the swap_space to 32GB. Note in most vLLM releases the default value is 4GB.

### DIFF
--- a/verl/trainer/config/ppo_trainer.yaml
+++ b/verl/trainer/config/ppo_trainer.yaml
@@ -103,6 +103,8 @@ actor_rollout_ref:
     do_sample: True
     # number of responses (i.e. num sample times)
     n: 1 # > 1 for grpo
+    engine_kwargs: # inference engine parameters
+      swap_space: null # null means "use the engine default value" (usually 4 GB), setting it to, e.g., 32 means 32 GB
     val_kwargs:
       # sampling parameters for validation
       top_k: -1 # 0 for hf rollout, -1 for vllm rollout
--- a/verl/workers/rollout/vllm_rollout/vllm_rollout.py
+++ b/verl/workers/rollout/vllm_rollout/vllm_rollout.py
@@ -100,7 +100,7 @@ class vLLMRollout(BaseRollout):
                              please increase max_num_batched_tokens or disable chunked prefill')
 
         # copy it to avoid secretly modifying the engine config
-        engine_kwargs = OmegaConf.to_container(deepcopy(config.engine_kwargs))
+        engine_kwargs = {} if 'engine_kwargs' not in config else OmegaConf.to_container(deepcopy(config.engine_kwargs))
         # For each vLLM engine parameter, 
         # - `None` means not setting it, so we pop it, and leave it to vLLM default value 
         #    (which can vary across different vLLM versions);
--- a/verl/workers/rollout/vllm_rollout/vllm_rollout.py
+++ b/verl/workers/rollout/vllm_rollout/vllm_rollout.py
@@ -102,9 +102,9 @@ class vLLMRollout(BaseRollout):
         # copy it to avoid secretly modifying the engine config
         engine_kwargs = OmegaConf.to_container(deepcopy(config.engine_kwargs))
         if engine_kwargs.get('swap_space', None) is None:
-            # vLLM has its own default `swap_space` value, which can change in the future versions.
-            # Not setting it means leave it to vLLM default value, 
-            # while setting it explicitly in config means the swap space in GB you want.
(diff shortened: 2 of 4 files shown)


verl/workers/rollout/vllm_rollout/vllm_rollout.py
```diff
@@ -112,6 +118,7 @@ def __init__(self, actor_module: nn.Module, config: DictConfig, tokenizer, model
             disable_log_stats=config.disable_log_stats,
             max_num_batched_tokens=max_num_batched_tokens,
             enable_chunked_prefill=config.enable_chunked_prefill,
+            **vllm_kwargs
```
eric-haibin-lin: what about adding a `engine_kwargs` field to https://github.com/volcengine/verl/blob/main/verl/trainer/config/ppo_trainer.yaml#L112, such as 
```
engine_kwargs:
    swap_space: 4
```
and then pass the entire engine_kwargs to `LLM(actor_module, ..., enable_chunked_prefill, **engine_kwargs)`

this way users can pass arbitrary new args to vllm via `+actor_ref_rollout.rollout.engine_kwargs.xxx`

eric-haibin-lin: Also, please update the doc in https://github.com/volcengine/verl/blob/main/docs/examples/config.rst

pengsun: Adding `engine_kwargs` and passing `**engine_kwags` looks elegant, will update it. 

But I'm not sure whether we should explicitly write `swap_space: 4` as default value in the verl config. In current vLLM, the default value is indeed '4', but it can change in future vLLM versions, and a verl user may just want to use as is the vLLM default by not setting it at all...

Let me know how do you think we should modify the code here. In particular, if you feel okay to explicitly write 4 as default swap_space value in verl config, then I can do it; Otherwise, we can write something like
```
engine_kwargs:
    swap_space: null
```
and in the py code delete `swap_space` from `engine_kwargs` so as to strictly use the vLLM default in any version.

pengsun: I've updated the code and doc, let me know if it looks better or needs more editing :)

verl/workers/rollout/vllm_rollout/vllm_rollout.py
```diff
@@ -98,6 +99,13 @@ def __init__(self, actor_module: nn.Module, config: DictConfig, tokenizer, model
             raise ValueError('Enable chunked prefill, max_num_batched_tokens is smaller than max_model_len, \
                              please increase max_num_batched_tokens or disable chunked prefill')
 
+        # copy it to avoid secretly modifying the engine config
+        engine_kwargs = OmegaConf.to_container(deepcopy(config.engine_kwargs))
+        if engine_kwargs.get('swap_space', None) is None:
+            # vLLM has its own default `swap_space` value, which can change in the future versions.
```
eric-haibin-lin: shall we loop over all keys and pop keys with None values?

verl/trainer/config/ppo_trainer.yaml
```diff
@@ -103,6 +103,8 @@ actor_rollout_ref:
     do_sample: True
     # number of responses (i.e. num sample times)
     n: 1 # > 1 for grpo
+    engine_kwargs: # inference engine parameters
+      swap_space: null # null means "use the engine default value" (usually 4 GB), setting it to, e.g., 32 means 32 GB
```
eric-haibin-lin, line 107: there's also megatron_ppo_trainer

verl/workers/rollout/vllm_rollout/vllm_rollout.py
pengsun: Sure, even better! Code updated.

verl/trainer/config/ppo_trainer.yaml
pengsun: done.

eric-haibin-lin:
i've kicked off CI. please sign CLA

pengsun:
> i've kicked off CI. please sign CLA

* add a commit for initial empty `engine_kwargs` in case it is missing the yaml config
* yapf all files

eric-haibin-lin (approved):

PR #970, opened 2025-04-08
[Mcore] context parallel

ISEEKYAN:
support context parallel for mcore backend.
Changes on:
* configs
* model loader
* checkpint
* single control dispatcher
* forward preprocess and postprocess

### DIFF
--- a/verl/utils/megatron_utils.py
+++ b/verl/utils/megatron_utils.py
@@ -26,6 +26,7 @@ from megatron.core.distributed import DistributedDataParallelConfig
 from megatron.core.enums import ModelType
 from megatron.core.optimizer import OptimizerConfig
 from megatron.core.transformer import TransformerConfig
+from megatron.core.transformer.enums import AttnBackend
 from megatron.core.transformer.module import Float16Module
 from megatron.core.utils import get_attr_wrapped_model
 from omegaconf import DictConfig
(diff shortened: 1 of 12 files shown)


ccclyu:
thanks a ton for quick support! May I know whether you have done some benchmarking or testing of training efficiency upon the context parallel?

ISEEKYAN:
> thanks a ton for quick support! May I know whether you have done some benchmarking or testing of training efficiency upon the context parallel?

I tried with 1 node with 8 H100, comparing tp4dp2cp1 with tp4dp1cp2. cp2(gray line) is slower than cp1 in this test. The result is reasonable because it is not a memory limited situation, with less data parallel and more communication. CP would be useful when the sequence length is larger, but so far, I have only focused on whether the functions have been implemented, and I haven't had the time to pay attention to further performance testing.

截屏2025-04-08 15 09 42

verl/utils/megatron_utils.py
```diff
@@ -285,8 +285,10 @@ def get_optimizer_checkpoint_path(checkpoint_path, use_distributed_optimizer=Tru
         return os.path.join(checkpoint_path, "optim", "optim.pt")
     pp_rank = mpu.get_pipeline_model_parallel_rank()
     tp_rank = mpu.get_tensor_model_parallel_rank()
+    cp_rank = mpu.get_context_parallel_rank()
+    dp_rank = mpu.get_data_parallel_rank()
     #TODO: support ep
-    return os.path.join(checkpoint_path, f"optim", f"distrib_optim_pp{pp_rank}_tp{tp_rank}.pt")
+    return os.path.join(checkpoint_path, f"optim", f"distrib_optim_pp{pp_rank}_tp{tp_rank}_cp{cp_rank}_dp{dp_rank}.pt")
```
ISEEKYAN, line 291: @ETOgaosion due to the optimizer states are distributed across all gpus, the dp rank also should be saved separately.

ETOgaosion: Get~ Later will sync to the doc.

PR #976, opened 2025-04-08
[tuning] docs: record the resource requirements for 70b model

BearBiscuit05:


eric-haibin-lin (approved):

docs/perf/device_tuning.rst
```diff
@@ -0,0 +1,106 @@
+Resource Needed for verl RL
```
eric-haibin-lin, line 1: we should add this page to index.rst as well

PR #979, opened 2025-04-08
misc: update kl penaty names

richardodliu:
Rename the KL penalty parameter to be consistent with common sense

### DIFF
--- a/verl/trainer/ppo/core_algos.py
+++ b/verl/trainer/ppo/core_algos.py
@@ -449,7 +449,7 @@ def compute_value_loss(vpreds, returns, values, response_mask, cliprange_value):
 
 def kl_penalty(logprob: torch.FloatTensor, ref_logprob: torch.FloatTensor, kl_penalty) -> torch.FloatTensor:
     """Compute KL divergence given logprob and ref_logprob.
-    Copied from https://github.com/huggingface/trl/blob/main/trl/trainer/ppo_trainer.py#L1104
+    reference from https://github.com/OpenRLHF/OpenRLHF/blob/main/openrlhf/models/utils.py#L7
 
     Args:
         logprob:
--- a/verl/trainer/ppo/core_algos.py
+++ b/verl/trainer/ppo/core_algos.py
@@ -458,18 +458,18 @@ def kl_penalty(logprob: torch.FloatTensor, ref_logprob: torch.FloatTensor, kl_pe
     Returns:
 
     """
-    if kl_penalty == "kl":
+    if kl_penalty == "k1":
         return logprob - ref_logprob
 
     if kl_penalty == "abs":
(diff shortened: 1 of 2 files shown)


vermouth1992:
Could you point the reference to the common sense?

vermouth1992:
Nice, could you add a reference of naming in the code? Thanks!

eric-haibin-lin (commented):
Thank you for the PR! We'd like to keep backward compatibility. Could you also add the checks for the old name as well. We can update example and docs with the better name to guide new users

verl/trainer/ppo/core_algos.py
```diff
@@ -449,7 +449,7 @@ def compute_value_loss(vpreds, returns, values, response_mask, cliprange_value):
 
 def kl_penalty(logprob: torch.FloatTensor, ref_logprob: torch.FloatTensor, kl_penalty) -> torch.FloatTensor:
     """Compute KL divergence given logprob and ref_logprob.
-    Copied from https://github.com/huggingface/trl/blob/main/trl/trainer/ppo_trainer.py#L1104
+    reference from https://github.com/OpenRLHF/OpenRLHF/blob/main/openrlhf/models/utils.py#L7
```
hiyouga, line 452: this function originated from https://github.com/huggingface/trl/blob/v0.11.0/trl/trainer/ppo_trainer.py#L1150-L1164

richardodliu: Yes， but these three methods are from joschu's blog.  Trl named these methods in a wrong way. I am trying to correct them.

verl/trainer/ppo/core_algos.py
```diff
@@ -458,18 +458,18 @@ def kl_penalty(logprob: torch.FloatTensor, ref_logprob: torch.FloatTensor, kl_pe
     Returns:
```
eric-haibin-lin, line 458: please update the list of values in `docs/examples/config.rst`.  remove the old names in the doc since we do not want others to use them. 

also please search in the codebase for existing values used in scripts. for example:
- examples/reinforce_plus_plus_trainer/run_qwen2-7b_math_rf.sh 
- examples/reinforce_plus_plus_trainer/run_qwen2-7b_math_rf_baseline.sh

verl/trainer/ppo/core_algos.py
```diff
@@ -458,18 +458,18 @@ def kl_penalty(logprob: torch.FloatTensor, ref_logprob: torch.FloatTensor, kl_pe
...
 
     """
-    if kl_penalty == "kl":
+    if kl_penalty == "k1":
```
eric-haibin-lin, line 461: we can simply do `if kl_penalty in ('kl', 'k1'):`
and then remove the verbose changes in verl/trainer/main_ppo.py

PR #984, opened 2025-04-09
fix: replace '@' with '_at_' in metric names to comply with MLflow naming constraints

corgilee:
Fix MLflow metric name errors by replacing '@' with '_at_' during MLflow logging

MLflow rejects metric names with '@' as below

`mlflow.exceptions.RestException: INVALID_PARAMETER_VALUE: Invalid metric name: 'val-aux/semantic_matching/reward/mean@1'. Names may only contain alphanumerics, underscores (_), dashes (-), periods (.), spaces ( ), and slashes (/).`

### DIFF
--- a/verl/utils/tracking.py
+++ b/verl/utils/tracking.py
@@ -139,6 +139,7 @@ class _MlflowLoggingAdapter:
 
     def log(self, data, step):
         import mlflow
+        results = {k.replace('@', '_at_'): v for k, v in data.items()}
         mlflow.log_metrics(metrics=data, step=step)
 
 


vermouth1992:
Could you add mlflow support for special characters in the mlflow logger. Maybe do a string replace before logging to mlflow?

eric-haibin-lin (dismissed):

eric-haibin-lin:
> Could you add mlflow support for special characters in the mlflow logger. Maybe do a string replace before logging to mlflow?

+1

ccclyu:
+1  string replace as the following way. 

https://github.com/volcengine/verl/blob/eda9f0e9be5ee841940205826db9c856cafbb3c8/verl/utils/tracking.py#L142C16-L142C27        

```
        results = {k.replace('@', '_at_'): v for k, v in data.items()}
        mlflow.log_metrics(metrics=results, step=step)
```

corgilee:
@eric-haibin-lin @vermouth1992 @ccclyu Thanks for the comments, revert the previous commit and "add mlflow support for special characters in the mlflow logger" as suggested

vermouth1992 (approved):

ccclyu:
@vermouth1992  seems merged commit has bug as I just commented.

vermouth1992:
> @vermouth1992 seems merged commit has bug as I just commented.

woops, what's the bug? It's just one line change

ccclyu:
https://github.com/volcengine/verl/blob/0a4f4b3cc12d2623313324c03ff5fd4f62aced95/verl/utils/tracking.py#L142-L143

In line 143, it should be `metrics=results`.

vermouth1992:
https://github.com/volcengine/verl/pull/1073 I create a new PR to address this

PR #987, opened 2025-04-09
[vllm] fix oom when vllm wakeup (vllm >=0.8.3)

BearBiscuit05:
This is a memory optimization method implemented based on this [fix](https://github.com/vllm-project/vllm/pull/15500). I just successfully ran a 72B model on 8*H800 cards. Before the fix, I would encounter an OOM issue. Please note that this fix is only effective for vLLM >= 0.8.3.

### DIFF
--- a/scripts/model_merger.py
+++ b/scripts/model_merger.py
@@ -28,7 +28,6 @@ try:
 except ImportError:
     from torch.distributed._tensor import DTensor
 
-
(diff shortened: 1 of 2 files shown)


hiyouga (approved):

JingbiaoMei:
Hi,
Could you share some of your scripts where you managed to run 72B with 8 H800?

Many Thanks.

Issue #742, opened 2025-03-25
[Feature Request]Support Dr. GRPO for Unbiased Optimization in RL Training


Peng-YM:
## Paper
[Dr. GRPO Paper](https://github.com/sail-sg/understand-r1-zero)

## Motivation/Benefits

- Fixes optimization bias while maintaining reasoning performance
- Reduces average incorrect response length by 38% (Fig.5 in paper)
- Backward-compatible with existing GRPO workflows

Peng-YM:
See Also: [trl@huggingface's PR](https://github.com/huggingface/trl/pull/3135), we could also add an option `scale_rewards=False` for GRPO algorithm.

eric-haibin-lin:
PR is welcome!

sunjin-k:
`scale_grpo_adv` is the first part of Dr.GRPO.

The second part is already implemented through `loss_agg_mode == "seq-mean-token-sum"`, but in the paper's implementation they multiply it by a constant factor `MAX_TOKENS` (`data.max_response_length` in verl), probably to be able to reuse hyperparameters between GRPO to some extent.

Should we should add `... / max_response_length` to `loss_agg_mode == "seq-mean-token-sum"`? 
Or add another option for that?




![Image](https://github.com/user-attachments/assets/f46c2c7d-33e5-4680-9695-f3d57bc18d6e)

Peng-YM:
Issue #990 has implemented the Dr. GRPO algorithm, many thanks for your awesome work! @sunjin-k @eric-haibin-lin

PR #990, opened 2025-04-09
Add scale_grpo_adv option 

sunjin-k:
https://github.com/volcengine/verl/issues/742

- Add an option for disabling standard-deviation normalization of advantages in GRPO.
- This completes one out of two algorithmic changes made by Dr.GRPO to GRPO, the other one being the removal of sequence-length averaging during loss aggregation.

### DIFF
--- a/examples/split_placement/split_monkey_patch.py
+++ b/examples/split_placement/split_monkey_patch.py
@@ -141,11 +141,13 @@ def fit(self):
                         batch.batch['token_level_rewards'] = batch.batch['token_level_scores']
 
                     # compute advantages, executed on the driver process
+                    scale_grpo_adv = self.config.algorithm.get('scale_grpo_adv', True)
                     batch = compute_advantage(batch,
                                               adv_estimator=self.config.algorithm.adv_estimator,
                                               gamma=self.config.algorithm.gamma,
                                               lam=self.config.algorithm.lam,
-                                              num_repeat=self.config.actor_rollout_ref.rollout.n)
+                                              num_repeat=self.config.actor_rollout_ref.rollout.n,
+                                              scale_grpo_adv=scale_grpo_adv)
 
                 # update critic
                 if self.use_critic:
--- a/recipe/dapo/src/dapo_ray_trainer.py
+++ b/recipe/dapo/src/dapo_ray_trainer.py
@@ -242,11 +242,13 @@ class RayDAPOTrainer(RayPPOTrainer):
 
                     with _timer('adv', timing_raw):
                         # compute advantages, executed on the driver process
+                        scale_grpo_adv = self.config.algorithm.get('scale_grpo_adv', True)
                         batch = compute_advantage(batch,
                                                   adv_estimator=self.config.algorithm.adv_estimator,
                                                   gamma=self.config.algorithm.gamma,
                                                   lam=self.config.algorithm.lam,
-                                                  num_repeat=self.config.actor_rollout_ref.rollout.n)
+                                                  num_repeat=self.config.actor_rollout_ref.rollout.n,
+                                                  scale_grpo_adv=scale_grpo_adv)
 
                     # update critic
                     if self.use_critic:
--- a/verl/trainer/ppo/core_algos.py
+++ b/verl/trainer/ppo/core_algos.py
@@ -107,7 +107,8 @@ def compute_gae_advantage_return(token_level_rewards: torch.Tensor, values: torc
 def compute_grpo_outcome_advantage(token_level_rewards: torch.Tensor,
                                    response_mask: torch.Tensor,
                                    index: np.ndarray,
-                                   epsilon: float = 1e-6):
+                                   epsilon: float = 1e-6,
+                                   scale_grpo_adv: str = True):
     """
     Compute advantage for GRPO, operating only on Outcome reward 
     (with only one scalar reward for each response).
@@ -116,6 +117,10 @@ def compute_grpo_outcome_advantage(token_level_rewards: torch.Tensor,
             shape: (bs, response_length)
         response_mask: `(torch.Tensor)`
             shape: (bs, response_length)
+        scale_grpo_adv: (bool)
+            whether to scale the GRPO advantage. 
+            If True, the advantage is scaled by the std, as in the original GRPO.
+            If False, the advantage is not scaled, as in Dr.GRPO (https://arxiv.org/abs/2503.20783).
     
     Returns:
         advantages: `(torch.Tensor)`
@@ -143,7 +148,10 @@ def compute_grpo_outcome_advantage(token_level_rewards: torch.Tensor,
             else:
                 raise ValueError(f"no score in prompt index: {idx}")
         for i in range(bsz):
-            scores[i] = (scores[i] - id2mean[index[i]]) / (id2std[index[i]] + epsilon)
+            if scale_grpo_adv:
+                scores[i] = (scores[i] - id2mean[index[i]]) / (id2std[index[i]] + epsilon)
+            else:
+                scores[i] = (scores[i] - id2mean[index[i]])
         scores = scores.unsqueeze(-1) * response_mask
 
     return scores, scores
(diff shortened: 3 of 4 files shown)


eric-haibin-lin (commented):
@tongyx361 does the naming of the option look good to you?

lkevinzc:
The scaling factor or original GRPO consists of two parts (please see the right-bottom equation in the Fig). It would be better if we could remove the bias altogether, by a single flag (argument), such that we could compare Dr. GRPO with GRPO fairly.

image

sunjin-k:
@lkevinzc

Oh, as I mentioned in https://github.com/volcengine/verl/issues/742, the removal of the $\frac{1}{|o_i|}$ term is already implemented in verl by the `loss_agg_mode == "seq-mean-token-sum"` option. 

Although that implementation is different from the original paper's implementation by a constant factor, it is equivalent, as in you just need to input different kl and stepsize hyperparameters.
(The authors multiply the loss/advantage by $\frac{1}{|MAX \textunderscore SEQ \textunderscore GENERATION \textunderscore LENGTH|}$, whereas verl multiplies by $\frac{1}{|BATCH \textunderscore SIZE|}$ (the *mean* part of seq_mean_token_sum). Perhaps their choice was for numerical stability or so they could be able to use similar hyperparameters with GRPO)

Please correct me if wrong

sunjin-k:
So, by making a `DrGRPO` flag, it will overlap with `loss_agg_mode`.

Also considering that DAPO wasn't implemented as a single flag but as a combination of multiple (`use_kl_loss`, `loss_agg_mode`, etc), I made Dr. GRPO like this.

But I do agree that as new RL algorithms keep appearing, there will be more and more `algorithm` options to keep track of, and perhaps we can give users both the option to use a simple flag for a single algorithm, as well as more fine-grained arguments.

lkevinzc:
@sunjin-k 

> Although that implementation is different from the original paper's implementation by a constant factor, it is equivalent, as in you just need to input different kl and stepsize hyperparameters.

I have tested both summed loss and summed loss / max_gen_len; I would suggest using the latter to keep the grad norm small (not being clipped). I agree that theoretically we can reduce lr to make them equivalent, but in practice, say, if we use lr / 4k (4k gen length), this will lead to 1e-10 scale of lr, which I am afraid that it has numerical problems.

---

> Also considering the fact that DAPO wasn't implemented as a single flag but as a combination of multiple (use_kl_loss, loss_agg_mode, etc), and also comments by others, I decided to make Dr. GRPO like this.

If currently we are not ready to use a single flag for each algo, I would suggest that we can use a new `loss_agg_mode`, e.g., `seq-mean-token-sum-norm`, which can be implemented as `seq_losses = torch.sum(loss_mat * loss_mask, dim=-1) / loss_mask.shape[-1]`, where I assume all `loss_mask` has the same dimension of the last axis.

---

Finally, is there any chance you could modify the README to tell users that Dr. GRPO is now an option (similar to other algos listed [here](https://github.com/volcengine/verl?tab=readme-ov-file#key-features)), and maybe provide an example (to specify the flags) with the link to that example (maybe the easiest way is to add a new drgrpo yaml, while keep the original dapo yaml unchanged)?

---

Thanks again for your time to make this PR happen.

tongyx361:
Thanks a lot for all the awesome discussions! Here are my thoughts:

1. About the option name, would `norm_adv_by_std_in_grpo` be better since it is clearer about the semantic of the bool value, though a little more cumbersome.
2. The suggestion of "seq-mean-token-sum-norm" considering the numerical precision by @lkevinzc sounds great! About the assumption that "all loss_mask has the same dimension of the last axis", it is true for now, but maybe it would be better to use `prompt_len + response_len` instead of `loss_mask.shape[-1]` in case of possible introduction of different loss_mask shape in the future?
3. About the algorithm-related options, I think, for a neutral framework like verl, it might be better to provide options to manipulate basic components like advantage and loss and implement algorithms as examples of option composition or "frontend options" which might be parsed into basic options. But indeed, verl is in a mixed status now. What are your opinions about this?

lkevinzc:
Thanks @tongyx361 for your review and thoughtful input!

1. `norm_adv_by_std_in_grpo` sounds perfect to me because it distinguishes from another scaling factor which is by length.

2. I thinking `prompt_len + response_len` is not equivalent to a constant normalizer - this summation is still a length-dependent variable. Here we'd better to have a constant normalizer, and the ideal way is to let users to define it but defaults to MAX_GEN_LEN (please see this [tweet](https://x.com/QGallouedec/status/1908741708021457357) on how the constant scale could affect performance - basically it controls the update strength). I did not suggest user-defined normalizer in my previous comment because it will introduce API changes of the [`agg_loss` function](https://github.com/volcengine/verl/blob/d9df9bbb5fc8e07a3215f788b08fee18f7fb764d/verl/trainer/ppo/core_algos.py#L309), and thus I proposed to exploit the shape to imply the **constant** context length. But please take user-defined normalizer as consideration if possible :)

3. I totally agree and support that verl should be a neutral framework. In that case, maybe for each algorithm, we could **provide self-contained example (and highlight the algo-specific flags)** while keeping all the options properly designed for each component. If we do so, users also have the flexibility to combine options in other ways and possible find better combination!

sunjin-k:
Another way is completely removing `loss_agg_mode` and replacing it with two arguments

$$ \frac{1}{\square} \sum_{b} \left( \frac{1}{\triangle} \sum_{t} \text{ClippedAdv} \right)$$

and we can get GRPO, DAPO, DrGRPO, etc by just specifying $\square$ and $\triangle$.

But it 
- might be complicated for users
- might cause numerical inconsistency (eg if in low precision) compared to `torch.mean`

But if you're alright with `norm_adv_by_std_in_grpo` and `seq-mean-token-sum-norm`, I'll implement it as soon as everyone agrees on the best way to calculate MAX_GEN_LEN

sunjin-k:
Yup, I've 
- renamed `scale_grpo_adv` to `norm_adv_by_std_in_grpo`
- added `seq-mean-token-sum-norm`
- added recipe with config instructions and short summary of paper to `/recipe/drgrpo`

lkevinzc:
Looks perfect to me, thanks again for your contribution! @sunjin-k

lkevinzc:
Hi @eric-haibin-lin , could you please have a review and hopefully we could merge this? I'm looking forward to running DrGRPO experiments with verl!

eric-haibin-lin:
@sunjin-k could you run scripts/format.sh to fix lint? thanks

mertunsall:
Looking forward to this!

eric-haibin-lin:
sorry we just made code format more strict 😂 
would you mind fixing lint according to https://github.com/volcengine/verl?tab=readme-ov-file#code-linting-and-formatting

eric-haibin-lin:
the CI warning should be dismissed after https://github.com/volcengine/verl/pull/1168

eric-haibin-lin (approved):

eric-haibin-lin:
Thank you all for the discussions and patience to address all the comments. PR is now merged~

lwp-hhxx:
> @lkevinzc
> 
> Oh, as I mentioned in #742, the removal of the 1 | o i | term is already implemented in verl by the `loss_agg_mode == "seq-mean-token-sum"` option.
> 
> Although that implementation is different from the original paper's implementation by a constant factor, it is equivalent, as in you just need to input different kl and stepsize hyperparameters. (The authors multiply the loss/advantage by 1 | M A X _ S E Q _ G E N E R A T I O N _ L E N G T H | , whereas verl multiplies by 1 | B A T C H _ S I Z E | (the _mean_ part of seq_mean_token_sum). Perhaps their choice was for numerical stability or so they could be able to use similar hyperparameters with GRPO)
> 
> Please correct me if wrong

I think the "seq-mean-token-sum-norm" mode in DRGRPO should be implemented like this:
      seq_losses = torch.sum(loss_mat * loss_mask, dim=-1) / loss_mask.shape[-1]
      loss = torch.mean(seq_losses)  # seq-mean

But in VERL, it is currently written as:

seq_losses = torch.sum(loss_mat * loss_mask, dim=-1)
loss = torch.sum(seq_losses) / loss_mask.shape[-1]  # The divisor

It seems the averaging step is missing. Am I misunderstanding something?

PR #998, opened 2025-04-10
fix: loading HF model in rank0 for mcore megatron model

ann-qin-lu:
[bug fix] Loading 72B Qwen model for mcore megatron is causing OOM. 
Extracted out the HF loading logic to a helper function (and disable `device_map=auto`), and refactored the legacy function `load_megatron_model_weights` and the new function `load_megatron_gptmodel_weights`.  `load_megatron_model_weights` should be deprecated once the the class RewardModelWorker is also migrated to Mcore.

### DIFF
--- a/verl/utils/model.py
+++ b/verl/utils/model.py
@@ -277,12 +277,8 @@ def _get_parallel_model_architecture_from_config(config: PretrainedConfig, value
                      f"Supported architectures: {ModelRegistry.get_supported_archs()}")
 
 
-def load_megatron_model_weights(config,
-                                model_config,
-                                parallel_model,
-                                params_dtype,
-                                is_value_model=False,
-                                local_cache_path='~/.cache/verl/rlhf'):
+def _load_hf_model(config, model_config, is_value_model, local_cache_path):
+    """Helper function containing the loading hf model logic"""
     assert hasattr(model_config, "architectures"), "architectures cannot be empty when load weight!"
     architectures = getattr(model_config, "architectures", [])
     local_cache_path = os.path.expanduser(local_cache_path)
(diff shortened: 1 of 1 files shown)


verl/utils/model.py
```diff
@@ -303,19 +299,38 @@ def load_megatron_model_weights(config,
...
             is_value_model = True
         else:
-            model = AutoModelForCausalLM.from_pretrained(local_model_path,
-                                                         torch_dtype="auto",
-                                                         device_map="auto",
-                                                         low_cpu_mem_usage=True)
+            model = AutoModelForCausalLM.from_pretrained(
+                local_model_path,
+                torch_dtype="auto",
+                # device_map="auto", # disable auto device_map, the HF weight is only loaded to CPU in root rank
```
ETOgaosion: Seems still find CPU and GPU resources? Because a CI shows that device is `meta`

verl/utils/model.py
```diff
@@ -296,26 +296,46 @@ def load_megatron_model_weights(config,
...
     with init_context(), warnings.catch_warnings():
         warnings.simplefilter("ignore")
+        # TODO: to find a better way to load mistral7b-rm lm_head
         if 'mistral7b-rm' in config.model.path:
             model = MistralForSequenceClassification.from_pretrained(
-                local_model_path, device_map="auto", low_cpu_mem_usage=True)  # use score head instead of lm_head
+                local_model_path,
+                torch_dtype="auto",
+                # device_map="auto",  # disable auto device_map, the HF weight is only loaded to CPU in src_rank
+                # low_cpu_mem_usage=True
```
ccclyu, line 307: can we remove this line for `low_cpu_mem_usage` as we do not use this flag as default?

verl/utils/model.py
```diff
@@ -296,26 +296,46 @@ def load_megatron_model_weights(config,
...
         else:
-            model = AutoModelForCausalLM.from_pretrained(local_model_path,
-                                                         torch_dtype="auto",
-                                                         device_map="auto",
-                                                         low_cpu_mem_usage=True)
+            model = AutoModelForCausalLM.from_pretrained(
+                local_model_path,
+                torch_dtype="auto",
+                # device_map="auto", # disable auto device_map, the HF weight is only loaded to CPU in src_rank
+                # low_cpu_mem_usage=True
```
ccclyu, line 319: the same as Line 310.

ann-qin-lu:
Rebased on the latest main. @ETOgaosion if you can help review the changes? Thanks!

verl/utils/model.py
ann-qin-lu: I intentionally leave it here. If someone adds this flag back, it might lead to error

verl/utils/model.py
ann-qin-lu: one is MistralForSequenceClassification, and this line is AutoModelForCausalLM

verl/utils/model.py
ann-qin-lu: meta device contains no actual data per https://pytorch.org/docs/stable/meta.html. this is more like a placeholder when we init the model with  [init_empty_weights](https://huggingface.co/docs/accelerate/v0.11.0/en/big_modeling#accelerate.init_empty_weights)

Issue #523, opened 2025-03-10
Failed to register worker to Raylet on slurm cluster.


yiding-w:
I am trying to run grpo training for 0.5B model on a slurm cluster with 2*A100-80G, I didn't use the docker image. And get such error:

> + VLLM_ATTENTION_BACKEND=XFORMERS
+ python3 -m verl.trainer.main_ppo algorithm.adv_estimator=grpo data.train_files=/home/mrm2vx/data/gsm8k/train.parquet data.val_files=/home/mrm2vx/data/gsm8k/test.parquet data.train_batch_size=256 data.max_prompt_length=512 data.max_response_length=256 actor_rollout_ref.model.path=Qwen/Qwen2.5-0.5B-Instruct actor_rollout_ref.actor.optim.lr=1e-6 actor_rollout_ref.model.use_remove_padding=True actor_rollout_ref.actor.ppo_mini_batch_size=64 actor_rollout_ref.actor.ppo_micro_batch_size_per_gpu=2 actor_rollout_ref.actor.use_kl_loss=True actor_rollout_ref.actor.kl_loss_coef=0.001 actor_rollout_ref.actor.kl_loss_type=low_var_kl actor_rollout_ref.model.enable_gradient_checkpointing=True actor_rollout_ref.actor.fsdp_config.param_offload=False actor_rollout_ref.actor.fsdp_config.optimizer_offload=False actor_rollout_ref.rollout.log_prob_micro_batch_size_per_gpu=2 actor_rollout_ref.rollout.tensor_model_parallel_size=1 actor_rollout_ref.rollout.name=vllm actor_rollout_ref.rollout.gpu_memory_utilization=0.4 actor_rollout_ref.rollout.n=3 actor_rollout_ref.ref.log_prob_micro_batch_size_per_gpu=2 actor_rollout_ref.ref.fsdp_config.param_offload=True algorithm.kl_ctrl.kl_coef=0.001 trainer.critic_warmup=0 'trainer.logger=[console,wandb]' trainer.project_name=verl_grpo_example_gsm8k trainer.experiment_name=qwen2.5_0.5b_function_rm trainer.n_gpus_per_node=2 trainer.nnodes=1 trainer.save_freq=-1 trainer.test_freq=5 trainer.total_epochs=1
2025-03-10 00:05:24,602	INFO worker.py:1841 -- Started a local Ray instance.
[...]

PR #1009, opened 2025-04-10
[ray] fix: ray hang due to num_cpus

tongyx361:
Fixing #523 according to https://github.com/volcengine/verl/issues/523#issuecomment-2723652147

~Concern: will `num_cpus=1` limit the performance of the cluster scheduler?~

See https://github.com/volcengine/verl/pull/1009#issuecomment-2807943422 for the final design.


hongpeng-guo:
@tongyx361 @eric-haibin-lin  I think #523 is a unique problem to SLURM with Ray. This fix may introduce negative side effect for non-SLURM ray scenario.

By default `ray.init(num_cpus=None)` will try to use every possible cpu cores on an machine. This also defines the number of concurrency a ray node can achieve.

The cause of  #523 is due to some SLURM setting. I.e., SLURM added some constraints on how the CPUs are visible by on job. While `ray.init()` tries to launch as many core processes as the number of CPU cores of the machine. Some constraints of SLURM restricts the `core-workers` seeing the raylet process on the node, leading to the problem.

For non-SLURM scenarios, it's fine and actually works best of performance to use `ray.init()` without specifying `num_cpus`, because it utilize the maximum possible cpus on the mahcine.

The SLURM [support](https://docs.ray.io/en/latest/cluster/vms/user-guides/community/slurm.html) on ray is community built and non-official. I would suggest close this PR as it's bad for more general verl use case. But instead adding the workaround into the issue and convert it as a discussion for later users.

verl/trainer/main_generation.py
```diff
@@ -45,7 +45,7 @@ def run_generation(config) -> None:
 
     if not ray.is_initialized():
         # this is for local ray cluster
-        ray.init(runtime_env={'env_vars': {'TOKENIZERS_PARALLELISM': 'true', 'NCCL_DEBUG': 'WARN'}})
+        ray.init(runtime_env={'env_vars': {'TOKENIZERS_PARALLELISM': 'true', 'NCCL_DEBUG': 'WARN'}}, num_cpus=1)
```
eric-haibin-lin: every ray remote call is scheduled by raylet. can we make it an option that does not affect the default behavior. if slurm users come across this issue, we can recommend the config in the FAQ page. by default we just set it to null

eric-haibin-lin (changes requested):

tongyx361:
> @tongyx361 @eric-haibin-lin I think #523 is a unique problem to SLURM with Ray. The fix in this PR may introduce negative side effect for non-SLURM ray scenario.
> 
> By default `ray.init(num_cpus=None)` will try to use every possible cpu cores on an machine. This also defines the number of concurrency a ray node can achieve.
> 
> The cause of #523 is due to some SLURM setting. I.e., SLURM added some constraints on how the CPUs are shared on a node. While `ray.init()` tries to launch as many worker processes as the number of CPU cores of the machine. Some constraints of SLURM restricts the `core-workers` seeing the `raylet` process, leading to the problem.
> 
> For non-SLURM scenarios, it's fine and actually works best of performance to use `ray.init()` without specifying `num_cpus`, because it utilizes the maximum possible cpus on the machine.
> 
> The SLURM [support](https://docs.ray.io/en/latest/cluster/vms/user-guides/community/slurm.html) on ray is community built and non-official. It's probably not ideal to include SLURM specify fix into verl, either.
> 
> I would suggest to **close** this PR as it has side effects for more general verl use cases. Instead, we can add the workaround into the issue and convert it as a discussion for later SLURM users.

Thanks for your awesome comment! It resolves my concern in the description. According to this, I think we can set the `num_cpus` as a config term as @eric-haibin-lin suggested, e.g.,

```yaml
ray_init:
    num_cpus: null # `None` means using all CPUs, which might cause hang if limited in systems like SLURM. Please set to a number allowed then.
```

```python
    if not ray.is_initialized():
        # this is for local ray cluster
        ray.init(runtime_env={'env_vars': {'TOKENIZERS_PARALLELISM': 'true', 'NCCL_DEBUG': 'WARN'}}, num_cpus=config.ray_init.num_cpus)
```

hongpeng-guo (approved):

eric-haibin-lin (dismissed):
lgtm. please resolve conflict

eric-haibin-lin (approved):

PR #1026, opened 2025-04-11
[fsdp] feat: support fsdp2 training and inference in fsdp_workers

lxg2015:
# What does this PR do?

This PR supports fsdp2 for fsdp_worker. Torch version 2.4 or higher is required.

# Usage Example

```
sh examples/grpo_trainer/run_qwen2-7b.sh \
    actor_rollout_ref.ref.strategy=fsdp2 \
    actor_rollout_ref.actor.strategy=fsdp2 
```
To save more memory, you can add the parameter below to enable the fsdp2 OffloadPolicy:
``` 
actor_rollout_ref.actor.offload_policy=True  
```
You can see the profile comparison between fsdp1 and fsdp2 here:  https://github.com/volcengine/verl/pull/1026#issuecomment-2824343860

### DIFF
--- a/docs/experiment/ppo.rst
+++ b/docs/experiment/ppo.rst
@@ -27,6 +27,7 @@ NVIDIA GPUs
 .. _Qwen0.5b PRIME Script: https://github.com/volcengine/verl/blob/main/recipe/prime/run_prime_qwen.sh
 .. _Qwen0.5b PRIME Wandb: https://api.wandb.ai/links/zefan-wang-thu-tsinghua-university/rxd1btvb
 .. _Megatron Qwen2 7b GRPO Script with Math and GSM8k: https://github.com/eric-haibin-lin/verl-data/blob/experiments/gsm8k/qwen2-7b_math_megatron.log
+.. _Qwen7b GRPO FSDP2 Script and Logs: https://github.com/eric-haibin-lin/verl-data/blob/experiments/gsm8k/qwen2-7b-fsdp2.log
 
 +----------------------------------+------------------------+------------+-----------------------------------------------------------------------------------------------+
 | Model                            | Method                 | Test score |  Details                                                                                      |
@@ -46,6 +47,8 @@ NVIDIA GPUs
 | deepseek-ai/deepseek-llm-7b-chat | PPO (Megatron)         | 69.5 [1]_  |   `Megatron PPO Command and Logs`_, `Megatron wandb`_                                         |
 +----------------------------------+------------------------+------------+-----------------------------------------------------------------------------------------------+
 | Qwen/Qwen2-7B-Instruct           | GRPO                   | 89         |   `Qwen7b GRPO Script`_                                                                       |
++----------------------------------+------------------------+------------+-----------------------------------------------------------------------------------------------+
+| Qwen/Qwen2-7B-Instruct           | GRPO (FSDP2)           | 89.8       |  `_Qwen7b GRPO FSDP2 Script and Logs`_                                                                       |
 +----------------------------------+------------------------+------------+-----------------------------------------------------------------------------------------------+
 | Qwen/Qwen2-7B-Instruct           | GRPO (Megatron)        | 89.6       |   `Megatron Qwen2 7b GRPO Script with Math and GSM8k`_                                        |
 +----------------------------------+------------------------+------------+-----------------------------------------------------------------------------------------------+
--- a/verl/trainer/config/ppo_trainer.yaml
+++ b/verl/trainer/config/ppo_trainer.yaml
@@ -30,7 +30,7 @@ actor_rollout_ref:
     use_remove_padding: False
     use_liger: False
   actor:
-    strategy: fsdp  # This is for backward-compatibility
+    strategy: fsdp  # [fsdp, fsdp2], This is for backward-compatibility
     ppo_mini_batch_size: 256
     ppo_micro_batch_size: null # will be deprecated, use ppo_micro_batch_size_per_gpu
     ppo_micro_batch_size_per_gpu: null
--- a/verl/trainer/config/ppo_trainer.yaml
+++ b/verl/trainer/config/ppo_trainer.yaml
@@ -67,6 +67,7 @@ actor_rollout_ref:
         min_num_params: 0
       param_offload: False
       optimizer_offload: False
+      offload_policy: False # only for fsdp2, offload param\grad\optimizer during train
       fsdp_size: -1
   ref:
     strategy: fsdp
@@ -129,7 +130,7 @@ actor_rollout_ref:
 
 critic:
   rollout_n: ${actor_rollout_ref.rollout.n}
-  strategy: fsdp
+  strategy: fsdp # [fsdp, fsdp2]
   optim:
     lr: 1e-5
     lr_warmup_steps_ratio: 0.  # the total steps will be injected during runtime
@@ -147,6 +148,7 @@ critic:
     fsdp_config:
       param_offload: False
       optimizer_offload: False
+      offload_policy: False # only for fsdp2, offload param\grad\optimizer during train
       wrap_policy:
         # transformer_layer_cls_to_wrap: None
         min_num_params: 0
--- a/verl/trainer/config/ppo_trainer.yaml
+++ b/verl/trainer/config/ppo_trainer.yaml
@@ -68,11 +68,13 @@ actor_rollout_ref:
       param_offload: False
       optimizer_offload: False
       offload_policy: False # only for fsdp2, offload param\grad\optimizer during train
+      reshard_after_forward: True # only for fsdp2, [True, False, int between 1 and fsdp_size]
       fsdp_size: -1
   ref:
     strategy: fsdp
     fsdp_config:
       param_offload: False
+      reshard_after_forward: True # only for fsdp2, [True, False, int between 1 and fsdp_size]
       wrap_policy:
         # transformer_layer_cls_to_wrap: None
         min_num_params: 0
@@ -149,6 +151,7 @@ critic:
       param_offload: False
       optimizer_offload: False
       offload_policy: False # only for fsdp2, offload param\grad\optimizer during train
+      reshard_after_forward: True # only for fsdp2, [True, False, int between 1 and fsdp_size]
       wrap_policy:
         # transformer_layer_cls_to_wrap: None
         min_num_params: 0
@@ -181,6 +184,7 @@ reward_model:
       wrap_policy:
         min_num_params: 0
       param_offload: False
+      reshard_after_forward: True # only for fsdp2, [True, False, int between 1 and fsdp_size]
       fsdp_size: -1
   micro_batch_size: null # will be deprecated, use micro_batch_size_per_gpu
   micro_batch_size_per_gpu: null # set a number
--- a/verl/utils/fsdp_utils.py
+++ b/verl/utils/fsdp_utils.py
@@ -385,10 +385,50 @@ def fsdp2_sharding_strategy(device_mesh):
     return sharding_strategy
 
 
+def fsdp2_load_full_state_dict(model: torch.nn.Module, full_sd: dict):
+    """ refer accelerate
+    Loads the full state dict (could be only on rank 0) into the sharded model. This is done by broadcasting the
+    parameters from rank 0 to all other ranks. This function modifies the model in-place.
+
+    Args:
+        model (`torch.nn.Module`): The model to load the state dict into
+        full_sd (`dict`): The full state dict to load, can only be on rank 0
+    """
+    from torch.distributed.tensor import distribute_tensor
+
+    sharded_sd = model.state_dict()
+
+    if dist.get_rank() == 0:
+        for (param_name, full_param), sharded_param in zip(full_sd.items(), sharded_sd.values()):
+            full_param = full_param.detach().cuda()
+            mesh = sharded_param.device_mesh
+            dist.broadcast(full_param, src=0, group=mesh.get_group())
+            sharded_tensor = distribute_tensor(full_param, mesh, sharded_param.placements)
+            sharded_sd[param_name] = sharded_tensor
+    else:
+        model.to_empty(device=torch.cuda.current_device())
+        for param_name, sharded_param in sharded_sd.items():
+            full_tensor = torch.empty(sharded_param.size(), device="cuda", dtype=sharded_param.dtype)
+            mesh = sharded_param.device_mesh
+            dist.broadcast(full_tensor, src=0, group=mesh.get_group())
+            sharded_tensor = distribute_tensor(full_tensor, mesh, sharded_param.placements)
+            sharded_sd[param_name] = sharded_tensor
+        
+    model.load_state_dict(sharded_sd)
+
+
+def prepare_for_cpu_offload(model: torch.nn.Module, cpu_offload=None):
+    if cpu_offload:
+        model.to('cpu')
+        for buf in model.buffers():
+            buf.to(torch.cuda.current_device())
+    
+
 def apply_fsdp2(model, fsdp_kwargs):
     '''model: AutoModelForCausalLM
     '''
     assert CPUOffloadPolicy is not None, "PyTorch version >= 2.4 is required for using fully_shard API (FSDP2)"
+    assert hasattr(model.model, 'layers'), "TransformerBlock layer was not found in model.model, please check model structure"
 
     fsdp_mesh = fsdp_kwargs.get('mesh')
 
--- a/verl/utils/fsdp_utils.py
+++ b/verl/utils/fsdp_utils.py
@@ -29,6 +29,8 @@ from torch.distributed.fsdp._runtime_utils import _lazy_init
 from torch.distributed.fsdp.wrap import size_based_auto_wrap_policy, transformer_auto_wrap_policy
 from transformers.trainer_pt_utils import get_module_class_from_name
 from packaging import version
+from torch.distributed.tensor import DTensor
+
 if version.parse(torch.__version__) >= version.parse('2.6'):
     from torch.distributed.fsdp import fully_shard, MixedPrecisionPolicy, FSDPModule, CPUOffloadPolicy
 elif version.parse(torch.__version__) >= version.parse('2.4'):
@@ -139,7 +141,8 @@ def offload_fsdp_model_to_cpu(model: FSDP, empty_cache: bool = True):
 
 @torch.no_grad()
 def offload_fsdp2_model_to_cpu(model, empty_cache: bool = True):
-    model.to('cpu', non_blocking=True)
+    for param in model.parameters():
+        param.data = param.data.to(torch.device('cpu'), non_blocking=True)
     if empty_cache:
         torch.cuda.empty_cache()
 
@@ -164,8 +167,9 @@ def load_fsdp_model_to_gpu(model: FSDP):
 
 @torch.no_grad()
 def load_fsdp2_model_to_gpu(model):
-    device_id = torch.cuda.current_device()
-    model.to(f"cuda:{device_id}", non_blocking=True)
+    device = torch.cuda.current_device()
+    for param in model.parameters():
+        param.data = param.data.to(device, non_blocking=True)
 
 @torch.no_grad()
 def offload_fsdp_optimizer(optimizer):
@@ -175,7 +179,7 @@ def offload_fsdp_optimizer(optimizer):
         for param in param_group["params"]:
             state = optimizer.state[param]
             for key, value in state.items():
-                if isinstance(value, torch.Tensor):
+                if isinstance(value, (torch.Tensor, DTensor)):
                     state[key] = value.to("cpu", non_blocking=True)
 
 
@@ -187,7 +191,7 @@ def load_fsdp_optimizer(optimizer, device_id):
         for param in param_group["params"]:
             state = optimizer.state[param]
             for key, value in state.items():
-                if isinstance(value, torch.Tensor):
+                if isinstance(value, (torch.Tensor, DTensor)):
                     state[key] = value.to(device_id, non_blocking=True)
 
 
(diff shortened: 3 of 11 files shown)


verl/workers/fsdp_workers.py
```diff
@@ -251,18 +253,37 @@ def _build_model_optimizer(self,
...
+                device_id=torch.cuda.current_device(),
+                sharding_strategy=sharding_strategy,  # zero3
+                mixed_precision=mixed_precision,
+                sync_module_states=True,
+                device_mesh=self.device_mesh,
+                forward_prefetch=False)
+        elif fsdp_strategy == 'fsdp2':
+            assert CPUOffloadPolicy is not None, "PyTorch version >= 2.4 is required for using fully_shard API (FSDP2)"
+            mp_policy = MixedPrecisionPolicy(param_dtype=param_dtype, reduce_dtype=reduce_dtype, cast_forward_inputs=True)
+            cpu_offload = None if role == 'actor' else CPUOffloadPolicy(pin_memory=True)
```
vermouth1992: Does FSDP2 support gradient accumulation and CPUOffload at the same time?

verl/utils/fsdp_utils.py
```diff
@@ -340,3 +363,86 @@ def init_fn(sub_mod: torch.nn.Module, recurse: bool = True):
...
+    assert CPUOffloadPolicy is not None, "PyTorch version >= 2.4 is required for using fully_shard API (FSDP2)"
+    assert hasattr(model.model, 'layers'), "TransformerBlock layer was not found in model.model, please check model structure"
+
+    fsdp_mesh = fsdp_kwargs.get('mesh')
+
+    # refer torchtitan
+    for idx, layer in enumerate(model.model.layers):
+        reshard_after_forward = fsdp2_sharding_strategy(fsdp_mesh)
+        if idx == len(model.model.layers) - 1:
+            reshard_after_forward = False
```
vermouth1992: For inference only model such as reference policy, we do want reshard_after_forward for the last layer

verl/utils/fsdp_utils.py
```diff
@@ -340,3 +363,86 @@ def init_fn(sub_mod: torch.nn.Module, recurse: bool = True):
...
+        model.to('cpu')
+        for buf in model.buffers():
+            buf.to(torch.cuda.current_device())
+    
+
+def apply_fsdp2(model, fsdp_kwargs):
+    '''model: AutoModelForCausalLM
+    '''
+    assert CPUOffloadPolicy is not None, "PyTorch version >= 2.4 is required for using fully_shard API (FSDP2)"
+    assert hasattr(model.model, 'layers'), "TransformerBlock layer was not found in model.model, please check model structure"
```
vermouth1992: Is there any way to inspect the _no_split_modules just like in FSDP1 to determine the fsdp wrap layer?

vermouth1992:
Thanks for the PR! Also, just wonder are there any benchmark results compared with FSDP1?

verl/utils/fsdp_utils.py
gali-leilei: There is no wrapper. From the [torchtitan documentation](https://github.com/pytorch/torchtitan/blob/main/docs/fsdp.md#fsdp1--fsdp2-api-differences):

- `fully_shard(module)` adds an `FSDPState` object on `module`, accessible via `fully_shard.state(module)`, instead of being an `nn.Module` wrapper. This is done via the @contract decorator.
- Calling `model.named_parameters()` for a `model` with FSDP2 applied returns unchanged parameter names and `DTensor` sharded parameters. This means that the optimizer and gradient norm clipping see `DTensors`

lxg2015:
I used the run_qwen2-7b.sh script to conduct tests on a single machine with 8 GPUs.
 
FSDP2 is 6% slower than FSDP1, and there is no difference in the maximum GPU memory usage (however, there is an issue with the return value of torch.cuda.max_memory_reserved() here).

image
image

verl/utils/fsdp_utils.py
lxg2015: An is_infer parameter was added to do this

verl/workers/fsdp_workers.py
lxg2015: I tested enabling offload for actor, but I encountered the following error, which hasn't been resolved yet.

`RuntimeError: FSDP parameters should be materialized on the CPU when enabling CPU offloading. For example, load a CPU state dict or call module.to_empty(device="cpu"). Found the following parameters on a non-CPU device: [('model.embed_tokens.weight', device(type='cuda', index=0)), ('model.norm.weight', device(type='cuda', index=0)), ('lm_head.weight', device(type='cuda', index=0))]`

eric-haibin-lin:
pls remember to sign CLA. thanks

lxg2015:
Hi  @eric-haibin-lin @lei-lei-shanda 

I've fixed some issues and tested that the performance of FSDP2 is superior to FSDP1 (run_qwen2-7b.sh  on 4 gpus). Please review it again. Thank you.

- In the same param_offload and optimizer_offload mode, FSDP2 reduces the max_memory_reserved peak value by 8% compared to FSDP1, and the training speed is also improved, although not significantly. 
- when FSDP2 enables CPUoffload, the memory usage is reduced by 27% compared to FSDP1, but the training speed drops by nearly 50%.  (FSDP2's CPUoffload is compatible with gradient accumulation. I verified in a single test script that the gradients are completely consistent when CPU offload is enabled and disabled, and the loss aligns in the training tasks.)

image

image

verl/utils/fsdp_utils.py
```diff
@@ -137,9 +149,19 @@ def offload_fsdp_model_to_cpu(model: FSDP, empty_cache: bool = True):
...
 
+@torch.no_grad()
+def offload_fsdp2_model_to_cpu(model, empty_cache: bool = True):
+    for param in model.parameters():
+        param.data = param.data.to(torch.device('cpu'), non_blocking=True)
+    if empty_cache:
+        torch.cuda.empty_cache()
 
 @torch.no_grad()
 def load_fsdp_model_to_gpu(model: FSDP):
```
weifengpy, line 152: what's the callsite for `load_fsdp_model_to_gpu`? when applying `FSDP` or `fully_shard`, we move model to cuda

verl/utils/fsdp_utils.py
```diff
@@ -162,7 +189,7 @@ def offload_fsdp_optimizer(optimizer):
         for param in param_group["params"]:
             state = optimizer.state[param]
             for key, value in state.items():
-                if isinstance(value, torch.Tensor):
+                if isinstance(value, (torch.Tensor, DTensor)):
```
weifengpy: nit: DTensor is an instance of torch.Tensor

would `isinstance(value, torch.Tensor)` be enough already?

verl/utils/fsdp_utils.py
```diff
@@ -348,3 +375,101 @@ def init_fn(sub_mod: torch.nn.Module, recurse: bool = True):
...
+    else:
+        return nullcontext()
+
+
+def fsdp2_sharding_strategy(device_mesh):
+    sharding_strategy = False # zero1,2
+    if device_mesh.ndim == 1:
+        sharding_strategy = True # zero3
+    elif device_mesh.ndim == 2:
+        sharding_strategy = torch.cuda.device_count() # hsdp
```
weifengpy: why `reshard_after_forward=N` is decided by device_mesh.ndim? I might miss some context here

for hsdp, fully_shard(device_mesh=(replicate, shard)), we can do reshard_after_forward=True/False/Int. device_mesh and reshard_after_forward are orthogonal to me

verl/utils/fsdp_utils.py
```diff
@@ -348,3 +375,101 @@ def init_fn(sub_mod: torch.nn.Module, recurse: bool = True):
...
+    for name, module in model.named_modules():
+        if module.__class__.__name__ in fsdp_transformer_layer_cls_to_wrap or isinstance(module, nn.Embedding):
+            modules.append(module)
+    
+    for idx, module in enumerate(modules):
+        fully_shard(module, **fsdp_kwargs)
+    fully_shard(model, **fsdp_kwargs) # fsdp2 will not reshard_after_forward for root module
+
+
+def fsdp2_clip_grad_norm_(parameters, max_norm, norm_type=2.0, error_if_nonfinite=False, foreach=None):
```
weifengpy, line 438: is following code copied from torchtitan?  we calculate total_norm in torchtitan because of using pipeline parallel https://github.com/pytorch/torchtitan/blob/9cf88aa1a1834670a6e91e8ba4a2e9af8dd74bf6/torchtitan/distributed/utils.py#L278

For FSDP2 without pipeline parallel, we can call `torch.nn.utils.clip_grad_norm_` directly https://github.com/pytorch/pytorch/blob/562328501e167206dc7d4b16895b5ae538520e06/test/distributed/_composable/fsdp/test_fully_shard_clip_grad_norm_.py#L66
```
total_norm = torch.nn.utils.clip_grad_norm_(
    model.parameters(),
    max_norm=max_norm,
    norm_type=norm_type,
)
```

weifengpy: the suggestion is optional. if you want to enable pipeline parallel, feel free to use current implementation

weifengpy (commented):
finished review from FSDP2 perspective. only concern is how `reshard_after_forward=int` is related to hsdp 2d device mesh.

verl/utils/fsdp_utils.py
lxg2015: [This runs on the actor model.](https://github.com/volcengine/verl/blob/main/verl/workers/fsdp_workers.py#L507) When we disable the offload_policy of fsdp2 while enabling param_offload, it will call load_fsdp_model_to_gpu before actor.update_policy and will call offload_fsdp_model_to_cpu after actor.update_policy, and this acts like fsdp1

verl/utils/fsdp_utils.py
lxg2015: Yes, it's enough already, I will restore this

verl/utils/fsdp_utils.py
lxg2015: verl will initialize the device_mesh as described [in the init](https://github.com/volcengine/verl/blob/main/verl/workers/fsdp_workers.py#L59) with 2 dim for hsdp , so a specific judgement related to the device_mesh is made here, [same as fsdp1](https://github.com/volcengine/verl/blob/main/verl/workers/fsdp_workers.py#L65)

or should it be changed like this
```
sharding_strategy = device_mesh.get_group(-1).size()
```

verl/utils/fsdp_utils.py
lxg2015: It's copied from [pytorch](https://github.com/pytorch/pytorch/blob/78953ee1223391df5c162ac6d7e3eb70294a722e/torch/nn/utils/clip_grad.py#L180), it will raise the following Error when enable CPUOffloadPolicy for fsdp2  if i use torch.nn.utils.clip_grad_norm_ directly. so i move grad_norm to cuda before _clip_grads_with_norm_
```
 File "verl/workers/actor/dp_actor.py", line 190, in _optimizer_step
    grad_norm = torch.nn.utils.clip_grad_norm_(self.actor_module.parameters(), max_norm=self.config.grad_clip)
  File ".local/lib/python3.9/site-packages/torch/nn/utils/clip_grad.py", line 34, in _no_grad_wrapper
    return func(*args, **kwargs)
  File ".local/lib/python3.9/site-packages/torch/nn/utils/clip_grad.py", line 216, in clip_grad_norm_
    _clip_grads_with_norm_(parameters, max_norm, total_norm, foreach)
  File ".local/lib/python3.9/site-packages/torch/nn/utils/clip_grad.py", line 34, in _no_grad_wrapper
    return func(*args, **kwargs)
  File ".local/lib/python3.9/site-packages/torch/nn/utils/clip_grad.py", line 155, in _clip_grads_with_norm_
    clip_coef = max_norm / (total_norm + 1e-6)
  File ".local/lib/python3.9/site-packages/torch/_tensor.py", line 39, in wrapped
    return f(*args, **kwargs)
  File ".local/lib/python3.9/site-packages/torch/_tensor.py", line 1077, in __rdiv__
    return self.reciprocal() * other
  File ".local/lib/python3.9/site-packages/torch/_compile.py", line 32, in inner
    return disable_fn(*args, **kwargs)
  File ".local/lib/python3.9/site-packages/torch/_dynamo/eval_frame.py", line 745, in _fn
    return fn(*args, **kwargs)
  File ".local/lib/python3.9/site-packages/torch/distributed/tensor/_api.py", line 346, in __torch_dispatch__
    return DTensor._op_dispatcher.dispatch(
  File ".local/lib/python3.9/site-packages/torch/distributed/tensor/_dispatch.py", line 182, in dispatch
    self.redistribute_local_args(
  File ".local/lib/python3.9/site-packages/torch/distributed/tensor/_dispatch.py", line 318, in redistribute_local_args
    resharded_local_tensor = redistribute_local_tensor(
  File ".local/lib/python3.9/site-packages/torch/distributed/tensor/_redistribute.py", line 208, in redistribute_local_tensor
    new_local_tensor = partial_spec._reduce_value(
  File ".local/lib/python3.9/site-packages/torch/distributed/tensor/_ops/_math_ops.py", line 126, in _reduce_value
    reduced_tensor = super()._reduce_value(tensor, mesh, mesh_dim)
  File ".local/lib/python3.9/site-packages/torch/distributed/tensor/placement_types.py", line 599, in _reduce_value
    return funcol.all_reduce(
  File ".local/lib/python3.9/site-packages/torch/distributed/_functional_collectives.py", line 176, in all_reduce
    tensor = torch.ops._c10d_functional.all_reduce(self, reduceOp.lower(), group_name)
  File ".local/lib/python3.9/site-packages/torch/_ops.py", line 1123, in __call__
    return self._op(*args, **(kwargs or {}))
RuntimeError: No backend type associated with device type cpu
```

lxg2015:
fsdp2 CPUOffloadPolicy allows us to train larger models. The slow speed is mainly due to the optimizer running on the CPU, perhaps we can optimize this time later.

verl/utils/fsdp_utils.py
lxg2015: Hi @weifengpy 
Let me explain what I understand.  `reshard_after_forward` only controls the number of shards for the parameters, and in `device_mesh=(replicate, shard)`, the size of `shard` determines the number of shards for the optimizer and the gradient.  so they are orthogonal.
Did I understand correctly?  If I did,  should I expose the reshard_after_forward parameter to user regardless of whether it is HSDP or not. Thanks

verl/utils/fsdp_utils.py
weifengpy: @lxg2015 we can do `dist.init_process_group(backend="cpu:gloo,cuda:nccl")` to resolve error `No backend type associated with device type cpu`. torchtitan does this as well

https://github.com/pytorch/torchtitan/blob/f27a1843a503fadf06876a3797bd7305098917a7/torchtitan/distributed/utils.py#L223-L225

verl/utils/fsdp_utils.py
weifengpy: > Did I understand correctly? If I did, should I expose the reshard_after_forward parameter to user regardless of whether it is HSDP or not. Thanks

your understanding is correct, they are orthogonal

> should I expose the reshard_after_forward parameter to user regardless of whether it is HSDP or not. Thanks

yes, it should be exposed to user. practically we set reshard_after_forward=False for FSDP2 + pipeline parallel

lxg2015: Thanks, I have exposed this parameter in ppo_trainer.yaml. since Verl does not currently support pipeline parallel, I have not set this

verl/utils/fsdp_utils.py
lxg2015: I see that Verl calls init_process_group in many workers(actor\critic\...), and these workers can be collapsed into one process group, which makes it uncertain which init_process_group is called first. Also, it's not easy to obtain the entire configuration to decide whether to add cpu:gloo during init_process_group in many worker initializations.

I think the current implementation is okay. because it's only called once during a whole training step. I have profiled both methods and found there is no difference in training time.

lxg2015:
Hi @vermouth1992 @eric-haibin-lin 

I have resolved the issues mentioned above.  Can this be merged, or are there any other suggestions?

verl/trainer/config/ppo_trainer.yaml
```diff
@@ -142,6 +145,8 @@ critic:
     fsdp_config:
       param_offload: False
       optimizer_offload: False
+      offload_policy: False # only for fsdp2, offload param\grad\optimizer during train
```
ZihengJiang, line 153: umm, can you infer the offload_policy with the `param_offload` and `optimizer_offload` arguments instead of adding a new one?
Also, how does reshard_after_forward work?

verl/workers/fsdp_workers.py
```diff
@@ -49,6 +49,9 @@
 from verl.utils.model import compute_position_id_with_mask
 from verl.workers.sharding_manager.fsdp_ulysses import FSDPUlyssesShardingManager
 
+from verl.utils.fsdp_utils import CPUOffloadPolicy, MixedPrecisionPolicy, fsdp_version, apply_fsdp2, \
```
ZihengJiang: Will this introduce higher version requirement for user's torch package? Like low version torch will cause the ImportError

lxg2015: I have checked this in[ fsdp_utils.py](https://github.com/lxg2015/verl/blob/main_fsdp2_meta_tensor_1/verl/utils/fsdp_utils.py#L34)
```
if version.parse(torch.__version__) >= version.parse('2.6'):
    from torch.distributed.fsdp import fully_shard, MixedPrecisionPolicy, FSDPModule, CPUOffloadPolicy
elif version.parse(torch.__version__) >= version.parse('2.4'):
    from torch.distributed._composable.fsdp import fully_shard, MixedPrecisionPolicy, FSDPModule, CPUOffloadPolicy
else:
    fully_shard, MixedPrecisionPolicy, FSDPModule, CPUOffloadPolicy = None, None, None, None
```

verl/trainer/config/ppo_trainer.yaml
lxg2015: Actually, param & optimizer offload is not equivalent to fsdp2 offload policy. The fsdp2 offload policy will run `optimizer.step` on the CPU, which saves more GPU memory but runs much slower than param & optimizer offload.

reshard_after_forward is only about parameters; it determines whether and how to shard parameters between the forward and backward passes. After the backward pass, the parameters will be sharded according to the mesh parameter passed to the fully_shard API.

eric-haibin-lin (commented):
please add one record in https://github.com/volcengine/verl/blob/main/docs/experiment/ppo.rst

eric-haibin-lin:
pls update PR description according to https://github.com/volcengine/verl/blob/main/.github/PULL_REQUEST_TEMPLATE.md and show example usage

verl/workers/fsdp_workers.py
```diff
@@ -771,19 +799,38 @@ def _build_critic_model_optimizer(self, config):
...
+                
+            fsdp_kwargs = {
+                "mesh": fsdp_mesh,
+                "mp_policy": mp_policy,
+                "offload_policy": offload_policy,
+                "reshard_after_forward": fsdp_config.reshard_after_forward,
+            }
+            full_state = critic_module.state_dict()
+            apply_fsdp2(critic_module, fsdp_kwargs, fsdp_config)
+            fsdp2_load_full_state_dict(critic_module, full_state, fsdp_mesh, offload_policy)
```
PeterSH6, line 873: Why do we need `fsdp2_load_full_state_dict` function? 
Loading the full parameter in rank 0 would easily cause OOM for models larger than 32B.
Correct me if I'm wrong: The critic_module already has the pretrained parameter. We can simply call `apply_fsdp2` to shard it with the pretrained parameter, similar to FSDP1

verl/utils/fsdp_utils.py
```diff
@@ -333,3 +360,91 @@ def init_fn(sub_mod: torch.nn.Module, recurse: bool = True):
...
+
+    modules = []
+    for name, module in model.named_modules():
+        if module.__class__.__name__ in fsdp_transformer_layer_cls_to_wrap or \
+            (isinstance(module, nn.Embedding) and not model.config.tie_word_embeddings):
+            modules.append(module)
+    
+    for idx, module in enumerate(modules):
+        fully_shard(module, **fsdp_kwargs)
+    fully_shard(model, **fsdp_kwargs) # fsdp2 will not reshard_after_forward for root module
```
wuxibin89: From pytorch fully_shard doc:
> Users generally should not call [fully_shard()](https://pytorch.org/docs/stable/distributed.fsdp.fully_shard.html#torch.distributed.fsdp.fully_shard) only on the topmost root module.

Why do we need shard root module here? And what does `fsdp2 will not reshard_after_forward for root module` mean?

lxg2015: It's said we should not call fully_shard `only` on root module, so we call fully_shard on every transformer block and the root module.  the root module here only contains output layer parameters actually. 

In FSDP2, the reshard_after_forward parameter for the root module, which contains only the output layer, will be ignored. You may refer to the[ FSDP2 source code ](https://github.com/pytorch/pytorch/blob/main/torch/distributed/fsdp/_fully_shard/_fsdp_state.py#L181)for more details.

docs/experiment/ppo.rst
```diff
@@ -47,6 +48,8 @@ NVIDIA GPUs
 +----------------------------------+------------------------+------------+-----------------------------------------------------------------------------------------------+
 | Qwen/Qwen2-7B-Instruct           | GRPO                   | 89         |   `Qwen7b GRPO Script`_                                                                       |
 +----------------------------------+------------------------+------------+-----------------------------------------------------------------------------------------------+
+| Qwen/Qwen2-7B-Instruct           | GRPO (FSDP2)           | 89.8       |  `_Qwen7b GRPO FSDP2 Script and Logs`_                                                                       |
++----------------------------------+------------------------+------------+-----------------------------------------------------------------------------------------------+
```
ccclyu, line 52: use the same number of `-` to make the table have correct render and also remove `_` in `_Qwen7b GRPO FSDP2 Script and Logs` for indexing.

verl/workers/fsdp_workers.py
duomicoding: @PeterSH6 What is the reason mentioned above?

Issue #1031, opened 2025-04-11
AssertionError: score: len(lst)=60, len(sample_scores)=560


YqjMartin:
![Image](https://github.com/user-attachments/assets/4039ee38-00c6-49ed-9bdc-11ed4d8c293d)

Xuekai-Zhu:
the same error:
`AssertionError: score: len(lst)=480, len(sample_scores)=876`

0x404:
Hi @Xuekai-Zhu, could you share the launching script and the which reward function you used?

Xuekai-Zhu:
thanks for your quick replay, the following is my test script:

```
#!/bin/bash

# export VLLM_ATTENTION_BACKEND=XFORMERS

WORKDIR=/fs-computility/plm/shared/zhuxuekai/reasoning_flow
PRETRAINED_MODEL=$WORKDIR/pre_trained_model/Qwen/Qwen2.5-7B
n_gpus_per_node=2
save_freq=-1

# train_path=$WORKDIR/data/math/train.parquet
# test_path=$WORKDIR/data/r1_bench_plus/test.parquet

# gsm8k_train_path=$WORKDIR/data/gsm8k/train.parquet
# gsm8k_test_path=$WORKDIR/data/gsm8k/test.parquet
# math_train_path=$WORKDIR/data/math/train.parquet
# math_test_path=$WORKDIR/data/math/test.parquet
# train_files="['$gsm8k_train_path', '$math_train_path']"
# test_files="['$gsm8k_test_path', '$math_test_path']"

math_train_path=$WORKDIR/data/math/train.parquet

# dapo_train_path=$WORKDIR/data/dapo/dapo-math-17k.parquet
r1_test_path=$WORKDIR/data/r1_bench_plus/test.parquet

experiment_name='test-grpo-math'

set -x

python3 -m verl.trainer.main_ppo \
    algorithm.adv_estimator=grpo \
    data.train_files=$math_train_path \
    data.val_files=$r1_test_path \
    data.train_batch_size=64 \
    data.max_prompt_length=1024 \
    data.max_response_length=2048 \
    data.truncation='left' \
    actor_rollout_ref.model.path=$PRETRAINED_MODEL \
    actor_rollout_ref.actor.optim.lr=1e-6 \
    actor_rollout_ref.model.use_remove_padding=True \
    actor_rollout_ref.actor.ppo_mini_batch_size=32 \
[...]

PR #1044, opened 2025-04-12
fix: correctly handle the inconsistency in reward_extra_info and reward_tensor

0x404:
If we are using multiple datasources for training/test, some reward function may return float and some may return dict (including extra_info) depending on the datasource.

In this case, the `reward_tensor` and `reward_extra_info` returned by `RewardManager` will have different shape, this causes the error in https://github.com/volcengine/verl/issues/1031.

this PR fix the inconsistency by correctly padding a place holder 'unkown' for those reward function only return float, this place holder will not be processed in and has no effect on the correctness:

https://github.com/volcengine/verl/blob/dc1714a428a5b7f8be435fac2d853e6d93ad0f09/verl/trainer/ppo/metric_utils.py#L227-L233


0x404:
Hi, @vermouth1992, Could you please help review this PR? Much appreciated!

PR #1047, opened 2025-04-12
reset default tp size

ETOgaosion:
default tp size use 1


BearBiscuit05:
I will double check whether the default TP value affects other scripts.

vermouth1992 (approved):

PR #1050, opened 2025-04-13
[mcore] change default sequence_parallel setting

BearBiscuit05:
Supplementary change for #1047: For sp, if tp defaults to 1, sp should also be set to false. [Reference](https://github.com/NVIDIA/Megatron-LM/blob/main/README.md#distributed-pretraining)


vermouth1992:
Actually, I guess it's better to always turn on sequence parallel to save memory. And when tp=1, turn off sequence parallel in verl code.

BearBiscuit05:
Sounds reasonable.

PR #1074, opened 2025-04-14
[misc] Fix typo in colocate

lei-lei-shanda:
- Force both usages of `colocate` and `collocate` to `colocate`, to be consistent with [vllm terminology](https://docs.vllm.ai/en/latest/getting_started/examples/rlhf_colocate.html).
- both `ResourcePool` and `RayResourcePool` use the same default value for `max_colocate_count` to avoid surprises.


eric-haibin-lin:
thanks for the contribution!

PR #1075, opened 2025-04-14
fix time for dapo

lilei199908:
When training with Dapo, because there is a continuous filter and dynamic sampling, each iteration involves multiple samplings. The time should be summed up to represent the total sampling time for one iteration.


verl/trainer/ppo/ray_trainer.py
```diff
@@ -235,7 +235,7 @@ def compute_advantage(data: DataProto, adv_estimator, gamma=1.0, lam=1.0, num_re
 def _timer(name: str, timing_raw: Dict[str, float]):
     with Timer(name=name, logger=None) as timer:
         yield
-    timing_raw[name] = timer.last
+    timing_raw[name] += timer.last
```
vermouth1992, line 240: Does this cause error if name is not in timing_raw?

lilei199908: Yes, sorry, I didn’t notice that the dict in ray_trainer.py didn’t use defaultdict. Should I change all the dicts to defaultdict, or should I add the following code?
`if name not in timing_raw:
    timing_raw[name] = 0
`

lilei199908: Or redefine a _timer for methods like dapo that require multiple generations？

vermouth1992 (approved):

verl/trainer/ppo/ray_trainer.py
lilei199908: @vermouth1992 is there something wrong? why checks have error?

PR #1082, opened 2025-04-14
Add bool-type index selection for DataProto

mansicer:
After the last change, current DataProto cannot use bool-type index due to hard-coded batch_size equal to idxs.shape[0]. 

This patch changes the new batch_size for bool-type idx to idxs.sum(). It's useful when users filter the batch with bool-type masks.

### DIFF
--- a/tests/utility/test_tensor_dict_utilities.py
+++ b/tests/utility/test_tensor_dict_utilities.py
@@ -297,3 +297,67 @@ def test_seqlen_balancing():
     reverse_idx_map = torch.tensor(reverse_idx_map)
     new_batch = batch[reverse_idx_map]
     torch.testing.assert_close(new_batch, dataproto.batch)
+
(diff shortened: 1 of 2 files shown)


mansicer:
@eric-haibin-lin can someone review this quick fix?

eric-haibin-lin (changes requested):
thanks! would you mind adding a test similar to `tests/verl/utility/test_tensor_dict_utilities.py` for the case you fixed? Thanks! 

Also please merge with main to resolve conflicts

mansicer:
Thanks for your review! I rebase to the main branch, add a test of DataProto indexing to `tests/verl/utility/test_tensor_dict_utilities.py`, and meanwhile fix a bug in boo-type list idxs input.

eric-haibin-lin:
thanks!

PR #1085, opened 2025-04-14
data: use hf template for vlm models

hiyouga:
## What does this PR do?

In this PR, we use huggingface's chat template (i.e., `processor.apply_chat_template`) to compute input ids for VLMs, it could be generalized to more model architectures compared with the earlier implementation.

## Who can review?

@vermouth1992 @eric-haibin-lin

### DIFF
--- a/tests/rollout/test_vllm_spmd.py
+++ b/tests/rollout/test_vllm_spmd.py
@@ -150,7 +150,7 @@ def test_vllm_spmd():
         skip_tokenizer_init=False,
         enable_prefix_caching=True,
         trust_remote_code=True,
-        seed=0,
+        seed=1,
     )
 
     outputs = llm.generate(preencode_prompts, sampling_params=sampling_params, use_tqdm=False)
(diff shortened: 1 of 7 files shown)


tests/rollout/test_vllm_spmd.py
```diff
@@ -150,7 +150,7 @@ def test_vllm_spmd():
         skip_tokenizer_init=False,
         enable_prefix_caching=True,
         trust_remote_code=True,
-        seed=int(os.getenv("RANK", "0")) // tensor_parallel_size,
```
eric-haibin-lin, line 153: is this change intentional?

verl/utils/dataset/rl_dataset.py
```diff
@@ -151,68 +151,82 @@ def resume_dataset_state(self):
     def __len__(self):
         return len(self.dataframe)
```
eric-haibin-lin, line 153: please update test_rl_dataset with a dataset with text & image

eric-haibin-lin (changes requested):

tests/rollout/test_vllm_spmd.py
hiyouga: Sure, we don't need to use different seeds across dp groups. The earlier implementation is somehow misleading

verl/utils/dataset/rl_dataset.py
hiyouga: done

eric-haibin-lin (approved):

PR #1087, opened 2025-04-14
CI: add vlm test for sglang rollout

JustinTong0323:


eric-haibin-lin (changes requested):
could u merge the test to https://github.com/volcengine/verl/blob/main/.github/workflows/e2e_ppo_trainer.yml#L146 ? we just refactored the CI configs. thanks.

PR #1091, opened 2025-04-15
[feat] multimodal inputs for reward model

mingruimingrui:
Reward model was missing multimodal inputs. It's already been added for the actor & critic workers.

### DIFF
--- a/verl/workers/fsdp_workers.py
+++ b/verl/workers/fsdp_workers.py
@@ -1025,6 +1025,14 @@ class RewardModelWorker(Worker):
             attention_mask = micro_batch['attention_mask']
             position_ids = micro_batch['position_ids']
 
+            multi_modal_inputs = {}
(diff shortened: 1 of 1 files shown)


verl/workers/fsdp_workers.py
```diff
@@ -1025,6 +1025,14 @@ def _forward_micro_batch(self, micro_batch):
             attention_mask = micro_batch['attention_mask']
             position_ids = micro_batch['position_ids']
 
+            multi_modal_inputs = {}
```
hiyouga, line 1028: It is unused actually

hiyouga (changes requested):
See the above comment

verl/workers/fsdp_workers.py
mingruimingrui: My bad! Fixed

hiyouga (approved):

PR #1092, opened 2025-04-15
fix: make prime reward manager compatible with naive manager

### LINKED ISSUE #1070: PRIME reward manager implementation out of sync
I think the prime reward manager is out of sync with the main pipeline. 

For example

Image

It should be important to add check/test for all reward managers at least prime and naive.

0x404:
Currently the naive reward manager will return `reward_extra_info` if some reward function reward a dict, and `return_dict` in manager is set to True.

This PR enable `PrimeRewardManager` to have the same behavior to compatible with the naive manger, also, this can fix if the prime manager use some reward function that return a dict.

Fix https://github.com/volcengine/verl/issues/1070


TranSirius:
Great, this snippet of codes works for me

0x404:
Hi @eric-haibin-lin, could you help take a look of this PR?

PR #1100, opened 2025-04-15
Move entropy to comput log probs to reduce peak memory when calculating entropy.

ETOgaosion:
Actor do not calculate Entropy loss if `entropy_coeff==0`,  and move the calculation of entropy to `compute_log_probs`

Tested configuration:

```sh
    data.max_prompt_length=$((1024 * 2)) \
    data.max_response_length=$((1024 * 10)) \
    actor_rollout_ref.rollout.max_num_batched_tokens=$((1024 * 12)) \
    context_parallel_size=2 \
```

### DIFF
--- a/verl/workers/fsdp_workers.py
+++ b/verl/workers/fsdp_workers.py
@@ -580,7 +580,7 @@ class ActorRolloutRefWorker(Worker):
         data.meta_info['use_dynamic_bsz'] = self.config.ref.log_prob_use_dynamic_bsz
         with self.ulysses_sharding_manager:
             data = self.ulysses_sharding_manager.preprocess_data(data)
-            output, _, _ = self.ref_policy.compute_log_prob(data=data, calculate_entropy=False)
+            output, _ = self.ref_policy.compute_log_prob(data=data, calculate_entropy=False)
             output = DataProto.from_dict(tensors={'ref_log_prob': output})
             output = self.ulysses_sharding_manager.postprocess_data(output)
 
--- a/verl/workers/megatron_workers.py
+++ b/verl/workers/megatron_workers.py
@@ -443,7 +443,7 @@ class ActorRolloutRefWorker(MegatronWorker):
         micro_batch_size = self.config.ref.log_prob_micro_batch_size_per_gpu
         data.meta_info['micro_batch_size'] = micro_batch_size
         data.meta_info['temperature'] = self.config.rollout.temperature
-        output, _, _ = self.ref_policy.compute_log_prob(data=data, calculate_entropy=False)
+        output, _ = self.ref_policy.compute_log_prob(data=data, calculate_entropy=False)
         output = DataProto.from_dict(tensors={'ref_log_prob': output})
         output = output.to('cpu')
         if self._is_offload_param:


verl/utils/megatron/tensor_parallel.py
```diff
@@ -106,6 +107,7 @@ def forward(ctx, vocab_parallel_logits: torch.Tensor) -> torch.Tensor:
         def mul_reduce(a, b):
             return (a * b).sum(dim=-1, keepdim=True)
 
+        log_gpu_memory_usage('before entropy forward')
```
vermouth1992: Remove this

verl/utils/megatron/tensor_parallel.py
```diff
@@ -117,10 +119,12 @@ def mul_reduce(a, b):
         dist.all_reduce(sum_softmax_times_logits, group=mpu.get_tensor_model_parallel_group())
         entropy = logits_max + normalized_sum_exp_logits.log() - sum_softmax_times_logits
         ctx.save_for_backward(vocab_parallel_logits, softmax_logits, sum_softmax_times_logits)
+        log_gpu_memory_usage('after entropy forward')
```
vermouth1992: Remove this

verl/workers/actor/dp_actor.py
```diff
@@ -52,13 +53,17 @@ def __init__(
...
         self.use_ulysses_sp = self.ulysses_sequence_parallel_size > 1
+        self.role = role
 
         self.compute_entropy_from_logits = (
             torch.compile(verl_F.entropy_from_logits, dynamic=True)
             if self.config.get('use_torch_compile', True)  #  use torch compile by default
             else verl_F.entropy_from_logits)
 
-    def _forward_micro_batch(self, micro_batch, temperature) -> Tuple[torch.Tensor, torch.Tensor]:
+    def _forward_micro_batch(self,
```
vermouth1992, line 56: We don't need a role for this. For compute_log_prob, we always compute entropy. For update actor, we only include entropy when entropy_coeff > 0

verl/utils/megatron/tensor_parallel.py
ETOgaosion: Done~

verl/utils/megatron/tensor_parallel.py
ETOgaosion: Done~

verl/workers/actor/dp_actor.py
ETOgaosion: Change the control way from inside check to outside control.

verl/workers/actor/dp_actor.py
```diff
@@ -215,22 +224,31 @@ def compute_log_prob(self, data: DataProto) -> torch.Tensor:
...
 
+            response_mask = micro_batch['attention_mask'][:, -micro_batch['responses'].size(-1):]
             with torch.no_grad():
-                _, log_probs = self._forward_micro_batch(micro_batch, temperature=temperature)
+                entropy, log_probs = self._forward_micro_batch(micro_batch,
+                                                               temperature=temperature,
+                                                               calculate_entropy=calculate_entropy)
             log_probs_lst.append(log_probs)
-        log_probs = torch.concat(log_probs_lst, dim=0)
+            if calculate_entropy:
```
vermouth1992, line 238: Can we return full entropy to the driver and the driver compute entropy? We may need per-token entropy for analysis

recipe/prime/prime_ray_trainer.py
```diff
@@ -398,6 +400,24 @@ def fit(self):
                     # recompute old_log_probs
                     with _timer('old_log_prob', timing_raw):
                         old_log_prob = self.actor_rollout_wg.compute_log_prob(batch)
+                        entropys = old_log_prob.batch['entropys']
+                        response_masks = old_log_prob.batch['response_masks']
+                        entropy_list = torch.chunk(entropys, batch.batch.batch_size[0], dim=0)
+                        response_mask_list = torch.chunk(response_masks, batch.batch.batch_size[0], dim=0)
+                        entropy_loss_dict = {}
+                        for entropy, response_mask in zip(entropy_list, response_mask_list):
```
vermouth1992: Why do we need to chunk here? We can compute the output using full batch?

verl/workers/actor/dp_actor.py
```diff
@@ -215,22 +224,35 @@ def compute_log_prob(self, data: DataProto) -> torch.Tensor:
...
+            entropys = torch.concat(entropy_lst, dim=0)
+            response_masks = torch.concat(response_mask_lst, dim=0)
         if use_dynamic_bsz:
             indices = list(itertools.chain.from_iterable(indices))
             assert len(indices) == log_probs.size(0), f"{len(indices)} vs. {log_probs.size()}"
             revert_indices = torch.tensor(get_reverse_idx(indices), dtype=torch.long)
             log_probs = log_probs[revert_indices]
 
-        return log_probs
+        return log_probs, entropys, response_masks
```
vermouth1992: We don't need to return response_mask. The response_mask can be retrieved from the batch on the single controller.

vermouth1992 (approved):

PR #1101, opened 2025-04-15
fix: gpu memory report in vllm sleep mode

hiyouga:
## What does this PR do?

When we use vLLM's sleep mode, we can no longer obtain the GPU memory statistics from the PyTorch API because vLLM uses cuMem to manage the memory space. We compute the freed memory of vLLM offloading to estimate the memory correctly.

see https://github.com/vllm-project/vllm/pull/11743#issuecomment-2754338119

Before:
image

After:
image

## Who can review?

@vermouth1992 @tongyx361 @BearBiscuit05

### DIFF
--- a/verl/utils/debug/performance.py
+++ b/verl/utils/debug/performance.py
@@ -19,8 +19,10 @@ import logging
 
 def log_gpu_memory_usage(head: str, logger: logging.Logger = None, level=logging.DEBUG, rank: int = 0):
     if (not dist.is_initialized()) or (rank is None) or (dist.get_rank() == rank):
-        memory_allocated = torch.cuda.memory_allocated() / 1024**3
-        memory_reserved = torch.cuda.memory_reserved() / 1024**3
+        free_mem, total_mem = torch.cuda.mem_get_info()
+        vllm_sleep_mem = torch.cuda.memory_reserved() - (total_mem - free_mem)
+        memory_allocated = (torch.cuda.memory_allocated() - vllm_sleep_mem) / 1024**3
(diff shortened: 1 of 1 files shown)


verl/utils/debug/performance.py
```diff
@@ -19,8 +19,11 @@
 
 def log_gpu_memory_usage(head: str, logger: logging.Logger = None, level=logging.DEBUG, rank: int = 0):
     if (not dist.is_initialized()) or (rank is None) or (dist.get_rank() == rank):
-        memory_allocated = torch.cuda.memory_allocated() / 1024**3
-        memory_reserved = torch.cuda.memory_reserved() / 1024**3
+        # take care of vllm sleep mem, see https://github.com/volcengine/verl/pull/1101
```
vermouth1992, line 22: This function is a general one. We shouldn't modify it here. Instead, we should modify it in the vllm part

hiyouga: But there is only one function to log the vram usage, should we create another function for vllm part? Besides, the sleep mode affects the memory log in both vllm side and the fsdp worker side

hiyouga: I think it is quite reasonable to check the mismatched values of torch.cuda.memory and torch.cuda.get_mem_info

vermouth1992: Does torch.cuda.empty_cache release the GPU memory?

hiyouga: Why do you think torch.cuda.empty_cache is related to this topic

PR #1104, opened 2025-04-15
[megatron] fix optimizer config

ccclyu:
### Changes: 

- Remove unused config in the optimizer 
- Add TODO for OptimizerParamScheduler and will submit another PR for LR scheduler in megatron actor/critic etc.


eric-haibin-lin (dismissed):

ccclyu:
resolved conflict with main branch.  can we merge this now? @eric-haibin-lin

verl/utils/megatron/optimizer.py
```diff
@@ -16,26 +16,22 @@
 import importlib
 from packaging.version import Version
 
-from apex.optimizers import FusedAdam as Adam
-from apex.optimizers import FusedSGD as SGD
-
 from megatron.core.optimizer import OptimizerConfig
-
+from megatron.core.optimizer_param_scheduler import OptimizerParamScheduler
```
eric-haibin-lin: is this unused import intended?

ccclyu: yes. we can remove now and it should be used for LR scheduler config, which I will add in another PR.

eric-haibin-lin (approved):
LGTM. just waiting for all tests to finish

PR #1114, opened 2025-04-16
[feat] compute reward during log_prob for ppo

thibautbar:


mertunsall:
@eric-haibin-lin We would be really happy to merge this - this is extremely helpful in speeding up when reward computation is computationally expensive (LLM-judge, compilation of a project, etc.)

verl/trainer/ppo/ray_trainer.py
```diff
@@ -238,6 +238,98 @@ def _timer(name: str, timing_raw: Dict[str, float]):
     timing_raw[name] = timer.last
 
 
+def get_custom_reward_fn(config):
```
hiyouga: can we move these functions to a new file? like verl/trainer/ppo/reward.py

thibautbar: done, thanks

verl/trainer/ppo/reward.py
```diff
@@ -0,0 +1,109 @@
...
+
+    reward_kwargs = dict(reward_fn_config.get("reward_kwargs", {}))
+
+    def wrapped_fn(*args, **kwargs):
+        return raw_fn(*args, **kwargs, **reward_kwargs)
+
+    return wrapped_fn
+
+
+def load_reward_manager(config, tokenizer, num_examine, take_reward_kwargs):
```
eric-haibin-lin: what about changing the interface to 
```
def load_reward_manager(config, tokenizer, num_examine, **reward_kwargs):
```

and invoke with 
```
load_reward_manager(config, tokenizer, num_examine=0, **config.reward_model.get("reward_kwargs", {})
```

verl/trainer/config/ppo_trainer.yaml
```diff
@@ -179,6 +179,7 @@ reward_model:
   use_dynamic_bsz: ${critic.use_dynamic_bsz}
   forward_max_token_len_per_gpu: ${critic.forward_max_token_len_per_gpu}
   reward_manager: naive
+  reward_manager_during_log_prob: False # custom reward function executed async on CPU, during log_prob
```
eric-haibin-lin, line 184: can we change it to 

launch_reward_fn_async: False  /  True?

eric-haibin-lin (changes requested):
thx!

verl/trainer/config/ppo_trainer.yaml
thibautbar: done, thanks

eric-haibin-lin:
do you mind merge with main again? we fixed some pre-commit config where it reports errors that are not changed by the PR

thibautbar:
> do you mind merge with main again? we fixed some pre-commit config where it reports errors that are not changed by the PR

Done, thanks for the review !

verl/trainer/ppo/reward.py
```diff
@@ -0,0 +1,116 @@
...
+    except Exception as e:
+        print(f"Error in reward_fn: {e}")
+        reward_tensor = reward_fn(data)
+        reward_extra_infos_dict = {}
+
+    return reward_tensor, reward_extra_infos_dict
+
+
+@ray.remote(num_cpus=1)
+def compute_reward_async(data: DataProto, config, tokenizer):
```
hiyouga, line 110: why not use the already loaded reward manager in async computing reward? what is the intuition of reloading the reward manager

thibautbar: I am reloading the reward manager because it is not possible to give to a worker a function loaded using importlib if the imported python code is importing other locally installed packages, even if it is given using https://docs.ray.io/en/latest/ray-core/handling-dependencies.html#runtime-environments-api-ref.

hiyouga: got it

hiyouga (approved):

verl/trainer/ppo/reward.py
hiyouga: would you like to provide a comparison of the reward computation duration before/after this PR?

thibautbar: I added a plot in the description. The duration of the computation is not changed, but the overall throughput is improved as it reduces the step time. Let me know if you need anything else

hiyouga:
@eric-haibin-lin I think this feature makes sense, could you review it again?

eric-haibin-lin:
looks good. thanks!

PR #1118, opened 2025-04-16
[profile]print cuda system memory and offload actor model after init

gzpan:


hiyouga (dismissed):

verl/workers/fsdp_workers.py
```diff
@@ -403,6 +403,10 @@ def init_model(self):
             # get the original unwrapped module
             self.actor_module = self.actor_module_fsdp._fsdp_wrapped_module
 
+            if self._is_offload_param:
+                offload_fsdp_model_to_cpu(self.actor_module_fsdp)
```
eric-haibin-lin, line 469: i remember fsdp offloading (not sure if it's gradient or param) is not compatible with gradient accumulation

hiyouga: No, it isn't fsdp offloading. We have already turned off fsdp offload here: https://github.com/volcengine/verl/blob/d6821a051a259db6d5ae2254eb2ec0ef62059bd7/verl/workers/fsdp_workers.py#L250-L253

BearBiscuit05:
This PR is important for verl's memory analysis. Can we push it forward quickly? Thanks!

hiyouga:
Sure @BearBiscuit05  let me rebase this PR

PR #1124, opened 2025-04-16
Support unbiased off-policy GRPO

LYMDLUT:


verl/trainer/ppo/core_algos.py
```diff
@@ -467,6 +467,10 @@ def kl_penalty(logprob: torch.FloatTensor, ref_logprob: torch.FloatTensor, kl_pe
     if kl_penalty == "mse":
         return 0.5 * (logprob - ref_logprob).square()
 
+    if kl_penalty == "k3_off_policy":
```
vermouth1992, line 470: Could you give a reference here?

LYMDLUT: You can refer to the following article:
https://zhuanlan.zhihu.com/p/1892008158626546312
https://hongyuzang.notion.site/The-critical-implementation-detail-of-KL-loss-in-GRPO-1ae3fe2c1ff9809a9307c5402e190373
https://tongyx361.github.io/blogs/posts/kl-rel-to-ref-in-rl-zh/
The corresponding implementation code in OpenRLHF is as follows:
https://github.com/OpenRLHF/OpenRLHF/pull/840

vermouth1992: Could you write those reference inside the code so that people can find its source? Thanks

LYMDLUT: Sure, no problem. I wonder if Verl can provide some computing support to test the performance within Verl.

PR #1128, opened 2025-04-17
[misc] change logging prefix for WorkerDict

gali-leilei:
Net effect is that `(WorkerDict pid=xxx)` becomes `(WorkerDict[ActorRolloutRefWorker] pid=xxx)` in when using `print()` statement with console logging.

changes include:
- move `def _unwrap_ray_remote` to private class method of `class RayClassWithInitArgs`.
- add `def __repr__()` to `class WorkerDict`, according to [guide here](https://docs.ray.io/en/latest/ray-observability/user-guides/configure-logging.html#customizing-prefixes-for-actor-logs)
- refactoring of input validation logic of `def create_colocated_worker_cls()`.


vermouth1992:
Could you show a screen shot of what it becomes after this PR? Thanks.

gali-leilei:
closed due to staleness.

PR #1139, opened 2025-04-17
[mcore] qwen2moe support

ISEEKYAN:
support qwen2moe structure to run with megatron-core
including:
* qwen2moe config converter 
* qwen2moe model initializer
* refactor the online weight converter from mcore to vllm
* qwen2moe online weight converter
* qwen2moe offline weight conversion script from hf to mcore
* a script to run training qwen1.5moe_a2.7b with 4 nodes

TODO
add option to freeze the MoE router weight during training


ISEEKYAN:
There seems to be a bug with computing gsm8k value over multiple nodes traning.
I trained qwen2-7b with 1/2/4 nodes, their scores are 0.88/0.44/0.22.
I trained qwen1.5-moe-a2.7b with 4 nodes(OOM with less nodes), which is reported to achieve [0.625](https://qwenlm.github.io/blog/qwen-moe/) on GSM8k, the score is about 0.159, roughly 1/4 of 0.625.

Microsoft Edge 2025-04-17 21 22 13

vermouth1992:
Could you fix weight format? Also, we have a qwen moe weight loader patcher yesterday..

ISEEKYAN:
> Could you fix weight format? Also, we have a qwen moe weight loader patcher yesterday..

can you show more details about your qwen moe weight loader patcher, I will check and merge the two implementations.

vermouth1992:
> > Could you fix weight format? Also, we have a qwen moe weight loader patcher yesterday..
> 
> can you show more details about your qwen moe weight loader patcher, I will check and merge the two implementations.

Here: https://github.com/volcengine/verl/pull/1137

ISEEKYAN:
> Great Job! Note that there seems to be multiple weight converters currently, we may need to uniform them to a single functional unit. Besides, transformer config needs some refine. And maybe current dist checkpoint and converter need some ci test. This PR is huge enough, merge first.

about conveter
I notice that the weight layout and names are difference across the different model structures, e.g. Qwen2ForCausalLM/Qwen2MoeForCausalLM/DeepseekV3ForCausalLM. So I introduced an OOP design, which is, every structure has a converter class that inherit from the base class, making it easy to reuse the common codes such as the attention parts.

about transformer config, what direction is the refinement to?

about the CI, the CI is necessary, let's add some in the next days.

verl/models/mcore/model_initializer.py
```diff
@@ -50,31 +56,94 @@ def init_mcore_model_dense(
...
+    pre_process=None,
+    post_process=None,
+    share_embeddings_and_output_weights=False,
+    value=False,
+    freeze_moe_router=True,
+    **extra_kwargs,
 ):
-    return init_mcore_model_dense(
-        tfconfig, hf_config, pre_process, post_process, share_embeddings_and_output_weights, value
+    from megatron.core.models.gpt.gpt_layer_specs import get_gpt_decoder_block_spec
```
ccclyu, line 68: as we use ``megatron.core`` in the all init functions, can we move to the header of this file to avoid duplicate import?

verl/models/mcore/model_initializer.py
```diff
@@ -50,31 +56,94 @@ def init_mcore_model_dense(
...
+        tfconfig.moe_router_load_balancing_type = "none"
+
+    def patch_layer_spec(transformer_layer_spec):
+        # shared_experts.gate=True
+        for i in range(len(transformer_layer_spec.layer_specs)):
+            transformer_layer_spec.layer_specs[i].submodules.mlp.submodules.shared_experts.params["gate"] = True
+        return transformer_layer_spec
+
+    assert tfconfig.normalization == "RMSNorm", "only RMSNorm is supported for now"
+    transformer_layer_spec = get_gpt_decoder_block_spec(tfconfig, use_transformer_engine=use_te)
```
ccclyu, line 82: nit: can we directly set `use_transformer_engine` to True instead of using one variable `use_te`?

ETOgaosion: Thanks a lot. We may take this into consideration.

verl/models/mcore/model_initializer.py
ETOgaosion: Thanks! We will try if it does not influence the CPU initialization process.

ISEEKYAN: good suggestions, we may refine the code in the next PRs

BearBiscuit05:
> Could you fix code format? Also, we have a qwen moe weight loader patcher yesterday..

I have optimized the previous code in the new PR #1200.

PR #1140, opened 2025-04-17
[feat] add file level weighted sampling for rl_dataset

pengsun:
Allow weighted sampling for multiple dataset files during RL training.

Suppose you have two datasets `GSM8K` (\~= 8K instances) and `BFG10K` (\~= 10K instances) in RL, and you want to mix in balanced data from these two datasets when making a mini batch at each iteration, then you can specify the data file weights by a list of numbers like this:
```
...
 data.train_files="['/path/to/gsm8k.parquet','/path/to/bfg10k.parquet']" \
 data.shuffle=True \
 +data.train_file_weights="[10,8]" \
...
```
as this ensures that the "total unnormalized probability mass (= num instances * weight)" equalizes for the two datasets: 8K * 10 = 10K * 8. 

Of course, other weights and other length of list can be set. Note also that the underlying torch `WeightedRandomSampler` accepts unnormalized weights as input.

One can get back to the conventional random shuffling (uniformly random sampling):
```
data.train_files="['/path/to/gsm8k.parquet','/path/to/bfg10k.parquet']" \
data.shuffle=True \
```
or sequential iterating (deterministic sampling):
```
data.train_files="['/path/to/gsm8k.parquet','/path/to/bfg10k.parquet']" \
data.shuffle=False \
```
by NOT setting the `data.train_file_weights` argument.

### DIFF
--- a/verl/trainer/ppo/ray_trainer.py
+++ b/verl/trainer/ppo/ray_trainer.py
@@ -30,7 +30,7 @@ import numpy as np
 import ray
 from codetiming import Timer
 from omegaconf import OmegaConf, open_dict
-from torch.utils.data import Dataset, RandomSampler, SequentialSampler
+from torch.utils.data import Dataset, RandomSampler, SequentialSampler, WeightedRandomSampler
 from torchdata.stateful_dataloader import StatefulDataLoader
 from tqdm import tqdm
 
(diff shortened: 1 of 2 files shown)


pengsun:
Just squashed and rebased on latest main branch.

pengsun:
Hi, I've updated the commits to resolve any conflict, and the code style should meet the requirement of the new tool.

verl/trainer/ppo/ray_trainer.py
```diff
@@ -467,26 +495,22 @@ def _create_dataloader(self):
 
         self.train_dataset = dataset_cls(
             data_files=self.config.data.train_files,
+            data_file_weights=self.config.data.get("train_file_weights", None),
```
eric-haibin-lin, line 498: no need to add additional arguments since config itself is passed to dataset

verl/trainer/ppo/ray_trainer.py
```diff
@@ -467,26 +495,22 @@ def _create_dataloader(self):
...
             config=self.config.data,
         )
 
         # use sampler for better ckpt resume
-        if self.config.data.shuffle:
-            train_dataloader_generator = torch.Generator()
-            train_dataloader_generator.manual_seed(self.config.data.get("seed", 1))
-            sampler = RandomSampler(data_source=self.train_dataset, generator=train_dataloader_generator)
-        else:
-            sampler = SequentialSampler(data_source=self.train_dataset)
```
eric-haibin-lin, line 481: a better way would be make the sampler choices extensible. similar to dataset, is it possible to set sampler class via config, and load a custom sampler class from ray trainer? we just need to define a general interface for the sampler constructor

verl/trainer/ppo/ray_trainer.py
pengsun: I see. But that way, the `dataset_cls` has to tell whether it is called for training (can be a list of files, see [code here](https://github.com/volcengine/verl/blob/ea4cd319871425df7d8a9b600a177cce185fe3ec/verl/trainer/ppo/ray_trainer.py#L430)) or for validating (can be only a single one dataset, [code here](https://github.com/volcengine/verl/blob/ea4cd319871425df7d8a9b600a177cce185fe3ec/verl/trainer/ppo/ray_trainer.py#L454)), which requires an extra arg called, say, `mode = 'train' or 'valid' `... So I guess it could be more elegant to pass the `dataset_file_weights` directly?

verl/trainer/ppo/ray_trainer.py
pengsun: Yeah, I believe it would be more flexible to allow the sampler be configurable. But I also found several reasons for not doing so:
- When we specify a specific `sampler`, we may also specify the according parameters. `RandomSampler` and `SequentialSampler` happen to be "parameterless", but `WeightedRandomSampler` needs a parameter, i.e., the weights for each instance, which can be inferred from `data_file_weights` in current implementation and can be unrealistic to be specified through config.
- Still, there is a problem of whether the dataset/dataloder is called for training or for validating, as the previous discussion. This way, we need configure the sampler for both training and validating, requiring more code logic and config args...

Considering mainly these two points, I decided to just extend the previous code logic to add the creation code for an extra sampler when submitted this PR.

pengsun:
Per the discussion offline and the PR #1282, closing this.

PR #1146, opened 2025-04-18
[profile] add profile for megatron train

BearBiscuit05:
###Motivation
This is a new feature that adds the functionality of collecting profiles during the training phase. Since the RL  process repeatedly enters the training process, by default, the profile temporarily captures the results of the first `update_policy`. Moreover, this modification should be seamlessly integrated into other training frameworks.

### DIFF
--- a/verl/trainer/config/ppo_megatron_trainer.yaml
+++ b/verl/trainer/config/ppo_megatron_trainer.yaml
@@ -74,6 +74,13 @@ actor_rollout_ref:
       use_dist_checkpointing: False
       dist_checkpointing_path: null
       seed: 1
+    profile:
+      use_profile: False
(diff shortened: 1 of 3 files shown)


verl/trainer/config/ppo_megatron_trainer.yaml
```diff
@@ -73,6 +73,12 @@ actor_rollout_ref:
       use_dist_checkpointing: False
       dist_checkpointing_path: null
       seed: 1
+    profile:
```
eric-haibin-lin: Please add comments to new options

BearBiscuit05: Got it, I have added more comments.

Issue #968, opened 2025-04-08
[FR] Filtration/Option for (validation) metrics


tongyx361:
### Summary

Request from https://github.com/volcengine/verl/pull/927#issuecomment-2781407256.

We might need some filteration/option to keep the metrics succinct.

zhongmingyuan:
@tongyx361 Not sure if I can contribute to this as a learning exercise. Thanks

tongyx361:
> [@tongyx361](https://github.com/tongyx361) Not sure if I can contribute to this as a learning exercise. Thanks

Of course! This is a call for contribution and also a good first issue. Welcome to have a try!

tongyx361:
> [@tongyx361](https://github.com/tongyx361) is it preferable for this to be a cmdline flag or an env var?

I think we config via the hydra yaml config files and CLI overriding by default.

PR #1159, opened 2025-04-19
FIx: filteration/option to keep the metrics succinct.

Hecate0821:
Fix for #968


Ghd-2077:
When I was reviewing the latest version of the source code, I didn’t come across this change. Do you happen to know why that might be?

PR #1162, opened 2025-04-19
[mcore] add offload param and opt function for magetron

BearBiscuit05:
## Motivation
This is a PR that supports offload in Megatron. Currently, parameters, gradients, and optimizers can be offloaded to the CPU when not needed. I have successfully tested the feasibility of the function using the memory snap tool. Further accuracy testing is still in progress.

## TODO
- [ ] Accuracy testing


BearBiscuit05:
> thanks for great work and will finish the testing on serval larger models like Qwen72b, Mixtral etc by the end of this week. cc: @ann-qin-lu

thx, but please note that there is currently an issue with the logger printing for GPU memory, mainly because vllm's memory release cannot be detected by torch. If you want to fix the issue, you can check PR #1118

ccclyu (approved):
LGTM. can first merge for feature dev.

PR #1164, opened 2025-04-19
add torch_compile  param for ref model

as12138:


as12138:
Configuring the torch_compile parameter of the ref model can make it easier for users to compare the differences between graph mode and non graph mode

verl/trainer/config/ppo_trainer.yaml
```diff
@@ -75,6 +75,7 @@ actor_rollout_ref:
       wrap_policy:
         # transformer_layer_cls_to_wrap: None
         min_num_params: 0
+    use_torch_compile: ${actor_rollout_ref.actor.use_torch_compile}
```
eric-haibin-lin, line 78: thanks for the PR. would u mind sharing how much perf improvement did you observe?

sunyi0505: self.compute_entropy_from_logits = (
            torch.compile(verl_F.entropy_from_logits, dynamic=True)
            if self.config.get('use_torch_compile', True) 
            else verl_F.entropy_from_logits)  
I tested this use case on A100 80G, and the graph mode showed a performance improvement of approximately 2-3 times compared to the eager mode

hiyouga (approved):

PR #1177, opened 2025-04-20
refactor: use 'from e' for exception stack trace

YeonwooSung:
Use the 'from e' inside the try-except statement to keep the stack trace of the error


eric-haibin-lin (approved):
Thx for the contribution!

PR #1180, opened 2025-04-20
[dev] fix: instructions about merging from before using ruff

tongyx361:
Our pre-commit hook and CI action only check the changes for now.

In this PR,

1. We apply `ruff check --fix` and `ruff format`.
2. We remove the unnecessary pipeline from the immigration warning, since directly merging without applying `ruff`, which might cause extra conflicts, is the best way to avoid introducing extra file changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -207,9 +207,8 @@ Contributions from the community are welcome! Please check out our [project road
 > [!WARNING]
 > We are [immigrating to `ruff` as the linter and formatter and `pre-commit` as the managing tool](https://github.com/volcengine/verl/pull/1010).
 >
-> If your branch is based on a previous commit using `yapf` and `pylint`,
-> please try applying `ruff check --fix` and `ruff format` before merging
-> to minimize the linting errors.
+> You are only expected to fix the linting errors in the files you changed.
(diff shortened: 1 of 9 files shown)


eric-haibin-lin:
does ruff check still produces lots of errors that should have been ignored?

tongyx361:
> does ruff check still produces lots of errors that should have been ignored?

No, most of the errors `ruff check` reports are reasonable, but it's difficult to fix all of them at once, so we only request the contributors to fix linting errors only in the files they changed.

This PR just fixes the instructions to help contributors restrict linting within the files they changed.

eric-haibin-lin:
Can u put something like "you are only expected to fix lint errors related to your changes while pre-commit checks the whole codebase"?

tongyx361:
> Can u put something like "you are only expected to fix lint errors related to your changes while pre-commit checks the whole codebase"?

Our pre-commit hook and CI action actually only check the changes.

This PR is mainly for correction that directly merging without applying `ruff`, which might cause extra conflicts, is the best way to avoid introducing other file changes.

eric-haibin-lin (approved):

PR #1184, opened 2025-04-21
[SGLang] Add support for IPv6

BearBiscuit05:
## Motivation
In the current usage of SGLang, if verl is under an IPv6 network, it will cause an address construction error, leading to a failure. This is because we need to handle the address based on IP network judgment.


hiyouga (approved):

BearBiscuit05:
Implemented at #1203

PR #1198, opened 2025-04-22
[llama megatron] support: ring attention

naturespeaker:
The code is functional in an environment where flash-attn (version >= v2) is installed and enabled. Without flash-attn, the code will fail to run due to the absence of the RoPE implementation. This file implements the Ring Attention functionality that supports unpadded inputs, but it strongly depends on the flash-attn library.


ETOgaosion:
As now we only maintain the GPTModel based mcore implementation, hand-written llama and qwen architecutures have been deprecated. Still, appreciate for your efforts, you are free to leave a link to your repo for users who hope to try without flash attn support~

Issue #1134, opened 2025-04-17
hf_rollout has bug when rollout.n > 1


QAQdev:
https://github.com/volcengine/verl/blob/19d0d0732920eb1717e191e40a27eb0af16f170b/verl/trainer/ppo/ray_trainer.py#L850-L852

When using `hf_rollout`, `gen_batch_output` has the same batch size as train batch size, which seems does not rollout for n times.

https://github.com/volcengine/verl/blob/19d0d0732920eb1717e191e40a27eb0af16f170b/verl/trainer/ppo/ray_trainer.py#L872-L874

So, when repeat train batch and union, error occurs

0x404:
Yes, there is an inconsistency in `hf_rollout`. In `vllm_rollout` or `sglang_rollout`, the `SamplingParams` is initialized using the `actor_rollout_ref.rollout` config, where we set `sampling_params.n` to `actor_rollout_ref.rollout.n`, and the generated result `gen_batch_output` shape should be the same as `batch` repeated `actor_rollout_ref.rollout.n` times.

In `hf_rollout`, the `GenerationConfig.num_return_sequences` is not overridden by `actor_rollout_ref.rollout.n`, which leads to this error.

QAQdev:
It is very nice of you to fix this! I will try it

Kirrito-k423:
中文：这个问题看起来已经由 #1199 修复，并已合并到主分支。维护者此前确认 `hf_rollout` 对 `rollout.n > 1` 的处理与 vllm/sglang 不一致；#1199 针对该问题更新了 `hf_rollout` 并添加了 `n=2` 的验证测试。当前 `main` 中，训练路径会在调用 rollout 前按 `rollout.n` repeat `gen_batch`，生成后再 repeat 原 batch 并 union；验证路径也会按 `val_kwargs.n` repeat test batch。因此该 issue 可以作为已修复关闭。

English: This appears to have been fixed by #1199, which has already been merged into the main branch. The maintainer previously confirmed that `hf_rollout` handled `rollout.n > 1` inconsistently with vllm/sglang; #1199 updated `hf_rollout` and added an `n=2` validation test. On current `main`, the training path repeats `gen_batch` by `rollout.n` before rollout and then repeats the original batch before union; the validation path similarly repeats the test batch by `val_kwargs.n`. So this issue can be closed as fixed.

This response was generated by a scheduled Codex GPT5.5 issue processing task.

PR #1199, opened 2025-04-22
feat: support rollout.n > 1 in hf_rollout

0x404:
Currently, the hf rollout backend only support `rollout.n == 1`, when `rollout.n > 1` it will lead to an error (https://github.com/volcengine/verl/issues/1134)

This PR make hf rollout support `do_sample` and `is_validate` to make it consistent with vllm and sglang backend, and correctly support `rollout.n > 1`.

### DIFF
--- a/tests/rollout/test_hf_rollout.py
+++ b/tests/rollout/test_hf_rollout.py
@@ -1,0 +1,180 @@
+# Copyright 2024 Bytedance Ltd. and/or its affiliates
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


vermouth1992 (approved):

eric-haibin-lin (commented):
would u mind add some example usage in tests/rollout?

0x404:
> would u mind add some example usage in tests/rollout?

@eric-haibin-lin Sure, I have added a test to test hf rollout, please have a check.

It seems only vllm rollout are tested in CI now, should we also add sglang and hf rollout into CI?

0x404:
> yeah a minimal test would be great

Currently we have `tests/rollout/test_sglang_spmd.py` and `tests/rollout/test_hf_rollout.py` added in this PR. Are these two tests suitable to add to our CI, or should we just keep them as they are?

eric-haibin-lin:
leaving them in this PR is fine. Thank you for contributing!

eric-haibin-lin (approved):

PR #1206, opened 2025-04-22
Add ray-based distributed reward manager

rawsh:
Add ray reward manager to support cross-node parallelization for verification

Tested with math-verify, dropped verification time to ~3s consistently for 128 x 16 = 2048 generations and ~30s for 256 x 64 = 16384 generations, with naive reward manager taking 200+ seconds


rawsh:
apologies for the update noise, cleaned it up a good bit

rawsh:
I think this approach is cleaner https://github.com/volcengine/verl/pull/1693

PR #1207, opened 2025-04-22
docker: update Dockerfile.sglang

BaiqingL:
Install ray[default] to include missing components


zhaochenyang20:
@ocss884 take a look plz

PR #1212, opened 2025-04-23
Feat/memory optimized loss

mingruimingrui:
## What's new?
Fused entropy & log probs function.

## How does it work?
The method to reduce memory cost during loss calculations is to not materialize the logits tensor.
To demonstrate, let's suppose you have the following configs
- token_per_batch = 32k
- vocab_size = 128k

Then assuming that we keep logits in fp16, we will have

logits_memory_size = 32k * 128k * 2 / 1e6 = 8GB

But wait... during loss calculations, we usually have to calculate probabilities, then when doing backwards, we have to store gradients. So the peak memory consumption will be 32GB! Then if we want better accuracy, we need to double that to 64GB.

To solve the problem of high memory consumption, we can use chunking. Specifically we chunk on the sequence length dimension. Calculate the forward and backward for each chunk then accumulate the gradients.

Suppose we use chunk_size = 512 and fp32 for loss calculations, we will have

peak_memory_usage = 512 * 128k * 4 / 1e6 * 4 ~= 1GB

This way we massively save on memory consumption

## Changes
- Added the options `return_last_hidden_state` and `use_fused_loss`
- monkey patch to make `model.forward` return `last_hidden_state` and not calculate `logits`
- Fused entropy and token_log_probs functions in verl/utils/experimental/torch_functional.py

### DIFF
--- a/verl/utils/experimental/torch_functional.py
+++ b/verl/utils/experimental/torch_functional.py
@@ -1,9 +1,14 @@
+from typing import Tuple
 import torch
 
 
 class FusedEntropy(torch.autograd.Function):
     @staticmethod
-    def entropy_fn(hidden_states: torch.Tensor, vocab_weights: torch.Tensor) -> torch.Tensor:
+    def entropy_fn(
+        hidden_states: torch.Tensor,
+        vocab_weights: torch.Tensor,
+        temperature: float = 1.0,
+    ) -> torch.Tensor:
         """Naive entropy function
 
         Args:
@@ -15,6 +20,7 @@ class FusedEntropy(torch.autograd.Function):
         """
         output_dtype = hidden_states.dtype
         logits = torch.einsum("bth,vh->btv", hidden_states, vocab_weights)
+        logits.div_(temperature)
         logits = logits.to(torch.float32)
         pd = torch.nn.functional.softmax(logits, dim=-1)
         entropy = torch.logsumexp(logits, dim=-1) - torch.sum(pd * logits, dim=-1)
@@ -25,6 +31,7 @@ class FusedEntropy(torch.autograd.Function):
         ctx,
         hidden_states: torch.Tensor,
         vocab_weights: torch.Tensor,
+        temperature: float = 1.0,
         chunk_size: int = 512,
     ) -> torch.Tensor:
         B, T, _ = hidden_states.shape
@@ -40,19 +47,25 @@ class FusedEntropy(torch.autograd.Function):
             chunk_end = min(chunk_start + chunk_size, T)
             chunk_hidden = hidden_states[:, chunk_start:chunk_end, :]
 
-            chunk_entropy = FusedEntropy.entropy_fn(chunk_hidden, vocab_weights)
+            chunk_entropy = FusedEntropy.entropy_fn(
+                chunk_hidden,
+                vocab_weights,
+                temperature=temperature,
+            )
 
             entropy[:, chunk_start:chunk_end] = chunk_entropy
 
         # Save necessary tensors for backward
         ctx.save_for_backward(hidden_states, vocab_weights)
         ctx.chunk_size = chunk_size
+        ctx.temperature = temperature
 
         return entropy
 
     @staticmethod
     def backward(ctx, grad_output):
         hidden_states, vocab_weights = ctx.saved_tensors
+        temperature = ctx.temperature
         chunk_size = ctx.chunk_size
         B, T, H = hidden_states.shape
         V = vocab_weights.shape[0]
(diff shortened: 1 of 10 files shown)


hiyouga:
Could you please look the failed tests?

hiyouga (commented):
It looks great to me. Wait for ci check passed. It would be better to provide some experiemental results with this PR

mingruimingrui:
+1 I think I'm still a bit far from merging.
Definitely need some use case where the response is super long to warrant such a feature.

hiyouga:
doc is also needed

mingruimingrui:
@vermouth1992 I've made 2 test scripts
fused enropy test: https://gist.github.com/mingruimingrui/8284bf2793b58ffbb5a02e0903e3ae30
fused log probs test: https://gist.github.com/mingruimingrui/c8db66b636ae99e6d1c0ebc36b772635

Here's the results

```
python test_fused_log_probs.py \
    --dtype bfloat16 \
    --device cuda \
    --batch_size 16

>>> Doing inference for the last hidden state
>>> last_hidden_state.shape=torch.Size([1, 21274, 896])
>>> vocab_weights.shape=torch.Size([151936, 896])
>>> lp_is_close=tensor(True, device='cuda:0') dh_is_close=tensor(True, device='cuda:0') dv_is_close=tensor(False, device='cuda:0')
>>> log_probs similarity: 1.0
>>> hidden_states grad similarity: 1.0078125
>>> vocab_weights grad similarity: 1.0

python test_fused_log_probs.py \
    --dtype bfloat16 \
    --device cuda \
    --batch_size 16 \
    --only_do_naive \
    --print_cuda_memory_stats

>>> Max memory allocated 43.67GB

python test_fused_log_probs.py \
    --dtype bfloat16 \
    --device cuda \
    --batch_size 16 \
    --only_do_fused \
    --print_cuda_memory_stats

>>> Max memory allocated 3.91GB
```

```
python test_fused_entropy.py \
    --dtype bfloat16 \
    --device cuda \
    --batch_size 8

>> Doing inference for the last hidden state
>> last_hidden_state.shape=torch.Size([1, 11674, 896])
>> vocab_weights.shape=torch.Size([151936, 896])
>> ent_is_close=tensor(True, device='cuda:0') dh_is_close=tensor(True, device='cuda:0') dv_is_close=tensor(False, device='cuda:0')
>> entropy similarity: 1.0
>> hidden_states grad similarity: 0.99609375
>> vocab_weights grad similarity: 1.0

python test_fused_entropy.py \
    --dtype bfloat16 \
    --device cuda \
    --batch_size 8 \
    --only_do_naive \
    --print_cuda_memory_stats

>>> Max memory allocated 50.94GB

python test_fused_entropy.py \
    --dtype bfloat16 \
    --device cuda \
    --batch_size 8 \
    --only_do_fused \
    --print_cuda_memory_stats

>>> Max memory allocated 3.47GB
```

mingruimingrui:
Next up I think I need a end-to-end training example.

hiyouga:
Please see the failed test

hiyouga:
e2e test is also needed

mingruimingrui:
@hiyouga Here's the results
- test scores and losses are about the same (not significantly different)
- actor update MFU took a hit from 0.365 -> 0.343 which is about a 5% difference
- overall training time is also slightly slower
  - 3h26m19s -> 3h27m29s (0.5% slower)
  - on average 3s slower per step
- peak memory usage is lower 88GB -> 72.4GB (this seems wrong)

hiyouga:
@vermouth1992 @eric-haibin-lin this PR is ready for merge, what do you think?

hiyouga (approved):
LGTM, this PR indeed reduces the VRAM usage

eric-haibin-lin (commented):
Please update the PR description following https://github.com/volcengine/verl/blob/main/.github/PULL_REQUEST_TEMPLATE.md

mingruimingrui:
> Please update the PR description following https://github.com/volcengine/verl/blob/main/.github/PULL_REQUEST_TEMPLATE.md

Done, please check

mingruimingrui:
@eric-haibin-lin @hiyouga another visualization for you guys, I added torch.cuda.memory._record_memory_history() to update_policy of dp_actor.py and got these diagrams

Without fused losses
image

With fused losses
image

The memory spikes due to loss computation are completely eliminated.

verl/utils/experimental/torch_functional.py
```diff
@@ -0,0 +1,317 @@
...
+        return (
+            grad_hidden,  # hidden_states
+            grad_vocab,  # vocab_weights
+            None,  # input_ids
+            None,  # temperature
+            None,  # chunk_size
+        )
+
+
+def fused_log_probs(
```
vermouth1992: Could you add a test of this function?

vermouth1992:
Could you merge main to resolve the conflict? Thanks.

PR #1225, opened 2025-04-23
tracker average token lengths for both positive and negative samples

paulinebourigault:


paulinebourigault:
Example plots:
neg_pos

verl/trainer/ppo/reward.py
```diff
@@ -85,32 +85,34 @@ def load_reward_manager(config, tokenizer, num_examine, **reward_kwargs):
     )
 
 
-def compute_reward(data: DataProto, reward_fn):
+def compute_reward(data: DataProto, reward_fn, tracker=None, step=None):
```
vermouth1992: Instead of passing tracker and step to the reward function, we can calculate the info using the full batch on driver and directly log to the tracker

verl/trainer/ppo/ray_trainer.py
```diff
@@ -969,9 +972,9 @@ def fit(self):
                             batch = batch.union(reward_tensor)
 
                         if self.config.reward_model.launch_reward_fn_async:
-                            future_reward = compute_reward_async.remote(batch, self.config, self.tokenizer)
+                            future_reward = compute_reward_async.remote(batch, self.config, self.tokenizer, tracker=self.tracker, step=self.global_steps)
                         else:
-                            reward_tensor, reward_extra_infos_dict = compute_reward(batch, self.reward_fn)
+                            reward_tensor, reward_extra_infos_dict = compute_reward(batch, self.reward_fn, tracker=self.tracker, step=self.global_steps)
```
vermouth1992: Return all the necessary information to calculate positive/negative seqlen, and then add to logger here.

paulinebourigault:
Thank you for the comments, I simplified
Now to use this, one would need to add to a training script for example:
`trainer.track_token_lengths=true \`
`trainer.positive_reward_threshold=0.5 \`
`trainer.token_length_tag_prefix="test-1.0" \`

patrik-bartak:
Wouldn't it make more sense to put this to the end of the fit function where the metrics dict is being updated?

verl/trainer/ppo/ray_trainer.py
```diff
@@ -1006,6 +1058,8 @@ def fit(self):
                             reward_tensor, reward_extra_infos_dict = ray.get(future_reward)
                         batch.batch["token_level_scores"] = reward_tensor
 
+                        self.log_token_lengths(batch, self.global_steps)
```
mertunsall, line 1061: there must be a `if` statement here to turn logging on and off, right?

verl/trainer/config/ppo_trainer.yaml
```diff
@@ -210,6 +210,9 @@ trainer:
   nnodes: 1
   n_gpus_per_node: 8
   save_freq: -1
+  track_token_lengths: true
+  positive_reward_threshold: 0.5
+  token_length_tag_prefix: "test-1.0"
```
mertunsall: maybe change the default to something else?

paulinebourigault: changed default to token_length

verl/trainer/ppo/ray_trainer.py
paulinebourigault: there is the condition at the beginning of the log_token_lengths method to check if token length tracking is enabled
`if not self.config.trainer.get('track_token_lengths', False):`
`    return`

mertunsall:
Hi @hiyouga, what do you think about merging this? The implementation is quite simplified now.

PR #1227, opened 2025-04-23
[FIX] fix device in sglang broadcast

coding-famer:
Based on the new `force_cpu_device` flag in [sglang's recent changes](https://github.com/sgl-project/sglang/pull/5416)


BearBiscuit05:
I think it has been fixed

Issue #1167, opened 2025-04-19
Ref Model Sync


mertunsall:
I realized that there is no functionality to sync reference model after certain number of iterations - didn't this cause any problems in long trainings for DAPO, ReTool, etc.? 

@eric-haibin-lin We are hoping to make a PR for this soon - we saw that this would cause our model to plateau during training after a while as KL divergence got too large.

gameofdimension:
https://github.com/volcengine/verl/pull/794 
maybe you can take a look at this pr

Superskyyy:
Recent approaches all tend to choose disable ref model and therefore this is not a problem.

thibautbar:
We did a small PR here to solve that issue: https://github.com/volcengine/verl/pull/1231
Let us know what you think

Kirrito-k423:
### Scheduled issue cleanup / 定时 issue 清理

中文：该 issue 命中定时清理规则：最近一次非当前运行账号的公开回复或引用已经超过 6 个月。因此按定时清理标准先关闭。若作者或社区成员仍能提供新的复现、证据或需求背景，欢迎继续评论请求重开，或创建带有最新信息的新 issue。

证据：
- Issue: https://github.com/verl-project/verl/issues/1167
- 最近一次非当前运行账号的公开回复或公开引用是 `2025-04-23T23:15:57Z` UTC（comment by @thibautbar），截至 2026-07-05 已超过 6 个月。
- 已检查 comments 来源：https://api.github.com/repos/verl-project/verl/issues/1167/comments
- 已检查 timeline 来源：https://api.github.com/repos/verl-project/verl/issues/1167/timeline

English: This issue matches the scheduled cleanup rule: the latest public reply or reference by an account other than the current runner is older than 6 months. Therefore, this is being closed under the scheduled cleanup policy. If the author or community members can still provide a fresh reproduction, evidence, or requirement context, please comment to request reopening or open a new issue with updated information.

Evidence:
- Issue: https://github.com/verl-project/verl/issues/1167
- The latest public reply or public reference from an account other than the current runner was `2025-04-23T23:15:57Z` UTC (comment by @thibautbar), which is more than 6 months before 2026-07-05.
- Checked comments source: https://api.github.com/repos/verl-project/verl/issues/1167/comments
- Checked timeline source: https://api.github.com/repos/verl-project/verl/issues/1167/timeline

This response was generated by a scheduled Codex GPT5.5 issue processing task.

PR #1231, opened 2025-04-23
[feat] PPO: synchronise the reference model with the actor model every X steps

thibautbar:
It can improve the training to synchronise the reference model every X steps, as described in this issue: https://github.com/volcengine/verl/issues/1167

In this PR we are adding the following:
- An option to sync the ref with actor every X steps (it can be weighted using alpha)
- If the reference model is sync with the actor, the model is saved / loaded using a checkpoint manager as for the actor

The solution here is to add a method to get the state_dict from the actor worker and to pass it to the reference worker


verl/workers/fsdp_workers.py
```diff
@@ -640,36 +649,103 @@ def compute_ref_log_prob(self, data: DataProto):
...
+        # Step 1: Force full state dict mode
+        state_dict_cfg = ShardedStateDictConfig(offload_to_cpu=True)
+
+        with FSDP.state_dict_type(self.actor_module_fsdp, StateDictType.SHARDED_STATE_DICT, state_dict_cfg):
+            model_state_dict = self.actor_module_fsdp.state_dict()
+
+        if self._is_offload_param:
+            offload_fsdp_model_to_cpu(self.actor_module_fsdp)
+
+        return self.rank, model_state_dict
```
vermouth1992, line 756: The model_state_dict can be huge. We can return it directly to the driver. Actually, if we colocate actor and ref in the same worker, we can directly operate on the weight

verl/trainer/ppo/ray_trainer.py
```diff
@@ -1027,6 +1043,17 @@ def fit(self):
...
+                    if (
+                        self.use_reference_policy
+                        and self.config.actor_rollout_ref.ref.sync_actor
+                        and self.global_steps % self.config.actor_rollout_ref.ref.sync_steps == 0
+                    ):
+                        with _timer("update_ref", timing_raw):
+                            self.ref_policy_wg.sync_with_actor(
+                                self.actor_rollout_wg.export_actor_weights(),
+                                self.config.actor_rollout_ref.ref.sync_actor_alpha,
+
```
jychen630, line 1056: This line looks like a syntax error: the comma on the line 1056 should be a closing bracket.

PR #1247, opened 2025-04-24
[FIX] Add support for reward_score dict return in prime reward manager

patrik-bartak:
Currently, the prime reward manager is outdated because it does not support dict return types from the reward score. This PR adds support for it, making it more similar to the naive reward manager.


patrik-bartak:
I looks like this is already addressed by #1070

PR #1248, opened 2025-04-24
[FIX] metric_utils log best, worst, maj only for n_resps > 1

patrik-bartak:
Solves #1249

Instead of logging best@1/mean and worst@1/mean, which is identical to mean@1, just do not log it when there is only one validation response per prompt (`n_resps == 1`). Same applies to std. 

Otherwise we get many duplicated plots that show the same thing. 

The only change is the addition of the `if n_resps > 1:` statement.


tongyx361 (dismissed):

ZihengJiang:
Please resolve the conflicts

patrik-bartak:
I resolved the conflicts. Like before, the only change should be the addition of the `if` statement to prevent redundant metrics being logged

tongyx361:
Nice job! Thanks for your contribution!

PR #1250, opened 2025-04-24
[FEAT] Add step and epoch metrics

patrik-bartak:
Solves #1251

Right now the current global step and current epoch are not being logged. This would be a useful feature.


eric-haibin-lin (approved):
thanks!

PR #1253, opened 2025-04-24
clean packages.

zhanluxianshen:


requirements.txt
```diff
@@ -8,7 +8,6 @@ hydra-core
 liger-kernel
 numpy
 pandas
```
eric-haibin-lin, line 10: we still need it. e.g. 
```
examples/data_preprocess/geo3k.py:import datasets
```
is it already a dependency of other pkgs ?

zhanluxianshen: I know we need it , i clean it , because it is duplicate package, you can see the code above.

zhanluxianshen: like line 4, has imported package:  ```datasets```

vermouth1992 (approved):

eric-haibin-lin (approved):

zhanluxianshen:
Thanks review.

Issue #1216, opened 2025-04-23
how to support multi-turn online rollout generation?


yuleiqin:
https://retool-rl.github.io/

The recently released ReTool project is a great work that introduces online multi-turn rollout generation which could optimizes the LLMs in an end-to-end manner.
To realize this,
we need to allow the rollout process to support multi-turn generation, where the LLM first generates the function call and waits for the observation/feedback from the tool calling results, and then continues to generate the second response.

Is there any detailed date/plan for supporting this feature?

Thank you in advance for your work.

OswaldoBornemann:
+1

lxg2015:
Hi @yuleiqin @OswaldoBornemann  I have implemented this feature at https://github.com/volcengine/verl/pull/1258， Does this meet the requirements?

Kirrito-k423:
### Scheduled issue cleanup / 定时 issue 清理

中文：该 issue 命中定时清理规则：最近一次非当前运行账号的公开回复或引用已经超过 6 个月。因此按定时清理标准先关闭。若作者或社区成员仍能提供新的复现、证据或需求背景，欢迎继续评论请求重开，或创建带有最新信息的新 issue。

证据：
- Issue: https://github.com/verl-project/verl/issues/1216
- 最近一次非当前运行账号的公开回复或公开引用是 `2025-04-25T09:30:56Z` UTC（mentioned by @OswaldoBornemann），截至 2026-07-05 已超过 6 个月。
- 已检查 comments 来源：https://api.github.com/repos/verl-project/verl/issues/1216/comments
- 已检查 timeline 来源：https://api.github.com/repos/verl-project/verl/issues/1216/timeline

English: This issue matches the scheduled cleanup rule: the latest public reply or reference by an account other than the current runner is older than 6 months. Therefore, this is being closed under the scheduled cleanup policy. If the author or community members can still provide a fresh reproduction, evidence, or requirement context, please comment to request reopening or open a new issue with updated information.

Evidence:
- Issue: https://github.com/verl-project/verl/issues/1216
- The latest public reply or public reference from an account other than the current runner was `2025-04-25T09:30:56Z` UTC (mentioned by @OswaldoBornemann), which is more than 6 months before 2026-07-05.
- Checked comments source: https://api.github.com/repos/verl-project/verl/issues/1216/comments
- Checked timeline source: https://api.github.com/repos/verl-project/verl/issues/1216/timeline

This response was generated by a scheduled Codex GPT5.5 issue processing task.

PR #1258, opened 2025-04-25
support multi-turn online rollout generation and tool calling

lxg2015:
## Support multi-turn rollout and tool calling

allow the rollout process to support multi-turn generation, where the LLM first generates the function call and waits for the observation/feedback from the tool calling results, and then continues to generate the second response. https://github.com/volcengine/verl/issues/1216

## Key Features
- support early stop for vllm generate
- support early stop for multi-turn rollout
- support online tool calling
- create loss mask for tool calling info 

## Exampels
```
step 0 batch, init prompt 
[
'蓝色养乐多',
 '1+1'
]

step 1 batch, init prompt with llm rollout and tool calling info
 [
'蓝色养乐多<search>蓝色养乐多产品信息</search>\n\n<information>工具调用(蓝色养乐多产品信息)</information>\n\n\n'
]

final multi-turn result:
 [
'蓝色养乐多<search>蓝色养乐多产品信息</search>\n\n<information>工具调用(蓝色养乐多产品信息)</information>\n\n\n<answer>蓝色养乐多通常指的是一种低糖版本的养乐多饮料。</answer>', 
'1+1<answer>1+1=2</answer>'
]
```


vermouth1992:
Thank you for your wonderful work! However, we have integrated AsyncLLM into verl rollout, that supports multi-turn natively. By doing so, we don't need to manually interact with the tool in batch manner. https://github.com/volcengine/verl/pull/1138

mertunsall:
Can you explain a little bit how to achieve this with AsyncLLM - interface seems more cumbersome @vermouth1992

vermouth1992:
> Can you explain a little bit how to achieve this with AsyncLLM - interface seems more cumbersome @vermouth1992

Hi @mertunsall,

We can use OpenAI-compatible server to iterate with vllm to generate rollouts just like online serving, where tool calling is supported by default. We will give some examples later. Stay tuned!

mertunsall:
I see what you mean - I understand that tools will be executed on the “server” side rather than on client, right?

vermouth1992:
> I see what you mean - I understand that tools will be executed on the “server” side rather than on client, right?

I guess it can happen on both side. If tool calling happens on "server" side, users can be freed from calling logics. If tool calling happens on the "client" side, it's more flexible to implement customized logics, such as parallel tool calling or advanced post processing.

verl/utils/multi_turn/loop_process.py
```diff
@@ -0,0 +1,292 @@
...
+        loss_mask_list = []
+        length_list = []
+        
+        for idx, responses in enumerate(self.loop_responses_token):
+            loss_mask = [[0] * len(responses[0])] # init_prompt loss
+            prompts_list = list(itertools.chain.from_iterable(responses[1:]))
+            # responses_token: [prompt_token, reponse_token_1, info_token_1, response_token_2....]
+            for turn_idx in range(len(responses[1:])): 
+                length = len(responses[turn_idx])
+                loss_mask.extend([turn_idx % 2] * length)
```
Zero1002, line 192: I noticed that when calculating loss_mask, the model response was set to 0 and the tool call was set to 1. Are the settings here reversed?

lxg2015: Yes, the settings are reversed. The response should be set to 1.

PR #1260, opened 2025-04-25
[single controller] feat: support directly register dispatch device mesh

vermouth1992:
The current dispatch methods are quite limited:
- In hybrid engine, we would like to dispatch infer_tp x infer_dp for generation and train_tp x train_dp x train_pp for training. However, currently implementation can only dispatch train_tp x train_dp x train_pp for training and dp for generation and perform allgather inside the workergroup.
- When two megatron models colocate, their device mesh has to be identical.
- We have to subclass RayWorkerGroup in order to implement various distributed strategies. This makes create_colocated_worker hacky to implement in the future.

The difference in this implementation:
- The nD device mesh is not obtained from worker, then broadcast to workergroup. It is directly registered to the worker group in driver. It's the user's responsibility to make sure that the registered device mesh matches with the device mesh established internally.

By doing so, we can completely remote MegatronWorker and the necessity to subclass RayWorkerGroup in the future to implement flexible dispatch methods.

Currently, we only support tp-cp-ep-dp-pp. This meets the current requirements of megatron and vllm/sglang. We can introduce more flexible device mesh like dispatch method in the future.

### DIFF
--- a/verl/single_controller/base/decorator.py
+++ b/verl/single_controller/base/decorator.py
@@ -318,6 +318,89 @@ def collect_dp_compute_data_proto(worker_group, output):
     output = collect_dp_compute(worker_group, output)
     return _concat_data_proto_or_future(output)
 
+from verl.single_controller.base.worker_group import WorkerGroup
+from verl.single_controller.base.worker import DistGlobalInfo, DistRankInfo
+
+
+def dispatch_nd_compute(device_mesh, worker_group, *args, **kwargs):
+    assert isinstance(worker_group, WorkerGroup)
+
+    assert device_mesh in worker_group._dist_global_info, f"expecting {device_mesh} in {worker_group._dist_global_info}"
+    assert device_mesh in worker_group._dist_workers_info, f"expecting {device_mesh} in {worker_group._dist_workers_info}"
+
+    dist_global_info: DistGlobalInfo = worker_group._dist_global_info[device_mesh]
+    dist_workers_info: list[DistRankInfo] = worker_group._dist_workers_info[device_mesh]
+    
+    all_args = []
+    for arg in args:
+        assert isinstance(arg, (Tuple, List)) and len(arg) == dist_global_info.dp_size
+        transformed_args = []
+        for i in range(worker_group.world_size):
+            local_dp_rank = dist_workers_info[i].dp_rank
+            transformed_args.append(arg[local_dp_rank])
+        all_args.append(transformed_args)
+    all_args = tuple(all_args)
+
+    all_kwargs = {}
+    for k, v in kwargs.items():
+        assert isinstance(v, (Tuple, List)) and len(v) == dist_global_info.dp_size
+        transformed_v = []
+        for i in range(worker_group.world_size):
+            local_dp_rank = dist_workers_info[i].dp_rank
+            transformed_v.append(v[local_dp_rank])
+        all_kwargs[k] = transformed_v
+    return all_args, all_kwargs
+
+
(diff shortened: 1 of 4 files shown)


verl/single_controller/base/decorator.py
```diff
@@ -319,6 +319,91 @@ def collect_dp_compute_data_proto(worker_group, output):
...
+
+    dist_global_info: DistGlobalInfo = worker_group._dist_global_info[mesh_name]
+    dist_workers_info: list[DistRankInfo] = worker_group._dist_workers_info[mesh_name]
+
+    output_in_dp = []
+    pp_size = dist_global_info.pp_size
+    for global_rank in range(worker_group.world_size):
+        local_rank_info = dist_workers_info[global_rank]
+        if local_rank_info.tp_rank == 0 and local_rank_info.pp_rank == pp_size - 1 and local_rank_info.cp_rank == 0 and local_rank_info.ep_rank == 0:
+            output_in_dp.append(output[global_rank])
```
wuxibin89, line 379: As an optimization, maybe we can set the result of workers other than tp=0,pp=pp_size-1,cp=0,ep=0 to None to reduce communication cost.

verl/single_controller/base/decorator.py
```diff
@@ -319,6 +319,91 @@ def collect_dp_compute_data_proto(worker_group, output):
...
+            local_dp_rank = dist_workers_info[i].dp_rank
+            transformed_args.append(arg[local_dp_rank])
+        all_args.append(transformed_args)
+    all_args = tuple(all_args)
+
+    all_kwargs = {}
+    for k, v in kwargs.items():
+        assert isinstance(v, (Tuple, List)) and len(v) == dist_global_info.dp_size
+        transformed_v = []
+        for i in range(worker_group.world_size):
```
wuxibin89, line 347: Since we use star topology, compare to all-gather in dp group, this amplify communication cost in single-controller.

verl/single_controller/base/decorator.py
vermouth1992: Can this be done from the driver side? Or user have to be aware of this?

verl/single_controller/base/decorator.py
vermouth1992: I guess it should be the same? Assume we have N data on driver, with TP + DP, where n_gpus = TP * DP. 
The current implementation:
- Split N with n_gpus, where each gpu obtains N/n_gpus data point. Then, perform allgather inside the worker. Each gpu finally obtains N/DP data.
New dispatch method:
- Split N with DP. Replicate the data in tp group to obtain the same dp data.

One missing optimization is to use ray.put to pickle the data in advance to avoid the same dp pickling the data tp times.

vermouth1992: One drawback of all-gather in tp group is that, N must be divisible by n_gpus, where in fact, N only needs to be divisible by DP. This limits the minimum number of data, which is quite annoying.

verl/single_controller/base/decorator.py
wuxibin89: It's better done in client side, we can set result to None if worker is not tp=0,pp=pp_size-1,cp=0,ep=0
https://github.com/volcengine/verl/blob/main/verl/single_controller/ray/base.py#L453

verl/single_controller/base/decorator.py
wuxibin89: The total communication cost in driver
- current implementation: (N/n_gpus) * n_gpus = N
- new implementation: (N/DP) * n_gpus = TP * N

Issue #390, opened 2025-02-26
[Question] How to use customized chat template


zpqiu:
What is the currently recommended method for customizing chat templates? I see there are two approaches, but both have some issues:

- In `rl_dataset.py`, it defaults to using the tokenizer's chat template to generate prompts, so one method is to modify the tokenizer's config.
- I notice that the `RLHFDataset` initialization function supports `chat_template_func`, but this function is not being used and there's no documentation or comments explaining how to define it?

patrik-bartak:
Some models, such as `deepseek-ai/deepseek-coder-1.3b-base`, do not support applying a chat template. When calling the function, I get 

```python
ValueError: Cannot use chat template functions because tokenizer.chat_template is not set and no template argument was passed! For information about writing templates and setting the tokenizer.chat_template attribute, please see the documentation at https://huggingface.co/docs/transformers/main/en/chat_templating
```

Because I need the ability to use models without a chat template, and it would be good to be able to overwrite the chat template, I will implement this.

PR #1261, opened 2025-04-25
Add custom chat template option

patrik-bartak:
#390 

Right now the RL and SFT datasets use the chat template of the tokenizer to format the prompt. Some models do not have a chat template (for example `deepseek-ai/deepseek-coder-1.3b-base`), and sometimes it can be useful to override this template. This adds the option to override the template when loading the tokenizer. 

I have added the required code to make this work with the RL dataset, SFT dataset, and main_ppo / main_ppo_split. If there are other parts of the code that should be updated to make this option work, let me know. 

More info here https://huggingface.co/docs/transformers/main/en/chat_templating

### DIFF
--- a/examples/split_placement/config/ppo_trainer_split.yaml
+++ b/examples/split_placement/config/ppo_trainer_split.yaml
@@ -3,6 +3,7 @@ data:
   train_files: ~/data/rlhf/gsm8k/train.parquet
   val_files: ~/data/rlhf/gsm8k/test.parquet
   prompt_key: prompt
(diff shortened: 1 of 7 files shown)


vermouth1992:
Could you help modify one test to use custom chat template? Maybe passing the chat template from the tokenizer just for testing?

patrik-bartak:
@vermouth1992 
I don't really see a test that I could modify. This PR just lets you override the chat_template using a key in the config.

PR #1267, opened 2025-04-26
Qwen2moe[part1]: add cpu converter option, add CI test for current solutions temporarily

ETOgaosion:
Temporarily use CPU to initialize larger models for huggingface to dist_ckpt conversion.

And Support GQA Moe model.

May not require CI as this function can be dependent to VeRL, but current solution may need.


vermouth1992:
Could you please fix the conflict?

scripts/converter_hf_to_mcore.py
```diff
@@ -169,7 +183,12 @@ def megatron_model_provider(pre_process, post_process):
     if test:
```
ccclyu, line 221: can we revise the `test` part like become a single function and change variable name to avoid `k` etc.

ccclyu:
Weight conversion using CPU looks good to me and I have tested the conversion on Mixtral8x22B and Mixtral8X7B. It can be merged after revising the test part.

scripts/converter_hf_to_mcore.py
ETOgaosion: OK, thanks for suggestion! Will refactor this part~

scripts/converter_hf_to_mcore.py
```diff
@@ -54,6 +54,37 @@ def __init__(self):
...
+            assert (dut_data == ref_data).all(), f"{name} is not equal"
+    for name in ref_state_dict.keys():
+        if ref_state_dict[name] is None:
+            continue
+        ref_data = ref_state_dict[name].data
+        if name in dut_state_dict:
+            dut_data = dut_state_dict[name].data
+            assert dut_data.shape == ref_data.shape, f"{name=} {dut_data.shape=} {ref_data.shape=}"
+            assert (dut_data == ref_data).all(), f"{name} is not equal"
+
```
ccclyu, line 95: can we add one log to indicate that the conversion testing is passed or similar statement?

ETOgaosion: Good suggestion.

PR #1269, opened 2025-04-26
[bugfix] fix: add `await` for  `_validate()`

nwiad:
As titled.


wuxibin89 (approved):

recipe/prime/prime_ray_trainer.py
```diff
@@ -501,7 +501,7 @@ async def fit(self):
                         and self.global_steps % self.config.trainer.test_freq == 0
                     ):
                         with _timer("testing", timing_raw):
-                            val_metrics: dict = self._validate()
+                            val_metrics: dict = await self._validate()
```
wuxibin89, line 504: #L523 should be also await

wuxibin89 (approved):

PR #1278, opened 2025-04-27
[feat] add FusedWorker

zTonyZhao:
on behalf of @zw0610 

FusedWorker is designed to enhance the ability of colocated workers.

FusedWorker keeps most of the interfaces as colocated workers: Users shall use `create_colocated_worker_cls_fused` to create colocated worker class, use `spawn` to split FusedWorker to dict of workers.

In colocated workers, access the methods of child workers is done by using `spawn` then access via worker dict or calling `{worker_group}.{worker}_{method}`. In FusedWorker, the first method was preserved, while the latter was change to a new way: First use `{worker_group}.fuse(prefixes)` to bind workers to the worker group, then use `{worker_group}.{worker}.foo()` to access child workers.


vermouth1992:
Could you write some documentation about this? Thanks.

zTonyZhao:
Usage of FusedWorker is added as comments in code.

vermouth1992:
It's weird that the CI can't be trigger..

vermouth1992:
Could you close it and resubmit it again?

PR #1282, opened 2025-04-28
[trainer] breaking: pass dataset as required args to SFTTrainer; also change ppo ray trainer to take custom datasets as inputs

eric-haibin-lin:
`FSDPSFTTrainer` now must take tokenizer, train_dataset and val_dataset as inputs. Example usage: 

```
    local_model_path = copy_to_local(src=config.model.partial_pretrain, verbose=True)
    tokenizer = hf_tokenizer(local_model_path, trust_remote_code=config.model.trust_remote_code)
    train_dataset = create_sft_dataset(config.data.train_files, config.data, tokenizer)
    val_dataset = create_sft_dataset(config.data.val_files, config.data, tokenizer)

    return FSDPSFTTrainer(config=config, device_mesh=device_mesh, ulysses_device_mesh=ulysses_device_mesh, tokenizer=tokenizer, train_dataset=train_dataset, val_dataset=val_dataset)
```

It makes it easier for others to customize the data loading logic for SFT & RL.

For the RL ray trainer, it takes a few additional inputs. If they are not provided, they are initialized the same as they were so the RayTrainer for RL is still backward compatible. 

```
        train_dataset: Optional[Dataset] = None,
        val_dataset: Optional[Dataset] = None,
        collate_fn=None,
        train_sampler: Optional[Sampler] = None,
```


tongyx361 (approved):
It seems reasonable to expose these simple interfaces to users for customization!

PR #1288, opened 2025-04-28
[FSDPCheckpointManager] feat: save huggingface model when 'hf_model' in checkpoint_contents

0x404:
Before, `FSDPCheckpointManager` will not save hf model when `hf_model` is given in `checkpoint_contents`, instead, it only save the hf model's config.

This PR correctly save the huggingface model when 'hf_model' is in `checkpoint_contents`.

### DIFF
--- a/verl/utils/checkpoint/fsdp_checkpoint_manager.py
+++ b/verl/utils/checkpoint/fsdp_checkpoint_manager.py
@@ -18,9 +18,9 @@ from typing import Optional, Union
 
 import torch
 import torch.distributed
+from torch.distributed.fsdp import FullStateDictConfig, ShardedOptimStateDictConfig, ShardedStateDictConfig, StateDictType
 from torch.distributed.fsdp import FullyShardedDataParallel as FSDP
-from torch.distributed.fsdp import ShardedOptimStateDictConfig, ShardedStateDictConfig, StateDictType
-from transformers import PreTrainedTokenizer, ProcessorMixin
+from transformers import AutoConfig, GenerationConfig, PreTrainedTokenizer, ProcessorMixin
 
 from verl.utils.fs import copy_to_local, is_non_local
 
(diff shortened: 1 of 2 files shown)


verl/utils/checkpoint/fsdp_checkpoint_manager.py
```diff
@@ -149,16 +151,54 @@ def save_checkpoint(self, local_path: str, hdfs_path: str = None, global_step: i
...
+
+                if save_model.can_generate():
+                    try:
+                        save_model.generation_config = GenerationConfig.from_pretrained(self.local_model_path)
+                    except OSError:
+                        print(f"Warning: {self.__class__.__name__}.save_checkpoint: Generation config file not found in {self.local_model_path}, using a generation config created from the model config when saving hf_model.")
+                        pass
+
+                save_model.save_pretrained(hf_local_path, state_dict=state_dict)
                 self.processing_class.save_pretrained(hf_local_path)
```
ETOgaosion, line 211: Here is also a bug! Could you please help to move tokenizer saving to default saving path above? Outside the `if "hf_model" in self.checkpoint_contents:` effective field.

Currently FSDP checkpoint do not save tokenizer without `"hf_model"` added, which is inconvenient for checkpoint resume.

Thanks a lot for help~

0x404: @ETOgaosion Hi, I have update the code to save tokenizer/processor in FSDP checkpoint dir.

I think we can keep this line of code here to also save tokenizer/processor in hf_model dir, since some user may want to direcly use `AutoTokenizer.from_pretrained("/path/to/hf_model")` and don't want to perform an extra copy operation.

0x404:
Hi @ETOgaosion, I have merged the main branch and resolve some conflicts since FSDP2 is merged. Please take a look when you have time :)

verl/utils/checkpoint/fsdp_checkpoint_manager.py
ETOgaosion: It's OK, I get it.

0x404:
looks like there are several CI failed due to `Error: Connection timed out (api.github.com:443)`

0x404:
> Seems here is a bug? https://github.com/volcengine/verl/actions/runs/14829253212/job/41690118956?pr=1288

Yes, in `test_fsdp_ckpt` the model is initialized using `from_config` thus it doesn't have the `name_or_path` and therefore it doesn't have the pretrained generation config. In this case, we don't need to save the generation config. I have fixed this and let's rerun the CI.

hiyouga (approved):

Issue #144, opened 2025-01-27
Additional memory optimization features


SalmanMohammadi:
- Activation offloading (see implementation [here](https://github.com/pytorch/torchtune/blob/main/torchtune/training/_activation_offloading.py))
- Fusing optimizer step into backward pass (see implementation [here](https://github.com/pytorch/torchtune/blob/main/torchtune/training/memory.py#L219))
- Utilize `full_shard` `reshard_after_forward` (see [here](https://github.com/pytorch/torchtitan/blob/main/docs/fsdp.md#fsdp1--fsdp2-api-differences)). I wasn't 100% sure if I could see this already implemented in veRL.

These optimizations largely trade off decreased peak memory useage for additional compute, so may only be useful for training larger models, and in GPU-constrained settings.

vermouth1992:
# 3 is already implemented in our FSDP1 implementation. We haven't tested FSDP2 for large scale training yet so it's a little bit risk to migrate.

eric-haibin-lin:
Thanks for sharing. Fusing optimizer step into backward pass sounds interesting!

tinuademargaret:
I just created a draft PR for fusing optimizer step into backward pass  [here](https://github.com/volcengine/verl/pull/1295) @eric-haibin-lin

Kirrito-k423:
These are great suggestions for memory optimization! Let me address each:

1. **Activation offloading**: The torchtune implementation you linked is a solid reference. veRL already uses activation checkpointing, but the offloading pattern (moving activations to CPU when not needed) would provide additional memory savings at the cost of some CPU-GPU transfer overhead.

2. **Fusing optimizer step into backward pass**: This is the "gradient accumulation" pattern where optimizer state is updated during the backward pass rather than in a separate optimizer step. This can reduce peak memory by not storing the full optimizer state simultaneously with gradients. The torchtune implementation at the link you provided would be a good reference.

3. **Full Shard with `reshard_after_forward`**: veRL does support FSDP HybridShard (`fsdp2` mode), but `reshard_after_forward=True` is indeed more memory-efficient as it only keeps the full model in the forward pass and redistributes after. You may want to check the latest veRL FSDP configuration options.

For a PR implementing these, the key files would be:
- `verl/trainer/ppo/fsdp_trainer.py` for FSDP config
- `verl/trainer/ppo/core_algos.py` for loss computation

Feel free to open a PR if you would like to contribute these optimizations!

yurun00:
[...]

PR #1295, opened 2025-04-28
Optimizer-in-backward Fusion Implementation

tinuademargaret:
This PR implements optimizer-in-backward fusion (suggested in this [issue](https://github.com/volcengine/verl/issues/144)), which integrates optimizer updates directly into the backward pass to decrease peak memory usage by eliminating the need to store gradients. It uses pytorch's  `_apply_optimizer_in_backward `to fuse optimizer steps with backpropagation.

This feature only works with the sft trainer. 

Optimizer-in-backward fusion is incompatible with gradient accumulation and is controlled by a `bwd_hook` flag in the optimizer config.

There is added validation to ensure that training configurations that rely on gradient accumulation does not enable the optimizer fusion option.


eric-haibin-lin:
thanks! Just to double check, this supports gradient accumulation?

tinuademargaret:
> thanks! Just to double check, this supports gradient accumulation?

 No, It uses pytorch's `apply_optimizer_in_backward`  which does not support gradient accumulation.

eric-haibin-lin (changes requested):
Thanks for the contribution! For RL i think it's very likely that we need gradient accumulation to train with large batch sizes. But this feature would still be useful for SFT. Do you think it make sense if you only keep the changes for SFT trainer, while keeping the RL trainer simple for now? Thx

PR #1296, opened 2025-04-28
Fix pickle error in prime reward manager when prime in main_ppo.py

### LINKED ISSUE #1293: Error: Can't pickle local object 'get_custom_reward_fn.<locals>.wrapped_fn In Prime Reward Manager
# Problem
Recently, I implemented a custom reward function that requests a sandbox to obtain scores. 

When using naive reward manager, the verl pipeline work properly. However, when switching to prime reward manager, the pipeline raise an error (Can't pickle local object 'get_custom_reward_fn.<locals>.wrapped_fn) and eventually hit OOM error.

I asked CHatGPT why verl raised this error. The answer is that in `parallel_compute_score_async` in Prime Reward Manager uses `ProcessPoolExecutor` to manage  parallel request. `ProcessPoolExecutor` rely on multiprocessing-when a function runs at Process level,all arguments must be picklable. In following case,  `evaluation_func, completion, reference, task, task_extra_info` need to be picklable, but `evaluation_func` failes  because it's a wrapper function.
```
async def parallel_compute_score_async(evaluation_func, completions, references, tasks, extra_info=None, num_processes=64):
    scores = []
    with ProcessPoolExecutor(max_workers=num_processes) as executor:
        if extra_info is None:
            extra_info = [None] * len(tasks)
        # Create tasks for all rows
        tasks_async = [single_compute_score(evaluation_func, completion, reference, task, task_extra_info, executor, timeout=300.0) for completion, reference, task, task_extra_info in zip(completions, references, tasks, extra_info)]
        # to prevent very occasional starvation caused by some anomalous programs ( like infinite loop ), the exceptions in async programs will instantly halt the evaluation, and all summoned processes will be killed.
        try:
            results = await asyncio.gather(*tasks_async, return_exceptions=False)
        except:
            for pid, proc in executor._processes.items():
                try:
                    proc.kill()
                except Exception as kill_err:
                    print("shut down failed: " + str(kill_err))
            raise

    # Process results
    for result, completion, reference, task in zip(results, completions, references, tasks):
        if isinstance(result, Exception) or result is None:
            # Handle failed or timed-out tasks
            scores.append(0.0)
        elif isinstance(result[0], (int, float, bool)):
            scores.append(float(result[0]))
        else:
            scores.append(float(result[0][0]))
    return scores
```

I looked through every issue, but mine seems unique. Guess I'm the first one to stumble into this! 😅

## My solution (Most Simple, Only for IO intensive)

I try to hack the code. 

The reason of multipreocessing must need pickle is that params need to  share in multi-processes. So, I directly use mem-shared———Threading (DO NOT consider pickle problem). I replace ProcessPoolExecutor to ThreadPoolExcecutor.

It's interesting, my code is running sucessfully.
```
from concurrent.futures import ThreadPoolExecutor
async def parallel_compute_score_async(evaluation_func, completions, references, tasks, extra_info=None, num_processes=64):
    scores = []
    with ThreadPoolExecutor(max_workers=num_processes) as executor:
        ...
```

## The best way to fix
The core issue is that `wrapped_fn` need to put at the top level, rather than function intern. So, We only modify return function wrap `verl/trainer/main_ppo.py`.
```
...
from functools import partial

def get_custom_reward_fn(config):
    import importlib.util, sys
    reward_fn_config = config.get("custom_reward_function") or {}
    file_path = reward_fn_config.get("path")
    if not file_path:
        return None

    if not os.path.exists(file_path):
        raise FileNotFoundError(f"Reward function file '{file_path}' not found.")

    spec = importlib.util.spec_from_file_location("custom_module", file_path)
    module = importlib.util.module_from_spec(spec)
    try:
        sys.modules["custom_module"] = module
        spec.loader.exec_module(module)
    except Exception as e:
        raise RuntimeError(f"Error loading module from '{file_path}': {e}")

    function_name = reward_fn_config.get("name")
    if not hasattr(module, function_name):
        raise AttributeError(f"Reward function '{function_name}' not found in '{file_path}'.")

    print(f"using customized reward function '{function_name}' from '{file_path}'")
    raw_fn = getattr(module, function_name)

    reward_kwargs = dict(reward_fn_config.get("reward_kwargs", {}))
    # ========modification===========
    #def wrapped_fn(*args, **kwargs):
    #    return raw_fn(*args, **kwargs, **reward_kwargs)

    #return wrapped_fn
    return partial(raw_fn, **reward_kwargs)

```
A detailed explanations: https://chatgpt.com/share/680f6ec1-55e0-8000-9575-b4d75d30154e

cgpeter96:
fixed https://github.com/volcengine/verl/issues/1293


verl/workers/reward_manager/prime.py
```diff
@@ -48,7 +48,7 @@ async def single_compute_score(evaluation_func, completion, reference, task, tas
 
 async def parallel_compute_score_async(evaluation_func, completions, references, tasks, extra_info=None, num_processes=64):
     scores = []
-    with ProcessPoolExecutor(max_workers=num_processes) as executor:
+    with ThreadPoolExecutor(max_workers=num_processes) as executor:
```
wuxibin89: ThreadPoolExecutor is not a good idea since compute_score is compute intensive, python GIL prevents it to use multi cores.

cgpeter96: > ThreadPoolExecutor is not a good idea since compute_score is compute intensive, python GIL prevents it to use multi cores.

Understood, I will update the code shortly.

patrik-bartak:
I actually ran into the same issue as you, and I think your solution is unnecessarily complex. I also got the issue that the function is not pickleable - the issue is that the `wrapped_fn` if defined within a function, and not top level. 

A simple solution that works for me is replacing the wrapped function with a partial over the raw function. 
```
...
    def wrapped_fn(*args, **kwargs):
        return raw_fn(*args, **kwargs, **reward_kwargs)

    return wrapped_fn
```
->
```
...
    return partial(raw_fn, **reward_kwargs)
```

No other changes are needed. Do you agree?

patrik-bartak:
Also, the function `get_custom_reward_fn` is duplicated across `main_dapo`, `main_eval`, `main_ppo`. If you want I can make a PR to share the implementation of this function.

vadimkantorov:
Hi @wuxibin89 @vermouth1992 ! We also stumbled in this issue with a custom reward function defined in a sidekick python file: `[Error] Task failed: Can't get local object 'get_custom_reward_fn.<locals>.wrapped_fn', completion: ...`

We tried the `partial(...)` solution but then hit another bug: `[Error] Task failed: Can't pickle <function compute_score at 0x7f23310b4ea0>: it's not the same object as custom_module.compute_score, completion: ...`

We're calling verl trainer as follows:
```shell
    custom_reward_function.path=/mnt/fs/verl/custom_function.py \
    custom_reward_function.name=compute_score \
    reward_model.reward_manager=prime \
    ...
```

Would you have any suggestions how to fix this?

zhourunlong:
> Hi @wuxibin89 @vermouth1992 ! We also stumbled in this issue with a custom reward function defined in a sidekick python file: `[Error] Task failed: Can't get local object 'get_custom_reward_fn.<locals>.wrapped_fn', completion: ...`
> 
> We tried the `partial(...)` solution but then hit another bug: `[Error] Task failed: Can't pickle <function compute_score at 0x7f23310b4ea0>: it's not the same object as custom_module.compute_score, completion: ...`
> 
> We're calling verl trainer as follows:
> 
> ```shell
> python -m verl.trainer.main_ppo \
>     custom_reward_function.path=/mnt/fs/verl/custom_function.py \
>     custom_reward_function.name=compute_score \
>     reward_model.reward_manager=prime \
>     ...
> ```
> 
> Would you have any suggestions how to fix this?

Meeting exactly the same issue.

george1459:
I ran into this issue as well (even with #2239). Here is [my fix for the issue](https://github.com/volcengine/verl/compare/main...george1459:verl:wip/fix_custom_reward_ray) which worked for me.
Let me know if you guys are interested in having this as a PR.
@vadimkantorov @zhourunlong perhaps you guys can give this a try if interested.

warmsnow-sh:
This code seems to be inconsistent with the latest version. Are there any latest solutions to this problem at present?

warmsnow-sh:
> I ran into this issue as well (even with #2239). Here is [my fix for the issue](https://github.com/volcengine/verl/compare/main...george1459:verl:wip/fix_custom_reward_ray) which worked for me. Let me know if you guys are interested in having this as a PR. @vadimkantorov @zhourunlong perhaps you guys can give this a try if interested.



I tried your plan, but it reported an error
(TaskRunner pid=1138472) [Error] Task failed: 'list' object has no attribute 'get',
I'm still checking if there are any other potential issues with my code. I wonder why this problem(prime manager) hasn't been fixed for so long

PR #1297, opened 2025-04-28
[rollout] misc: add demo chat completion scheduler described in ReTool paper

wuxibin89:


PeterSH6 (approved):
The multi-turn and tool-calling examples LGTM.
However, for the single-turn generation, I think we could simplify the workflow without using the ChatScheduler.
But we can merge this PR in advance.

I believe we should setup a meeting discuss a better "ChatScheduler" to generalize multi-turn/tool calling scenarios

PR #1316, opened 2025-04-29
[vllm] add moe patch for qwen3-moe

BearBiscuit05:
# What does this PR do?

Add moe patch for qwen3-moe. Fix the weight loader issue in vLLM MoE models. This isn’t a permanent solution, and we may need to contribute code to vLLM to address the problem caused by FusedMoE. I’m already seeking suggestions for this.

# ChangeLog:

- Add Qwen3MoeForCausalLM class for moe_patch

### DIFF
--- a/verl/utils/vllm_utils.py
+++ b/verl/utils/vllm_utils.py
@@ -32,8 +32,9 @@ def patch_vllm_moe_model_weight_loader(model):
     # (False, 'model.layers.0.mlp.experts.w2_weight')          use mlp.experts.weight_loader
     from vllm.model_executor.models.deepseek_v2 import DeepseekV2ForCausalLM, DeepseekV3ForCausalLM
     from vllm.model_executor.models.qwen2_moe import Qwen2MoeForCausalLM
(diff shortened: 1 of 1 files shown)


verl/utils/vllm_utils.py
```diff
@@ -32,8 +32,9 @@ def patch_vllm_moe_model_weight_loader(model):
     # (False, 'model.layers.0.mlp.experts.w2_weight')          use mlp.experts.weight_loader
     from vllm.model_executor.models.deepseek_v2 import DeepseekV2ForCausalLM, DeepseekV3ForCausalLM
     from vllm.model_executor.models.qwen2_moe import Qwen2MoeForCausalLM
+    from vllm.model_executor.models.qwen3_moe import Qwen3MoeForCausalLM
```
vermouth1992: Shall we make sure Qwen3MoeForCausalLM exists in vllm?

vermouth1992: People may use old version of vllm

BearBiscuit05: done

vermouth1992 (approved):

PR #1319, opened 2025-04-29
[bugfix]:CPU-only Workers Failed to Initiallize

softBread233:
Before,in the `test_worker_group_basics.py` file, when the `use_gpu` parameter is set to `False`, the program encounters an error. The error message `"RuntimeError: No CUDA GPUs are available"` is thrown, preventing the code from running properly in CPU-only environments. 
The root cause of this problem is that the original code attempts to access `torch.cuda.get_device_name()` without first checking if CUDA GPUs are available. The `torch.cuda.get_device_name()` function can only be called when `torch.cuda.is_available()` returns `True`.
The code has been modified to first check the availability of CUDA GPUs using `torch.cuda.is_available()` before attempting to access `torch.cuda.get_device_name()`.
This PR ensures that  the code can now run smoothly in both GPU-enabled and CPU-only environments, resolving the original error.

### DIFF
--- a/tests/ray/test_worker_group_CPUonly.py
+++ b/tests/ray/test_worker_group_CPUonly.py
@@ -1,0 +1,106 @@
+'''
+Author: [Mr.L]
+Date: 2025-05-03 15:48:12
+LastEditors: [Mr.l]
+LastEditTime: 2025-05-03 15:48:54
+Description: 
+'''
+# Copyright 2024 Bytedance Ltd. and/or its affiliates
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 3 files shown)


ZihengJiang (commented):
Thanks for the PR. Can you add a test to check the CPU-only worker initialization case?

Mr-chen233:
> Thanks for the PR. Can you add a test to check the CPU-only worker initialization case?

already add.

eric-haibin-lin:
Would u mind changing the test file path to tests/single_controller/base/test_worker.py thx

Mr-chen233:
> Would u mind changing the test file path to tests/single_controller/base/test_worker.py thx

already change the file path, and should I name the file test_worker_CPU_only.py? thx

eric-haibin-lin:
the CPU_only suffix is not needed. And it seems that your test is duplicated once in this PR..

PR #1320, opened 2025-04-29
cancel bootstrapping for n=n_samples

mertunsall:
# What does this PR do?

The validation metrics currently bootstraps its estimates by randomly sampling 1,2,4,8,16,...,n_samples results out of n_samples results. However, this bootstrapping doesn't make sense for `n=n_samples` as you cannot have more information about the estimate for `pass@n_samples` if you only have `n_samples` samples.

This results in weird results when doing RL with only one problem in the validation set (best@N is a value between 0 and 1 instead of 0 or 1)

This PR turns off bootstrapping for n=n_samples case and leaves rest of the computations the same.


mertunsall:
@hiyouga another very quick PR, if you could take a look! Thanks a lot!

hiyouga (approved):

PR #1331, opened 2025-04-30
fix: ray worker exit with SYSTEM_ERROR caused by SIGALRM from math re…

chenhaiq:
…ward

Since SIGALRM only works in main thread, if it is fired in a sub thread, the ray worker will exit with SYSTEM_ERROR.
Fixed this problem by using multiprocessing.Process instead of SIGALRM handling.

# What does this PR do?

bug fix for ray worke exit with SYSTEM_ERROR when timeout in prime math

# ChangeLog:

Fixed this problem by using multiprocessing.Process instead of SIGALRM handling.

# Usage

- see tests/utility/test_timeout_decorator.py


tongyx361 (dismissed):
LGTM! But could you please add a CI for the test?

tongyx361:
Btw, just curious about how you locate the root cause of `SYSTEM_ERROR`?

chenhaiq:
> Btw, just curious about how you locate the root cause of `SYSTEM_ERROR`?

1. strace -o trace.log -p 3024935 （pid of the runner process）
2. when the process is exited, check tailing lines of trace.log：
```
epoll_wait(35, 0x7ffe6fdd5d40, 128, -1) = -1 EINTR (Interrupted system call)
--- SIGALRM {si_signo=SIGALRM, si_code=SI_KERNEL} ---
+++ killed by SIGALRM +++
```

chenhaiq:
@tongyx361 although timeout is implemented by a different method, I find that the upper caller did not handle TimeoutError exception, so the training still will fail when are_equal_under_sympy timeout.
I will update this commit to return false when timeout in upper callers, please hold this fix.

tongyx361:
Is the original error only triggered if we use `threading`? But if I understand correctly, `threading` can not accelerate `sympy` workload due to the GIL. Why do you want to use it with `threading`? Or did I misunderstand?

tongyx361:
Could you:

1. add your test to the CI following other workflow `.yml`?
2. provide an end-to-end training test record using your new timeout decorator?

chenhaiq:
> Is the original error only triggered if we use `threading`? But if I understand correctly, `threading` can not accelerate `sympy` workload due to the GIL. Why do you want to use it with `threading`? Or did I misunderstand?

I just hit this error when run the sample ppo trainning after pull main code without any code change. 
I do not want to use it with threading either. There should be a change from recent commits that introduced threading.

chenhaiq:
> Could you:
> 
> 1. add your test to the CI following other workflow `.yml`?
> 2. provide an end-to-end training test record using your new timeout decorator?

1. the new pytest has already been included in .github/workflows/sanity.yml (pytest -s -x tests/utility)
2. this is a job which have been running for 12 steps with this fix:
截屏2025-05-06 20 51 21

tongyx361:
> > > Could you:
> > > 
> > > 1. add your test to the CI following other workflow `.yml`?
> > > 2. provide an end-to-end training test record using your new timeout decorator?
> > 
> > 
> > 
> > 1. the new pytest has already been included in .github/workflows/sanity.yml (pytest -s -x tests/utility)
> > 2. this is a job which have been running for 12 steps with this fix:
> > 
> > 截屏2025-05-06 20 51 21
> 
> For 1, there seems only `pytest -s -x tests/sanity`?
> 
> https://github.com/volcengine/verl/blob/d60499d170bcd3cc39cbc618475d9732947de142/.github/workflows/sanity.yml#L45

The unit tests workflow is recently moved to https://github.com/volcengine/verl/blob/d60499d170bcd3cc39cbc618475d9732947de142/.github/workflows/verl_unit_test.yml#L48-L51

Let's just wait for the CI passing and merge!

PR #1342, opened 2025-04-30
refactor: DataProto.response_mask & compute_response_mask

tongyx361:
# What does this PR do?

This PR refactor `compute_response_mask(data: DataProto)` to be also usable by `batch: TensorDict`.

# Change Log

- `DataProto.response_mask` is for `data: DataProto`.
- `compute_response_mask(response_ids: torch.Tensor, attention_mask: torch.Tensor)` is for `batch: TensorDict`.

# Before submitting

- [x] Did you read the [Contribute Guide](https://github.com/volcengine/verl?tab=readme-ov-file#contribution-guide) and finish the [code format check](https://github.com/volcengine/verl?tab=readme-ov-file#code-linting-and-formatting)? 
- [x] Did you make sure to update the documentations with your changes in the [docs](https://github.com/volcengine/verl/tree/main/docs) especially for breaking config etc?
- [x] Did you write any test cases if neccessary? Please add CI tests to your new feature.  

# Additional Info

- **Issue Number**: none
- **Training**: none
- **Inference**: none


vermouth1992:
In general, this is not a good idea because it makes the general data structure DataProto not general any more. DataProto is an extension of Tensordict to support non-tensor. We can actually replace DataProto with Tensordict now as Tensordict supports non-tensor.

tongyx361:
> In general, this is not a good idea because it makes the general data structure DataProto not general any more. DataProto is an extension of Tensordict to support non-tensor. We can actually replace DataProto with Tensordict now as Tensordict supports non-tensor.

Thanks for your reminder! I also considered the generality before, but I think it might be good to have **some utility for training-specific tensors to calculate some tensor like `response_mask` from existing ones `responses` and `attention_mask` to keep the principle of single source of truth, thus avoiding that any tensor is outdated**.

From this perspective, `TensorDict` is not enough. But combining this thought and your suggestion together, maybe it's best to inherit from `TensorDict` and add these training-specific utilities, while still naming it as `DataProto`?

For now, could we go on with leaving these utilities under `DataProto`?

PR #1347, opened 2025-05-01
[New Recipe] SPIN algorithm implementationSpin pr

cedricbeta:
This PR introduces an implementation of the Self-Play Fine-Tuning (SPIN) algorithm, adapting the existing PPO framework within verl.

You can find more information about SPIN here: https://github.com/uclaml/SPIN

This implementation adapts the PPO framework for SPIN/Online DPO, involving these core changes:

* **Objective & Loss:**
    * PPO maximizes cumulative reward via policy gradients and value estimates.
    * SPIN/Online DPO directly optimizes preference likelihood using a **DPO-specific loss function** (e.g., sigmoid loss).
    * **Code Change:** The primary logic change is in the actor's update step (`dp_actor.py: SPINDataParallelPPOActor.update_policy_dpo_with_ref`, `fsdp_workers.py: SPINRolloutRefWorker.update_actor_dpo`) and the loss calculation (`core_algos.py: compute_online_dpo_loss`).

* **Model Requirements:**
    * PPO uses Actor, Critic, and optionally Reward/Reference models.
    * SPIN/Online DPO uses Actor and a **mandatory Reference Model** for the loss calculation, plus a reward source for *preference labeling*. **No Critic is needed.**
    * **Code Change:** The `ray_trainer.py` logic was modified to manage/update the reference model and *not* initialize/use the critic worker. `fsdp_workers.py` was updated for reference model initialization and checkpointing.

* **Update Signal:**
    * PPO relies on **Advantage estimates** derived from rewards and the critic's value function.
    * SPIN/Online DPO uses the **log probability difference** between chosen/rejected pairs under the policy and reference models.
    * **Code Change:** Advantage calculation (`compute_advantage`) was removed from the training loop in `ray_trainer.py`. Preference determination (`compute_onlineDPO_pref` in `core_algos.py`) was added.

* **Data:**
    * PPO uses (prompt, response, reward, value) tuples.
    * SPIN/Online DPO effectively uses (prompt, chosen_response, rejected_response) tuples, requiring preference data generation.
    * **Code Change:** Data processing in `ray_trainer.py` (`fit_dpo`) was adapted to handle preference pairs and prepare the specific inputs needed for `update_policy_dpo`.


recipe/spin/config/spin_trainer.yaml
```diff
@@ -0,0 +1,221 @@
...
+        min_num_params: 0
+    log_prob_micro_batch_size: null # will be deprecated, use log_prob_micro_batch_size_per_gpu
+    log_prob_micro_batch_size_per_gpu: null
+    log_prob_use_dynamic_bsz: ${actor_rollout_ref.actor.use_dynamic_bsz}
+    log_prob_max_token_len_per_gpu: ${actor_rollout_ref.actor.ppo_max_token_len_per_gpu}
+    ulysses_sequence_parallel_size: ${actor_rollout_ref.actor.ulysses_sequence_parallel_size} # sp size
+  rollout:
+    name: sglang
+    temperature: 1.0
+    top_k: -1 # 0 for hf rollout, -1 for vllm rollout
```
hebiao064: does this mean we also need to set it to -1 for sglang? It would be nice to replace `vllm` with `sglang` in the comment to avoid confusion

recipe/spin/config/spin_trainer.yaml
```diff
@@ -0,0 +1,221 @@
...
+    enable_chunked_prefill: True # may get higher throughput when set to True. When activated, Please increase max_num_batched_tokens or decrease max_model_len.
+    # for hf rollout
+    do_sample: True
+    # number of responses (i.e. num sample times)
+    n: 2 # > 1 for grpo
+    engine_kwargs: # inference engine parameters
+      swap_space: null # null means "use the engine default value" (usually 4 GB), setting it to, e.g., 32 means 32 GB
+    val_kwargs:
+      # sampling parameters for validation
+      top_k: -1 # 0 for hf rollout, -1 for vllm rollout
```
hebiao064: same here

recipe/spin/config/spin_trainer.yaml
zhaochenyang20: I changed it

eric-haibin-lin:
could you review this RFC? Thanks https://github.com/volcengine/verl/issues/1354

cedricbeta:
> could you review this RFC? Thanks #1354
Thanks for the guideline! Just updated the PR to meet the structure requirement!

recipe/spin/config/spin_trainer.yaml
tongyx361, line 1: Could you please follow dapo_trainer.yaml or prime_trainer.yaml to use the `default` grammar to make this take ppo_trainer.yaml as default to reduce redundant code?

recipe/spin/config/spin_trainer.yaml
```diff
@@ -0,0 +1,221 @@
...
+      top_p: 1.0
+      temperature: 0
+      n: 2
+      do_sample: False # default eager for validation
+
+critic:
```
hebiao064: maybe we can copy your doc to here to avoid confusion?

```yaml
# The value function critic model used in PPO is not required and not used in SPIN
# we keep following critic configs just as placeholder to keep consistent with config format
```

recipe/spin/dp_actor.py
zhaochenyang20, line 1: Could  this be clean up?

zhaochenyang20 (changes requested):

zhaochenyang20:
Great Job. Let me clean up all the following part.

PR #1349, opened 2025-05-01
Explicitly initialize optimizer states to prevent OOM during rollout

llkn-2:
# What does this PR do?

Optimizer states are lazily initialized on the first step, which can lead to an over-allocation of the KV cache when `optimizer_offload=False`, potentially causing an out-of-memory (OOM) error during the rollout phase.

# ChangeLog:

- Manually initialize optimizer state before building the rollout.

## Before submitting

- [x] Did you read the [Contribute Guide](https://github.com/volcengine/verl?tab=readme-ov-file#contribution-guide) and finish the [code format check](https://github.com/volcengine/verl?tab=readme-ov-file#code-linting-and-formatting)? 
- [x] Did you make sure to update the documentations with your changes in the [docs](https://github.com/volcengine/verl/tree/main/docs) especially for breaking config etc?
- [x] Did you write any test cases if neccessary? Please add CI tests to your new feature.  

# Additional Info: 
- **Issue Number**: none
- **Training**: FSDP
- **Inference**: none

### DIFF
--- a/verl/utils/fsdp_utils.py
+++ b/verl/utils/fsdp_utils.py
@@ -168,6 +168,37 @@ def load_fsdp_optimizer(optimizer, device_id):
                     state[key] = value.to(device_id, non_blocking=True)
 
 
+@torch.no_grad()
+def init_optimizer_state(optimizer):
+    # Adapt from https://github.com/pytorch/pytorch/blob/v2.6.0/torch/optim/adam.py#L157
+    from torch.optim.optimizer import _get_scalar_dtype
+
+    for group in optimizer.param_groups:
+        for p in group["params"]:
+            state = optimizer.state[p]
+            # State initialization
+            if len(state) == 0:
+                # note(crcrpar): [special device hosting for step]
+                # Deliberately host `step` on CPU if both capturable and fused are off.
+                # This is because kernel launches are costly on CUDA and XLA.
+                state["step"] = (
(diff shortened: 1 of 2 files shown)


verl/utils/fsdp_utils.py
```diff
@@ -168,6 +168,37 @@ def load_fsdp_optimizer(optimizer, device_id):
                     state[key] = value.to(device_id, non_blocking=True)
 
 
+@torch.no_grad()
+def init_optimizer_state(optimizer):
```
vermouth1992, line 172: Is this function only applicable for Adam?

llkn-2: Yes, currently `fsdp_workers.py` only supports AdamW，maybe I could add an assertion？

llkn-2: Another solution maybe use `optimizer.step()` after initialization, like this:
```
for group in optimizer.param_groups:
    for p in group['params']:
        p.grad = torch.zeros_like(p.data)
optimizer.step()
optimizer.zero_grad()
```

vermouth1992: Agree. Maybe this solution is better.

llkn-2: @vermouth1992 I’ve already made the updates, are there any other issues blocking the merge?

PR #1356, opened 2025-05-02
[feat] Add auto padding for DataProto

hongpeng-guo:
### Checklist Before Starting

Coming from #577 , credit to @zw0610 

### What does this PR do?

Today, users must manually duplicate (repeat) a DataProto so its batch size matches the data‑parallel (dp) size of the target WorkerGroup. This PR enables `auto_padding` to pad the `DataProto` when chunk is called.

### Specific Changes

* Enriched the `DataProto` so that it can have context of padding during chunking;
* Modified the `decorator.py` that a DataProto can be automatically padded and chunked with  `dispatch_dp_compute_data_proto`;
* Added unit tests under `tests/ray/test_auto_padding.py`.

### API

Two new API under `DataProto` are introduced, which are `padding` and `is_padding_enabled`

### Test

Tests added to `tests/ray/test_auto_padding.py`

### Additional Info.


verl/single_controller/base/decorator.py
```diff
@@ -46,17 +46,40 @@ class Execute(Enum):
     RANK_ZERO = 1
 
 
-def _split_args_kwargs_data_proto(chunks, *args, **kwargs):
+def _split_args_kwargs_data_proto(chunks, *args, enable_auto_padding=False, **kwargs):
```
hongpeng-guo: for this function, I still think a readable arg name would be better. See https://github.com/volcengine/verl/pull/557#discussion_r2071020679 for more related discussion. what do you think? @vermouth1992

hongpeng-guo: never mind. I simply split the two scenarios into two helper functions, and there is no need of the addition args.

vermouth1992:
In fact, we can actually deprecate DataProto since Tensordict 0.8.0 is out. We can merge this first and discuss this later.

PR #1358, opened 2025-05-02
[BREAKING] feat: decouple PPO epochs from critic

emergenz:
### Checklist Before Starting

### What does this PR do?

Currently, `ppo_epochs` is used for both the actor, as well as the critic. While this is the case for many PPO implementations (https://github.com/EdanToledo/Stoix, https://github.com/openai/baselines), there is no inherent reason why this should be the case. This is probably due to historic reasons, since OpenAI's PPO implementation had a single loss for both actor and critic (https://github.com/openai/baselines/blob/ea25b9e8b234e6ee1bca43083f8f3cf974143998/baselines/ppo2/model.py; inherent coupling of `num_epochs` between actor and critic).

This PR simply decouples the actor and critic epochs.

### API

The only API change is that I renamed `critic.ppo_epochs` to `critic.epochs` to reflect the new behaviour.

### Additional Info.

- **Training**: [FSDP, Megatron]
- **Inference**: []


eric-haibin-lin (commented):
the existing config still allows custom epoch num for critic via setting critic.ppo_epochs. does it not work for you?

Issue #1354, opened 2025-05-02
[RFC] verl recipe and extensions


eric-haibin-lin:
# motivation

RL training can be quite complicated. The goal of verl is to provide a RL library for LLMs that is easy to hack for researchers for exploration, meanwhile providing the right level of abstraction so that it can scale for production usage. verl needs a place to put end2end RL training pipelines that are promising for the community to reproduce and further explore, while maintaining a relatively stable verl core. 

There are ongoing PRs for adding more functionality and training pipeline to verl. We need consensus on how they should be maintained, and this RFC discuss the recipes verl currently maintains. 

# proposed change

verl shall maintain a few folders:
- `/verl` : contains the most common building blocks / utils for RL training
- `/examples`: scripts showing how to use verl for common use cases
- **`/recipe`**: end-to-end RL recipes from data preparation, reward definition, env/tool and the training algorithm, usually extending verl's APIs. Once a certain recipe is mature, we move shareable components back to `/verl`. 

To make verl easier to extend, we need to define the interfaces. Let's break it down:

## data format
The RLDataset currently provides data in the format described in https://verl.readthedocs.io/en/latest/preparation/prepare_data.html#prepare-data-for-post-training. For other tasks such as browser use / code / chat preferences, verl should **document the data format** and provide example Dataset implementation. 

TODO:
- review the current data format, and provide more documentation for the vision format
- docs for SFT single/multi-turn dataset format

## trainer
The RayPPOTrainer provides an example training loop for a typical on-policy RL algorithm with a reward model/function. Since https://github.com/volcengine/verl/pull/1282 **dataset construction is done outside of the trainer**, so it is more flexible in dataset customization. For any new algorithm that requires a different training loop (e.g. async RL, DPO), or requires considerable changes to the existing trainer, we highly recommend try **inheriting from the existing trainer** first, or simply **making a copy of the trainer** for your own exploration. 

The RayPPOTrainer is expected to be maintained as an API with backward compatibility. For plugins or integrations that should be added to RayPPOTrainer interface to improve its functionality/performance, please leave a comment in this thread for discussions. 

## entry script
You should always make a copy of the entry script such as `main_ppo.py` should you need to make change. 

## recipe code structure

To add a new recipe, a few things are expected from recipe maintainers:

- a `README.md` file that describes what this recipe is about plus any background/paper for educational purpose 
- the verl version needed to run the recipe (either a stable version or requires verl commit xxx), requirements.txt for dependencies or recommended image. 
[...]

PR #1407, opened 2025-05-05
[New Recipe] SPIN algorithm implementation

cedricbeta:
This PR introduces an implementation of the Self-Play Fine-Tuning (SPIN) algorithm, adapting the existing PPO framework within verl.

You can find more information about SPIN here: https://github.com/uclaml/SPIN

This implementation adapts the PPO framework for SPIN/Online DPO, involving these core changes:

* **Objective & Loss:**
    * PPO maximizes cumulative reward via policy gradients and value estimates.
    * SPIN/Online DPO directly optimizes preference likelihood using a **DPO-specific loss function** (e.g., sigmoid loss).
    * **Code Change:** The primary logic change is in the actor's update step (`dp_actor.py: SPINDataParallelPPOActor.update_policy_dpo_with_ref`, `fsdp_workers.py: SPINRolloutRefWorker.update_actor_dpo`) and the loss calculation (`core_algos.py: compute_online_dpo_loss`).

* **Model Requirements:**
    * PPO uses Actor, Critic, and optionally Reward/Reference models.
    * SPIN/Online DPO uses Actor and a **mandatory Reference Model** for the loss calculation, plus a reward source for *preference labeling*. **No Critic is needed.**
    * **Code Change:** The `ray_trainer.py` logic was modified to manage/update the reference model and *not* initialize/use the critic worker. `fsdp_workers.py` was updated for reference model initialization and checkpointing.

* **Update Signal:**
    * PPO relies on **Advantage estimates** derived from rewards and the critic's value function.
    * SPIN/Online DPO uses the **log probability difference** between chosen/rejected pairs under the policy and reference models.
    * **Code Change:** Advantage calculation (`compute_advantage`) was removed from the training loop in `ray_trainer.py`. Preference determination (`compute_onlineDPO_pref` in `core_algos.py`) was added.

* **Data:**
    * PPO uses (prompt, response, reward, value) tuples.
    * SPIN/Online DPO effectively uses (prompt, chosen_response, rejected_response) tuples, requiring preference data generation.
    * **Code Change:** Data processing in `ray_trainer.py` (`fit_dpo`) was adapted to handle preference pairs and prepare the specific inputs needed for `update_policy_dpo`.


cedricbeta:
@PeterSH6 Hi Guangming, we revised the code according to the new recipe RFC! Hope this works!

recipe/spin/fsdp_workers.py
```diff
@@ -0,0 +1,1183 @@
...
+        if self._is_offload_param:
+            offload_fsdp_model_to_cpu(self.actor_module_fsdp)
+        if self._is_offload_optimizer:
+            offload_fsdp_optimizer(optimizer=self.actor_optimizer)
+
+        return output
+    
+
+# class ActorRolloutRefWorker(Worker):
+#     """
```
hebiao064: remove this?

recipe/spin/run_spin.sh
```diff
@@ -0,0 +1,30 @@
...
+  data.val_files=$HOME/data/gsm8k/test.parquet \
+  data.train_batch_size=1024 \
+  data.max_prompt_length=1024 \
+  data.max_response_length=1024 \
+  actor_rollout_ref.model.path=Qwen/Qwen2.5-0.5B-Instruct \
+  actor_rollout_ref.actor.optim.lr=1e-6 \
+  actor_rollout_ref.actor.ppo_mini_batch_size=64 \
+  actor_rollout_ref.actor.ppo_micro_batch_size=8 \
+  actor_rollout_ref.rollout.log_prob_micro_batch_size=64 \
+  actor_rollout_ref.rollout.tensor_model_parallel_size=1 \
```
hebiao064, line 17: just for my curiosity, why we set tp to 1 during rollout?

recipe/spin/README.md
```diff
@@ -0,0 +1,150 @@
...
+3.  **Update:** This preference tuple (`prompt`, `chosen_response`, `rejected_response`) is used to update the actor model using the DPO loss function, comparing against a reference model.
+
+**Connection to SPIN:**
+While this recipe uses the DPO loss, the online generation loop where the current model generates data used for its own update shares conceptual similarities with the self-play idea in SPIN. The periodic update of the reference model (potentially using weights from the actor) further aligns with SPIN's iterative self-improvement concepts.
+
+**Reference Papers:**
+* **SPIN:** [Self-Play Fine-Tuning Converts Weak Language Models to Strong Language Models](https://arxiv.org/abs/2401.01335) (Chen et al., 2024)
+* **DPO:** [Direct Preference Optimization: Your Language Model is Secretly a Reward Model](https://arxiv.org/abs/2305.18290) (Rafailov et al., 2023)
+
+## Implementation within Verl
```
ZihengJiang: According to the RFC(https://github.com/volcengine/verl/issues/1354), let's add a note about the verl version needed to run the recipe (either a stable version or requires verl commit xxx)

eric-haibin-lin (dismissed):
```
2025-05-21T16:57:18Z	info-0	trufflehog	scanning repo	{"source_manager_worker_id": "14K0i", "unit_kind": "dir", "unit": "/tmp/", "repo": "https://github.com/volcengine/verl", "base": "0d3360a12d2311435ba869e90f0fb50eafa61f47", "head": "a491803cf2e9bb7d47bf4a503d5b96a806426dc3"}
Warning: Found verified WeightsAndBiases result 🐷🔑
```
please either revoke this wandb key or squash history commits into one so the key info is not revealed

eric-haibin-lin (approved):
in general i think there's lots of potential for verl API improvement (on the verl side). for now i support merging this PR first. we can refactor some of the APIs from verl later

zhaochenyang20:
@eric-haibin-lin Thanks. I will take a second look and PR some refactoring if needed.

cedricbeta:
> Thanks for your great work on this project! I’ve been exploring the code and I have a question regarding the data generation process in self.actor_rollout_wg.generate_sequences(gen_batch).
> 
> From what I can see, this function seems to generate only one output per prompt. However, for preference-based methods like DPO (Direct Preference Optimization), we typically need pairs of outputs (e.g., a “better” and “worse” answer) to form preference pairs for training.
> 
> Could you clarify:
> 
> Why does generate_sequences generate only one answer per prompt? How do you construct the preference pairs from these single outputs? Are you comparing across batches, sampling additional generations, or relying on some scoring mechanism? If I wanted to modify the code to generate multiple answers per prompt (for explicit pairwise comparison), where would be the best place to adjust? Thanks a lot for your help! Looking forward to your guidance.

Hi @TianL123, thanks for the support! 

- We actually generate a response with the base model and generate another one with the reference model, then compare their response with the reward function. The reference model is initialized as the base model and can be updated with user-specific update ratio.
-  After the two response get generated and stored in the same batch, we call the compute_onlineDPO_pref to compare the two reponses.
- If you prefer not to use the ref model you can directly use the rollout.n to configure the rollout number or call the generate function multiple times and construct the batch by your self.

Hope these helps!

PR #1408, opened 2025-05-06
WIP [worker] Adding `destroy` API to `WorkerGroup` and `RayWorkerGroup`

hongpeng-guo:
### Checklist Before Starting

- [x] Search for similar PR(s). credit to LX

### What does this PR do?

Introduce one new abstract API to worker and worker group. 

* `WorkerGroup::destroy()` which will destroy the worker group and release related resource; 

### API

Adding a new API `WorkerGroup::destroy()`

### Usage Example

```python
# destroy the worker group and cleanup related resources.
wg.destroy()
```

### DIFF
--- a/verl/single_controller/base/worker.py
+++ b/verl/single_controller/base/worker.py
@@ -229,3 +229,16 @@ class Worker(WorkerHelper):
     def execute_func_rank_zero(self, func, *args, **kwargs):
         result = func(*args, **kwargs)
         return result
+
+    @register(dispatch_mode=Dispatch.ONE_TO_ALL)
+    def ready(self):
+        # Whether the worker is ready to be used for computing.
+        # Subclasses can override this method to implement custom readiness checks,
+        # e.g., for additional model initialization
+        return True
+
+    @register(dispatch_mode=Dispatch.ONE_TO_ALL)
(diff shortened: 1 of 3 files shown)


verl/single_controller/ray/base.py
```diff
@@ -429,6 +429,15 @@ def workers(self):
     def world_size(self):
         return self._world_size
 
+    def destroy(self):
```
eric-haibin-lin, line 432: could u add a unit test that calls this function?

verl/single_controller/base/worker.py
```diff
@@ -29,15 +29,15 @@ class DistRankInfo:
     tp_rank: int
     dp_rank: int
     pp_rank: int
-    cp_rank: int
+    cp_rank: int = 0
```
ZihengJiang, line 32: why this change?

hongpeng-guo: I was thinking about adding a default value for cp, as cp might not be as widely used as the first three parallelism. Let me revert the change.

hongpeng-guo:
Close this PR for now, will reopen this PR as the related feature is more stable.

Issue #1313, opened 2025-04-29
Support Qwen3？


XueruiSu:
What I should do to make verl support Qwen3? I found the following error:
![Image](https://github.com/user-attachments/assets/f5155f78-25a3-4846-9562-4b54e96c97ac)

Do I just need to find the corresponding function or do I need to make a lot of engineering changes?
![Image](https://github.com/user-attachments/assets/fa78987f-2bb4-4fb9-8a4b-2075149420fd)

XueruiSu:
After upgrading vllm to a higher version, can the lower version files in the https://github.com/volcengine/verl/tree/main/verl/third_party/vllm directory still be used?

mihirp1998:
seems like by default enable_thinking is turned on? is there a simple way to turn it off?

XueruiSu:
There is no separate treatment for Qwen3 here. Does this potential problem (maybe) have a significant impact on the experimental results?
![Image](https://github.com/user-attachments/assets/5ed0b84e-10bc-4f66-b86a-3b72f3df26c7)

BearBiscuit05:
> There is no separate treatment for Qwen3 here. Does this potential problem (maybe) have a significant impact on the experimental results? ![Image](https://github.com/user-attachments/assets/5ed0b84e-10bc-4f66-b86a-3b72f3df26c7)

It just make the mfu always 0. And will not impact the experimental results.

vermouth1992:
Shall we add flops counter for QWen3 model?

BearBiscuit05:
> Shall we add flops counter for QWen3 model?

I think we need to support it. I’ll work on it.

XueruiSu:
Very glad to your support. This will be a great contribution for the whole community

changyi7231:
> Shall we add flops counter for QWen3 model?

Training with reinforcement learning, Qwen3 is much slower than Qwen2.5. Is this normal?"

dirtyDan0:
> seems like by default enable_thinking is turned on? is there a simple way to turn it off?

![Image](https://github.com/user-attachments/assets/e0c7baa9-2209-467c-a124-82b9979baf05)

JoyTim-777:
Does Verl support grpo training for the qwen3-235B-A22B?

hebiao064:
> Does Verl support grpo training for the qwen3-235B-A22B?

+1

Kirrito-k423:
[...]

PR #1409, opened 2025-05-06
[misc] add support for qwen3 model (dense/moe)

BearBiscuit05:
### Checklist Before Starting

### What does this PR do?

>  add mfu compute function for qwen3 model

### Additional Info.

- **Issue Number**: Fixes issue #1313 


verl/utils/flops_counter.py
```diff
@@ -156,6 +158,43 @@ def _estimate_deepseek_v3_flops(self, tokens_sum, batch_seqlens, delta_time):
 
         return flops_achieved
 
+    def _estimate_qwen3_moe_flops(self, tokens_sum, batch_seqlens, delta_time):
```
vermouth1992, line 161: Shall we add a test for mfu calculation of all the models?

BearBiscuit05: I’ve thought about this too, but I’m unsure. For a test, what’s the expected result? For example, if we provide a model and sequence length, do we expect a specific MFU value?

BearBiscuit05: done

vermouth1992 (approved):

PR #1419, opened 2025-05-06
[Metric] fix: boostrap with `n == n_resps` since with replacement

tongyx361:
### Checklist Before Starting

### What does this PR do?

This PR fixes https://github.com/volcengine/verl/pull/1320 since bootstrapping is done with replacement, which makes it still meaningful even when `n == n_resps`

### Additional Info.

- **Issue Number**: https://github.com/volcengine/verl/pull/1320
- **Training**: none
- **Inference**: none

### DIFF
--- a/verl/trainer/ppo/metric_utils.py
+++ b/verl/trainer/ppo/metric_utils.py
@@ -214,20 +214,14 @@ def process_validation_metrics(data_sources: list[str], sample_inputs: list[str]
                 metric[f"mean@{n_resps}"] = np.mean(var_vals)
 
                 if n_resps > 1:
-                    # n = n_resps
                     metric[f"std@{n_resps}"] = np.std(var_vals)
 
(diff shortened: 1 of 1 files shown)


tongyx361:
> The problem I had was that when there is one problem in your test-set (for example, when applying test-time RL for a single problem) the pass@k accuracy is a value between 0 and 1, instead of 0 or 1. I am not sure about the statistics theory behind this but information-wise I felt you shouldn't be able to gain more information by just sampling with replacement. maybe I am wrong.

I think a pass@k accuracy between 0 and 1 for a one-problem test set is still reasonable. The key is that pass@k is still a random variable and we want to estimate its expectation. When you sample k solutions to the problem, sometimes you have at least one correct solution, while sometimes you have none, then you can't say that the expectation of pass@k is 1. What do you think of this view? @mertunsall

vermouth1992 (approved):

PR #1424, opened 2025-05-07
[single_controller][decorator] Define a `DynamicEnum` class to make `Dispatch` and `Execute` extensible.

hongpeng-guo:
### Checklist Before Starting

### What does this PR do?

Today, extending `verl` in proprietary usage large requires forking it, and padding code changes in the private fork. 
For example, the current `verl` API doesn't support adding new `Dispatch` and `Execute` mode in runtime. The only way to achieve it is to make a new private fork.

This PR replace the static `Enum` type of `Dispatch` and `Execute` into a new `"DynamicEnum"` type, that the users can use new APIs `register_dispatch_mode` and `update_dispatch_mode` to adding and define new distributed mode at runtime using native `verl` API, instead of making a fork.

### Specific Changes

* Defined `DynamicEnum` class in `utils.py_functional.py`;
* Re-defined `Dispatch` and `Execute` classes, all existing Enum API and usage are still usbale;
* Added `register_dispatch_mode` and `update_dispatch_mode` for users to register new dispatch modes at runtime;
* nit: `pre-commit` automatically fixed part of code format in another PR #1331 

### Usage Example

```python
def test_register_new_dispatch_mode():
    # Test registration
    def dummy_dispatch(worker_group, *args, **kwargs):
        return args, kwargs

    def dummy_collect(worker_group, output):
        return output

    register_dispatch_mode("TEST_MODE", dummy_dispatch, dummy_collect)

    # Verify enum extension
    _check_dispatch_mode(Dispatch.TEST_MODE)

    # Verify registry update
    assert get_predefined_dispatch_fn(Dispatch.TEST_MODE) == {"dispatch_fn": dummy_dispatch, "collect_fn": dummy_collect}
```

### Test

Added `tests/verl/test_decorator.py`

### DIFF
--- a/tests/verl/test_decorator.py
+++ b/tests/verl/test_decorator.py
@@ -15,7 +15,7 @@
 import pytest
 
 import verl.single_controller.base.decorator as decorator_module
-from verl.single_controller.base.decorator import DISPATCH_MODE_FN_REGISTRY, Dispatch, _check_dispatch_mode, register_dispatch_mode, update_dispatch_mode
+from verl.single_controller.base.decorator import DISPATCH_MODE_FN_REGISTRY, Dispatch, _check_dispatch_mode, get_predefined_dispatch_fn, register_dispatch_mode, update_dispatch_mode
 
 
 @pytest.fixture
@@ -42,7 +42,7 @@ def test_register_new_dispatch_mode(reset_dispatch_registry):
     _check_dispatch_mode(Dispatch.TEST_MODE)
 
     # Verify registry update
-    assert DISPATCH_MODE_FN_REGISTRY[Dispatch.TEST_MODE] == {"dispatch_fn": dummy_dispatch, "collect_fn": dummy_collect}
+    assert get_predefined_dispatch_fn(Dispatch.TEST_MODE) == {"dispatch_fn": dummy_dispatch, "collect_fn": dummy_collect}
     # Clean up
     Dispatch.remove("TEST_MODE")
 
@@ -62,5 +62,5 @@ def test_update_existing_dispatch_mode(reset_dispatch_registry):
     update_dispatch_mode(original_mode, new_dispatch, new_collect)
 
     # Verify update
-    assert DISPATCH_MODE_FN_REGISTRY[original_mode]["dispatch_fn"] == new_dispatch
-    assert DISPATCH_MODE_FN_REGISTRY[original_mode]["collect_fn"] == new_collect
+    assert get_predefined_dispatch_fn(original_mode)["dispatch_fn"] == new_dispatch
+    assert get_predefined_dispatch_fn(original_mode)["collect_fn"] == new_collect
--- a/tests/verl/test_decorator.py
+++ b/tests/verl/test_decorator.py
@@ -1,3 +1,17 @@
+# Copyright 2024 Bytedance Ltd. and/or its affiliates
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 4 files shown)


tongyx361 (dismissed):

tests/verl/test_decorator.py
```diff
@@ -0,0 +1,66 @@
...
+        return args, kwargs
+
+    def new_collect(worker_group, output):
+        return output
+
+    # Test update=
+    update_dispatch_mode(original_mode, new_dispatch, new_collect)
+
+    # Verify update
+    assert DISPATCH_MODE_FN_REGISTRY[original_mode]["dispatch_fn"] == new_dispatch
```
ZihengJiang: should we provide a `get_dispatch_mode` function instead of manipulate the DISPATCH_MODE_FN_REGISTRY directly?

tests/verl/test_decorator.py
```diff
@@ -0,0 +1,66 @@
...
+    def dummy_dispatch(worker_group, *args, **kwargs):
+        return args, kwargs
+
+    def dummy_collect(worker_group, output):
+        return output
+
+    register_dispatch_mode("TEST_MODE", dummy_dispatch, dummy_collect)
+
+    # Verify enum extension
+    _check_dispatch_mode(Dispatch.TEST_MODE)
```
ZihengJiang, line 42: Registering with string then retrieving with a new enum type looks a bit of odd... Is this for being compatible with the previous Dispatch.XXX usage?

tongyx361: For registration, previous usage doesn't allow dynamic registration at all, i.e., all the items are hard-coded as members of the`Enum`. https://github.com/volcengine/verl/blob/312a8cbcebf964a7b2d55713c44b122f9182a88b/verl/single_controller/base/decorator.py#L27-L42

For accessing, previous usage just uses the attribute grammar, e.g., https://github.com/volcengine/verl/blob/312a8cbcebf964a7b2d55713c44b122f9182a88b/verl/single_controller/base/decorator.py#L372-L414

Therefore, to enable dynamic registration and accessing with the attribute, registering with the string of attribute name seems the only approach?

tests/verl/test_decorator.py
tongyx361: We have similar functions in the main code: https://github.com/volcengine/verl/blob/312a8cbcebf964a7b2d55713c44b122f9182a88b/verl/single_controller/base/decorator.py#L372-L414

I am not sure if such functions are necessary since they just perform indexing.

But it would be also nice to replicate the existing behavior in tests.

ZihengJiang: Depending on whether you want to expose `DISPATCH_MODE_FN_REGISTRY` this implementation detail to the user or provide an api so that you can change the internal implementation in the future without breaking those code. Generally speaking, the latter is better, but if you're currently only using it in tests, that's fine too.

For the code you point, you can also use the API you provide `register_dispatch_mode` to register them in the initial function.

tests/verl/test_decorator.py
hongpeng-guo: @ZihengJiang Yes, this is for the compatibility with the previous `Enum` usage. When registering, we use a name string, which will later become a member of `Dispatch` that can be used similar to `Enum` with attribute

tests/verl/test_decorator.py
hongpeng-guo: yes, we already have this function defined before this PR, which is `get_predefined_dispatch_fn`
Now, it looks like this: https://github.com/volcengine/verl/blob/880077c83265d1977225f7a2b941d4c2962a5c96/verl/single_controller/base/decorator.py#L430-L431

hongpeng-guo: Just updated the unit test using public API `get_predefined_dispatch_fn`

ZihengJiang: thanks, sounds good to me

tongyx361 (approved):

PR #1429, opened 2025-05-07
[feat] Sandbox: support sandbox fusion on FaaS & localhost

chenhaiq:
### Checklist Before Starting

- [ *] Search for similar PR(s).

### What does this PR do?

Implement sandbox fusion backend on FaaS. For example, reward score using a FaaS instance on volcengine.com. It have better performance and security comparing to local sandbox.

### Specific Changes

Added a code branch in _default_compute_score to choose sandbox according to sandbox_fusion_url configuration. 

### Usage Example

examples/ppo_trainer/run_deepseek7b_llm_sandbox_fusion.sh

### Test

tests/reward_score/test_sandbox_fusion.py
However, the new testcase requires to setting Sandbox API URL in env SANDBOX_FUSION_URL. If the env is not set, most testcases will be skipped.

### Additional Info.

Using sandbox on Faas have save 60% time on reward process comparing local sandbox:
截屏2025-05-07 20 37 05

- [*] Read the [Contribute Guide](https://github.com/volcengine/verl?tab=readme-ov-file#contribution-guide).
- [*] Apply [pre-commit checks](https://github.com/volcengine/verl?tab=readme-ov-file#code-linting-and-formatting).
- [*] Add `[BREAKING]` to the PR title if it breaks any API.
- [*] Update the documentation about your changes in the [docs](https://github.com/volcengine/verl/tree/main/docs).
- [*] Add CI test(s) if neccessary.


verl/utils/reward_score/sandbox_fusion/utils.py
```diff
@@ -0,0 +1,378 @@
...
+         logger.warning(f"Mismatch between number of inputs ({len(inputs)}) and outputs ({len(expected_outputs)}).")
+         # Return error based on the number of inputs provided
+         return [-1] * num_cases, [{"error": "Input/output count mismatch", "case_index": i} for i in range(num_cases)]
+
+
+    first_compile_error_index = -1
+
+    # Use ThreadPoolExecutor for concurrent execution
+    # Adjust max_workers based on expected load and API limits
+    with concurrent.futures.ThreadPoolExecutor() as executor:
```
wuxibin89: For multi turn rollout, each chat completion may independently call check_correctness, and we may have thousands of chat completion running concurrently. So we better have a global ThreadPoolExecutor, or more ideally use asyncio to avoid invoke hundreds of thread.

A `SandboxFusionManager` may a better option for  thread pool/event loop, rate limit control, etc.
```python
class SandboxFusionManager:
    def submit(self, in_outs: Optional[dict], generation: str, ...):
        pass
```

chenhaiq: 1. implemented a global concurrent limit to fusion api calls by using a thread semaphore
2. since prime.py calls compute_score method in subprocess concurrently, it can not work with asyncio in sandbox fusion. I will submit another version of sandbox for asyncio soon with SandboxFusionManager.

verl/trainer/config/ppo_megatron_trainer.yaml
```diff
@@ -229,7 +229,8 @@ reward_model:
   use_dynamic_bsz: ${critic.use_dynamic_bsz}
   max_length: null
   launch_reward_fn_async: False # custom reward function executed async on CPU, during log_prob
-
```
eric-haibin-lin, line 232: let's make it hierarchical:
```
sandbox_fusion:
   url:
   xxxx: 
```

chenhaiq: Pushed a new commit to fix this. Please review again.

tests/reward_score/test_sandbox_fusion.py
```diff
@@ -0,0 +1,540 @@
...
+    # For timeout, stdout might be None or empty depending on when the timeout occurred
+    # assert metadata_list[first_timeout_index]["stdout"] is None or metadata_list[first_timeout_index]["stdout"] == ""
+
+
+# --- Unit test cases (using mock) ---
+
+
+@patch("verl.utils.reward_score.sandbox_fusion.utils.call_sandbox_api")
+def test_unit_concurrency_order(mock_call_sandbox_api):
+    """单元测试：验证并发执行时结果顺序是否正确"""
```
eric-haibin-lin: please leave docs in English, thank you. Same for L238, L244, L255, L257, L265, and a few more occurrences

verl/utils/reward_score/sandbox_fusion/__init__.py
```diff
@@ -0,0 +1,95 @@
+import json
+import logging
+import traceback
+
+from .utils import check_correctness
+
+"""
+Verify code correctness using the sandbox_fusion FaaS service provided by public clouds. Public clouds covered by testing include:
+- volcengine.com
```
eric-haibin-lin: is it possible to deploy sandbox fusion locally, and does this PR support that kind of usage as well? How do users develop and debug sandbox before onboarding with volc-engine?

tests/reward_score/test_sandbox_fusion.py
chenhaiq: > please leave docs in English, thank you. Same for L238, L244, L255, L257, L265, and a few more occurrences

fixed

verl/utils/reward_score/sandbox_fusion/__init__.py
chenhaiq: updated this comment

chenhaiq:
@eric-haibin-lin ：
1. added a document about how to use sandbox fusion
2. fixed a performance issue because reasoning LLM may generate very long context without code.

eric-haibin-lin (approved):

verl/utils/reward_score/prime_code/__init__.py
```diff
@@ -50,7 +50,7 @@ def compute_score(completion, test_cases, continuous=False):
             metadata_list = []
             res_list = []
             for test_case_id, test_case in enumerate(test_cases_list):
-                res, metadata = apps_check_correctness(in_outs=test_case, generation=solution, timeout=5, debug=False)
+                res, metadata = apps_check_correctness(in_outs=test_case, generation=solution, timeout=10, debug=False)
```
eric-haibin-lin, line 53: why changing the timeout? What is the unit used here (second, minute?)

chenhaiq: It is because the local sandbox changed timeout from 10 to 5. Using the same default value to have the same effect.

chenhaiq:
@eric-haibin-lin the test failed because HTTPSConnectionPool(host='hf-mirror.com', port=443): Read timed out. Can we rerun it?

PR #1434, opened 2025-05-08
Update sglang multi-turn readme to use the correct preprocessing script

BaiqingL:
As the title states, the readme instructs users to use the normal gsm8k preprocessing function, but we should be using the formatted version instead.


vermouth1992 (approved):

BaiqingL:
No need, readme is already fixed upstream!

PR #1439, opened 2025-05-08
add pip dependances

chenhaiq:
Add 2 pip dependances to install correctly.


eric-haibin-lin:
what depends on them?

chenhaiq:
> what depends on them?

verl/workers/rollout/async_server.py

PR #1442, opened 2025-05-08
add `engine_kwargs`  in vllm_rollout_spmd to set params in vllm engine

zhangyongxin121:
### Checklist Before Starting

### What does this PR do?

> add `engine_kwargs`  in vllm_rollout_spmd to set swap_space in vllm engine. 

### Specific Changes

> add `engine_kwargs` in vllm_rollout_spmd, which can be set in config file. Same changes has been made in vllm_rollout. As the version of vllm is update to 0.8, the default vllm_rollout worker becomes vllm_rollout_spmd, which does not have `engine_kwargs` as in vllm_rollout, so this RP want to complete it.

### Usage Example

> users can set vllm engine param such as `swap_space`, `seed` through the `engine_kwargs` in config file. For example, if  one want to set the swap_space=32 in vllm, he can set the item in config like this
```bash
actor_rollout_ref.rollout.engine_kwargs.swap_space=32
```

### DIFF
--- a/verl/workers/rollout/vllm_rollout/vllm_rollout_spmd.py
+++ b/verl/workers/rollout/vllm_rollout/vllm_rollout_spmd.py
@@ -29,12 +29,13 @@ When working with Megatron:
 import logging
 import os
 from contextlib import contextmanager
+from copy import deepcopy
 from typing import Any, Dict, List, Union
 
 import numpy as np
 import torch
 import torch.distributed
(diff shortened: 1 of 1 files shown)


eric-haibin-lin (approved):

ZihengJiang:
@eric-haibin-lin @vermouth1992 shall we merge?

PR #1444, opened 2025-05-08
[megatron] Qwen3moe-part 2: Allow Infer and train tp to be different with CI tests, Fix vllm resharding process

ETOgaosion:
### Checklist Before Starting

### What does this PR do?

1. This PR eliminates the micro-dp group as the article says, and support train-infer tp to be different.
2. Side Effect: able to run Qwen3moe on megatron aligned with FSDP.
3. CI tests have been added to check the effect.

### High-Level Design

This PR eliminates the micro-dp group as the article says, since the `generate_sequence` process only relates to inference engine, there is no need for us to consider the training side.

The only problem now is that the `dispatch/collect` function cannot directly use the inference parallel size, so current solution is that we define a new `MEGATRON_ALL_DP` dispatch method to view all ranks as Data Parallel rank, which is the same as FSDP.

So we follow the way of FSDP to pre/post-process the data.

### Specific Changes

Mainly in `megatron_vllm.py`

### API

None

### Usage Example

```sh
actor_rollout_ref.actor.megatron.tensor_model_parallel_size=2 \
actor_rollout_ref.rollout.tensor_model_parallel_size=4 \

# or

actor_rollout_ref.actor.megatron.tensor_model_parallel_size=4 \
```

### Test

Added CI tests.

For e2e test with Qwen 2.5 7B, please refer to `examples/grpo_trainer/run_qwen2_5-7b_math_megatron_diff_tp.sh`

### Additional Info.

- **Training**: Megatron
- **Inference**: vLLM


ccclyu:
@ann-qin-lu tested and seems that the hanging issue has not been resolved. How about making it WIP and dive deep for debug?

verl/workers/sharding_manager/megatron_vllm.py
```diff
@@ -350,6 +331,9 @@ def tensor_generator():
 
         # lazy load tensor for full model
         for cur_pp_rank, scan_vpp_idx, idx, name in layer_list_meta:
+            if self.model_config.tie_word_embeddings and ("output_layers" in name):
+                print("[WARNING] Current model sharing word and embedding weights, skip output layer conversio")
```
eric-haibin-lin: conversio -> conversion

eric-haibin-lin: use warnings.warn? do they get printed many times?

eric-haibin-lin (commented):
nitpick (not critical for merge)

verl/workers/sharding_manager/megatron_vllm.py
ETOgaosion: Get it! I will see how to print warning in other places.

verl/single_controller/base/decorator.py
```diff
@@ -328,6 +344,23 @@ def collect_megatron_pp_as_dp_data_proto(worker_group, output):
...
+    # Note: enable auto padding for dp compute DatapProto
+    splitted_args, splitted_kwargs = _split_args_kwargs_data_proto(
+        worker_group.world_size,
+        *args,
+        **kwargs,
+    )
+    return splitted_args, splitted_kwargs
+
+
+def collect_megatron_all_dp_data_proto(worker_group, output):
```
vermouth1992: Shall we reuse existing ones?

ETOgaosion: I think we may consider for future use, inference engine may have pp or ep support, so by then we may need to use a better implementation in Megatron side. So it's a space-occupation interface.

ETOgaosion: But FSDP may also need the modification, if you suggest that there is no need for maintaining two functions for each, I can remove it.

vermouth1992: I guess we can simply dispatch DP outside and perform allgather inside. In this way, the single controller doesn't care the parallelism strategy inside the workers.

vermouth1992: In fact, we are considering removing MegatronWorkerGroup in this PR https://github.com/volcengine/verl/pull/1260

vermouth1992: The current MegatronWorkerGroup design is quite limited and needs generalization

ETOgaosion: OK, get it.

vermouth1992 (approved):

PR #1463, opened 2025-05-09
[constant scheduler] fix: model won't be updated on first training step

0x404:
### What does this PR do?

I found that when using the FSDP checkpoint test introduced by https://github.com/volcengine/verl/pull/1288, after one step of training, both comparisons pass the tests. This includes comparing the merged FSDP checkpoint with the verl-saved HF model, and comparing the merged FSDP checkpoint with the original HuggingFace model. This means the FSDP model is not being updated after one step of training.

However, the training log shows that in the first step, the learning rate is 1e-6, which is weird. I found two issues in the existing code:

1. There's a problem with `get_constant_schedule_with_warmup`: when `num_warmup_steps=0`, at the first step (`current_step=0`), it will return `0.0` instead of `1.0`. This is wrong and inconsistent with the existing constant LR definition: https://github.com/huggingface/transformers/blob/774dc274ac966f4bccbcd90d55bba23f6cca37ae/src/transformers/optimization.py#L72

2. The log saves the learning rate after `actor_lr_scheduler.step()`, which is incorrect since it records the next step's LR, thus hiding the problem with `get_constant_schedule_with_warmup`.

This PR fixes these issues and decrease tolerance in FSDP checkpoint test.

### Additional Info.

- **Training**: FSDP
- **Inference**: both


eric-haibin-lin (approved):

0x404:
Hi @eric-haibin-lin, Could you re-review this, just resolve conflicts several days ago

eric-haibin-lin (approved):

PR #1464, opened 2025-05-09
[rollout] fix: Make `free_cache_engine` option workable in latest vLLM/SGLang

HollowMan6:
### Checklist Before Starting

### What does this PR do?

Make `free_cache_engine` option workable in latest vLLM/SGLang

### High-Level Design

It looks like `actor_rollout_ref.rollout.free_cache_engine` control option only works for vLLM version 0.5.4 and 0.6.3, the sleep / wake up mode in vLLM engine, as well as release / resume memory occupation in SGLang is enabled by default and there's no way to turn them off.

While always alllowing inference engine to free cache can be ideal, it's unfortunately not supported on some devices, such as AMD Mi250x, since it doesn't support virtual memory management: https://github.com/vllm-project/vllm/pull/12695#issuecomment-2633919751

So we would need to be able to turn it off so that verl can run on those devices.

In addition, it looks like we no longer need to enforce eager in latest vLLM when we choose to free cache, so this PR also lifted this restriction.

### Additional Info.

- **Issue Number**: None
- **Training**: both FSDP and Megatron
- **Inference**: both vLLM and SGLang

- [X] Add CI test(s) if neccessary.


zhaochenyang20:
We gonna to review SGLang part these days.

zhaochenyang20:
@yushengsu-thu yusheng could you help on the AMD part. SInce release KV cache and offload works slow on ROCM

yushengsu-thu:
@zhaochenyang20 , actually me and @YangWang92 working on this.
vllm part: https://github.com/vllm-project/vllm/pull/12695#issuecomment-2633919751
sglang part: https://github.com/sgl-project/sglang/issues/5093#issuecomment-2785084577, https://github.com/fzyzcjy/torch_memory_saver/issues/9#issuecomment-2850012653
This issue can be solved. Stay tuned.

HollowMan6:
> Btw, That said, this feature seems more like a general one, as the issue of memory not being properly released can occur when certain kernels lack proper support (I’ve seen others encounter similar problems on different GPUs as well) so we can add it.

@yushengsu-thu Yes, and not only about that, I think we will also need this PR to make verl support older AMD GPUs, such as the MI250x I'm using, since they unfortunately don't support virtual memory management APIs because of the hardware. https://github.com/ROCm/hip-tests/blob/84a460d96bafb00615969304cc5eaddc3b20bc3d/catch/unit/memory/hip_vmm_common.hh#L27-L37 

I was the original author of https://github.com/vllm-project/vllm/pull/12695, and I've been following your progress after I found out about the issue for supporting sleep mode on MI250x.
It would be hard to support it unless we no longer rely on virtual memory management APIs / find an alternative.

verl/workers/rollout/vllm_rollout/vllm_rollout_spmd.py
```diff
@@ -126,7 +130,7 @@ def __init__(self, model_path: str, config: DictConfig, tokenizer, model_hf_conf
 
         self.inference_engine = LLM(
             model=model_path,
-            enable_sleep_mode=True,
+            enable_sleep_mode=config.free_cache_engine,
```
yushengsu-thu, line 149: @HollowMan6 Have you ever tried to set `enable_sleep_mode` as `False` in vllm>=0.8. I remember that we tried but it will encounter another issue, (@YangWang92 @gowtham1997) right?

verl/workers/rollout/sglang_rollout/sglang_rollout.py
```diff
@@ -195,7 +193,8 @@ def __init__(
         )
 
         # offload
-        self.inference_engine.release_memory_occupation()
+        if config.free_cache_engine:
```
yushengsu-thu: @HollowMan6 Have you ever tried disabling `release_memory_occupation()` and running any of the example scripts? If so, could you please share the script with me? I remember running the same script and reducing its `gpu_memory_utilization` to below 0.2, but it still encountered an OOM issue.

HollowMan6:
@yushengsu-thu To the best of my knowledge, I'm not aware of different ROCm APIs for achieving the same feature, but this is definitely something that's worth considering, and I would assume that you can check with your AMD colleagues, and they might have something in mind.

For other designs, when I initially opened the feature request for sleep mode at the vLLM side: https://github.com/vllm-project/vllm/issues/10714, I did send a temporary PR of my solution for destroying all KV cache during runtime: https://github.com/vllm-project/vllm/pull/10810. This one is device-agnostic as it doesn't use any lower API and fully relies on the garbage collection, but as it needs to make sure we clear out all the reference counters, it's not very stable and fully transparent to the application, and might be hard to maintain in the long run. So I guess that's why they choose to come up with a more elegant solution https://github.com/vllm-project/vllm/pull/11743, that involves this low-level virtual memory management API.

yushengsu-thu (approved):
LGTM

verl/workers/rollout/sglang_rollout/sglang_rollout.py
yushengsu-thu: [Done]

verl/workers/rollout/vllm_rollout/vllm_rollout_spmd.py
yushengsu-thu: [Done]

zhaochenyang20:
> LGTM

I am running the CI. We can merge it after your approval. thanks!

HollowMan6:
The latest CI failure https://github.com/volcengine/verl/actions/runs/14958501396/job/42018309219?pr=1464 doesn't seem to be this PR specific, maybe it's a good idea to rerun it again.

> (WorkerDict pid=15082) [2025-05-11 19:51:54 TP0] Scheduler hit an exception: Traceback (most recent call last):
(WorkerDict pid=15082)   File "/usr/local/lib/python3.10/dist-packages/sglang/srt/managers/scheduler.py", line 2014, in run_scheduler_process
(WorkerDict pid=15082)     scheduler = Scheduler(server_args, port_args, gpu_id, tp_rank, dp_rank)
(WorkerDict pid=15082)   File "/usr/local/lib/python3.10/dist-packages/sglang/srt/managers/scheduler.py", line 258, in __init__
(WorkerDict pid=15082)     self.tp_worker = TpWorkerClass(
(WorkerDict pid=15082)   File "/usr/local/lib/python3.10/dist-packages/sglang/srt/managers/tp_worker_overlap_thread.py", line 63, in __init__
(WorkerDict pid=15082)     self.worker = TpModelWorker(server_args, gpu_id, tp_rank, dp_rank, nccl_port)
(WorkerDict pid=15082)   File "/usr/local/lib/python3.10/dist-packages/sglang/srt/managers/tp_worker.py", line 75, in __init__
(WorkerDict pid=15082)     self.model_runner = ModelRunner(
(WorkerDict pid=15082)   File "/usr/local/lib/python3.10/dist-packages/sglang/srt/model_executor/model_runner.py", line 177, in __init__
(WorkerDict pid=15082)     min_per_gpu_memory = self.init_torch_distributed()
(WorkerDict pid=15082)   File "/usr/local/lib/python3.10/dist-packages/sglang/srt/model_executor/model_runner.py", line 360, in init_torch_distributed
(WorkerDict pid=15082)     init_distributed_environment(
(WorkerDict pid=15082)   File "/usr/local/lib/python3.10/dist-packages/sglang/srt/distributed/parallel_state.py", line 1052, in init_distributed_environment
(WorkerDict pid=15082)     torch.distributed.init_process_group(
(WorkerDict pid=15082)   File "/usr/local/lib/python3.10/dist-packages/torch/distributed/c10d_logger.py", line 81, in wrapper
(WorkerDict pid=15082)     return func(*args, **kwargs)
(WorkerDict pid=15082)   File "/usr/local/lib/python3.10/dist-packages/torch/distributed/c10d_logger.py", line 95, in wrapper
(WorkerDict pid=15082)     func_return = func(*args, **kwargs)
(WorkerDict pid=15082)   File "/usr/local/lib/python3.10/dist-packages/torch/distributed/distributed_c10d.py", line 1714, in init_process_group
(WorkerDict pid=15082)     store, rank, world_size = next(rendezvous_iterator)
(WorkerDict pid=15082)   File "/usr/local/lib/python3.10/dist-packages/torch/distributed/rendezvous.py", line 226, in _tcp_rendezvous_handler
(WorkerDict pid=15082)     store = _create_c10d_store(
(WorkerDict pid=15082)   File "/usr/local/lib/python3.10/dist-packages/torch/distributed/rendezvous.py", line 194, in _create_c10d_store
(WorkerDict pid=15082)     return TCPStore(
(WorkerDict pid=15082) RuntimeError: The server socket has failed to listen on any local network address. port: 30856, useIpv6: 0, code: -98, name: EADDRINUSE, message: address already in use

docs/README_vllm0.7.md
```diff
@@ -49,7 +49,7 @@ After installation, examples using FSDP as training backends can be used. By def
 
 ```
 actor_rollout_ref.rollout.enforce_eager=False \
-actor_rollout_ref.rollout.free_cache_engine=False \
+actor_rollout_ref.rollout.free_cache_engine=True \
```
eric-haibin-lin, line 52: @ZSL98 BTW why is free_cache_engine set to False for the initial integration?

HollowMan6: My guess is because it's a history problem. For vllm v0.5.4 and v0.6.3, we need to disable the usage of CUDAGraph (set `enforce_eager` to True) when we set `free_cache_engine` to True because of the custom implementation limitation. But when the sleep mode is introduced in 0.7.2, this is no longer needed, but somehow this limit was forgotten to be lifted.

https://github.com/volcengine/verl/blob/8e5ad4688a13de81727c014a3c2e2fb26324bc20/verl/workers/rollout/vllm_rollout/vllm_rollout.py#L79

That's why I added a vLLM version check for this assertion in this PR.

HollowMan6:
@zhaochenyang20 @yushengsu-thu @eric-haibin-lin Any plan to merge this soon? New conflicts constantly arise, and I hope to have this quickly merged so that I don't need to spend time resolving them.

yushengsu-thu:
> @zhaochenyang20 @yushengsu-thu @eric-haibin-lin Any plan to merge this soon? New conflicts constantly arise, and I hope to have this quickly merged so that I don't need to spend time resolving them.

@eric-haibin-lin @zhaochenyang20 . It looks good to me. Since the current version uses a hybrid engine design, some potential issues have arisen that were not present in the disaggregated design. Releasing and resuming memory at the kernel level is one such issue (AMD side also has this issue, but we have already solved it, and we will release the updated Docker image soon). I’m aware that other company GPUs face similar problems if they haven't modified their memory allocation implementation (here: https://github.com/fzyzcjy/torch_memory_saver/issues/9#issuecomment-2910919609), and they might also encounter this issue. Therefore, merging this feature could enable more users (who are not using NVIDIA GPUs) to train RL models on their setups (it just requires more memory during training).

verl/workers/rollout/vllm_rollout/vllm_rollout_spmd.py
```diff
@@ -87,7 +87,11 @@ def __init__(self, model_path: str, config: DictConfig, tokenizer, model_hf_conf
         """
         super().__init__()
         self.config = config
-        assert not (not config.enforce_eager and config.free_cache_engine), "disable CUDA graph (enforce_eager = False) if free cache engine"
+        if vllm_version in (
```
eric-haibin-lin: nit: better to keep the versions in verl/workers/rollout/vllm_rollout/ as a constant and refer to this constant in all these places instead (or have a util function for version check).

eric-haibin-lin (approved):
left a minor comment. overall looks good to me

verl/workers/rollout/vllm_rollout/vllm_rollout_spmd.py
HollowMan6: Put a `customized_vllm` flag under `verl/third_party/vllm/__init__.py`

https://github.com/volcengine/verl/pull/1464/commits/e63fb27c81c55f31a97c2b41f683d6e1647f13d7

eric-haibin-lin (dismissed):

yushengsu-thu (approved):

eric-haibin-lin (approved):
hey @HollowMan6 , sorry for the late review. Somehow this PR was neglected by me in the past few weeks and thank you for resolving the conflict with verl main multiple times.  In verl main the wake_up/sleep call of the async rollout is moved to the `generate_sequence` API instead of leaving it to the trainer (this way trainer code is simpler)https://github.com/volcengine/verl/pull/2124/files . I've adjusted your code based on that recent change (https://github.com/volcengine/verl/pull/1464/commits/753a69c6f02a1e184a5c1a7ff4aad7feec4e5a6a). Overall it looks good to me. We can merge it once all CI tests pass.

verl/workers/rollout/sglang_rollout/sglang_rollout.py
```diff
@@ -1074,7 +1072,7 @@ def _req_level_generate_sequences(self, prompts: DataProto, **kwargs) -> DataPro
         )
 
         # free cache engine
-        if self.config.free_cache_engine and self._engine is not None and self._tp_rank == 0:
+        if self._engine is not None and self._tp_rank == 0:
```
eric-haibin-lin, line 1075: hmmm @HollowMan6 why removing this condition?

HollowMan6: The `free_cache_engine` configuration will be misinterpreted if we don't remove it here. As `flush_cache()` is to flush the radix cache in SGLang, it's not related to release/resume memory occupation (as per the meaning in vLLM), and we should still flush the radix cache even if release/resume memory occupation is not enabled, to make sure the radix cache is cleared and not reused in the next generation.

cc: @zhaochenyang20 for confirmation.

eric-haibin-lin:
Merged for now. Thanks for the contribution. @HollowMan6 in the next PR would you mind adding an unit test with free_cache_engine=False and run `generate_sequence`, just to make sure this option will not be broken again in the future? Testing with the latest sglang/vllm should be sufficient. Thanks!

Cysu:
This seems to break when setting `free_cache_engine` to `False`, at least for the async vLLM rollout -- the first rollout was good, but after the actor update, the second rollout would be corrupted (generating random tokens). Setting `free_cache_engine` to `True` is fine.

HollowMan6:
> This seems to break when setting `free_cache_engine` to `False`, at least for the async vLLM rollout -- the first rollout was good, but after the actor update, the second rollout would be corrupted (generating random tokens). Setting `free_cache_engine` to `True` is fine.

One possible reason can be that some code forgets to do `free_cache_engine` checks before `sleep` calls, but if I remember correctly, vLLM should give out errors if it's in sleep mode and there are new requests. Feel free to submit a PR if you somehow spot the culprit.

PR #1470, opened 2025-05-10
[refactor][single_controller] Small refactor and fixes in `worker.py` and `ray.base.py` 

hongpeng-guo:
### Checklist Before Starting

### What does this PR do?

1. The existing `WorkerHelper::_get_pid()` returns nothing; this PR fix this nit by returning `os.getpid()`;
2. The `WorkerMeta` class in `worker.py` is only used in `Worker.init()` and no where else. This class just maintains a list of env keys and  wraps a dict named `store`. This PR delete this class and move the contents inside the `Worker` class. In this way, it would be easier if a user want to subclass Worker, with different env keys;
3. In `merge_resource_pool` function, instead of directly return a `RayResourcePool`, this PR changes the return type to be `type(rp1)`. In this way, the function can be applied to not only `RayResourcePool`, but also any subclass of `RayResourcePool`.

### Note

This PR splits off some small nits and refactors from #1454, so that the small things here could be reviewed and merged sooner before we decide on the structural PR #1454


eric-haibin-lin (commented):
how to extend to more env keys?

hongpeng-guo:
> how to extend to more env keys?

In case of subclassing this Worker, we can simply override its `env_keys()` method, i.e.,

```
class MyWorker(Worker):
    @classmethod
    def env_keys():
        return super().env_keys() + ["MY_KEY_1", "MY_KEY_2"]
```
The old way of using `WorkerMeta` will require subclassing `WorkerMeta`, which makes the process over complicated.

vermouth1992 (approved):

PR #1505, opened 2025-05-13
[Doc/Docker Image] Update mcore image to use vLLM which support qwen3 and rewrite installation from conda

ETOgaosion:
### Checklist Before Starting

### What does this PR do?

Update mcore image to use vLLM which support qwen3 and rewrite installation from conda

### High-Level Design

### Specific Changes

Docker image and docs

### API

### Usage Example

```python
```

### Test

### Additional Info.

- **Training**: both
- **Inference**: both


install.sh
```diff
@@ -0,0 +1,48 @@
+#!/bin/bash
```
vermouth1992: What's the usage of this file? Why should it be put in the root folder?

vermouth1992: Shall we move this file to another location?

ETOgaosion: It's actually the installation script needed for local install verl, I will figure out how other projects place scripts like this

ETOgaosion: Can we place the install script in the `dev` folder? Maybe later we can remove the scripts and refactor the `setup.py` to better install  these dependencies.

vermouth1992: How about under scripts?

vermouth1992: Nevermind, I guess the current path is fine. Let's merge this and not trigger CI again :)

ETOgaosion: > How about under scripts?

I suggest that current `scripts` folder may better rename to sth like `tools` :), and we can place some shell scripts to the original `scripts`. Maybe we can try when refactoring codes.

vermouth1992 (approved):

PR #1509, opened 2025-05-13
[fix][DataProto] Make classmethod  `from_single_dict` return a cls not the class name

hongpeng-guo:
### Checklist Before Starting

### What does this PR do?

`from_single_dict` is a classmethod, right now, it directly returns `DataProto.from_dict(.....)`. This PR changes it to `cls.from_dict(.....)`, In this way any subclass of `DataProto` may reuse this classmethod to instantiate a subclass.

In the current implementation, when subclass, i.e., `MyDataProto.from_single_dict()` is called, it returns a parent class instance, i.e., a `DataProto`, but not a `MyDataProto` instance.

### DIFF
--- a/verl/protocol.py
+++ b/verl/protocol.py
@@ -153,7 +153,7 @@ def fold_batch_dim(data: "DataProto", new_batch_size):
     for key, val in non_tensor.items():
         non_tensor[key] = np.reshape(val, newshape=(new_batch_size, -1, *val.shape[1:]))
 
-    return DataProto(batch=tensor, non_tensor_batch=non_tensor, meta_info=data.meta_info)
+    return type(data)(batch=tensor, non_tensor_batch=non_tensor, meta_info=data.meta_info)
 
 
 def unfold_batch_dim(data: "DataProto", batch_dims=2):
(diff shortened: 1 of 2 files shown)


verl/protocol.py
```diff
@@ -338,7 +338,7 @@ def from_single_dict(cls, data: Dict[str, Union[torch.Tensor, np.ndarray]], meta
             else:
                 raise ValueError(f"Unsupported type in data {type(val)}")
 
-        return DataProto.from_dict(tensors=tensors, non_tensors=non_tensors, meta_info=meta_info, auto_padding=auto_padding)
+        return cls.from_dict(tensors=tensors, non_tensors=non_tensors, meta_info=meta_info, auto_padding=auto_padding)
```
eric-haibin-lin, line 341: we can add a dummy sub class with a test?

hongpeng-guo: @eric-haibin-lin  Updated the unit test. Also fixed a few instance methods that directly return `DataProto(.....)`, now they are returning `type(self)(......)`. In this way, the subclass can use the parent classes.

vermouth1992 (approved):

Issue #1221, opened 2025-04-23
[RFC] TorchRPC as Single-Controller Backend


zTonyZhao:
# Motivation

verl was designed in layered architecture, using WorkerGroups with Ray backend as the single-controller of the RL workflow. Ray is well known for its easy expansion and pythonic interface. However, there exist users that are not familiar with Ray, hoping for an alternative backend for verl. Thus, an implementation of WorkerGroup using TorchRPC from PyTorch offers this community a second choice.

# Proposed Add-Ons

New low-level modules including `TorchRPCResourcePool`, `TorchRPCClassWithInitArgs` and `TorchRPCWorkerGroup` will be implemented using TorchRPC in replace of Ray.

We try our best to keep APIs aligned to those in Ray-backend. Certain features like detached workers and named workers will not be implemented due to limitations of TorchRPC.

# Demo

Here follows an equivalent test of [tests/ray/test_worker_group_basics.py](https://github.com/volcengine/verl/tree/main/tests/ray/test_worker_group_basics.py) using torchrpc backend.

No modifications other than module importing are needed, with the `@register` decorator and the dispatch-collect method stay the same.

```python
import os
import torch
import torch.distributed.rpc as rpc

from verl.single_controller.base.decorator import Dispatch, Execute, collect_all_to_all, register
from verl.single_controller.base.worker import Worker
from verl.single_controller.torchrpc import TorchRPCResourcePool, TorchRPCClassWithInitArgs, TorchRPCWorkerGroup, rref_to_here

def two_to_all_dispatch_fn(worker_group, *args, **kwargs):
    """
    Assume the input is a list of 2. Duplicate the input interleaved and pass to each worker.
    """
    for arg in args:
        assert len(arg) == 2
        for i in range(worker_group.world_size - 2):
            arg.append(arg[i % 2])
    for k, v in kwargs.items():
        assert len(v) == 2
        for i in range(worker_group.world_size - 2):
            v.append(v[i % 2])
    return args, kwargs

class TestActor(Worker):
    def __init__(self, x) -> None:
        super().__init__()
        self._x = x

    def foo(self, y):
        return self._x + y

    @register(dispatch_mode=Dispatch.ALL_TO_ALL, execute_mode=Execute.RANK_ZERO)
    def foo_rank_zero(self, x, y):
        return self._x + y + x

    @register(Dispatch.ONE_TO_ALL, blocking=False)
    def foo_one_to_all(self, x, y):
        return self._x + y + x

[...]

PR #1519, opened 2025-05-14
[feat] TorchRPC as Single-Controller Backend

zTonyZhao:
### Checklist Before Starting

### What does this PR do?

add TorchRPC as single-controller backend, with new modules including `TorchRPCResourceGroup`, `TorchRPCClassWithInitArgs`, `TorchRPCWorkerGroup`, as described in #1221.

### High-Level Design

We use TorchRPC to build a process-isolated worker runtime(at single-controller/torchrpc/utils.py) and a resource dispatch service(at single-controller/torchrpc/node.py), which are used to achieve similar abilities of Ray backend. `TorchRPCResourceGroup`, `TorchRPCClassWithInitArgs`, `TorchRPCWorkerGroup` are built on top of them, these modules keep nearly identical APIs of Ray equivalents, as described below.

### API

**TORCHRPC BACKEND ONLY WORKS ON PATCHED PYTORCH.**

Colocated worker could not be pickled by default pickler used by TorchRPC. Patch is needed by changing the default pickler from pickle to dill in `site-packages/torch/distributed/rpc/internal.py`. You can change it manually or run the patch script located at verl/single_controller/torchrpc/patch_torchrpc.py.

Current APIs using Ray will not be impacted at all.

New APIs of TorchRPC backend:

#### `@torchrpc_remote`

this decorator is needed on the entry function and is used for TorchRPC initialization. Usage of this decorator is further explained in example section.

#### `TorchRPCResourceGroup`

```
TorchRPCResourceGroup(
        process_on_nodes: List[int] = None,
        use_gpu: bool = True,
        max_colocate_count: int = 5,
)
```

create a resource group with `process_on_nodes` workers.

`name_prefix` and `detached` is removed due to lack of support on detached and named resource.

#### `TorchRPCClassWithInitArgs`

```
TorchRPCClassWithInitArgs(cls, *args, **kwargs)
```

A worker class with init args.

underlying APIs keep the same with Ray.

#### `TorchRPCWorkerGroup`

```
TorchRPCWorkerGroup(
        resource_pool: TorchRPCResourcePool = None,
        cls_with_init: TorchRPCClassWithInitArgs = None,
        **kwargs)
```

create a workergroup with `cls_with_init` worker on `resource_pool`.

since colocate worker group with FusedWorker has `spawn` implemented using deep copy and no longer needs detach workers, detached or named actors are no longer required.

underlying APIs keep the same with Ray.

### Usage Example

Here is an example partly copied from tests/torchrpc/tests_worker_group_basics.py:

```python
class TestActor(Worker):
    def __init__(self, x) -> None:
        super().__init__()
        self._x = x

    def foo(self, y):
        return self._x + y

    @register(dispatch_mode=Dispatch.ALL_TO_ALL, execute_mode=Execute.RANK_ZERO)
    def foo_rank_zero(self, x, y):
        return self._x + y + x

    @register(Dispatch.ONE_TO_ALL, blocking=False)
    def foo_one_to_all(self, x, y):
        return self._x + y + x
        
@torchrpc_remote
def test_basics():
    resource_pool = TorchRPCResourcePool([2,2], use_gpu=True)
    class_with_args = TorchRPCClassWithInitArgs(cls=TestActor, x=2)
    worker_group = TorchRPCWorkerGroup(
        resource_pool=resource_pool, cls_with_init=class_with_args
    )

    output = worker_group.execute_all_sync("foo", y=3)
    assert output == [5, 5, 5, 5]

    # this is a list of object reference. It won't block.
    output_ref = worker_group.execute_all_async("foo", y=4)
    assert rref_to_here(output_ref) == [6, 6, 6, 6]
    
    output_ref = worker_group.foo_one_to_all(x=1, y=2)
    assert rref_to_here(output_ref) == [5, 5, 5, 5]
```

Each custom Worker class derived from the Worker class does not need to add the `@ray.remote()` decorator.

By adding the `@register()` decorator before the member function, it can be directly called in the WorkerGroup and run on each Worker according to the distribution and execution strategy specified by the decorator.

Add the decorator `@torchrpc_remote` above the entry function of the program. This decorator is used for the initialization process of TorchRPC. Under this decorator, each node initializes TorchRPC connection first. Then the function wrapped will be run only at master node. It also closes RPC connection upon finish.

When running the task, it is necessary to run the entry program on each node and add the following environment variables:

```bash
MASTER_ADDR=2001::1 \
MASTER_PORT=12345 \
TORCHRPC_RANK=0 \
TORCHRPC_WORLD_SIZE=2 \
python3 tests/torchrpc/test_worker_group_basics.py
```

These environment variables will be used to initialize TorchRPC. The meanings of each environment variable are as follows:

- `MASTER_ADDR`: Address of master node
- `MASTER_PORT`: Port of master node
- `TORCHRPC_RANK`: Index o current node, ranging from [0, TORCHRPC_WORLD_SIZE)
- `TORCHRPC_WORLD_SIZE`: Total number of nodes

### Test

This implementation has already pass single-controller tests, including:

- test_worker_group_basics.py
- test_worker_group_torch.py
- test_colocated_workers.py
- test_data_transfer.py
- test_driverfunc_to_worker.py
- test_high_level_scheduling_api.py

Tests are located in `tests/torchrpc` folder.

RayPPOTrainer is also migrated to TorchRPCPPOTrainer, which has passed e2e test `arithmetic_sequence` with tests/e2e/arithmetic_sequence/main_trainer_torchrpc.py. The reward curves using Ray and TorchRPC are identical, which proves the correctness of this implementation.

### Additional Info.

- **Issue Number**: #1221 
- **Training**: both
- **Inference**: both


zTonyZhao:
Added github CI tests for torchrpc backend.

verl/single_controller/torchrpc/patch_torchrpc.py
```diff
@@ -0,0 +1,36 @@
...
+file_to_patch = os.path.join(module_path, "internal.py")
+print("Patching", file_to_patch)
+with open(file_to_patch, 'r', encoding='utf-8') as f:
+    lines = f.readlines()
+modified = False
+new_lines = []
+for line in lines:
+    stripped_line = line.strip()
+    if not modified and stripped_line == 'import pickle':
+        new_lines.append('import dill as pickle\n')
```
ezyang, line 27: There's some precedent for using dill in PyTorch when it is available (torch/utils/data/datapipes/datapipe.py) so I think it might be possible to just upstream this? It's a little weird though because datapipe.py seems to try to pickle first (for BC?) before using dill, which doesn't seem like it would be very fast, although I see that it would improve BC. Another option is to modify the APIs to take a pickle module as argument, similar to how torch.serialization does it. You can tag me on a PR for this, although you should probably ping me from here as I'm on parental leave and not rigorously checking my PyTorch notifications right now.

zTonyZhao: After extensive research, I have found that TorchRPC provides an interface `_use_rpc_pickler` for pickler override. However replacing the underlying pickle engine using that interface is quite complicated, and more importantly **that interface was broken**. A bugfix as well as a new interface for overriding underlying pickle engine was provided at https://github.com/pytorch/pytorch/pull/153821. Thank you so much for your review!

PR #1529, opened 2025-05-15
[misc] fix: no need to use world_size to decide whether to use full_tensor in FSDP2

PeterSH6:
[misc] fix: no need to use world_size to decide whether to use full_tensor() for FSDP2 state_dict() when world_size==1

### Checklist Before Starting

### What does this PR do?

This PR simplifies the parameter loading logic within the `FSDPVLLMShardingManager` by removing an unnecessary `world_size` check when determining whether to call `full_tensor()` on parameters obtained from an FSDP2 model's `state_dict()`. As the FSDP2 parameters are all `DTensor`.

### High-Level Design

The change modifies the update_params method. When loading weights into the vLLM model, parameters from the FSDP state_dict() (which might be ShardedTensor or DTensor instances under FSDP2 when world_size == 1) are converted to full tensors using param.full_tensor(). This PR ensures this conversion happens if the full_tensor() method is available on the parameter, without an additional, potentially incorrect, check against world_size == 1.

### Specific Changes

Skip. See file changes

### API

No

### Usage Example

No

### Test

No CI changes

### Additional Info.

- **Issue Number**: No
- **Training**: [Note which backend this PR will affect: FSDP
- **Inference**: [Note which backend this PR will affect: vLLM

### DIFF
--- a/verl/workers/sharding_manager/fsdp_vllm.py
+++ b/verl/workers/sharding_manager/fsdp_vllm.py
@@ -188,7 +188,6 @@ class FSDPVLLMShardingManager(BaseShardingManager):
     def update_params(self, updated_params):
         model = self.model_runner.model
         patch_vllm_moe_model_weight_loader(model)
-        world_size = torch.distributed.get_world_size()
         device = torch.cuda.current_device()  # used when fsdp2 set cpu_offload_policy
-        loaded_params = model.load_weights(((name, param.to(device, non_blocking=True).full_tensor() if world_size != 1 and hasattr(param, "full_tensor") else param) for name, param in updated_params.items()))
+        loaded_params = model.load_weights(((name, param.to(device, non_blocking=True).full_tensor() if hasattr(param, "full_tensor") else param) for name, param in updated_params.items()))
         logger.info("vLLM load weights, loaded_params: %d", len(loaded_params))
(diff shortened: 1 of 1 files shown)


verl/workers/sharding_manager/fsdp_vllm.py
```diff
@@ -188,7 +188,6 @@ def postprocess_data(self, data: DataProto) -> DataProto:
     def update_params(self, updated_params):
         model = self.model_runner.model
         patch_vllm_moe_model_weight_loader(model)
-        world_size = torch.distributed.get_world_size()
         device = torch.cuda.current_device()  # used when fsdp2 set cpu_offload_policy
-        loaded_params = model.load_weights(((name, param.to(device, non_blocking=True).full_tensor() if world_size != 1 and hasattr(param, "full_tensor") else param) for name, param in updated_params.items()))
+        loaded_params = model.load_weights(((name, param.to(device, non_blocking=True).full_tensor() if hasattr(param, "full_tensor") else param) for name, param in updated_params.items()))
```
vermouth1992: How about isinstance(param, DTensor)?

vermouth1992 (approved):

verl/workers/sharding_manager/fsdp_vllm.py
PeterSH6: Good suggestion. Let me fix it.

PR #1467, opened 2025-05-09
[megatron] support megatron expert parallel

ISEEKYAN:
### What does this PR do?

support expert parallel in megatron

### High-Level Design

introduce EPsize and ETPsize
ETPsize is the TPsize for MoE parts, recommended to set 1, meaning that MoE parts not use TP

### Specific Changes

1. mcore model initilize
2. megatron vllm parameter transfer

### Usage Example

```bash
LLM=models/Qwen1.5-MoE-A2.7B-Chat
NODES=1
PP=2
TP=4
VLLM_TP=4
EP=4
ETP=1

python3 -m verl.trainer.main_ppo --config-path=./config --config-name='ppo_megatron_trainer'\
    algorithm.adv_estimator=gae \
    data.train_files="$train_files" \
    data.val_files="$test_files" \
    data.train_batch_size=128 \
    data.max_prompt_length=1024 \
    data.max_response_length=512 \
    data.filter_overlong_prompts=True \
    data.truncation='error' \
    actor_rollout_ref.model.path=$LLM \
    actor_rollout_ref.actor.optim.lr=1e-6 \
    actor_rollout_ref.actor.ppo_mini_batch_size=32 \
    actor_rollout_ref.actor.ppo_micro_batch_size_per_gpu=1 \
    actor_rollout_ref.actor.use_kl_loss=False \
    actor_rollout_ref.rollout.log_prob_micro_batch_size_per_gpu=2 \
    actor_rollout_ref.rollout.name=vllm \
    actor_rollout_ref.rollout.gpu_memory_utilization=0.7 \
    critic.optim.lr=1e-5 \
    critic.model.path=$LLM \
    critic.model.enable_gradient_checkpointing=False \
    critic.ppo_micro_batch_size_per_gpu=1 \
    algorithm.use_kl_in_reward=False \
    trainer.critic_warmup=0 \
    trainer.logger=['console','wandb'] \
    trainer.project_name='verl_megatron_gsm8k_examples' \
    trainer.experiment_name='qwen_moe_instruct_1node_ep' \
    trainer.n_gpus_per_node=8 \
    trainer.nnodes=$NODES \
    trainer.save_freq=-1 \
    trainer.test_freq=5 \
    actor_rollout_ref.rollout.tensor_model_parallel_size=$VLLM_TP \
    actor_rollout_ref.actor.megatron.pipeline_model_parallel_size=$PP \
    actor_rollout_ref.ref.megatron.pipeline_model_parallel_size=$PP \
[...]

PR #1602, opened 2025-05-20
[sglang] Fix megatron support in sglang and add sglang_async support & CI tasks

SwordFaith:
### Checklist Before Starting

### What does this PR do?

- Fix sglang megatron support
- Add sglang_async megatron support
- Add CI task to protect megatron-sglang impl

### Test

https://wandb.ai/swordfaith/gsm8k_async_rl/runs/6h7apmbn?nw=nwuserswordfaith

### Additional Info.


.github/workflows/e2e_ppo_trainer_megatron.yml
```diff
@@ -265,3 +265,229 @@ jobs:
       - name: clean up
         run: |
           rm -rf checkpoints
+
+  e2e_ppo_trainer_megatron-qwen-sgl:
```
ETOgaosion: Now this CI file has become so complicate that is hard to maintain both vLLM and SGLang tests. My suggestion is that we can add both vLLM and SGLang tests in a single shell script test file [https://github.com/volcengine/verl/blob/main/tests/e2e/run_ppo_trainer_megatron.sh](https://github.com/volcengine/verl/blob/main/tests/e2e/run_ppo_trainer_megatron.sh), and we simply need to call this file once.

tests/e2e/run_ppo_trainer_megatron.sh
```diff
@@ -9,6 +9,8 @@ MODEL_ID=${MODEL_ID:-Qwen/Qwen2.5-0.5B}
 MODEL_PATH=${MODEL_PATH:-${HOME}/models/${MODEL_ID}}
 huggingface-cli download "${MODEL_ID}" --local-dir "${MODEL_PATH}"
 
+ENGINE=${ENGINE:-vllm}
```
ETOgaosion: I mean here we can also include SGLang python scripts below the vLLM, so that we can test two systems with the same configuration.

verl/workers/sharding_manager/megatron_sglang.py
```diff
@@ -34,61 +54,152 @@
...
+    @GPUMemoryLogger(role="megatron sglang sharding_manager", logger=logger)
+    def preprocess_data(self, data: DataProto) -> DataProto:
+        # DP_COMPUTE_PROTO: all training ranks are dp, the same as fsdp
+        if self.infer_tp_size == 1:
+            return data
+        all_gather_data_proto(data, self.device_mesh["tp"].get_group())
+        return data
+
+    @GPUMemoryLogger(role="megatron sglang sharding_manager", logger=logger)
+    def postprocess_data(self, data: DataProto) -> DataProto:
```
ETOgaosion, line 132: I read somewhere that the SGLang's TP rank output might be different from vLLM's, as now the dispatch method is the same with FSDP, can we just borrow the fsdp_sglang.py's implementation to avoid misalignment?
In [here](https://github.com/volcengine/verl/blob/15b1b15f9963e178b68368c9b3996c60637a5156/verl/workers/sharding_manager/fsdp_sglang.py#L168)

tests/e2e/run_ppo_trainer_megatron.sh
SwordFaith: OK, I will modify it configuration after CI test finished

verl/workers/sharding_manager/megatron_sglang.py
SwordFaith: It seems that we currently lack a unified weight conversion method for FSDP as Megatron Core, so we cannot merge these two classes for now. Are there any better approaches to implement this?

tests/e2e/run_ppo_trainer_megatron.sh
SwordFaith: Ok, I'll modify this.

verl/workers/sharding_manager/megatron_sglang.py
ETOgaosion: Oh, I mean not merging these classes, we can borrow the implementation there, just like this vLLM's implementation

SwordFaith:
> @SwordFaith Thanks for contribution, could you rebase main and add some megatron_sglang support of training side expert parallel as #1467 ? And you can use the latest image to add some tests~

For sglang, it seems that support has been added in utils/megatron_utils as part of 1467. It would be better to share the megatron_utils per_tensor_generator implementation between megatron vllm and sglang. However, I believe such a refactor would require end-to-end training verification, which could take days to modify and debug. For now, we can address sglang support in the current PR and plan the refactor for the future.

PR #1616, opened 2025-05-21
Expose `engine_kwargs` from SGLang to Verl configuration

hebiao064:
### Checklist Before Starting

### What does this PR do?

Expose `engine_kwargs` from SGLang to Verl configuration

This PR enables RL users to configure `engine_kwargs` directly through Verl, providing more control and flexibility over inference behavior.

### High-Level Design

One key motivation is the choice of attention backend, which can significantly affect rollout performance. The SGLang team has observed that different attention backends perform better in different phases:

FA3 tends to be more efficient during the prefill stage.

FlashInfer or Triton generally offer better performance during decode.

Moreover, the optimal backend may change across versions of SGLang. By exposing these parameters, we allow users to tune their setup based on the specific use case and version, ultimately improving performance and adaptability.

### Specific Changes

> List the specific changes.

### API

### Usage Example

```python
```

### Test

In my setup about QWen 2.5 7B Instruct on H200:
```
timing_s/step:106.761 (flash infer)
timing_s/step:100.520 （fa3)
timing_s/step:100.364 (triton)
```

Hence, I would suggest our team to use fa3 or triton for now.

### Additional Info.

### DIFF
--- a/docs/examples/config.rst
+++ b/docs/examples/config.rst
@@ -170,6 +170,7 @@ Actor/Rollout/Reference Policy
       do_sample: True
       engine_kwargs: # inference engine parameters
         swap_space: null # null means "use the engine default value" (usually 4 GB), setting it to, e.g., 32 means 32 GB
+        attention_backend: fa3 # null means use the engine default value, available options: flashinfer, triton, flashmla
       # number of responses (i.e. num sample times)
       n: 1 # > 1 for grpo, rloo
       val_kwargs:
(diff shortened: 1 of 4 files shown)


zhaochenyang20:
great work. I think for VLM, FA3 could be even better?

SwordFaith:
Which backend is best as default ?

zhaochenyang20:
> Which backend is best as default ?

right now any of it is okay. I think fa3 is default?

hebiao064:
> Which backend is best as default ?

It's determined by SGLang's logic, for now, it's fa3 on hopper, flashinfer on other arch.

But even on hopper or ampere, sometimes we found triton is faster, that's why we need this granularity.

hebiao064:
> great work. I think for VLM, FA3 could be even better?

It depends on the batch size, context size and response length, at least we can provide the freedom to user to decide.

vermouth1992:
@zhaochenyang20 Shall we merge this?

zhaochenyang20 (approved):

zhaochenyang20:
@vermouth1992 thanks, merged

PR #1617, opened 2025-05-21
[feat][BREAKING] Megatron support dynamic batch size, to rebalance the workloads

ETOgaosion:
### Checklist Before Starting

### What does this PR do?

1. Megatron support dynamic batch size, to rebalance the workloads.
2. Fix missing critic metrics.

### High-Level Design

Follow the FSDP's dynamic batch size.

### Specific Changes

Use the `rearrange_micro_batches` API, but compatible with Megatron VPP constraints.

```py
vpp_size = mpu.get_virtual_pipeline_model_parallel_world_size()
if vpp_size is not None and vpp_size > 1:
    microbatch_group_size_per_vp_stage = self.tf_config.microbatch_group_size_per_vp_stage
    micro_batches, indices = rearrange_micro_batches(batch=mini_batch.batch, num_batches_devided_by=microbatch_group_size_per_vp_stage, max_token_len=max_token_len)
    assert len(micro_batches) % self.tf_config.microbatch_group_size_per_vp_stage == 0, f"micro_batches {micro_batches} must be divisible by microbatch_group_size_per_vp_stage {microbatch_group_size_per_vp_stage} for megatron backend"
else:
    micro_batches, indices = rearrange_micro_batches(batch=mini_batch.batch, max_token_len=max_token_len)
```

@vermouth1992 please check whether it makes sense.

Megatron's constraint when using interleaving pipeline:

```py
# If the final micro-batch group has fewer micro-batches than pipeline-parallel size,
    # the pipeline will have dependency bubbles.
    final_microbatch_group_size = num_microbatches % config.microbatch_group_size_per_vp_stage
    if 0 < final_microbatch_group_size < pipeline_parallel_size:
        msg = 'The remainder of M (the total micro-batches) divided by N (number of '
        msg += 'contiguous micro-batches in a virtual pipeline stage) should be 0, '
        msg += 'or larger than or equal to the pipeline-parallel size, but it is '
        msg += f'{final_microbatch_group_size}. '
        msg += 'Otherwise, it introduces dependency bubbles in the pipeline '
        msg += 'and reduces throughput.'
        raise RuntimeError(msg)
```

### API

Megatron forward_backward_batch has changed input, and the output has become a dict, containing original `output` and the `indices` needed for compute_old_log_probs.

### Usage Example

```bash
    actor_rollout_ref.actor.use_dynamic_bsz=${USE_DYNAMIC_BSZ} \
    actor_rollout_ref.actor.ppo_max_token_len_per_gpu=${ppo_max_token_len_per_gpu} \
    critic.ppo_max_token_len_per_gpu=${forward_max_token_len_per_gpu} \
```

Other models will directly copy the config.

### Test

### Additional Info.

### DIFF
--- a/verl/utils/seqlen_balancing.py
+++ b/verl/utils/seqlen_balancing.py
@@ -216,7 +216,7 @@ def roundup_divisible(a, b):
     return ((a + b - 1) // b) * b
 
 
-def rearrange_micro_batches(batch, max_token_len, dp_group=None, num_batches_devided_by=None, same_micro_num_in_dp=True, min_num_micro_batch=None):
+def rearrange_micro_batches(batch, max_token_len, dp_group=None, num_batches_divided_by=None, same_micro_num_in_dp=True, min_num_micro_batch=None):
     """
     Split a batch into micro-batches by total token count, with optional DP sync and padding.
 
@@ -224,7 +224,7 @@ def rearrange_micro_batches(batch, max_token_len, dp_group=None, num_batches_dev
         batch (TensorDict): must include "attention_mask" (B*S); other fields are sliced similarly.
         max_token_len (int): max sum of attention_mask per micro-batch.
         dp_group (optional): torch.distributed group for data-parallel sync.
-        num_batches_devided_by (optional): virtual pipeline parallel size, for megatron.
+        num_batches_divided_by (optional): virtual pipeline parallel size, for megatron.
         same_micro_num_in_dp (bool): if True and dp_group set, pad all ranks to the same count.
         min_num_micro_batch (int, optional): force at least this many splits (pads empty ones).
 
@@ -246,8 +246,8 @@ def rearrange_micro_batches(batch, max_token_len, dp_group=None, num_batches_dev
         num_micro_batches = torch.tensor([num_micro_batches], device="cuda")
         dist.all_reduce(num_micro_batches, op=dist.ReduceOp.MAX, group=dp_group)
         num_micro_batches = num_micro_batches.cpu().item()
-    if num_batches_devided_by is not None:
-        num_micro_batches = roundup_divisible(num_micro_batches, num_batches_devided_by)
+    if num_batches_divided_by is not None:
+        num_micro_batches = roundup_divisible(num_micro_batches, num_batches_divided_by)
 
     seq_len_effective = seq_len_effective.tolist()
     assert num_micro_batches <= len(seq_len_effective)
--- a/verl/workers/reward_model/megatron/reward_model.py
+++ b/verl/workers/reward_model/megatron/reward_model.py
@@ -220,7 +220,7 @@ class MegatronRewardModel(BasePPORewardModel):
             vpp_size = mpu.get_virtual_pipeline_model_parallel_world_size()
             if vpp_size is not None and vpp_size > 1:
                 microbatch_group_size_per_vp_stage = self.tf_config.microbatch_group_size_per_vp_stage
-                micro_batches, indices = rearrange_micro_batches(batch=mini_batch.batch, num_batches_devided_by=microbatch_group_size_per_vp_stage, max_token_len=max_token_len)
+                micro_batches, indices = rearrange_micro_batches(batch=mini_batch.batch, num_batches_divided_by=microbatch_group_size_per_vp_stage, max_token_len=max_token_len)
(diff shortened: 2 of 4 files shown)


verl/workers/reward_model/megatron/reward_model.py
```diff
@@ -184,27 +200,37 @@ def compute_reward(self, data: DataProto) -> DataProto:
...
+        indices = None
+        if use_dynamic_bsz:
+            assert max_token_len is not None, "max_token_len must be set when use_dynamic_bsz is True"
+            vpp_size = mpu.get_virtual_pipeline_model_parallel_world_size()
+            if vpp_size is not None and vpp_size > 1:
+                microbatch_group_size_per_vp_stage = self.tf_config.microbatch_group_size_per_vp_stage
+                micro_batches, indices = rearrange_micro_batches(batch=mini_batch.batch, num_batches_devided_by=microbatch_group_size_per_vp_stage, max_token_len=max_token_len)
+                assert len(micro_batches) % self.tf_config.microbatch_group_size_per_vp_stage == 0, f"micro_batches {micro_batches} must be divisible by microbatch_group_size_per_vp_stage {microbatch_group_size_per_vp_stage} for megatron backend"
+            else:
+                micro_batches, indices = rearrange_micro_batches(batch=mini_batch.batch, max_token_len=max_token_len)
```
vermouth1992, line 226: Does len(mini_batch) has to be greater than pp_size?

ETOgaosion: For the interleaving pipeline, it's true, while for the normal pipeline, there is no such constraint or assertion.

ETOgaosion: Do we need assertions here?

vermouth1992: Got it.

ISEEKYAN (approved):

verl/utils/seqlen_balancing.py
```diff
@@ -212,14 +212,19 @@ def ceildiv(a, b):
     return -(a // -b)
 
 
-def rearrange_micro_batches(batch, max_token_len, dp_group=None, same_micro_num_in_dp=True, min_num_micro_batch=None):
+def roundup_divisible(a, b):
+    return a + (a % b)
+
+
+def rearrange_micro_batches(batch, max_token_len, dp_group=None, num_batches_devided_by=None, same_micro_num_in_dp=True, min_num_micro_batch=None):
```
ccclyu: should be a typo? ``num_batched_devided_by`` -> ``num_batched_divided_by``.  there are multiple places appearing.

verl/utils/seqlen_balancing.py
```diff
@@ -212,14 +212,19 @@ def ceildiv(a, b):
...
     """
     Split a batch into micro-batches by total token count, with optional DP sync and padding.
 
     Args:
         batch (TensorDict): must include "attention_mask" (B*S); other fields are sliced similarly.
         max_token_len (int): max sum of attention_mask per micro-batch.
         dp_group (optional): torch.distributed group for data-parallel sync.
+        vpp_size (optional): virtual pipeline parallel size, for megatron.
```
ccclyu: is ``vpp_size`` in the ``rearrange_micro_batches`` definition?

verl/utils/seqlen_balancing.py
```diff
@@ -212,14 +212,19 @@ def ceildiv(a, b):
     return -(a // -b)
 
 
-def rearrange_micro_batches(batch, max_token_len, dp_group=None, same_micro_num_in_dp=True, min_num_micro_batch=None):
+def roundup_divisible(a, b):
+    return a + (a % b)
```
ccclyu: i might think the implantation is not correct. how about the following one?  you can double check it.

``
    return ((a + b - 1) // b) * b
``

verl/utils/seqlen_balancing.py
ETOgaosion: Thanks for reviewing~ Maybe I already fixed in previous commit?

verl/utils/seqlen_balancing.py
ETOgaosion: Yeah, this is correct, I tried to use `%`, which has error logic

verl/utils/seqlen_balancing.py
ETOgaosion: I used to pass vpp_size, which is an error, use `num_batches_devided_by` instead~

verl/utils/seqlen_balancing.py
ccclyu: no.  it should be `divided` not `devided`.   previous commit did not fix this.  there are multiple places using `devided`.

vermouth1992 (approved):

PR #1623, opened 2025-05-21
Copy torch.Tensor from "multi_modal_inputs" to device too

GHGmc2:
### Checklist Before Starting

### What does this PR do?

### High-Level Design

### Specific Changes

> List the specific changes.

### API

### Usage Example

```python
```

### Test

### Additional Info.

### DIFF
--- a/verl/protocol.py
+++ b/verl/protocol.py
@@ -389,6 +389,11 @@ class DataProto:
         """
         if self.batch is not None:
             self.batch = self.batch.to(device)
+        # copy torch.Tensor from "multi_modal_inputs" to device too
+        if "multi_modal_inputs" in self.non_tensor_batch.keys():
+            for input in self.non_tensor_batch["multi_modal_inputs"]:
+                for key in input.keys():
+                    input[key] = input[key].to(device=device)
         return self
 
     def select(self, batch_keys=None, non_tensor_batch_keys=None, meta_info_keys=None, deepcopy=False) -> "DataProto":


GHGmc2:
@vermouth1992 @hiyouga any comments?

For the data passed to `ActorRolloutRefWorker`, we copy it to the GPU device firstly:
```python
data = data.to(get_torch_device().current_device())
```

But the `torch.Tensor` in "non_tensor_batch" field is not moved to GPU, and we should execute `torch.cat()` on GPU instead of CPU for better performance:
```python
class DataParallelPPOActor(BasePPOActor):
    def _forward_micro_batch():
        if "multi_modal_inputs" in micro_batch:
            for key in micro_batch["multi_modal_inputs"][0].keys():
                multi_modal_inputs[key] = torch.cat([inputs[key] for inputs in micro_batch["multi_modal_inputs"]], dim=0)
```

verl/protocol.py
```diff
@@ -389,6 +389,11 @@ def to(self, device) -> "DataProto":
         """
         if self.batch is not None:
             self.batch = self.batch.to(device)
+        # copy torch.Tensor from "multi_modal_inputs" to device too
+        if "multi_modal_inputs" in self.non_tensor_batch.keys():
```
hiyouga, line 393: we should avoid introducing some logic in protocol.py, can we move them to fsdp_worker and megatron worker?

GHGmc2: > we should avoid introducing some logic in protocol.py, can we move them to fsdp_worker and megatron worker?

May be we should move `torch.Tensor` in "multi_modal_inputs" from "non_tensor_batch" field of DataProto, which is align with the the name "non_tensor"?

hiyouga: no

Issue #1517, opened 2025-05-14
[Bug]FSDP2 failed to load large model state_dict


GHGmc2:
I found that FSDP2 failed to load large(32B or 72B) model state_dict. And it works after I changed the "fsdp2" part in the cmd below to "fsdp":
```bash
actor_rollout_ref.actor.strategy=fsdp \
actor_rollout_ref.ref.strategy=fsdp \
```

Config
- GPU: 64 H20-96GB cards
- model: Qwen2.5-VL-32B
- dataset: geo3k

Command w/ FSDP2
```bash
MODEL_PATH=/path/to/Qwen2.5-VL-32B
DATA_PATH=/path/to/geo3k

python3 -m verl.trainer.main_ppo \
      algorithm.adv_estimator=grpo \
      data.train_files=$DATA_PATH/train.parquet \
      data.val_files=$DATA_PATH/test.parquet \
      data.train_batch_size=512 \
      data.max_prompt_length=1024 \
      data.max_response_length=2048 \
      data.filter_overlong_prompts=True \
      data.truncation='error' \
      data.image_key=images \
      actor_rollout_ref.model.path=$MODEL_PATH \
      actor_rollout_ref.model.use_remove_padding=True \
      actor_rollout_ref.model.enable_gradient_checkpointing=True \
      actor_rollout_ref.actor.strategy=fsdp2 \
      actor_rollout_ref.actor.optim.lr=1e-6 \
      actor_rollout_ref.actor.ppo_mini_batch_size=128 \
      actor_rollout_ref.actor.ppo_micro_batch_size_per_gpu=10 \
      actor_rollout_ref.actor.use_kl_loss=True \
      actor_rollout_ref.actor.kl_loss_coef=0.01 \
      actor_rollout_ref.actor.kl_loss_type=low_var_kl \
      actor_rollout_ref.actor.entropy_coeff=0 \
      actor_rollout_ref.rollout.name=vllm \
      actor_rollout_ref.rollout.log_prob_micro_batch_size_per_gpu=20 \
      actor_rollout_ref.rollout.tensor_model_parallel_size=4 \
      actor_rollout_ref.rollout.gpu_memory_utilization=0.6 \
      actor_rollout_ref.rollout.enable_chunked_prefill=True \
      actor_rollout_ref.rollout.enforce_eager=False \
      actor_rollout_ref.rollout.free_cache_engine=False \
      actor_rollout_ref.rollout.n=5\
[...]

PR #1667, opened 2025-05-24
[fsdp2] fix: oom issue when loading model state dict in fsdp2

0x404:
### Checklist Before Starting

### What does this PR do?

When using FSDP2, loading the model consumes huge reserved memory, causing OOM issues. See the issue at https://github.com/volcengine/verl/issues/1517.

There are several issues:
1. The offload function won't actually reduce used GPU memory: https://github.com/volcengine/verl/blob/5dc64391fec0c7a829b5d175e25cf8d7056a49d5/verl/utils/fsdp_utils.py#L151-L156
Actually, simple `model.to("cpu")` works.

2. Using `set_model_state_dict` will consume a large amount of reserved memory, which cannot be released simply using `torch.cuda.empty_cache()`. I found that offloading FSDP2 module to CPU then back to GPU can release this memory and the memory consumption is consistent with FSDP1.
But I think this is a temporary solution since it requires one more offload compared to FSDP1. If someone has a better solution please let me know.

3. I think we should have consistent parameters among `param_offload`, `optimizer_offload` and `offload_policy`, since some users will set `param_offload=True and offload_policy=False` and expect their model to be offloaded, but actually it won't.

### Test
Maybe we should also add FSDP e2e test into CI?

### Additional Info.

- **Issue Number**: Fixes issue #1517
- **Training**: FSDP2
- **Inference**: none


vermouth1992:
Thank you for your contribution! I also find FSDP2 init very difficult as it removes the param_init_fn in FSDP1. It seems that the ultimate solution is to transform the pretraining weights to dcp in advance

vermouth1992:
CPU init can only scale to 32b and we have to find alternative solutions for larger models

weifengpy:
@0x404 thanks for the investigation. this is a bug in torch==2.6.0. we fixed it in torch==2.7.0 https://github.com/pytorch/pytorch/pull/134025

eric-haibin-lin:
patched by torch team in https://github.com/volcengine/verl/pull/2606. @0x404 thank you for the efforts

PR #1668, opened 2025-05-24
[WIP][Async SGLang Rollout] Efficient and model-agnostic multi-turn messages tokenization and masking

jybsuper:
### Checklist Before Starting

### What does this PR do?

Implement efficient, model-agnostic multi-turn message tokenization and masking based solely on the chat template defined in the model config.

### High-Level Design

### Specific Changes

> List the specific changes.

### API

### Usage Example

```python
```

#### Benchmark of theoretical acceleration
```python
conversation = data[589]["messages"]#*30
print(f"Total Turns: {len(conversation)}; Total Tokens: {len(tokenizer.apply_chat_template(conversation, tokenize=True))}")
repeat = 10

# SFT implementation
elapsed1 = 0
for _ in tqdm(range(repeat), total=repeat, desc="SFT"):
    for i in range(2, len(conversation)):
        start = time.time()
        tokenizer.apply_chat_template(conversation[: i-1], tokenize=True, add_generation_prompt=True, return_tensors="pt")
        tokenizer.apply_chat_template(conversation[:i], tokenize=True, add_generation_prompt=False, return_tensors="pt")
        elapsed1 += (time.time() - start)
elapsed1 /= repeat

# New RL implementation
elapsed2 = 0
for _ in tqdm(range(repeat), total=repeat, desc="RL"):
    for i in range(2, len(conversation)):
        start = time.time()
        pos = len(tokenizer.apply_chat_template(conversation[: i-1], tokenize=False, add_generation_prompt=True))
        new_message = tokenizer.apply_chat_template(conversation[:i], tokenize=False, add_generation_prompt=False)[pos:]
        tokenizer.encode(new_message, add_special_tokens=False)
        elapsed2 += (time.time() - start)
elapsed2 /= repeat

print(f"Full Tokenization: {elapsed1:.4f}s, Additive Tokenization: {elapsed2:.4f}s")
```
Result on M1 Pro(Observerd bigger improvement on AMD EPYC 9554)
```
Total Turns: 403; Total Tokens: 32899
Full Tokenization: 60.6303s, Additive Tokenization: 1.9332s
```
#### Correctness validation
1. Validated on my own multiturn RL project(QwQ-32B + GRPO)
2. [WIP] Validating with Verl's multiturn GSM8K example
3. [WIP] Validating single dataset with async SGLang rollout

#### E2E Benchmark
WIP

### Additional Info.


verl/workers/rollout/schemas.py
```diff
@@ -67,7 +71,9 @@ class AsyncRolloutRequest(BaseModel):
     request_id: str
     state: AsyncRolloutRequestStateEnum
     messages: List[Message]
-    tools: Optional[List[OpenAIFunctionToolSchema]] = None
+    messages_dumps: List[Dict[str, Any]]
```
SwordFaith: Why do we need `message_dumps` here? I assume the `Message` data class primarily serves as a validator to prevent implicit schema changes. If iteratively dumping `Message` objects incurs significant cost, it could be limited to usage only before adding any messages to the messages list, while maintaining the messages list as `List[Dict[str, Any]]`.

verl/workers/rollout/schemas.py
```diff
@@ -67,7 +71,9 @@ class AsyncRolloutRequest(BaseModel):
...
+    tool_schemas: Optional[List[OpenAIFunctionToolSchema]] = None
```
SwordFaith, line 76: same for tool_schema

verl/trainer/config/ppo_megatron_trainer.yaml
```diff
@@ -164,7 +164,12 @@ actor_rollout_ref:
       enable: False  # should set rollout.name to sglang_async if True
       max_turns: null  # null for no limit (default max_length // 3)
       tool_config_path: null  # null for no tool
-      format: chatml  # chatml, more formats will be supported in the future
+      format: chatml  # deprecated, no effect anymore
```
SwordFaith: Could we remove this unnecessary option to avoid misleading users into thinking it offers deprecated support instead of being fully removed?

verl/trainer/config/ppo_trainer.yaml
```diff
@@ -136,7 +136,12 @@ actor_rollout_ref:
       enable: False  # should set rollout.name to sglang_async if True
       max_turns: null  # null for no limit (default max_length // 3)
       tool_config_path: null  # null for no tool
-      format: chatml  # chatml, more formats will be supported in the future
+      format: chatml  # deprecated, no effect anymore
```
SwordFaith: same

verl/workers/rollout/schemas.py
```diff
@@ -82,127 +88,121 @@ class AsyncRolloutRequest(BaseModel):
...
-        return tokenizer.apply_chat_template(  # type: ignore
-            conversation=[msg.model_dump() for msg in self.messages],
-            tools=[tool.model_dump() for tool in self.tools] if self.tools else None,
-            add_generation_prompt=True,
-            tokenize=True,
-        )
+    tokenization_mode: Literal["fast", "full", "sanity_check"]
+    generation_prompt_ids: List[int]
+
+    @model_validator(mode="before")
```
SwordFaith, line 102: Good practice, learned a lot from this.

verl/workers/rollout/schemas.py
```diff
@@ -82,127 +88,121 @@ class AsyncRolloutRequest(BaseModel):
...
+        if not (max_prompt_len := values.get("max_prompt_len")):
+            raise ValueError("max_prompt_len is required for AsyncRolloutRequest initialization")
+        if not (tokenizer := values.pop("tokenizer", None)):
+            raise ValueError("tokenizer is required for AsyncRolloutRequest initialization")
+
+        values["messages"] = [Message.model_validate(msg) for msg in messages]
+        values["messages_dumps"] = [msg.model_dump() for msg in values["messages"]]
+
+        if tool_schemas := values.get("tool_schemas"):
+            tools = values["tools"] = [tool.model_dump() for tool in tool_schemas]
```
SwordFaith: Can tools []/None be used directly to switch here? Rather than use if-else. And I'm a little confused why we need pre configured input_ids and attention mask here, it seems can be supported as lazy tokenization (only do tokenize when missing) ?

verl/workers/rollout/schemas.py
```diff
@@ -82,127 +88,121 @@ class AsyncRolloutRequest(BaseModel):
...
+        else:
+            tokens_without_prompt = tokenizer.apply_chat_template(messages, add_generation_prompt=False, tokenize=True)
+            values["prompt_ids"], values["prompt_attention_mask"] = values["input_ids"], values["attention_mask"]
+
+        values["position_ids"] = values["prompt_position_ids"] = compute_position_id_with_mask(torch.tensor(values["attention_mask"])).tolist()
+        values["loss_mask"] = values["prompt_loss_mask"] = [0] * len(values["input_ids"])
+        values["generation_prompt_ids"] = values["input_ids"][len(tokens_without_prompt) :]
+        return values
+
+    def _update_input_ids(self, new_input_ids: List[int], attention_mask: bool, loss_mask: bool, full_tokens: bool = False) -> None:
```
SwordFaith: Great refactor! It's advisable to add docstrings for both `_update_input_ids` and `_tokenize_all_messages`, as they update`ids` in different contexts, which can be somewhat confusing. One potential improvement is to create a `tokenize_messages` method that switches between the current supported tokenization modes, enhancing readability.

verl/workers/rollout/schemas.py
```diff
@@ -82,127 +88,121 @@ class AsyncRolloutRequest(BaseModel):
...
+
+    def _append_messages(self, messages: list[Message]) -> None:
+        self.messages.extend(messages)
+        self.messages_dumps.extend([msg.model_dump() for msg in messages])
+
+    def _tokenize_all_messages(self, tokenizer: PreTrainedTokenizer, delta_input_ids_to_check: Optional[list[int]], add_generation_prompt: bool = False) -> None:
+        full_input_ids = tokenizer.apply_chat_template(self.messages_dumps, tools=self.tools, add_generation_prompt=add_generation_prompt, tokenize=True)
+        if self.tokenization_mode == "sanity_check" and delta_input_ids_to_check is not None:
+            assert full_input_ids == self.input_ids + delta_input_ids_to_check, (
+                f"Sanity check failed.\nFull tokenization result:\n{tokenizer.decode(full_input_ids, skip_special_tokens=False)}\nFast tokenization result:\n{tokenizer.decode(self.input_ids + delta_input_ids_to_check, skip_special_tokens=False)}"
```
SwordFaith: Is there a method to separate enabling or disabling sanity from full tokenization or fast tokenization?

verl/workers/rollout/schemas.py
```diff
@@ -82,127 +88,121 @@ class AsyncRolloutRequest(BaseModel):
...
-            last_position_id = self.position_ids[-1]
-            _position_ids = [pos_id + last_position_id for pos_id in _delta_position_ids]
-            self.loss_mask += [0] * len(append_token_ids)
-            self.position_ids += _position_ids
-        else:
-            raise ValueError(f"Unsupported format: {format}")
-        assert len(self.input_ids) == len(self.attention_mask) == len(self.position_ids) == len(self.loss_mask), f"""Request {self.request_id} has different length of {len(self.input_ids)=}, 
-            {len(self.attention_mask)=}, {len(self.position_ids)=}, {len(self.loss_mask)=}"""
+        if self.tokenization_mode != "full":
+            if tool_calls or not content_ids:
```
SwordFaith: Should we integrate the logic for handling `tool_calls` into `_update_input_ids` or other function specifically for fast tokenization handling? It will be better as tokenization mode handling function input.

verl/workers/rollout/schemas.py
```diff
@@ -82,127 +88,121 @@ class AsyncRolloutRequest(BaseModel):
...
+
+    def _tokenize_all_messages(self, tokenizer: PreTrainedTokenizer, delta_input_ids_to_check: Optional[list[int]], add_generation_prompt: bool = False) -> None:
+        full_input_ids = tokenizer.apply_chat_template(self.messages_dumps, tools=self.tools, add_generation_prompt=add_generation_prompt, tokenize=True)
+        if self.tokenization_mode == "sanity_check" and delta_input_ids_to_check is not None:
+            assert full_input_ids == self.input_ids + delta_input_ids_to_check, (
+                f"Sanity check failed.\nFull tokenization result:\n{tokenizer.decode(full_input_ids, skip_special_tokens=False)}\nFast tokenization result:\n{tokenizer.decode(self.input_ids + delta_input_ids_to_check, skip_special_tokens=False)}"
+            )
+        self._update_input_ids(full_input_ids, attention_mask=True, loss_mask=False, full_tokens=True)
+
+    def get_prompt_ids(self, tokenizer: PreTrainedTokenizer) -> list[int]:
```
SwordFaith: Why do we still need tokenization here? The name `prompt_ids` seems to follow the sglang naming convention, which might cause confusion between the token ids of verl's `prompt/response` and sglang's `prompt_ids`. A better name could be something like `get_generation_prompt_ids` or another name that clearly distinguishes it from verl's prompt token ids.

verl/workers/rollout/schemas.py
```diff
@@ -82,127 +88,121 @@ class AsyncRolloutRequest(BaseModel):
...
+        return values
+
+    def _update_input_ids(self, new_input_ids: List[int], attention_mask: bool, loss_mask: bool, full_tokens: bool = False) -> None:
+        message_len_delta = (len(new_input_ids) - len(self.input_ids)) if full_tokens else len(new_input_ids)
+        self.input_ids = new_input_ids if full_tokens else (self.input_ids + new_input_ids)
+        attention_mask = [int(attention_mask)] * message_len_delta
+        self.attention_mask += attention_mask
+        _delta_position_ids = compute_position_id_with_mask(torch.tensor(attention_mask)).tolist()
+        last_position_id = self.position_ids[-1]
+        _position_ids = [pos_id + last_position_id for pos_id in _delta_position_ids]
```
SwordFaith: _position_ids = (_delta_position_ids + (last_position_id + 1)).tolist() may be a good fix. Before impl I made a bug here, but won't trigger due to regenerated position_ids when build batch. https://github.com/volcengine/verl/blob/main/verl/workers/rollout/sglang_rollout/async_sglang_rollout.py#L413

verl/workers/rollout/schemas.py
jybsuper: You are right, pydantic model to dict should be fast. `message_dumps ` removed now

verl/workers/rollout/schemas.py
jybsuper: removed

verl/trainer/config/ppo_megatron_trainer.yaml
jybsuper: Done

verl/trainer/config/ppo_trainer.yaml
jybsuper: Done

verl/workers/rollout/schemas.py
jybsuper: I was trying to copy the logic from async_sglang_rollout, but I agree with you. reusing the input ids is better. logic updated. Thanks!

verl/workers/rollout/schemas.py
jybsuper: Thanks for the suggestion! Done

verl/workers/rollout/schemas.py
jybsuper: I moved sanity check codes to a separate case in `_tokenize_messages`. Does this new implementaion look good to you?

verl/workers/rollout/schemas.py
jybsuper: Sure. Moved it to a separate `_fast_tokenize` method

verl/workers/rollout/schemas.py
jybsuper: renamed it to `get_generation_prompt_ids`.

verl/workers/rollout/schemas.py
jybsuper: Good catch!

SwordFaith:
1. sanity check default on, and add `skip_sanity_check` is an option
2. data & rollout config have chat_template option, default null means use origin chat template, otherwise should be a jinja file , pre-placed jinjas (QwQ & Qwen3)
3. User specific chat template e.g. search r1 / retool original version , fast concat logic support, our special finetune tag.
4. White list for known models, store chat template hash due to mismatch in train/inference / can't pass sanity check. Replace chat template by default.

docs/sglang_multiturn/multiturn.rst
```diff
@@ -40,6 +40,82 @@ Finally, set the ``tools_config_file`` in your rollout config:
...
+If the discrepancy is expected, you can disable the sanity check via:
+
+``actor_rollout_ref.rollout.multi_turn.enable_tokenization_sanity_check=False``
+
+Special Cases
+^^^^^^^^^^^^^
+
+Some models (e.g., Qwen/QwQ-32B and Qwen3 series) remove internal reasoning content during chat template rendering. As a result, the message content can vary across turns, making the delta-based tokenization inaccurate.
+
+To handle this, we fall back to a **fixed base conversation** containing only a single system and user message. Since this base doesn’t include assistant messages or reasoning content, it remains consistent across turns.
```
SwordFaith, line 81: Since this base doesn’t include assistant messages or reasoning content, it remains consistent across turns. 

We should emphasize the differences between the turns managed by the last assistant turn (QwQ-32B) and those managed by the assistant turn after the user's last input (Qwen3), as these discrepancies lead to trajectory variations. Clarifying this will make the current implementation easier to understand for future users and developers.

docs/sglang_multiturn/multiturn.rst
```diff
@@ -40,6 +40,82 @@ Finally, set the ``tools_config_file`` in your rollout config:
...
+This mismatch can affect model performance in unpredictable ways. To avoid it, we default to using the full response (including reasoning) for both training and rollout.
+
+However, this approach comes with trade-offs:
+
+1. Long reasoning contents can easily exceed the model’s context window, especially in multi-turn rollout.
+2. There’s a mismatch between rollout and production environment now—models will not have reasoning content from past turns if you use the default chat template in production.
+
+We are still evaluating the impact of these issues. If you experience context length problems or prefer rollouts that match production (i.e., exclude reasoning), you can enable:
+
+``actor_rollout_ref.rollout.multi_turn.use_inference_chat_template = True``
```
SwordFaith, line 117: In the current design, how can we implement a sanity check feature for users to test their models? Additionally, `enable_tokenization_sanity_check` is a crucial feature for developers. Could you assist in adding the related documentation?

verl/workers/rollout/schemas.py
```diff
@@ -221,6 +186,19 @@ def finalize(
     ) -> None:
         self.state = AsyncRolloutRequestStateEnum.COMPLETED
         self.reward_scores = reward_scores
+        if self.enable_tokenization_sanity_check:
```
SwordFaith, line 189: Sorry, I didn’t quite understand your point about reducing the sanity check overhead from the previous discussion. I think there are two approaches to perform sanity checks: one is to check only at the end of the rollout, and the other is to check at each turn (which allows for early termination if unexpected divergence occurs). If divergence happens, perhaps both approaches could include options for logging (with wandb metrics) and raising exceptions, which might be more effective. I’d love to hear your thoughts on this. For now, I suggest keeping the current implementation as is and creating an issue to track potential future improvements.

jybsuper: The reason I changed to this implementation is that:
1. santy check at each round means we need to apply chat template on the converstaion again and again and this makes all the tokenization slower.
2. I think it might be better to log warnings than throw errors as models like QwQ will always cause divergence and throwing error means users have to turn off sanity check for all those models although the divergence actually is ok for those models. Turning off sanity check in their shell script means there is higher chance they forgot to enable it again for other models and this is the last thing we want to see. Therefore, sanity check in each turn only produces more warnings instead of early termination in this design.

Does it make sense to you?

docs/sglang_multiturn/multiturn.rst
jybsuper, line 186: Users do not have to do any development to sanity check tokenization for their models. They just run experiements with new models as usual. If there is a consistency issue, they will see many warnings in their log.

Could you elaborate a bit more why `enable_tokenization_sanity_check` is crucial for developer? I think it is pretty static logic in our codebase and neither users or developer gonna have to change it. What do you think will be helpful other than the descriptions in the doc?

SwordFaith: Sorry for missing out above doc.

verl/workers/rollout/schemas.py
SwordFaith: I agree. The current approach is a good practice to maintain an always-on version for sanity checks.

SwordFaith:
LGTM

casper-hansen:
Hi @vermouth1992 this PR broke async vLLM Server. Considering this was an update for SGLang, I am quite surprised.

CC @wuxibin89

Here is the problem:
- omegaconf.errors.ConfigAttributeError: Key 'format' is not in struct due to removal of it in this PR.
- But vLLM Async Server depends on this parameter. And when you try to add `actor_rollout_ref.rollout.multi_turn.format=hermes`, it doesn't work

https://github.com/volcengine/verl/blob/49b08e95090fc1924161c640d190637337a747b9/verl/workers/rollout/vllm_rollout/vllm_async_server.py#L206

casper-hansen:
CC @wuxibin89

jybsuper:
> Hi @vermouth1992 this PR broke async vLLM Server. Considering this was an update for SGLang, I am quite surprised.
> 
> CC @wuxibin89
> 
> Here is the problem:
> 
> * omegaconf.errors.ConfigAttributeError: Key 'format' is not in struct due to removal of it in this PR.
> * But vLLM Async Server depends on this parameter. And when you try to add `actor_rollout_ref.rollout.multi_turn.format=hermes`, it doesn't work
> 
> https://github.com/volcengine/verl/blob/49b08e95090fc1924161c640d190637337a747b9/verl/workers/rollout/vllm_rollout/vllm_async_server.py#L206

Hi @casper-hansen, thanks for reporting this issue!

The `format` field was removed 2 weeks ago when this PR was created, at which point the vLLM async server wasn't utilizing this field. All vLLM tests passed during the PR review process, which is why the removal proceeded.

That said, I'm quite surprised that vLLM has been repurposing the chat template `format` field for tool parser functionality. As you may have noticed, the default value `chatml` is actually a chat template format name from SGLang's multiturn rollout - it's not even a valid tool parser identifier in vLLM's server. This creative reuse of fields intended for different purposes does create unnecessary confusion. Moving forward, it would be beneficial if vLLM could add a dedicated field for the tool parser rather than overloading existing fields with unrelated functionality. This would help prevent such integration issues.

Given the collaborative nature of VeRL and the number of moving parts, tracking cross-PR dependencies is indeed challenging, especially when PRs have long review cycles. This is exactly where proper tests become critical. I'd suggest updating the test [here](https://github.com/volcengine/verl/blob/main/tests/workers/rollout/test_vllm_chat_scheduler.py#L31) to include `OmegaConf.set_struct(config, True)` to catch any unexpected removal of config fields.

Let me know if you'd like help fixing this—happy to contribute a patch to resolve the issue properly.

casper-hansen:
@jybsuper I would appreciate a patch for this so that the ChatCompletionScheduler can use similar config arguments. I do have a preference for the scheduler because of how easy it is for me to implement custom multi-turn training workflows.

jybsuper:
> @jybsuper I would appreciate a patch for this so that the ChatCompletionScheduler can use similar config arguments. I do have a preference for the scheduler because of how easy it is for me to implement custom multi-turn training workflows.

Sounds good. I will create a PR for fix soon.

Issue #1636, opened 2025-05-22
Implement sglang async rollout and multi-turn using AsyncServerBase


chenhaiq:
There are 2 different sync LLM in current:
1. rollout.name=vllm + rollout.mode=async from  https://github.com/volcengine/verl/pull/1138
2. rollout.name=sglang_async from https://github.com/volcengine/verl/pull/1037

The implementation of vllm async mode adopts a server mode, which facilitates future extensions. For example, it can add load balancing capabilities to reduce GPU idle time caused by long-tail requests.

An AsyncSGLangServer(AsyncServerBase) can be implemented to unify the asynchronous modes of the two engines.

zhaochenyang20:
Thanks. Jin, could you take a look?

jhinpan:
Gotcha, we are currently doing Server-based support for SGLang rollout in verl. You are welcome to cooperate with us to implement the AsyncServerBase for AsyncSGLangServer. My vx id is: Jhin_P.

SwordFaith:
A sharing interface could enhance user comprehension of asynchronous multi-turn logic and simplify the process of switching between different rollout types. I'm curious about how the current design handles tool logic. How are previously considered multi-model names, partial rollouts, and the replay buffer implemented? Looking forward to hearing your thoughts!

PR #1698, opened 2025-05-26
sglang_async_rollout: Implement sglang async rollout and multi-turn u…

chenhaiq:
…sing AsyncServerBase

Implemented AsyncSglangServer similar with AsyncvLLMServer.

Tested run_qwen2-7b_seq_balance_sglang.sh with TP=1, but still has some todos:

TODO

- [ ] improve performance when TP>1. Current implementation is slow because sglang_engine.async_generate is called in sequence for each request.
- [ ] test in multi node deployment.
- [ ] add an unit test

### Checklist Before Starting

- [done] Search for similar PR(s).

### What does this PR do?

resolve issue: https://github.com/volcengine/verl/issues/1636

### High-Level Design

https://github.com/volcengine/verl/issues/1636

### Specific Changes

add AsyncSglangServer

### API

N/A

### Usage Example

### Test

### Additional Info.

- **Issue Number**: Fixes issue 1636
- **Training**: [none]
- **Inference**: [SGLang]

- [done ] Read the [Contribute Guide](https://github.com/volcengine/verl?tab=readme-ov-file#contribution-guide).
- [ done] Apply [pre-commit checks](https://github.com/volcengine/verl?tab=readme-ov-file#code-linting-and-formatting).
- [ done] Add `[BREAKING]` to the PR title if it breaks any API.
- [ done] Update the documentation about your changes in the [docs](https://github.com/volcengine/verl/tree/main/docs).
- [ done] Add CI test(s) if necessary.

### DIFF
--- a/verl/workers/fsdp_workers.py
+++ b/verl/workers/fsdp_workers.py
@@ -407,41 +407,68 @@ class ActorRolloutRefWorker(Worker):
             log_gpu_memory_usage("After building sharding manager", logger=logger)
 
         elif rollout_name == "sglang":
-            from verl.workers.rollout.sglang_rollout import SGLangRollout
+            if self.config.rollout.mode == "sync":
+                from verl.workers.rollout.sglang_rollout import SGLangRollout
 
-            # NOTE(linjunrong): Due to recent fp8 support in SGLang. Now importing any symbol relate to
-            # SGLang's model_runner would check CUDA device capability. However, due to verl's setting,
-            # the main process of ray can not find any CUDA device, which would potentially lead to:
-            # "RuntimeError: No CUDA GPUs are available".
-            # For this reason, sharding_manager.__init__ should not import FSDPSGLangShardingManager and
-            # we import it here use the abs path.
-            # check: https://github.com/sgl-project/sglang/blob/00f42707eaddfc2c0528e5b1e0094025c640b7a0/python/sglang/srt/layers/quantization/fp8_utils.py#L76
-            from verl.workers.sharding_manager.fsdp_sglang import FSDPSGLangShardingManager
+                # NOTE(linjunrong): Due to recent fp8 support in SGLang. Now importing any symbol relate to
+                # SGLang's model_runner would check CUDA device capability. However, due to verl's setting,
+                # the main process of ray can not find any CUDA device, which would potentially lead to:
+                # "RuntimeError: No CUDA GPUs are available".
+                # For this reason, sharding_manager.__init__ should not import FSDPSGLangShardingManager and
+                # we import it here use the abs path.
+                # check: https://github.com/sgl-project/sglang/blob/00f42707eaddfc2c0528e5b1e0094025c640b7a0/python/sglang/srt/layers/quantization/fp8_utils.py#L76
+                from verl.workers.sharding_manager.fsdp_sglang import FSDPSGLangShardingManager
 
-            log_gpu_memory_usage(f"Before building {rollout_name} rollout", logger=logger)
-            local_path = copy_to_local(self.config.model.path)
-            rollout = SGLangRollout(
-                actor_module=local_path,
-                config=self.config.rollout,
-                tokenizer=self.tokenizer,
-                model_hf_config=self.actor_model_config,
-                trust_remote_code=trust_remote_code,
-            )
-            log_gpu_memory_usage(f"After building {rollout_name} rollout", logger=logger)
+                log_gpu_memory_usage(f"Before building {rollout_name} rollout", logger=logger)
+                local_path = copy_to_local(self.config.model.path)
+                rollout = SGLangRollout(
+                    actor_module=local_path,
+                    config=self.config.rollout,
+                    tokenizer=self.tokenizer,
+                    model_hf_config=self.actor_model_config,
+                    trust_remote_code=trust_remote_code,
+                )
+                log_gpu_memory_usage(f"After building {rollout_name} rollout", logger=logger)
 
-            if torch.distributed.get_world_size() == 1:
-                self.config.rollout.load_format = "dummy_hf"
-            rollout_sharding_manager = FSDPSGLangShardingManager(
-                module=self.actor_module_fsdp,
-                inference_engine=rollout.inference_engine,
-                model_config=self.actor_model_config,
-                full_params="hf" in self.config.rollout.load_format,
-                device_mesh=rollout_device_mesh,
-                offload_param=self._is_offload_param,
-            )
-            log_gpu_memory_usage("After building sharding manager", logger=logger)
+                if torch.distributed.get_world_size() == 1:
+                    self.config.rollout.load_format = "dummy_hf"
+                rollout_sharding_manager = FSDPSGLangShardingManager(
+                    module=self.actor_module_fsdp,
+                    inference_engine=rollout.inference_engine,
+                    model_config=self.actor_model_config,
+                    full_params="hf" in self.config.rollout.load_format,
+                    device_mesh=rollout_device_mesh,
+                    offload_param=self._is_offload_param,
+                )
+                log_gpu_memory_usage("After building sharding manager", logger=logger)
+            elif self.config.rollout.mode == "async":
+                from verl.workers.rollout.sglang_rollout import AsyncSGLangRollout
+                from verl.workers.sharding_manager.fsdp_sglang import FSDPAsyncSGLangShardingManager
 
+                local_path = copy_to_local(self.config.model.path)
+                log_gpu_memory_usage(f"Before building {rollout_name} rollout", logger=None)
+                rollout = AsyncSGLangRollout(
+                    actor_module=local_path,
+                    config=self.config.rollout,
+                    tokenizer=self.tokenizer,
+                    model_hf_config=self.actor_model_config,
+                    trust_remote_code=trust_remote_code,
+                )
+                log_gpu_memory_usage(f"After building {rollout_name} rollout", logger=None)
+
+                if torch.distributed.get_world_size() == 1:
+                    self.config.rollout.load_format = "dummy_hf"
+                rollout_sharding_manager = FSDPAsyncSGLangShardingManager(
+                    module=self.actor_module_fsdp,
+                    inference_engine=rollout._engine,
+                    model_config=self.actor_model_config,
+                    full_params="hf" in self.config.rollout.load_format,
+                    device_mesh=rollout_device_mesh,
+                    offload_param=self._is_offload_param,
+                )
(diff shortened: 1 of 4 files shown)


tao-githup:
Why choose the server approach instead of the engine approach? Are there any advantages to the former?

chenhaiq:
> Why choose the server approach instead of the engine approach? Are there any advantages to the former?
server approach can resolve those problems better than engine:
1. in mult-turn mode, tool calling involve delays. 
2. each problems in training dataset have different length of response, which involve delays to wait for the slowest rank.

SwordFaith:
Simple Notes for 250527 2:00 p.m. -> 4:00 p.m.
1. move broadcat to other place rather than generate sequences
2. tool use api/mcp(optional) ? colocate(worker-tool) + resource(oss:// base, avoid resend hot point) issue
  - tool registry + worker separation
  - tool response has image with GUI Agent (20-50turns, each turn has tool)
  - ray aware req source, and route back, only transfer multi-modal data between processes, avoid all data go through `chat scheduler` and `tool registry`.
  - ray affinity(ray remote call has affinity scheduler) is optimized or need a lot work to do?
3. engine back press get query from queue (need optimize in chat scheduler)
4. malformed tool call fix (`parse_args` in `BaseTool` -> how to support in chat scheduler & tool registry version)

verl/workers/rollout/sglang_rollout/async_sglang_server.py
```diff
@@ -0,0 +1,71 @@
...
+from starlette.requests import Request
+from starlette.responses import JSONResponse
+
+from verl.workers.rollout.async_server import AsyncServerBase
+
+logger = logging.getLogger(__file__)
+
+
+@ray.remote(num_cpus=1)
+class AsyncSglangServer(AsyncServerBase):
```
wuxibin89, line 28: Add a test for AsyncSglangServer?

chenhaiq: 加了测试
tests/workers/rollout/test_async_sglang_server.py

verl/workers/fsdp_workers.py
```diff
@@ -407,41 +407,68 @@ def _build_rollout(self, trust_remote_code=False):
...
+                rollout_sharding_manager = FSDPAsyncSGLangShardingManager(
+                    module=self.actor_module_fsdp,
+                    inference_engine=rollout._engine,
+                    model_config=self.actor_model_config,
+                    full_params="hf" in self.config.rollout.load_format,
+                    device_mesh=rollout_device_mesh,
+                    offload_param=self._is_offload_param,
+                )
+                log_gpu_memory_usage("After building sharding manager", logger=None)
         elif rollout_name == "sglang_async":
```
SwordFaith, line 470: Unavailable after PR 1717 is merged.

verl/workers/rollout/sglang_rollout/async_sglang_server.py
```diff
@@ -0,0 +1,71 @@
...
+            # send to all works in this tp group, because sglang is SPMD
+            if current_rank >= self._dp_rank * self._tp_size and current_rank < (self._dp_rank + 1) * self._tp_size:
+                self.workers.append(ray.get_actor(**matched_actor))
+
+    async def chat_completion(self, raw_request: Request):
+        request = await raw_request.json()
+
+        output_dp_lst = []
+        for worker in self.workers:
+            output_future = worker.execute_method.remote("chat_completion", request)
```
SwordFaith, line 56: Maybe we can refer to the implementation in the following link: https://github.com/SwordFaith/verl/blob/refactor/merge_sgl_rollouts_and_bump_to_0.4.6.post4/verl/workers/rollout/vllm_rollout/vllm_async_server.py#L213. We could only retrieve results from TP0, meaning non-TP0 workers would return a special response and be skipped by default.

verl/workers/rollout/sglang_rollout/async_sglang_rollout.py
```diff
@@ -768,3 +771,84 @@ def _preprocess_prompt_to_async_rollout_requests(self, prompts: DataProto, n: in
...
+            if self._tp_rank == 0:
+                loop = asyncio.get_event_loop()
+                output = loop.run_until_complete(
+                    self._engine.async_generate(
+                        prompt=prompt_str,
+                        sampling_params=sampling_params_dict,
+                        return_logprob=True,
+                    )
+                )
+            output = broadcast_pyobj(
```
SwordFaith, line 802: It can be removed if we only need TP0 to return all results from the current TP group and assemble them in the chat scheduler.

verl/workers/rollout/sglang_rollout/async_sglang_rollout.py
```diff
@@ -768,3 +771,84 @@ def _preprocess_prompt_to_async_rollout_requests(self, prompts: DataProto, n: in
...
+            }
+        else:
+            raise ValueError(f"not supported method : {method}")
+
+        # this function is left for uniform train-inference resharding
+
+    def resume(self):
+        if not self.is_sleep:
+            return
+        self.sharding_manager.__enter__()  # pylint: disable=C2801
```
SwordFaith, line 844: Is there any better implementation way ?

verl/workers/rollout/sglang_rollout/async_sglang_rollout.py
```diff
@@ -768,3 +771,84 @@ def _preprocess_prompt_to_async_rollout_requests(self, prompts: DataProto, n: in
...
+            sampling_params_dict = {
+                "n": json_request.get("n", 1),
+                "max_new_tokens": json_request.get("max_completion_tokens", self.config.response_length),
+                "temperature": json_request.get("temperature", 1.0),
+                "top_p": json_request.get("top_p", 1.0),
+            }
+            output = None
+            if self._tp_rank == 0:
+                loop = asyncio.get_event_loop()
+                output = loop.run_until_complete(
```
SwordFaith, line 795: Would it be better to use _async_rollout_a_request?

verl/workers/rollout/sglang_rollout/async_sglang_rollout.py
```diff
@@ -768,3 +771,84 @@ def _preprocess_prompt_to_async_rollout_requests(self, prompts: DataProto, n: in
                 req_list.append(req)
 
         return req_list
+
+    def execute_method(self, method: Union[str, bytes], *args, **kwargs):
```
SwordFaith, line 775: Is it possible to implement this as an async function?

chenhaiq: > Is it possible to implement this as an async function?

sure. It will be done in https://github.com/volcengine/verl/issues/1721
It have to be sync method at this moment.

verl/workers/rollout/sglang_rollout/async_sglang_rollout.py
chenhaiq: > Would it be better to use _async_rollout_a_request?

It will be replaced soon in https://github.com/volcengine/verl/issues/1721

verl/workers/rollout/sglang_rollout/async_sglang_rollout.py
chenhaiq: > Is there any better implementation way ?

vllm part also has code like this. I will create a new issue for this.

verl/workers/rollout/sglang_rollout/async_sglang_rollout.py
chenhaiq: > It can be removed if we only need TP0 to return all results from the current TP group and assemble them in the chat scheduler.

yes

verl/workers/rollout/sglang_rollout/async_sglang_server.py
chenhaiq: > Maybe we can refer to the implementation in the following link: https://github.com/SwordFaith/verl/blob/refactor/merge_sgl_rollouts_and_bump_to_0.4.6.post4/verl/workers/rollout/vllm_rollout/vllm_async_server.py#L213. We could only retrieve results from TP0, meaning non-TP0 workers would return a special response and be skipped by default.

yes

verl/workers/fsdp_workers.py
chenhaiq: > Unavailable after PR 1717 is merged.

let's merge this PR to avoid further conflict

verl/workers/rollout/sglang_rollout/async_sglang_rollout.py
chenhaiq: > Is there any better implementation way ?

https://github.com/volcengine/verl/issues/1734

verl/workers/rollout/sglang_rollout/async_sglang_server.py
wuxibin89: After removing broadcast_pyobj, is it just to call TP0 worker.execute_method?

SwordFaith: > After removing broadcast_pyobj, is it just to call TP0 worker.execute_method?

If we have a proper registration mechanism, it would be possible.

tao-githup:
> > Why choose the server approach instead of the engine approach? Are there any advantages to the former?
> > server approach can resolve those problems better than engine:
> 
> 1. in mult-turn mode, tool calling involve delays.
> 2. each problems in training dataset have different length of response, which involve delays to wait for the slowest rank.

yeah, that's why we use async multi-turn mode. But why did we choose the server instead of  async vllm engine ? may be due to the load balance?

SwordFaith:
> > > Why choose the server approach instead of the engine approach? Are there any advantages to the former?
> > > server approach can resolve those problems better than engine:
> > 
> > 
> > 
> > 1. in mult-turn mode, tool calling involve delays.
> > 2. each problems in training dataset have different length of response, which involve delays to wait for the slowest rank.
> 
> yeah, that's why we use async multi-turn mode. But why did we choose the server instead of async vllm engine ? may be due to the load balance?

A centralized chat scheduler enables more efficient load balancing compared to pre-splitting tasks among DP groups.

HoweYoung:
> > Why choose the server approach instead of the engine approach? Are there any advantages to the former?
> > server approach can resolve those problems better than engine:
> 
> 1. in mult-turn mode, tool calling involve delays.
> 2. each problems in training dataset have different length of response, which involve delays to wait for the slowest rank.

所以说用engine的话是在外面组好batch一次性丢进去的（不会存在llm和调用工具并行的情况是吗），而sever based是一条一条的，这样调用工具和llm模型推理能够流水线并行？可以这么理解吗？

PR #1705, opened 2025-05-26
Add dstack multinode example in README

Bihan:


Bihan:
Will send a cleaner commit

PR #1712, opened 2025-05-27
[misc] feat: support logging rollout prob vs. actor probs for debugging purpose

vermouth1992:
### Checklist Before Starting

### What does this PR do?

- Support logging rollout probs vs. actor probs for debugging purpose
- Support both vllm and sglang async

### High-Level Design

### Specific Changes

> List the specific changes.

### API

### Usage Example

```python
```

### Test

### Additional Info.

### DIFF
--- a/recipe/dapo/test_dapo_7b_math.sh
+++ b/recipe/dapo/test_dapo_7b_math.sh
@@ -54,6 +54,8 @@ offload=True
 gen_tp=4
 fsdp_size=32
 
+# remember to set VLLM_ALLOW_LONG_MAX_MODEL_LEN=1 for this model
+
 python3 -m verl.trainer.main_ppo \
     data.train_files="${TRAIN_FILE}" \
     data.val_files="${TEST_FILE}" \


PeterSH6:
Do you observe that setting logprob=1 in vllm may be slower than logprob=0?

I observed this behavior in vllm v0.7. While vllm 0.6.3 did not incur too much overhead when setting logprob>0

Not sure if it's still a potential problem in vllm v0.8

verl/workers/rollout/vllm_rollout/vllm_rollout.py
```diff
@@ -229,11 +229,11 @@ def generate_sequences(self, prompts: DataProto, **kwargs) -> DataProto:
             # TODO(sgm): disable logprob when recompute_log_prob is enable
             # if n = 1: (bs, response_length) ; if n > 1: (bs * n, response_length)
             response = output[0].to(idx.device)
-            # log_probs = output[1].to(idx.device)
+            log_probs = output[1].to(idx.device)
```
PeterSH6, line 232: Do we need to check if logprob is >0 here?

vermouth1992: logprob will never be > 0?

PeterSH6: What I refer to is here: 

```python3
kwargs = dict(
            n=1,
            logprobs=0,  # can be set to 0 and let actor to recompute
            max_tokens=config.response_length,
        )
```
We may need to set logprobs > 0 to get logprob returns in vllm

PeterSH6: logprobs == 0 will return the highest logprob. So it would be fine here

GHGmc2:
Can we add an option (maybe default `False`) for it? Sometimes we may want as less as possible data to be sent/redv over ray controller on large clusters.

vermouth1992:
Sure. I guess you can draft a PR that only pass useful keys to workers to minimize communication cost.

GHGmc2:
> Sure. I guess you can draft a PR that only pass useful keys to workers to minimize communication cost.

Please help to review: https://github.com/volcengine/verl/pull/2072

Issue #1676, opened 2025-05-25
How can I set training without validation?


tzjtatata:
I want to train a model without any validation. But I found that "must validation" is hard-coding in verl. HOW CAN I AVOID IT?

vermouth1992:
You can set the `test_freq` to a huge value

tzjtatata:
I don't have val dataset. But verl force me to put one...



---Original---
From: "Chi ***@***.***&gt;
Date: Sun, May 25, 2025 16:30 PM
To: ***@***.***&gt;;
Cc: "Yuanze ***@***.******@***.***&gt;;
Subject: Re: [volcengine/verl] How can I set training without validation?(Issue #1676)


vermouth1992 left a comment (volcengine/verl#1676)
 
You can set the test_freq to a huge value

ccclyu:
simply copy some samples from the training set as the validation set. It is weird to discard the validation stage in terms of the functionality of training framework.

HareshKarnan:
training seems to hang at the step iteration that matches test_freq, makes me wonder if there's a bug causing training to hang at the validation stage

PR #1754, opened 2025-05-29
fix: support empty val dataset

shinytang6:
### Checklist Before Starting

### What does this PR do?

We have same requirements as https://github.com/volcengine/verl/issues/1676
This PR support user training with empty val dataset

### High-Level Design

### Specific Changes

> List the specific changes.

### API

### Usage Example

```python
```

### Test

### Additional Info.

### DIFF
--- a/recipe/spin/spin_trainer.py
+++ b/recipe/spin/spin_trainer.py
@@ -549,22 +549,24 @@ class RaySPINTrainer:
             collate_fn=collate_fn,
             sampler=train_sampler,
         )
-
-        val_batch_size = self.config.data.val_batch_size  # Prefer config value if set
-        if val_batch_size is None:
-            val_batch_size = len(self.val_dataset)
-
-        self.val_dataloader = StatefulDataLoader(
-            dataset=self.val_dataset,
-            batch_size=val_batch_size,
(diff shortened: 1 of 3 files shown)


shinytang6:
@vermouth1992 pls help take a look, thx

recipe/spin/spin_trainer.py
```diff
@@ -549,22 +549,24 @@ def _create_dataloader(self, train_dataset, val_dataset, collate_fn, train_sampl
...
-            num_workers=self.config.data.get("dataloader_num_workers", 8),
-            shuffle=False,
-            drop_last=False,
-            collate_fn=collate_fn,
-        )
-
         assert len(self.train_dataloader) >= 1, "Train dataloader is empty!"
-        assert len(self.val_dataloader) >= 1, "Validation dataloader is empty!"
+
+        if self.val_dataset is None:
```
vermouth1992, line 554: It seems that self.val_dataset will never be None according to the line 535?

shinytang6: `create_rl_dataset` will return None is `self.config.data.val_files` is not set

PR #1761, opened 2025-05-29
[DataProto] Supporting new operations for `DataProto`

hongpeng-guo:
### Checklist Before Starting

### What does this PR do?

Adding/ Enriching new operations on `DataProto` data class:

1. Making `DataProto` compitable with `self.batch is None`, this is useful when we are using a `DataProto` to contain non-tensor data only, i.e., images for vlm use cases;
2. `sample_level_repeat`： this function repeat the rows in DataProto multiple times in sample level;
3.  `unfold_column_chunks`: this function split along the second dim into `n_splits` folds. Useful in passing grouped tensors that doesn't want to be shuffled in dataset.

### API & Test

Please check the usage from the added unit test files: `tests/test_protocol.py`. There are three unit tests added, which are: `test_dataproto_no_batch`, `test_sample_level_repeat`, and `test_dataproto_unfold_column_chunks`.


eric-haibin-lin:
some broken megatron-sglang tests will be fixed by #1717

eric-haibin-lin (approved):

Issue #1757, opened 2025-05-29
vllm cannot load model after megatron training


KAKSIS:
After megatron training and convert to hf model, i want to infer using vllm, which meet problem when loading.
File "/python3.11/site-packages/vllm/model_executor/models/utils.py", line 250, in _load_module
   raise ValueError(msg)
ValueError: There is no module or parameter named 'decoder' in Qwen3ForCausalLM
After model_merger.py process, the name of parameters changed.

PR #1774, opened 2025-05-30
[bugfix] fix megatron model merger

ShareLer:
### Checklist Before Starting

### What does this PR do?

Fix megatron model merger.

### High-Level Design

### Specific Changes

- Fix get rank method to support just TP.
- Fix state_dict keys after convert.
- Add mla/moe convert support.

### API

### Usage Example

```python
```

### Test

Test with Qwen3-8B and Qwen2.5-7B.

### Additional Info.

- **Issue Number**: Fixes issue #1757

### DIFF
--- a/scripts/model_merger.py
+++ b/scripts/model_merger.py
@@ -475,11 +475,16 @@ class MegatronModelMerger(BaseModelMerger):
         """
         Checks if the key is a valid Megatron state key.
 
-        Now the model merger only supports keys that start with "decoder" in TransformerLayer.
+        Now the model merger only supports keys that start with "decoder/embedding/output_layer" in TransformerLayer.
+        Shall not use key starts with "model."
         """
+        if key.startswith("model."):
+            raise ValueError(f"Invalid key {key} in Megatron state_dict. Expected keys to start with 'decoder/embedding/output_layer' in TransformerLayer.")
+
         skip_checking_keys = ["embedding.word_embeddings", "output_layer"]
         for skip_key in skip_checking_keys:
             if skip_key in key:
+                print(f"skip checking key {key}")
                 return
 
         # Exclude extra state keys


ShareLer:
> @ShareLer Thanks a lot for helping us to fix this, it helps a lot~
> 
> Could you briefly point out what causes vllm inference failure? Seems also involved a lot of refactorization. Is it caused by the missing parameters transferring?

Three main reasons:
1. The name of the converted layer has not been modified in _merge_state_dicts().
`model.decoder.layers.xxx` in converted ckpt, but the actual is supposed to be `model.layers.xxx`.
This is also the reason for the failure in the mentioned issue.

2. The name of qkv in the attention layer is error after converted.
`linear_qkv` is converted to linear_q/linear_k/linear_v, but in reality it should be q_proj/k_proj/v_proj

3. The weight of the final output layer is processed incorrectly.
When is_value_model=False, the `output_layer` is ColumnParallelLinear, but the weights of different TP were not merged in _merge_state_dicts().
(When submitting for the first time, my method ignored the value_model, resulting in the failure of CI. It has just been fixed.)

eric-haibin-lin (commented):
is it possible to add a test that reproduces the issue?

ShareLer:
> is it possible to add a test that reproduces the issue?

You can reproduce this problem very simply by using the CI script (like job 'e2e_ppo_trainer_megatron-qwen3' in e2e_ppo_trainer_megatron.yml) just change command option in merger：
You need change the `test` operation in `python scripts/model_merger.py test --backend megatron` to `merge`.

There were no problems in the previous CI test because different logics were used in the `test` and `merge` options: 
First of all, they all obtained the merged weights through the `_merge_state_dicts()` method. However, it should be noted that there are some problems with state_dicts at this time (the three problems in the previous reply). 
Next, in the `merge` option, this problematic state_dicts was directly saved as the final ckpt. But these problematic layer names were corrected in the test option (remove the decoder and correct the name of the qkv) which used in CI.

ETOgaosion:
> The name of the converted layer has not been modified in _merge_state_dicts().
model.decoder.layers.xxx in converted ckpt, but the actual is supposed to be model.layers.xxx.
This is also the reason for the failure in the mentioned issue.

@ShareLer Could you add some assertions for checking whether naming prefix is valid as `model_merger`'s request?

Or whether it can have some robustness to get elements from string, like fetch the layer_index from [-3] index of split strings?

ShareLer:
> > The name of the converted layer has not been modified in _merge_state_dicts().
> > model.decoder.layers.xxx in converted ckpt, but the actual is supposed to be model.layers.xxx.
> > This is also the reason for the failure in the mentioned issue.
> 
> @ShareLer Could you add some assertion for checking whether naming prefix is valid as `model_merger`'s request?
> 
> Or whether it can have some robustness to get elements from string, like fetch the layer_index from [-3] index of split strings?

Sorry, I don't quite understand what you mean.
Do you mean that we should perform some checks on the keys in merged_state_dict before saving the merged ckpt to ensure its validity?

scripts/model_merger.py
```diff
@@ -444,28 +485,28 @@ def _merge_state_dicts(self, model_state_dict_lst: list[list[dict]], tp_size: in
                         print("skip lm_head and reward_head loading because of tie_word_embeddings")
                         continue
```
ETOgaosion, line 509: Maybe at here we do some check on the mcore's key to check whether it's valid for `_replace_name` to work?

ShareLer: Yes, we can add an assert to the result of `_replace_name` to ensure that it is not None, because we define the layer transformation relationship in self.params_mapping, and normally the result should not appear None.
Do you think this is feasible?

ETOgaosion: Maybe I didn't make myself clear(qaq). I can help to add some checking and warnings~

ShareLer:
@ETOgaosion Hi, I have added additional judgments for embedding layer and output_layer on the basis of your code.

ETOgaosion:
@ShareLer Thanks for helping, I was about to test on my machine~

PR #1775, opened 2025-05-30
fix import issue for mcore package

jinqinn:
fix import issue for mcore package in `patch_v012.py`


jinqinn:
The CI test failed. it seems like it's not related to my commit. plz take a look at it and let me know if i need to make any other updates. Thank you.

vermouth1992 (approved):

Issue #891, opened 2025-04-03
[Potential Bug] KL compute in `low_var_kl (Cause KL NaN and !!!!!!!!!!!!!!! output)


yushuiwx:
Hi, 

A few days ago, I reported this issue in the discussion [751](https://github.com/volcengine/verl/issues/751) and now there are some updates.

similar issue:
- https://github.com/volcengine/verl/issues/747
- https://github.com/volcengine/verl/issues/721
-  .......

After careful analysis, I found that the issue lies in the calculation of `low_var_kl`. In the problematic step, the loss has normal values (without NaN or Inf), but during backpropagation, the `exp` operation in `ratio = torch.exp(kl)` within `low_var_kl` （https://github.com/volcengine/verl/blob/8cae42dc29736d0802ded43c5ecf67a809d56bd8/verl/trainer/ppo/core_algos.py#L386）causes extreme values, leading to NaN in `kl`. My temporary solution is to use `torch.clamp(kl, min=-5, max=5)` to constrain `kl` before computing `ratio`. **This approach works, but I feel uncertain about it. Could you kindly help confirm whether this is an appropriate solution?** 

 _Originally posted by @yushuiwx in [#751](https://github.com/volcengine/verl/issues/751#issuecomment-2774700868)_

vermouth1992:
The final clamp does not take effect?

great-luao:
For me I always have the similar !!! problem after the first training step. I found that my training grad_norm is 0, which caused the training to crash, I solve this issue refer to https://github.com/volcengine/verl/issues/405. Hope this will help somebody lol.
Besides, this leads to another issue that the assertion added actually won't work since we don't use "ppo_micro_batch_size" anymore. The assert code will be skipped and no one can notice.

![Image](https://github.com/user-attachments/assets/ff7850b6-1040-450e-bdbf-54b38fd77092)

chuangzhidan:
> For me I always have the similar !!! problem after the first training step. I found that my training grad_norm is 0, which caused the training to crash, I solve this issue refer to [#405](https://github.com/volcengine/verl/issues/405). Hope this will help somebody lol. Besides, this leads to another issue that the assertion added actually won't work since we don't use "ppo_micro_batch_size" anymore. The assert code will be skipped and no one can notice.
> 
> ![Image](https://github.com/user-attachments/assets/ff7850b6-1040-450e-bdbf-54b38fd77092)

how did u fix it ?    
2gpus,my setting is like this,
[...]

Issue #721, opened 2025-03-23
output is  !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!


BeSkyer:
While training with GRPO, I encountered an issue where the model output abruptly degrades into a sequence like !!!!!!!!!!!!!!!. Just before this degradation, the kl_loss becomes NaN. Upon further investigation, I found that old_log_prob is turning into NaN, which appears to be the root cause.

Additionally, the occurrence of this issue is inconsistent across training steps, indicating that it likely isn’t related to the dataset (I'm using GSM8K). I’ve also noticed similar reports from others using GRPO or REINFORCE++.

Are there any plans to address this instability in future updates?

step:59 - global_seqlen/min:16268.000 - global_seqlen/max:20607.000 - global_seqlen/minmax_diff:4339.000 - global_seqlen/balanced_min:17865.000 - global_seqlen/balanced_max:19434.000 - global_seqlen/mean:18257.750 - actor/kl_loss:nan - actor/kl_coef:0.001 - actor/entropy_loss:nan - actor/pg_loss:nan - actor/pg_clipfrac:0.002 - actor/ppo_kl:nan - actor/grad_norm:nan - perf/mfu/actor:0.042 - perf/max_memory_allocated_gb:27.050 - perf/max_memory_reserved_gb:28.637 - perf/cpu_memory_used_gb:43.031 -

Enoch202:
I encountered the same problem when training Qwen with GRPO or reinforce++. 
The crash during training was very sudden: the previous step was normal, but in the next step, the loss became NaN.

vermouth1992:
Could you provide a script that can reproduce?

Reason-Wang:
Same problem here. Has anyone investigated this problem? It happened on my side on math problems with the dataset from [Open-Reasoner-Zero](https://github.com/Open-Reasoner-Zero/Open-Reasoner-Zero). It can happen on any step during training.

yushuiwx:
Hi, same problem, @BeSkyer May I ask how you managed to solve it, please?

BeSkyer:
> Hi, same problem, [@BeSkyer](https://github.com/BeSkyer) May I ask how you managed to solve it, please?
Sorry for the late reply. I spent a long time checking the code but still couldn't solve the issue. In the end, I set KL_loss to False, and that prevented the NaN problem from occurring.

[...]

PR #1779, opened 2025-05-30
Stabilize loss calculations by clamping KL divergence values

syo093c:
### Summary

This PR improves the numerical stability of PPO training in `verl` by clamping KL divergence-related values in the loss calculations. Specifically:

- In `compute_policy_loss`, the `negative_approx_kl` value is now clamped to the range \([-10, 10]\) before exponentiation and further use.
- In `kl_penalty` (for the `"low_var_kl"` mode), the KL value is also clamped to \([-10, 10]\) before further calculations.

### Motivation

During PPO training, extreme log-probability differences can occasionally occur, leading to numerical instabilities or exploding/vanishing gradients. By clamping these values, we ensure more stable and reliable training dynamics, especially in edge cases.

### Changes

- Added `torch.clamp` to `negative_approx_kl` in `compute_policy_loss`.
- Added `torch.clamp` to KL values in `kl_penalty` for `"low_var_kl"` mode.
- Both are clamped to the range \([-10, 10]\).

### Related Issues

#891 #721

### DIFF
--- a/verl/trainer/ppo/core_algos.py
+++ b/verl/trainer/ppo/core_algos.py
@@ -447,7 +447,7 @@ def compute_policy_loss(
 
     negative_approx_kl = log_prob - old_log_prob
     # Clamp negative_approx_kl for stability
-    negative_approx_kl = torch.clamp(negative_approx_kl, min=-10.0, max=10.0)
+    negative_approx_kl = torch.clamp(negative_approx_kl, min=-20.0, max=20.0)
     ratio = torch.exp(negative_approx_kl)
     ppo_kl = verl_F.masked_mean(-negative_approx_kl, response_mask)
 
@@ -544,7 +544,7 @@ def kl_penalty(logprob: torch.FloatTensor, ref_logprob: torch.FloatTensor, kl_pe
     if kl_penalty == "low_var_kl":
         kl = ref_logprob - logprob
         # For numerical stability
-        kl = torch.clamp(kl, min=-10, max=10)
+        kl = torch.clamp(kl, min=-20, max=20)
         ratio = torch.exp(kl)
         kld = (ratio - kl - 1).contiguous()
         return torch.clamp(kld, min=-10, max=10)


verl/trainer/ppo/core_algos.py
```diff
@@ -446,6 +446,8 @@ def compute_policy_loss(
     assert clip_ratio_c > 1.0, "The lower bound of the clip_ratio_c for dual-clip PPO should be greater than 1.0," + f" but get the value: {clip_ratio_c}."
 
     negative_approx_kl = log_prob - old_log_prob
+    # Clamp negative_approx_kl for stability
+    negative_approx_kl = torch.clamp(negative_approx_kl, min=-10.0, max=10.0)
```
vermouth1992: Is [-10, 10] too strict for log probs?

vermouth1992: Maybe we want to choose the maximum value such as exp is not nan?

syo093c: Hi, thank you for your comment!

Float32 max value is about ~3.4e38.
So, we can use `80` as limit (exp(80)=5.5406e+34).
But, choose `80` will still have risk to overflow pg_loss`= advantages*ratio`.

Consideration of numerical stability and exploration space,
I think maybe at most `20` or `30` is enough.

```
exp(20)=4.8517e+08
exp(30)=1.0686e+13
```

vermouth1992: Could you make the bound to be [-20, 20] and let's merge this

syo093c: Thank you for your advice!
I have set the bound to [-20,20]!

syo093c:
@vermouth1992 

Hi, apologies for reaching out during your busy schedule. I’ve adjusted the range to [-20, 20]. Could we proceed with merging this PR? I’d really appreciate it!

I also noticed that some tests were canceled. Is there anything I can assist with to help resolve this? Thank you so much!

hiyouga:
@vermouth1992 I think we can merge it, WDYT?

vermouth1992:
> @vermouth1992 I think we can merge it, WDYT?

I didn't have a time to verify it. I guess it's better to find a case that will actually cause Nan and verify this fix. In my training experiences, most nan occurs due to machine problems...

vermouth1992:
> > @vermouth1992 I think we can merge it, WDYT?
> 
> I didn't have a time to verify it. I guess it's better to find a case that will actually cause Nan and verify this fix. In my training experiences, most nan occurs due to machine problems...

Oh, I saw a case from EasyR1. Then, we can merge it!

vermouth1992 (approved):

PR #1794, opened 2025-06-01
[bugfix] fix select_idxs function in DataProto

Moyu-42:
### Checklist Before Starting

### What does this PR do?

Fix the batch_size type error when using DataProto.select_idxs, which originally causes the TypeError when using DataProto.chunk

### High-Level Design

### Specific Changes

Fix the batch_size type of select_idxs func.

### API

### Usage Example

```python
import torch
from verl import DataProto
import numpy as np

data = {"random_array": torch.randn(4, 2)}
batch = DataProto.from_dict(data)
valid_mask = np.array([True, False, True, False])

batch_select_idxs = batch[valid_mask]

batch.chunk(2) # correct

batch_select_idxs.chunk(2) # incorrect, raising TypeError

# with tensordict version == 0.6.2
# Traceback (most recent call last):
#   File "<stdin>", line 1, in <module>
#   File "./verl/verl/protocol.py", line 667, in chunk
#     batch_lst = self.batch.chunk(chunks=chunks, dim=0)
#   File "/opt/conda/envs/verl/lib/python3.10/site-packages/tensordict/base.py", line 2134, in chunk
#     return self.split(split_size, dim=dim)
#   File "/opt/conda/envs/verl/lib/python3.10/site-packages/tensordict/_td.py", line 1715, in split
#     raise TypeError(WRONG_TYPE)
# TypeError: split(): argument 'split_size' must be int or list of ints
```

### Test

### Additional Info.

- **Issue Number**: N/A.

### DIFF
--- a/tests/test_protocol.py
+++ b/tests/test_protocol.py
@@ -18,7 +18,6 @@ import numpy as np
 import pytest
 import torch
 from tensordict import TensorDict
-
 from verl import DataProto
 from verl.protocol import union_numpy_dict, union_tensor_dict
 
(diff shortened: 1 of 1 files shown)


eric-haibin-lin (commented):
thanks! could u add a unit test that cover this case to https://github.com/volcengine/verl/blob/main/tests/test_protocol.py ?

Moyu42:
> thanks! could u add a unit test that cover this case to https://github.com/volcengine/verl/blob/main/tests/test_protocol.py ?

I have added a unit test named `test_dataproto_chunk_after_index()`, which checks whether `DataProto.batch.batch_size` is of type `int`, as this was the root cause of the bug.

vermouth1992 (approved):

PR #1819, opened 2025-06-03
fix: train config add `model_dtype: bfloat16` to config model_dtype

weedge:
### Checklist Before Starting

### What does this PR do?

>  train config add `model_dtype: bfloat16` to config model_dtype , e.g.: `critic.model.fsdp_config.model_dtype=bfloat16`

### High-Level Design

### Specific Changes

> List the specific changes.

### API

### Usage Example

```python
```

### Test

### Additional Info.


ETOgaosion:
@weedge Thanks for contribution, we may actively add this field in configuration.

However, directly using `bfloat16` may causing divergence with current implementation:

https://github.com/volcengine/verl/blob/fba8f3463a4c1c6dff0f63a704501affa04f68d6/verl/workers/fsdp_workers.py#L198-L207

Could you check this further?

verl/trainer/config/ppo_trainer.yaml
```diff
@@ -73,6 +73,7 @@ actor_rollout_ref:
       total_training_steps: -1  # must be override by program
       weight_decay: 0.01
     fsdp_config:
+      model_dtype: bfloat16
```
hiyouga, line 76: We should not set `model_dtype` to BF16 in PPO training because it sets the optimizer states to the same dtype, but PPO requires FP32 optimizer states.
https://github.com/volcengine/verl/blob/2fe47f71ab16eec06a5fcc97f10af40e2ad36752/verl/workers/fsdp_workers.py#L205

PR #1820, opened 2025-06-03
[perf] feat: Add verl profiling support from Nvidia Nsight System

davidmlw:
Add verl profiling support from Nvidia Nsight System

### Checklist Before Starting

### What does this PR do?

Add verl profiling support from Nvidia Nsight System

### High-Level Design

This PR add config fileds to trigger Nsight profiling. If `trainer.profile_steps` is set, Nsight system will be triggered to profiling the corresponding steps. In each task role, other config fields control also control the profiling details.

The profiling tasks include the single_controller process and the worker process. Single_controller process uses the re-designed `marked_timer` to record each task range in NVTX. 

The worker processes dumps the GPU execution details. Since veRL has hybrid-engine mode and supports split mode, there are two profiling modes, discrete or not. Discrete mode means each task will generate a dedicate database; otherwise a whole giant database will be generated. Nsight system supports to import and align multiple databases automatically.

### Specific Changes

`verl.utils.debug.profile` add general profling interface and `verl.utils.debug.nvtx_profile` implements the interface.

### API

`verl.utils.debug.performance._timer` has been changed to `simple_timer`, and `marked_timer` is added to support profiler range marker.

`verl.utils.debug.profile` wrappers the basic profiler interfaces, including mark_*_range, mark_annotate, ProfilerConfig, WorkerProfiler, and WorkerProfilerExtension. `verl.utils.debug.nvtx_profile` implements the interfaces when nvtx is available.

### Usage Example

Two examples are added in `/examples/ppo_trainer/run_deepseek_math_gsm8k_megatron_nsys.sh`  `/examples/ppo_trainer/run_qwen2-7b_rm_seq_balance_nsys.sh`

### Test

There should be no functional changes and performance changes.

### Additional Info.

- **Training**: both FSDP, Megatron will be affected.
- **Inference**: both vLLM, SGLang will be affected.

- [X] Add CI test(s) if necessary.


verl/trainer/ppo/ray_trainer.py
```diff
@@ -993,19 +1012,23 @@ def fit(self):
                     batch_keys=batch_keys_to_pop,
                     non_tensor_batch_keys=non_tensor_batch_keys_to_pop,
                 )
+                nvtx.end_range(nvtx_batch)
 
                 is_last_step = self.global_steps >= self.total_training_steps
 
                 with _timer("step", timing_raw):
                     # generate a batch
+                    nvtx_gen = nvtx.start_range(message="gen", color="green")
```
eric-haibin-lin: what about adding the nvtx code to _timer instead?

verl/trainer/config/ppo_megatron_trainer.yaml
```diff
@@ -41,6 +41,9 @@ actor_rollout_ref:
     trust_remote_code: False
   actor:
     strategy: megatron  # This is for backward-compatibility
+    profile_ranks: null # [0,1]
+    profile_ranks_all: False
+    profile_discrete: False
```
eric-haibin-lin: what does profile_discrete mean?

recipe/dapo/dapo_ray_trainer.py
```diff
@@ -34,13 +34,15 @@
     reduce_metrics,
 )
 from verl.trainer.ppo.ray_trainer import AdvantageEstimator, RayPPOTrainer, _timer, apply_kl_penalty, compute_advantage, compute_response_mask
+import nvtx
```
eric-haibin-lin: nvtx should be an optional dependency.

eric-haibin-lin: can we wrap nvtx.annotate as a function under verl.utils.debug (e.g. @verl.utils.debug.tx_annotate)
it should perform a no-op if nvtx is not present

verl/trainer/config/ppo_megatron_trainer.yaml
davidmlw: In discrete=False, each process, each step, will generate one `worker_process_<PID>.<n>.nsys-rep`.
But since verl combines mulitple engines and a single step might be huge. In discrete=True, each process, each step, each task (generate, update_actor, ref_log, ...) will generate one `worker_process_<PIC>.<n>.nsys-rep`.

verl/workers/fsdp_workers.py
```diff
@@ -592,6 +609,10 @@ def init_model(self):
 
     @register(dispatch_mode=Dispatch.DP_COMPUTE_PROTO)
     def update_actor(self, data: DataProto):
+        if self.profile:
```
ETOgaosion: To reduce the redundancy codes, is it better to use a decorator defined in this class or some other classes in utils like:

```py
    def profile(foo):
        def nvtx_profile( self ) :
            if self.profile:
                # ...
```

```py
@register(dispatch_mode=Dispatch.DP_COMPUTE_PROTO)
@profile(message="update_actor", color="red")
```

davidmlw: done, and indeed much better idea.

recipe/dapo/dapo_ray_trainer.py
davidmlw: done, and I create backup empty_annotations for nvtx backup.

verl/trainer/ppo/ray_trainer.py
davidmlw: done. Indeed a better idea.

verl/utils/debug/empty_annotations.py
```diff
@@ -0,0 +1,15 @@
+
+
+import nvtx
```
stevewx: We will use empty_annotations when nvtx is not available, should not try to import it here.

verl/trainer/main_ppo.py
```diff
@@ -18,6 +18,9 @@
 
 import hydra
 import ray
+import socket
+import nvtx
```
stevewx: nit: nvtx import not needed here

stevewx:
Verified megatron rloo training with Nsight profiling enabled and was able to get detailed profiling data. This is very helpful, thanks for the change!

verl/utils/debug/empty_annotations.py
davidmlw: Right. it was a typo
Done

verl/trainer/main_ppo.py
davidmlw: done

ETOgaosion:
For the docker image, next version will install the requested nsys in the base image~

verl/utils/debug/empty_annotations.py
```diff
@@ -0,0 +1,14 @@
+
```
ETOgaosion, line 14: New files shall include bytedance and NVIDIA's License. Can copy from other files.

davidmlw: done. Nice reminder.

verl/trainer/config/ppo_trainer.yaml
```diff
@@ -96,7 +96,7 @@ actor_rollout_ref:
 
   # common configs for the model
   model:
-  
+
```
ETOgaosion, line 99: Could you remove this empty spaces, remove this needless changes?

verl/utils/debug/performance.py
```diff
@@ -105,7 +110,7 @@ def log_print(ctn: Any):
 
 
 @contextmanager
-def _timer(name: str, timing_raw: Dict[str, float]):
+def _timer(name: str, timing_raw: Dict[str, float], color: str = None):
```
ETOgaosion: `color` seems unclear, could you consider to use a `marked_timer` to package `_timer`?

```py
def _timer(name: str, timing_raw: Dict[str, float]):
    # _timer is a inner function

@contextmanager
def simple_timer(name: str, timing_raw: Dict[str, float]):
    # ...

@contextmanager
def marked_timer(name: str, timing_raw: Dict[str, float], color: str = None):
    # ...
```

verl/trainer/config/ppo_trainer.yaml
davidmlw: My editor removes all tailing spaces for every line when saving files. I also notice these empty updates, it was due to the empty spaces were deleted.

verl/utils/debug/performance.py
davidmlw: Done. There are many `_timer` user code.
In the single controller code, I use `marked_timer` in ppo, dapo, `simple_timer` in sppo, prime. `recipe/spin/spin_trainer.py` defines its own `_timer`, and I didn't touch it. Do you think I need use `marked_timer` in sppo and prime?

I use `simple_timer` in all worker code.

ETOgaosion: Maybe sometimes `simple_timer` is OK~

verl/trainer/config/ppo_trainer.yaml
ETOgaosion: It's OK, we may remove these empty spaces.

recipe/dapo/main_dapo.py
```diff
@@ -37,7 +40,17 @@ def run_ppo(config) -> None:
             num_cpus=config.ray_init.num_cpus,
         )
 
-    runner = TaskRunner.remote()
+    if OmegaConf.select(config.trainer, "profile_steps") is not None and len(OmegaConf.select(config.trainer, "profile_steps")) > 0:
+        runner = TaskRunner.options(runtime_env={"nsight": {
+            "t": "cuda,nvtx,cublas,cublas-verbose,cusparse,cusparse-verbose,cudnn,opengl,opengl-annotations,openacc,openmp,osrt,mpi,nvvideo,vulkan,vulkan-annotations,oshmem,ucx",
+            "cuda-memory-usage": "true",
+            "cuda-graph-trace": "graph",
+            #"capture-range": "nvtx",
```
eric-haibin-lin: please remove unused code

eric-haibin-lin:
this PR breaks the sanity check. for some reason it did throw an error 
image

recipe/dapo/main_dapo.py
eric-haibin-lin: can we move these env var to somewhere else? this expose too much detail to ordinary scientists. what about puting them in the launch script?

verl/utils/debug/__init__.py
```diff
@@ -13,5 +13,11 @@
...
 from .performance import GPUMemoryLogger, log_gpu_memory_usage, log_print
+from ..import_utils import is_nvtx_available
 
-__all__ = ["GPUMemoryLogger", "log_gpu_memory_usage"]
+if is_nvtx_available():
+    from .nvtx_annotations import mark_start_range, mark_end_range, mark_annotate
+else:
+    from .empty_annotations import mark_start_range, mark_end_range, mark_annotate
+
+__all__ = ["GPUMemoryLogger", "log_gpu_memory_usage", "mark_start_range", "mark_end_range", "mark_annotate"]
```
eric-haibin-lin: for newly added APIs in the namespace, doc string is required to explain what the function does, the argument and return types

verl/utils/debug/performance.py
```diff
@@ -118,12 +136,26 @@ def _timer(name: str, timing_raw: Dict[str, float]):
...
+    accumulates the timing information, and adds NVTX markers for profiling.
 
+    Args:
+        name (str): The name/identifier for this timing measurement.
+        timing_raw (Dict[str, float]): Dictionary to store timing information.
+        color (Optional[str]): Color for the NVTX marker. Defaults to None.
+
+    Yields:
+        None: This is a context manager that yields control back to the code block.
+    """
```
eric-haibin-lin, line 155: can we import from verl.xxx here instead of checking `is_nvtx_available` at the beginning of the file

verl/single_controller/base/worker.py
```diff
@@ -135,7 +138,7 @@ def env_keys(cls):
         """The keys of the environment variables that are used to configure the Worker."""
         return ["WORLD_SIZE", "RANK", "LOCAL_WORLD_SIZE", "LOCAL_RANK", "MASTER_ADDR", "MASTER_PORT", "CUDA_VISIBLE_DEVICES"]
 
-    def __init__(self, cuda_visible_devices=None) -> None:
+    def __init__(self, cuda_visible_devices=None, profile_discrete=False, profile_ranks=None, profile_ranks_all=False) -> None:
         """Initialize the worker with environment settings and device configuration.
```
eric-haibin-lin: you need to add docs for new arguments added

recipe/dapo/main_dapo.py
davidmlw: Just removed the unused lines.
Since ray manages and starts the processes, the trick I currently find is to add the env var in the Actor `options` before start it. 
One of the reasons that the community didn't run Nsight partially because it cannot be trigger from launch script like before.

verl/utils/debug/performance.py
davidmlw: I tried `from verl.utils.debug import mark_start_range, mark_end_range`, since in the `__init__.py` has the `if.. else..` logics. But this will meet an error, because `performance.py` is part of `debug` module and it will be circle dependent. So I copied the `if.. else..` logics here.

verl/single_controller/base/worker.py
davidmlw: Done

verl/utils/debug/__init__.py
davidmlw: done, I only added doc string in nvtx_annotation.py not in empty_annotation.py for less redundant information.

docs/perf/nsight_profiling.rst
```diff
@@ -0,0 +1,122 @@
+Nsight Profiling in verl
```
eric-haibin-lin: please add it to index.rst

recipe/dapo/main_dapo.py
davidmlw: I thought of an idea, we may allocate a config group for Nsight and load the configs here. What do you think?
It may need two nsight config group, one for the single controller, and one for the workers.

davidmlw: I have implemented this idea. 
Two config group, controller_nsight_options and worker_nsight_options are in configs now.

verl/utils/debug/performance.py
eric-haibin-lin: no, i mean import it inside `def marked_timer`. is there still circular dependency?

verl/workers/fsdp_workers.py
```diff
@@ -1042,6 +1071,7 @@ def init_model(self):
         )
 
     @register(dispatch_mode=Dispatch.DP_COMPUTE_PROTO)
+    @Worker.profile_annotate(color="cyan")
```
eric-haibin-lin: profile_annotate should be a general annotator instead of under `Worker` class

verl/single_controller/base/worker.py
```diff
@@ -295,3 +309,57 @@ def execute_func_rank_zero(self, func, *args, **kwargs):
         """
         result = func(*args, **kwargs)
         return result
+
```
eric-haibin-lin: Profiler should be a class that is unit-testable. It should be initialized with something like: 
```
@dataclass
ProfilerConfig:
    enabled # true/false
    ranks: Optional[list[int]]
    all_ranks # true/false 
    discrete # true/false
```
the config can be initalized via sth like
```
# Convert nested OmegaConf to dataclass
prof_conf = ProfilerConfig(**OmegaConf.to_object(conf.actor.profile))
```

There should be a unit test that create such as profiler object and test if it works. 

dp_actor, dp_critic, megatron_actor, megatron_critic holds the profiler instance, and calls it in each related function.

eric-haibin-lin: I do not think it's necessary to change the worker base class. the profiler can get rank from `torch.distributed.get_rank()` since we always refer to the global rank for profiler. There should be no change to trainer either (we want to keep trainer with only minimal code)

eric-haibin-lin (changes requested):

verl/workers/fsdp_workers.py
davidmlw: Partially done.
I move most of the profile logic to verl.utils.debug.profile.[ProfilerConfig, NsightSystemProfiler]. I still put the annotate function in a class NsightSystemProfiler since it is specific to annotate a class's member function which has a member profiler. The Worker class is much clean now.
For your consideration.

verl/single_controller/base/worker.py
davidmlw: A better idea indeed. I add a data class `ProfilerConfig` and use it to initiate a NsightSystemProfiler inside a worker.
For your consideration.

verl/utils/debug/profile.py
```diff
@@ -91,3 +94,99 @@ def stop_trace(self):
...
+class NsightSystemsProfiler:
+    """
+    Nsight system profiler. Installed in a worker to control the Nsight system profiler.
+    """
+
+    def __init__(self, rank: int, config: Optional[ProfilerConfig] = None):
+        config = config or ProfilerConfig()
+        self.this_step: bool = False
+        self.discrete: bool = config.discrete
+        self.this_rank: bool = rank in (config.ranks or []) or config.all_ranks
```
eric-haibin-lin: For nsight the profiler object should have the counter itself Map["function_name"-> step_count]. What about adding the same arguments from torch profiler (warmup, wait, active). Since the state is kept in the profiler, you can remove the start/stop calls from the trainer

eric-haibin-lin: this way the annotated functions can always call start/stop before and after the function call based on the step count maintained by itself. and you can remove the changes in trainer

verl/single_controller/base/worker.py
eric-haibin-lin: instead of creating `NsightSystemsProfiler` directly, could you have a profiler factory that creates profiler based on backend. For now backend only support 'nsys', and we may add 'torch' backend to directly use torch profiler in the future.

eric-haibin-lin: please remove the changes in `base.Worker` as verl wants to keep the functionality of base worker minimal. we can keep the profiler object in megatron and fsdp worker directly

yushengsu-thu (commented):
@davidmlw 
This is a great feature! I just have a small suggestion — would it be possible to add a condition 
```
if torch.cuda.is_available():
    device_name = torch.cuda.get_device_name()
    if 'NVIDIA' in device_name.upper():
        xxxxx
    else:
        xxxxx
```
so that the profiling code only runs when the current GPU is from NVIDIA?
Right now, many different GPUs support VERL, and since this feature seems to be NVIDIA-specific, it might cause issues if users try to run VERL on other devices.

recipe/dapo/dapo_ray_trainer.py
```diff
@@ -305,6 +316,15 @@ def fit(self):
                 # TODO: make a canonical logger that supports various backend
                 logger.log(data=metrics, step=self.global_steps)
 
+                if do_profile:
```
zw0610, line 319: May I raise a naive question? What if the experiment failed suddenly without executing L319-L326?

verl/utils/debug/profile.py
davidmlw: I would like to support two modes, whole step mode (discrete=false) and task discrete mode. 
In whole step mode, we need a scheme for the controller to inform the workers the training step start and stop.
In discrete mode, the annotated task function can maintain its own step counter, but there are chances that the mulitple counters not are synchronized and increase complexity. For example, we saw in DAPO, the generation may run more times that actor_update due to generation fails to collect enough qualified sequences. It is better to get a synchronized singal of training step from the verl framework as an infrastructure.

eric-haibin-lin: you can track & config them separately with `Map["function_name"-> step_count]`

recipe/dapo/dapo_ray_trainer.py
davidmlw: It is better to let the target program exit gracely for Nsight to dump data from Memory to Disk. I do see that the dumped *.nsys-rep lack detailed information when the target program crashed.

We can use `capture-range-end` to control Nsight Systems' behavior, so if it is repeatable, we can capture more data before it crashes.. https://docs.nvidia.com/nsight-systems/UserGuide/index.html

davidmlw:
> @davidmlw This is a great feature! I just have a small suggestion — would it be possible to add a condition
> 
> ```
> if torch.cuda.is_available():
>     device_name = torch.cuda.get_device_name()
>     if 'NVIDIA' in device_name.upper():
>         xxxxx
>     else:
>         xxxxx
> ```
> 
> so that the profiling code only runs when the current GPU is from NVIDIA? Right now, many different GPUs support VERL, and since this feature seems to be NVIDIA-specific, it might cause issues if users try to run verl on other devices.

Now I think I solved this problem more clear in utils.debug. see https://github.com/davidmlw/verl-nsys/blob/bench/verl/utils/debug/__init__.py#L19.

Outside `verl.utils.debug`, the user code only need to import `verl.utils.debug.[WorkerProfiler, mark_annotate, mark_end_range, mark_start_range, marked_timer]`, and it will route to the right implementation. Other platforms only need to implement the 5 interface functions/classes.

yushengsu-thu:
> Outside `verl.utils.debug`, the user code only need to import `verl.utils.debug.[WorkerProfiler, mark_annotate, mark_end_range, mark_start_range, marked_timer]`, and it will route to the right implementation. Other platforms only need to implement the 5 interface functions/classes.

yushengsu-thu:
> > @davidmlw This is a great feature! I just have a small suggestion — would it be possible to add a condition
> > ```
> > if torch.cuda.is_available():
> >     device_name = torch.cuda.get_device_name()
> >     if 'NVIDIA' in device_name.upper():
> >         xxxxx
> >     else:
> >         xxxxx
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > so that the profiling code only runs when the current GPU is from NVIDIA? Right now, many different GPUs support VERL, and since this feature seems to be NVIDIA-specific, it might cause issues if users try to run verl on other devices.
> 
> Now I think I solved this problem more clear in utils.debug. see https://github.com/davidmlw/verl-nsys/blob/bench/verl/utils/debug/__init__.py#L19.
> 
> Outside `verl.utils.debug`, the user code only need to import `verl.utils.debug.[WorkerProfiler, mark_annotate, mark_end_range, mark_start_range, marked_timer]`, and it will route to the right implementation. Other platforms only need to implement the 5 interface functions/classes.

thanks. See this. I get it.

docs/perf/nsight_profiling.rst
```diff
@@ -0,0 +1,122 @@
...
+``trainer.profile_steps``
+    List of step numbers at which profiling should be performed. For example: [1, 2, 5] will profile steps 1, 2, and 5. And ``null`` means no profiling.
+
+verl has one single controller process and multiple worker processes. Both controller and worker processes can be profiled. Since the controller process can be executed in any nodes in the cluster, there is message printed in the logging to indicate the controller process node hostname and process id.
+
+2. Profiling rank control
+----------------
+
+By default, verl profiles the whole training process in a single `` worker_process_<PID>.<step>.nsys-rep`` file for each process rank. If you want to profile only part of the process ranks, you can set the ``profile_ranks`` parameter for each component. ``null`` and ``[]`` means no ranks, and ``[0,1]`` means profile ranks 0 and 1. If you want to profile all ranks, you can set the ``profile_ranks_all`` parameter for each component.
+
```
eric-haibin-lin: FYI the doc website supports markdown as well. I think displaying them with bullet points makes more sense? 
image

davidmlw: done. I use markdown now.
The documentation also needs a rewrite to reflect the new design.

davidmlw: I also update the PR description.

PR #1826, opened 2025-06-03
[Feat]: Implement partial rollout

stargazerZJ:
# partial rollout PR draft

### Checklist Before Starting

### What does this PR do?

Supporting the partial rollout feature: Set vllm's maximum output length to an integer fraction of `config.response_length`. If one model response is not completed in the current round, its generation can continue in the next iteration with updated weights. This trick unlocked significant rollout time reduction without sacrifacing model performance.

partial rollout with 4K max response length, compared to baseline:
image
image
image
Note on the 3rd plot: We didn't enable testing on 2GPU baseline runs because they were already so slowly. We will update the 4GPU baseline runs after it finishes in a few hours.

### High-Level Design

image

### Specific Changes

- Add buffer pool and filter logic in the `fit` function
- Concat partial responses to prompt inputs in the `vLLMRollout` class
- Add `split` method to `DataProto`
- Add configuration entry

### API

There are mainly two areas where compatibility is potentially tricky:
1.  In the original architecture, the main loop sends only one copy of each prompt to vllm, and vllm uses the `SamplingParams.n` parameter to generate multiple responses before returning them. In our implementation, we must duplicate the prompts in the main loop and send multiple copies to vllm.
2.  The `fit` function in verl takes a batch from the dataloader in each iteration and continuously adds keys to it. In our implementation, there's a step where the `partial_batch` (partially generated by vllm) needs to be merged with the initial `batch` taken from the dataloader.

### Usage Example

```python
    algorithm.partial_rollout_max_split=2
```

### Test

Setup: qwen3 0.6b base; MATH dataset; batch size 1024; rollout n = 5; max response length 4K; 2 H100 GPUs; Megatron trainer

We run:
- Partial rollout 4K 2 split
- Partial rollout 4K 4 split
- Baseline 4K
- Baseline 2K: max response length reduced to 2k, for reference
- Baseline 1K 
- Baseline 4K 4GPU

Results, partial rollout 4K 2 split compared to baseline 4K:
* Overall experiment speed (16h vs 48h) and generation speed (estimated at 20%) significantly improved; Both are even faster than `baseline 2K` and `baseline 4K 4GPU`.
* Reward curves are basically consistent after around 15 steps (out of 105 steps in total). (No discernible difference among baseline 4K, 2K, 1K, and partial 4K, etc.)
* In the early stages of training, the reward increase of `partial 4K 2 split` lags behind the baseline by about 1.5 steps, and `partial 4K 4 split` lags even more.
* In the middle and late stages of training, the mean response length of `partial 4K` and `baseline_4096_4GPU` show an upward trend, while other baselines tend to stabilize.

image

We also tested these settings
qwen3 0.6b base, dataset is gsm8k; batch size 1024; rollout n = 5; max prompt length 512; 2 GPUs; FSDP trainer
3 experiments:
`baseline_512`
`partial_512`
`baseline_256` max response length 256
In these experiments, the speed gain is about 30% overall and 50% in gen. the aforementioned two problems in training dynamics are less visible.
image
image
image
image

### Additional Info.

- **Training**: none
- **Inference**: vLLM

### DIFF
--- a/verl/protocol.py
+++ b/verl/protocol.py
@@ -34,9 +34,9 @@ from packaging import version
 from tensordict import TensorDict
 from torch.utils.data import DataLoader
 
+from verl.utils.device import get_torch_device
 from verl.utils.py_functional import union_two_dict
 from verl.utils.torch_functional import allgather_dict_tensors
-from verl.utils.device import get_torch_device
 
 __all__ = ["DataProto", "union_tensor_dict"]
 
@@ -750,6 +750,34 @@ class DataProto:
             non_tensor_batch=repeated_non_tensor_batch,
             meta_info=self.meta_info,
         )
+
+    @staticmethod
+    def split(data_proto: "DataProto", filter_mask) -> tuple["DataProto", "DataProto"]:
+        """
+        Split a DataProto into two based on a boolean mask.
+
+        Args:
+            data_proto: The DataProto to split
+            filter_mask: Boolean tensor/array where True values go to the first DataProto
+
+        Returns:
+            Tuple[DataProto, DataProto]: First DataProto with items where mask is True,
+                                        Second DataProto with items where mask is False
+        """
+        # Convert to tensor if it's a list or numpy array
+        if isinstance(filter_mask, list):
+            filter_mask = torch.tensor(filter_mask, dtype=torch.bool)
+        elif isinstance(filter_mask, np.ndarray):
+            filter_mask = torch.from_numpy(filter_mask)
+
+        # Create inverse mask
+        inverse_mask = ~filter_mask
+
+        # Split into two DataProtos
+        first_proto = data_proto.select_idxs(filter_mask)
+        second_proto = data_proto.select_idxs(inverse_mask)
+
+        return first_proto, second_proto
 
 
 @dataclass
--- a/verl/workers/rollout/vllm_rollout/vllm_rollout_spmd.py
+++ b/verl/workers/rollout/vllm_rollout/vllm_rollout_spmd.py
@@ -37,7 +37,7 @@ import torch
 import torch.distributed
 from omegaconf import DictConfig, OmegaConf
 from tensordict import TensorDict
-from vllm import LLM, SamplingParams
+from vllm import LLM, RequestOutput, SamplingParams
 from vllm.distributed import parallel_state as vllm_ps
 from vllm.worker.worker_base import WorkerWrapperBase
 
@@ -235,12 +235,18 @@ class vLLMRollout(BaseRollout):
         if batch_size != len(non_tensor_batch["raw_prompt_ids"]):
             raise RuntimeError("vllm sharding manager is not work properly.")
 
+        raw_prompt_ids = non_tensor_batch.pop("raw_prompt_ids")
+        if "raw_response_ids" in non_tensor_batch:
+            raw_response_ids = non_tensor_batch.pop("raw_response_ids")
+        else:
+            raw_response_ids = np.fromiter(([] for _ in range(batch_size)), dtype=object)
+
         if "multi_modal_data" in non_tensor_batch:
             vllm_inputs = []
             for raw_prompt_ids, multi_modal_data in zip(non_tensor_batch.pop("raw_prompt_ids"), non_tensor_batch.pop("multi_modal_data")):
                 vllm_inputs.append({"prompt_token_ids": raw_prompt_ids, "multi_modal_data": multi_modal_data})
         else:
-            vllm_inputs = [{"prompt_token_ids": raw_prompt_ids} for raw_prompt_ids in non_tensor_batch.pop("raw_prompt_ids")]
+            vllm_inputs = [{"prompt_token_ids": raw_prompt_ids_ + raw_response_ids_} for raw_prompt_ids_, raw_response_ids_ in zip(raw_prompt_ids, raw_response_ids)]
 
         # ensure the type of `prompt_token_ids` passed to vllm is list[int]
         # https://github.com/volcengine/verl/pull/772
@@ -269,10 +275,15 @@ class vLLMRollout(BaseRollout):
                 "temperature": self.config.val_kwargs.temperature,
                 "n": 1,  # if validate, already repeat in ray_trainer
             }
+        else:
+            kwargs = {
+                "n": 1,  # also repeated in ray_trainer
+                "max_tokens": self.config.response_length // self.config.partial_rollout_max_split,
+            }
 
         # users can customize different sampling_params at different run
         with self.update_sampling_params(**kwargs):
-            outputs = self.inference_engine.generate(
+            outputs: list[RequestOutput] = self.inference_engine.generate(
                 prompts=vllm_inputs,  # because we have already convert it to prompt token id
                 sampling_params=self.sampling_params,
                 use_tqdm=False,
@@ -282,15 +293,22 @@ class vLLMRollout(BaseRollout):
             # if n = 1: (bs, response_length) ; if n > 1: (bs * n, response_length)
 
             response = []
+            finished = []
             rollout_log_probs = []
             for output in outputs:
                 for sample_id in range(len(output.outputs)):
                     response_ids = output.outputs[sample_id].token_ids
-                    response.append(response_ids)
+                    filtered_response = [id if id < 151669 else 0 for id in response_ids]
+                    response.append(filtered_response)
+                    finished.append(output.outputs[sample_id].finish_reason != "length")
                     curr_log_prob = []
                     for i, logprob in enumerate(output.outputs[sample_id].logprobs):
                         curr_log_prob.append(logprob[response_ids[i]].logprob)
                     rollout_log_probs.append(curr_log_prob)
+            non_tensor_batch["finished"] = np.array(finished)
+            response = raw_response_ids + np.fromiter(response, dtype=object)
+            non_tensor_batch["raw_response_ids"] = response
+            non_tensor_batch["raw_prompt_ids"] = raw_prompt_ids
 
             response = pad_2d_list_to_length(response, self.pad_token_id, max_length=self.config.response_length).to(idx.device)
             rollout_log_probs = pad_2d_list_to_length(rollout_log_probs, -1, max_length=self.config.response_length).to(idx.device)
@@ -330,7 +348,7 @@ class vLLMRollout(BaseRollout):
                 "prompts": idx,
                 "responses": response,
                 "input_ids": seq,  # here input_ids become the whole sentences
-                'rollout_log_probs': rollout_log_probs, # we will recompute old log prob with actor
+                "rollout_log_probs": rollout_log_probs,  # we will recompute old log prob with actor
                 "attention_mask": attention_mask,
                 "position_ids": position_ids,
             },
--- a/verl/trainer/config/ppo_megatron_trainer.yaml
+++ b/verl/trainer/config/ppo_megatron_trainer.yaml
@@ -129,6 +129,7 @@ actor_rollout_ref:
     top_p: 1
     prompt_length: ${data.max_prompt_length}  # for xperf_gpt
     response_length: ${data.max_response_length}
+    partial_rollout_max_split: ${algorithm.partial_rollout_max_split}
     # for vllm rollout
     dtype: bfloat16 # should align with FSDP
     gpu_memory_utilization: 0.5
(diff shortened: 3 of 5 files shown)


feifeibear:
Hi @stargazerZJ , thank you for your design diagrams and draft code.

We originally planned to develop the partial rollout feature after completing the server-based rollout refactoring. #1636 
The server-based rollout will enable more efficient inter-request asynchrony and load balancing between inference dp ranks, which is a prerequisite for partial rollout.

Would you mind joining us in the refactoring of the rollout module? We'd like to hear your thoughts first.

lilei199908:
Great work! I'm wondering if there have been any tests on the effectiveness of larger models (due to multi-step asynchrony from partial rollouts) and the efficiency on larger clusters.

eric-haibin-lin:
Are you on slack? Looking forward to more discussions.

eric-haibin-lin (commented):
A CI test is needed

eric-haibin-lin:
please also resolve conflicts with main, thx

stargazerZJ:
I'm back! I have resolved conflicts at 
https://github.com/stargazerZJ/verl/tree/partial-2
and will test it tomorrow

verl/workers/rollout/vllm_rollout/vllm_rollout_spmd.py
```diff
@@ -282,15 +293,22 @@ def generate_sequences(self, prompts: DataProto, **kwargs) -> DataProto:
             # if n = 1: (bs, response_length) ; if n > 1: (bs * n, response_length)
 
             response = []
+            finished = []
             rollout_log_probs = []
             for output in outputs:
                 for sample_id in range(len(output.outputs)):
                     response_ids = output.outputs[sample_id].token_ids
-                    response.append(response_ids)
+                    filtered_response = [id if id < 151669 else 0 for id in response_ids]
```
eric-haibin-lin, line 313: what is this magic number? 151669

eric-haibin-lin:
Split based on vllm max gen token does not seem general, as the actual rollout time may be affected by tool call/env interaction time. This needs careful consideration

verl/workers/rollout/vllm_rollout/vllm_rollout_spmd.py
protossw512: looks like the vocab size of a certain model, but why do you need filter here? did you encounter model to generate token id larger than what's in the tokenizer?

stargazerZJ: Yes. Qwen3 will do that, and will cause vllm to raise an exception with Qwen's output is fed back to its input. I'm sorry that I have forgotten to mention it when writing the PR.

stargazerZJ: The same happens to Qwen 2.5 Math too.

verl/protocol.py
```diff
@@ -757,6 +757,34 @@ def repeat(self, repeat_times=2, interleave=True):
...
+            filter_mask = torch.tensor(filter_mask, dtype=torch.bool)
+        elif isinstance(filter_mask, np.ndarray):
+            filter_mask = torch.from_numpy(filter_mask)
+
+        # Create inverse mask
+        inverse_mask = ~filter_mask
+
+        # Split into two DataProtos
+        first_proto = data_proto.select_idxs(filter_mask)
+        second_proto = data_proto.select_idxs(inverse_mask)
```
protossw512, line 784: You will running into issues here if filter_mask is either all True or all False:
`RuntimeError: batch dimension mismatch, got self.batch_size=torch.Size([32]) and value.shape=torch.Size([0, 4096]).`

Basically select_idxs won't support empty selection

stargazerZJ: Thank you. I'll test further as soon as the machines are vacant. I remembered encountering and fixing similar issues during early developing. I think during later training runs there were cases when all/no prompts are finished, and there was no RE.

protossw512: nvm, turned out I was using an older version of verl (0.3). I think this issue has been fixed. thanks!

chenjiaoAngel:
Excuse me, i use this mr to implement on my repo, but it run when partial_rollout_max_split>1, it will occur crush. the error info is :
model: Qwen2.5-VL-7B-Instruct
datasets: geo3k
max_response_length=2k
partial_rollout_max_split=2
RuntimeError: Sizes of tensors must match except in dimension 0. Expected size 3072 but got size 1024 for tensor number 1 in the list.
image

chenxia-han:
Can we specify any maximum response length, or must the original maximum response length be divisible by it?

stargazerZJ:
> Excuse me, i use this mr to implement on my repo, but it run when partial_rollout_max_split>1, it will occur crush. the error info is : model: Qwen2.5-VL-7B-Instruct datasets: geo3k max_response_length=2k partial_rollout_max_split=2 RuntimeError: Sizes of tensors must match except in dimension 0. Expected size 3072 but got size 1024 for tensor number 1 in the list. image

Sorry, multi-modal input support is not implemented in this draft pr. Support it requires adding the related fields to `partial_batch` and other batches manually. It should be trivial, but I am just not familiar with multimodal training at the moment.

BTW, sglang is not supported as well.

stargazerZJ:
> Can we specify any maximum response length, or must the original maximum response length be divisible by it?

the original maximum response length be divisible by it. Otherwise an assertion error is raised.

chenxia-han:
> > Can we specify any maximum response length, or must the original maximum response length be divisible by it?
> 
> the original maximum response length be divisible by it. Otherwise an assertion error is raised.

I’m a bit confused. Instead of relying on `age == max_split`, we could simply check the response length to determine whether it has finished. That would allow us to use any maximum response length.

stargazerZJ:
> > > Can we specify any maximum response length, or must the original maximum response length be divisible by it?
> > 
> > 
> > the original maximum response length be divisible by it. Otherwise an assertion error is raised.
> 
> I’m a bit confused. Instead of relying on `age == max_split`, we could simply check the response length to determine whether it has finished. That would allow us to use any maximum response length.

You are correct that `any maximum response length` can work this way, but that requires iterating the `raw_input_ids` to get the length of the response. Also the `vllm` sampling param needs to be copied per request to allow for different max_length for each request. It can work, just slightly more complex.

stargazerZJ:
Moreover, arbitrary long output limit is difficult to achieve because we need a preconfigured `maximum response length` to determine the size of the tensors, such as `input_ids` to be fed into the training engine and the rollout engine throughout the `fit` function. `FSDPTrainer` also need a fixed size tensor when SP is not turned on (I may misremember). Moreover, even if a fixed size `input_ids` is never needed in a specific training setting, it's transferred here and there and needs a size.

chenxia-han:
> Moreover, arbitrary long output limit is difficult to achieve because we need a preconfigured `maximum response length` to determine the size of the tensors, such as `input_ids` to be fed into the training engine and the rollout engine throughout the `fit` function. `FSDPTrainer` also need a fixed size tensor when SP is not turned on (I may misremember). Moreover, even if a fixed size `input_ids` is never needed in a specific training setting, it's transferred here and there and needs a size.

By “any maximum response length,” I mean a fixed (not dynamic) value that doesn’t necessarily divide evenly into the original maximum response length. For instance, if the original maximum is 40K, the current implementation doesn’t support 7K because 40K % 7K ≠ 0.

stargazerZJ:
Thanks for clarification. As I previously mentioned, it's possible, just slightly more complex.

chenxia-han:
> Thanks for clarification. As I previously mentioned, it's possible, just slightly more complex.

I see. Thanks for the reply.

chenjiaoAngel:
> > Excuse me, i use this mr to implement on my repo, but it run when partial_rollout_max_split>1, it will occur crush. the error info is : model: Qwen2.5-VL-7B-Instruct datasets: geo3k max_response_length=2k partial_rollout_max_split=2 RuntimeError: Sizes of tensors must match except in dimension 0. Expected size 3072 but got size 1024 for tensor number 1 in the list. image
> 
> Sorry, multi-modal input support is not implemented in this draft pr. Support it requires adding the related fields to `partial_batch` and other batches manually. It should be trivial, but I am just not familiar with multimodal training at the moment.
> 
> BTW, sglang is not supported as well.



> > Excuse me, i use this mr to implement on my repo, but it run when partial_rollout_max_split>1, it will occur crush. the error info is : model: Qwen2.5-VL-7B-Instruct datasets: geo3k max_response_length=2k partial_rollout_max_split=2 RuntimeError: Sizes of tensors must match except in dimension 0. Expected size 3072 but got size 1024 for tensor number 1 in the list. image
> 
> Sorry, multi-modal input support is not implemented in this draft pr. Support it requires adding the related fields to `partial_batch` and other batches manually. It should be trivial, but I am just not familiar with multimodal training at the moment.
> 
> BTW, sglang is not supported as well.

but i test on other model， such as Qwen2.5-32B model , it will cause this problem (i just run success on Qwen2.5-7B model, but the advantage is not obvious)
the first problem is on concat, shape is not match, i use padding method to work (self.padding(chunks - len(self) % chunks, "first"))
image 
image


the seconde proble is:
image

stargazerZJ:
> > > Excuse me, i use this mr to implement on my repo, but it run when partial_rollout_max_split>1, it will occur crush. the error info is : model: Qwen2.5-VL-7B-Instruct datasets: geo3k max_response_length=2k partial_rollout_max_split=2 RuntimeError: Sizes of tensors must match except in dimension 0. Expected size 3072 but got size 1024 for tensor number 1 in the list. image
> > 
> > 
> > Sorry, multi-modal input support is not implemented in this draft pr. Support it requires adding the related fields to `partial_batch` and other batches manually. It should be trivial, but I am just not familiar with multimodal training at the moment.
> > BTW, sglang is not supported as well.
> 
> > > Excuse me, i use this mr to implement on my repo, but it run when partial_rollout_max_split>1, it will occur crush. the error info is : model: Qwen2.5-VL-7B-Instruct datasets: geo3k max_response_length=2k partial_rollout_max_split=2 RuntimeError: Sizes of tensors must match except in dimension 0. Expected size 3072 but got size 1024 for tensor number 1 in the list. image
> > 
> > 
> > Sorry, multi-modal input support is not implemented in this draft pr. Support it requires adding the related fields to `partial_batch` and other batches manually. It should be trivial, but I am just not familiar with multimodal training at the moment.
> > BTW, sglang is not supported as well.
> 
> but i test on other model， such as Qwen2.5-32B model , it will cause this problem (i just run success on Qwen2.5-7B model, but the advantage is not obvious) the first problem is on concat, shape is not match, i use padding method to work (self.padding(chunks - len(self) % chunks, "first")) image image
> 
> the seconde proble is: image

Ah, did you `export VERL_AUTO_PADDING=1`? This is required.

Sorry for no documentation.

stargazerZJ:
During later testing and communication with the verl team, we discovered that this trick is not very scalable with more GPUs, larger model, and the DAPO algorithm with the overlong buffer.

When tested with 8 H100 on a single node, `Qwen2.5-Math-7B`, using `https://github.com/volcengine/verl/blob/main/recipe/dapo/test_dapo_7b_math.sh`, partial rollout with 2, 4 or 8 splits results in at most 1~2% speed gain (the fastest is the 4 split setting).

We can conclude that this trick works best in smaller models, less GPUs and no overlong buffer.

I decided to close the PR for this reason. Thank you to everyone that helped me during my SJTU [CS 2916](https://gair-nlp.github.io/cs2916/docs/category/2025) final project!

chenxia-han:
> During later testing and communication with the verl team, we discovered that this trick is not very scalable with more GPUs, larger model, and the DAPO algorithm with the overlong buffer.
> 
> When tested with 8 H100 on a single node, `Qwen2.5-Math-7B`, using `https://github.com/volcengine/verl/blob/main/recipe/dapo/test_dapo_7b_math.sh`, partial rollout with 2, 4 or 8 splits results in at most 1~2% speed gain (the fastest is the 4 split setting).
> 
> We can conclude that this trick works best in smaller models, less GPUs and no overlong buffer.
> 
> I decided to close the PR for this reason. Thank you to everyone that helped me during my SJTU [CS 2916](https://gair-nlp.github.io/cs2916/docs/category/2025) final project!

Could you elaborate on why it doesn’t scale? I’m genuinely curious about the bottleneck.

chenjiaoAngel:
> export VERL_AUTO_PADDING=1

OK，i will use this to try. but use this export VERL_AUTO_PADDING=1, it will influence its profiler?

PR #1861, opened 2025-06-05
[feat] add train dataset weights

Hasuer:
### Checklist Before Starting

### What does this PR do?

> Add support for weighted sampling over different training datasets, allowing users to specify sampling probabilities per dataset for more flexible and effective multi-source training.

### High-Level Design

> Based on the existing design, we extend the sampler by implementing a `WeightedRandomSampler` that assigns sampling weights to each dataset according to the number of prompts after filtering. This ensures that datasets with more usable prompts are sampled more frequently, improving training efficiency and balance.

### Specific Changes

> `verl/verl/trainer/main_ppo.py`: modified function `create_rl_sampler()`
> `verl/verl/trainer/config/ppo_trainer.yaml`: add `train_weights in data config
> `verl/verl/utils/dataset/rl_dataset.py`: modified the `__init__` of class `RLHFDataset`

### Usage Example

> If `train_weights` is provided, it should be the same size of `train_files`, and train_files must be a list with more than one dataset.
```
        data.train_files=['dataset1', 'dataset2`] \
        data.train_weights=['0.5','1'] \
        data.val_files=/data/aime24/aime24.parquet \
        data.train_batch_size=128 \
...
```
The sum of `train_weights` does not need to be 1 as it will be normalized atuomatically.

### Test

### Additional Info.

- [ ] Add `[BREAKING]` to the PR title if it breaks any API. (not needed)

### DIFF
--- a/verl/trainer/config/ppo_trainer.yaml
+++ b/verl/trainer/config/ppo_trainer.yaml
@@ -1,6 +1,7 @@
 data:
   tokenizer: null
   use_shm: False
+  train_weights: null
   train_files: ~/data/rlhf/gsm8k/train.parquet
   val_files: ~/data/rlhf/gsm8k/test.parquet
   prompt_key: prompt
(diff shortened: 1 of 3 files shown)


0x0C001:
老哥，你把百度的代理加到代码里了

Issue #1418, opened 2025-05-06
[Performance] `trainer.balance_batch=True` brings much serialization overhead for multi-modal dataset


GHGmc2:
When running GRPO with multi-model dataset, we found that throughput is much lower with `trainer.balance_batch=True` compared to `trainer.balance_batch=False`, and it's even worse if we use dataset with multi images per prompt (we didn't test video dataset yet).

Below is the example with Qwen2.5-VL-7B and geo3k dataset:
 - orange: balance_batch=True
 - blue: balance_batch=False
![Image](https://github.com/user-attachments/assets/34936175-a16e-46dc-b62b-e8bd6b4626c0)

We notice that the `non_tensor_batch` field of `DataProto` stores the inputs from multi-modal dataset as "multi_modal_inputs" part, and there're indeed `torch.Tensor` in it, but we only set `consolidate()` for `batch` field of `DataProto` now:

```python
@dataclass
class DataProto:
    def __getstate__(self):
        import io
        buffer = io.BytesIO()
        if tensordict.__version__ >= '0.5.0' and self.batch is not None:
            self.batch = self.batch.contiguous()
            self.batch = self.batch.consolidate()
        torch.save(self.batch, buffer)
        buffer_bytes = buffer.getvalue()
        return buffer_bytes, self.non_tensor_batch, self.meta_info

```

Kirrito-k423:
定时清理检查结论：该 issue 最近一次由当前运行账号（Kirrito-k423）之外账号产生的公开回复或公开引用是 2025-06-06T06:41:52Z，来源为 cross-referenced（GHGmc2）：https://github.com/verl-project/verl/issues/1418。截至 2026-07-06，该时间已经超过 6 个月。

我已检查该 issue 的公开 comments 和 timeline（issue 页面：https://github.com/verl-project/verl/issues/1418；comments/timeline 对应 GitHub 公开 issue 记录）。根据本次定时清理规则，这类长时间没有新的非本人公开互动的 open issue 将先关闭。如果作者或社区成员仍能提供新的复现、证据或需求背景，请继续评论请求重开，或创建新的 issue 并引用这里的上下文。

---

Scheduled cleanup result: the latest public reply or public reference from an account other than the current runner (Kirrito-k423) was at 2025-06-06T06:41:52Z, from cross-referenced by GHGmc2: https://github.com/verl-project/verl/issues/1418. As of 2026-07-06, that is more than 6 months ago.

[...]

PR #1862, opened 2025-06-05
[single controller] feat: mitigate pickle cost

vermouth1992:
### Checklist Before Starting

### What does this PR do?

ray put all the args in advance to avoid duplicate serialization cost for megatron dispatch.

### High-Level Design

### Specific Changes

> List the specific changes.

### API

### Usage Example

```python
```

### Test

### Additional Info.

### DIFF
--- a/verl/single_controller/base/decorator.py
+++ b/verl/single_controller/base/decorator.py
@@ -158,8 +158,8 @@ def dispatch_megatron_compute(worker_group, *args, **kwargs):
     # ray put all the args in advance to avoid duplicate serialization cost
     import ray
 
-    args = [ray.put(arg) for arg in args]
-    kwargs = {k: ray.put(v) for k, v in kwargs.items()}
+    args = [[ray.put(dp_arg) for dp_arg in arg] for arg in args]
+    kwargs = {k: [ray.put(dp_v) for dp_v in v] for k, v in kwargs.items()}
 
     all_args = []
     for arg in args:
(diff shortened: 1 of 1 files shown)


eric-haibin-lin (commented):
Why does it avoid serialization cost?

ann-qin-lu:
Changes LGTM. Tested with 40 nodes setup (TP=8, PP=10, CP=2, DP=2) and after this fix, the training time reduced by 50% :)

eric-haibin-lin (approved):

hiyouga:
@vermouth1992 will this feature reduce the time of VLMs fsdp dispatch?

vermouth1992:
No, it won't reduce fsdp dispatch as there is no redundancy in fsdp dispatch method. Eventually, we have to use uri based methods for images/videos.

ccclyu:
> ray.put serializes each argument once and returns an ObjectRef. Subsequent calls pass only the lightweight references. If these lines were removed and the arguments were large, each dispatch to the remote workers would cause Ray to serialize and transfer the full objects repeatedly—one copy per worker. For large payloads, that duplicate serialization and network transfer can significantly increase memory usage and slow down the dispatch, or even exceed available resources. Using ray.put avoids that overhead by storing the large data once and reusing the references.

Quote from [OpenAI Codex agent](https://openai.com/index/introducing-codex/) that helped me understand the code logic LOL  cc: @eric-haibin-lin

hiyouga:
@vermouth1992 sure, I'll refer to this [PR](https://github.com/hiyouga/EasyR1/pull/318) and migrate it into verl

GHGmc2:
> @vermouth1992 sure, I'll refer to this [PR](https://github.com/hiyouga/EasyR1/pull/318) and migrate it into verl

Glad to hear that, I believe this can fix the performance issue from multi-modal datasets as described here: https://github.com/volcengine/verl/issues/1418

PR #1886, opened 2025-06-06
[feat] add validation shuffle

mlpod:
### Checklist Before Starting

### What does this PR do?

In scenarios involving multiple validation sets, where the difficulty levels of these sets differ significantly and the generated content lengths vary notably, the order in which the validation sets are processed can have a substantial impact on the validation speed.

### High-Level Design

add validation shuffle

### Usage Example

```python
validation_shuffle: True
```

### Test

Validation speed increase of over 10%.

### DIFF
--- a/verl/trainer/config/ppo_trainer.yaml
+++ b/verl/trainer/config/ppo_trainer.yaml
@@ -13,6 +13,7 @@ data:
   return_raw_chat: False
   return_full_prompt: False
   shuffle: True
+  validation_shuffle: True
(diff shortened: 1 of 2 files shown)


verl/trainer/config/ppo_trainer.yaml
```diff
@@ -13,6 +13,7 @@ data:
   return_raw_chat: False
   return_full_prompt: False
   shuffle: True
+  validation_shuffle: True
```
vermouth1992: Shall we not modify the default behavior?

0x0C001: 👌

vermouth1992 (approved):

PR #1889, opened 2025-06-06
[feat] Add qwen3 multi-turn sft support

SwordFaith:
### Checklist Before Starting

### What does this PR do?

- Add enable thinking keys
- Fix multi-turn support only support user & assistant issue

### Specific Changes

> List the specific changes.

- Add `enable_thinking_keys` and `tool_keys` to config sample-level data apply chat template behavior.
- Add `convert_nested_value_to_list_recursive` to fix multi-turn type mismatch issue.

### Usage Example

ref: examples/sft/multiturn/run_qwen3_4b_sp4.sh

### Additional Info.

- **Training**: None
- **Inference**: None


examples/data_preprocess/retool_multiturn.py
```diff
@@ -0,0 +1,96 @@
+# Copyright 2023-2024 SGLang Team
```
vermouth1992: Could you move this to recipe/retool?

verl/utils/dataset/multiturn_sft_dataset.py
```diff
@@ -95,29 +119,67 @@ def __getitem__(self, item):
...
+            elif cur_messages["role"] == "system":
+                if i == 0:
+                    prev_applied_text = ""
+                else:
+                    raise ValueError("System message should be the first message")
+                    # prev_applied_text = tokenizer.apply_chat_template(messages[:i], tokenize=False, add_generation_prompt=False)
+                cur_applied_text = tokenizer.apply_chat_template(messages[:i+1], tokenize=False, add_generation_prompt=False)
+                i += 1
+            else:
+                raise ValueError(f"Unknown role: {cur_messages['role']}")
```
vermouth1992, line 279: I guess we can concatenate each token chunk and verify that it's the same as full_tokens. If it doesn't match, we can give user a warn as this is likely due to tokenizer issue.

vermouth1992: It also helps to make sure that the loss_mask is correct if the tokenizer doesn't tokenize string from two chunks as a single token.

examples/data_preprocess/retool_multiturn.py
SwordFaith: sure.

verl/utils/dataset/multiturn_sft_dataset.py
SwordFaith: Agree, a warning seems better than silence fail.

examples/data_preprocess/retool_multiturn.py
vermouth1992: You can move my previous scripts into deprecated.

examples/sft/multiturn/run_qwen2.5_32b_sp8.sh
```diff
@@ -0,0 +1,34 @@
+#!/bin/bash
```
vermouth1992: Could you move all these scripts to recipe/retool? Thanks.

SwordFaith: Ok, will do.

vermouth1992 (approved):

verl/trainer/config/sft_trainer.yaml
```diff
@@ -32,7 +33,7 @@ model:
     cpu_offload: False
     offload_params: False
   external_lib: null
-  enable_gradient_checkpointing: False
+  enable_gradient_checkpointing: True
```
eric-haibin-lin, line 36: Is this a intended breaking change?

SwordFaith: It won't affect training but will lower memory usage, allowing for a larger batch size or model size during SFT.

PR #1896, opened 2025-06-06
Hlin/dpsk

eric-haibin-lin:
### Checklist Before Starting

### What does this PR do?

Add a doc for large moe models


docs/perf/dpsk.md
```diff
@@ -0,0 +1,47 @@
...
+- optimizer offloading
+- various debugging metrics and utils
+
+and the megatron backend now has a wider list of models supported:
+- DeepSeek-V3
+- Moonlight
+- Qwen3
+- Qwen2.5-VL (to be merged soon)
+- Qwen2
+
```
ccclyu, line 20: can we also include Mixtral here in case anyone might need Mixtral series MoE models ?

vermouth1992 (approved):

PR #1739, opened 2025-05-28
[vlm] Support ulysses sequence parallelism for vlm

popomen:
### Checklist Before Starting

- [x] Search for similar PR(s).

### What does this PR do?

Only apply Ulysses sequence parallel to the LLM part of the VLM model, which is the main component, to avoid `the Image features and image tokens do not match` issue from occurring before `masked_scatter`.

### High-Level Design

> Demonstrate the high-level design if this PR is complex.

### Specific Changes

1. For the VLM model, we only pad the inputs before forward pass without slicing them; instead, we perform slicing after the embedding stage.  
2. In cases where ViT and LLM share/reuse FlashAttention, distinguish the ViT scenario and skip the Ulysses logic.

### API

> Demonstrate how the API changes if any.

### Usage Example

> Provide usage example(s) for easier usage.

```python
# Add code snippet or script demonstrating how to use this 
```

### Test

```
python -m verl.trainer.main_ppo \
    algorithm.adv_estimator=grpo \
    data.train_files=/mnt/hdfs/zhudelin123/data/geo3k/train.parquet \
    data.val_files=/mnt/hdfs/zhudelin123/data/geo3k/test.parquet \
    data.train_batch_size=64 \
    data.max_prompt_length=2048 \
    data.max_response_length=2048 \
    data.filter_overlong_prompts=True \
    data.truncation=error \
    data.image_key=images \
    actor_rollout_ref.model.path=/mnt/hdfs/Qwen2.5-VL-7B-Instruct \
    actor_rollout_ref.actor.optim.lr=1e-6 \
    actor_rollout_ref.model.use_remove_padding=True \
    actor_rollout_ref.actor.ppo_mini_batch_size=64 \
    actor_rollout_ref.actor.ppo_micro_batch_size_per_gpu=8 \
    actor_rollout_ref.actor.use_kl_loss=True \
    actor_rollout_ref.actor.kl_loss_coef=0.01 \
    actor_rollout_ref.actor.kl_loss_type=low_var_kl \
    actor_rollout_ref.actor.entropy_coeff=0 \
    actor_rollout_ref.model.enable_gradient_checkpointing=True \
    actor_rollout_ref.actor.fsdp_config.param_offload=False \
    actor_rollout_ref.actor.fsdp_config.optimizer_offload=False \
    actor_rollout_ref.model.use_fused_kernels=True \
    actor_rollout_ref.actor.ulysses_sequence_parallel_size=2 \
[...]

PR #1639, opened 2025-05-22
[feat] support kimi_vl VLM model

ShareLer:
### Checklist Before Starting

- [x] Search for similar PR(s).
Some code will conflict with this PR #1613 

### What does this PR do?

Add initial support for Kimi_vl;
Add sp patch for kimi_vl.

### High-Level Design

> Demonstrate the high-level design if this PR is complex.

### Specific Changes

- Add some minor changes to be compatible with kimi_vl
- Add patch to support ulysses_sequence_parallel

### API

> Demonstrate how the API changes if any.

### Usage Example

```bash

python3 -m verl.trainer.main_ppo \
    algorithm.adv_estimator=grpo \
    data.train_files=$DATA_PATH/geo3k/test.parquet \
    data.val_files=$DATA_PATH/geo3k/test.parquet \
    data.train_batch_size=16 \
    data.max_prompt_length=2048 \
    data.max_response_length=4096 \
    data.filter_overlong_prompts=True \
    data.truncation='error' \
    data.image_key=images \
    data.shuffle=False \
    +data.trust_remote_code=True \
    actor_rollout_ref.model.path=moonshotai/Kimi-VL-A3B-Instruct \
    actor_rollout_ref.actor.optim.lr=1e-6 \
    actor_rollout_ref.model.use_remove_padding=True \
    actor_rollout_ref.actor.ulysses_sequence_parallel_size=2 \
    actor_rollout_ref.actor.ppo_mini_batch_size=8 \
    actor_rollout_ref.actor.ppo_micro_batch_size_per_gpu=1 \
    actor_rollout_ref.actor.use_kl_loss=True \
    actor_rollout_ref.actor.kl_loss_coef=0.01 \
    actor_rollout_ref.actor.kl_loss_type=low_var_kl \
    actor_rollout_ref.actor.entropy_coeff=0 \
    actor_rollout_ref.model.enable_gradient_checkpointing=False \
    actor_rollout_ref.model.trust_remote_code=True \
    actor_rollout_ref.actor.fsdp_config.param_offload=True \
    actor_rollout_ref.actor.fsdp_config.optimizer_offload=True \
    actor_rollout_ref.rollout.log_prob_micro_batch_size_per_gpu=1 \
    actor_rollout_ref.rollout.tensor_model_parallel_size=8\
    actor_rollout_ref.rollout.name=vllm \
    actor_rollout_ref.rollout.gpu_memory_utilization=0.6 \
    actor_rollout_ref.rollout.enable_chunked_prefill=False \
[...]

PR #1899, opened 2025-06-07
fix sequence parallelism conflict in kimiVL

ShareLer:
### Checklist Before Starting

### What does this PR do?
Fix sequence parallelism conflict in kimiVL patch.

Background:
A recent VLM-related PR(#1739 ) has modified the sequence parallelism logic of VLM: Split inputs_embeds after the model's embedding layer instand of spliting input_ids and position_ids before forward.
However, the SP logic I implemented in KimiVL's PR(#1639 ) was still implemented in accordance with the old logic.  And split the image token at the combination of image_token and text_token to avoid the problem of 'the Image features and image tokens do not match'.  
Since these two PR were developed in parallel which led to logical conflicts after the PR were merged.

### High-Level Design

### Specific Changes

- Delete the patch for _merge_with_image_features which to  assign the image token to the corresponding SP rank.
- Adjust the processing related to position_ids in _ulysses_flash_attn_forward.

### API

### Usage Example

```python
```

### Test

image

### Additional Info.


eric-haibin-lin (commented):
thanks! would u mind adding a unit test in tests/models/test_transformers_ulysses.py that reproduce this error?

vermouth1992 (approved):

ShareLer:
> thanks! would u mind adding a unit test in tests/models/test_transformers_ulysses.py that reproduce this error?

It might not be possible to add the unit test of KimiVL because the unit tests in tests/models/test_transformers_ulysses.py rely on importing the model's config from transformers (ie: 'KimiVLConfig') to create a mock model by `from_config`, while in transformers the model and config of kimiVL are not defined.

It is necessary to add ulysses-related unit tests for VLM. Perhaps I can first try to add a unit test related to ulysses for QwenVL.

hiyouga:
@ShareLer The unittest for qwen2vl ulysses already exists https://github.com/volcengine/verl/blob/85fef90d518577b44abca3015249c6cde52b0be9/.github/workflows/e2e_ppo_trainer.yml#L200-L208

I think current state is sufficient

PR #1900, opened 2025-06-07
[doc]fix vllm engine default args config only support for multimodal model

777ki:
### Checklist Before Starting

### What does this PR do?

> fix doc!
As bollow, the default vllm rollout config is only support multimodal model, and the  way to fix that problem need user use `engine_kwargs` config to cover default value, this approach is unfriendly to users who are not familiar with vllm. so, we’d better to give a clearly config to show how to cover vllm default value。
bug

### High-Level Design

### Specific Changes

> List the specific changes.

### API

### Usage Example

```python
```

### Test

### Additional Info.

- [ x] Read the [Contribute Guide](https://github.com/volcengine/verl?tab=readme-ov-file#contribution-guide).
- [ x] Apply [pre-commit checks](https://github.com/volcengine/verl?tab=readme-ov-file#code-linting-and-formatting).
- [x ] Update the documentation about your changes in the [docs](https://github.com/volcengine/verl/tree/main/docs).


docs/examples/config.rst
```diff
@@ -175,6 +175,11 @@ Actor/Rollout/Reference Policy
       engine_kwargs: # inference engine parameters
         vllm:
           swap_space: null # null means "use the engine default value" (usually 4 GB), setting it to, e.g., 32 means 32 GB
+          enable_sleep_mode: True
```
eric-haibin-lin, line 178: could u help explain when users would want to change each one of these configs, and do you expect lots of users would need to change them?

docs/examples/config.rst
```diff
@@ -175,6 +175,11 @@ Actor/Rollout/Reference Policy
...
+          disable_custom_all_reduce: True
+          disable_mm_preprocessor_cache: True
```
yzlnew, line 180: Why disable_mm_preprocessor_cache is default to True, resulting in breaking a lot of existing text-only llm configs

hiyouga:
Already fixed in https://github.com/volcengine/verl/pull/2068

PR #1907, opened 2025-06-07
[Feat] Add GenRM for verify answer during training and validation

Hasuer:
### Checklist Before Starting

- [x] Searched for similar PR(s)

---

### What does this PR do?

> Adds `GenRewardModelWorker` for verifying rollout answers during training and validation.

---

### High-Level Design

> This PR introduces a new `GenRewardModelWorker` class modeled after the `ActorRolloutRefWorker`’s rollout logic.  

> It is compatible with the VLLM backend.  

> `GenRewardModelWorker` can replace the previous reward function mechanism to judge rollout results by transforming the GenRM output into binary values (0/1).

---

### Specific Changes

- `verl/trainer/config/ppo_trainer.yaml`: Added configuration options for `gen_reward_model`.
- `verl/workers/fsdp_workers.py`: Initialized GenRM.
- `verl/trainer/main_ppo.py`, `verl/trainer/ppo/ray_trainer.py`: Integrated GenRM invocation.
- `verl/workers/rollout/vllm_rollout/vllm_rollout_spmd.py`: 
  - Disabled sleep mode (`enable_sleep_mode: False`) to allow multiple VLLM instances in a single process.
  - Fixed a bug in `generate_sequences()` when using VLLM > 0.6.3 to resolve tensor shape mismatches (`non_tensor_batch` alignment issues).

---

### API Changes

> No changes to the public API.

---

### Usage Example

```bash
# Disclaimer: The model path below is for academic demonstration purposes only.
set -x

export HYDRA_FULL_ERROR=1
export VLLM_ATTENTION_BACKEND=XFORMERS  # flash_attn has known issues with qwen2-7b

train_files=/examples/data/math/train.parquet
test_files=/data/aime24/test_aime24_modefied_box.parquet

    data.train_files="$train_files" \
    data.val_files="$test_files" \
    data.train_batch_size=16 \
    data.max_prompt_length=1024 \
    data.max_response_length=512 \
    data.truncation='error' \
    data.return_raw_chat=True \
    actor_rollout_ref.model.path="/models/Qwen/Qwen2.5-0.5B" \
    actor_rollout_ref.actor.optim.lr_warmup_steps_ratio=0.1 \
    actor_rollout_ref.actor.ppo_mini_batch_size=16 \
    actor_rollout_ref.actor.ppo_micro_batch_size_per_gpu=16 \
    actor_rollout_ref.actor.use_kl_loss=False \
    actor_rollout_ref.rollout.log_prob_micro_batch_size_per_gpu=16 \
    actor_rollout_ref.rollout.n=8 \
    gen_reward_model.model.path="/models/IAAR-Shanghai/xVerify-0.5B-I" \
    gen_reward_model.model.use_remove_padding=True \
    gen_reward_model.model.enable_gradient_checkpointing=True \
    gen_reward_model.prompt_length=32704 \
    gen_reward_model.response_length=64 \
    gen_reward_model.truncation='error' \
    gen_reward_model.enable=True \
    gen_reward_model.log_prob_micro_batch_size_per_gpu=16 \
    trainer.critic_warmup=0 \
    trainer.logger=['console','wandb'] \
    trainer.project_name='math_rl' \
    trainer.experiment_name='qwen2_5_math_7b_instruct_grpo' \
    trainer.val_before_train=True \
    trainer.n_gpus_per_node=8 \
    trainer.nnodes=1 \
    trainer.save_freq=20 \
    trainer.test_freq=5 \
    trainer.total_epochs=15 $@

### DIFF
--- a/verl/trainer/config/ppo_trainer.yaml
+++ b/verl/trainer/config/ppo_trainer.yaml
@@ -1,6 +1,7 @@
 data:
   tokenizer: null
   use_shm: False
+  train_weights: null
   train_files: ~/data/rlhf/gsm8k/train.parquet
   val_files: ~/data/rlhf/gsm8k/test.parquet
   prompt_key: prompt
--- a/verl/trainer/config/ppo_trainer.yaml
+++ b/verl/trainer/config/ppo_trainer.yaml
@@ -228,6 +228,57 @@ custom_reward_function:
   path: null
   name: compute_score
 
+gen_reward_model:
+  enable: False
+  strategy: fsdp
+  mode: sync # sync: LLM, async: AsyncLLM
+  name: vllm
+  temperature: 1.0
+  top_k: -1 # 0 for hf rollout, -1 for vllm rollout
+  top_p: 1
+  use_fire_sampling: False # https://arxiv.org/abs/2410.21236
+  prompt_length: ${data.max_prompt_length}  # not use for opensource
+  response_length: ${data.max_response_length}
+  model:
+    path: ~/models/deepseek-llm-7b-chat
+    external_lib: null
+    override_config: { }
+    enable_gradient_checkpointing: True 
+    use_remove_padding: False
+    use_liger: False
+  fsdp_config:
+      wrap_policy:
+        # transformer_layer_cls_to_wrap: None
+        min_num_params: 0
+      param_offload: False
+      optimizer_offload: False
+      fsdp_size: -1
+  # for vllm rollout
+  dtype: bfloat16 # should align with FSDP
+  gpu_memory_utilization: 0.5
+  ignore_eos: False
+  enforce_eager: True
+  free_cache_engine: True
+  load_format: dummy_dtensor
+  tensor_model_parallel_size: 1
+  max_num_batched_tokens: 32768 #8192
+  max_model_len: null
+  max_num_seqs: 1024
+  log_prob_micro_batch_size: null # will be deprecated, use log_prob_micro_batch_size_per_gpu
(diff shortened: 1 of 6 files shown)


vermouth1992:
Thank you for your contribution. The way to implement GenRM using hybridengine is overkill if the model doesn't train. There are two options here:
- Directly invoke vllm or sglang server on separate GPUs, and use them directly via OpenAI client.
- Remove the FSDP part of GenRM here. Just use the vLLMRollout inside a worker group, read from checkpoint and use wakeup/sleep to share GPU memory with others.

Hasuer:
Thanks for your reply!
Could you briefly describe the implementation path of the second method？ 
Should I make all vllm instances into async mode? Any other advices?
Thanks in advance!

Hasuer:
> Thank you for your contribution. The way to implement GenRM using hybridengine is overkill if the model doesn't train. There are two options here:
> 
> * Directly invoke vllm or sglang server on separate GPUs, and use them directly via OpenAI client.
> * Remove the FSDP part of GenRM here. Just use the vLLMRollout inside a worker group, read from checkpoint and use wakeup/sleep to share GPU memory with others.

The vllm version I use is `0.8.5.post1+cu121`. In that case, the sleep model of vLLMRollout is enabled when I build vLLMRollout  for GenRM, `vllmAssertionError: Sleep mode can only be used for one instance per process.`, this assertion shows.

When I change set the sleep mode to be False in `verl/verl/workers/rollout/vllm_rollout/vllm_rollout_spmd.py`, it works well in one node, but it shows `AssertionError: Sleep mode is not enabled in the model config`. Why it works when in one node?

PR #1908, opened 2025-06-07
[recipe] feat: char count

vermouth1992:
### Checklist Before Starting

### What does this PR do?

Add a tiny recipe char count that can be run on a consumer GPU with only 8GB.

### High-Level Design

### Specific Changes

> List the specific changes.

### API

### Usage Example

```python
```

### Test

### Additional Info.


verl/trainer/config/sft_trainer.yaml
```diff
@@ -7,8 +7,8 @@ data:
   # Single-turn settings
   prompt_key: question
   response_key: answer
-  prompt_dict_keys: ['question']
-  response_dict_keys: ['answer']
```
eric-haibin-lin, line 11: is this intentional? this would be a breaking change

eric-haibin-lin:
recorded in https://github.com/volcengine/verl/discussions/1902

PR #1911, opened 2025-06-08
Support Multi-stage Awake

hebiao064:
### Checklist Before Starting

### What does this PR do?

Support Multi-stage Awake for SGLang

vLLM has supported Multi-Stage Awake for vllm engine: https://github.com/vllm-project/vllm/issues/15254

But in SGLang, we are using `torch_memory_saver` for holding the model weight and KV Cache virtual address (to make sure CUDA Graph works across different rollouts)

And `torch_memory_saver` is a singleton, which make it hard for SGLang to support Multi-Stage Awake which is critical in RL use case

Without this feature, we can only set KV Cache mem frac rate to <= 0.7 or even lower (e.g 0.3)

Currently it's failing during resume:

```bash
(WorkerDict pid=1143757) resume_memory_occupation weights
(WorkerDict pid=1143757) resume_memory_occupation weights done
(WorkerDict pid=1143757) resume_memory_occupation kv_cache
(WorkerDict pid=1143757) [torch_memory_saver.cpp] CUresult error  result=1 file=csrc/torch_memory_saver.cpp func=resume line=206
```

I doubt whether it's because the multiple instance are sharing the virtual address.

Related PR:
- Torch Memory Saver: https://github.com/fzyzcjy/torch_memory_saver/pull/17
- SGLang: https://github.com/sgl-project/sglang/pull/6962/files

### High-Level Design

### Specific Changes

> List the specific changes.

### API

### Usage Example

```python
```

### Test

### Additional Info.

### DIFF
--- a/verl/workers/rollout/sglang_rollout/sglang_rollout.py
+++ b/verl/workers/rollout/sglang_rollout/sglang_rollout.py
@@ -139,21 +139,21 @@ class AsyncEngine(sglang.srt.entrypoints.engine.Engine):
         # default to use dummy load format, which need to reload weights in first time
         self._need_reload = True
 
-    async def release_memory_occupation(self):
+    async def release_memory_occupation(self, tags: Optional[List[str]] = None):
         """Release GPU occupation temporarily."""
-        obj = ReleaseMemoryOccupationReqInput()
+        obj = ReleaseMemoryOccupationReqInput(tags=tags)
         return await self.tokenizer_manager.release_memory_occupation(obj, None)
 
-    async def resume_memory_occupation(self):
+    async def resume_memory_occupation(self, tags: Optional[List[str]] = None):
         """Resume GPU occupation."""
 
         # because __init__ is a sync method, it can not call the async release_memory_occupation
         # have to move release_memory_occupation from __init__ to here
         if self._need_reload:
-            await self.release_memory_occupation()
+            await self.release_memory_occupation(tags=tags)
             self._need_reload = False
 
-        obj = ResumeMemoryOccupationReqInput()
+        obj = ResumeMemoryOccupationReqInput(tags=tags)
         return await self.tokenizer_manager.resume_memory_occupation(obj, None)
 
     async def update_weights_from_tensor(
--- a/verl/trainer/config/ppo_trainer.yaml
+++ b/verl/trainer/config/ppo_trainer.yaml
@@ -388,6 +388,9 @@ actor_rollout_ref:
     # number of responses (i.e. num sample times). > 1 for grpo
     n: 1
 
+    # Whether to wake up inference engine in multi-stage. (Wake up model weights first, then resume kv cache)
+    multi_stage_wake_up: false
+
     # Extra inference engine arguments (vllm, sglang).
     engine_kwargs:
 
--- a/verl/trainer/ppo/ray_trainer.py
+++ b/verl/trainer/ppo/ray_trainer.py
@@ -945,7 +945,7 @@ class RayPPOTrainer:
 
         # perform validation before training
         # currently, we only support validation using the reward_function.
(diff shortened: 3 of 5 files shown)


hebiao064:
With the PR of torch_memory_saver and SGLang

```python
    async def update_weights(self, params):
        if self.device_mesh["infer_tp"].get_local_rank() == 0:
            # when we resume weight and kv cache together, it will work
            await self.inference_engine.resume_memory_occupation()
```

If we still use the old api (which will under the hood resume weight and kv cache together (but using two tms instances), it will work.

However, if we explicitly resume then by passing different tags, it will fail
```
        # When we resume them separately, it will fail with " CUresult error  result=1 file=csrc/torch_memory_saver.cpp func=resume line=206"
        #     if self.multi_stage_wake_up:
        #         print("resume_memory_occupation weights")
        #         await self.inference_engine.resume_memory_occupation(tags=["weights"])
        #         print("resume_memory_occupation weights done")
        #     else:
        #         await self.inference_engine.resume_memory_occupation()

        # if self.device_mesh["infer_tp"].get_local_rank() == 0 and self.multi_stage_wake_up:
        #     print("resume_memory_occupation kv_cache")
        #     await self.inference_engine.resume_memory_occupation(tags=["kv_cache"])
        #     print("resume_memory_occupation kv_cache done")
```

hebiao064:
Another finding, for current GRPO training for QWEN 7B, mem-fraction 0.9 would fail with: 
```
(WorkerDict pid=3022678) [torch_memory_saver.cpp] CUresult error  result=2 file=csrc/torch_memory_saver.cpp func=cu_mem_create line=103
```

Our goal is to make GRPO QWen 7B work with 0.9 or even 0.95 mem fraction by leveraging multi-stage awake

verl/workers/rollout/sglang_rollout/sglang_rollout.py
```diff
@@ -152,7 +152,7 @@ async def resume_memory_occupation(self, tags: Optional[List[str]] = None):
         # because __init__ is a sync method, it can not call the async release_memory_occupation
         # have to move release_memory_occupation from __init__ to here
         if self._need_reload:
-            await self.release_memory_occupation(tags=tags)
+            await self.release_memory_occupation()
```
hebiao064, line 144: @fzyzcjy FYI, this is why I got CURESULT=1 

Previously, we release both weights + kv cache together, so the very first time when this method being called, it will release both weights + kv cache and then resume them.

After I separated it, when I call  `release_memory_occupation(tags='weights')` it will release `weights` memory and set `_need_reload` to True.
When I call  `release_memory_occupation(tags='kv_cache')`, since `_need_reload` is False now, so we won't release `kv_cache`. Hence we are trying resume `kv_cache` while `kv_cache` is not paused.

fzyzcjy: I see. this also reminds me, one way may be add sanity checks in torch memory saver to avoid double pause / double resume

hebiao064:
> Another finding, for current GRPO training for QWEN 7B, mem-fraction 0.9 would work (peak memory goes up to 97%) on H200
> 
> When we set mem-fraction to 0.95, it will fail with:
> 
> ```
> (WorkerDict pid=3022678) [torch_memory_saver.cpp] CUresult error  result=2 file=csrc/torch_memory_saver.cpp func=cu_mem_create line=103
> ```
> 
> Our goal is make GRPO QWen 7B work with 0.95 mem fraction by leveraging multi-stage awake

Screenshot 2025-06-19 at 12 13 14 PM

Screenshot 2025-06-19 at 12 16 19 PM

zhaochenyang20:
Great work！

zhaochenyang20:
This feature is supported in sglang main and torch mem savor main. We will release 0.4.8 these days.

zhaochenyang20 (approved):
Awesome

PR #1913, opened 2025-06-08
WIP: [model] refactor: [1/N] Move model related code into DataParallelPPOActor

vermouth1992:
### Checklist Before Starting

### What does this PR do?

We are in the process of refactor to implement Model as an Engine so that the training part of the code will become more modular and reusable.

To implement Model as an Engine, we propose the following high level steps to enable smooth migration without breaking APIs and make code review much easier.
1. Move model related code into DataParallelPPOActor
2. Abstract DataParallelPPOActor into DataParallelModelEngine and a set of data preprocess/postprocess/loss function specific to PPOActor
3. Implement DataParallelPPOCritic using DataParallelModelEngine and critic specific data preprocess/postprocess/loss function
4. Implement RewardModelWorker using DataParallelModelEngine and ORM specific data preprocess/postprocess/loss function
5. Replace the OmegaConfig with explicit dataclass config

This is part to implement the first step:
- move model construct inside DataParallelPPOActor
- move FSDPCheckpointManager inside DataParallelPPOActor
- move lora config into DataParallelPPOActor
- move load/offload into DataParallelPPOActor

In the next PR, ulysses data processing will be moved into the DataParallelPPOActor

### High-Level Design

### Specific Changes

> List the specific changes.

### API

### Usage Example

```python
```

### Test

### Additional Info.

### DIFF
--- a/verl/workers/actor/dp_actor.py
+++ b/verl/workers/actor/dp_actor.py
@@ -145,7 +145,7 @@ class DataParallelPPOActor(BasePPOActor):
                 fsdp_config = self.config.fsdp_config
             else:
                 optim_config = None
-                fsdp_config = OmegaConf.create()
+                fsdp_config = self.config.fsdp_config
 
             override_model_config = OmegaConf.to_container(self.config.model.get("override_config", OmegaConf.create()))
 
--- a/verl/workers/actor/dp_actor.py
+++ b/verl/workers/actor/dp_actor.py
@@ -18,25 +18,55 @@ Single Process Actor
 """
 
 import itertools
+import json
 import logging
 import os
+import warnings
+from dataclasses import asdict
 from typing import Tuple
 
 import torch
+import torch.distributed as dist
+from omegaconf import OmegaConf
+from peft import LoraConfig, TaskType, get_peft_model
+from safetensors.torch import save_file
 from torch import nn
+from torch.distributed.device_mesh import init_device_mesh
 from torch.distributed.fsdp import FullyShardedDataParallel as FSDP
 
 import verl.utils.torch_functional as verl_F
 from verl import DataProto
+from verl.models.transformers.monkey_patch import apply_monkey_patch
 from verl.trainer.ppo.core_algos import agg_loss, compute_policy_loss, kl_penalty
-from verl.utils.debug import GPUMemoryLogger
+from verl.utils.activation_offload import enable_activation_offloading
+from verl.utils.checkpoint.fsdp_checkpoint_manager import FSDPCheckpointManager
+from verl.utils.debug import GPUMemoryLogger, log_gpu_memory_usage
 from verl.utils.device import get_device_name, get_torch_device, is_cuda_available, is_npu_available
-from verl.utils.fsdp_utils import FSDPModule, fsdp2_clip_grad_norm_
-from verl.utils.py_functional import append_to_dict
+from verl.utils.fs import copy_to_local
+from verl.utils.fsdp_utils import (
+    CPUOffloadPolicy,
+    FSDPModule,
+    MixedPrecisionPolicy,
+    apply_fsdp2,
+    fsdp2_clip_grad_norm_,
+    fsdp2_load_full_state_dict,
+    fsdp_version,
+    get_fsdp_wrap_policy,
+    get_init_weight_context_manager,
+    init_fn,
+    layered_summon_lora_params,
+    load_fsdp_model_to_gpu,
+    load_fsdp_optimizer,
+    offload_fsdp_model_to_cpu,
+    offload_fsdp_optimizer,
+)
+from verl.utils.import_utils import import_external_libs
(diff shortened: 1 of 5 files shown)


verl/workers/actor/dp_actor.py
```diff
@@ -74,6 +124,318 @@ def __init__(self, config, actor_module: nn.Module, actor_optimizer: torch.optim
...
+        # so that DataParallelActor can be instantiated solely base on config. We keep compatible with old approach
+        # that constructing model and optimizer, when pass them as arguments. We will remove the compatibility in v0.6
+        self.rank = torch.distributed.get_rank()
+        self.inference_only = inference_only
+        self._lora_rank = self.config.model.get("lora_rank", 0)
+        self._is_lora = self._lora_rank > 0
+        # build device mesh for FSDP
+        world_size = torch.distributed.get_world_size()
+        # TODO(sgm): support FSDP hybrid shard for larger model
+        self.device_mesh = create_device_mesh(world_size=world_size, fsdp_size=self.config.fsdp_config.fsdp_size)
```
eric-haibin-lin, line 137: the device mesh was still in https://github.com/volcengine/verl/blob/0eae03ae2bb369012c230edcb44de244e520e12b/verl/workers/fsdp_workers.py#L90

eric-haibin-lin (commented):
@ZihengJiang fyi..

verl/workers/actor/dp_actor.py
```diff
@@ -50,8 +79,29 @@
...
+    elif device_mesh.ndim == 2:
+        sharding_strategy = ShardingStrategy.HYBRID_SHARD
+    else:
+        raise NotImplementedError(f"Get device mesh ndim={device_mesh.ndim}, but only support 1 or 2")
+    return sharding_strategy
+
+
 class DataParallelPPOActor(BasePPOActor):
-    def __init__(self, config, actor_module: nn.Module, actor_optimizer: torch.optim.Optimizer = None):
+    def __init__(self, config, actor_module: nn.Module = None, actor_optimizer: torch.optim.Optimizer = None, inference_only=False, tokenizer=None, processor=None):
```
eric-haibin-lin, line 104: is it possible to create a unit test for dp actor now?

vermouth1992: Yes. Let's on hold this PR and discuss the details until we reach a clear path of transition

dataproblems, line 104: Can you help me understand - in what cases would we have a need for initializing a DataParallelPPOActor with no nn.Module? Also is there a reason to not use Optional[nn.Module] instead?

verl/workers/actor/dp_actor.py
```diff
@@ -74,6 +124,318 @@ def __init__(self, config, actor_module: nn.Module, actor_optimizer: torch.optim
         )
         self.device_name = get_device_name()
 
+        # Note(vermouth1992): we move the device mesh constructor into the DataParallelActor
+        # so that DataParallelActor can be instantiated solely base on config. We keep compatible with old approach
+        # that constructing model and optimizer, when pass them as arguments. We will remove the compatibility in v0.6
+        self.rank = torch.distributed.get_rank()
```
dataproblems, line 130: Let's create strategies instead? 

For e.g. 

```
class OptimizerStrategy(ABC):
    def __init__(
        self, *, role: str, sharded_model: nn.Module, optim_config: Optional[DictConfig] = None
    ):
        self._sharded_model = sharded_model
        self.role = role
        self.optim_config = optim_config
        self.optimizer = None

    @property
    def optimizer(self):
        pass

    def zero_grad(self):
        if self.optimizer is not None:
            self.optimizer.zero_grad(set_to_none=True)
        else:
            raise RuntimeError("Optimizer not initialized.")

```

Then use it in the constructor to compose the actor instead of relying on the actor constructor to do everything? 


Similar logic for parallelism strategy to deal with fsdp / megatron.

dataproblems (changes requested):

verl/workers/actor/dp_actor.py
vermouth1992: This is for backward compatible. The original design of DataParallelPPOActor requires users to init the FSDP model outside DataParallelPPOActor. Now we move the init inside DataParallelPPOActor if actor_module is None.

verl/workers/actor/dp_actor.py
vermouth1992: Nice suggestion. Let's do this one PR after another PR. For refactor, each PR should only modify a small portion of code to make review and testing easy.

verl/workers/actor/dp_actor.py
dataproblems: I see - perhaps in the future we can try to make it explicit? I'm still not sure I understand the use case where the nn.Module would be None ( in that case, what would the actor do? )

PR #1916, opened 2025-06-08
[trainer] refactor: refactor reward manager, advantage estimator

eric-haibin-lin:
### Checklist Before Starting

### What does this PR do?

Refactor advantage estimator extensions, and reward manager extensions

### High-Level Design

Add register APIs for these two classes

### API

Advantage estimator:
```
from verl.trainer.ppo.core_algos import register_adv_est
@register_adv_est("another_adv_est")
def compute_gae_advantage_return(
    token_level_rewards: torch.Tensor,
    response_mask: torch.Tensor,
    config=None,
    **kwargs):
    # custom logic goes here
    return advantages, returns

fn = get_adv_estimator_fn("another_adv_est")
advantages, returns = fn(token_level_rewards, response_mask, config=config, ...)
```

Reward manager: usage is similar

### Test

### Additional Info.

### DIFF
--- a/.github/workflows/cpu_unit_tests.yml
+++ b/.github/workflows/cpu_unit_tests.yml
@@ -55,5 +55,5 @@ jobs:
           pytest -s -x .
       - name: Running worker tests
         run: |
-          cd tests/workers
+          cd tests/workers/reward_manager
           pytest -s -x .


vermouth1992 (approved):

recipe/dapo/main_dapo.py
```diff
@@ -118,21 +118,12 @@ def run(self, config):
...
-            from verl.workers.reward_manager import NaiveRewardManager
-
-            reward_manager_cls = NaiveRewardManager
-        elif reward_manager_name == "prime":
-            from verl.workers.reward_manager import PrimeRewardManager
-
-            reward_manager_cls = PrimeRewardManager
-        elif reward_manager_name == "dapo":
-            from verl.workers.reward_manager import DAPORewardManager
+        from verl.workers.reward_manager import get_reward_manager_cls
```
dataproblems, line 121: nice! I like centralizing the reward manager initialization into a registry!

recipe/spin/main_spin.py
```diff
@@ -113,25 +113,12 @@ def run(self, config):
...
-            from verl.workers.reward_manager import DAPORewardManager
+        from verl.workers.reward_manager import get_reward_manager_cls
 
-            reward_manager_cls = DAPORewardManager
-        else:
-            raise NotImplementedError
+        # Note(haibin.lin): please make sure custom reward managers are imported and
+        # registered via `verl.workers.reward_manager.register`
+        reward_manager_name = config.reward_model.get("reward_manager", "naive")
+        reward_manager_cls = get_reward_manager_cls(reward_manager_name)
```
dataproblems, line 121: do you think it makes sense to create a factory method that also gives a reward manager instance? instead of getting just the class and separating out the initialization into a factory of its own?

eric-haibin-lin: That's something we can consider in the future. Currently I'm not sure if the reward manager constructor interface is stable

eric-haibin-lin: actually as long as the factory method provides backward compatibility, i welcome contributions. currently i think the more concerning thing is to make megatron impl follow https://github.com/volcengine/verl/discussions/1560 so that it can be unit-tested. for fsdp it has started in https://github.com/volcengine/verl/pull/1913

PR #1927, opened 2025-06-09
[FSDP] feat: Add FSDP forward pefetch and recompute chunking entropy

CurryRice233:
### Checklist Before Starting

### What does this PR do?

1. Add fsdp1 forward pefetch configuration.
2. Add chunk entropy computation.
3. Add torch.checkpoint to entropy computation.
4. Move data to device from `ActorRolloutRefWorker.update_actor` to `DataParallelPPOActor.update_policy`.
5. Add `npu_cross_entropy_loss` fusion kernel.

### High-Level Design

1. More detail see [FSDP forward_pefetch](https://docs.pytorch.org/docs/stable/fsdp.html#module-torch.distributed.fsdp)
2. `logits` usually is a large tensor [bsz\*seq_len, voc], on `compute_entropy_from_logits` will use [bsz\*seq_len, voc] * (4(float32) + 2(autocast of softmax+logsumexp) + 1(output of softmax)) memory. To reduce this memory peak, we can use chunk calculation, changing [bsz*seq_len, voc] to [chunk_size(2048), voc].
3. During the training phase, `enable_gradient_checkpointing=True` is not applicable to entropy calculation, so add the recomputation function of entropy to reduce the memory peak during the training phase.
4. On `ActorRolloutRefWorker.update_actor` all batch data is moved to the device, but this is unnecessary, `DataParallelPPOActor.update_policy` will move the data to the device for each micro batch.

### Specific Changes

> List the specific changes.

### API

Add 3 new configurations in actor/ref, 1 new configuration in critic/reward.

- actor_rollout_ref.actor.fsdp_config.forward_prefetch: False
- actor_rollout_ref.actor.entropy_from_logits_with_chunking: False
- actor_rollout_ref.actor.entropy_checkpointing: False
- actor_rollout_ref.ref.fsdp_config.forward_prefetch: False
- actor_rollout_ref.ref.entropy_from_logits_with_chunking: False
- actor_rollout_ref.ref.entropy_checkpointing: False
- critic.model.fsdp_config.forward_prefetch: False
- reward_model.model.fsdp_config.forward_prefetch: False

### Usage Example

```python
```

### Test

### Additional Info.

### DIFF
--- a/verl/trainer/config/ppo_trainer.yaml
+++ b/verl/trainer/config/ppo_trainer.yaml
@@ -201,6 +201,12 @@ actor_rollout_ref:
     # Sequence parallelism size for Ulysses-style model parallelism
     ulysses_sequence_parallel_size: 1
 
+    # calculate entropy with chunking to reduce memory peak
+    entropy_from_logits_with_chunking: False
+
+    # recompute entropy
+    entropy_checkpointing: False
+
     # checkpoint configs
     checkpoint:
     
@@ -259,6 +265,10 @@ actor_rollout_ref:
       # Number of GPUs in each FSDP shard group; -1 means auto
       fsdp_size: -1
 
+      # Only for FSDP1: FSDP1 configuration, prefetch the next forward-pass all-gather
+      # before the current forward computation.
+      forward_prefetch: False
+
   # Reference model config.
   # Reference model will be enabled when actor.use_kl_loss or/and algorithm.use_kl_in_reward is/are True.
   ref:
@@ -275,6 +285,10 @@ actor_rollout_ref:
       # whether to perform reshard after model forward to save memory.
       # only for fsdp2, [True, False, int between 1 and fsdp_size]
       reshard_after_forward: True
+
+      # Only for FSDP1: FSDP1 configuration, prefetch the next forward-pass all-gather
+      # before the current forward computation.
+      forward_prefetch: False
 
       # the wrap policy for FSDP model
       wrap_policy:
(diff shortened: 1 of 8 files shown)


verl/trainer/config/ppo_trainer.yaml
```diff
@@ -201,6 +201,12 @@ actor_rollout_ref:
     # Sequence parallelism size for Ulysses-style model parallelism
     ulysses_sequence_parallel_size: 1
 
+    # calculate entropy with chunking to reduce memory peak
+    entropy_from_logits_with_chunking: False
```
vermouth1992, line 205: Can we remove these two options? I guess they should be on by default for NPU

CurryRice233: > Can we remove these two options? I guess they should be on by default for NPU

not only for NPU, but also for GPU, chunk+recompute can reduce entropy memory utilization(e2e memory peak). If memory is sufficient, its recommended to disable these two options.

eric-haibin-lin:
@CurryRice233 thx for the contribution! Would you mind adding these options to https://verl.readthedocs.io/en/latest/perf/perf_tuning.html (https://github.com/volcengine/verl/blob/main/docs/perf/perf_tuning.rst) ?

CurryRice233:
> @CurryRice233 thx for the contribution! Would you mind adding these options to https://verl.readthedocs.io/en/latest/perf/perf_tuning.html (https://github.com/volcengine/verl/blob/main/docs/perf/perf_tuning.rst) ?

No problem, I will create a PR within this week.

PR #1932, opened 2025-06-09
`[fsdp, worker] fix: added offload for compute_ref_log_prob`

LiyuanLucasLiu:
`ref offload` is added as an option, but not implemented. This patches the missing offload functionality for ref model.

### Checklist Before Starting

- [x] Checked PR Title format
  - [x] In format of: [modules] type: Title
  - [x] modules are in `fsdp, megatron, sglang, vllm, rollout, trainer, tests, training_utils, recipe, hardware, deployment, ray, worker, single_controller, misc, perf, model, algo, env, tool, ckpt`
  - [x] type is in `feat, fix, doc, refactor, chore`
  - [ ] can involve multiple modules, seperated by `,` or space, like `[megatron, fsdp] feat: xxx`

### What does this PR do?

`ref offload` is added as an option, but not implemented. This patches the missing offload functionality for ref model

### High-Level Design

The patch follows the same implementation for `compute_log_prob`

### Specific Changes
added 5 lines in `fsdp_workers.py`

### DIFF
--- a/verl/workers/fsdp_workers.py
+++ b/verl/workers/fsdp_workers.py
@@ -697,6 +697,9 @@ class ActorRolloutRefWorker(Worker):
             data = DataProto.from_dict(tensors={"ref_log_prob": data.batch["old_log_probs"]})
             return data
         assert self._is_ref
+        if self._is_offload_param:
+            load_fsdp_model_to_gpu(self.ref_module_fsdp)
+            
         # else:
         # otherwise, the class have a standalone ref model
         # Support all hardwares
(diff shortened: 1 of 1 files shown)


vermouth1992:
Hi @LiyuanLucasLiu, reference policy already uses CPUOffload provided by FSDP. So there is no need to manually offload it.

vadimkantorov:
Is ref model implemented by FSDP? Is it possible to compute ref log probs also with vllm? (e.g. to benefit from fp8-quantization)

I also wonder if verl has any benchmarks for comparing vllm/sglang/naive pytorch for rollouts... I guess if naive can be made okay-ish (maybe using torchao or gpt-fast techniques), it's a good baseline for not having to depend on extra heavy framework like vllm/sglang

Issue #1940, opened 2025-06-10
hf rollout error


bliao-git:
I use hf for rollout instead of vllm or sglang, but got error. Please refine the following code that handles sharding manager for hf rollout:

volcengine/verl/verl/workers/fsdp_workers.py
```py
    def _build_rollout(self, trust_remote_code=False):
        from torch.distributed.device_mesh import init_device_mesh

        # TODO(sgm): support FSDP hybrid shard for larger model
        infer_tp = self.config.rollout.tensor_model_parallel_size
        dp = self.world_size // infer_tp
        assert self.world_size % infer_tp == 0, f"rollout world_size: {self.world_size} is not divisible by infer_tp: {infer_tp}"
        rollout_device_mesh = init_device_mesh(device_name, mesh_shape=(dp, infer_tp), mesh_dim_names=["dp", "infer_tp"])
        rollout_name = self.config.rollout.name
        if rollout_name == "hf":
            from verl.workers.rollout import HFRollout
            from verl.workers.sharding_manager.base import BaseShardingManager

            rollout = HFRollout(module=self.actor_module_fsdp, config=self.config.rollout)
            rollout_sharding_manager = BaseShardingManager()
            # TODO: a sharding manager that do nothing?

        elif rollout_name == "vllm":
```

bliao-git:
The error is something like : BaseShardingManager has no timing attribute

bliao-git:
The complete command and error are as follows:

```bash
PYTHONUNBUFFERED=1 python3 -m verl.trainer.main_ppo \
 algorithm.adv_estimator=grpo \
 data.train_files=$HOME/data/gsm8k/train.parquet \
 data.val_files=$HOME/data/gsm8k/test.parquet \
 data.train_batch_size=256 \
 data.max_prompt_length=512 \
 data.max_response_length=256 \
 actor_rollout_ref.model.path=Qwen/Qwen2.5-0.5B-Instruct \
 actor_rollout_ref.actor.optim.lr=1e-6 \
 actor_rollout_ref.actor.ppo_mini_batch_size=64 \
 actor_rollout_ref.actor.ppo_micro_batch_size_per_gpu=1 \
 actor_rollout_ref.rollout.name=hf \
 actor_rollout_ref.rollout.log_prob_micro_batch_size_per_gpu=1 \
 actor_rollout_ref.rollout.tensor_model_parallel_size=1 \
 actor_rollout_ref.rollout.gpu_memory_utilization=0.4 \
[...]

Issue #1926, opened 2025-06-09
vllm async is blocked by timing_raw.update(gen_batch_output.meta_info["timing"])


chenhaiq:
Run with:
rollout_mode="async"
rollout_name="vllm" 

fails at:

```
Traceback (most recent call last):
  File "/root/verl/verl/trainer/main_ppo.py", line 27, in main
    run_ppo(config)
  File "/root/verl/verl/trainer/main_ppo.py", line 46, in run_ppo
    ray.get(runner.run.remote(config))
  File "/usr/local/lib/python3.10/dist-packages/ray/_private/auto_init_hook.py", line 21, in auto_init_wrapper
    return fn(*args, **kwargs)
  File "/usr/local/lib/python3.10/dist-packages/ray/_private/client_mode_hook.py", line 103, in wrapper
    return func(*args, **kwargs)
  File "/usr/local/lib/python3.10/dist-packages/ray/_private/worker.py", line 2771, in get
    values, debugger_breakpoint = worker.get_objects(object_refs, timeout=timeout)
  File "/usr/local/lib/python3.10/dist-packages/ray/_private/worker.py", line 919, in get_objects
    raise value.as_instanceof_cause()
ray.exceptions.RayTaskError(KeyError): ray::TaskRunner.run() (pid=63403, ip=192.168.111.15, actor_id=84861f30d48e73810c87da6601000000, repr=<main_ppo.TaskRunner object at 0x7fbfc17864d0>)
  File "/root/verl/verl/trainer/main_ppo.py", line 179, in run
    trainer.fit()
  File "/root/verl/verl/trainer/ppo/ray_trainer.py", line 947, in fit
    timing_raw.update(gen_batch_output.meta_info["timing"])
KeyError: 'timing'
```

Related to https://github.com/volcengine/verl/pull/1834/files#diff-b910a162fd62b8b2a6b5c30557fee4d28d36cce12adcce62816901da27157024R640

@ETOgaosion

vermouth1992:
Do we have a vllm async e2e CI?

sharonyu-115:
I ran into the same error. Worked around temporarily by adding an if condition

```
if "timing" in gen_batch_output.meta_info:
     timing_raw.update(gen_batch_output.meta_info["timing"])
     gen_batch_output.meta_info.pop("timing", None)
```

PR #1945, opened 2025-06-10
[rollout] fix: rollout timing exceptions

ETOgaosion:
### Checklist Before Starting

- [ ] Checked PR Title format
  - [ ] In format of: [modules] type: Title
  - [ ] modules are in `fsdp, megatron, sglang, vllm, rollout, trainer, tests, training_utils, recipe, hardware, deployment, ray, worker, single_controller, misc, perf, model, algo, env, tool, ckpt`
  - [ ] type is in `feat, fix, doc, refactor, chore`
  - [ ] can involve multiple modules, seperated by `,` or space, like `[megatron, fsdp] feat: xxx`

### What does this PR do?

HFRollout skip weight sync time record to avoid error.

### High-Level Design

### Specific Changes

> List the specific changes.

### API

### Usage Example

```python
```

### Test

### Additional Info.

- **Issue Number**:
    - #1926 , @wuxibin89 @chenhaiq @sharonyu-115 please review
    - #1940 , @bliao-git please review

### DIFF
--- a/recipe/sppo/sppo_ray_trainer.py
+++ b/recipe/sppo/sppo_ray_trainer.py
@@ -185,7 +185,8 @@ class RaySPPOTrainer(RayPPOTrainer):
                         if not self.async_rollout_mode:
                             gen_batch_output = self.actor_rollout_wg.generate_sequences(gen_batch)
                         else:
-                            self.async_rollout_manager.wake_up()
+                            with _timer("wake_up", timing_raw):
+                                self.async_rollout_manager.wake_up()
                             gen_batch_output = self.async_rollout_manager.generate_sequences(gen_batch)
                             self.async_rollout_manager.sleep()
                         timing_raw.update(gen_batch_output.meta_info.get("timing", {}))


verl/workers/sharding_manager/base.py
```diff
@@ -19,6 +19,9 @@
 
 
 class BaseShardingManager:
+    def __init__(self):
+        self.timing = {}
```
vermouth1992: This is hacky...

ETOgaosion: Actually I think that initialize timer in the `__enter__` function may cause some access problems as well, so I think maybe we directly make timing a class variable, do you think that use an attribute is better than initializing in `__init__`?

ETOgaosion:
Now align the sync and async rollout, it is hard to meassure each worker's wake timing, so we directly measure ray timing of wakeup. Combined with #1949 , will solve this misalignment.

The BaseShardingManager has no timing, so use `has_attr` to check without hacking base class.

cc @vermouth1992 @chenhaiq @wuxibin89

verl/workers/fsdp_workers.py
```diff
@@ -633,7 +633,9 @@ def generate_sequences(self, prompts: DataProto):
 
             output = self.rollout_sharding_manager.postprocess_data(output)
 
-        timing_generate.update(self.rollout_sharding_manager.timing)
+        if hasattr(self.rollout_sharding_manager, "timing"):
+            # BaseShardingManager does not have timing, so we skip it
```
chenhaiq, line 637: BaseShardingManager has timing attr now. Do we need this part of code?

ETOgaosion: No need for now, historical and outdated~

Issue #1371, opened 2025-05-02
[RFC] engine interface for training backends (FSDP, FSDP2, torchtitan, Megatron, PAI-Megatron, etc)


eric-haibin-lin:
To RFC requesters: Please take a look at previous [RFCs](https://github.com/volcengine/verl/issues?q=is%3Aissue%20state%3Aopen%20label%3ARFC) for reference.
 
# Motivation
Terms:
- Roles: actor, critic, reward, etc
- Training Engines: fsdp1, fsdp2, megatron, etc

Currently, the training engines (Megatron/FSDP) are coupled with RL roles (e.g actor/critic). This makes it hard to write unit test for each module. Ideally for FSDP/Megatron engines, for development we want to have standalone test that simply runs forward/loss/backward/update. To achieve the separation of concerns, the engines better be a standalone module separated from `ActorRolloutRefWorker`. This way it would be much easier for the community to integrate different training backends, such as internal Megatron forks, torchtitan or other parallel training engines. 

On the other hand, in the codebase there's repeated code used to wrap modules with fsdp or megatron for each role (e.g. [reward](https://github.com/volcengine/verl/blob/main/verl/workers/fsdp_workers.py#L1105), [critic](https://github.com/volcengine/verl/blob/main/verl/workers/fsdp_workers.py#L842)). To achieve DRY, we need to modulize distributed model creation and put FSDP/Megatron model fwd/bwd code into a single class.
 
# Proposed Changes

Refactor actors (e.g. actor.py) agnostic to training engine backends, while training engines expose the same interface to actors.

```
worker
    engine (Model computation layer, includes various parallelism strategies. Only contains basic operations like forward, backward, and optimizer step. The definition of the loss is in the outermost driver code, passed down via a set function to the actor/critic worker layer in workers/actors, where it’s actually called.)
        megatron
            model.py (Base class for policy/value models. Includes most common functions that rarely need modification: optimizer_step, optimizer_zero_grad, etc.)
        fsdp
            model.py (Same as above)
    actors (A lightweight forwarding layer of the engine, used to unify the functions callable by the driver layer for both training and inference backends. Be cautious when adding new elements. It also serves as the server-facing interface layer — this design can still be refined.)
        actor.py (Unifies FSDP/Megatron/etc. Contains sharding/gather manager, handles some optimization-related context manager work for FSDP like sp group, and in the future, for Megatron — e.g., optimizations requiring context management such as supporting different models and parallel strategies under HE.)
[...]

Issue #1049, opened 2025-04-12
Error when using SGLang rollout


yujianll:
Hi, I get this error when using SGLang for rollout.
```
(WorkerDict pid=2199833) [2025-04-12 21:11:01 TP0] Scheduler hit an exception: Traceback (most recent call last):
(WorkerDict pid=2199833)   File "/proj/long-multi/midi/miniforge3/envs/exp_t/lib/python3.10/site-packages/sglang/srt/managers/scheduler.py", line 1999, in run_scheduler_process
(WorkerDict pid=2199833)     scheduler = Scheduler(server_args, port_args, gpu_id, tp_rank, dp_rank)
(WorkerDict pid=2199833)   File "/proj/long-multi/midi/miniforge3/envs/exp_t/lib/python3.10/site-packages/sglang/srt/managers/scheduler.py", line 249, in __init__
(WorkerDict pid=2199833)     self.tp_worker = TpWorkerClass(
(WorkerDict pid=2199833)   File "/proj/long-multi/midi/miniforge3/envs/exp_t/lib/python3.10/site-packages/sglang/srt/managers/tp_worker_overlap_thread.py", line 63, in __init__
(WorkerDict pid=2199833)     self.worker = TpModelWorker(server_args, gpu_id, tp_rank, dp_rank, nccl_port)
(WorkerDict pid=2199833)   File "/proj/long-multi/midi/miniforge3/envs/exp_t/lib/python3.10/site-packages/sglang/srt/managers/tp_worker.py", line 74, in __init__
(WorkerDict pid=2199833)     self.model_runner = ModelRunner(
(WorkerDict pid=2199833)   File "/proj/long-multi/midi/miniforge3/envs/exp_t/lib/python3.10/site-packages/sglang/srt/model_executor/model_runner.py", line 174, in __init__
(WorkerDict pid=2199833)     min_per_gpu_memory = self.init_torch_distributed()
(WorkerDict pid=2199833)   File "/proj/long-multi/midi/miniforge3/envs/exp_t/lib/python3.10/site-packages/sglang/srt/model_executor/model_runner.py", line 316, in init_torch_distributed
(WorkerDict pid=2199833)     before_avail_memory = get_available_gpu_memory(self.device, self.gpu_id)
(WorkerDict pid=2199833)   File "/proj/long-multi/midi/miniforge3/envs/exp_t/lib/python3.10/site-packages/sglang/srt/utils.py", line 277, in get_available_gpu_memory
[...]

PR #1977, opened 2025-06-11
WIP: [Refactor] Training Engine Interface and Development Plan

ZihengJiang:
# [Refactor] Training Engine Interface and Development Plan

## Motivation  
See the original RFC for background: https://github.com/volcengine/verl/issues/1371  

Modernizing our training loop requires that we:

- **Decouple** training-backend implementation from algorithm code so each can evolve independently  
- **Unify** on a single, well-defined `Engine` interface across FSDP/Megatron/etc backends  
- **Enable** unit-testing of each backend implementation in isolation  
- **Guarantee** algorithm “roles” (Critic, Actor, Rollout, Ref) remain completely engine-agnostic.
- **Preserve** the existing Roles interface exactly—so prior algorithms (e.g., `PPOTraining`) continue to work without change 

---

### Classic Training Loop with the New Interface

```python
# 1. Build and initialize engine
engine = FSDPEngine(config)
engine.init_model()
engine.set_loss_fn(loss_fn)

# 2. Training loop
for epoch in range(config.num_epochs):
    for batch in train_loader:
        # a) zero gradients
        engine.optimizer_zero_grad()

        # b) forward + backward
        with engine.train_mode():
            preds, loss, ctx = engine.forward_backward_step(
                batch,
                ctx,
                forward_only=False,
                preprocess_fn=preprocess_fn,
                postprocess_fn=postprocess_fn
            )

        # c) update and schedule
        grad_norm = engine.optimizer_step()
        current_lr = engine.lr_scheduler_step()

# 3. Evaluation
with engine.eval_mode():
    for micro_batch in data:
        preds, ctx = engine.forward_backward_step(
            micro_batch,
            ctx,
            forward_only=True,
            preprocess_fn=preprocess_fn,
            postprocess_fn=postprocess_fn
        )
```

### Detailed BaseEngine Interface
We now introduce an abstract base class, `BaseEngine`, exposing core methods:

```python

class BaseEngine(object):
    """
    Abstract base class defining the interface for model training engines.

    Engine implementations must subclass BaseEngine and provide concrete behavior for all methods.
    """
    def __init__(self, config):
        """
        Initialize the BaseEngine.

        Args:
            config: Configuration object containing parameters for engine setup.
        """
        raise NotImplementedError

    def init_model(self):
        """
        Instantiate or load the model, optimizer, and learning rate scheduler.

        Should prepare all components necessary for training or evaluation.
        """
        raise NotImplementedError

    def train_mode(self):
        """
        Context manager entry for switching the engine and model into training mode.

        Usage:
            with engine.train_mode():
                # runs in training mode
        """
        raise NotImplementedError

    def eval_mode(self):        
        """
        Context manager entry for switching the engine and model into evaluation mode.

        Usage:
            with engine.eval_mode():
                # runs in evaluation mode
        """
        raise NotImplementedError

    def forward_backward_step(self, 
                              batch, 
                              ctx=None, 
                              forward_only=False, 
                              preprocess_fn=None, 
                              postprocess_fn=None):
        """
        Execute a forward pass (and optional backward pass) over a batch of data.

        Args:
            batch: Raw batch data (e.g., tensors or mappings) to process.
            ctx: Optional context dict passed to preprocess/postprocess functions.
            forward_only: If True, skip gradient computation and backward pass.
            preprocess_fn: Function(batch, ctx) -> (inputs, ctx), applied before model call.
            postprocess_fn: Function(outputs, ctx) -> (predictions, ctx), applied after model call.

        Returns:
            If forward_only:
                (predictions, ctx)
            Else:
                (predictions, loss, ctx)
        """
        raise NotImplementedError

    def optimizer_zero_grad(self):
        """
        Zero out gradients of all parameters before starting a new backward pass.
        """
        raise NotImplementedError

    def optimizer_step(self):
        """
        Perform an optimization step to update model parameters based on accumulated gradients.

        Returns:
            grad_norm (float): The norm of the gradients before clipping or update.
        """
        raise NotImplementedError

    def lr_scheduler_step(self):
        """
        Advance the learning rate scheduler by one step.

        Returns:
            current_lr (float or list[float]): Updated learning rate(s).
        """
        raise NotImplementedError

    def shard_data(self, data):
        """
        Shard or partition data for distributed training or parallel execution.

        Args:
            data: Data structure to be sharded across devices/workers.

        Returns:
            Sharded data in the same format as input.
        """
        raise NotImplementedError

    def unshard_data(self, data):
        """
        Reconstruct or gather sharded data back to a unified format.

        Args:
            data: Sharded data structure to reconstruct.

        Returns:
            Unsharded, combined data.
        """
        raise NotImplementedError
        

    def set_loss_fn(self, loss_fn):
        """
        Set the loss function to be used during training.

        Args:
            loss_fn: Callable(data, predictions, ctx) -> (loss_tensor, new_ctx)
        """
        raise NotImplementedError

    def to(self, device: str, model: bool = True, optimizer: bool = True):
        """
        Move model parameters, optimizer states, or both to the specified device.

        Args:
            device: Target device identifier (e.g., "cuda" or "cpu").
            model: If True, move the model.
            optimizer: If True, move the optimizer states.
        """
        raise NotImplementedError


    def save_checkpoint(self, local_path, hdfs_path=None, global_step=0, max_ckpt_to_keep=None):
        """
        Save model, optimizer, and scheduler states to a checkpoint.

        Args:
            local_path: Local filesystem path to save checkpoint.
            hdfs_path: Optional HDFS path to copy checkpoint.
            global_step: Integer training step number for naming.
            max_ckpt_to_keep: Maximum number of recent checkpoints to retain.
        """
        raise NotImplementedError


    def load_checkpoint(self, local_path, hdfs_path=None, del_local_after_load=True):
        """
        Load model, optimizer, and scheduler states from a checkpoint.

        Args:
            local_path: Local filesystem path of the checkpoint.
            hdfs_path: Optional HDFS path where checkpoint is stored.
            del_local_after_load: Whether to delete local copy after loading.
        """
        raise NotImplementedError
```

### FSDPEngine Implementaion

A concrete `FSDPEngine` implements all methods using PyTorch FullyShardedDataParallel, supporting all the features that FSDP DPCritic Worker support:

- Multi-GPU/model sharding  
- Activation- and optimizer-offload  
- LoRA & sequence parallelism  
- Dynamic batch size and remove padding

### CriticWorker Implementation based on the FSDPEngine
- Unchanged public API 
- Each role calls only BaseEngine methods (init_model, train_mode/eval_mode, forward_backward_step, etc.)
- No modifications needed in existing algorithms (e.g., PPOTraining)
- New roles can be plugged in identically to legacy code

## Development Plan
We’ll roll this out in three gated phases, controlled by a feature-flag (`use_legacy_worker_impl`).

- Refactor Critic, Actor, Rollout, Ref to use only BaseEngine APIs
- Design a hierarchical, immutable config system for engine/backends
- Ensure PPO training curves and final accuracy match legacy implementation

- Enforce new interface for all feature work
- Gather benchmarks, bug reports, and performance data

- Remove legacy worker code and flags
- Delete deprecated paths
- Finalize documentation, update changelogs, close deprecation notices

Please review this refactor and share any feedback or concerns!

### DIFF
--- a/verl/trainer/config/ppo_trainer.yaml
+++ b/verl/trainer/config/ppo_trainer.yaml
@@ -819,6 +819,9 @@ trainer:
   # Device to run training on (e.g., "cuda", "cpu")
   device: cuda
 
+  # whether to use legacy role implementation
+  use_legacy_worker_impl: False
+
 # configs related to ray initialization
 ray_init:
 
--- a/verl/trainer/main_ppo.py
+++ b/verl/trainer/main_ppo.py
@@ -89,7 +89,17 @@ class TaskRunner:
         if config.actor_rollout_ref.actor.strategy in ["fsdp", "fsdp2"]:
             assert config.critic.strategy in ["fsdp", "fsdp2"]
             from verl.single_controller.ray import RayWorkerGroup
-            from verl.workers.fsdp_workers import ActorRolloutRefWorker, AsyncActorRolloutRefWorker, CriticWorker
+            from verl.workers.fsdp_workers import ActorRolloutRefWorker, AsyncActorRolloutRefWorker
+
+            print("use_legacy_worker_impl", config.trainer.use_legacy_worker_impl)
+            assert config.trainer.use_legacy_worker_impl == False
+            if config.trainer.use_legacy_worker_impl:
+                import warnings
+                warnings.warn(f"Legacy worker impl is going to be deprecated, will be removed in the future. \
+                              Please use the new worker impl supported for PPO trainer.")
+                from verl.workers.fsdp_workers import CriticWorker
+            else:
+                from verl.workers.roles import CriticWorker
 
             actor_rollout_cls = AsyncActorRolloutRefWorker if config.actor_rollout_ref.rollout.mode == "async" else ActorRolloutRefWorker
             ray_worker_group_cls = RayWorkerGroup
--- a/verl/workers/engine/__init__.py
+++ b/verl/workers/engine/__init__.py
@@ -1,0 +1,3 @@
+from .base import BaseEngine
+
+__all__ = ["BaseEngine"]
--- a/verl/workers/engine/base.py
+++ b/verl/workers/engine/base.py
@@ -1,10 +1,3 @@
-from contextlib import nullcontext
-
-"""
-TODO:
-- checkpointing
-"""
-
 class BaseEngine(object):
     def __init__(self, config):
         """
@@ -19,7 +12,7 @@ class BaseEngine(object):
         """
         raise NotImplementedError
 
-    def init_model_and_optimizer(self):
+    def init_model(self):
         """
         Initialize the model and its corresponding optimizer.
         """
@@ -39,12 +32,15 @@ class BaseEngine(object):
         """
         raise NotImplementedError
 
-    def forward_backward_step(self, batch, forward_only=False, ctx=None):
+    def forward_backward_step(self, 
+                              batch, 
+                              ctx=None, 
+                              forward_only=False, 
+                              preprocess_fn=None, 
+                              postprocess_fn=None):
         """
-        return
-        - preds
-        - loss
-        - out_ctx
+        inputs, ctx = preprocess_fn(batch, ctx)
+        preds, ctx = postprocess_fn(outputs, ctx)
         """
         raise NotImplementedError
 
@@ -81,24 +77,11 @@ class BaseEngine(object):
 
     def unshard_data(self, data):
         raise NotImplementedError
-
-    def set_preprocess_fn(self, preprocess_fn):
-        """
-        preprocess_fn(data, ctx) -> inputs, ctx
-        """
-        raise NotImplementedError
-
-
-    def set_postprocess_fn(self, postprocess_fn):
-        """
-        postprocess_fn(outputs, ctx) -> preds, ctx
-        """
-        raise NotImplementedError
         
 
     def set_loss_fn(self, loss_fn):
         """
-        loss_fn(data, preds, ctx) -> loss, out_ctx
+        loss_fn(data, preds, ctx) -> loss, ctx
         """
         raise NotImplementedError
 
--- a/verl/workers/engine/base.py
+++ b/verl/workers/engine/base.py
@@ -1,34 +1,61 @@
+# Copyright 2024 Bytedance Ltd. and/or its affiliates
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""
+The abstract base class defining the interface for model training engines.
+"""
+
+
 class BaseEngine(object):
+    """
+    Abstract base class defining the interface for model training engines.
+
+    Engine implementations must subclass BaseEngine and provide concrete behavior for all methods.
+    """
     def __init__(self, config):
         """
-        Initialize the BaseEngine instance.
-
-        This method serves as the constructor for the BaseEngine class. BaseEngine 
-        is likely an abstract base class.
+        Initialize the BaseEngine.
 
         Args:
-            config: A configuration object containing parameters and settings
-                    required for initializing the engine.
+            config: Configuration object containing parameters for engine setup.
         """
         raise NotImplementedError
 
     def init_model(self):
         """
-        Initialize the model and its corresponding optimizer.
+        Instantiate or load the model, optimizer, and learning rate scheduler.
+
+        Should prepare all components necessary for training or evaluation.
         """
         raise NotImplementedError
 
     def train_mode(self):
         """
-        Set the engine to training mode. This method is designed to switch the engine 
-        and its associated model(s) into training mode. 
+        Context manager entry for switching the engine and model into training mode.
+
+        Usage:
+            with engine.train_mode():
+                # runs in training mode
         """
         raise NotImplementedError
 
-    def eval_mode(self):
+    def eval_mode(self):        
         """
-        Set the engine to eval mode. This method is designed to switch the engine 
-        and its associated model(s) into eval mode. 
+        Context manager entry for switching the engine and model into evaluation mode.
+
+        Usage:
+            with engine.eval_mode():
+                # runs in evaluation mode
         """
         raise NotImplementedError
 
@@ -39,59 +66,113 @@ class BaseEngine(object):
                               preprocess_fn=None, 
                               postprocess_fn=None):
         """
-        inputs, ctx = preprocess_fn(batch, ctx)
-        preds, ctx = postprocess_fn(outputs, ctx)
+        Execute a forward pass (and optional backward pass) over a batch of data.
+
+        Args:
+            batch: Raw batch data (e.g., tensors or mappings) to process.
+            ctx: Optional context dict passed to preprocess/postprocess functions.
+            forward_only: If True, skip gradient computation and backward pass.
+            preprocess_fn: Function(batch, ctx) -> (inputs, ctx), applied before model call.
+            postprocess_fn: Function(outputs, ctx) -> (predictions, ctx), applied after model call.
+
+        Returns:
+            If forward_only:
+                (predictions, ctx)
+            Else:
+                (predictions, loss, ctx)
         """
         raise NotImplementedError
 
     def optimizer_zero_grad(self):
         """
-        Zero out the gradients of all the parameters optimized by the optimizer.
-        Before starting a new round of gradient computation for a fresh batch of data,
-        it's necessary to zero out the gradients to prevent the gradients from being
-        added to the previous ones.
+        Zero out gradients of all parameters before starting a new backward pass.
         """
         raise NotImplementedError
 
     def optimizer_step(self):
         """
-        Perform a single optimization step.
-        This method is designed to update the model's parameters based on the computed gradients.
+        Perform an optimization step to update model parameters based on accumulated gradients.
+
+        Returns:
+            grad_norm (float): The norm of the gradients before clipping or update.
         """
         raise NotImplementedError
 
     def lr_scheduler_step(self):
         """
-        Perform a single step of the learning rate scheduler.
+        Advance the learning rate scheduler by one step.
 
-        In the process of model training, the learning rate scheduler is used to adjust 
-        the learning rate dynamically over time, which can help the model converge better.
-        This method is designed to trigger the learning rate adjustment mechanism of the scheduler.
-        For example, it can be used to decay the learning rate after a certain number of epochs 
-        or when the validation loss plateaus.
+        Returns:
+            current_lr (float or list[float]): Updated learning rate(s).
         """
         raise NotImplementedError
 
     def shard_data(self, data):
+        """
+        Shard or partition data for distributed training or parallel execution.
+
+        Args:
+            data: Data structure to be sharded across devices/workers.
+
+        Returns:
+            Sharded data in the same format as input.
+        """
         raise NotImplementedError
 
     def unshard_data(self, data):
+        """
+        Reconstruct or gather sharded data back to a unified format.
+
+        Args:
+            data: Sharded data structure to reconstruct.
+
+        Returns:
+            Unsharded, combined data.
+        """
         raise NotImplementedError
         
 
     def set_loss_fn(self, loss_fn):
         """
-        loss_fn(data, preds, ctx) -> loss, ctx
+        Set the loss function to be used during training.
+
+        Args:
+            loss_fn: Callable(data, predictions, ctx) -> (loss_tensor, new_ctx)
         """
         raise NotImplementedError
 
     def to(self, device: str, model: bool = True, optimizer: bool = True):
+        """
+        Move model parameters, optimizer states, or both to the specified device.
+
+        Args:
+            device: Target device identifier (e.g., "cuda" or "cpu").
+            model: If True, move the model.
+            optimizer: If True, move the optimizer states.
+        """
         raise NotImplementedError
 
 
     def save_checkpoint(self, local_path, hdfs_path=None, global_step=0, max_ckpt_to_keep=None):
+        """
+        Save model, optimizer, and scheduler states to a checkpoint.
+
+        Args:
+            local_path: Local filesystem path to save checkpoint.
+            hdfs_path: Optional HDFS path to copy checkpoint.
+            global_step: Integer training step number for naming.
+            max_ckpt_to_keep: Maximum number of recent checkpoints to retain.
+        """
         raise NotImplementedError
 
 
     def load_checkpoint(self, local_path, hdfs_path=None, del_local_after_load=True):
+        """
+        Load model, optimizer, and scheduler states from a checkpoint.
+
+        Args:
+            local_path: Local filesystem path of the checkpoint.
+            hdfs_path: Optional HDFS path where checkpoint is stored.
+            del_local_after_load: Whether to delete local copy after loading.
+        """
         raise NotImplementedError
--- a/verl/workers/engine/base.py
+++ b/verl/workers/engine/base.py
@@ -102,5 +102,13 @@ class BaseEngine(object):
         """
         raise NotImplementedError
 
-    def to():
+    def to(self, device: str, model: bool = True, optimizer: bool = True):
+        raise NotImplementedError
+
+
+    def save_checkpoint(self, local_path, hdfs_path=None, global_step=0, max_ckpt_to_keep=None):
+        raise NotImplementedError
+
+
+    def load_checkpoint(self, local_path, hdfs_path=None, del_local_after_load=True):
         raise NotImplementedError
--- a/verl/workers/engine/base.py
+++ b/verl/workers/engine/base.py
@@ -35,9 +35,12 @@ class BaseEngine(object):
 
     def lr_scheduler_step(self):
         raise NotImplementedError
-    
-    def sharding_manager(self):
-        pass
+
+    def shard_data(self, data):
+        raise NotImplementedError
+
+    def unshard_data(self, data):
+        raise NotImplementedError
 
     def set_preprocess_fn(self, preprocess_fn):
         """
--- a/verl/workers/engine/base.py
+++ b/verl/workers/engine/base.py
@@ -1,0 +1,63 @@
+from contextlib import nullcontext
+
+"""
+TODO:
+- checkpointing
+"""
+
+class BaseEngine(object):
+    def __init__(self, config):
+        raise NotImplementedError
+
+    def init_model_and_optimizer(self):
+        raise NotImplementedError
+
+    def train_mode(self):
+        raise NotImplementedError
+
+    def eval_mode(self):
+        raise NotImplementedError
+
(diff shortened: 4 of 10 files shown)


vermouth1992:
@ccclyu Please given some feedbacks as the same process will need for MegatronWorker. The current design philosophy is

- Megatron and FSDP will use the same worker with only difference in the ModelEngine. (There are some issues to be solved with this because Megatron and FSDP must have different data dispatch mode). This means that we have to choose different data dispatch mode given the computation backend of CriticWorker
- Keep the legacy CriticWorker and new CriticWorker for some time and remove the legacy CriticWorker. There are several issues: 1) the CI pressure doubles. 2) how to maintain consistency when new features are added? 3) how to handle PRs that contribute to the legacy worker?

ZihengJiang:
@vermouth1992 @eric-haibin-lin @PeterSH6 @tongyx361 @ETOgaosion @hongpeng-guo @wwwjn @tianyu-l
Please review this refactor and share any feedback or concerns!

vermouth1992:
Checklist
- [ ] FSDP self-contain: whether actor/ref/critic/reward model can be easily implemented using this API @vermouth1992 @PeterSH6 
- [ ] Whether lora can be implemented using this API
- [ ] Whether VLM can be easily implemented using this API @hiyouga 
- [ ] Whether Megatron-LM can use the ModelEngine design @ISEEKYAN 
- [ ] Whether TorchTitan can use the ModelEngine design
- [ ] Whether MindSpeed can use the ModelEngine design
- [ ] How to handle data dispatch problem? Currently, it's mixed. single controller dispatch + internal resharding. @vermouth1992

vadimkantorov:
> using PyTorch FullyShardedDataParallel.

Meaning FSDP1?

ZihengJiang:
> > using PyTorch FullyShardedDataParallel.
> 
> Meaning FSDP1?

I also migrated the FSDP2 implementation in the PR. https://github.com/volcengine/verl/blob/a7459481a43596e942102b878012672e900612d5/verl/workers/engine/fsdp/engine_impl.py#L248

verl/workers/roles/critic.py
```diff
@@ -0,0 +1,362 @@
...
+            values = pad_input(values_rmpad, indices=ctx["indices"], batch=ctx["bs"], seqlen=ctx["seqlen"]).squeeze(-1)
+            values = values[:, -response_length - 1 : -1]
+            return values, ctx
+
+
+        self.use_remove_padding = self.config.model.get("use_remove_padding", False)
+        if self.use_remove_padding:
+            return preprocess_fn_with_rmpad, postprocess_fn_with_rmpad
+        else:
+            return preprocess_fn_without_rmpad, postprocess_fn_without_rmpad
```
ISEEKYAN: this part is training engine dependent, and this is FSDP's processes. Megatron use a different way. 
Now that the processes are callbacks for forward_backward_step and only be called within , I would suggest to let the Engine to have a method `set_use_rmpad()` or some option for that and all the corresponding processes run within the engine

verl/workers/roles/critic.py
```diff
@@ -0,0 +1,362 @@
...
+            micro_batches = get_micro_batches(data)
+
+            values_lst = []
+            for micro_batch in micro_batches:
+                if isinstance(micro_batch, DataProto):
+                    micro_batch = {**micro_batch.batch, **micro_batch.non_tensor_batch}
+
+                with torch.no_grad():
+                    # TODO: should not access module in the engine
+                    use_value_head_model = hasattr(self.engine.module, "v_head")
```
ISEEKYAN: this is not used in megatron

verl/workers/roles/critic.py
```diff
@@ -0,0 +1,362 @@
...
+        import torch.distributed
+
+        if not torch.distributed.is_initialized():
+            torch.distributed.init_process_group(backend="nccl" if is_cuda_available else "hccl")
+        self.config = config
+        self.ulysses_sequence_parallel_size = self.config.get("ulysses_sequence_parallel_size", 1)
+
+        # normalize config
+        self.config.ppo_mini_batch_size *= self.config.rollout_n
+        self.config.ppo_mini_batch_size //= torch.distributed.get_world_size() // self.ulysses_sequence_parallel_size
```
ISEEKYAN: ulysses_sequence_parallel_size is a concept of FSDP only, megatron do not use it

verl/workers/roles/critic.py
```diff
@@ -0,0 +1,362 @@
...
+            assert self.config.ppo_mini_batch_size % self.config.ppo_micro_batch_size_per_gpu == 0, f"normalized ppo_mini_batch_size {self.config.ppo_mini_batch_size} should be divisible by ppo_micro_batch_size_per_gpu {self.config.ppo_micro_batch_size_per_gpu}"
+            assert self.config.ppo_mini_batch_size // self.config.ppo_micro_batch_size_per_gpu > 0, f"normalized ppo_mini_batch_size {self.config.ppo_mini_batch_size} should be larger than ppo_micro_batch_size_per_gpu {self.config.ppo_micro_batch_size_per_gpu}"
+
+        self.engine = FSDPEngine(self.config)
+
+        def loss_fn(batch, vpreds, ctx):
+            responses = batch["responses"]
+            attention_mask = batch["attention_mask"]
+            values = batch["values"]
+            returns = batch["returns"]
```
ISEEKYAN: do we have a protocol for batch?

verl/workers/roles/critic.py
```diff
@@ -0,0 +1,362 @@
...
+                        elif self.config.use_dynamic_bsz:
+                            max_token_len = self.config.ppo_max_token_len_per_gpu * self.ulysses_sequence_parallel_size
+                            micro_batches, _ = rearrange_micro_batches(batch=mini_batch, max_token_len=max_token_len)
+                        else:
+                            micro_batches = mini_batch.split(self.config.ppo_micro_batch_size_per_gpu)
+                            self.gradient_accumulation = self.config.ppo_mini_batch_size // self.config.ppo_micro_batch_size_per_gpu
+
+                        self.engine.optimizer_zero_grad()
+
+                        for micro_batch in micro_batches:
```
ISEEKYAN: the iteration through mini-batch is wrapped by callbacks in megatron, instead of directly iterating every micro batch and calling a `forward_backward_step`. 
In megatron it is more proper to call a function `forward_backward_mini_batch`.

ISEEKYAN (commented):
Hello, I have some comments from aspect of megatron compatibility.

verl/workers/roles/critic.py
vermouth1992: agreed. ulysses_sequence_parallel_size should be moved into FSDPEngine

verl/workers/engine/fsdp/engine_impl.py
```diff
@@ -0,0 +1,541 @@
...
+
+
+    def unshard_data(self, data):
+        """
+        Postprocess data from sharded format back to full format.
+        """
+        return self.ulysses_sharding_manager.postprocess_data(data)
+
+
+    def forward_backward_step(self,
```
vermouth1992: forward_backward_step should train a batch, run forward + backward + optimizer step

ZihengJiang:
@vermouth1992 @ISEEKYAN made several updates:
- adapt the forward-backward-step interface for mini-batch instead of microbatch
- move ulysses config into engine implementation and left microbatch process logic also to the engine.

Let me know whether current interface looks good to you guys.

eranhirs:
Just putting here some feedback, it seems like the right place: it would be useful to be able to run non-hybrid engines, which currently throws `NotImplementedError` (see [here](https://github.com/volcengine/verl/blob/cb3dcc6f2cad898964aed893147e74b7ffb0e4df/verl/trainer/ppo/ray_trainer.py#L832)). One motivation for this is [#1049](https://github.com/volcengine/verl/issues/1049).

verl/workers/roles/critic.py
```diff
@@ -0,0 +1,288 @@
...
+        Worker.__init__(self)
+        DistProfilerExtension.__init__(
+            self, DistProfiler(rank=self.rank, config=omega_conf_to_dataclass(config.get("profiler")))
+        )
+        import torch.distributed
+
+        if not torch.distributed.is_initialized():
+            torch.distributed.init_process_group(backend="nccl" if is_cuda_available else "hccl")
+        self.config = config
+        self.engine = FSDPEngine(self.config)
```
vermouth1992: Could you make some abstract interface for megatron backend? For example, we should select the engine based on strategies

verl/workers/roles/critic.py
```diff
@@ -0,0 +1,288 @@
...
+
+            return loss, ctx
+
+        self.engine.set_loss_fn(loss_fn)
+
+    @register(dispatch_mode=Dispatch.ONE_TO_ALL)
+    def init_model(self):
+        self.engine.init_model()
+
+    def get_microbatch_process_fn(self):
```
vermouth1992: Also, make the megatron to "fill in the blank" for this function as well

verl/workers/roles/critic.py
```diff
@@ -0,0 +1,288 @@
...
+        with self.engine.eval_mode():
+            data = self.engine.shard_data(data=data)
+            output, _ = self.engine.infer_batch(
+                data, ctx=ctx, preprocess_fn=preprocess_fn, postprocess_fn=postprocess_fn
+            )
+            output = self.engine.unshard_data(data=output)
+        output = output.to("cpu")
+        return output
+
+    @register(dispatch_mode=Dispatch.DP_COMPUTE_PROTO)
```
vermouth1992, line 123: We will need to decide how we dispatch here.

verl/workers/roles/critic.py
vermouth1992: Shall we move those preprocess/postprocess function into a separate file as I expect it to be very large

vermouth1992: Maybe we can make them a class so that other backends can subclass directly

verl/workers/engine/fsdp/engine_impl.py
```diff
@@ -0,0 +1,667 @@
...
+
+
+class FSDPEngine(BaseEngine):
+    """
+    Concrete Engine implementation using PyTorch FullyShardedDataParallel (FSDP).
+
+    Supports model sharding, activation/optimizer offloading, LoRA, and sequence parallelism.
+    """
+
+    def __init__(self, config):
```
vermouth1992, line 90: Shall we define a dataclass for this config?

vermouth1992: And convert hydra config to dataclass config at worker level

verl/workers/engine/fsdp/engine_impl.py
```diff
@@ -0,0 +1,667 @@
...
+                else:
+                    inputs = batch
+
+                # forward execution
+                inputs["use_cache"] = False
+                outputs = self.module(**inputs)
+
+                # optional microbatch postprocess
+                if postprocess_fn is not None:
+                    preds, ctx = postprocess_fn(outputs, ctx)
```
vermouth1992: There may be multiple preds. In actor, there will be logprobs and entropy

verl/workers/engine/fsdp/engine_impl.py
```diff
@@ -0,0 +1,667 @@
...
+
+        if use_dynamic_bsz:
+            indices = list(itertools.chain.from_iterable(indices))
+            assert len(indices) == values.size(0), f"{len(indices)} vs. {values.size()}"
+            revert_indices = torch.tensor(get_reverse_idx(indices), dtype=torch.long)
+            values = values[revert_indices]
+
+        response_mask = data.batch["response_mask"]
+        values = values * response_mask  # Only action tokens have values
+        output = DataProto.from_dict(tensors={"values": values})
```
vermouth1992: critic-specific components such as "value" should appear in engine.

verl/workers/engine/fsdp/engine_impl.py
```diff
@@ -0,0 +1,667 @@
...
+            values_lst.append(preds)
+        values = torch.concat(values_lst, dim=0)
+
+        if use_dynamic_bsz:
+            indices = list(itertools.chain.from_iterable(indices))
+            assert len(indices) == values.size(0), f"{len(indices)} vs. {values.size()}"
+            revert_indices = torch.tensor(get_reverse_idx(indices), dtype=torch.long)
+            values = values[revert_indices]
+
+        response_mask = data.batch["response_mask"]
```
vermouth1992: We should move those logics to worker as they are critic specific.

verl/workers/engine/fsdp/engine_impl.py
```diff
@@ -0,0 +1,667 @@
...
+            preprocess_fn: Function(batch, ctx) -> (inputs, ctx), applied before model call.
+            postprocess_fn: Function(outputs, ctx) -> (predictions, ctx), applied after model call.
+
+        Returns:
+            (predictions, loss, ctx)
+        """
+        assert self.mode == "train"
+        # split batch into micro_batches
+        mini_batch = data
+        select_keys = ["input_ids", "responses", "response_mask", "attention_mask", "position_ids", "values", "returns"]
```
vermouth1992: values and returns shouldn't be here as they are critic-specific.

verl/workers/roles/critic.py
```diff
@@ -0,0 +1,288 @@
...
+            self, DistProfiler(rank=self.rank, config=omega_conf_to_dataclass(config.get("profiler")))
+        )
+        import torch.distributed
+
+        if not torch.distributed.is_initialized():
+            torch.distributed.init_process_group(backend="nccl" if is_cuda_available else "hccl")
+        self.config = config
+        self.engine = FSDPEngine(self.config)
+
+        def loss_fn(batch, vpreds, ctx):
```
eric-haibin-lin: can we avoid nested function and make it a member function and pass self.loss_fn to engine? We should make `__init__` more modular and also make it easier for others to override the loss fn in their extensions
```
def loss_fn(self, batch, vpreds, ctx)
```

verl/workers/roles/critic.py
```diff
@@ -0,0 +1,288 @@
...
+elif is_npu_available:
+    from transformers.integrations.npu_flash_attention import index_first_axis, pad_input, rearrange, unpad_input
+
+from verl.utils.config import omega_conf_to_dataclass
+from verl.utils.profiler import DistProfiler, DistProfilerExtension
+
+logger = logging.getLogger(__file__)
+logger.setLevel(os.getenv("VERL_LOGGING_LEVEL", "WARN"))
+
+
```
eric-haibin-lin, line 41: can we add some unit test for worker and engine in this PR?

eric-haibin-lin (commented):
nice!

verl/workers/roles/critic.py
ZihengJiang: use class is a good idea

verl/workers/roles/critic.py
ZihengJiang: can you elaborate on this?

ZihengJiang:
Hi @vermouth1992 @eric-haibin-lin 

I've made several updates to the PR and addressed most of the comments. The infer_batch and train_batch interfaces have gone through a few iterations and now support multi-output and callback hooks, should making them flexible enough for various worker implementations.

Could you let me know the preferred approach for testing? Do we have a common used e2e test workload?

verl/trainer/main_ppo.py
```diff
@@ -132,7 +132,15 @@ def run(self, config):
         if config.actor_rollout_ref.actor.strategy in {"fsdp", "fsdp2"}:
             assert config.critic.strategy in {"fsdp", "fsdp2"}
             from verl.single_controller.ray import RayWorkerGroup
-            from verl.workers.fsdp_workers import ActorRolloutRefWorker, AsyncActorRolloutRefWorker, CriticWorker
+            from verl.workers.fsdp_workers import ActorRolloutRefWorker, AsyncActorRolloutRefWorker
+
+            if config.trainer.use_legacy_worker_impl:
+                # import warnings
+                # warnings.warn(f"Legacy worker impl is going to be deprecated, will be removed in the future. \
+                #               Please use the new worker impl supported for PPO trainer.")
```
eric-haibin-lin: explicitly say "please set trainer.use_legacy_worker_impl = false to switch to the new worker implementation"

verl/workers/engine/__init__.py
```diff
@@ -0,0 +1,27 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from . import fsdp, megatron
+from .base import BaseEngine
+
+
+def get_training_engine(key):
```
eric-haibin-lin: for APIs included in `__init__` docstring is required

verl/workers/engine/__init__.py
```diff
@@ -0,0 +1,27 @@
...
+    if key == "fsdp":
+        return fsdp.FSDPEngine
+    elif key == "megatron":
+        raise megatron.MegatronEngine
```
eric-haibin-lin: return megatron.MegatronEngine?

verl/workers/roles/critic.py
```diff
@@ -0,0 +1,198 @@
...
+        Worker.__init__(self)
+        DistProfilerExtension.__init__(
+            self, DistProfiler(rank=self.rank, config=omega_conf_to_dataclass(config.get("profiler")))
+        )
+        import torch.distributed
+
+        if not torch.distributed.is_initialized():
+            torch.distributed.init_process_group(backend="nccl" if is_cuda_available else "hccl")
+        self.config = config
+        engine_cls = get_training_engine(self.config.strategy)
```
eric-haibin-lin: maybe change it to factory pattern like `TrainEngineRegistry.new(config, *args, **kwargs)`

verl/workers/roles/critic.py
```diff
@@ -0,0 +1,198 @@
...
+        self.engine = engine_cls(self.config)
+
+
+    @register(dispatch_mode=Dispatch.ONE_TO_ALL)
+    def init_model(self):
+        self.engine.init_model()
+
+
+    def post_fn_values(self, micro_batch, preds):
```
eric-haibin-lin: should this be private `_post_fn_values`?

verl/workers/engine/base.py
```diff
@@ -0,0 +1,182 @@
...
+    def eval_mode(self):
+        """
+        Context manager entry for switching the engine and model into evaluation mode.
+
+        Usage:
+            with engine.eval_mode():
+                # runs in evaluation mode
+        """
+        raise NotImplementedError
+
```
eric-haibin-lin, line 68: do we expect pre_fn argument for infer/train?

verl/trainer/config/ppo_trainer.yaml
```diff
@@ -925,6 +925,9 @@ trainer:
   # Device to run training on (e.g., "cuda", "cpu")
   device: cuda
 
+  # whether to use legacy role implementation
+  use_legacy_worker_impl: True
```
eric-haibin-lin: can we default it to null or "auto", so that we can automatically switch to the new impl if user do not set this config

verl/workers/roles/critic.py
ZihengJiang: add a small single gpu e2e test to do the sanity check.

eric-haibin-lin (approved):

eric-haibin-lin:
@OpenHands please sync with main branch and fix sanity test if there's any failure for https://github.com/volcengine/verl/pull/1977

verl/workers/roles/critic.py
eric-haibin-lin: we'll work on that for megatron backend

NenoL2001:
Hello， I am working at add deep speed  as a new backend engine.  Could we connect to align with the interface and config design?

zhengyaogu:
Is this feature stable? Are researchers encouraged to use this feature?

PR #2021, opened 2025-06-14
[trainer] Adding multithread_reward_computation

gty1829:
### Checklist Before Starting

### What does this PR do?
In the text-to-SQL domain, reward computation involves querying the database, making CPU processing a performance bottleneck. The compute_reward function has been enhanced with multithreading support.

### Test

736673ca672ebcc67a0d0214a7abc74
This is a comparison of computation efficiency under the same time period. The yellow line represents the number of steps computed by the original compute_reward function, while the blue line represents the steps computed after the modification.

93e69205642ca7b12e6214b40473f2d
1b05e9ca7d14d2ac85c15448434d8f3
This is a comparison of single-step computation efficiency before and after the modification.

### Specific Changes

1. Added multithreading support to the `compute_reward` function for parallel reward computation.
2. Adding `reward_fn_workers: 0` to `ppo_trainer.yml` and `ppo_megatron_trainer.yml`
3. Modified the corresponding interface of the `compute_reward` function in ray_trainer.py.
4. Adding test examples in tests/reward_score/test_compute_reward_multithread.py

### API
Original: compute_reward(data: DataProto, reward_fn)
Modify: compute_reward(data: DataProto, reward_fn, num_workers: int)
`num_workers` is the number of threads.

### Usage Example

    ...
   reward_model.reward_fn_workers=16

### NOTE
`reward_model.reward_fn_workers=0` indicates that reward computation is performed in the main process, while `reward_model.reward_fn_workers=n` (where n >= 1) means that n threads are used for reward computation.
It is recommended to use `reward_model.reward_fn_workers`  <= 16 otherwise you may get `ray.exceptions.ActorUnavailableError: The actor xxxxxxxxxxxxxxxx is unavailable: The actor is temporarily unavailable: RpcError: RPC Error message: keepalive watchdog timeout; RPC Error details:  rpc_code: 14. The task may or maynot have been executed on the actor.
`

```python
    .......
    reward_model.reward_fn_workers=16
```

### DIFF
--- a/tests/reward_score/test_compute_reward_multithread.py
+++ b/tests/reward_score/test_compute_reward_multithread.py
@@ -1,0 +1,119 @@
+# Copyright 2025 Bytedance Ltd. and/or its affiliates
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import numpy as np
+import ray
+import torch
+
+from verl import DataProto
+from verl.trainer.ppo.reward import compute_reward, compute_reward_async
+
+
+def main():
+    # -------------------------
+    # 1. Build a DataProto
+    # -------------------------
+    tensors = {
+        "input_ids": torch.tensor([[1, 2, 3], [4, 5, 6], [7, 8, 9], [10, 11, 12]], dtype=torch.float32),  # shape (4, 3)
+    }
+    non_tensors = {
+        "text": np.array(["a", "b", "c", "d"], dtype=object),
+    }
+    meta_info = {"split": "demo"}
+
+    dp = DataProto.from_dict(
+        tensors=tensors,
+        non_tensors=non_tensors,
+        meta_info=meta_info,
+    )
+
+    # -------------------------
+    # 2. Define a simple reward function
+    # -------------------------
+    def simple_reward_fn(data_proto, return_dict=False):
+        # Sum each row in input_ids → shape (batch_size,)
+        batch = data_proto.batch["input_ids"]
+        rewards = batch.sum(dim=1)
+        if return_dict:
+            return {"reward_tensor": rewards, "reward_extra_info": {"note": "sum_of_inputs"}}
+        return rewards
+
+    # -------------------------
+    # 3. Synchronous: main_process vs multi-thread
+    # -------------------------
+    r_single, info_single = compute_reward(dp, simple_reward_fn, num_workers=0)
+    r_multi, info_multi = compute_reward(dp, simple_reward_fn, num_workers=4)
+
+    # If multi-thread returns a list, concatenate into a Tensor
+    if isinstance(r_multi, list):
+        r_multi = torch.cat(r_multi, dim=0)
+
+    print("Single-worker rewards:", r_single)
+    print("Multi-worker  rewards:", r_multi)
+    print("Single-worker info:", info_single)
+    print("Multi-worker  info:", info_multi)
+
+    assert torch.allclose(r_single, r_multi), "Mismatch between main process and multi-thread outputs!"
+    assert info_single == info_multi, "Mismatch in extra_info between main process and multi-thread!"
+
+    # -------------------------
+    # 4. Asynchronous: Ray remote call
+    # -------------------------
+    ray.init(local_mode=True, ignore_reinit_error=True)
+
+    # Minimal config stub for compute_reward_async.remote
+    class CFG:
+        def __init__(self):
+            self.reward_model = type("M", (), {"get": lambda self, k, default=None: {"reward_manager": "naive", "sandbox_fusion": None, "reward_kwargs": {}}.get(k, default), "reward_fn_workers": 4})()
+            self.data = type("D", (), {"reward_fn_key": "reward_tensor"})
+
+    cfg = CFG()
+    tokenizer = None  # Not used in this example
+
+    # Monkey-patch load_reward_manager to return our simple_reward_fn
+    import verl.trainer.ppo.reward as Rmod
+
+    original_loader = Rmod.load_reward_manager
+    Rmod.load_reward_manager = lambda config, tokenizer, num_examine, **kw: simple_reward_fn
+
+    # Remote invocation
+    obj_ref = compute_reward_async.remote(dp, cfg, tokenizer)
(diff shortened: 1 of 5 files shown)


gty1829:
@ezyang @pcmoritz @corbt @WillemJiang Hi, just wanted to check in on this PR.  Would love to hear any feedback or next steps!

tongyx361:
Could you consider unify the design with https://github.com/volcengine/verl/pull/2021 ?

gty1829:
@tongyx361 Sorry, what do you mean of unify design with #2021?

tongyx361:
Sorry I had a typo. I meant unifying with https://github.com/volcengine/verl/pull/1871

gty1829:
> Sorry I had a typo. I meant unifying with #1871

I previously tried using the `compute_reward_async` function, but didn’t observe much performance improvement. It was based on `@ray.remote(num_cpus=1)` and the `naive reward manager`, which didn’t seem to include any optimization logic inside the reward manager— so I explored a multithreaded implementation instead.

tongyx361:
> > Sorry I had a typo. I meant unifying with #1871
> 
> I previously tried using the `compute_reward_async` function, but didn’t observe much performance improvement. It was based on `@ray.remote(num_cpus=1)` and the `naive reward manager`, which didn’t seem to include any optimization logic inside the reward manager— so I explored a multithreaded implementation instead.

I wonder if it's possible to unify the API to keep the API concise?

gty1829:
> > > Sorry I had a typo. I meant unifying with #1871
> > 
> > 
> > I previously tried using the `compute_reward_async` function, but didn’t observe much performance improvement. It was based on `@ray.remote(num_cpus=1)` and the `naive reward manager`, which didn’t seem to include any optimization logic inside the reward manager— so I explored a multithreaded implementation instead.
> 
> I wonder if it's possible to unify the API to keep the API concise?

You mean unify API `compute_reward` and `compute_reward_async` into the same API(e.g. `compute_reward`)?

tongyx361:
@chrisyao19 Do you have any suggestion on the concurrency design?

tongyx361:
> > > > Sorry I had a typo. I meant unifying with #1871
> > > 
> > > 
> > > I previously tried using the `compute_reward_async` function, but didn’t observe much performance improvement. It was based on `@ray.remote(num_cpus=1)` and the `naive reward manager`, which didn’t seem to include any optimization logic inside the reward manager— so I explored a multithreaded implementation instead.
> > 
> > I wonder if it's possible to unify the API to keep the API concise?
> 
> You mean unify API `compute_reward` and `compute_reward_async` into the same API(e.g. `compute_reward`)?

No need for now. I just think that it's best if you and @chrisyao19 can have a discussion on the concurrency design and @chrisyao19 can also help review.

jybsuper:
@gty1829 From what I understand, your task is IO-bound, right? If so, you might benefit from the async reward support introduced in #1871 — coroutines are generally more efficient than multithreading for IO-bound workloads. If your task is CPU-bound instead, you could consider using the `PrimeRewardManager`, which supports multi-process execution.

As for the `compute_reward_async` setting, it simply overlaps reward computation and ref logp calculation using Ray. While this can help reduce some bottlenecks, it won't significantly improve performance if your reward function is still much slower than the reference model's forward pass — which is likely why you're not seeing much gain.

tongyx361:
> @gty1829 From what I understand, your task is IO-bound, right? If so, you might benefit from the async reward support introduced in #1871 — coroutines are generally more efficient than multithreading for IO-bound workloads. If your task is CPU-bound instead, you could consider using the `PrimeRewardManager`, which supports multi-process execution.
> 
> As for the `compute_reward_async` setting, it simply overlaps reward computation and ref logp calculation using Ray. While this can help reduce some bottlenecks, it won't significantly improve performance if your reward function is still much slower than the reference model's forward pass — which is likely why you're not seeing much gain.

Indeed, `compute_reward_async` is somehow a bad name, since it actually relies on RPC instead of `asyncio`.
For IO-bound tasks, we prefer `asyncio` instead of `multithreading` for less overhead.
But it's also fine to have `multithreading` support?

vermouth1992:
Instead of adding complicated multi-theading reward computation, shall we create an abstraction of batch reward computation so that users can define their own custom logics?

gty1829:
> @gty1829 From what I understand, your task is IO-bound, right? If so, you might benefit from the async reward support introduced in #1871 — coroutines are generally more efficient than multithreading for IO-bound workloads. If your task is CPU-bound instead, you could consider using the `PrimeRewardManager`, which supports multi-process execution.
> 
> As for the `compute_reward_async` setting, it simply overlaps reward computation and ref logp calculation using Ray. While this can help reduce some bottlenecks, it won't significantly improve performance if your reward function is still much slower than the reference model's forward pass — which is likely why you're not seeing much gain.

Thanks. I'll take a look.

gty1829:
> Instead of adding complicated multi-theading reward computation, shall we create an abstraction of batch reward computation so that users can define their own custom logics?

Original function `compute_reward` can do batch computation in the main process. My PR decompose the batch data into small chunks. The number of chunks equals to the number of threads. The function `_compute_chunk` is the same as the original version of `compute_reward`, and now `compute_reward` can compute in both multithread mode and main process mode(set `num_workers=0`). Is this the abstraction of batch reward computation you mean?

gty1829:
> > @gty1829 From what I understand, your task is IO-bound, right? If so, you might benefit from the async reward support introduced in #1871 — coroutines are generally more efficient than multithreading for IO-bound workloads. If your task is CPU-bound instead, you could consider using the `PrimeRewardManager`, which supports multi-process execution.
> > As for the `compute_reward_async` setting, it simply overlaps reward computation and ref logp calculation using Ray. While this can help reduce some bottlenecks, it won't significantly improve performance if your reward function is still much slower than the reference model's forward pass — which is likely why you're not seeing much gain.
> 
> Indeed, `compute_reward_async` is somehow a bad name, since it actually relies on RPC instead of `asyncio`. For IO-bound tasks, we prefer `asyncio` instead of `multithreading` for less overhead. But it's also fine to have `multithreading` support?

I've been reading through the `PrimeRewardManager`implementation, and I noticed that the `verify`  function essentially invokes an asynchronous coroutine. I’m curious about the design choice here—specifically, why coroutine-based asynchronous execution is preferred in this context when doing parallel computation. From my understanding, async approaches are particularly effective for I/O-bound tasks, and `compute_reward_async` seems to take advantage of that. Given this, I wonder if there's an opportunity to align the synchronous `compute_reward` function with this design as well—perhaps by internally adopting async execution along with multi-threading or multi-processing. 

If both versions could share a unified, transparent backend (as @vermouth1992 suggested), it might simplify the interface and improve maintainability without sacrificing flexibility (we can abstract the multi-processing or multi-thread executor in `prime.py`).

What're your opinions? @tongyx361 @jybsuper Thanks.

jybsuper:
> Instead of adding complicated multi-theading reward computation, shall we create an abstraction of batch reward computation so that users can define their own custom logics?

@vermouth1992 
We already have a [batch reward manager](https://github.com/volcengine/verl/blob/main/verl/workers/reward_manager/batch.py) for this purpose, right?

PR #2039, opened 2025-06-16
[ci] feat: ci rerun

plutoZZZZ:
### Checklist Before Starting

- [x] Checked PR Title format
  - type is in `feat, fix, refactor, chore`

### What does this PR do?

In #1979, we migrated the CI tasks to the Vemlp platform. However, this left a problem. If we need to `rerun failed jobs` on GitHub, it will only rerun **the failed jobs and their downstream jobs**. The setup job is the task of creating a GitHub runner on VEMLP. If it is not executed, the subsequent jobs will have no runners available. Therefore, we modified the workflow and created an independent setup and cleanup job for each job. If we need to rerun failed jobs, we only need to rerun the corresponding setup job. To address the code redundancy, we encapsulated a GitHub Action and defined its inputs and outputs.
image

### Test

### High-Level Design

### Specific Changes

> List the specific changes.

### API

### Usage Example

```python
```

### DIFF
--- a/.github/workflows/e2e_ppo_trainer_megatron.yml
+++ b/.github/workflows/e2e_ppo_trainer_megatron.yml
@@ -76,34 +76,107 @@ env:
   DYNAMIC_RUNNER_ENDPOINT: "https://sd10g3clalm04ug7alq90.apigateway-cn-beijing.volceapi.com/runner"
 
 jobs:
-  setup:
+  setup0:
     runs-on: ubuntu-latest
     outputs:
-      runner-label: ${{ steps.create-runner.outputs.runner_label }}
-      mlp-task-id: ${{ steps.create-runner.outputs.mlp_task_id }}
+      runner-label: ${{ steps.create-runner.outputs.runner-label }}
+      mlp-task-id: ${{ steps.create-runner.outputs.mlp-task-id }}
     steps:
-      - name: create runner
-        id: create-runner
-        shell: bash
-        run: |
-          if [[ "${{ github.event.repository.full_name }}" != "volcengine/verl" ]]; then
-            echo "no need create runner, skip"
-            exit 0
-          fi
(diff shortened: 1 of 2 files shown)


.github/workflows/e2e_ppo_trainer_megatron.yml
```diff
@@ -76,34 +76,107 @@ env:
   DYNAMIC_RUNNER_ENDPOINT: "https://sd10g3clalm04ug7alq90.apigateway-cn-beijing.volceapi.com/runner"
 
 jobs:
-  setup:
+  setup0:
```
eric-haibin-lin, line 79: is it possible to have a template so that setup1/2/3/4 do not require lengthy configs?

plutoZZZZ: We currently do not have a solution to create templates for GitHub Actions jobs that can eliminate the need for lengthy configs across multiple setup jobs (setup0/1/2/3/4).
GitHub Actions does not natively support template inheritance in the way we need, and while reusable workflows  can reduce duplication to some extent, they still require defining each job instance explicitly in the workflow file.
If we are unable to find a better solution in the future, we can provide veCi templates and documentation to standardize the setup process. This would involve creating a set of predefined configurations that teams can reuse with minimal modifications.
cc @htc070011

plutoZZZZ: You can see our template at `.github/workflows/README.md`

plutoZZZZ: @eric-haibin-lin @vermouth1992 @tongyx361 
Would it be acceptable to provide a template like this? If not, we can discuss further. Alternatively, we could upload a version that uses an action without redundancy, encapsulating the FaaS interfaces first, but this would not support rerunning specific failed jobs.
cc @htc070011

PR #2056, opened 2025-06-17
[logging] feat: wandb entity logging for collaboration in teams

rhinosaur0:
### Checklist Before Starting

- [✅] Searched for similar PR(s).
- [✅] Checked PR Title format
  - type is in `feat, fix, refactor, chore`

### What does this PR do?

Passes in the `entity` parameter into logging, enabling syncing of wandb logs in teams.

### Specific Changes

Everywhere with logger = Tracking, pass the entity parameter

```python
        logger = Tracking(
            project_name=self.config.trainer.project_name,
            experiment_name=self.config.trainer.experiment_name,
            entity=self.config.trainer.get('entity', None),
            default_backend=self.config.trainer.logger,
            config=OmegaConf.to_container(self.config, resolve=True),
        )
```

- [✅] Read the [Contribute Guide](https://github.com/volcengine/verl?tab=readme-ov-file#contribution-guide).
- [✅] Apply [pre-commit checks](https://github.com/volcengine/verl?tab=readme-ov-file#code-linting-and-formatting).
- [✅] Add `[BREAKING]` to the PR title `description` if it breaks any API.
- [✅] Update the documentation about your changes in the [docs](https://github.com/volcengine/verl/tree/main/docs).
- [✅] Rely on existing unit tests on CI that covers the code path.

### DIFF
--- a/verl/trainer/config/ppo_trainer.yaml
+++ b/verl/trainer/config/ppo_trainer.yaml
@@ -802,6 +802,9 @@ trainer:
   # Logging backends to use: "console", "wandb", etc.
   logger: [ 'console', 'wandb' ]
 
+  # When using wandb, the group name for runs
+  entity: verl
+
   # Number of generations to log during validation
   log_val_generations: 0
 
--- a/docs/workers/ray_trainer.rst
+++ b/docs/workers/ray_trainer.rst
@@ -120,6 +120,7 @@ To extend to other RLHF algorithms, such as DPO, GRPO, please refer to
 
        logger = Tracking(project_name=self.config.trainer.project_name,
                            experiment_name=self.config.trainer.experiment_name,
+                           entity=self.config.trainer.get('entity', None),
                            default_backend=self.config.trainer.logger,
                            config=OmegaConf.to_container(self.config, resolve=True))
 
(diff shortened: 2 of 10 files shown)


vermouth1992 (approved):

verl/trainer/config/ppo_trainer.yaml
```diff
@@ -802,6 +802,9 @@ trainer:
   # Logging backends to use: "console", "wandb", etc.
   logger: [ 'console', 'wandb' ]
```
eric-haibin-lin, line 804: since verl integrates with multiple tracking backends, each may have different options. what about having this in the config instead:

```
logger_kwargs:
    entity: null
```
In the implementation, each backend only pass the kwargs to it's constructor if the value is not null.

ccclyu: highly agree with @eric-haibin-lin and as `entity` is wandb specific config and please do not setup it in the `__init__` param list. 

```
logger_cfg = config.get('logger_kwargs', {})

proxy = logger_cfg.get("wandb_proxy", None)
entity = logger_cfg.get("entity",  None)
```

mathewjhan:
@rhinosaur0 @vermouth1992 any updates to this? would be interested in this feature as well! 

maybe another solution is to read `WANDB_ENTITY` from the env right before `wandb.init`? for some reason wandb's SDK doesn't do this by default

Issue #1923, opened 2025-06-09
ValueError: `disable_mm_preprocessor_cache` is only supported for multimodal models.


FlAmmmmING:
Hello!I'm a beginner in reinforcement learning. When running the demo, I encountered this problem. Is it an issue with my environment or with my parameter settings?
'''cmd
Traceback (most recent call last):
  File "/home2/jwj/miniconda3/envs/verl/lib/python3.11/site-packages/verl/trainer/main_ppo.py", line 27, in main
    run_ppo(config)
  File "/home2/jwj/miniconda3/envs/verl/lib/python3.11/site-packages/verl/trainer/main_ppo.py", line 46, in run_ppo
    ray.get(runner.run.remote(config))
  File "/home2/jwj/miniconda3/envs/verl/lib/python3.11/site-packages/ray/_private/auto_init_hook.py", line 21, in auto_init_wrapper
    return fn(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^
  File "/home2/jwj/miniconda3/envs/verl/lib/python3.11/site-packages/ray/_private/client_mode_hook.py", line 103, in wrapper
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/home2/jwj/miniconda3/envs/verl/lib/python3.11/site-packages/ray/_private/worker.py", line 2822, in get
    values, debugger_breakpoint = worker.get_objects(object_refs, timeout=timeout)
                                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home2/jwj/miniconda3/envs/verl/lib/python3.11/site-packages/ray/_private/worker.py", line 930, in get_objects
    raise value.as_instanceof_cause()
ray.exceptions.RayTaskError(ValueError): ray::TaskRunner.run() (pid=367439, ip=159.226.47.239, actor_id=b4ab27d77122883c9b5e255c01000000, repr=<main_ppo.TaskRunner object at 0x7d0fe0163650>)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home2/jwj/miniconda3/envs/verl/lib/python3.11/site-packages/verl/trainer/main_ppo.py", line 177, in run
    trainer.init_workers()
[...]

PR #2068, opened 2025-06-17
modify the default value of disable_mm_preprocessor_cache

yyDing1:
All scripts using LLM (Non-VLM + vllm rollout backend) break (Error details can be found at issue https://github.com/volcengine/verl/issues/1923, also mentioned in PR https://github.com/volcengine/verl/pull/1900)
This error currently occurs in vllm>=0.9.0).

The reason is that `disable_mm_preprocessor_cache=True` only works for VLM, and will cause errors for non-VLM models.

It appears that the default value in vllm is `False` and it's recommended to be set to False, even for VLM, according to official guidelines below:

https://github.com/vllm-project/vllm/blob/ca94d7fa007d113f2daf74e2f2278c03be676e93/vllm/config.py#L380C5-L382

Therefore, it's would be better to set `disable_mm_preprocessor_cache` to `False` here.

### DIFF
--- a/verl/workers/rollout/vllm_rollout/vllm_async_server.py
+++ b/verl/workers/rollout/vllm_rollout/vllm_async_server.py
@@ -172,7 +172,7 @@ class AsyncvLLMServer(AsyncServerBase):
             enforce_eager=config.enforce_eager,
             gpu_memory_utilization=config.gpu_memory_utilization,
             disable_custom_all_reduce=True,
-            disable_mm_preprocessor_cache=True,
+            disable_mm_preprocessor_cache=False,
             skip_tokenizer_init=False,
             max_model_len=max_model_len,
             load_format="auto",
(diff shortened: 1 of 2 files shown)


hiyouga:
Set `disable_mm_preprocessor_cache` to False may cause CPU memory leak in our previous experiments. Could we use a config entry to control this behavior and disable mm processor cache by default for VLMs?

eric-haibin-lin (commented):
what about including this arg in vllm.engine_kwargs ?

yyDing1:
Good suggestion. I have made the following changes:

- mv `disable_mm_preprocessor_cache` to vllm engine args. It can be activated by adding 
`--actor_rollout_ref.rollout.engine_kwargs.vllm.disable_mm_preprocessor_cache=True`

- update docs

hiyouga:
I think it's necessary to disable the mm processor cache in the vlm examples: 
https://github.com/volcengine/verl/blob/main/examples/grpo_trainer/
https://github.com/volcengine/verl/blob/main/examples/grpo_trainer/run_qwen2_5_vl-7b.sh

Could you fix them along with this changes?

yyDing1:
Fixed. I also notice that `engine_kwargs` seems to be used only for vllm engine, so it may not cause error for sglang engine.

hiyouga (approved):
LGTM

PR #2092, opened 2025-06-18
[megatron] feat: patch mcore 0.12.0 distributed optimizer to save sharded optimizer state from GPUs to disk directly without DP gathering to CPU memory

stevewx:
### Checklist Before Starting

- [x] Checked PR Title format

### What does this PR do?

For models with large parameter count, gathering optimizer states to RANK 0 CPU memory poses OOM risk.

This PR patches mcore 0.12.0 distributed optimizer to save sharded optimizer state from GPUs to disk directly without DP gathering to CPU memory.

### Test
E2e run with checkpoint saving, loading, and resumption using mixtral model megatron rloo training, and verify rewards trend, pg_loss and entropy_loss.

### High-Level Design
NA.

### Specific Changes
Added new methods for Megatron ChainedOptimizer and DistributedOptimizer to save sharded optimizer state from GPUs to disk directly without DP gathering to CPU memory

### API

+ DistributedOptimizer.get_parameter_state_local
+ DistributedOptimizer.load_parameter_state_local
+ DistributedOptimizer.save_sharded_parameter_state
+ DistributedOptimizer.load_sharded_parameter_state
+ ChainedOptimizer.save_sharded_parameter_state
+ ChainedOptimizer.load_sharded_parameter_state

### Usage Example
included in megatron_checkpoint_manager.py


verl/models/mcore/patch_v012.py
```diff
@@ -198,3 +199,200 @@ def qkv_up_proj_and_rope_apply(q_compressed, kv_compressed, k_pos_emb, rotary_po
         return query, key, value
 
     MLASelfAttention.get_query_key_value_tensors = patch_get_query_key_value_tensors
+
+
+def apply_optimizer_sharded_save_load_patches():
```
ETOgaosion: From my point of view, the cases where we use monkey patches are those where it's hard to initialize instances with specific class, for example, `model = AutoModel.from_pretrained`. Its dynamicity is not a good way of design.

But in case of Megatron, since there is a good abstraction of `GPTModel` and other transformer classes, we can directly inherit them and replace the function. Then in `model_initializer`, we can build models with our own classes. So I suggest that in Megatron, patch can be custom models inheriting GPTModels, maybe need some refactor.

Here we may use a custom `DistributedOptimizer` and `ChainedOptimizer` to inherit the original classes? Also we need to modify the `get_megatron_optimizer`.

@ccclyu May I ask what's your opinion of these monkey patches?

ETOgaosion: Like the GPTModelWithFusedLayer in https://github.com/volcengine/verl/pull/2103/files#diff-805f5deec9595136ccf2612b71128ef80f2093db968f2d7cd5e94430a0413add

stevewx: Thanks for the feedback! Taking a look

stevewx:
Thanks @ETOgaosion for the suggestion, updated the monkey patch to inheritance based class solution. Can you please take another look?

ETOgaosion:
LGTM, this is what I expected.

ETOgaosion:
@stevewx Dist_ckpt has been merged, can you rebase main and resolve conflicts?

stevewx:
Oh nice, thanks for the pointer I see we have shifted to dist_checkpointing for megatron overall. Since we are no longer using `save_parameter_state` in megatron checkpoint manager, we can close this PR. Lmk if we want to keep this patch in place for future optimizer related customizations.

PR #2107, opened 2025-06-19
[DAPO] chore: add negative expected_len warning

LeavesLei:
#### Summary
Added a warning mechanism to detect and alert users when the overlong buffer configuration is invalid, specifically when `overlong_buffer_len > max_response_length` which causes `expected_len` to be negative.

#### Problem
When `overlong_buffer_len` is greater than `max_response_length`, the calculated `expected_len` becomes negative:
```
expected_len = max_resp_len - overlong_buffer_len  # Results in negative value
```
This causes all reasonable response lengths to be penalized.

#### Solution
Added a `UserWarning` that triggers when `expected_len < 0`:
```
if expected_len < 0:
    warnings.warn(
        f"Overlong buffer configuration issue: expected_len = {expected_len} < 0. "
        f"This means overlong_buffer_len ({overlong_buffer_len}) > max_resp_len ({self.max_resp_len}). "
        f"All reasonable response lengths will be penalized. "
        f"Consider reducing overlong_buffer_len or increasing max_resp_len."
    )
```

#### Changes Made
1. File: `verl/workers/reward_manager/dapo.py`
2. Added: `import warnings` at the top


LeavesLei:
Hi, I have fixed the import error:)

verl/workers/reward_manager/dapo.py
```diff
@@ -107,6 +108,15 @@ def __call__(self, data: DataProto, return_dict: bool = False):
             if self.overlong_buffer_cfg.enable:
                 overlong_buffer_len = self.overlong_buffer_cfg.len
                 expected_len = self.max_resp_len - overlong_buffer_len
+
+                if expected_len < 0:
+                    warnings.warn(
```
wuxibin89: Move to `DAPORewardManager.__init__` as an assertion?

LeavesLei: Thanks for your suggestion. I have opened a new pull request, please refer to https://github.com/volcengine/verl/pull/2330

PR #1946, opened 2025-06-10
[rollout] feat: Add LangGraph chat scheduler for agent training with tool use support

waleko:
### Checklist Before Starting

- [x] Search for similar PR(s).

> #1831 also leverages async server's openai api, but doesn't simplify agent creation
> #1668 focuses on simplifying loss masking, but in a more low-level way

### What does this PR do?

Add LangGraph chat scheduler for simpler agent training and native support for tool use

### High-Level Design

This PR introduces a [LangGraph](https://langchain-ai.github.io/langgraph/)-based chat completion scheduler that extends the existing `ChatCompletionScheduler` to support complex multi-turn reasoning and tool calling workflows. The implementation includes:
- **LangGraphChatCompletionScheduler**: A new chat scheduler that uses LangGraph for orchestrating complex chat completions
- **Masking calculation**: `loss_mask` is calculated by using `return_assistant_tokens_mask` to identify assistant tokens from HuggingFace's `apply_chat_template`. This makes it possible to reuse SGLang's multi-turn options in the `ray_trainer`
- **Retool Experiment with SandboxFusion and self-reflection**: an example of using the new chat scheduler with SandboxFusion tool and self-reflection agentic setup

### Specific Changes

1. **Added LangGraph Chat Scheduler** (`examples/ppo_trainer/langgraph_chat_scheduler.py`):
   - New `LangGraphChatCompletionScheduler` class extending `ChatCompletionScheduler`. Given the efforts of #1769, it is possible to train it not only with vllm, but also with SGLang.
   - Async rollout generation with arbitrary LangGraph runnables. By using Hydra configuration, it is possible to import an existing LangGraph agent, and use it in VeRL's rollout. This helps avoids duplicating the code for the training, and allows to use the same tools / multi-turn logic / prompts during training and inference. Furthermore, LangChain/LangGraph ecosystem is very vast, and it is possible to use a lot of existing tools including MCP.
[...]

PR #1630, opened 2025-05-22
[sglang] feat: Support async multi-turn rollout with simulation feedback in sglang

kinza99:
### Checklist Before Starting

- [x] Searched for similar PR(s).
- [x] Checked PR Title format
  - In format of: [modules] type: Title
  - modules are in `fsdp, megatron, sglang, vllm, rollout, trainer, ci, training_utils, recipe, hardware, deployment, ray, worker, single_controller, misc, perf, model, algo, env, tool, ckpt, doc, data`
  - type is in `feat, fix, refactor, chore, test`
  - can involve multiple modules, seperated by `,` or space, like `[megatron, fsdp, doc] feat: xxx`

### What does this PR do?

> Implements multi-turn interaction system for reinforcement learning training, enabling dynamic conversational feedback and iterative problem-solving scenarios with extensible agent-based architecture.

### Test

> Comprehensive testing has been performed including:
> - **Unit Tests**: Complete test suite in `tests/interactions/test_gsm8k_interaction.py` with 15+ test cases covering:
>   - Initialization and configuration validation
>   - Interaction lifecycle (start → generate → calculate → finalize)
>   - Correct/incorrect answer handling with GSM8K scoring integration
>   - Edge cases (empty messages, malformed content, concurrent sessions)
>   - Resource cleanup and error handling
> - **Integration Tests**: SGLang rollout integration in `tests/workers/rollout/test_sglang_async_rollout_w_interaction.py`:
>   - Multi-GPU distributed testing (requires 2+ GPUs)
>   - FSDP model sharding with SGLang inference engine
>   - End-to-end interaction workflow with `interaction_kwargs` parameter passing
>   - Comparison with HuggingFace baseline for output validation
> - **Real-world Validation**: Training script testing with Qwen2.5-0.5B model demonstrating:
>   - GRPO algorithm integration with interaction-based rewards
>   - Multi-turn conversation handling in production training loops

### High-Level Design

> **Multi-Turn Interaction System Architecture:**
> 
> The system introduces a flexible, async-based interaction framework designed for RL training scenarios with the following key components:
> 
> 1. **BaseInteraction Class**: Core abstraction layer providing async interface for interaction agents
> 2. **Instance Management**: Stateful session management with unique instance IDs for concurrent interactions
> 3. **SGLang Integration**: Seamless integration with SGLang rollout system for multi-turn conversations
> 4. **Configuration-Driven Loading**: Dynamic agent loading via YAML configuration files
[...]

PR #2124, opened 2025-06-20
[rollout] feat: add agent loop

wuxibin89:
### Checklist Before Starting

- [ ] Checked PR Title format

### What does this PR do?

Add AgentLoopBase and AgentLoopManager for agentic rollout.

### Test

### High-Level Design

### Specific Changes

> List the specific changes.

### API

### Usage Example

```python
```

### DIFF
--- a/verl/agent/agent_loop.py
+++ b/verl/agent/agent_loop.py
@@ -26,6 +26,7 @@ import torch
 from cachetools import LRUCache
 from omegaconf import DictConfig
 from openai.types.chat.chat_completion import ChatCompletion
+from pydantic import BaseModel
 from tensordict import TensorDict
 from transformers import AutoTokenizer
 
@@ -117,6 +118,15 @@ class AsyncLLMServerManager:
             await session.close()
 
 
+class AgentLoopOutput(BaseModel):
+    """Agent loop output."""
+
+    prompt_ids: List[int]
+    response_ids: List[int]
+    response_mask: List[int]
+    num_turns: int
+
+
 class AgentLoopBase(ABC):
     """An agent loop takes a input message, chat with OpenAI compatible LLM server and interact with various environments."""
 
@@ -133,7 +143,7 @@ class AgentLoopBase(ABC):
         self.tokenizer = tokenizer
 
     @abstractmethod
-    async def run(self, messages: List[Dict[str, Any]], sampling_params: Dict[str, Any]) -> List[List[Dict[str, Any]]]:
+    async def run(self, messages: List[Dict[str, Any]], sampling_params: Dict[str, Any]) -> List[Dict[str, Any]]:
         """Run agent loop to interact with LLM server and environment.
 
         Args:
@@ -141,21 +151,18 @@ class AgentLoopBase(ABC):
             sampling_params (Dict[str, Any]): LLM sampling params.
 
         Returns:
-            List[List[Dict[str, Any]]]: List of `n` choices messages.
+            List[Dict[str, Any]]]: List of multi-turn messages.
         """
         raise NotImplementedError
 
-    async def tokenize(self, messages: List[Dict[str, str]]) -> Dict[str, List]:
+    async def tokenize(self, messages: List[Dict[str, str]]) -> AgentLoopOutput:
         """Tokenize messages to ids.
 
         Args:
             messages: multi-turn messages
 
         Returns:
-            Dict[str, List[int]]:
-            - prompt_ids: prompt ids
-            - response_ids: response ids
-            - response_mask: response mask
+            AgentLoopOutput: tokenized output
         """
         raise NotImplementedError
 
@@ -199,26 +206,23 @@ class AgentLoopWorker:
         if batch.meta_info.get("validate", False):
             sampling_params["top_p"] = config.val_kwargs.top_p
             sampling_params["temperature"] = config.val_kwargs.temperature
-            sampling_params["n"] = 1
 
+        n = 1 if batch.meta_info.get("validate", False) else config.n
         tasks = []
-        for agent_name, messages in zip(batch.non_tensor_batch["agent_name"], batch.non_tensor_batch["raw_prompt"]):
-            tasks.append(asyncio.create_task(self._run_agent_loop(agent_name, messages, sampling_params)))
-        result = await asyncio.gather(*tasks)
-        outputs = [choice for choices in result for choice in choices]
+        agent_names = batch.non_tensor_batch["agent_name"].repeat(n, axis=0)
+        raw_prompts = batch.non_tensor_batch["raw_prompt"].repeat(n, axis=0)
+        for agent_name, messages in zip(agent_names, raw_prompts):
+            tasks.append(asyncio.create_task(self._run_agent_loop(agent_name, messages.tolist(), sampling_params)))
+        outputs = await asyncio.gather(*tasks)
 
         output = self._postprocess(outputs)
         return output
 
-    async def _run_agent_loop(self, agent_name: str, messages: List[Dict[str, Any]], sampling_params: Dict[str, Any]) -> List[Dict[str, List[int]]]:
+    async def _run_agent_loop(self, agent_name: str, messages: List[Dict[str, Any]], sampling_params: Dict[str, Any]) -> List[AgentLoopOutput]:
         agent_loop_class = self.get_agent_loop_class(agent_name)
         agent_loop = agent_loop_class(self.config, self.server_manager, self.tokenizer)
-        choices = await agent_loop.run(messages, sampling_params)
-        tasks = []
-        for choice in choices:
-            tasks.append(asyncio.create_task(agent_loop.tokenize(choice)))
-        outputs = await asyncio.gather(*tasks)
-        return outputs
+        messages = await agent_loop.run(messages, sampling_params)
+        return await agent_loop.tokenize(messages)
 
     def get_agent_loop_class(self, agent_name: str) -> Type[AgentLoopBase]:
         # TODO: add tool agent registrary
@@ -231,7 +235,7 @@ class AgentLoopWorker:
             return ToolAgentLoop
         raise ValueError(f"Unknown agent_name: {agent_name}")
 
-    def _postprocess(self, inputs: List[Dict[str, List]]) -> DataProto:
+    def _postprocess(self, inputs: List[AgentLoopOutput]) -> DataProto:
         # NOTE: consistent with batch version of generate_sequences in vllm_rollout_spmd.py
         # prompts: left pad
         # responses: right pad
@@ -242,7 +246,7 @@ class AgentLoopWorker:
         # prompts
         self.tokenizer.padding_side = "left"
         outputs = self.tokenizer.pad(
-            [{"input_ids": input["prompt_ids"]} for input in inputs],
+            [{"input_ids": input.prompt_ids} for input in inputs],
             padding="max_length",
             max_length=self.config.actor_rollout_ref.rollout.prompt_length,
             return_tensors="pt",
@@ -253,7 +257,7 @@ class AgentLoopWorker:
         # responses
         self.tokenizer.padding_side = "right"
         outputs = self.tokenizer.pad(
-            [{"input_ids": input["response_ids"]} for input in inputs],
+            [{"input_ids": input.response_ids} for input in inputs],
             padding="max_length",
             max_length=self.config.actor_rollout_ref.rollout.response_length,
             return_tensors="pt",
@@ -263,7 +267,7 @@ class AgentLoopWorker:
 
         # response_mask
         outputs = self.tokenizer.pad(
-            [{"input_ids": input["response_mask"]} for input in inputs],
+            [{"input_ids": input.response_mask} for input in inputs],
             padding="max_length",
             max_length=self.config.actor_rollout_ref.rollout.response_length,
             return_tensors="pt",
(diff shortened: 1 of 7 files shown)


vermouth1992:
@ccclyu @SwordFaith Could you help review this design. Basically, it converts recursive llm interactions to loop-based design to make writing complex agent loop more easily. The original idea is from @zw0610

vermouth1992:
@casper-hansen Could you please also take a look and give some comments? Thanks.

zw0610:
@vermouth1992 This is a compromised version of what we've discussed. I would still suggest move the agent loop of out the AsyncLLMServer, and manipulate it directly on driver process.

vermouth1992:
> @vermouth1992 This is a compromised version of what we've discussed. I would still suggest move the agent loop of out the AsyncLLMServer, and manipulate it directly on driver process.

It runs on another ray actor. If we remove the ray actor, it directly runs on the driver process

zw0610:
> > @vermouth1992 This is a compromised version of what we've discussed. I would still suggest move the agent loop of out the AsyncLLMServer, and manipulate it directly on driver process.
> 
> It runs on another ray actor. If we remove the ray actor, it directly runs on the driver process

My concern lays more on the flexibility of debugging and customizing agent loop. Moreover, if we consider agent loops that consumes much more local resources, transforming them into distributed ray actors will release the resource pressure on either the ray actor you mentioned under the this implementation or driver process.

casper-hansen:
My feedback below.

1. Code Quality Improvement: Agent loop effectively refactors ChatCompletionScheduler by replacing recursion with a while loop, making the implementation simpler for the end-user.
2. Missing Extensibility: Agent loop lacks the extensibility that ChatCompletionScheduler provides via `config.multi_turn.completion_callback`. Adding similar functionality would make this more useful.
3. Development Experience Issue (Agree with @zw0610): Both implementations suffer from slow iteration cycles due to Ray dependency. I've implemented a private interoperable CompletionCallback that works with both training and inference, allowing direct calls to o4-mini or Qwen3 without booting veRL (which takes minutes).

I hope the feedback is useful. Point 3 is a general challenge that I think can be overcome in a future PR (scope is too big), so I would suggest adding it to the roadmap.

vermouth1992:
> My feedback below.
> 
> 1. Code Quality Improvement: Agent loop effectively refactors ChatCompletionScheduler by replacing recursion with a while loop, making the implementation simpler for the end-user.
> 2. Missing Extensibility: Agent loop lacks the extensibility that ChatCompletionScheduler provides via `config.multi_turn.completion_callback`. Adding similar functionality would make this more useful.
> 3. Development Experience Issue (Agree with @zw0610): Both implementations suffer from slow iteration cycles due to Ray dependency. I've implemented a private interoperable CompletionCallback that works with both training and inference, allowing direct calls to o4-mini or Qwen3 without booting veRL (which takes minutes).
> 
> I hope the feedback is useful. Point 3 is a general challenge that I think can be overcome in a future PR (scope is too big), so I would suggest adding it to the roadmap.

Thanks for your feedback! For number 2, I guess we can provide a registration mechanism for agent loop so that you can define by your self outside verl codebase? The main motivation of loop-based agent is for easy implementation when the agent interacts with an environment (e.g., pokemon game). Just wonder do you still prefer recursive style or loop style? For number 3, I agree that it requires substantial optimization in order to make it fast.

verl/agent/agent_loop.py
```diff
@@ -0,0 +1,390 @@
...
+        Args:
+            config (DictConfig): YAML config.
+            server_addresses (List[str]): OpenAI compatible LLM server addresses.
+            max_cache_size (int, optional): max cache size for request_id to address mapping. Defaults to 10000.
+        """
+        self.config = config
+        self.server_addresses = server_addresses
+
+        # Least requests load balancing
+        self.weighted_addresses = [[0, address] for address in server_addresses]
```
SwordFaith: least request load balancing may also results to shard equal data size to each dp group if without rate limit or running requests aware back press mechanism.

verl/agent/agent_loop.py
```diff
@@ -0,0 +1,390 @@
...
+        heapq.heapify(self.weighted_addresses)
+
+        # LRU cache to map request_id to address
+        self.request_id_to_address = LRUCache(maxsize=max_cache_size)
```
SwordFaith: If it possible to use radix-aware or prefix-aware routing hybrid with least requests and max running reqs aware.

verl/agent/agent_loop.py
```diff
@@ -0,0 +1,390 @@
...
+            self.weighted_addresses[0][0] += 1
+            heapq.heapreplace(self.weighted_addresses, self.weighted_addresses[0])
+
+        # use new request_id to avoid duplicate request_id problem
+        request_id = uuid4().hex
+        self.request_id_to_address[request_id] = address
+
+        return await self._chat_completions(address, request_id, messages=messages, tools=tools, **sampling_params, **kwargs)
+
+    async def _chat_completions(self, address: str, request_id: str, **chat_complete_request) -> ChatCompletion:
```
SwordFaith: So we finally choose to keep http server routing rather than ray based ?

verl/agent/agent_loop.py
```diff
@@ -0,0 +1,390 @@
...
+class AgentLoopOutput(BaseModel):
+    """Agent loop output."""
+
+    prompt_ids: List[int]
+    response_ids: List[int]
+    response_mask: List[int]
+    num_turns: int
+
+
+class AgentLoopBase(ABC):
```
SwordFaith: At least we need abort here. I wonder the main difference between agent loop and async rollout request is a normal class or a data class work with rollout or server+tool registry, why we need yet another interaction with env impl here ?

verl/agent/agent_loop.py
```diff
@@ -0,0 +1,390 @@
...
+                "position_ids": position_ids,  # [bsz, prompt_length + response_length]
+            },
+            batch_size=len(input_ids),
+        )
+
+        num_turns = np.array([input.num_turns for input in inputs], dtype=np.int32)
+        return DataProto(batch=batch, non_tensor_batch={"__num_turns__": num_turns})
+
+
+class AgentLoopManager:
```
SwordFaith: It will be better to consider this as replay buffer or rollout manager, which is the coordinator for partial rollout.

SwordFaith (commented):
Looking forward to discuss further.

SwordFaith (commented):
Looking forward to discuss further.

verl/agent/agent_loop.py
vermouth1992: Did you observe a performance diff?

verl/agent/agent_loop.py
vermouth1992: There is a fundamental difference between the two programming abstractions. In real world agent applications, rollout is just a server. The agent receives a prompt and starts to query LLMs for instructions in an agent loop. The tool calling is implemented inside the agent loop because there is no way to implement in the server (e.g., OpenAI models). The agent loop abstraction makes existing agent system easily pluggable, and not just restricted to tools (e.g., agentless or predefined workflows)

verl/agent/agent_loop.py
SwordFaith: It was suffered with concurrency when with sgl-router, and make Zilin Zhu switch to http2 in several weeks before. But may be ok because we only use between each server. 
Another consideration is resource affinity inside message list e.g. tool create resource like image in each turn. 
Using ray will help on auto-affinity scheduling, using http will require user equip with s3 like oss service and cause many network traffic during multi-turn generation. I guess support ray-based routing can improve user experience with less infrastructure requirements.
Using http-based routing can more easily porting with existing open source openai compatible server impl, and keep verl simplier, which is also an advantage.

casper-hansen:
> For number 2, I guess we can provide a registration mechanism for agent loop so that you can define by your self outside verl codebase?

Yes, allow users to define a custom agent loop easily.

> Just wonder do you still prefer recursive style or loop style?

Loop style is preferred. It’s easier to implement your own loop than it is to implement with recursion. Also, looking at the code, loop style is easier to transfer to inference after training.

verl/agent/agent_loop.py
SwordFaith: For predefined workflow, in my opinion, we should constraint our discussion about shared context / seperate context, hybrid model / single model. Of course, current loop abstraction with http server can make it easier to port existing workflow, but using messages list rather than dict of message list, seems can't support seperate context case as well. Using current AsyncLLMServerManager seems still suffer to add yet another model which not include in current verl training. In #1630 we can support user train single node in a hybrid workflow, no matter it depends on which model for other node. For e2e agent shared model and shared context, current impl works fine. A key challenge is how to train seperate context and single model workflow in verl. Which If there is an agentic loop, I want it to accomplish.

vermouth1992: Let me take a close look at

verl/agent/agent_loop.py
wuxibin89: > So we finally choose to keep http server routing rather than ray based ?

The message protocol `chat_complete_request ` between client and server should be [OpenAI compatible](https://platform.openai.com/docs/api-reference/chat/create), but the transport protocol can either be http or ray remote call. So there will be 2 _chat_completions implementation, the second one is in develop
- _chat_completions_http
- _chat_completions_ray

verl/agent/agent_loop.py
wuxibin89: Yes, we need a more server pressure aware load balancing strategy, @chenhaiq is investigating it.

verl/agent/agent_loop.py
wuxibin89: With collecting server performance metrics(e.g current requests, gpu cache usage, cpu cache usage, etc), we can implement better routing strategy.

wuxibin89:
> My feedback below.
> 
> 1. Code Quality Improvement: Agent loop effectively refactors ChatCompletionScheduler by replacing recursion with a while loop, making the implementation simpler for the end-user.
> 2. Missing Extensibility: Agent loop lacks the extensibility that ChatCompletionScheduler provides via `config.multi_turn.completion_callback`. Adding similar functionality would make this more useful.
> 3. Development Experience Issue (Agree with @zw0610): Both implementations suffer from slow iteration cycles due to Ray dependency. I've implemented a private interoperable CompletionCallback that works with both training and inference, allowing direct calls to o4-mini or Qwen3 without booting veRL (which takes minutes).
> 
> I hope the feedback is useful. Point 3 is a general challenge that I think can be overcome in a future PR (scope is too big), so I would suggest adding it to the roadmap.

@casper-hansen Thanks for you feedback! For point 2, I'm adding AgentLoop registry, allowing user to register their custom AgentLoop.

casper-hansen:
@wuxibin89 Two minor issues to consider:
1. When using Qwen3, you have to specify `enable_thinking=False` in the `_postprocess()` function because it will otherwise add empty `<think> </think>` blocks in your messages since it's enabled by default. In the ChatCompletionScheduler, I am able to override this because we have direct access to modifying `_postprocess()`. How would we achieve the same thing for the Agent loop?
2. Implementing streaming for inference with the recursive callbacks is pretty challenging. The Agent loop seems easier because we can modify it to yield messages as they appear. Consider if there should be a streaming implementation for this in an example.

wuxibin89:
@casper-hansen 
1. Due to the problem describe in: https://github.com/volcengine/verl/pull/1946#issuecomment-2995459715 , I have replace `/v1/chat_completions` to `AsyncLLM.generate`, instead of passing text messages, now we pass prompt_ids and response_ids directly between client and server, so there's no need to tokenize messages in `_postprocess` any more.
2. Yes, streaming responses is more easy to implement in AgentLoop, but I don't get why we need streaming responses, because we need `<EOS>` token or other stop tokens to extract tool calls.

casper-hansen:
> 1. Due to the problem describe in: [[rollout] feat: Add LangGraph chat scheduler for agent training with tool use support #1946 (comment)](https://github.com/volcengine/verl/pull/1946#issuecomment-2995459715) , I have replace `/v1/chat_completions` to `AsyncLLM.generate`, instead of passing text messages, now we pass prompt_ids and response_ids directly between client and server, so there's no need to tokenize messages in `_postprocess` any more.

I see now, great update on this! The removal of tokenization is reassuring and should be compatible with all models now.

> 2. Yes, streaming responses is more easy to implement in AgentLoop, but I don't get why we need streaming responses, because we need `<EOS>` token or other stop tokens to extract tool calls.

After I train, I want to move to an inference system where I can see the intermediate messages from the trained agent as this can take 5-10 minutes. This loop makes it easier to create an API in front of the trained model.

verl/experimental/agent_loop/single_agent_loop.py
```diff
@@ -0,0 +1,50 @@
+# Copyright 2024 Bytedance Ltd. and/or its affiliates
```
vermouth1992, line 1: Can we rename this file to single_turn_agent_loop? `single_agent_loop` is confusing.

vermouth1992 (approved):

casper-hansen:
@wuxibin89 I see this has been merged now. Will you followup with a PR on extensibility so we can use our CustomAgentLoop by registering it through the config?

waleko:
@wuxibin89 Does this change mean we can no longer use a chat scheduler or completion_callback? That was quite helpful for our use case. Is there any way to restore or support this functionality again?

yinzhangyue:
Perfect Work! Is there any guide how to use the tool_agent_loop in the Training? Thx~

leoleoasd:
Does `AgentLoopManager` completely replace async sglang rollout manager when in async mode? If it's true, does this
https://verl.readthedocs.io/en/latest/sglang_multiturn/multiturn.html
need to be updated?

PR #2125, opened 2025-06-20
[megatron] feat: Support of dist checkpoint

ETOgaosion:
### Checklist Before Starting

- [ ] Checked PR Title format

### What does this PR do?

Support of dist checkpoint in saving, loading and model merger.

### Test

Algorithm:

image

### High-Level Design

### Specific Changes

> List the specific changes.

### API

### Usage Example

```python
```

### DIFF
--- a/.github/workflows/e2e_ppo_trainer.yml
+++ b/.github/workflows/e2e_ppo_trainer.yml
@@ -159,8 +159,8 @@ jobs:
         run: |
           export EXP_NAME="qwen2.5-0.5b-function-reward-minimal"
           ls checkpoints/verl-test/${EXP_NAME}/global_step_1/actor
-          cat checkpoints/verl-test/${EXP_NAME}/global_step_1/actor/config.json
-          python3 -m verl.model_merger merge --backend fsdp --local_dir checkpoints/verl-test/${EXP_NAME}/global_step_1/actor/ --target_dir checkpoints/verl-test/${EXP_NAME}/global_step_1/actor/hf
+          cat checkpoints/verl-test/${EXP_NAME}/global_step_1/actor/huggingface/config.json
+          python3 -m verl.model_merger merge --backend fsdp --local_dir checkpoints/verl-test/${EXP_NAME}/global_step_1/actor/ --target_dir checkpoints/verl-test/${EXP_NAME}/global_step_1/actor/huggingface
       - name: Running GSM8K E2E training tests on 8 L20 GPUs with grpo lora using function rm with use_shm and layered_summon with fsdp2
         run: |
           ray stop --force


.github/workflows/e2e_ppo_trainer.yml
```diff
@@ -109,15 +109,15 @@ jobs:
       - name: Test merging FSDP checkpoints (Qwen Actor) 
         run: |
           exp_name="qwen2.5-0.5b-function-reward-minimal-fsdp8"
-          python scripts/model_merger.py test --backend fsdp --local_dir checkpoints/verl-test/${exp_name}/global_step_1/actor --test_hf_dir checkpoints/verl-test/${exp_name}/global_step_1/actor/huggingface
+          python -m scripts.model_merger test --backend fsdp --local_dir checkpoints/verl-test/${exp_name}/global_step_1/actor --test_hf_dir checkpoints/verl-test/${exp_name}/global_step_1/actor/huggingface
```
eric-haibin-lin: this means that `scripts` become a pkg namespace. if you want to enable `python -m ` i would suggest using `python -m verl.xxx` instead.

verl/utils/checkpoint/fsdp_checkpoint_manager.py
```diff
@@ -62,7 +62,7 @@ def __init__(
         optimizer: Optional[torch.optim.Optimizer] = None,
         lr_scheduler: Optional[torch.optim.lr_scheduler.LRScheduler] = None,
         processing_class: Union[PreTrainedTokenizer, ProcessorMixin] = None,
-        checkpoint_contents: DictConfig = None,
```
eric-haibin-lin, line 65: is this a breaking change? if renaming is desired, we should record in https://github.com/volcengine/verl/discussions/1902

verl/utils/checkpoint/checkpoint_manager.py
```diff
@@ -49,10 +49,11 @@ def __init__(
         optimizer: torch.optim.Optimizer,
         lr_scheduler: torch.optim.lr_scheduler.LRScheduler = None,
         processing_class: Union[PreTrainedTokenizer, ProcessorMixin] = None,
-        checkpoint_contents: DictConfig = None,
+        checkpoint_config: DictConfig = None,
```
eric-haibin-lin, line 50: In the next PR, can we use a dataclass instead of DictConfig for `checkpoint_config`?

Yangruipis:
Excellent work! This is exactly what i am looking forward to.  

however, for dpsk 671B, load all weigts into one node is slow and will easily cause RAM OOM. I have implemented a distributed merger for large models, as every single node only processes a partition of model pipeline, and save as HF model separately(will not ensure every .safetensors file the same size, but, emmm, it's fast), if you are interested, i'm willing to contribute based on you PR~

eric-haibin-lin:
> Excellent work! This is exactly what i am looking forward to.
> 
> however, for dpsk 671B, load all weigts into one node is slow and will easily cause RAM OOM. I have implemented a distributed merger for large models, as every single node only processes a partition of model pipeline, and save as HF model separately(will not ensure every .safetensors file the same size, but, emmm, it's fast), if you are interested, i'm willing to contribute based on you PR~

@Yangruipis that sounds great, looking forward to your contribution

ETOgaosion:
@Yangruipis Maybe we can split into 2 PRs, I left some spaces for multi-node merger extensions, Thanks a lot for contribution~

verl/utils/checkpoint/fsdp_checkpoint_manager.py
ETOgaosion: Actually, from namimg issue, yes, since from now checkpoint config can include some other configurations besides contents.

verl/utils/checkpoint/checkpoint_manager.py
ETOgaosion: OK, maybe define a checkpoint config class.

verl/model_merger/base_model_merger.py
```diff
@@ -0,0 +1,264 @@
...
+            print(f"Saving processor to {self.config.target_dir}")
+            processor.save_pretrained(self.config.target_dir)
+        if tokenizer is not None:
+            print(f"Saving tokenizer to {self.config.target_dir}")
+            tokenizer.save_pretrained(self.config.target_dir)
+
+    def upload_to_huggingface(self):
+        from huggingface_hub import HfApi
+
+        api = HfApi()
```
ccclyu, line 248: we might need to handle authentication issues if `hf_token` is not provided.

verl/model_merger/base_model_merger.py
```diff
@@ -0,0 +1,264 @@
...
+        api.create_repo(repo_id=self.config.hf_upload_path, private=self.config.private, exist_ok=True)
+        api.upload_folder(folder_path=self.config.target_dir, repo_id=self.config.hf_upload_path, repo_type="model")
```
ccclyu: some issues like network etc may cause upload folder failure. can you add some try/except to catch any potential failures?

verl/model_merger/megatron_model_merger.py
```diff
@@ -0,0 +1,369 @@
...
+        os.environ["WORLD_SIZE"] = "1"
+        os.environ["MASTER_ADDR"] = "localhost"
+        os.environ["MASTER_PORT"] = "12355"
+        torch.distributed.init_process_group(get_nccl_backend())
+        mpu.initialize_model_parallel(
+            tensor_model_parallel_size=1,
+            virtual_pipeline_model_parallel_size=None,
+            context_parallel_size=1,
+            expert_model_parallel_size=1,
+        )
```
ccclyu, line 100: add one comment for single rank distributed set up and multiple rank loading is currently not supported. in the future consider warp environment into a reusable method.

verl/model_merger/megatron_model_merger.py
```diff
@@ -0,0 +1,369 @@
...
+                q_lst.append(q)
+                k_lst.append(k)
+                v_lst.append(v)
+
+            q = torch.cat(q_lst, dim=0)
+            k = torch.cat(k_lst, dim=0)
+            v = torch.cat(v_lst, dim=0)
+            return [q, k, v]
+        else:
+            return tensor
```
ccclyu: return `tensor` is not consistent with return type `list[torch.Tensor]`

verl/model_merger/megatron_model_merger.py
```diff
@@ -0,0 +1,369 @@
...
+            up_lst.append(up)
+            gate = torch.cat(gate_lst, dim=0)
+            up = torch.cat(up_lst, dim=0)
+            return [gate, up]
+        elif "self_attention.linear_qkv." in key and "layer_norm" not in key:
+            # if the tensor is qkv, for each param on tp, split into q, k, v
+            # concat q, k, v separately.
+            q_lst = []
+            k_lst = []
+            v_lst = []
```
ccclyu: will `q_lst, k_lst, v_lst = [], [], [] ` be better?

verl/model_merger/megatron_model_merger.py
```diff
@@ -0,0 +1,369 @@
...
+                ]
+                q, k, v = chunk.split(split_size)
+                q_lst.append(q)
+                k_lst.append(k)
+                v_lst.append(v)
+
+            q = torch.cat(q_lst, dim=0)
+            k = torch.cat(k_lst, dim=0)
+            v = torch.cat(v_lst, dim=0)
+            return [q, k, v]
```
ccclyu: return [torch.cat(q_lst, dim=0), torch.cat(k_lst, dim=0), torch.cat(v_lst, dim=0)]

ETOgaosion: Thanks a lot, all fixed~

verl/model_merger/megatron_model_merger.py
ETOgaosion: Thanks a lot, all fixed~

verl/model_merger/megatron_model_merger.py
ETOgaosion: Thanks a lot, all fixed~

verl/model_merger/megatron_model_merger.py
ETOgaosion: I added a comment here.

verl/model_merger/base_model_merger.py
ETOgaosion: Would you check my try-catch is what you expected?

verl/model_merger/base_model_merger.py
ETOgaosion: GPT says that error code 401 means auth error, does it make sense?

eric-haibin-lin:
@dataproblems could u help review as well, thanks

verl/model_merger/base_model_merger.py
```diff
@@ -0,0 +1,292 @@
...
+        config (ModelMergerConfig): The configuration object passed during initialization.
+        hf_model_config_path (str): Path to the HuggingFace model configuration files.
+        model_config (PretrainedConfig): Loaded HuggingFace model configuration.
+    """
+
+    def __init__(self, config: ModelMergerConfig):
+        self.config = config
+        self.hf_model_config_path = config.hf_model_config_path
+
+        if config.hf_model_path:
```
dataproblems: I see that we are overriding the `self.hf_model_config_path` if the user has provided the deprecated `hf_model_path`, should we add a note to say that we're overriding that value with what's provided under `hf_model_path` ?

verl/model_merger/fsdp_model_merger.py
```diff
@@ -0,0 +1,264 @@
...
+            backend="fsdp",
+            local_dir="path/to/fsdp/checkpoints",
+            target_dir="path/to/output"
+        )
+        merger = FSDPModelMerger(config)
+        merger.merge_and_save()
+        ```
+    """
+
+    def _get_world_size(self) -> int:
```
dataproblems, line 92: is it possible to also ask the users to include the world size in the config.json? We can check the file name pattern for sure but it would also be easier if the user can provide an explicit value. What do you think?

verl/model_merger/fsdp_model_merger.py
```diff
@@ -0,0 +1,264 @@
...
+                raise ValueError("test_hf_dir must be provided for test operation")
+            self._test_state_dict(merged_state_dict)
+        elif self.config.operation == "merge":
+            self.save_hf_model_and_tokenizer(merged_state_dict)
+            if self.config.hf_upload:
+                self.upload_to_huggingface()
+        else:
+            raise ValueError(f"Unknown operation: {self.config.operation}")
+
+    def _test_state_dict(self, state_dict: dict[str, torch.Tensor]):
```
dataproblems: nit: Let's call it `validate_state_dict` or something equivalent instead of `test_state_dict` to avoid the commonly used test_method_name for unit tests? But optional - not a blocker for release.

verl/model_merger/megatron_model_merger.py
```diff
@@ -0,0 +1,364 @@
...
+        for skip_key in skip_checking_keys:
+            if skip_key in key:
+                print(f"skip checking key {key}")
+                return
+
+        # Exclude extra state keys
+        if not key.startswith("decoder"):
+            raise ValueError(f"Invalid key {key} in Megatron state_dict. Expected keys to start with 'decoder' in TransformerLayer.")
+
+    def _split_tensors(self, key: str, tensor: torch.Tensor, config: PretrainedConfig, is_value_model: bool = False) -> list[torch.Tensor]:
```
dataproblems, line 220: This is a great start, what do you think about offering `MergeSplitStrategy` for Megatron layers - something that users can provide for additional layers as well? I'm not sure if the layers we have right now in the params mapping are an exhaustive list of the layers, but at the same time we can offer the users to provide their implementations for the ones we have not covered in this implementation. 

Something where you'd have `default_splitting_strategies: List[MergeSplitStrategy] = [# The current layer implementations here]`

And then you let users provide additional implementations should their models contain some other type of layer? What do you think?

verl/utils/checkpoint/megatron_checkpoint_manager.py
```diff
@@ -254,24 +347,55 @@ def save_checkpoint(self, local_path: str, hdfs_path: str = None, global_step: i
...
+
+        # Synchronize all async save requests
+        if not self.checkpoint_config.async_save:
+            assert async_save_request is None, "Async save request should be None when not using async save."
+            torch.distributed.barrier()
 
-        # Save Model
-        if self.should_save_model and mpu.get_data_parallel_rank() == 0:
-            state_dicts = []
+        def finalize_save_fn():
```
dataproblems: In the future,  we can make the checkpoint manager file system agnostic by offering an interface and letting the users provide the output path.

verl/model_merger/base_model_merger.py
ETOgaosion: Because in many other scripts we use `hf_model_path` as the huggingface model directory, I think we shall deprecate the `local_dir` and only use `hf_model_path` here for the continuity? I refactor the default-saved `tokenizer` and `hf_config` path to exactly huggingface model path, so the usage of `hf_model_path` is OK.

ETOgaosion: cc please @0x404

verl/model_merger/fsdp_model_merger.py
ETOgaosion: Yeah, nice idea. Since we directly save a TransformerConfig in megatron side, maybe we shall save a FSDP config as a runtime reference?

verl/model_merger/fsdp_model_merger.py
ETOgaosion: Get! Have changed.

verl/model_merger/base_model_merger.py
0x404: Hi @ETOgaosion, I remember the reason we keep `hf_model_path` before was because during model merge we needed to read the original huggingface config and tokenizer, but currently the checkpoints saved by verl already include hf config and other information, so there is no need for users to provide an extra `hf_model_path`. We keep this argument for backward compatibility since some old checkpoints may not contain hf config.

> should we add a note to say that we're overriding that value with what's provided under hf_model_path 

in the `model_merger.py` we will give a deprecation warning if users provide this arg and we will override `hf_model_config_path`:

https://github.com/volcengine/verl/blob/2a6212385a3078ac83abede0d8eec3fc6cacb288/scripts/model_merger.py#L102-L104

I think we can still keep this arg and give a notice to the users,  or we can remove this arg in this PR and thus deprecate this arg completely?

verl/model_merger/fsdp_model_merger.py
ETOgaosion: Currently I only save the FSDP version and distributed info, maybe we leave other configurations in the next PR?
Also cc @0x404 ~

0x404: Nice idea! I can draft a PR recently if this is not too urgent. With this config, we may also support resume from a checkpoint with different machine configuration. For example, a checkpoint is saved using one node and 8 GPUs, and now we are going to resume this checkpoint on 2 nodes and 16 total GPUs.

verl/model_merger/megatron_model_merger.py
ETOgaosion: I get what you mean. This is super extendable, I also think that it is useful as there can be all kinds of operators other than qkv or gate_up needed to be fixing from megatron to huggingface. Here we leave a TODO for designing such pattern for extendable usage.

verl/utils/checkpoint/megatron_checkpoint_manager.py
ETOgaosion: Actually I'm not quite clear about the `agnostic` and `output path` here. Do you mean that users shall offer paths to save all components?

verl/model_merger/base_model_merger.py
ETOgaosion: I get what you mean, can you help review the logic I changed in 

https://github.com/ETOgaosion/verl/blob/ef15cde94289e948b7863027985bc376678bd618/verl/utils/checkpoint/fsdp_checkpoint_manager.py#L227-L300

Now no matter whether `should_save_hf_model`, we save huggingface config and tokenizer in the `huggingface` path, and when saving huggingface model, we also place it in `huggingface` path and only save the model to avoid repeated works.

And we now can deprecate `hf_model_path` in the saving process as Megatron now can directly use the `${local_dir}/huggingface" to store the hf config and tokenizers.

Maybe we shall clarify that in the doc.

0x404: LGTM

verl/model_merger/fsdp_model_merger.py
dataproblems: Sounds good! Thank you! Could you cc me on that PR when you open it?

verl/model_merger/megatron_model_merger.py
dataproblems: I see. Yeah, I think extending it would be quite important as many users would have the need for it and we should minimize the amount of custom code / changes required to make that happen! Thanks for leaving a TODO.

verl/utils/checkpoint/megatron_checkpoint_manager.py
dataproblems: I was suggesting that we separate the storage logic from the checkpoint manager in to some `PersistenceManager` implementation, that way, if the user specifies a local path, we store it locally, an hdfs path, we use HDFSPersistenceManager, an s3 path, we use an S3PersistenceManager... etc. Does that help clarify things?

ETOgaosion: Ah, I see. Nice idea. It's more extendable to hold a `PersistenceManager` backend.

vermouth1992 (approved):

verl/model_merger/base_model_merger.py
qinwang2333: The previously trained model config and tokenizer are stored under _actor_, not _huggingface_ path. The latest code forces hf_model_config_path=/path/huggingface, and I cannot modify it through parameters _hf_model_path_.
image

0x404: @qinwang2333 Yes, this PR breaks the backward compatibility. I think a workaround is to create a `huggingface` directory under this actor dir, and move all files except *.pt to the `huggingface` directory.
```bash
mkdir huggingface
mv $(ls | grep -v "\.pt$\|huggingface") huggingface/
```
But the current model merger will require a `fsdp_config.json` under the `actor` dir, which includes two fields `FSDP_version` and `world_size`, so you can create a simple file in this dir. For example:
```json
{
    "FSDP_version": 1,
    "world_size": 2
}
```
Then, you can merge this checkpoint:
```bash
python -m verl.model_merger merge --backend fsdp --local_dir your/path/actor --target_dir your/path/actor/huggingface
```

Hi @ETOgaosion, do you think we should add backward compatibility into current merger to avoid users getting errors when merging old checkpoints? (for example, https://github.com/volcengine/verl/issues/2252) For example, if `fsdp_config.json` cannot be found, then use the previous method of getting word size based on file names.

ETOgaosion: @0x404 Can you help still keep the `model_merger` in the `scripts` folder in your PR #2251 ? But with `legacy_model_merger` and explain this in the doc please? This may work as a backward compatibility?

ETOgaosion: > Can you help still keep the `model_merger` in the `scripts` folder in your PR #2251 ? But with `legacy_model_merger` and explain this in the doc please? This may work as a backward compatibility?

@qinwang2333 Is it acceptable to you?

qinwang2333: Sounds good! Thank you!

PR #2132, opened 2025-06-21
[algo] feat:  A more Efficient implementation of GAE

Clement25:
### What does this PR do?

Find the inefficiency in current GAE implementation using "for loop".  Reformulate it as conv1d and implement it.

### Checklist Before Describing the Details

- [x] PR title is in the format of: `[modules] type: Title`
  - modules are in `fsdp, megatron, sglang, vllm, rollout, trainer, ci, training_utils, recipe, hardware, deployment, ray, worker, single_controller, misc, perf, model, algo, env, tool, ckpt, doc, data, cfg`
  - multiple modules are seperated by `,` or space, such as `[megatron, fsdp, doc] feat: xxx`

### Test
On single A100 GPU. The settings for tensor size are below:
```python
batch_size = 4, seq_len = 128
```
Result (first 10 elements):
```python
# print(advantages[0,:10])
Using conv: 
tensor([-0.1061,  0.3279,  0.7977,  0.6851,  0.8231,  0.7337,  1.0778, -0.1774,
         0.0533,  0.2212], device='cuda:0')
Original implementation: 
tensor([-0.1061,  0.3279,  0.7977,  0.6851,  0.8231,  0.7337,  1.0778, -0.1774,
         0.0533,  0.2212], device='cuda:0')
```

Time (print in terminal, 100 runs)
```
Using conv: 0.2193572111427784
Original implementation: 0.8121056966483593
```

### High-Level Design

### Specific Changes

Only in 1 file. See commit details below.

### API and Usage Example

```python
```

- [x] Apply [pre-commit checks](https://github.com/volcengine/verl?tab=readme-ov-file#code-linting-and-formatting): `pre-commit run --show-diff-on-failure --color=always --all-files`

### DIFF
--- a/verl/trainer/ppo/core_algos.py
+++ b/verl/trainer/ppo/core_algos.py
@@ -186,7 +186,7 @@ def compute_gae_advantage_return(
         device = values.device
         values_pad = pad(values, (0, 1), mode="constant", value=0.0)
         delta = token_level_rewards + gamma * values_pad[:, 1:] - values
-        delta_padded = pad(delta, (0, delta.shape[-1] - 1)).unsqueeze(1)    # (batch, 1, gen_len)
+        delta_padded = pad(delta, (0, delta.shape[-1] - 1)).unsqueeze(1)  # (batch, 1, gen_len)
         decay = (lam * gamma) ** torch.arange(gen_len, device=device)[None, None, :]  # (1, 1, gen_len)
 
         advantages = torch.conv1d(delta_padded, decay).squeeze(1)
--- a/verl/trainer/ppo/core_algos.py
+++ b/verl/trainer/ppo/core_algos.py
@@ -25,6 +25,7 @@ from enum import Enum
 
 import numpy as np
 import torch
+from torch.nn.functional import pad
 
 import verl.utils.torch_functional as verl_F
 
@@ -180,16 +181,15 @@ def compute_gae_advantage_return(
 
     """
     with torch.no_grad():
-        lastgaelam = 0
-        advantages_reversed = []
         gen_len = token_level_rewards.shape[-1]
 
-        for t in reversed(range(gen_len)):
-            nextvalues = values[:, t + 1] if t < gen_len - 1 else 0.0
-            delta = token_level_rewards[:, t] + gamma * nextvalues - values[:, t]
-            lastgaelam = delta + gamma * lam * lastgaelam
-            advantages_reversed.append(lastgaelam)
-        advantages = torch.stack(advantages_reversed[::-1], dim=1)
+        device = values.device
+        values_pad = pad(values, (0, 1), mode="constant", value=0.0)
+        delta = token_level_rewards + gamma * values_pad[:, 1:] - values
+        delta_padded = pad(delta, (0, delta.shape[-1] - 1)).unsqueeze(1)    # (batch, 1, gen_len)
+        decay = (lam * gamma) ** torch.arange(gen_len, device=device)[None, None, :]  # (1, 1, gen_len)
+
+        advantages = torch.conv1d(delta_padded, decay).squeeze(1)
 
         returns = advantages + values
         advantages = verl_F.masked_whiten(advantages, response_mask)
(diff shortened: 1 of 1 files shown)


tongyx361:
Could you add the reference information of the naive implementation, e.g.,

```python
    """Compute advantages and returns using GAE (Generalized Advantage Estimation, https://arxiv.org/abs/1506.02438)

    This implementation is efficient but not that easy to understand.
    For easier understanding, please refer to a simpler equivalent implementation in
    https://github.com/huggingface/trl/blob/559a99f053e5bda892dd19e9283831be5b90c46b/trl/trainer/ppo_trainer.py#L524-L532

    Args:
        token_level_rewards: `(torch.Tensor)`
            shape is (bs, response_length)
        values: `(torch.Tensor)`
            shape is (bs, response_length)
        response_mask: `(torch.Tensor)`
            shape is (bs, response_length). [EOS] mask. The token after [EOS] have mask zero.
        gamma is `(float)`
            discounted factor used in RL
        lam: `(float)`
            lambda value in GAE.

    Returns:
        advantages: `(torch.Tensor)`
            shape: (bs, response_length)
        Returns: `(torch.Tensor)`
            shape: (bs, response_length)
    """
```

Clement25:
Added. 

> Could you add the reference information of the naive implementation, e.g.,
> 
> ```python
>     """Compute advantages and returns using GAE (Generalized Advantage Estimation, https://arxiv.org/abs/1506.02438)
> 
>     This implementation is efficient but not that easy to understand.
>     For easier understanding, please refer to a simpler equivalent implementation in
>     https://github.com/huggingface/trl/blob/559a99f053e5bda892dd19e9283831be5b90c46b/trl/trainer/ppo_trainer.py#L524-L532
> 
>     Args:
>         token_level_rewards: `(torch.Tensor)`
>             shape is (bs, response_length)
>         values: `(torch.Tensor)`
>             shape is (bs, response_length)
>         response_mask: `(torch.Tensor)`
>             shape is (bs, response_length). [EOS] mask. The token after [EOS] have mask zero.
>         gamma is `(float)`
>             discounted factor used in RL
>         lam: `(float)`
>             lambda value in GAE.
> 
>     Returns:
>         advantages: `(torch.Tensor)`
>             shape: (bs, response_length)
>         Returns: `(torch.Tensor)`
>             shape: (bs, response_length)
>     """
> ```

vermouth1992:
We probably don't want to directly modify the original implementation because it is much more easier to modify. Also, the fraction of time gae computation contributed to the overall training time is basically negligible. I do have experience in class DeepRL, this optimization is critic, however, training/rollout llm is soooo time consuming and makes this optimization less necessary

Clement25:
Shall we add that as an alternative implementation somewhere? Since I believe my implementation could match the basic formula of GAE better than current one.

eric-haibin-lin (commented):
Would u mind adding a unit test that asserts the old and new implementations are consistent, in tests/? Thanks!

eric-haibin-lin:
How much time in percentage does this function take in a step?

Clement25:
> Would u mind adding a unit test that asserts the old and new implementations are consistent, in tests/? Thanks!

@eric-haibin-lin 
Sure. But which precision type should I write for? I tested both float32 and float16 for both implementations. If treating float32 as the ground truth, then after converting to float16 it seems that conv shows better precision than for loop. (see the screenshot below)

Screenshot 2025-06-24 at 1 00 06 AM

verl/trainer/ppo/core_algos.py
```diff
@@ -180,16 +183,15 @@ def compute_gae_advantage_return(
...
-            lastgaelam = delta + gamma * lam * lastgaelam
-            advantages_reversed.append(lastgaelam)
-        advantages = torch.stack(advantages_reversed[::-1], dim=1)
+        device = values.device
+        values_pad = pad(values, (0, 1), mode="constant", value=0.0)
+        delta = token_level_rewards + gamma * values_pad[:, 1:] - values
+        delta_padded = pad(delta, (0, delta.shape[-1] - 1)).unsqueeze(1)  # (batch, 1, gen_len)
+        decay = (lam * gamma) ** torch.arange(gen_len, dtype=values.dtype, device=device)[None, None, :]  # (1, 1, gen_len)
+
+        advantages = torch.conv1d(delta_padded, decay).squeeze(1)
```
wuxibin89, line 194: Since GAE computation time is trivial in a whole step, the readability is more important. Moreover, `torch.conv1d` is hard to skip observation tokens in multi-turn rollout.

PR #2142, opened 2025-06-22
[trainer, worker] fix: handle case when reference policy worker is not…

xingyunjohn1:
[trainer, worker] fix: handle case when reference policy worker is not available in fit function (when using LoRA)

### What does this PR do?

This PR adds a safeguard to ensure that the absence of a reference policy worker does not cause an error during training. Specifically, it checks whether the reference policy worker is available before attempting to use it in the `fit` function. This scenario commonly arises when using LoRA to update the actor.

The same logic has already been implemented in `verl/trainer/ppo/ray_trainer.py` at lines 1068–1071, and this PR applies a consistent fix elsewhere in the codebase to maintain robustness and prevent runtime issues.

https://github.com/volcengine/verl/blob/9b7bb69ea3165b691cc908d7f3f2f14c4a65a59e/verl/trainer/ppo/ray_trainer.py#L1068C29-L1071C97

### Checklist Before Starting

- [x] Search for similar PRs. 
*For multi-node multi-GPU training tasks where models with different roles are distributed across different nodes, there are relatively few reported issues, and no relevant issues have been found so far.*

### Test

### API and Usage Example

> None. This is just a simple bug fix involving a few lines of code.

```python
```

### High-Level Design

> This is just a simple bug fix involving a few lines of code.

### Specific Changes

> This is just a simple bug fix involving a few lines of code.

> [!IMPORTANT]


examples/split_placement/split_monkey_patch.py
```diff
@@ -127,7 +127,10 @@ def fit(self):
                 if self.use_reference_policy:
                     # compute reference log_prob
                     with _timer("ref", timing_raw):
-                        ref_log_prob = self.ref_policy_wg.compute_ref_log_prob(batch)
+                        if not self.ref_policy_wg:
```
eric-haibin-lin: did you mean `if self.ref_policy_wg`?

xingyunjohn1: Hi, thank you for pointing out my initial mistake and for the suggestion!

Actually, the intended logic is different. After checking my local codebase and testing, the correct condition is `if not self.ref_in_actor:`. This change keeps the logic consistent with what is used in `verl/trainer/ppo/ray_trainer.py`.

Here is the reasoning:

* The flag `self.ref_in_actor` is designed to be set to `True` when the actor is being trained with LoRA (specifically, when `lora_rank > 0`).
* The specific use case I tested was for GRPO, where the actor is indeed trained using LoRA.
* In this scenario, `self.ref_in_actor` becomes `True`, and the code within the `if not self.ref_in_actor:` block should be skipped. My latest commit correctly implements this behavior.

Thanks again for your careful review! It helped me fix this bug.

PR #2146, opened 2025-06-22
[tool] feat: Add support for tools that generate multimodal data

nanjiangwill:
### What does this PR do?

This PR adds support for tools to create and return multimodal data (images and videos) during rollout. It enhances the framework to properly handle multimodal inputs that are dynamically generated by tools during multi-turn conversations.

### Key Features

- Tools can now return images and videos as part of their response
- Added support for processing multimodal inputs in the rollout system
- Introduced a new configuration option `return_multi_modal_inputs` to control how multimodal inputs are processed
- Updated documentation with examples of how to implement tools that generate multimodal data

### API and Usage Example

```python
async def execute(self, ...) -> Tuple[str | Dict[str, Any], float, dict]:
    # Process images or videos
    from verl.utils.dataset.vision_utils import process_image, process_video

    img1 = process_image(img1)
    video1 = process_video(video1)

    # Return multimodal data
    return {"image": [img1, ...], "video": [video1, ...], "text": "..."}, 0, {}
```

In your dataset config, set:
```yaml
data:
  return_multi_modal_inputs: False
```

### Specific Changes

- Enhanced `AsyncRolloutRequest` to handle multimodal data from tools
- Updated `add_tool_response_messages` to process multimodal content
- Added documentation for multimodal tool support in the RST docs
- Fixed configuration in example YAML files
- Added proper handling of multimodal inputs in the rollout system

- [X] Add `[BREAKING]` to the PR title `description` if it breaks any API.
- [X] New CI unit test(s) are added to cover the code path.
- [X] Rely on existing unit tests on CI that covers the code path.


verl/workers/rollout/schemas.py
```diff
@@ -240,19 +246,48 @@ def add_assistant_message(
...
 
-    def add_tool_response_messages(self, processing_class: Union[PreTrainedTokenizer, PreTrainedTokenizerFast, ProcessorMixin], contents: list[str]) -> None:
+    def add_tool_response_messages(self, processing_class: Union[PreTrainedTokenizer, PreTrainedTokenizerFast, ProcessorMixin], contents: list[str | Dict[str, Any]]) -> None:
         if not contents:
             return
-
-        self.messages.extend([Message(role="tool", content=content) for content in contents])
+        # We also handle the case when tool returns image
+        # We require the processing of the image and video to be done at tool.execute() level
+        delta_multi_modal_data = {key: [] for key in self.multi_modal_keys}
```
SwordFaith, line 426: May be we need test for delta multi_modal_data part of logic ?

SwordFaith (commented):
We can include a two-step end-to-end training test and a delta multi-modal input unit test, if necessary, to safeguard our code through CI. Additional reasoning for image-based multi-turn tasks, such as Deep Eyes, can be incorporated in future PRs.

verl/workers/rollout/schemas.py
nanjiangwill: Thanks for reviewing! I already tested this on internal tasks, willing to add [DeepEyes](https://github.com/Visual-Agent/DeepEyes) in future with @xieck13

verl/workers/rollout/schemas.py
```diff
@@ -147,6 +150,11 @@ def initialize_request(cls, values):
                 # Only log the warning to avoid truncating in the middle of generation prompt. Consider raising an error for this case in the future.
                 logger.warning(f"Prompt {values['batch_data_id']} length {len(values['input_ids'])} greater than max_prompt_len {max_prompt_len} after applied chat template with tools.")
 
+            multi_modal_inputs = dict(tokenization_dict_with_prompt)
+            multi_modal_inputs.pop("input_ids", None)
+            multi_modal_inputs.pop("attention_mask", None)
+            values["multi_modal_inputs"] = multi_modal_inputs
+
         values["prompt_ids"], values["prompt_attention_mask"] = values["input_ids"], values["attention_mask"]
         values["position_ids"] = values["prompt_position_ids"] = compute_position_id_with_mask(torch.tensor(values["attention_mask"])).tolist()
```
mantle2048: Great PR! Should we also consider the Qwen-VL 3D RoPE position encoding, as seen in[ this commit](https://github.com/volcengine/verl/commit/d50c6cd66e77cf8c2afb531a1e99f8e882196f34#diff-b51188afaba947f32a00ea8b9c0458bf4d80965dd2c23365692316fb7eb5e498)?

nanjiangwill: thank you! could you elaborate a bit why that will affect the multi_modal_inputs logic? my understanding is that it does not affect any position_ids logic rn

vermouth1992:
Could you please resolve conflict?

verl/workers/rollout/schemas.py
mantle2048: Hi, I understand that in `schemas.py`, the `position_ids` are currently all calculated based on the `compute_position_id_with_mask` function. However, the vision tool will return an image, which means it requires 3D RoPE for the correct `position_id` calculation, much like how it's done in verl's [`rl_dataset.py`.](http://github.com/volcengine/verl/blob/main/verl/utils/dataset/rl_dataset.py#L263-L275)

Please correct me if my understanding is wrong~

nanjiangwill: good catch! let me add special handling if the model is qwen2.5 vl. thanks!

nanjiangwill: thanks for the explanation! i added the handling for qwen2.5vl! would love to have you checkout the latest change!

one small thing is that since in `AsyncRolloutRequest` we process everything in list not tensor, so it's a bit messy to update position_ids. i am working with the author who originally wrote this logic to figure out a better way to process. let me know if you have any suggestion!

nanjiangwill: working on changing from list to tensor, will push the change soon

verl/workers/rollout/schemas.py
```diff
@@ -73,7 +73,7 @@ class AsyncRolloutRequestStateEnum(str, Enum):
 class TokenizationSanityCheckModeEnum(str, Enum):
     """The enum for tokenization sanity check mode."""
 
-    OFF = "off"
```
eric-haibin-lin, line 76: Is it necessary to introduce backward incompatible changes? In general it would hurt users experience when doing the upgrade

verl/trainer/config/ppo_trainer.yaml
```diff
@@ -87,6 +87,9 @@ data:
 
     # The name of the dataset class within the specified file.
     name: null
+  
+  # Whether to process multi-modal inputs. Set to False if rollout generates new multi-modal inputs.
+  process_multi_modal_inputs_in_dataset: True
```
eric-haibin-lin: 1. since this field is already under `data.xxx`, it's better to drop the `_in_dataset` suffix to keep it succinct
2. from the code this field does not control the process logic, it looks like the process_image/video and processor functions are still invoked. maybe rename it to `return_multi_modal_inputs` or adjust the code logic?

verl/workers/rollout/schemas.py
nanjiangwill: yes this is needed, there is a parsing bug when using `off`. i will add more warnings

verl/trainer/config/ppo_trainer.yaml
nanjiangwill: thanks! will update this

verl/workers/rollout/schemas.py
```diff
@@ -460,35 +644,40 @@ def finalize(
...
+            if isinstance(self.position_ids[0], list) and isinstance(self.position_ids[0][0], int)
+            else self.position_ids[: self.max_model_len]
+        )
         self.loss_mask = self.loss_mask[: self.max_model_len]
         self.response_ids = self.input_ids[len(self.prompt_ids) :][: self.max_response_len]
         self.response_attention_mask = self.attention_mask[len(self.prompt_attention_mask) :][: self.max_response_len]
-        self.response_position_ids = self.position_ids[len(self.prompt_position_ids) :][: self.max_response_len]
+        self.response_position_ids = (
+            [
+                position_ids[len(self.prompt_position_ids) :][: self.max_response_len]
```
mantle2048: There seems to be a potential issue here. `self.prompt_position_ids` is also a list of position IDs with the shape [3, seq_len]. Taking `len(self.prompt_position_ids)` will always result in 3. We should use `self.prompt_position_ids[0]` here instead~?

nanjiangwill: thanks for the catch, previously i was not going to write this as it is very easy to write bug in list operation. but my concern is it will be too many changes in a single pr. I will create a new pr to refactor this to tensor like operation later. so we dont need to handle all these if else statements

nanjiangwill: having them in list is not ideal to keep code base clean. will make the new pr right after this pr gets merged

mantle2048: Awesome, thanks again for the great contribution! I learned a lot from your implementation.

nanjiangwill: https://github.com/volcengine/verl/pull/2362

this will be the new pr, it is just a draft for now.

eric-haibin-lin:
My concern is addressed

zhaochenyang20 (approved):

PR #1833, opened 2025-06-04
feat: Add MiniCPM-o 2.6 support

RanchiZhao:
### Checklist Before Starting

- [x] Search for similar PR(s).

### What does this PR do?

Add MiniCPM-o 2.6 multimodal model support to VERL framework for vision-language RL training.

### Specific Changes

- **verl/third_party/vllm/vllm_v_0_5_4/dtensor_weight_loaders.py**: Add MiniCPM-o weight loading
- **verl/third_party/vllm/vllm_v_0_6_3/dtensor_weight_loaders.py**: Add MiniCPM-o weight loading  
- **verl/utils/dataset/vision_utils.py**: Enhanced vision data processing
- **verl/utils/dataset/rl_dataset.py**: Multimodal dataset support
- **verl/utils/flops_counter.py**: Vision model FLOPS calculation
- **verl/workers/actor/dp_actor.py**: Multimodal model compatibility
- **examples/grpo_trainer/run_minicpmo2_6.sh**: Complete training example

### Usage Example

```bash
# Train MiniCPM-o 2.6 with GRPO
bash examples/grpo_trainer/run_minicpmo2_6.sh
```

### Test

- [x] Local testing with MiniCPM-o 2.6 on geo3k dataset
- [x] Verified weight loading for both vLLM versions
- [x] Training script validation

### Checklist Before Submitting

- [x] Read the Contribute Guide
- [ ] Apply pre-commit checks (will fix in follow-up if needed)
- [ ] No breaking API changes
- [ ] Documentation updates (if needed)
- [x] Rely on existing unit tests

PR #2178, opened 2025-06-24
[model] feat: Add MiniCPM-o 2.6 support

hiyouga:
@RanchiZhao We reverted the previous commit because we find a critical bug in #1833 https://github.com/volcengine/verl/pull/1833/files#diff-e06e73d3a7775a502b7aea91103e7911f6597eb48e4b898db558766cdd41daf9R119-R121

The indentation size of the if-else block is incorrect

### DIFF
--- a/verl/workers/actor/dp_actor.py
+++ b/verl/workers/actor/dp_actor.py
@@ -116,25 +116,26 @@ class DataParallelPPOActor(BasePPOActor):
                 else:
                     position_ids_rmpad = index_first_axis(rearrange(position_ids.unsqueeze(-1), "b s ... -> (b s) ..."), indices).transpose(0, 1)
 
-            if "multi_modal_inputs" in micro_batch:
-                # MiniCPM-o specific processing for image bounds and pixel values
-                if "image_bound" in multi_modal_inputs:
-                    # Adjust image bounds based on left padding and cumulative sequence lengths
-                    # This is necessary for MiniCPM-o's vision-language alignment
-                    left_padding_length = torch.argmax(attention_mask, dim=1)
-                    image_bounds = []
-                    for i in range(len(multi_modal_inputs["image_bound"])):
-                        image_bound = multi_modal_inputs["image_bound"][i].to(left_padding_length.device) - left_padding_length[i] + cu_seqlens[i]
-                        image_bounds.append(image_bound)
-                    multi_modal_inputs["image_bound"] = [torch.vstack(image_bounds)]
-                    # Flatten pixel values list for MiniCPM-o processing
-                    pixel_values = []
-                    for i in range(len(multi_modal_inputs["pixel_values"])):
-                        pixel_values.extend([p for p in multi_modal_inputs["pixel_values"][i]])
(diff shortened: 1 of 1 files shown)


verl/utils/dataset/rl_dataset.py
```diff
@@ -115,6 +115,12 @@ def __init__(
         self.need_tools_kwargs = config.get("need_tools_kwargs", False)
         self.filter_prompts = config.get("filter_prompts", True)
         self.serialize_dataset = False
+        self.processor_type = self.processor.image_processor.__class__.__name__ if self.processor is not None else None
```
vermouth1992, line 118: Is there any way to make this generalize?

hiyouga: For text LLMs, there is no processor, so the processor type is None. For VLMs, the processor type is a string that matches the image processor's class name. So I think this already covers most scenarios.

vermouth1992: Sorry, I mean line 119, where says if self.processor_type == "MiniCPMVImageProcessor":

vermouth1992: It seems it's model specific

RanchiZhao:
> @RanchiZhao We reverted the previous commit because we find a critical bug in #1833 https://github.com/volcengine/verl/pull/1833/files#diff-e06e73d3a7775a502b7aea91103e7911f6597eb48e4b898db558766cdd41daf9R119-R121
> 
> The indentation size of the if-else block is incorrect

@hiyouga Nice catch! And thanks for the quick fix—much appreciated.

verl/utils/dataset/rl_dataset.py
RanchiZhao: @vermouth1992 Valid observation! This mirror Qwen2VL approach (line 272) because:
• MiniCPMO requires custom visual feature handling
• Current abstraction works better at the pipeline level
Appreciate you flagging this - happy to refine if you see opportunities

nanjiangwill:
Hi @hiyouga, thanks for your contribution and the effort you put into this PR. I have a couple of concerns that I believe should be addressed to ensure the codebase remains clean and generalizable:

1. `verl/workers/actor/dp_actor.py:L94` does not properly handle the general case where `multi_modal_inputs['pixel_values']` is a list. Since tool-generated images are typically lists, adding a special-case logic here seems unnecessary. It would be cleaner to first convert to tensor and apply a consistent logic. Similarly, the logic starting at line 119 should also be adjusted to handle the general case. I will suggest to remove those special logics and convert the type back to what you desire in `MiniCPM` specific logic instead of in the underlying general `dp_actor.py`

2. In `rl_dataset.py`, the current structure introduces a separate code block for new model. This pattern will likely lead to a bloated and less maintainable codebase over time if we add special logic for each new model. I recommend refactoring `MiniCPM` code to be more modular and general-purpose to fit into current code base.

I really appreciate the work so far—just wanted to raise these points to help improve long-term maintainability. Let me know what you think.

hiyouga:
@nanjiangwill we'll fix this problem ASAP

PR #2239, opened 2025-06-27
[cfg] fix: pickleing error in multiprocessing in the reward_fn

none0663:
### What does this PR do?

> Fix "Can't pickle local object" error when using custom reward functions in multiprocessing

When I import `compute_score` from my own Python file, the `file_path` is not None and it will call the `wrapped_fn ` method.
You can view the relevant code at the following：
https://github.com/volcengine/verl/blob/e96f0fbf44a50a302ddea0118e76d40cc9aa4fff/verl/trainer/ppo/reward.py#L25-L57

When  I set` reward_model.reward_manager=prime`, it will call the `ProcessPoolExecutor` to use asyncio, leading to the error:

`"Can't pickle local object 'get_custom_reward_fn.<locals>.wrapped_fn'".`

Root Cause:

The nested closure `wrapped_fn` created in `get_custom_reward_fn() `is unpicklable for the following reasons:

- Python's `pickle` cannot serialize local functions (functions defined inside another function).

- The closure dynamically captures variables (`raw_fn` and `reward_kwargs`) from its outer scope.

This breaks multiprocessing workflows (e.g., `SubprocVecEnv`, `multiprocessing.Pool)` that rely on pickling.

### DIFF
--- a/verl/trainer/ppo/reward.py
+++ b/verl/trainer/ppo/reward.py
@@ -21,10 +21,12 @@ import ray
 from verl import DataProto
 from verl.utils.reward_score import default_compute_score
 
+
 # This function is used to merge additional keyword arguments with the original function's arguments.
 def _call_with_kwargs(raw_fn, extra_kwargs, *args, **kwargs):
     merged_kwargs = {**kwargs, **extra_kwargs}
     return raw_fn(*args, **merged_kwargs)
+
 
 def get_custom_reward_fn(config):
     import importlib.util
@@ -55,11 +57,7 @@ def get_custom_reward_fn(config):
 
     reward_kwargs = dict(reward_fn_config.get("reward_kwargs", {}))
 
-    return partial(
-        _call_with_kwargs,
-        raw_fn,
-        reward_kwargs
-    )
+    return partial(_call_with_kwargs, raw_fn, reward_kwargs)
 
 
 def load_reward_manager(config, tokenizer, num_examine, **reward_kwargs):
--- a/verl/trainer/ppo/reward.py
+++ b/verl/trainer/ppo/reward.py
@@ -57,8 +57,8 @@ def get_custom_reward_fn(config):
 
     return partial(
         _call_with_kwargs,
-        raw_fn=raw_fn,
-        extra_kwargs=reward_kwargs
+        raw_fn,
+        reward_kwargs
     )
 
 
--- a/verl/trainer/ppo/reward.py
+++ b/verl/trainer/ppo/reward.py
@@ -21,6 +21,10 @@ import ray
 from verl import DataProto
 from verl.utils.reward_score import default_compute_score
 
+# This function is used to merge additional keyword arguments with the original function's arguments.
+def _call_with_kwargs(raw_fn, extra_kwargs, *args, **kwargs):
+    merged_kwargs = {**kwargs, **extra_kwargs}
+    return raw_fn(*args, **merged_kwargs)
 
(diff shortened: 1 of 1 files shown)


none0663:
`[e2e_ppo_trainer / e2e_ppo_trainer_sglang_vlm (pull_request)](https://github.com/volcengine/verl/actions/runs/15950713439/job/45015193954?pr=2239)Failing after 5m`

verl/trainer/ppo/reward.py
```diff
@@ -22,6 +22,12 @@
 from verl.utils.reward_score import default_compute_score
 
 
+# This function is used to merge additional keyword arguments with the original function's arguments.
+def _call_with_kwargs(raw_fn, extra_kwargs, *args, **kwargs):
```
tongyx361: What about putting the comment into docstring for better IDE hints?

```suggestion
def _call_with_kwargs(raw_fn, extra_kwargs, *args, **kwargs):
    """This function is used to merge additional keyword arguments with the original function's arguments."""
```

vadimkantorov:
Does it also fix the error `[Error] Task failed: Can't pickle <function compute_score at 0x7f23310b4ea0>: it's not the same object as custom_module.compute_score, completion` from:
- https://github.com/volcengine/verl/pull/1296#issuecomment-3002092357
- https://github.com/volcengine/verl/pull/1296#issuecomment-3017376023

It seems that somehow ray is messing with pickling/unpickling

yuekaizhang:
> * Fix pickle error in prime reward manager when prime in main_ppo.py #1296 (comment)

@vadimkantorov I have met the issue even using the latest code. ThreadPoolExecutor works for me.

vadimkantorov:
> I have met the issue even using the latest code.

Could you please clarify about `ThreadPoolExecutor`? Would you have any suggestions on how to fix `[Error] Task failed: Can't pickle <function compute_score at 0x7f23310b4ea0>: it's not the same object as custom_module.compute_score, completion ...`?

yuekaizhang:
> > I have met the issue even using the latest code.
> 
> Could you please clarify about `ThreadPoolExecutor`? Would you have any suggestions on how to fix `[Error] Task failed: Can't pickle <function compute_score at 0x7f23310b4ea0>: it's not the same object as custom_module.compute_score, completion ...`?

https://github.com/volcengine/verl/compare/main...yuekaizhang:verl:thread It could fix your issue in my case.

yuekaizhang:
@eric-haibin-lin  What do you think what if we add another option to switch between  ThreadPoolExecutor and ProcessPoolExecutor in prime.py? 

In my case, I set up a remote LLM reward server, which makes compute_score not a compute intensive job anymore.

vadimkantorov:
@eric-haibin-lin @vermouth1992  we'll try again with the newly-merged PR (might be that module import must also be done in some other ray process), but yeah, probably a knob to select threadpoolexecutor would be great

Especially given that compute_score often is HTTP request to some server and does not do any heavy compute itself, so more lightweight ThreadPoolExecutor should make sense

vadimkantorov:
@eric-haibin-lin @vermouth1992 In Trinity they implemented a simple generic mechanism which can load all Python files in a given directory (`plugins/` by default). Trinity uses it for allowing the user provide/register their own reward function or workflow class. I've tested it and it works well

Could it be useful to simply upstream this mechanism / helper function from Trinity?

Relevant:
- https://github.com/modelscope/Trinity-RFT/issues/240#issuecomment-3247441005

PR #2258, opened 2025-06-29
[sglang] feat: Repeat sampling parameter n into requests of GRPO in SGLang

zhaochenyang20:
### What does this PR do?

For a large-scale GRPO with a huge sampling parameter n (say 128 or more), we take the sampling times n out to directly duplicate the requests.

This is beneficial if our n is relatively large. But we need to check the order of the input and output requests.

We create unique UIDs for each prompt to enable grouping in GRPO advantage computation

1. Each prompt gets a unique UID that is repeated n times along with the prompt
2. After generation, responses are aligned with prompts using UID matching
3. In GRPO advantage computation, UID groups responses from the same prompt

Note that we only enable this for sglang `_req_level`, i.e., in ma ulti-turn setting GRPO.

### Checklist Before Starting

### Test

Will add experiments on the change of rollout time.

### API and Usage Example

N/A

### High-Level Design

N/A

### Specific Changes

> List the specific changes.

> [!IMPORTANT]

### DIFF
--- a/verl/trainer/ppo/ray_trainer.py
+++ b/verl/trainer/ppo/ray_trainer.py
@@ -1074,7 +1074,10 @@ class RayPPOTrainer:
         last_val_metrics = None
         self.max_steps_duration = 0
 
-        repeat_sampling_sglang_grpo = self.config.actor_rollout_ref.rollout.name == "sglang" and self.config.actor_rollout_ref.rollout.multi_turn.enable
+        repeat_sampling_sglang_grpo = (
+            self.config.actor_rollout_ref.rollout.name == "sglang"
+            and self.config.actor_rollout_ref.rollout.multi_turn.enable
+        )
 
         for epoch in range(self.config.trainer.total_epochs):
             for batch_dict in self.train_dataloader:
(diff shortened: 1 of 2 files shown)


verl/trainer/ppo/ray_trainer.py
```diff
@@ -942,6 +942,8 @@ def fit(self):
         last_val_metrics = None
         self.max_steps_duration = 0
 
+        repeat_sampling_sglang_grpo = self.config.actor_rollout_ref.rollout.name == "sglang" and self.config.actor_rollout_ref.rollout.multi_turn.enable
```
vermouth1992: Why this feature relates to multiturn?

vermouth1992:
LGTM. Please resolve conflict

verl/trainer/ppo/ray_trainer.py
zhaochenyang20: If not multi-turn, the rollout will use `_batch_level_generate_senquence`, which already does what we want to do.

zhaochenyang20: So no need to repeat n in requests. But I shall find a way to unify them for SGLang later.

vermouth1992: Yes. We need to unify all the places to use the same logic to repeat at driver instead of passing n inside.

zhaochenyang20: Sure. I discussed with Haibin, will unify two ways of rollout in the following PRs:

https://github.com/volcengine/verl/issues/2309

2258 could be merged!

PR #2259, opened 2025-06-29
[algo] fix: correctly aggregate kl metrics in PPO actor

0x404:
### What does this PR do?

This PR fix an issue in dp_actor where `actor/kl_loss` and `actor/kl_coef` were being continuously overwritten during the micro-batch processing loop.

Previously, the long-lived `metrics` dictionary was updated directly, causing the value for these metrics to reflect only the final micro-batch of any given step, rather than an aggregation of all micro-batches within that step.

This change refactors the logic to align the collection of all metrics, now `kl_loss` is collected for each micro-batch, the same as other metrics like `pg_loss`.

> [!IMPORTANT]

### DIFF
--- a/verl/workers/actor/dp_actor.py
+++ b/verl/workers/actor/dp_actor.py
@@ -414,6 +414,8 @@ class DataParallelPPOActor(BasePPOActor):
                 self.actor_optimizer.zero_grad()
 
                 for data in micro_batches:
+                    micro_batch_metrics = {}
+
                     # Support all hardwares
                     if isinstance(data, DataProto):
                         data = {**data.batch.to(get_device_id()), **data.non_tensor_batch}
@@ -484,8 +486,8 @@ class DataParallelPPOActor(BasePPOActor):
                         kl_loss = agg_loss(loss_mat=kld, loss_mask=response_mask, loss_agg_mode=loss_agg_mode)
 
                         policy_loss = policy_loss + kl_loss * self.config.kl_loss_coef
-                        metrics["actor/kl_loss"] = kl_loss.detach().item()
-                        metrics["actor/kl_coef"] = self.config.kl_loss_coef
+                        micro_batch_metrics["actor/kl_loss"] = kl_loss.detach().item()
+                        micro_batch_metrics["actor/kl_coef"] = self.config.kl_loss_coef
 
                     if self.config.use_dynamic_bsz:
                         # relative to the dynamic bsz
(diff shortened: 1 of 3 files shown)


verl/workers/actor/dp_actor.py
```diff
@@ -414,6 +414,8 @@ def update_policy(self, data: DataProto):
                 self.actor_optimizer.zero_grad()
 
                 for data in micro_batches:
+                    micro_batch_metrics = {}
```
vermouth1992, line 417: Could you help modify dp_critic as well? Thanks.

0x404: Sure, done!

verl/workers/fsdp_workers.py
```diff
@@ -1123,9 +1123,9 @@ def update_critic(self, data: DataProto):
             estimated_flops, promised_flops = self.flops_counter.estimate_flops(global_num_tokens, delta_time)
             metrics["perf/mfu/critic"] = estimated_flops * self.config.ppo_epochs / promised_flops / self.world_size
 
-            self.critic_lr_scheduler.step()
             lr = self.critic_lr_scheduler.get_last_lr()[0]
             metrics["critic/lr"] = lr
+            self.critic_lr_scheduler.step()
```
0x404, line 1128: also fix an issue where previous recorded lr is actually the next step's lr instead of the current step. as https://github.com/volcengine/verl/pull/1463

vermouth1992:
One more thing, could you also fix the metrics aggregation in this PR like this https://github.com/volcengine/verl/pull/2144? In #2144, DataProto was modified to support this. As a general data structure, this is inappropriate. The reasonable approach is to perform allgather inside the worker, and aggregate on rank zero. Just like what typically torchrun programs did. Do you mind also fix this as well? Thanks!

0x404:
> One more thing, could you also fix the metrics aggregation in this PR like this #2144? In #2144, DataProto was modified to support this. As a general data structure, this is inappropriate. The reasonable approach is to perform allgather inside the worker, and aggregate on rank zero. Just like what typically torchrun programs did. Do you mind also fix this as well? Thanks!

Sure, I would like to fix this. Should I open a separate PR or push that fix into this?

vermouth1992:
Just push the fix into this PR is fine as it just fixes aggregation problem

verl/workers/fsdp_workers.py
```diff
@@ -628,11 +629,9 @@ def update_actor(self, data: DataProto):
...
             self.actor_lr_scheduler.step()
 
-            # TODO: here, we should return all metrics
-            output = DataProto(meta_info={"metrics": metrics})
-
-            output = self.ulysses_sharding_manager.postprocess_data(data=output)
-            output = output.to("cpu")
+            # Note: we should convert worker's metrics to DataProto's non_tensor_batch
+            # so that metrics from different workers can be correctly collected by worker group
+            worker_metrics = DataProto(non_tensor_batch=to_numpy_metrics(metrics))
```
0x404: Here we can return the worker's metrics as DataProto's non_tensor_batch, and utilize the worker group's collect function to allgather different workers' metrics from all workers, thereby avoiding the use of allgather for communication within workers.

vermouth1992: non_tensor_batch requires the batch size to be equal to the data's batch size. How can we ensure that the batch size of metrics is the same of the data?

0x404: In the workers' `update_xxx` (update_actor, update_critic), the worker only return metrics without batch data, thus the DataProto don't have batch_size.

0x404: Maybe a better solution is to add a generic field in DataProto, maybe with a name like `aggregated_data` or `aux_data`, which is not constrained by the batch_size, like `meta_info`, just a simple Dict[str, np.array]. The difference between this field and `meta_info` is that we will aggregate this field in worker group's collect_fn. This field can be used to aggregate metrics, and can also be used for aggregating other info. What do you think?

vermouth1992: I think create a field of aggreated_data makes sense. If this is the approach, then we may need to first introduce another PR, that
1. Create the field
2. Define how the behavior of this field in each DataProto API
3. Define the dispatch and collect_fn for each registered method
4. Write a testcase to protect the method

Then, we can use this field to implement auto metrics aggregation. We need to clearly define the aggregation function so that it can be customized.

0x404: OK, I will revert this PR to 3d8bf7a and open a separate PR.

PR #2308, opened 2025-07-01
[training_utils] feat: impl a metrics class for recording metrics and support worker sync [WIP]

0x404:
### What does this PR do?

### Checklist Before Starting

### Test

### API and Usage Example

```python
```

### High-Level Design

### Specific Changes

> List the specific changes.

> [!IMPORTANT]

### DIFF
--- a/verl/utils/metric/metric.py
+++ b/verl/utils/metric/metric.py
@@ -112,7 +112,7 @@ class Metrics:
         """
         name_lower = name.lower()
 
-        if any(word in name_lower for word in ("sum")):
+        if any(word in name_lower for word in ("sum",)):
             agg = MetricAggregationType.SUM
         elif any(word in name_lower for word in ("max", "peak")):
             agg = MetricAggregationType.MAX
@@ -122,7 +122,7 @@ class Metrics:
             # default to mean
             agg = MetricAggregationType.MEAN
 
-        if "lr" in name_lower:
+        if any(word in name_lower for word in ("lr", "learning_rate")):
             fmt = "{:.6f}"
         else:
             # default to 3 decimal places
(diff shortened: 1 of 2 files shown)


0x404:
Hi @vermouth1992, I found the approach we discussed in https://github.com/volcengine/verl/pull/2259#discussion_r2173753481 is somewhat tricky. We need more discussion to support `aggregated_data` for different dispatch_fn and collect_fn. And since we are going to deprecate `DataProto` in the future refactor, I think we shouldn't support `aggregated_data` for it anymore.

I think we could support a generic `Metrics` class, and we can set a config for each recorded metric, including the reduce type it should use, and currently also including the format it should use (this is to support console log, since currently each metric can only support the default `"{:.3f}"`, so learning rate always shows as 0.000). We can support synchronization and reduce between different ranks for this Metrics class, and can use this util to replace the currently used dict for recording metrics.

This PR is still work in progress. Do you think this approach is reasonable? It would be great to hear your insights.

verl/utils/metric/metric.py
```diff
@@ -0,0 +1,342 @@
+from collections import defaultdict
```
vermouth1992, line 1: How to we want to aggregate among dp?

vermouth1992: I guess we need a function to merge them? Actually, I guess allgather is a good choice because this is what typically done in pretraining.

0x404: Yes, I was thinking of creating a function `allgather_across_group(self, group)` for Metrics class, which would gather all other ranks' metrics objects on each dp rank, then merge them into a single Metrics through `Metrics.merge`.

What do you think? This PR is still a draft demo, if this makes sense, I will keep working on this.

PR #1020, opened 2025-04-11
Integrate AdaRFT (Adaptive Curriculum Reinforcement Finetuning) into veRL

maksimstw:
### Integrate AdaRFT (Adaptive Curriculum Reinforcement Finetuning) into veRL

📄 Paper: [Efficient Reinforcement Finetuning via Adaptive Curriculum Learning](https://arxiv.org/abs/2504.05520)  
🔗 Repo: https://github.com/uscnlp-lime/verl
🗂️ Data: [Datasets with Difficulty Scores for Curriculum Learning](https://huggingface.co/collections/lime-nlp/difficulty-estimation-math-datasets-67e85f875aeb74b76d763aef)

---

**Summary:**

This PR integrates **AdaRFT**, a lightweight curriculum learning extension designed for reinforcement finetuning (RFT). AdaRFT operates at the **data level**, dynamically sampling training examples based on difficulty without modifying:

- the reward function,
- model architecture, or
- the underlying RL algorithm.

As a result, AdaRFT is **fully compatible** with most RFT algorithms (e.g., PPO, GRPO, PRIME), and can be seamlessly integrated into existing RL pipelines.

---

**Key Features:**
- **Faster convergence and better performance:** Reduces training steps/time by up to 2x and achieves better performance
- **Data-level curriculum learning:** Adjusts training difficulty based on model performance on the fly.
- **Dynamic difficulty-based sampling:** Continuously updates the sampling distribution to keep training examples appropriately challenging.
- **No changes to RL internals:** Compatible with standard RFT algorithms out of the box.
- **Configurable scheduler:** Exposes hyperparameters (`η`, `α`, `β`) for controlling the curriculum dynamics.
- **Custom sampler**: introduces ```custom_sampler.py```, enabling users to design their own sampling strategies for training.

---

**Motivation:**

AdaRFT improves **sample and compute efficiency** by ensuring the model is neither overfitted on easy examples nor stuck on hard ones. The adaptive curriculum ensures smoother learning and faster convergence. Additionally, while there has been considerable research suggesting that curriculum learning can significantly boost RFT performance, no existing RFT codebase currently supports this feature. In addition to AdaRFT, this pull request also introduces ```custom_sampler.py```, enabling users to design their own sampling strategies for training. This addition expands verl’s utility, supporting more **dynamic and customizable** sampling techniques.

---
Let me know if you need any additional information about AdaRFT, or feel free to suggest a more appropriate way to integrate this feature into ```verl```.

PR #2314, opened 2025-07-01
[data] feat: add interface for user-defined curriculum sampler

frrad:
### What does this PR do?

This PR introduces a flexible interface that allows users to plug in their own Sampler implementations. This is particularly useful for advanced training strategies like curriculum learning, where the sampling policy evolves over time to progressively present the model with increasingly difficult tasks.

Curriculum learning can significantly accelerate training convergence and improve generalization, especially in complex domains. By decoupling the Sampler, users can implement task- or environment-specific curricula—for instance, starting with simpler examples and gradually incorporating harder ones, or adapting sampling based on the model’s competence.

### Checklist Before Starting

- [X] Search for similar PRs. Paste at least one query link here: ...

There have been previous attempts to add specific samplers for curriculum learning ( [search 1](https://github.com/volcengine/verl/pulls?q=is%3Apr+is%3Aopen+sampler) [search 2](https://github.com/volcengine/verl/pulls?q=is%3Apr+is%3Aopen+curriculum) ) but they commit to specific implementations of the curriculum.

This PR just adds an interface so that users can supply their own implementation. This approach was suggested in [this comment](https://github.com/volcengine/verl/pull/759/files#r2030220009).

### API and Usage Example

```bash
./examples/sglang_multiturn/run_qwen0.5b_gsm8k_multiturn_curriculum.sh
```

This is a run using the example sampler implementation. here is the diff of that script with an existing one in case we decide to omit the example runner in this PR.

```bash

diff ./examples/sglang_multiturn/run_qwen0.5b_gsm8k_multiturn_curriculum.sh ./examples/sglang_multiturn/run_qwen3-4b_gsm8k_multiturn.sh

15,17c15
<     data.curriculum.curriculum_class="RandomCurriculumSampler" \
<     data.curriculum.curriculum_class_path="verl.utils.dataset.curriculum_sampler" \
<     data.dataloader_num_workers=0 \
---
>     data.train_batch_size=256 \
20d17
<     data.train_batch_size=256 \
24c21
<     actor_rollout_ref.model.path=Qwen/Qwen2.5-0.5B-Instruct \
---
>     actor_rollout_ref.model.path=Qwen/Qwen3-4B \


```

 

### Specific Changes

This PR exposes a new interface so that users can implement their own `Sampler`. I also provide a trivial implementation of this interface - `RandomCurriculumSampler` as an example / for test purposes.

- [X] Apply [pre-commit checks](https://github.com/volcengine/verl?tab=readme-ov-file#code-linting-and-formatting): `pre-commit install && pre-commit run --all-files --show-diff-on-failure --color=always`
- [X] Add unit or end-to-end test(s) to [the CI workflow](https://github.com/volcengine/verl/tree/main/.github/workflows) to cover all the code. If not feasible, explain why: ...
- [X] Once your PR is ready for CI, send a message in [the `ci-request` channel](https://verl-project.slack.com/archives/C091TCESWB1) in [the `verl` Slack workspace](https://join.slack.com/t/verl-project/shared_invite/zt-3855yhg8g-CTkqXu~hKojPCmo7k_yXTQ).


Peng-YM:
Should the Sampler interface be aware of the current Trainer's global_steps, epochs, metrics, etc.?

frrad:
This seems useful to me. I am interested in feedback from the maintainers about the right interface. 

AdaRFT provides feedback to the curriculum sampler here

https://github.com/volcengine/verl/pull/1020/files#diff-cacab4f7bb28a5ef411968e425f32f5fc994c347715804af429c8c2c14303004L940-L941

our implementation does something similar. I think an interface like

```python
@abstractmethod
def update(self, batch:DataProto) -> None:
    pass
```

would make sense, and call it in as AdaRFT does if the `isinstance(sampler, AbstractSampler)`.

Happy to do this on this PR, or to iterate on the interface separately.

zhaochenyang20:
@frederrx could we modify the description of the PR to add the motivation, like what we expect to have in curriculum learning, thanks!

tests/utils/dataset/test_create_rl_sampler.py
```diff
@@ -0,0 +1,80 @@
...
+"""
+
+import pytest
+from omegaconf import OmegaConf
+from torch.utils.data import Dataset
+
+from verl.trainer.main_ppo import create_rl_sampler
+
+
+class FakeIncorrectSampler:
```
zhaochenyang20:
```python
class FakeIncorrectSampler:
"""A fake sampler class that does not adhere to the AbstractCurriculumSampler interface."""
    def __init__(self, data_source, data_config):
        pass
```

Also, I'd persoanlly call it `mock` rather than `fake` 🤔

verl/trainer/config/ppo_trainer.yaml
```diff
@@ -88,6 +91,15 @@ data:
     # The name of the dataset class within the specified file.
     name: null
 
+  # settings related to curriculum
```
zhaochenyang20: `# settings related to curriculum learning`

verl/trainer/config/ppo_trainer.yaml
```diff
@@ -88,6 +91,15 @@ data:
...
+  curriculum:
+
+    # the path to the module containing a curriculum class
```
zhaochenyang20: what's the definition of a `curriculum class`?

zhaochenyang20: Could we rename this like `curriculum sampler` or `curriculum sampler class`? `curriculum_sampler_class_path` like this

zhaochenyang20: I saw the file name in `dataset/curriculum_sampler.py` 🤔

verl/trainer/ppo/ray_trainer.py
```diff
@@ -561,10 +561,17 @@ def _create_dataloader(self, train_dataset, val_dataset, collate_fn, train_sampl
 
             collate_fn = default_collate_fn
 
+        num_workers = self.config.data["dataloader_num_workers"]
+        if (
+            self.config.data.curriculum is not None
+            and self.config.data.curriculum.get("curriculum_class_path", None) is not None
+        ):
+            assert num_workers == 0, "If using curriculum, num_workers must be 0 to prevent data caching."
```
zhaochenyang20: If we are using curriculum learning, we want to avoid to use the cached dataset, right?

zhaochenyang20 (changes requested):

verl/trainer/ppo/ray_trainer.py
frrad: yes. do you think this comment could be more clear?

zhaochenyang20: I think we should tell users why we avoid to use the cached dataset. more elaborate than `"If using curriculum, num_workers must be 0 to prevent data caching."`

zhaochenyang20 (changes requested):

tests/utils/dataset/test_create_rl_sampler.py
```diff
@@ -0,0 +1,82 @@
+# Copyright 2025 Amazon.com Inc and/or its affiliates
```
eric-haibin-lin: please add suffix `_on_cpu.py` so that this test gets picked up by cpu CI worker

frrad:
@Peng-YM I decided to add to the interface in this PR to reduce back and forth.

zhaochenyang20 (approved):
LGTM

verl/trainer/config/ppo_trainer.yaml
```diff
@@ -88,6 +91,15 @@ data:
     # The name of the dataset class within the specified file.
     name: null
 
+  # settings related to curriculum learning
```
eric-haibin-lin: what about renaming it to be `sampler` instead of `curriculum_sampler`? 
If it's an instance of `AbstractCurriculumSampler` then trainer should check the setting for `num_workers`

examples/sglang_multiturn/run_qwen0.5b_gsm8k_multiturn_curriculum.sh
```diff
@@ -0,0 +1,56 @@
...
+ulimit -n 65535
+
+PROJECT_DIR="$(pwd)"
+CONFIG_PATH="$PROJECT_DIR/examples/sglang_multiturn/config"
+
+python3 -m verl.trainer.main_ppo \
+    --config-path="$CONFIG_PATH" \
+    --config-name='gsm8k_multiturn_grpo' \
+    algorithm.adv_estimator=grpo \
+    data.curriculum_sampler.class_name="RandomCurriculumSampler" \
```
eric-haibin-lin: what is the expected output of this script, is it the same as not using curriculum sampler?

verl/trainer/main_ppo.py
```diff
@@ -272,9 +274,26 @@ def create_rl_sampler(data_config, dataset):
     import torch
     from torch.utils.data import RandomSampler, SequentialSampler
```
eric-haibin-lin, line 298: let's not restrict the custom class to be only curriculum samplers

verl/utils/dataset/curriculum_sampler.py
```diff
@@ -0,0 +1,56 @@
...
+        config: DictConfig,
+    ):
+        pass
+
+    @abstractmethod
+    def update(self, batch: DataProto) -> None:
+        pass
+
+
+class RandomCurriculumSampler(AbstractCurriculumSampler):
```
eric-haibin-lin: what's the purpose of this class? If it is functionally equivalent to random sampler, we should just reuse the one from pytorch and discard this class

verl/utils/import_utils.py
```diff
@@ -101,6 +101,28 @@ def load_extern_type(file_path: Optional[str], type_name: Optional[str]):
     return getattr(module, type_name)
 
 
+def load_type_from_module(module_path: str, type_name: str) -> Type:
```
eric-haibin-lin: please add an unit test for this function, since it looks useful and can potentially be reused in other places in the codebase

verl/trainer/config/ppo_trainer.yaml
```diff
@@ -88,6 +91,15 @@ data:
...
+  curriculum_sampler:
+
+    # the path to the module containing a curriculum class which implements the AbstractCurriculumSampler interface
```
eric-haibin-lin: please add example values to this field. In other custom configs file paths are used, but it seems here it refers to a module path. Having an example helps aovid the confusion. 
Furthermore, we should probably have a single util function that address both loading a class from a module `a.b.c` and from a file `a/b/c.py`?

eric-haibin-lin (changes requested):
thanks for the PR!

verl/utils/import_utils.py
eric-haibin-lin: actually, we should just extend `load_extern_type` with added module_path support

verl/trainer/config/ppo_trainer.yaml
zhaochenyang20: Thanks. I agree. We can provide general interface for sampler, not only for curriculum learning

examples/sglang_multiturn/run_qwen0.5b_gsm8k_multiturn_curriculum.sh
zhaochenyang20: Yeah. It just demonstrating if users create their sampler, how to enable it.

verl/trainer/config/ppo_trainer.yaml
frrad: Initially, I tried implementing this with the existing `load_extern_type`, but the `isinstanceof(RandomCurriculumSampler, AbstractCurriculumSampler)` failed when pointed at the existing `Random` class. I think the way the current `load_extern_type` works it only loads a file in isolation, removing information about its relationships to the rest of the modules. 

If we were to allow users to supply file paths instead of module paths they would be unable to ever subclass `AbstractCurriculumSampler` in a way that would be recognized. I think that it's better to leave the utilities separate and prevent the user from importing from a path to avoid this situation.

verl/utils/dataset/curriculum_sampler.py
frrad: This was for testing, will remove.

examples/sglang_multiturn/run_qwen0.5b_gsm8k_multiturn_curriculum.sh
frrad: This is just to exercise the new codepath. It's functionally the same. Do you think it's better to remove this file?

verl/trainer/config/ppo_trainer.yaml
frrad: great suggestion

verl/utils/import_utils.py
frrad: something like this was just merged in https://github.com/volcengine/verl/pull/2326 I have moved to using that.

eric-haibin-lin: actually load_extern_type just added support for `pkg://verl.xx.yy` that uses import_module`. So it should suffice your need?

frrad: I just noticed that due to a merge conflict

frrad:
@eric-haibin-lin thank you for your review. please take another look

eric-haibin-lin (approved):

verl/utils/dataset/sampler.py
```diff
@@ -0,0 +1,36 @@
...
+    def __init__(
+        self,
+        data_source: Sized,
+        config: DictConfig,
+    ):
+        pass
+
+
+class AbstractCurriculumSampler(AbstractSampler):
+    @abstractmethod
```
eric-haibin-lin, line 34: for curriculum learning, my gut is that it's quite hard to get it right at this point since the algorithm can be quite diverse and it changes quickly. the update API seems too general and I suspect it may get changed in the future. Shall we rename it to be a less ambiguous name such as `on_batch_end`? I also want to move this to `verl/experimental` since the API is likely to evolve overtime @frrad

PR #2321, opened 2025-07-02
[misc] feat: add nccl timeout configuration to fsdp workers

shinytang6:
### What does this PR do?

add nccl timeout config for fsdp backend

### DIFF
--- a/verl/workers/fsdp_workers.py
+++ b/verl/workers/fsdp_workers.py
@@ -15,6 +15,7 @@
 The main entry point to run the PPO algorithm
 """
 
+import datetime
 import json
 import logging
 import os
@@ -123,6 +124,7 @@ class ActorRolloutRefWorker(Worker, DistProfilerExtension):
                 backend=f"cpu:gloo,{get_device_name()}:{get_nccl_backend()}",
                 rank=rank,
                 world_size=world_size,
+                timeout=datetime.timedelta(seconds=self.config.get("nccl_timeout", 600)),
                 init_method=os.environ.get("DIST_INIT_METHOD", None),
             )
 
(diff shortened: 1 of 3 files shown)


verl/workers/fsdp_workers.py
```diff
@@ -919,11 +921,13 @@ def __init__(self, config):
         DistProfilerExtension.__init__(self, DistProfiler(rank=self.rank, config=profiler_config))
         import torch.distributed
 
+        self.config = config
         if not torch.distributed.is_initialized():
             torch.distributed.init_process_group(
-                backend=get_nccl_backend(), init_method=os.environ.get("DIST_INIT_METHOD", None)
+                backend=get_nccl_backend(),
+                timeout=datetime.timedelta(seconds=self.config.get("nccl_timeout", 600)),
```
vermouth1992, line 927: I guess it's better to use environment variable here

shinytang6: @vermouth1992 why? l think configuration is much more clear than using env

vermouth1992: To be compatible with init_method. Also, nccl related configs are all environment variables.

shinytang6: @vermouth1992 l'm ok with both choices, but l see you'r using configuration instead of env 
in megatron_workers, do l need to make it the same in this PR?

vermouth1992: Oh, my bad. I guess it's better to unify them

shinytang6: ok, let me do it

shinytang6: @vermouth1992 done, pls help review again

shinytang6: @vermouth1992 So the expectation is to checkout to the `nccl_timeout` configuration, right?

shinytang6: cc @vermouth1992

shinytang6:
@vermouth1992 @eric-haibin-lin can you pls take a look on this pr? l'm seeing a lot error related to this issue

PR #2324, opened 2025-07-02
[BREAKING][rollout] feat: repeat DataProto when n>1 in driver instead of rollout workers

wuxibin89:
### What does this PR do?

Before this PR, when `generate_sequences` with sampling param n>1, DataProto repeat is quit diverge.
- validation: DataProto is repeated by `n` in driver, then chunked and dispatched to rollout workers.
- training
  - batch mode: DataProto is chunked and dispatched to rollout workers, then repeated in rollout workers
  - server mode: DataProto is repeated by `n` in driver, then chunked and dispatched to rollout workers.

In batch mode, the `chunk-dispatch-repeat` pattern restricts GRPO training where we have more GPUs than batch_size. For example, `batch_size=128, n=16, world_size=256`:
- `chunk-dispatch-repeat`: DataProto(batch_size=128) can't be chunked to 256 shards.
- `repeat-chunk-dispatch`: after repeat, DataProto(batch_size=2048) can be successfully chunked.

After this PR, always repeat DataProto in driver whether it's validate or training, batch mode or server mode.

> [!IMPORTANT]
> This change breaks almost all recipes and projects using verl GRPO as submodules. 

### Checklist Before Starting

### Test

### API and Usage Example

```python
```

### High-Level Design

### Specific Changes

> List the specific changes.

> [!IMPORTANT]

### DIFF
--- a/verl/workers/rollout/sglang_rollout/sglang_rollout.py
+++ b/verl/workers/rollout/sglang_rollout/sglang_rollout.py
@@ -1190,7 +1190,6 @@ class SGLangRollout(BaseRollout):
             non_tensor_batch={
                 "messages": np.array(messages),
                 "reward_scores": np.array(reward_scores),
-                "uid": np.array([req.uid for req in sorted_output_req_list]),
                 "multi_modal_inputs": np.array(multi_modal_inputs, dtype=object),
             },
         )


verl/trainer/ppo/ray_trainer.py
```diff
@@ -1116,15 +1110,10 @@ def fit(self):
                     non_tensor_batch_keys=non_tensor_batch_keys_to_pop,
                 )
 
-                if repeat_sampling_sglang_grpo:
-                    uids_for_prompts = np.array([str(uuid.uuid4()) for _ in range(len(batch.batch))], dtype=object)
-                    batch.non_tensor_batch["uid"] = uids_for_prompts
-                    gen_batch.non_tensor_batch["uid"] = uids_for_prompts
-                    batch = batch.repeat(repeat_times=self.config.actor_rollout_ref.rollout.n, interleave=True)
+                # batch mode: generate n responses for each prompt with sampling_params(n>1).
```
vermouth1992: Shall we unify batch mode as well?

wuxibin89: I prefer keep batch mode as it is:
- repeat in batch mode introduces more communication cost between driver and worker
- it may also introduce BREAKING compatibility to other repo based on batch mode

vermouth1992: Actually, this helps train with small batch size with large n on large number of GPUs. say train_batch_size=128, n=16, n_gpus=256. In the current setting, it will throw an error.

zhaochenyang20: I think that for batch mode, we shall also repeat it. We repeat it just for "train_batch_size=128, n=16, n_gpus=256."

verl/workers/rollout/sglang_rollout/sglang_rollout.py
```diff
@@ -1140,7 +1140,6 @@ def _req_level_generate_sequences(self, prompts: DataProto, **kwargs) -> DataPro
             non_tensor_batch={
                 "messages": np.array(messages),
                 "reward_scores": np.array(reward_scores),
-                "uid": np.array([req.uid for req in sorted_output_req_list]),
```
zhaochenyang20, line 1203: If the uid is just removed, how to make sure the `union` of batch?

verl/trainer/ppo/ray_trainer.py
wuxibin89: OK, let me unify batch mode.

verl/workers/rollout/sglang_rollout/sglang_rollout.py
wuxibin89: In original batch mode with n>1, uid is added in original batch, it's not needed in gen_batch_output, just to make sure `gen_batch_output` has same order.
```
batch.non_tensor_batch["uid"] = np.array(
    [str(uuid.uuid4()) for _ in range(len(batch.batch))], dtype=object
)
# repeat to align with repeated responses in rollout
batch = batch.repeat(repeat_times=self.config.actor_rollout_ref.rollout.n, interleave=True)
batch = batch.union(gen_batch_output)
```

zhaochenyang20: ok. batch mode should always have the right order. But for a list of requests, this is not what happens. Make sense to me.

Thanks!

wuxibin89: I have check that `_req_level_generate_sequences` has order guarantee:
https://github.com/volcengine/verl/blob/main/verl/workers/rollout/sglang_rollout/sglang_rollout.py#L1031-L1040

zhaochenyang20:
```python
        logger.info(
            "n is deprecated for SGLang rollout since ray ppo trainer will repeat the prompts for rollout.n times"
        )
```

```python
            req = AsyncRolloutRequest(
                batch_data_id=data_idx,
                rollout_offset=0,
                request_id=str(uuid4()),
                uid=uid,
                state=AsyncRolloutRequestStateEnum.PENDING,
                messages=raw_prompt.tolist(),
                multi_modal_data=multi_modal_data,
                tool_schemas=_tool_schemas,
                tools_kwargs=_tools_kwargs,
                interaction_kwargs=_interaction_kwargs,
                input_ids=_input_ids,
                response_ids=[],
                attention_mask=_attention_mask,
                response_attention_mask=[],
                response_position_ids=[],
                response_loss_mask=[],
                reward_scores={},
                max_prompt_len=self.config.prompt_length,
                max_response_len=self.config.response_length,
                max_model_len=min(self.config.max_model_len, self.config.prompt_length + self.config.response_length),
                use_inference_chat_template=self.config.multi_turn.use_inference_chat_template,
                tokenization_sanity_check_mode=self.config.multi_turn.tokenization_sanity_check_mode,
                processing_class=self.processing_class,
            )
```

`rollout_offset` is set to alway 0. So the sorting is done on `x.batch_data_id`, which makes its order be kept. Great. You are correct.

zhaochenyang20: The gen batch is like:

```python
prompt 1 prompt 2 prompt 3
```

And we repeat it in ray ppo trainer.

```python
prompt 1 prompt 1 prompt 1 prompt 2 prompt 2 prompt 2 prompt 3 prompt 3 prompt 3 prompt 3
```

Then its `data_idx` is:

```python
prompt 1 (0) prompt 1 (1) prompt 1 (2) prompt 2 (3) prompt 2 (4) prompt 2 (5) prompt 3 (6) prompt 3 (7) prompt 3 (8) prompt 3 (9)
```

And we sort it by: 

```python
sorted_output_req_list = sorted(output_req_list, key=lambda x: x.batch_data_id)
```

The order is kept for GRPO group. 

Is this what actually happenning?

wuxibin89: Yes, you're right. There's a hard constraint that `generate_sequences` should keep output sequences same order as input prompts in DataProto, whether batch mode or server mode.

verl/trainer/ppo/ray_trainer.py
wuxibin89: @zhaochenyang20 batch mode is also repeat in driver.

verl/workers/rollout/vllm_rollout/vllm_rollout_spmd.py
```diff
@@ -192,7 +185,7 @@ def __init__(self, model_path: str, config: DictConfig, tokenizer, model_hf_conf
         for k in config.keys():
             if hasattr(SamplingParams(), str(k)):
                 kwargs[k] = config.get(k)
-
+        kwargs["n"] = 1  # already repeat in ray_trainer
         print(f"kwargs: {kwargs}")
```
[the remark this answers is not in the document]
vermouth1992: I guess print is fine. The problem is that it prints on every rank, that pollutes the log when the number of GPUs grow.

tests/workers/rollout/rollout_vllm/test_vllm_spmd.py
```diff
@@ -67,6 +68,7 @@ def are_lists_similar(a, b):
     return percentage_difference <= 15
 
 
+@pytest.mark.skip("https://github.com/vllm-project/vllm/issues/16993")
```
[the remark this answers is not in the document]
vermouth1992: Why skip this test?

verl/trainer/ppo/ray_trainer.py
zhaochenyang20: Okay, could you make a test on this setting `train_batch_size=16, n=16, n_gpus=8, rollout=sglang/vllm, multi-turn=true/false` (4 experiences indeed)

Make sure it works? If so, works to me!

tests/workers/rollout/rollout_vllm/test_vllm_spmd.py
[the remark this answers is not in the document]
eric-haibin-lin: same question. is it not stable on verl CI as well?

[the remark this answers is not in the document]
wuxibin89: vllm sleep mode incurs segment fault at exit, this test only works with vllm==0.7.3: 
https://github.com/vllm-project/vllm/issues/16993#issuecomment-2841972683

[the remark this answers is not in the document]
vermouth1992: nm, I got the point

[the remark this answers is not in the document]
wuxibin89: Skip this test temporary once this issue fixed, our ci image is vllm==0.8.5.

zhaochenyang20 (approved):

eric-haibin-lin:
If we break it, the change must be highlighted in the release note, since it impact almost all recipes and projects using verl  GRPO as submodules ...

wuxibin89:
> If we break it, the change must be highlighted in the release note, since it impact almost all recipes and projects using verl GRPO as submodules ...

@eric-haibin-lin @vermouth1992 @zhaochenyang20 Add breaking changes to v0.4.2 release notes.

vermouth1992 (approved):

PR #2330, opened 2025-07-02
[cfg] chore: add non-negative expected_len assertion

LeavesLei:
#### Summary
Added a assertion when the overlong buffer configuration is invalid, specifically when `overlong_buffer_len > max_response_length` which causes `expected_len` to be negative.

#### Problem
When `overlong_buffer_len` is greater than `max_response_length`, the calculated `expected_len` becomes negative:
```
expected_len = max_resp_len - overlong_buffer_len  # Results in negative value
```
This causes all reasonable response lengths to be penalized.

#### Solution
Added a `assert self.max_resp_len >= self.overlong_buffer_cfg.len` in DAPORewardManager

#### Changes Made
1. File: `verl/workers/reward_manager/dapo.py`

### DIFF
--- a/verl/workers/reward_manager/dapo.py
+++ b/verl/workers/reward_manager/dapo.py
@@ -45,6 +45,9 @@ class DAPORewardManager:
             assert self.max_resp_len is not None, (
                 f"max_resp_len must be provided if {overlong_buffer_cfg=}, but got None"
             )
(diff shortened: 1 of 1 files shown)


eric-haibin-lin:
please fix the pre-commit error

LeavesLei:
Thanks. I have fixed the pre-commit error.

PR #2343, opened 2025-07-03
[rollout, data] fix: force chat to end with EOS when add_generation_prompt==False

dirtyDan0:
### What does this PR do?

Removing everything after the last `EOS` when `apply_chat_template(add_generation_prompt=False)`

### Checklist Before Starting

### API and Usage Example

```python
def force_chat_end_with_eos(rendered_chat, eos_token):
    idx = rendered_chat.rfind(eos_token)
    if idx == -1:
        # This should never happen
        raise ValueError("No EOS found in the rendered chat.")
    return rendered_chat[: idx + len(eos_token)]

raw_prompt = tokenizer.apply_chat_template(
              messages,
              add_generation_prompt=False,
              tokenize=False,
          )
raw_prompt = force_chat_end_with_eos(raw_prompt, eos_token=tokenizer.eos_token)
```

### High-Level Design

We use `tokenizer/processor.apply_chat_template(messages, add_generation_prompt=False)` followed by encoding to reconstruct the tokens in previous turns/ the complete history and expect them to be identical to what llm has generated. However, some templates add `\n` after `EOS`, which may cause discrepancies especially when postprocessing messages in `CompletionCallback`and getting `generation_prompt_ids/text` in `AsyncRolloutRequest` and `MultiTurnSFTDataset`. A general solution is to manually strip everything that follows the last `EOS`, regardless of whether the chat template appends extra content after it.

You can run the following code to see the difference:

```python
from vllm import LLM, SamplingParams
from transformers import AutoTokenizer

model_path = "Qwen/Qwen3-0.6B"
llm = LLM(
    model=model_path,
    dtype="float16",
    max_model_len=256
)
sampling_params = SamplingParams(
    temperature=0.7,
    top_p=0.95,
    max_tokens=512,
)
messages = [
    {"role": "user", "content": "Introduce yourself."}
]
outputs = llm.chat(messages, sampling_params)

generated_ids = outputs[0].prompt_token_ids + list(outputs[0].outputs[0].token_ids)

tokenizer = AutoTokenizer.from_pretrained(model_path)

messages.append({"role":"assistant","content":outputs[0].outputs[0].text})
ids = tokenizer.apply_chat_template(messages,add_generation_prompt=False,tokenize=True)
print(ids[:-1] == generated_ids) #differ in the last \n

def force_chat_end_with_eos(rendered_chat, eos_token):
    idx = rendered_chat.rfind(eos_token)
    if idx == -1:
        # This should never happen
        raise ValueError("No EOS found in the rendered chat.")
    return rendered_chat[: idx + len(eos_token)]

raw_prompts = tokenizer.apply_chat_template(messages,add_generation_prompt=False,tokenize=False)
raw_prompts = force_chat_end_with_eos(raw_prompts,eos_token=tokenizer.eos_token)
ids = tokenizer(raw_prompts)["input_ids"]
print(ids == generated_ids)

```

### Specific Changes

* applying `force_chat_end_with_eos` for all `apply_chat_template(add_generation_prompt=False)`
* btw, In `ToolAgentLoop`, the system prompt may depend on `tools` (e.g. [[qwen3 system prompt need tools](https://huggingface.co/Qwen/Qwen3-8B?chat_template=default)](https://huggingface.co/Qwen/Qwen3-8B?chat_template=default))

> [!IMPORTANT]

- [x] Read the [[Contribute Guide](https://github.com/volcengine/verl?tab=readme-ov-file#contribution-guide)](https://github.com/volcengine/verl?tab=readme-ov-file#contribution-guide).
- [x] Apply [[pre-commit checks](https://github.com/volcengine/verl?tab=readme-ov-file#code-linting-and-formatting)](https://github.com/volcengine/verl?tab=readme-ov-file#code-linting-and-formatting): `pre-commit install && pre-commit run --all-files --show-diff-on-failure --color=always`
- [ ] Add / Update [[the documentation](https://github.com/volcengine/verl/tree/main/docs)](https://github.com/volcengine/verl/tree/main/docs).
- [x] Add unit or end-to-end test(s) to [[the CI workflow](https://github.com/volcengine/verl/tree/main/.github/workflows)](https://github.com/volcengine/verl/tree/main/.github/workflows) to cover all the code. If not feasible, explain why: ...

### DIFF
--- a/verl/experimental/agent_loop/tool_agent_loop.py
+++ b/verl/experimental/agent_loop/tool_agent_loop.py
@@ -112,7 +112,9 @@ class ToolAgentLoop(AgentLoopBase):
 
         cls.prompt_length = config.actor_rollout_ref.rollout.prompt_length
         cls.response_length = config.actor_rollout_ref.rollout.response_length
-        cls.system_prompt = tokenizer.apply_chat_template([{}], add_generation_prompt=False, tokenize=True)
+        cls.system_prompt = tokenizer.apply_chat_template(
+            [{}], tools=cls.tool_schemas, add_generation_prompt=True, tokenize=True
+        )
 
     async def run(self, messages: List[Dict[str, Any]], sampling_params: Dict[str, Any]) -> AgentLoopOutput:
         metrics = {}
--- a/verl/workers/rollout/schemas.py
+++ b/verl/workers/rollout/schemas.py
@@ -78,6 +78,14 @@ class TokenizationSanityCheckModeEnum(str, Enum):
     IGNORE_STRIPPABLE = "ignore_strippable"
 
 
+def force_chat_end_with_eos(rendered_chat, eos_token):
+    idx = rendered_chat.rfind(eos_token)
+    if idx == -1:
+        # This should never happen
+        raise ValueError("No EOS found in the rendered chat.")
+    return rendered_chat[: idx + len(eos_token)]
+
+
 class AsyncRolloutRequest(BaseModel):
     """The data model for async rollout."""
 
@@ -219,17 +227,28 @@ class AsyncRolloutRequest(BaseModel):
                 logger.warning(
                     "There is multi_modal_data but you are not using a processor. Multi-modal data will be ignored."
                 )
-            return processing_class.apply_chat_template(
+
+            raw_prompt = processing_class.apply_chat_template(
                 messages,
                 tools=tools,
                 add_generation_prompt=add_generation_prompt,
-                tokenize=tokenize,
-                return_dict=return_dict,
+                tokenize=False,
             )
+
+            if add_generation_prompt is False:
+                raw_prompt = force_chat_end_with_eos(raw_prompt, eos_token=processing_class.eos_token)
+
+            return (
+                processing_class(raw_prompt, return_dict=return_dict, return_tensors="pt") if tokenize else raw_prompt
+            )
+
         elif isinstance(processing_class, ProcessorMixin):
             raw_prompt = processing_class.apply_chat_template(
                 messages, tools=tools, add_generation_prompt=add_generation_prompt, tokenize=False
             )
+            if add_generation_prompt is False:
+                raw_prompt = force_chat_end_with_eos(raw_prompt, eos_token=processing_class.tokenizer.eos_token)
+
             if not tokenize:
                 return raw_prompt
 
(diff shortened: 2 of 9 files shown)


verl/workers/rollout/schemas.py
```diff
@@ -78,6 +78,14 @@ class TokenizationSanityCheckModeEnum(str, Enum):
     IGNORE_STRIPPABLE = "ignore_strippable"
 
 
+def force_chat_end_with_eos(rendered_chat, eos_token):
```
chenhaiq: Please describe the purpose of this function in detail in a pydoc str, because it can not be easily understand.

verl/experimental/agent_loop/tool_agent_loop.py
```diff
@@ -112,7 +112,9 @@ def init_class(cls, config, tokenizer):
 
         cls.prompt_length = config.actor_rollout_ref.rollout.prompt_length
         cls.response_length = config.actor_rollout_ref.rollout.response_length
-        cls.system_prompt = tokenizer.apply_chat_template([{}], add_generation_prompt=False, tokenize=True)
+        cls.system_prompt = tokenizer.apply_chat_template(
```
wuxibin89: `system_prompt` is used to remove prefix when encode tool response:
https://github.com/volcengine/verl/blob/main/verl/experimental/agent_loop/tool_agent_loop.py#L161-L167

Since we don't pass tool_schemas when encode tool response, neither do here. For Qwen3, it's ok that the system_prompt is empty.

dirtyDan0: Got it 👌🏻

verl/workers/rollout/schemas.py
dirtyDan0: Added 👌🏻

dirtyDan0:
For Qwen Base models (e.g., Qwen2.5 and Qwen3 Base), the tokenizer uses `<|endoftext|>` as the EOS token, while the `chat_template` uses `<|im_end|>`. As a result, the previous CI run raised a `ValueError` because `<|endoftext|>` was not found in the rendered chat output.

As clarified [here](https://github.com/QwenLM/Qwen3/issues/927#user-content-fn-3p-5481b9c93be76840bb8e07746e624824), Qwen base models are not designed for chat-based conversations. This means that `apply_chat_template` cannot faithfully reconstruct the actual tokens, since the model uses `<|endoftext|>` as the EOS token, while the chat template inserts `<|im_end|>` instead.

As a workaround, when the rendered chat does not contain the expected EOS token, we fall back to searching for `<|im_end|>` instead.

( Does this mean we should not use qwen base models for multi-turn rollout, since `apply_chat_template` is necessary for encoding previous conversations?)

dirtyDan0:
Please take a look :)

PR #2362, opened 2025-07-04
[sglang, rollout] refactor: use torch.Tensor in async rollout schemas

nanjiangwill:
### What does this PR do?

This PR refactors the `AsyncRolloutRequest` model to use `torch.Tensor` instead of `List[int]` for tensor-related fields. This change improves performance by avoiding unnecessary conversions between tensor and list types during processing.

### Checklist Before Starting

- [X] Search for similar PRs. Paste at least one query link here: https://github.com/volcengine/verl/pulls?q=is%3Apr+sglang+schema

### Test

Existing tests pass with these changes. The modifications ensure that the data model works correctly with tensor types.

### API and Usage Example

No API changes. The internal implementation now directly uses `torch.Tensor` types instead of converting to and from lists.

### Specific Changes

- Added `model_config = ConfigDict(arbitrary_types_allowed=True)` to allow `torch.Tensor` in Pydantic model
- Changed field types in `AsyncRolloutRequest` from `List[int]` to `torch.Tensor`
- Changed `multi_modal_inputs` type from `Dict[str, Any]` to `Dict[str, torch.Tensor]` for type safety
- Updated method implementations to work with tensor operations (e.g., using `torch.cat` instead of list concatenation)
- Updated shape checking to use tensor dimensions (e.g., `tensor.shape[-1]` instead of `len(list)`)
- Simplified position_ids handling by using tensor operations
- Improved truncation operations using tensor slicing

- [X] Apply [pre-commit checks](https://github.com/volcengine/verl?tab=readme-ov-file#code-linting-and-formatting): `pre-commit install && pre-commit run --all-files --show-diff-on-failure --color=always`
- [X] Add unit or end-to-end test(s) to [the CI workflow](https://github.com/volcengine/verl/tree/main/.github/workflows) to cover all the code. Existing tests cover these changes.
- [X] Once your PR is ready for CI, send a message in [the `ci-request` channel](https://verl-project.slack.com/archives/C091TCESWB1) in [the `verl` Slack workspace](https://join.slack.com/t/verl-project/shared_invite/zt-3855yhg8g-CTkqXu~hKojPCmo7k_yXTQ).


nanjiangwill:
Addressing reviewer comments:

Thank you for the feedback. I understand the concerns about using `torch.Tensor` in Pydantic models. I have added `model_config = ConfigDict(arbitrary_types_allowed=True)` to allow arbitrary types in the model, which addresses the runtime errors issue.

The main motivation for this change is to improve performance by avoiding unnecessary conversions between tensor and list types during processing, especially in the hot path of the async rollout system.

zhaochenyang20 (approved):
LGTM

eric-haibin-lin (commented):
I noticed that pydantic is used in rollout, and https://github.com/volcengine/verl/pull/2147 this PR introduces simple dataclass for verl internal config passing to minimize the reliance on omegaconf everywhere (especially in unit tests). while it's not urgent for now but later we need to probably think about whether to unify to pydantic or simple dataclass. Any thoughts?

nanjiangwill:
`pydantic` definitely is a good choice if all configs only include primitive types (int, list, dict, str, etc). but we should use a simple dataclass if, in the future, we decide to include more complex data types or type hinting.

from the perspective of keeping the code base clean while not reinventing the wheel, I would vote for `pydantic` for now and use supportive type hinting tools such as [torchtyping](https://github.com/patrick-kidger/torchtyping). I think in the future, as long as pydantic is adopted, there will be more and more extensions built upon it, and we can use them all.

from the perspective of moving fast with the `verl` ecosystem, if there is enough effort and enough people spending time maintaining the in-house dataclass, then I have no objection to building a simple dataclass by `verl`.

PR #2376, opened 2025-07-05
Revert "[misc] fix: invalid escape sequence '\*'"

vermouth1992:
Reverts volcengine/verl#2375


verl/utils/dataset/rl_dataset.py
```diff
@@ -43,7 +43,7 @@ def collate_fn(data_list: list[dict]) -> dict:
 
     Returns:
         Dict where tensor entries are stacked into a torch.Tensor of shape
-        (batch_size, *dims) and non-tensor entries are converted to
+        (batch_size, \*dims) and non-tensor entries are converted to
```
wuxibin89, line 46: Why we should escape this?

PR #2381, opened 2025-07-06
[data] refactor: change sampler api from update to a more specific one 

eric-haibin-lin:
### What does this PR do?

change sampler api from update to a more specific one and document that it is experimental 

### Checklist Before Starting

> [!IMPORTANT]

### DIFF
--- a/verl/utils/dataset/sampler.py
+++ b/verl/utils/dataset/sampler.py
@@ -21,6 +21,9 @@ from verl import DataProto
 
 
 class AbstractSampler(Sampler[int]):
+    """Abstract interface for custom curriculum learning samplers. The interface
+    is experimental"""
+
     @abstractmethod
     def __init__(
         self,
@@ -32,5 +35,5 @@ class AbstractSampler(Sampler[int]):
 
 class AbstractCurriculumSampler(AbstractSampler):
     @abstractmethod
-    def update(self, batch: DataProto) -> None:
+    def on_batch_end(self, batch: DataProto) -> None:
         pass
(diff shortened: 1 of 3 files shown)


verl/utils/dataset/sampler.py
```diff
@@ -21,16 +21,20 @@
...
 
 class AbstractSampler(Sampler[int]):
+    """Abstract interface for custom samplers."""
+
     @abstractmethod
     def __init__(
         self,
         data_source: Sized,
-        config: DictConfig,
+        data_config: DictConfig,
```
eric-haibin-lin, line 30: the abstract class use `data_config` instead of `config`

eric-haibin-lin: @frrad FYI

frrad:
thanks! I was planning to work on this Monday. One small thing = what do you think about calling it something like `update_on_batch_end` instead of just `on_batch_end` ? I think it would be useful for the name to communicate something about what this method does from the perspective of the sampler.

verl/utils/dataset/sampler.py
```diff
@@ -21,16 +21,20 @@
...
     ):
         pass
 
 
 class AbstractCurriculumSampler(AbstractSampler):
+    """Experimental interface for curriculum learning samplers."""
+
     @abstractmethod
-    def update(self, batch: DataProto) -> None:
+    def on_batch_end(self, batch: DataProto) -> None:
```
vermouth1992: Actually, I would prefer "update" as "on_batch_end" implicitly assumes that there will be a training loop, which doesn't really exists for async RL on the data sampling part.

eric-haibin-lin: just changed it back

PR #2466, opened 2025-07-10
[rollout] feat: implemented the persistable replay buffer feature

Xunmei-liu:
## **Implemented the replay buffer to have the following features:**
**1. Large-scale data storage**
Uses an in-memory cache combined with RocksDB, preventing OOM issues when handling large data storage and transfers, such as when storing rollout results.

**2. Customizable sampling strategies**
Supports pluggable samplers, with uniform sampling and p-time sampling already implemented.

**3. Data persistence**
Allows other trials or the same trial after restarting to reuse data in the replay buffer.

## **Performance:**
Achieved QPS comparable to Python dict. During rollout, the performance is similar to the baseline (using a python dict) when most data is in the in-memory cache. However, access times may increase as more data resides in RocksDB.

## **Links for other information:**
User manual: https://github.com/volcengine/verl/issues/2584
RFC: https://github.com/volcengine/verl/issues/2539


setup.py
```diff
@@ -41,6 +41,8 @@
     "transformers",
     "wandb",
     "packaging>=20.0",
+    "rocksdb-py",
+    "pympler",
 ]
```
eric-haibin-lin, line 44: please move to extras_require

tests/utils/replay_buffer/e2e_replay_buffer_perf_test.py
```diff
@@ -0,0 +1,103 @@
+import copy
+import logging
+import random
+import time
+
+import torch
+
+from verl import DataProto
+from verl.utils.replay_buffer.persistable_replay_buffer_client import PersistableReplayBufferClient
```
eric-haibin-lin: the import stmt is too long.. on the other hand, replay buffer can be quite complicated and usually require an RFC for the community to review. I'd suggest moving to 
`from verl.experimental.replay_buffer.persistable_client import XXX` instead

eric-haibin-lin: can u create a RFC with design choices, pro and cons?

burgerkingeater: @Xunmei-liu also include a user-facing doc on the usage of this feature, limitation, etc

verl/experimental/replay_buffer/samplers/uniform_key_sampler.py
```diff
@@ -0,0 +1,41 @@
...
+class UniformKeySampler(Sampler):
+    def __init__(self) -> None:
+        super().__init__({get_key_index_name()})
+
+    def build_index(self, key, value):
+        # build an index table which maps:
+        # indexed key to key.
+        self._index[get_key_index_name()][key] = key
+
+    def sample(self) -> Iterator:
```
PrinsYin, line 31: From my understanding, the keys list is captured once and never refreshed, meaning the sampler will continue yielding keys from the same list, right?

Zhou-jiecheng:
In VLM training, there are images and videos in DataProto, which makes DataProto dispatch and concat very time-consuming, perhaps we need to consider both scenarios together.

ccclyu:
good feature!  is there any follow-up on benchmarking the performance or any plan for integrating back to verl?

PR #2475, opened 2025-07-11
[misc] fix: Use `AutoModelForImageTextToText` instead of `AutoModelForVision2Seq`

kaln27:
### What does this PR do?
- According to https://github.com/huggingface/transformers/pull/38900, the `AutoModelForVision2Seq` will be deprecated soon. So verl need to use `AutoModelForImageTextToText` instead. 
- Also I found some model doesn't have 'generation_config.json` file, but the code assume it has. It will cause an error. So I check whether it is existed first (Already fix by other PR) .
- I also create a PR https://github.com/huggingface/transformers/pull/39199 in transformers to support GLM4.1v training in verl. 
- Add GLM4.1v flops counter function.  (Already done bu other PR)


kaln27:
@eric-haibin-lin Hi eric. Could you review this PR? Thanks a lot.

ruanhao566:
May I ask why there was no merge? AutoModelForVision2Seq is no longer available in the new version of transformers, and I am encountering the same error.

PR #2498, opened 2025-07-12
[single_controller] fix: replace unittest.mock.patch with context manager for env var handling

PeterSH6:
### What does this PR do?

Fixes a critical issue in the Ray worker initialization where environment variables were not being properly preserved when using `unittest.mock.patch`. This could lead to environment variables being unexpectedly deleted after worker initialization. The fix replaces the `patch` usage with a proper context manager for safer environment variable management.

### Checklist Before Starting

- [x] Search for similar PRs. Paste at least one query link here: https://github.com/volcengine/verl/pulls?q=is%3Apr+environment+variables+patch
- [x] Format the PR title as `[single_controller] fix: replace unittest.mock.patch with context manager for env var handling`

### Test

This change can be tested through the CI system since it affects core worker initialization functionality. The following test scenarios should be covered:
- Worker initialization with pre-existing environment variables
- Worker initialization without pre-existing environment variables
- Multiple worker initializations in sequence
- Error cases during worker initialization

### API and Usage Example

No API changes. Internal implementation change only. The fix uses a new context manager:

```python
@contextmanager
def temp_env_var(key: str, value: str):
    """Context manager for temporarily setting an environment variable."""
    original = os.environ.get(key)
    os.environ[key] = value
    try:
        yield
    finally:
        if original is None:
            os.environ.pop(key, None)
        else:
            os.environ[key] = original

# Usage in worker initialization
with temp_env_var("DISABLE_WORKER_INIT", "1"):
    worker = user_defined_cls(*args, **kwargs)
```

### Design & Code Changes

Changes made:
1. Removed dependency on `unittest.mock.patch`
2. Added new `temp_env_var` context manager for safe environment variable handling
3. Updated worker initialization code in two locations to use the context manager:
   - In `WorkerDict.__init__` for regular worker initialization
   - In `FusedWorker.__init__` for fused worker initialization
4. Ensures environment variables are properly restored even if initialization fails

- [x] Apply pre-commit checks
- [ ] Add / Update the documentation - N/A (internal implementation change)
- [x] Add unit tests to verify environment variable handling in worker initialization
- [ ] Request CI via Slack channel

### DIFF
--- a/tests/utils/test_temp_env_on_cpu.py
+++ b/tests/utils/test_temp_env_on_cpu.py
@@ -19,10 +19,10 @@ import pytest
 from verl.utils.py_functional import temp_env_var
 
 
-def setup_function():
-    """Set up function called before each test function."""
-    # Store original environment state to restore after each test
-    global original_env
+@pytest.fixture(autouse=True)
+def clean_env():
+    """Fixture to clean up environment variables before and after each test."""
+    # Store original environment state
     original_env = dict(os.environ)
 
     # Clean up any test variables that might exist
@@ -31,10 +31,10 @@ def setup_function():
         if var in os.environ:
             del os.environ[var]
 
+    # Yield control to the test function
+    yield
 
-def teardown_function():
-    """Tear down function called after each test function."""
-    # Restore original environment state
+    # Restore original environment state after test
     os.environ.clear()
     os.environ.update(original_env)
 
--- a/tests/utils/test_temp_env_on_cpu.py
+++ b/tests/utils/test_temp_env_on_cpu.py
@@ -1,0 +1,145 @@
+# Copyright 2024 Bytedance Ltd. and/or its affiliates
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


verl/single_controller/ray/base.py
```diff
@@ -753,6 +753,28 @@ def _determine_fsdp_megatron_base_class(mros: List):
...
+    """
+    original = os.environ.get(key)
+    os.environ[key] = value
+    try:
+        yield
+    finally:
+        if original is None:
+            os.environ.pop(key, None)
+        else:
+            os.environ[key] = original
```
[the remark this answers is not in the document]
PeterSH6: We may not have to worry about thread safe here?

verl/single_controller/ray/base.py
```diff
@@ -753,6 +753,28 @@ def _determine_fsdp_megatron_base_class(mros: List):
     raise ValueError(f"Cannot determine base class for {mros}")
 
 
+@contextmanager
+def temp_env_var(key: str, value: str):
```
vermouth1992: Shall we move this as a utility function?

vermouth1992: Please also add a test

tests/utils/test_temp_env_on_cpu.py
```diff
@@ -0,0 +1,145 @@
+# Copyright 2024 Bytedance Ltd. and/or its affiliates
```
vermouth1992, line 1: I guess we have to use pytest, otherwise it won't be invoked.

eric-haibin-lin: pytest is covering all test_xx files https://github.com/volcengine/verl/blob/main/.github/workflows/cpu_unit_tests.yml#L83

PR #2521, opened 2025-07-14
[ray] refactor: Use public method to get node IP

kevin85421:
### What does this PR do?

1. Currently, verl uses `ray._private.services.get_node_ip_address()` to get the node IP. However, it's better to avoid using functions under `_private`. Instead, we should use the public API `ray.util.get_node_ip_address()`. Both are equivalent: https://github.com/ray-project/ray/blob/c6e2080a96b39328c8b4b506390340e97a903688/python/ray/util/__init__.py#L6.

2. Update some methods in `class WorkerHelper` to be `@staticmethods` because they don't rely on the class's state.

### Checklist Before Starting

### API and Usage Example

No

### Design & Code Changes

No

> [!IMPORTANT]

### DIFF
--- a/verl/single_controller/base/worker.py
+++ b/verl/single_controller/base/worker.py
@@ -45,19 +45,10 @@ class DistGlobalInfo:
 class WorkerHelper:
     @staticmethod
     def _get_node_ip():
-        def get_node_ip_by_sdk():
-            if os.getenv("WG_BACKEND", None) == "ray":
-                return ray.util.get_node_ip_address()
-            else:
-                raise NotImplementedError("WG_BACKEND now just support ray mode.")
-
-        host_ipv4 = os.getenv("MY_HOST_IP", None)
-        host_ipv6 = os.getenv("MY_HOST_IPV6", None)
-        host_ip = host_ipv4 or host_ipv6
-        if host_ip is None:
-            host_ip = get_node_ip_by_sdk()
-
-        return host_ip
+        if os.getenv("WG_BACKEND", None) == "ray":
+            return ray.util.get_node_ip_address()
+        else:
+            raise NotImplementedError("WG_BACKEND now just support ray mode.")
 
     @staticmethod
     def _get_free_port():


verl/single_controller/base/worker.py
```diff
@@ -43,34 +43,31 @@ class DistGlobalInfo:
 
 
 class WorkerHelper:
-    def _get_node_ip(self):
+    @staticmethod
+    def _get_node_ip():
         def get_node_ip_by_sdk():
             if os.getenv("WG_BACKEND", None) == "ray":
-                import ray
```
kevin85421, line 49: Ray has already been imported in L22.

verl/single_controller/base/worker.py
```diff
@@ -43,34 +43,31 @@ class DistGlobalInfo:
...
-
-                return ray._private.services.get_node_ip_address()
+                return ray.util.get_node_ip_address()
             else:
                 raise NotImplementedError("WG_BACKEND now just support ray mode.")
 
         host_ipv4 = os.getenv("MY_HOST_IP", None)
```
kevin85421: I didn't find any documentation or code that uses `MY_HOST_IP` or `MY_HOST_IPV6`. Should we remove it if they are not used? 

Btw, Ray currently doesn't support ipv6.

verl/single_controller/base/worker.py
```diff
@@ -43,34 +43,31 @@ class DistGlobalInfo:
...
+    @staticmethod
+    def _get_free_port():
         with socket.socket() as sock:
             sock.bind(("", 0))
             return sock.getsockname()[1]
 
     def get_availale_master_addr_port(self):
         return self._get_node_ip(), str(self._get_free_port())
 
-    def _get_pid(self):
```
kevin85421, line 71: It is not used anywhere.

verl/single_controller/base/worker.py
```diff
@@ -43,34 +43,31 @@ class DistGlobalInfo:
...
-                return ray._private.services.get_node_ip_address()
+                return ray.util.get_node_ip_address()
             else:
                 raise NotImplementedError("WG_BACKEND now just support ray mode.")
 
         host_ipv4 = os.getenv("MY_HOST_IP", None)
         host_ipv6 = os.getenv("MY_HOST_IPV6", None)
-        host_ip_by_env = host_ipv4 or host_ipv6
-        host_ip_by_sdk = get_node_ip_by_sdk()
+        host_ip = host_ipv4 or host_ipv6
```
kevin85421: Only call `get_node_ip_by_sdk` when the env vars are not set.

verl/single_controller/base/worker.py
vermouth1992: This was meant for internal usage. Feel free to remove it

vermouth1992: When can Ray support ipv6? This is critical for us @wuxibin89

wuxibin89: `MY_HOST_IP` or `MY_HOST_IPV6` is safe to remove, we can always use `ray.util.get_node_ip_address()`. As our discussion with Anyscale Team, ipv6 support is in Q3. @kevin85421

vermouth1992 (approved):

verl/single_controller/base/worker.py
kevin85421: I heard users request for ipv6 multiple times recently https://github.com/ray-project/kuberay/issues/2999#issuecomment-2650298837. I’ll discuss it with our team and try to get it on the roadmap.

kevin85421: > This was meant for internal usage. Feel free to remove it

I will remove it.

kevin85421: Another question: Does it make sense for me to introduce an interface in the following PR to improve maintainability?

```python
class WorkerGroupBackend(ABC):
    @abstractmethod
    def get_node_ip(self):
        pass

    @abstractmethod
    def xxxxx(self):
        pass

class RayWorkerGroupBackend(WorkerGroupBackend):
    ...
```

verl/single_controller/base/worker.py
```diff
@@ -43,34 +43,22 @@ class DistGlobalInfo:
...
-                return ray._private.services.get_node_ip_address()
-            else:
-                raise NotImplementedError("WG_BACKEND now just support ray mode.")
-
-        host_ipv4 = os.getenv("MY_HOST_IP", None)
-        host_ipv6 = os.getenv("MY_HOST_IPV6", None)
-        host_ip_by_env = host_ipv4 or host_ipv6
-        host_ip_by_sdk = get_node_ip_by_sdk()
-
-        host_ip = host_ip_by_env or host_ip_by_sdk
```
hongpeng-guo, line 60: QQ: how do we set node_ip with env vars now? It seems `host_ip` can not receive as assignment from env vars right now.

kevin85421: I have the same question here:
https://github.com/volcengine/verl/pull/2521#discussion_r2204015326

verl/single_controller/base/worker.py
hongpeng-guo: This API is used internally via inheritance and might be useful in OSS later, can we still keep it alive :)

kevin85421: My recommendation is not to leave unused functions in the OSS codebase, and to add them only when they are actually used. In addition, [pass-through methods](https://www.mattduck.com/2021-04-a-philosophy-of-software-design.html) are usually not a recommended pattern. Is it OK to call `os.getpid()` directly instead of using `_get_pid()`?

The implementation isn't that important, and I’m guessing this change might create some overhead when your team rebases the OSS repo with the internal one. You can make the final decision whether we should remove the function or not, and I will do that.

hongpeng-guo: sg, let's remove this API. But keep a record on this issue: https://github.com/volcengine/verl/issues/2528. The `single_controller` module is usually modified by various teams internally subject their various underlying infras. Let's keep a record of the updates since v0.4.0~

hongpeng-guo (commented):
This PR looks to me in general.

Please record the API changes and functional updates in this issue: https://github.com/volcengine/verl/issues/2528

i.e., the removal of `_get_pid` and the env vars are not supported in `_get_host_ip`

verl/single_controller/base/worker.py
kevin85421: What information would you like me to record? There is no behavior change for the public functions.

kevin85421: Oh, the issue covers API changes and refactoring.

kevin85421:
The error in https://github.com/volcengine/verl/actions/runs/16272518188/job/45943502003?pr=2521 `RuntimeError: CUDA error: an illegal memory access was encountered` seems to be unrelated to this PR.

@hongpeng-guo is this test known to be flaky?

vermouth1992 (approved):

PR #2524, opened 2025-07-14
[fsdp] fix: vlm dynamic batch & unify dynamic batch api

hiyouga:
### What does this PR do?

The use of `len(data)` is incorrect since the data is a **dict** if we enable dynamic batch for VLMs. It will return the number of keys in the dict instead of the number of batch samples.

https://github.com/volcengine/verl/blob/0b508ab803497b1929727abb3cbade28361a600f/verl/workers/actor/dp_actor.py#L540-L542

https://github.com/volcengine/verl/blob/0b508ab803497b1929727abb3cbade28361a600f/verl/workers/actor/dp_actor.py#L432-L434

It can work correctly with pure-text LLMs because the data here is a **tensordict** that has a `len` API.

https://github.com/volcengine/verl/blob/0b508ab803497b1929727abb3cbade28361a600f/verl/workers/actor/dp_actor.py#L441-L443

To solve this problem, we use `response_mask.shape[0]` to get the number of samples in dynamic batch.

Nevertheless, I think the current implementation isn't elegant because the underlying object processed here can be either a **dict** or a **tensordict**. So I unify the APIs of dynamic batch and provide two functions: `prepare_dynamic_batch` and `restore_dynamic_batch`. They can be used for both computing log probs and updating actor. They remove the redundant code and make a clean view for the fsdp workers.

### Checklist Before Starting

### Test

Test with Qwen2.5-VL-3b dynamic batch on the Geo3k dataset

```bash
python examples/data_preprocess/geo3k.py --local_dir ~/data/geo3k

    data.train_files=$HOME/data/geo3k/train.parquet \
    data.val_files=$HOME/data/geo3k/test.parquet \
    data.train_batch_size=512 \
    data.max_prompt_length=1024 \
    data.max_response_length=2048 \
    data.truncation='error' \
    data.image_key=images \
    actor_rollout_ref.model.path=Qwen/Qwen2.5-VL-3B-Instruct \
    actor_rollout_ref.actor.ppo_mini_batch_size=128 \
    actor_rollout_ref.actor.use_dynamic_bsz=True \
    actor_rollout_ref.actor.ppo_max_token_len_per_gpu=6144 \
    actor_rollout_ref.actor.kl_loss_coef=0.01 \
    actor_rollout_ref.rollout.log_prob_micro_batch_size_per_gpu=20 \
    actor_rollout_ref.rollout.enable_chunked_prefill=False \
    actor_rollout_ref.rollout.enforce_eager=False \
    actor_rollout_ref.rollout.free_cache_engine=False \
    actor_rollout_ref.rollout.n=5 \
    actor_rollout_ref.ref.log_prob_max_token_len_per_gpu=6144 \
    trainer.critic_warmup=0 \
    trainer.logger=['console','wandb'] \
    trainer.project_name='verl_nightly_ci' \
    trainer.experiment_name='qwen2_5_vl_3b_function_rm' \
    trainer.n_gpus_per_node=4 \
    trainer.nnodes=1 \
    trainer.save_freq=-1 \
    trainer.test_freq=5 \
    trainer.total_epochs=15
```

Results: orange: before this PR, blue: after this PR

image

### API and Usage Example

```python
def prepare_dynamic_batch(data: DataProto, max_token_len: int) -> tuple[list[DataProto], list[list[int]]]:
    """
    Prepare a batch for dynamic batching.

    Args:
        data (DataProto): The input data.
        max_token_len (int): The maximum token length for dynamic batching.

    Returns:
        Tuple[List[DataProto], List[List[int]]]: A tuple containing a list of DataProto objects
        and a list of index lists.
    """
    ...

def restore_dynamic_batch(data: torch.Tensor, batch_idx_list: list[list[int]]) -> torch.Tensor:
    """
    Restore a batch from dynamic batching.

    Args:
        data (torch.Tensor): The input data.
        batch_idx_list (List[List[int]]): The list of index lists.

    Returns:
        torch.Tensor: The restored data.
    """
    ...
```

### Design & Code Changes

> [!IMPORTANT]

### DIFF
--- a/verl/utils/seqlen_balancing.py
+++ b/verl/utils/seqlen_balancing.py
@@ -14,10 +14,12 @@
 
 import copy
 import heapq
+from itertools import chain
 
 import torch
 from torch import distributed as dist
 
+from verl.protocol import DataProto
 from verl.utils.device import get_device_name
 
 
@@ -321,3 +323,41 @@ def get_reverse_idx(idx_map):
         reverse_idx_map[idx] = i
 
     return reverse_idx_map
+
+
+def prepare_dynamic_batch(data: DataProto, max_token_len: int) -> tuple[list[DataProto], list[list[int]]]:
+    """
+    Prepare a batch for dynamic batching.
+
+    Args:
+        data (DataProto): The input data.
+        max_token_len (int): The maximum token length for dynamic batching.
+
+    Returns:
+        Tuple[List[DataProto], List[List[int]]]: A tuple containing a list of DataProto objects
+        and a list of index lists.
+    """
+    batch, batch_idx_list = rearrange_micro_batches(data.batch, max_token_len=max_token_len)
+    micro_batches = []
+    for i, batch_idx in enumerate(batch_idx_list):
+        tensors = dict(batch[i])
+        non_tensors = {key: value[batch_idx] for key, value in data.non_tensor_batch.items()}
+        micro_batches.append(DataProto.from_dict(tensors, non_tensors))
+
+    return micro_batches, batch_idx_list
+
+
+def restore_dynamic_batch(data: torch.Tensor, batch_idx_list: list[list[int]]) -> torch.Tensor:
+    """
+    Restore a batch from dynamic batching.
+
+    Args:
+        data (torch.Tensor): The input data.
+        batch_idx_list (List[List[int]]): The list of index lists.
+
+    Returns:
+        torch.Tensor: The restored data.
+    """
+    indices = list(chain.from_iterable(batch_idx_list))
+    revert_indices = torch.tensor(get_reverse_idx(indices), dtype=torch.long)
+    return data[revert_indices]
--- a/tests/utils/test_seqlen_balancing.py
+++ b/tests/utils/test_seqlen_balancing.py
@@ -18,7 +18,13 @@ import torch.multiprocessing as mp
 
 from verl import DataProto
 from verl.utils.model import create_random_mask
-from verl.utils.seqlen_balancing import ceildiv, get_reverse_idx, rearrange_micro_batches
+from verl.utils.seqlen_balancing import (
+    ceildiv,
+    get_reverse_idx,
+    prepare_dynamic_batch,
+    rearrange_micro_batches,
+    restore_dynamic_batch,
+)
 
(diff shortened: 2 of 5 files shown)


verl/utils/seqlen_balancing.py
```diff
@@ -321,3 +323,44 @@ def get_reverse_idx(idx_map):
         reverse_idx_map[idx] = i
 
     return reverse_idx_map
+
+
+def prepare_dynamic_batch(data: DataProto, max_token_len: int) -> tuple[list[DataProto], list[list[int]]]:
```
vermouth1992, line 328: As a core functionality, could you please write a test case?

hiyouga: done~

ultmaster:
@hiyouga May I ask why uneven split is not supported?

https://github.com/hiyouga/verl/blob/8e8efd68267d303a259726bee9c421d485b24cd6/verl/workers/actor/dp_actor.py#L325

In my case, the batch could get a size of 146 with chunk size 4. My code used to work before but now crashes with the latest version:

```
  File "/scratch/amlt_code/verl/verl/workers/actor/dp_actor.py", line 325, in compute_log_prob
    micro_batches = data.split(micro_batch_size)
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/scratch/amlt_code/verl/verl/protocol.py", line 739, in split
    assert len(self) % split_size == 0, (
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError: only support equal split. Got size of DataProto 146 and chunk 4.
```

tensordict also has a split, and they do not require an even split.

https://github.com/pytorch/tensordict/blob/bc1b20abbc49ebadec296732c7e964d5ad529eb7/tensordict/_td.py#L1744

hiyouga:
@ultmaster how about using
```
return [self[i:i + split_size] for i in range(0, len(self), split_size)]
```
to replace the current code?

ultmaster:
@hiyouga I think it'll work. I'm not sure about the performance differences from the native pytorch split.

hiyouga:
@ultmaster I think they are comparable. If it can actually work, could you create a PR to fix this?

PR #2580, opened 2025-07-16
set use_kl_in_reward=True in reinforce_plus_plus

Titanpku:
set use_kl_in_reward=True in reinforce_plus_plus


Yangruipis:
@vermouth1992 would please review this pr? reinforce++ should set true for use_kl_in_reward and false for use_kl_loss , see #644 (comment)

vermouth1992:
> @vermouth1992 would please review this pr? reinforce++ should set true for use_kl_in_reward and false for use_kl_loss , see #644 (comment)

If that's the case, shall we add assertion in the code?

Yangruipis:
> > @vermouth1992 would please review this pr? reinforce++ should set true for use_kl_in_reward and false for use_kl_loss , see #644 (comment)
> 
> If that's the case, shall we add assertion in the code?

i'd rather not, sometimes people still want to break the common cases and do some exploration. 

we only need to ensure the example provided is a standard one, thats enough, perhaps.

vermouth1992 (approved):

PR #2583, opened 2025-07-16
[cfg] feat: add critic config class

eric-haibin-lin:
Added CriticConfig, MegatronCriticConfig, and FSDPCriticConfig dataclasses with a clear inheritance hierarchy for critic model configuration, and updated YAML files to support direct dataclass instantiation.

## Changes
  - Introduced dataclasses for critic configs, all inheriting from BaseConfig.
  - Added _target_ fields to critic YAML files for compatibility with omega_conf_to_dataclass.
  - Added unit tests to verify config instantiation and inheritance.

##  Special notice
both megatron and fsdp critic contains the following config:
- model
- optimizer

however, the config names in these two configs are not yet consistent. In this PR, they are retreated as `dict[str, Any]` for flexibility. We shall introduce model config and optimizer config are they are consolidated. 

I've also removed kl_cntrol from megatron critic config, they're not used @ETOgaosion


ZihengJiang:
With this config, can we run the current criticworker independently?

eric-haibin-lin:
the dataclass config can be created programmically. But I have not tried to run critic worker independently

tests/trainer/config/legacy_ppo_megatron_trainer.yaml
```diff
@@ -327,9 +327,6 @@ critic:
   data_loader_seed: ${actor_rollout_ref.actor.data_loader_seed}
   shuffle: ${actor_rollout_ref.actor.shuffle}
   cliprange_value: 0.5
-  kl_ctrl:
```
ETOgaosion, line 330: We don't need these?

verl/trainer/config/_generated_ppo_trainer.yaml
```diff
@@ -259,11 +259,11 @@ critic:
     warmup_style: constant
   model:
     path: ~/models/deepseek-llm-7b-chat
-    tokenizer_path: ${actor_rollout_ref.model.path}
+    tokenizer_path: ${oc.select:actor_rollout_ref.model.path,/models/deepseek-llm-7b-chat}
```
ETOgaosion: Is it `/model/`, or `~/model`?

verl/trainer/config/_generated_ppo_trainer.yaml
```diff
@@ -300,8 +300,9 @@ critic:
     discrete: false
     all_ranks: false
     ranks: []
-  forward_micro_batch_size: ${critic.ppo_micro_batch_size}
-  forward_micro_batch_size_per_gpu: ${critic.ppo_micro_batch_size_per_gpu}
+  _target_: verl.trainer.config.FSDPCriticConfig
```
ETOgaosion, line 303: Is it a comment? Maybe put it below `critic` to be more obvious?

verl/trainer/config/critic/critic.yaml
```diff
@@ -23,31 +23,31 @@ model:
   path: ~/models/deepseek-llm-7b-chat
 
   # Tokenizer path (defaults to actor's model path)
-  tokenizer_path: ${actor_rollout_ref.model.path}
+  tokenizer_path: ${oc.select:actor_rollout_ref.model.path,/models/deepseek-llm-7b-chat}
```
ETOgaosion: Same here `/model`

tests/trainer/config/legacy_ppo_megatron_trainer.yaml
eric-haibin-lin: no there's a copy in Algorithm

verl/trainer/config/_generated_ppo_trainer.yaml
eric-haibin-lin: It's autogenerated

verl/trainer/config/_generated_ppo_trainer.yaml
eric-haibin-lin: nice catch!

eric-haibin-lin:
@ETOgaosion updated the default value, please review again

vermouth1992 (approved):

PR #2614, opened 2025-07-18
[CI] feat: add `mypy` to pre-commit

frrad:
### What does this PR do?

Adds `mypy` typechecking to pre-commit. Only one module is whitelisted and it is configured to be very permissive.

### Test

this is a CI / testing change

### API and Usage Example

```python
```

### Design & Code Changes


zhaochenyang20:
@frrad CI is good. I can help to merge if haibin approve.

frrad:
copying comment here from Lark 

@eric-haibin-lin 
> i wonder why the type is corrected to "Any". in fact we pass DictConfig or AlgoConfig. 
> Does the mypy make the correction itself, or the change has to be done manually?

@frrad 
> I made that change manually. the `AlgoConfig` type is incorrect - none of the fields referenced on L939 https://github.com/volcengine/verl/blob/0fdbe54118729ab0a571e9f08c9929a6e9f7f375/verl/trainer/ppo/core_algos.py#L939 actually exist on the datatype https://github.com/volcengine/verl/blob/f252da34cffd66cb2eb48d41005e86e7da0c7120/verl/trainer/config/algorithm.py#L105-L114
> 
> I was not sure what the real type was so I changed it to `Any`. I can change it to `DictConfig` instead if you prefer.

eric-haibin-lin:
Let's change it to Optional AlgoConfig|DictConfig]

frrad:
please take another look

frrad:
In https://github.com/volcengine/verl/pull/2614/commits/cc2972c11431a5ab189813ee855eaa0429e1baf1 I have added an additional path to the mypy whitelist which make it necessary to change `AlgoConfig` to add the fields referenced from `core_algos`.  Happy to revert that commit to keep the scope of this first PR as small as possible if you want.

edit: originally I had linked the wrong commit in this message

zhaochenyang20:
I think this PR is almost done. Let me help on the merge! @frrad

verl/trainer/config/algorithm.py
```diff
@@ -112,3 +121,7 @@ class AlgoConfig(BaseConfig):
     use_pf_ppo: bool = False
     pf_ppo: Optional[PFPPOConfig] = None
     filter_groups: Optional[FilterGroupsConfig] = None
+    policy_loss: Optional[PolicyLossConfig] = field(default_factory=PolicyLossConfig)
+    clip_ratio_high: float = 0.2
+    clip_ratio_low: float = 0.2
+    clip_ratio: float = 0.2
```
eric-haibin-lin: There are two issues. One is that verl already has `actor_rollout_ref.actor.policy_loss` and `actor_rollout_ref.actor.clip_ratio_high/low`, duplicating them without changing existing example script would be confusing. 
Also restructuring algo config & actor config is not a high priority for now and moving them around and breaking others' recipes may lead to bad experience. If we do want to make breaking changes, we usually make announcement for deprecations, keep it for one version before breaking it.. 

I do not think it's necessary to couple mypy setup with breaking changes ..

frrad:
I have reverted that commit.

frrad:
this is probably a better fix. please take another look.

eric-haibin-lin (approved):

zhaochenyang20:
@frrad Let me merge it after the CI

PR #2619, opened 2025-07-18
[recipe] fix: fix dapo checkpoint, cannot save the checkpoint of last step before

rong-hash:
### What does this PR do?

### Checklist Before Starting

Similar PR
https://github.com/volcengine/verl/pull/2090

### Test

image
Can not save the last checkpoint. Only save the checkpoint with training steps % save_freq=0

### Design & Code Changes

dapo_ray_trainer.py only record training steps variable but not record generation steps.

### DIFF
--- a/recipe/dapo/dapo_ray_trainer.py
+++ b/recipe/dapo/dapo_ray_trainer.py
@@ -67,6 +67,7 @@ class RayDAPOTrainer(RayPPOTrainer):
         )
 
         self.global_steps = 0
+        self.gen_steps = 0
 
         # load checkpoint before doing anything
         self._load_checkpoint()
@@ -86,6 +87,7 @@ class RayDAPOTrainer(RayPPOTrainer):
 
         # we start from step 1
         self.global_steps += 1
+        self.gen_steps += 1
         last_val_metrics = None
 
         timing_raw = defaultdict(float)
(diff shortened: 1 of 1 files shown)


recipe/dapo/dapo_ray_trainer.py
```diff
@@ -126,7 +128,7 @@ def fit(self):
                     )
                 gen_batch = gen_batch.repeat(repeat_times=self.config.actor_rollout_ref.rollout.n, interleave=True)
 
-                is_last_step = self.global_steps >= self.total_training_steps
+                is_last_step = self.gen_steps >= self.total_training_steps
```
wuxibin89, line 131: What is the difference between `gen_steps` and `global_steps`?

rong-hash: dapo will filter the generated samples whose rewards/scores are all the same. For one training step (global_step), there may be multiple generation steps to provide enough valid samples to train. Therefore, most of the time global_steps is smaller than gen_steps. 

However, I also notice that save_checkpoint and load_checkpoint only save the global_steps, which may cause other issues such as the data loading and training progress bar management. I'm not sure whether I have the permission to modify this important code. Maybe we can write a new save_checkpoint and load_checkpoint specifically for dapo?

PR #2620, opened 2025-07-18
[fsdp, megatron, sglang] fix: Fixed a bug in the update_weight process where the GPU ID was being passed incorrectly.

SuperCB:
### What does this PR do?

When using verl+megatron+sglang, we observed extremely unusual phenomena. We found that some abnormal processes were always running on GPU0, occupying VRAM. 

image

When Ray launches each process, it sets `CUDA_VISIBLE_DEVICES` for each process. For example, rank 1 is set to GPU 1, and rank 2 is set to GPU 2. However, the problem is that from the perspective of process rank 2, it treats GPU 2 as its own GPU 0 (index=0). This leads to rank 2 actually sending a GPU ID of 0 when transferring Tensors across processes. Consequently, every Worker in SGlang receives a Tensor with an incorrect GPU ID, causing abnormal memory usage on GPU 0,at the same time, this also causes unnecessary memory copies, slowing down the update_weights process.

[PR](https://github.com/pytorch/pytorch/pull/149248) addresses the issue of transferring Tensors across processes. This requires a patch to be applied both during serialization in the sending process and deserialization in the receiving process. Verl hasn't implemented this. Referencing [PR](https://github.com/pytorch/pytorch/pull/149248) and the implementation in [slime](https://github.THUDM/slime.git), we resolved this problem.
We also modified  the [Sglang](https://github.com/sgl-project/sglang/pull/8150).
### Checklist Before Starting

### Test

### API and Usage Example

```python
```

### Design & Code Changes

> [!IMPORTANT]

### DIFF
--- a/verl/workers/rollout/sglang_rollout/sglang_rollout.py
+++ b/verl/workers/rollout/sglang_rollout/sglang_rollout.py
@@ -162,16 +162,14 @@ class AsyncEngine(sglang.srt.entrypoints.engine.Engine):
 
     async def update_weights_from_tensor(
         self,
-        named_tensors: List[Tuple[str, torch.Tensor]],  # noqa: UP006
+        named_tensors: List[str],  # noqa: UP006
         load_format: Optional[str] = None,
         flush_cache: bool = True,
     ):
         """Update weights from distributed source. If there are going to be more updates, set `flush_cache` to be false
         to avoid duplicated cache cleaning operation."""
         obj = UpdateWeightsFromTensorReqInput(
-            serialized_named_tensors=[
-                MultiprocessingSerializer.serialize(named_tensors) for _ in range(self.server_args.tp_size)
-            ],
+            serialized_named_tensors=named_tensors,
             load_format=load_format,
             flush_cache=flush_cache,
         )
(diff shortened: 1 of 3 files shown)


chenhaiq:
Does this issue result "CUDA error: an illegal memory access was encountered"?

zhaochenyang20:
> Does this issue result "CUDA error: an illegal memory access was encountered"?

This may not be the same issue. illegal mem access may be fixed by bumping a higher version of sglang  https://github.com/volcengine/verl/pull/2720

Right now it's hard for us to reproduce illigal mem access, but typically, just bump up sglang can fix it.

hebiao064:
> > Does this issue result "CUDA error: an illegal memory access was encountered"?
> 
> This may not be the same issue. illegal mem access may be fixed by bumping a higher version of sglang #2720

@chenhaiq @zhaochenyang20 #2720 is built on top of #2620, we can close #2620

For illegal memory issue, we hope higher version can solve that, but we haven't find a way to consistently repro it.

vermouth1992:
Could you resolve the conflict?

vermouth1992:
It seems that all the sglang tests passed. Could you help move the deprecated sglang tests out after rebase? Thanks!

hebiao064:
Duplicated with #2720

PR #2117, opened 2025-06-20
[config, perf] refactor: add omega_conf_to_dataclass API, rename WorkerProfiler to DistProfiler, add unit test based on ProfilerConfig

eric-haibin-lin:
### Checklist Before Starting

- [x] Searched for similar PR(s).
- [x] Checked PR Title format
  - In format of: [modules] type: Title
  - modules are in `fsdp, megatron, sglang, vllm, rollout, trainer, ci, training_utils, recipe, hardware, deployment, ray, worker, single_controller, misc, perf, model, algo, env, tool, ckpt, doc, data`
  - type is in `feat, fix, refactor, chore, test`
  - can involve multiple modules, seperated by `,` or space, like `[megatron, fsdp, doc] feat: xxx`

### What does this PR do?

Previously, most of individual components in verl takes omega conf dict as one of the input, making it tedious to setup unit tests. Now verl is gradually introducing dataclass for each sub module for configuration, with `verl.utils.omega_conf_to_dataclass` to make the conversion easier. 
This PR also provide example unit tests on how standalone classes with config as the input should be tested before using them end-to-end. 
Finally, this PR also renames WorkerProfiler to DistProfiler for clarity. 

### Test

Test cases for configuration utilities on CPU.
    1. Test basic OmegaConf to dataclass conversion for simple nested structures
    2. Test nested OmegaConf to dataclass conversion for complex hierarchical configurations
    3. Verify all configuration values are correctly converted and accessible

Test suite for NsightSystemsProfiler functionality
    1. Initialization: Verify profiler state after creation
    2. Basic Profiling: Test start/stop functionality
    3. Discrete Mode: Test discrete profiling behavior
    4. Annotation: Test the annotate decorator in both normal and discrete modes
    5. Config Validation: Verify proper config initialization from OmegaConf

### Usage Example

> Provide usage example(s) for easier usage.

```python
def omega_conf_to_dataclass(config: Union[DictConfig, dict], dataclass_type: Type[Any]) -> Any:
    """
    Convert an OmegaConf DictConfig to a dataclass.

    Args:
        config: The OmegaConf DictConfig or dict to convert.
        dataclass_type: The dataclass type to convert to.

    Returns:
        The dataclass instance.
    """
```

### Checklist Before Submitting

- [ ] Read the [Contribute Guide](https://github.com/volcengine/verl?tab=readme-ov-file#contribution-guide).
[...]

PR #2621, opened 2025-07-18
[cfg] refactor: add ActorConfig, EngineConfig, and ActorWorker unit test, refactor validation code

eric-haibin-lin:
As initially mentioned in https://github.com/volcengine/verl/discussions/1941, having structured configuration classes in verl makes argument passing easier for testing and validation. 

This is an extended thread on the current implementation of configuration schema in verl. Related PRs:
-  https://github.com/volcengine/verl/pull/2117
- https://github.com/volcengine/verl/pull/2621 

# Motivation 
By moving from loose `omegaconfig.DictConfig`-based parameters to structured dataclasses, we gain:
- Type safety & IDE support when accessing fields (e.g. cfg.optim.lr).
- Validation hooks via __post_init__ in each class.
- Immutable defaults with controlled mutability (e.g., an extra field).
- Seamless Hydra/OmegaConf integration and easy per-recipe extension.

# Core: BaseConfig

hydra natively provides support for converting DictConfig to dataclass, but dataclass does not support accessing attribute via `get()`. We introduce a base class to provide backward compatibility and make the change less abrupt for existing users.

All config dataclasses inherit from BaseConfig, which:
- Implements collections.abc.Mapping → dict-like iteration/access.
- Freezes attributes once set, unless listed in _mutable_fields.
- Provides an `extra: dict[str, Any]` for unchecked extensions.

```python
@dataclass
class BaseConfig(collections.abc.Mapping):
    """Dict-like, frozen dataclass with opt-in mutability."""
    _mutable_fields: set[str] = {"extra"}
    extra: dict[str, Any] = field(default_factory=dict)

    def __setattr__(self, name: str, value):
        if name in self.__dict__ and name not in self._mutable_fields:
            raise FrozenInstanceError(f"Field '{name}' is frozen")
        super().__setattr__(name, value)

    # Mapping methods: get, __getitem__, __iter__, __len__ …

```

# Example Config Classes (verl/trainer/config)

Each sub-component of the trainer has its own dataclass, inheriting BaseConfig.
```yaml:
critic:
  checkpoint:
    _target_: verl.trainer.config.CheckpointConfig
    save_contents: ["model","optimizer","extra"]
    load_contents: ["model","optimizer","extra"]
    async_save: false
```
Definition: 
```python
@dataclass
class CheckpointConfig(BaseConfig):
    """What to save/load and async behavior."""
    save_contents: list[str] = field(default_factory=lambda: ["model","optimizer","extra"])
    load_contents: list[str] = field(default_factory=lambda: ["model","optimizer","extra"])
    async_save: bool = False

    def __post_init__(self):
        # validation checks go here after initialization


ckpt_cfg = CheckpointConfig(async_save=True)
print(ckpt_cfg.save_contents)
print(ckpt_cfg.get("save_contents", default_value))
print(ckpt_cfg["save_contents"])

# converting hydra-generated omegaconf.DictConfig to the dataclass config:
from verl.utils.config import omegaconf_to_dataclass
ckpt_cfg_from_cli = omegaconf_to_dataclass(config.critic.checkpoint)
```

# Extending existing config classes
Because now configs become structured, unexpected keys would raise exceptions. To add new keys, there are two ways:
## Explicit class extensions:
```python
from verl.workers.config import FSDPActorConfig

@dataclass
class SPPOActorConfig(FSDPActorConfig):
    """Add SPPO-specific temperature/penalty."""
    sppo_eta: float = 1.0

```
When using yaml or from command line, update the target config class:
```yaml
hydra:
  searchpath:
    - file://verl/trainer/config
defaults:
  - ppo_trainer      # base trainer config
  - _self_               # then apply these overrides

actor_rollout_ref:
  actor:
    _target_:  recipe.sppo.config.SPPOActorConfig # **new target dataclass required for extension **
    sppo_eta: 1.0  
```
or directly from command line:
```bash
python main_sppo.py \
  actor_rollout_ref.actor._target_=recipe.sppo.config.SPPOActorConfig \
  actor_rollout_ref.actor.sppo_eta=1.0
```

## Leverage the `extra` field
Adding more keys to the `extra` field of any dataclass that inherits from `BaseConfig` also works. This way there's no need to define your own dataclass in python:
```yaml
hydra:
  searchpath:
    - file://verl/trainer/config
defaults:
  - ppo_trainer      # base trainer config
  - _self_               # then apply these overrides

actor_rollout_ref:
  actor:
    extra:
        sppo_eta: 1.0  
```

# Declaring mutable fields
For historical reasons some fields in the configs are mutated inplace in the codebase such as batch size for data/sequence parallelism. We are in the process of deprecating this kind of behavior. However, if you want to intentionally mutate one field, specify it with the `_mutable_fields` attr: 
```python
@dataclass
class CheckpointConfig(BaseConfig):
    """What to save/load and async behavior."""
    _mutable_fields = BaseConfig._mutable_fields | {"save_contents"} # mark save_contents as mutable.

    save_contents: list[str] = field(default_factory=lambda: ["model","optimizer","extra"])
    load_contents: list[str] = field(default_factory=lambda: ["model","optimizer","extra"])
    async_save: bool = False
```

# Other helpful resources
verl default trainer configs combines the following config files together, specified in the `_defaults_` field: https://github.com/volcengine/verl/blob/main/verl/trainer/config/ppo_trainer.yaml#L1-L36
- verl/trainer/config/ppo_trainer.yaml  # main config for entrypoint 
- verl/trainer/config/actor/dp_actor.yaml 
- verl/trainer/config/critic/dp_critic.yaml 
- verl/trainer/config/reward_model/dp_reward_model.yaml 
- verl/trainer/config/rollout/rollout.yaml 

To quickly peek the default full config in a single file, you can check the auto-generated full config in https://github.com/volcengine/verl/blob/main/verl/trainer/config/_generated_ppo_trainer.yaml 

# Change log and impact on existing code
This PR converts the following fields to structured dataclass in the training pipeline. More can be done in future PRs (contributions from the community is welcome) 
- [x] actor_rollout_ref.actor
- [x] critic 
- [ ] actor_rollout_ref.rollout
- [ ] actor_rollout_ref.ref
- [ ] reward_model
- [ ] data
- [ ] trainer

Changes needed for existing code that added new fields to config:
- see recipe/sppo for an example 
- `OmegaConf.to_container(self.config.model.get("override_config", OmegaConf.create()))` now has to manually changed to `self.config.model.get("override_config", {})`. Because OmegaConf.to_container expects a DictConfig but config.model.override_config is already a dict.

# Other Breaking Changes
critic.optim.lr for megatron changed from 1e-6 to 1e-5


verl/workers/config/engine.py
```diff
@@ -0,0 +1,109 @@
...
+    use_dist_checkpointing: bool = False
+    dist_checkpointing_path: Optional[str] = None
+    seed: int = 42
+    override_ddp_config: dict[str, Any] = field(default_factory=dict)
+    override_transformer_config: dict[str, Any] = field(default_factory=dict)
+    use_mbridge: bool = False
+
+
+@dataclass
+class FSDPEngineConfig(BaseConfig):
```
ZihengJiang, line 80: shall we add _frozen_fields for FSDPEngine also?

eric-haibin-lin: yes, I'll do that

verl/workers/critic/megatron_critic.py
```diff
@@ -84,9 +84,6 @@ def _validate_config(self, config) -> None:
         assert config.get("ulysses_sequence_parallel_size", 1) == 1
         if config.shuffle:
             assert config.data_loader_seed is not None, "If shuffle dataloader, seed must be manually set"
-        if config.megatron.tensor_model_parallel_size == 1:
```
eric-haibin-lin, line 87: moved to `MegatronEngineConfig.__post_init__`

verl/trainer/config/actor/dp_actor.yaml
```diff
@@ -32,8 +35,8 @@ entropy_checkpointing: False
 # optimizer configs
 optim:
```
eric-haibin-lin, line 37: moved to actor.yaml

verl/trainer/config/actor/megatron_actor.yaml
```diff
@@ -4,28 +4,23 @@ defaults:
...
 
+_target_: verl.workers.config.McoreActorConfig
+
 strategy: megatron
 
 data_loader_seed: null
 
 load_weight: True
 
-checkpoint:
```
eric-haibin-lin, line 13: moved to actor.yaml

wuxibin89:
> Adding more keys to the extra field of any dataclass that inherits from BaseConfig also works. This way there's no need to define your own dataclass in python:

We have some keys not defined in config yaml and are dynamically append in script, for example:
```bash
    reward_model.reward_manager=dapo \
    +reward_model.reward_kwargs.overlong_buffer_cfg.enable=${enable_overlong_buffer} \
    +reward_model.reward_kwargs.overlong_buffer_cfg.len=${overlong_buffer_len} \
    +reward_model.reward_kwargs.overlong_buffer_cfg.penalty_factor=${overlong_penalty_factor} \
    +reward_model.reward_kwargs.overlong_buffer_cfg.log=False \
    +reward_model.reward_kwargs.max_resp_len=${max_response_length} \
```
Should we move them to `extra` field, or make them explicitly define in class? I prefer to define explicitly and discourage user to use `extra` field.

wuxibin89:
> For historical reasons some fields in the configs are mutated inplace in the codebase such as batch size for data/sequence parallelism. We are in the process of deprecating this kind of behavior.

Will we deprecate mutating config fields? It's pretty useful to set fields which must derived from config in runtime.

verl/trainer/ppo/ray_trainer.py
```diff
@@ -366,21 +368,17 @@ def __init__(
         if self.config.algorithm.use_kl_in_reward:
             self.kl_ctrl_in_reward = core_algos.get_kl_controller(self.config.algorithm.kl_ctrl)
 
-        if self.config.algorithm.adv_estimator == AdvantageEstimator.GAE:
+        if config.critic.get("enable") is not None:
```
wuxibin89: Should we discourage to use `.get` since it's already define in `CriticConfig.enable`?

eric-haibin-lin:
> > Adding more keys to the extra field of any dataclass that inherits from BaseConfig also works. This way there's no need to define your own dataclass in python:
> 
> We have some keys not defined in config yaml and are dynamically append in script, for example:
> 
> ```shell
>     reward_model.reward_manager=dapo \
>     +reward_model.reward_kwargs.overlong_buffer_cfg.enable=${enable_overlong_buffer} \
>     +reward_model.reward_kwargs.overlong_buffer_cfg.len=${overlong_buffer_len} \
>     +reward_model.reward_kwargs.overlong_buffer_cfg.penalty_factor=${overlong_penalty_factor} \
>     +reward_model.reward_kwargs.overlong_buffer_cfg.log=False \
>     +reward_model.reward_kwargs.max_resp_len=${max_response_length} \
> ```
> 
> Should we move them to `extra` field, or make them explicitly define in class? I prefer to define explicitly and discourage user to use `extra` field.

currently both are supported. Defining explicitly sounds fine to me. @tongyx361 @ZihengJiang @vermouth1992 thoughts?

verl/trainer/ppo/ray_trainer.py
eric-haibin-lin: yes we should. there're other occurrences of using get in the codebase and we can leave the clean-up work to the next PR

wuxibin89:
Can we add DAPO related fields explicitly in this PR(or next PR), since we have quit a few scripts rely on it.

eric-haibin-lin:
@wuxibin89 as for +reward_model.xx, it is not yet converted to dataclass in this PR (see changelog section in the PR description). I took a look at all +critic.xx and +actor_rollout_ref.actor.xx, and added 2 options related to fsdp_config in the last commit. For other megatron related ones, i verified that they can be added as expected because the expected type is a general dict:

```
python3 scripts/print_cfg.py +actor_rollout_ref.actor.megatron.override_transformer_config.num_layers_in_first_pipeline_stage=3 actor_rollout_ref.actor.ppo_micro_batch_size=1
McoreActorConfig(extra={}, strategy='megatron', ppo_mini_batch_size=256, ppo_micro_batch_size=1, ppo_micro_batch_size_per_gpu=None, use_dynamic_bsz=False, ppo_max_token_len_per_gpu=16384, clip_ratio=0.2, clip_ratio_low=0.2, clip_ratio_high=0.2, policy_loss=PolicyLossConfig(extra={}, loss_mode='vanilla', clip_cov_ratio=0.0002, clip_cov_lb=1.0, clip_cov_ub=5.0, kl_cov_ratio=0.0002, ppo_kl_coef=0.1), clip_ratio_c=3.0, loss_agg_mode='token-mean', entropy_coeff=0, use_kl_loss=False, use_torch_compile=True, kl_loss_coef=0.001, kl_loss_type='low_var_kl', ppo_epochs=1, shuffle=False, checkpoint=CheckpointConfig(extra={}, save_contents=['model', 'optimizer', 'extra'], load_contents=['model', 'optimizer', 'extra'], async_save=False), optim=McoreOptimizerConfig(extra={}, lr=1e-06, lr_warmup_steps_ratio=0.0, total_training_steps=-1, weight_decay=0.01, lr_warmup_steps=-1, optimizer='adam', clip_grad=1.0, lr_warmup_init=0.0, lr_decay_steps=None, lr_decay_style='constant', min_lr=0.0, weight_decay_incr_style='constant', lr_wsd_decay_style='exponential', lr_wsd_decay_steps=None, use_checkpoint_opt_param_scheduler=False), use_fused_kernels=False, data_loader_seed=None, load_weight=True, megatron=McoreEngineConfig(extra={}, param_offload=False, grad_offload=False, optimizer_offload=False, tensor_model_parallel_size=1, expert_model_parallel_size=1, expert_tensor_parallel_size=None, pipeline_model_parallel_size=1, virtual_pipeline_model_parallel_size=None, context_parallel_size=1, sequence_parallel=False, use_distributed_optimizer=True, use_dist_checkpointing=False, dist_checkpointing_path=None, seed=42, override_ddp_config={}, override_transformer_config={'num_layers_in_first_pipeline_stage': 3}, use_mbridge=False), profile={'use_profile': False, 'profile_ranks': None, 'step_start': -1, 'step_end': -1, 'save_path': None})
```

eric-haibin-lin:
For reward_model, there are still dataclasses to be consolidated. I think it's better to split changes into multiple PRs to avoid a too large PR

wuxibin89 (approved):

vermouth1992 (approved):

eric-haibin-lin:
Failure of tests/trainer/config/test_legacy_config_on_cpu.py can be ignored since it being fixed in https://github.com/volcengine/verl/pull/2675/

ETOgaosion:
Megatron tests fail because of the mess up of remaining recomputation config, also fixed in #2675

eric-haibin-lin:
Let's hold this PR, make a v0.5 release before merging it

PR #2623, opened 2025-07-18
[hardware] feat: unify flash attention imports into verl.utils.kernel.flash_attn

eric-haibin-lin:
### What does this PR do?

unify flash attention imports into verl.utils.kernel.flash_attn

### Checklist Before Starting

### Test

### API and Usage Example

```python
```

### Design & Code Changes

> [!IMPORTANT]

### DIFF
--- a/verl/utils/kernel/flash_attn.py
+++ b/verl/utils/kernel/flash_attn.py
@@ -1,0 +1,31 @@
+# Copyright 2025 Bytedance Ltd. and/or its affiliates
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from verl.utils.device import is_cuda_available, is_npu_available
+
+if is_cuda_available:
+    try:
+        from flash_attn.bert_padding import index_first_axis, pad_input, rearrange, unpad_input
+    except ImportError as err:
+        raise ImportError("flash_attn is required when CUDA is available but not installed") from err
+elif is_npu_available:
+    try:
+        from transformers.integrations.npu_flash_attention import (  # noqa: F401
+            index_first_axis,
+            pad_input,
+            rearrange,
+            unpad_input,
+        )
+    except ImportError as err:
+        raise ImportError("transformers with NPU flash attention support is required when NPU is available") from err
--- a/recipe/prime/prime_dp_rm.py
+++ b/recipe/prime/prime_dp_rm.py
@@ -19,13 +19,13 @@ import itertools
 
 import torch
 import torch.distributed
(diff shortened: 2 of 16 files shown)


vermouth1992 (dismissed):

verl/utils/kernel/flash_attn.py
```diff
@@ -0,0 +1,31 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from verl.utils.device import is_cuda_available, is_npu_available
+
+if is_cuda_available:
```
vermouth1992, line 17: This is not ideal because we may import these functions in pure CPU machine as well

wuxibin89: import `flash_attn` or `torch_npu` on cpu only machine will raise Exception, so it's ok here.

vermouth1992: Yes. Sometimes we import this function on CPU-only machine on driver and the object is pickled to the worker with GPU

vermouth1992: Maybe it's actually a better choice to copy those function out and implement in the verl repo.

ErenAta16 (commented):
The sweep looks complete: every `from flash_attn.bert_padding import ...` in the tree is routed through the new module, including the deferred imports inside `init_model` and `forward`, and the four names re-exported (`index_first_axis`, `pad_input`, `unpad_input`, `rearrange`) match what the call sites actually ask for. `rearrange` is the easy one to drop, since it comes from einops via `bert_padding` rather than from flash-attn itself, and it is there.

**The shim has no branch for "neither".**

```python
if is_cuda_available:
    try:    from flash_attn.bert_padding import ...
    except: raise ImportError("flash_attn is required when CUDA is available but not installed")
elif is_npu_available:
    try:    from transformers.integrations.npu_flash_attention import ...
    except: raise ImportError("transformers with NPU flash attention support is required ...")
```

On a box that is neither CUDA nor NPU, both branches are skipped, no exception is raised, and the module imports successfully while binding none of the four names. The failure then lands on the importer as

```
ImportError: cannot import name 'index_first_axis' from 'verl.utils.kernel.flash_attn'
```

which does not say what is missing or why. That is worth closing because the module is otherwise careful: it raises a clear, actionable message for both cases it does handle, and the third case is the one that gets nothing.

It is a real configuration rather than a hypothetical. `verl/utils/device.py` has an explicit `else: device = "cpu"` path, so CPU is a supported target, and the same applies to a ROCm or XPU box where both flags are false. Before this PR those environments failed at `from flash_attn.bert_padding import ...` with `ModuleNotFoundError: No module named 'flash_attn'`, which at least named the missing package. So this is an error-quality regression rather than a new break, but it is one the shim is well placed to fix:

```python
else:
    raise ImportError(
        "flash-attention padding helpers are unavailable on this platform "
        "(neither CUDA nor NPU detected)"
    )
```

Two of the touched files are `tests/models/test_transformer.py` and `tests/models/test_transformers_ulysses.py`, so whichever runner collects those is the first place the opaque message would show up.

One process note: this was opened 2025-07-18 and `mergeable` currently reports `UNKNOWN`, so it likely needs a rebase before it can land. Given the diff is a mechanical import rewrite across sixteen files, the longer it sits the more conflicts it accumulates against files that have moved since.

PR #2629, opened 2025-07-19
[rollout, trainer] feat: Enabling Request Skewness Scheduler towards near-equal generated token in rollout

Tyizhanshen:
### What does this PR do?

We've observed severe workload imbalance during the rollout phase of verl, particularly when setting a large `data.max_response_length`.

image

and in some case showing below, the workload **varies greatly** across DP ranks and the time gap between the fastest and slowest DP instances can exceed 1800s (e.g., between rank-5 and rank-4), leading to NCCL timeouts and significant GPU resource waste.

image

To mitigate this, this PR provides a batch-based, simple but efficient, implementation of request skewness scheduler. As shown below, the proposed request skewness scheduler dramatically reduces idle time between DP ranks in long-tail batches, resolving the NCCL timeouts and boosing both GPU utilization and rollout efficiency.
411f9234-8ce4-4720-8ad2-397864c1bce8

### Proposed Design

We observe a strong positive correlation between the load on a DP rank and the number of tokens it generates. Consequently, a greater disparity in the final token counts across these ranks leads to a significant variance in their execution times.

To this end, in contrast to the conventional even_prompt policy, we propose a new load-balancing strategy, `even_token`, and the architectural concept of a `Request Scheduler`.

The core objective of this strategy is to **ensure that the total number of generated tokens handled by each DP instance within a batch is approximately equal.** This, in turn, minimizes the completion time disparity among different DP instances ([see scheduler diagram](#scheduler-design)).

To achieve this, the key is to estimate the generation length for each request. We have opted for a more direct statistical method for its engineering simplicity, rather than training a separate prediction model. 

The implementation is as follows:
* **Length Estimation**: We maintain a lookup table that records the actual response length for each request from the previous episode (Note the lookup table can be maintained and saved as a json file for feature experiments).

* **Scheduling Execution**: When distributing a new batch, the scheduler references this historical length data to estimate the computational load of each request. It then dynamically distributes the requests, aiming to make the sum of estimated generation lengths nearly identical for every DP instance.

* **Order Restoration**: After all DP instances have completed generation, a `reorder` operation is performed to restore the outputs to their original sequence, ensuring data integrity for subsequent stages.

By leveraging the actual generation lengths from the previous episode to balance the current batch's load, the even_token strategy can effectively reduce the computational "bubbles" caused by length disparities, leading to a significant improvement in overall training efficiency and resource utilization.

### Evaluation
* dataset: DAPO-Math-17k
* max_response_length: 20k
* model: DeepSeek-R1-Distill-Qwen-7B
* 8*powerful GPUs in Hopper architecture，DP_size=8
* algorithm base: DAPO

绘图2

* As the results demonstrate, enabling the Request Skewness Scheduler provides a significant performance uplift, which becomes more pronounced as the workload imbalance increases. At training step 7, for example, the scheduler achieves a 25% performance improvement.
* Critically, the baseline run without the scheduler fails during training step 8 due to an `NCCL timeout`, leading to a crucial challenge for training stability. In contrast, our method with the scheduler maintains training stability even with severe "long-tail" batch issues, highlighting its effectiveness in improving both performance and robustness.

-----

* As already demonstrated in  ([scheduler diagram](#scheduler-design)), by balancing the token generation load across GPUs, our scheduling strategy achieves uniform generation times, reducing idle time among data-parallel workers and boosts GPU utilization especially when handling batches with a severe long-tail distribution.

-----
|  |  |
|:---:|:---:|
| image-20250717182347698 | image-20250717182459552 |

* Enabling the Request Skewness Scheduler yields no significant difference for subsequent stages, which is attributable to our Order Restoration function.

### Discussion
In our design, we repeat DataProto (when n > 1) in rollout workers instead of the driver[(#2324)](https://github.com/volcengine/verl/pull/2324). This choice is motivated by two primary considerations:

*  **Reduced Communication Overhead:** Repeating the data on the driver would introduce more communication cost between driver and worker, which is particularly pronounced in our architecture.
*  **Optimized KV Cache Utilization:** Performing request scheduling on pre-repeated DataProto could lead to identical prompts being assigned to different DP rank. This would necessitate redundant KV cache computations, undermining performance. 

Regarding the issues raised in pr[#2324](https://github.com/volcengine/verl/pull/2324), we acknowledge them and have designated their resolution as future work.

### Checklist Before Starting

### Test

### API and Usage Example

You can find a script  example to train `DeepSeek-R1-Distill-Qwen-7B` with 8 GPUs in verl\recipe\req_shced\

And you can enable the Request Skewness Scheduler in your training script with following few simple steps:

```
# Initialize the scheduler once before the training loop
req_scheduler = ReqScheduler()

for batch in dataloader:
	# 1. BEFORE generation: Schedule the batch
	req_scheduler.sched()
	
	# --- Your Existing Code ---
        # The model generates on the efficiently sorted batch
        generated_sequences = model.generate()
        ......
        # ---
	
	# 2. AFTER generation: Restore order and update scheduler state
	req_scheduler.restore_order()
	req_scheduler.update_table # restore order and update table after ROLLOUT
	
        # --- Integration End ---
        # Your existing RL logic (reward calculation, PPO update, etc.)
        # remains completely untouched and works as before.
```

> [!IMPORTANT]

### CC
[hgl71964](https://github.com/hgl71964) , [HyperdriveHustle](https://github.com/HyperdriveHustle), [FortPercent](https://github.com/FortPercent), [qzweng](https://github.com/qzweng)

### DIFF
--- a/recipe/req_sched/main_sched_ppo.py
+++ b/recipe/req_sched/main_sched_ppo.py
@@ -22,6 +22,7 @@ import hydra
 import ray
 from omegaconf import OmegaConf
 
+from verl.trainer.constants_ppo import PPO_RAY_RUNTIME_ENV
 from verl.trainer.ppo.reward import load_reward_manager
 
 from .sched_ray_trainer import RaySchedTrainer
@@ -41,14 +42,7 @@ def run_ppo(config) -> None:
         # NCCL debug level, VLLM logging level, and allow runtime LoRA updating
         # `num_cpus` specifies the number of CPU cores Ray can use, obtained from the configuration
         ray.init(
-            runtime_env={
-                "env_vars": {
-                    "TOKENIZERS_PARALLELISM": "true",
-                    "NCCL_DEBUG": "WARN",
-                    "VLLM_LOGGING_LEVEL": "WARN",
-                    "VLLM_ALLOW_RUNTIME_LORA_UPDATING": "true",
-                }
-            },
+            runtime_env=PPO_RAY_RUNTIME_ENV,
             num_cpus=config.ray_init.num_cpus,
         )
 
@@ -101,17 +95,9 @@ class TaskRunner:
         # Used for multimodal LLM, could be None
         processor = hf_processor(local_path, trust_remote_code=trust_remote_code, use_fast=True)
 
-        # Version validation for vllm.
-        if config.actor_rollout_ref.rollout.name in ["vllm"]:
-            from verl.utils.vllm_utils import is_version_ge
-
-            if config.actor_rollout_ref.model.get("lora_rank", 0) > 0:
-                if not is_version_ge(pkg="vllm", minver="0.7.3"):
-                    raise NotImplementedError("PPO LoRA is not supported before vllm 0.7.3")
-
         # Define worker classes based on the actor strategy.
-        if config.actor_rollout_ref.actor.strategy in ["fsdp", "fsdp2"]:
-            assert config.critic.strategy in ["fsdp", "fsdp2"]
+        if config.actor_rollout_ref.actor.strategy in {"fsdp", "fsdp2"}:
+            assert config.critic.strategy in {"fsdp", "fsdp2"}
             from verl.single_controller.ray import RayWorkerGroup
             from verl.workers.fsdp_workers import AsyncActorRolloutRefWorker, CriticWorker
 
@@ -165,7 +151,7 @@ class TaskRunner:
         # finally, we combine all the rewards together
         # The reward type depends on the tag of the data
         if config.reward_model.enable:
-            if config.reward_model.strategy in ["fsdp", "fsdp2"]:
+            if config.reward_model.strategy in {"fsdp", "fsdp2"}:
                 from verl.workers.fsdp_workers import RewardModelWorker
             elif config.reward_model.strategy == "megatron":
                 from verl.workers.megatron_workers import RewardModelWorker
--- a/recipe/req_sched/main_sched_ppo.py
+++ b/recipe/req_sched/main_sched_ppo.py
@@ -47,7 +47,8 @@ def run_ppo(config) -> None:
 
     # Create a remote instance of the TaskRunner class, and
     # Execute the `run` method of the TaskRunner instance remotely and wait for it to complete
-    if OmegaConf.select(config.trainer, "profile_steps") is not None and len(OmegaConf.select(config.trainer, "profile_steps")) > 0:
+    if (OmegaConf.select(config.trainer, "profile_steps") is not None 
+        and len(OmegaConf.select(config.trainer, "profile_steps")) > 0):
         nsight_options = OmegaConf.to_container(config.trainer.controller_nsight_options)
         runner = TaskRunner.options(runtime_env={"nsight": nsight_options}).remote()
     else:
@@ -79,7 +80,8 @@ class TaskRunner:
 
         # Download the checkpoint from HDFS to the local machine.
         # `use_shm` determines whether to use shared memory, which could lead to faster model loading if turned on
-        local_path = copy_to_local(config.actor_rollout_ref.model.path, use_shm=config.actor_rollout_ref.model.get("use_shm", False))
+        local_path = copy_to_local(
+            config.actor_rollout_ref.model.path, use_shm=config.actor_rollout_ref.model.get("use_shm", False))
 
         # Instantiate the tokenizer and processor.
         from verl.utils import hf_processor, hf_tokenizer
@@ -104,7 +106,11 @@ class TaskRunner:
             from verl.workers.fsdp_workers import AsyncActorRolloutRefWorker, CriticWorker
             from .fsdp_workers import ActorRolloutRefWorker
 
-            actor_rollout_cls = AsyncActorRolloutRefWorker if config.actor_rollout_ref.rollout.mode == "async" else ActorRolloutRefWorker
+            actor_rollout_cls = (
+                AsyncActorRolloutRefWorker 
+                if config.actor_rollout_ref.rollout.mode == "async" 
+                else ActorRolloutRefWorker
+            )
             ray_worker_group_cls = RayWorkerGroup
 
         elif config.actor_rollout_ref.actor.strategy == "megatron":
@@ -112,7 +118,11 @@ class TaskRunner:
             from verl.single_controller.ray.megatron import NVMegatronRayWorkerGroup
             from verl.workers.megatron_workers import ActorRolloutRefWorker, AsyncActorRolloutRefWorker, CriticWorker
 
-            actor_rollout_cls = AsyncActorRolloutRefWorker if config.actor_rollout_ref.rollout.mode == "async" else ActorRolloutRefWorker
+            actor_rollout_cls = (
+                AsyncActorRolloutRefWorker 
+                if config.actor_rollout_ref.rollout.mode == "async" 
+                else ActorRolloutRefWorker
+            )
             ray_worker_group_cls = NVMegatronRayWorkerGroup
 
         else:
@@ -159,8 +169,12 @@ class TaskRunner:
             mapping[Role.RefPolicy] = global_pool_id
 
         # Load the reward manager for training and validation.
-        reward_fn = load_reward_manager(config, tokenizer, num_examine=0, **config.reward_model.get("reward_kwargs", {}))
-        val_reward_fn = load_reward_manager(config, tokenizer, num_examine=1, **config.reward_model.get("reward_kwargs", {}))
+        reward_fn = load_reward_manager(
+            config, tokenizer, num_examine=0, **config.reward_model.get("reward_kwargs", {})
+        )
+        val_reward_fn = load_reward_manager(
+            config, tokenizer, num_examine=1, **config.reward_model.get("reward_kwargs", {})
+        )
         resource_pool_manager = ResourcePoolManager(resource_pool_spec=resource_pool_spec, mapping=mapping)
 
         from verl.utils.dataset.rl_dataset import collate_fn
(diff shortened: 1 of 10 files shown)


recipe/req_sched/sched_ray_trainer.py
```diff
@@ -0,0 +1,857 @@
...
+
+            test_gen_batch_padded, _ = pad_dataproto_to_divisor(test_gen_batch, self.actor_rollout_wg.world_size)
+
+            test_output_gen_batch = self.actor_rollout_wg.generate_sequences(test_gen_batch_padded)
+            
+            self.req_scheduler.restore_order(
+                test_output_gen_batch, 
+                test_reqs_idx, 
+                n_samples=self.config.actor_rollout_ref.rollout.val_kwargs.n
+            )
```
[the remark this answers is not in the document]
Tyizhanshen: This suggestion is incorrect. We've modified the logic in `pad_dataproto_to_divisor`, and the new implementation ensures that the workers internally operate only on unpadded, valid data. Therefore, an extra unpadding step is no longer necessary.

wuxibin89:
> Length Estimation: We maintain a lookup table that records the actual response length for each request from the previous episode (Note the lookup table can be maintained and saved as a json file for feature experiments).

So the `reorder` start from episode=1? In episode=0, we don't have any prior knowledge about how long the response is.

wuxibin89:
> In our design, we repeat DataProto (when n > 1) in rollout workers instead of the driverhttps://github.com/volcengine/verl/pull/2324. This choice is motivated by two primary considerations:

> - Reduced Communication Overhead: Repeating the data on the driver would introduce more communication cost between driver and worker, which is particularly pronounced in our architecture.
> - Optimized KV Cache Utilization: Performing request scheduling on pre-repeated DataProto could lead to identical prompts being assigned to different DP rank. This would necessitate redundant KV cache computations, undermining performance.

If a prompt tend to have long response, should we evenly distribute this prompt across DP ranks instead of repeat it in a single DP?

Tyizhanshen:
> > Length Estimation: We maintain a lookup table that records the actual response length for each request from the previous episode (Note the lookup table can be maintained and saved as a json file for feature experiments).
> 
> So the `reorder` start from episode=1? In episode=0, we don't have any prior knowledge about how long the response is.

Sequence length estimation can be performed offline using the` load_table` and `log_seqlen` methods. If no historical length data is available, the sched function will start from episode 1.

Tyizhanshen:
> > In our design, we repeat DataProto (when n > 1) in rollout workers instead of the driverhttps://github.com/[/pull/2324](https://github.com/volcengine/verl/pull/2324). This choice is motivated by two primary considerations:
> 
> > * Reduced Communication Overhead: Repeating the data on the driver would introduce more communication cost between driver and worker, which is particularly pronounced in our architecture.
> > * Optimized KV Cache Utilization: Performing request scheduling on pre-repeated DataProto could lead to identical prompts being assigned to different DP rank. This would necessitate redundant KV cache computations, undermining performance.
> 
> If a prompt tend to have long response, should we evenly distribute this prompt across DP ranks instead of repeat it in a single DP?

1. From the perspective of our goal—"towards near-equal generated tokens in rollout"—I think the two approaches are equivalent, for we have the flexibility to choose different statistical strategies, such as `max`, `sum`, or `mean`, to define the standard for response length and guide the sched process.
2. We should also consider the impact of an "evenly distribute" strategy on KV cache hit rates. Since similar prompts are more likely to generate identical prefixes, the resulting redundant computation and storage of the same initial key-value pairs may, contrary to expectations, actually slow down inference speed.

wuxibin89:
@Tyizhanshen OK, so does the DAPO-Math-17k experiment utilize offline estimation? Training steps 0~13 should be on episode 0.

Tyizhanshen:
> @Tyizhanshen OK, so does the DAPO-Math-17k experiment utilize offline estimation? Training steps 0~13 should be on episode 0.

@wuxibin89  Yes, we utilize offline estimation in the  DAPO-Math-17k experiment with Request Skewness Scheduler

chenhaiq:
@Tyizhanshen  This is a good idea! There is one more thing you may need to consider:
The target is to balance generation time in each DP group, but generate time is also effect to current items. For example:

If there are 2 concurrent generation in a DP, the token per second may be 2 times than a DP with 1 concurrent if there are enough memory.

A DP with 2 samples with 10 and 20 tokens may runs faster than a DP with 1 sample with 30 tokens. May be it can be improved by combination with sample count and total tokens in a DP.

Tyizhanshen:
> @Tyizhanshen This is a good idea! There is one more thing you may need to consider: The target is to balance generation time in each DP group, but generate time is also effect to current items. For example:
> 
> If there are 2 concurrent generation in a DP, the token per second may be 2 times than a DP with 1 concurrent if there are enough memory.
> 
> A DP with 2 samples with 10 and 20 tokens may runs faster than a DP with 1 sample with 30 tokens. May be it can be improved by combination with sample count and total tokens in a DP.

@chenhaiq Thanks for your comments and questions! Regarding the concern about the mentioned problem, we believe our implemented scheduling methods already address this effectively. The scheduling methods we implemented, `even_token` (greedy partition) and `even_token_kk` (Karmarkar-Karp partition), are inherently designed to make the number of samples assigned to each DP rank as balanced as possible and the long-tail data is distributed evenly among the ranks. Our experimental results also confirm this: the sample distribution across ranks is nearly identical, with the deviation rarely exceeding 3 samples.

|      |      |      | 
| :--: | :--: | ---- | 
| image |  image |image |   

where `len(my_idx)` means the number of samples allocated to the DP rank.

Therefore, from a "concurrent" perspective, we do not anticipate extreme scenarios where a large discrepancy in sample counts between DP ranks would cause a slowdown in inference speed.

Tyizhanshen:
@wuxibin89 @chenhaiq 
Hi, for the two CI checks we failed:
(1)[model_rmpad / model_rmpad (pull_request)](https://github.com/volcengine/verl/actions/runs/16535866424/job/46830048668?pr=2629)

It's strange that this error occurs, because the changes in our PR should not affect this part, and we run the failed test script locally, the result shows that there is no problem

image

(2)[e2e_ppo_trainer / e2e_ppo_trainer_sglang_multiturn_with_tool (pull_request)](https://github.com/volcengine/verl/actions/runs/16535866414/job/46830131375?pr=2629)

There seems to be something wrong with this test? a problem occurs at the initial model download stage
image

recipe/req_sched/main_sched_ppo.py
```diff
@@ -0,0 +1,291 @@
...
+# Define a function to run the PPO-like training process
+def run_ppo(config) -> None:
+    # Check if Ray is not initialized
+    if not ray.is_initialized():
+        # Initialize Ray with a local cluster configuration
+        # Set environment variables in the runtime environment to control tokenizer parallelism,
+        # NCCL debug level, VLLM logging level, and allow runtime LoRA updating
+        # `num_cpus` specifies the number of CPU cores Ray can use, obtained from the configuration
+        ray.init(
+            runtime_env={
```
eric-haibin-lin: these env vars were moved to verl/trainer/constants_ppo.py

recipe/req_sched/main_sched_ppo.py
```diff
@@ -0,0 +1,291 @@
...
+        # Instantiate the tokenizer and processor.
+        from verl.utils import hf_processor, hf_tokenizer
+
+        trust_remote_code = config.data.get("trust_remote_code", False)
+        tokenizer = hf_tokenizer(local_path, trust_remote_code=trust_remote_code)
+        # Used for multimodal LLM, could be None
+        processor = hf_processor(local_path, trust_remote_code=trust_remote_code, use_fast=True)
+
+        # Version validation for vllm.
+        if config.actor_rollout_ref.rollout.name in ["vllm"]:
```
eric-haibin-lin: this can be removed since verl requires vllm version >= 0.7.3 now

recipe/req_sched/sched_ray_trainer.py
```diff
@@ -0,0 +1,890 @@
...
+                            batch.meta_info["multi_turn"] = self.config.actor_rollout_ref.rollout.multi_turn.enable
+                            actor_output = self.actor_rollout_wg.update_actor(batch)
+                        actor_output_metrics = reduce_metrics(actor_output.meta_info["metrics"])
+                        metrics.update(actor_output_metrics)
+
+                    # Log rollout generations if enabled
+                    rollout_data_dir = self.config.trainer.get("rollout_data_dir", None)
+                    if rollout_data_dir:
+                        with marked_timer("dump_rollout_generations", timing_raw, color="green"):
+                            print(batch.batch.keys())
```
eric-haibin-lin: no need to print

recipe/req_sched/sched_ray_trainer.py
```diff
@@ -0,0 +1,890 @@
...
+
+    def _validate(self):
+        data_source_lst = []
+        reward_extra_infos_dict: dict[str, list] = defaultdict(list)
+
+        sample_inputs = []
+        sample_outputs = []
+        sample_scores = []
+
+        for test_data in self.val_dataloader:
```
eric-haibin-lin, line 110: general comments: 
could you highlight the key changes in this file that enable the scheduler via leaving comments in the code? This will greatly help others adopting your changes in their own trainer subclasses. 
e.g. the places where `self.req_scheduler.sched`, `self.req_scheduler.restore_order`, `self.req_scheduler.log_seqlen` are called

recipe/req_sched/sched_ray_trainer.py
```diff
@@ -0,0 +1,890 @@
...
+                    non_tensor_batch_keys=non_tensor_batch_keys_to_pop,
+                )
+
+                is_last_step = self.global_steps >= self.total_training_steps
+
+                idx = gen_batch.batch["input_ids"]  # (bs, prompt_length)
+                reqs_idx = gen_batch.non_tensor_batch["reqs_idx"]
+                raw_prompt_ids = gen_batch.non_tensor_batch["raw_prompt_ids"]  # (bs, varlen)
+                batch.non_tensor_batch["raw_prompt_ids"] = raw_prompt_ids
+                print(
```
eric-haibin-lin: would this produce too many logs?

recipe/req_sched/sched_ray_trainer.py
```diff
@@ -0,0 +1,890 @@
...
+            # Get only the tokens before padding
+            original_response = padded_response[:pad_start_idx]
+        except ValueError:
+            # No padding found, use the full response
+            original_response = padded_response
+
+        unpadded_responses.append(original_response)
+    return unpadded_responses
+
+
```
eric-haibin-lin: it's better to move ReqScheduler to a separate file

Tyizhanshen: Thanks for all your helpful suggestions. I'm working on incorporating all of them, including the other points you raised.

Tyizhanshen:
@eric-haibin-lin I've pushed the latest changes. The code has been tidied up, and to make the review easier, I've highlighted the key design changes with comments prefixed by [ReqScheduler].

Please let me know if you have any other questions or feedback!

Tyizhanshen:
The CI is failing with two errors that appear to be unrelated to my changes: a `requests.exceptions.ProxyError` and an unexpected OOM error in the Ulysses GPU tests. All other checks have passed, and I haven't touched the relevant code for these failing steps. Could a maintainer please take a look? @eric-haibin-lin @wuxibin89 @chenhaiq

PR #2700, opened 2025-07-23
Revert "[fsdp,megatron,sglang] fix: Fix torch reduce to speed up update weights"

hebiao064:
Reverts volcengine/verl#2692 since https://github.com/sgl-project/sglang/pull/8267 is merged yet, we need to wait for 8267 and sglang release


vermouth1992:
We don't need to revert it as it doesn't break the code

PR #2739, opened 2025-07-24
[recipe] feat:  Add sleep/wakeup mode for gen rm vllm service and add tqdm  showing process

none0663:
### What does this PR do?

Add sleep/wakeup mode for gen rm vllm service and add tqdm  showing process.

This capability is particularly beneficial when the model server shares resources with a training workload on the same machine. It allows the reward model service to be temporarily offloaded (to free up GPU memory) during intensive training sessions and reloaded when the service is required again.

### DIFF
--- a/recipe/genrm_remote/README.md
+++ b/recipe/genrm_remote/README.md
@@ -7,8 +7,16 @@
 Deploy the pretrained GenRM model using vLLM. Skip this step if you want to use an external api service.
 
 ```bash 
-vllm serve verl-team/GenRM-CI-Test-1.5B --served-model-name genrm-demo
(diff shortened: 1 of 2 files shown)


wuxibin89:
For external reward service, I think they should always use separate GPU resource?

none0663:
> For external reward service, I think they should always use separate GPU resource?

Yes, it’s generally best practice to use dedicated GPU resources for the external reward service when resources are plentiful. However, when GPU resources are constrained, especially considering that many GPUs allocated for training might be idle during reward model scoring, we prefer to utilize those idle GPU resources for the reward service.

wuxibin89 (approved):

ETOgaosion:
This PR seems to break the genrm test. @none0663 @wuxibin89 

Suprisingly, Itself didn't test this.

For https://github.com/volcengine/verl/pull/2794 to solve all CI faults in main, we may revert this first and add some robustness check.

PR #2775, opened 2025-07-28
[algo] feat: add GSPO policy loss computation function [WIP]

0x404:
### What does this PR do?

This PR implements the GSPO policy loss calculation proposed by paper https://arxiv.org/pdf/2507.18071

### Test

Work in progress.

Will compare the effectiveness and training curves of GRPO and GSPO under the same settings.

### API and Usage Example

To use GSPO, users only need to set `actor_rollout_ref.actor.policy_loss.loss_mode` to `gspo`.

```shell
  ... \
  actor_rollout_ref.actor.policy_loss.loss_mode="gspo" \
  ...
```

> [!IMPORTANT]

### DIFF
--- a/verl/trainer/ppo/core_algos.py
+++ b/verl/trainer/ppo/core_algos.py
@@ -31,6 +31,7 @@ from omegaconf import DictConfig
 import verl.utils.torch_functional as verl_F
 from verl.trainer.config import AlgoConfig
 from verl.utils.import_utils import deprecated
+from verl.workers.config import ActorConfig
 
 PolicyLossFn = Callable[
     [
@@ -882,6 +883,63 @@ def compute_policy_loss_vanilla(
 
     pg_losses = torch.where(advantages < 0, clip_pg_losses2, clip_pg_losses1)
     pg_loss = agg_loss(loss_mat=pg_losses, loss_mask=response_mask, loss_agg_mode=loss_agg_mode)
+
+    return pg_loss, pg_clipfrac, ppo_kl, pg_clipfrac_lower
+
+
+@register_policy_loss("gspo")
+def compute_policy_loss_gspo(
+    old_log_prob: torch.Tensor,
+    log_prob: torch.Tensor,
+    advantages: torch.Tensor,
+    response_mask: torch.Tensor,
+    loss_agg_mode: str = "token-mean",
+    config: Optional[DictConfig | ActorConfig] = None,
+) -> tuple[torch.Tensor, torch.Tensor, torch.Tensor, torch.Tensor]:
+    """
+    Compute the clipped policy objective and related metrics for GSPO.
+
+    See https://arxiv.org/pdf/2507.18071 for more details.
+
+    Args:
+        old_log_prob (torch.Tensor):
+            Log-probabilities of actions under the old policy, shape (batch_size, response_length).
+        log_prob (torch.Tensor):
+            Log-probabilities of actions under the current policy, shape (batch_size, response_length).
+        advantages (torch.Tensor):
+            Advantage estimates for each action, shape (batch_size, response_length).
+        response_mask (torch.Tensor):
+            Mask indicating which tokens to include in the loss, shape (batch_size, response_length).
+        loss_agg_mode (str, optional):
+            Aggregation mode for `agg_loss`. Defaults to "token-mean".
+    """
+
+    assert config is not None
+    assert not isinstance(config, ActorConfig)
+    clip_ratio = config.clip_ratio
+    clip_ratio_low = config.clip_ratio_low if config.clip_ratio_low is not None else clip_ratio
+    clip_ratio_high = config.clip_ratio_high if config.clip_ratio_high is not None else clip_ratio
+
+    # compute sequence-level importance ratio
+    seq_lengths = torch.sum(response_mask, dim=-1)
+    normalized_seq_log_prob = torch.sum(log_prob * response_mask, dim=-1) / seq_lengths
+    normalized_old_seq_log_prob = torch.sum(old_log_prob * response_mask, dim=-1) / seq_lengths
+    negative_approx_kl_seq = normalized_seq_log_prob - normalized_old_seq_log_prob
+    seq_ratio = torch.exp(negative_approx_kl_seq)
+
+    pg_losses1 = -advantages * seq_ratio
+    pg_losses2 = -advantages * torch.clamp(seq_ratio, 1 - clip_ratio_low, 1 + clip_ratio_high)
+    pg_losses = torch.maximum(pg_losses1, pg_losses2)
+    pg_loss = agg_loss(loss_mat=pg_losses, loss_mask=response_mask, loss_agg_mode=loss_agg_mode)
+
+    # For compatibility, return zero for pg_clipfrac_lower (not used in standard GSPO)
+    pg_clipfrac = verl_F.masked_mean(torch.gt(pg_losses2, pg_losses1).float(), response_mask)
+    pg_clipfrac_lower = torch.tensor(0.0, device=pg_loss.device)
+
+    # compute token-level mean kl divergence metric for compatibility
+    negative_approx_kl = log_prob - old_log_prob
+    negative_approx_kl = torch.clamp(negative_approx_kl, min=-20.0, max=20.0)
+    ppo_kl = verl_F.masked_mean(-negative_approx_kl, response_mask)
 
     return pg_loss, pg_clipfrac, ppo_kl, pg_clipfrac_lower
 
--- a/verl/trainer/ppo/core_algos.py
+++ b/verl/trainer/ppo/core_algos.py
@@ -915,10 +915,9 @@ def compute_policy_loss_gspo(
     """
 
     assert config is not None
-    assert not isinstance(config, ActorConfig)
-    clip_ratio = config.clip_ratio
-    clip_ratio_low = config.clip_ratio_low if config.clip_ratio_low is not None else clip_ratio
-    clip_ratio_high = config.clip_ratio_high if config.clip_ratio_high is not None else clip_ratio
+    assert isinstance(config, ActorConfig)
+    clip_ratio_low = config.clip_ratio_low if config.clip_ratio_low is not None else config.clip_ratio
+    clip_ratio_high = config.clip_ratio_high if config.clip_ratio_high is not None else config.clip_ratio
 
     # compute sequence-level importance ratio
     seq_lengths = torch.sum(response_mask, dim=-1)
@@ -926,20 +925,24 @@ def compute_policy_loss_gspo(
     normalized_old_seq_log_prob = torch.sum(old_log_prob * response_mask, dim=-1) / seq_lengths
     negative_approx_kl_seq = normalized_seq_log_prob - normalized_old_seq_log_prob
     seq_ratio = torch.exp(negative_approx_kl_seq)
+    seq_ratio_sg = seq_ratio.detach()  # stop gradient
 
-    pg_losses1 = -advantages * seq_ratio
-    pg_losses2 = -advantages * torch.clamp(seq_ratio, 1 - clip_ratio_low, 1 + clip_ratio_high)
+    negative_approx_kl = log_prob - old_log_prob
+    negative_approx_kl = torch.clamp(negative_approx_kl, min=-20.0, max=20.0)
+    ppo_kl = verl_F.masked_mean(-negative_approx_kl, response_mask)
+    token_ratio = torch.exp(negative_approx_kl)  # (batch_size, response_length)
+
+    seq_ratio_expanded = seq_ratio_sg.unsqueeze(1)  # (batch_size, 1)
+    combined_ratio = seq_ratio_expanded * token_ratio  # (batch_size, response_length)
+
+    pg_losses1 = -advantages * combined_ratio
+    pg_losses2 = -advantages * torch.clamp(combined_ratio, 1 - clip_ratio_low, 1 + clip_ratio_high)
     pg_losses = torch.maximum(pg_losses1, pg_losses2)
     pg_loss = agg_loss(loss_mat=pg_losses, loss_mask=response_mask, loss_agg_mode=loss_agg_mode)
 
     # For compatibility, return zero for pg_clipfrac_lower (not used in standard GSPO)
     pg_clipfrac = verl_F.masked_mean(torch.gt(pg_losses2, pg_losses1).float(), response_mask)
     pg_clipfrac_lower = torch.tensor(0.0, device=pg_loss.device)
-
-    # compute token-level mean kl divergence metric for compatibility
-    negative_approx_kl = log_prob - old_log_prob
-    negative_approx_kl = torch.clamp(negative_approx_kl, min=-20.0, max=20.0)
-    ppo_kl = verl_F.masked_mean(-negative_approx_kl, response_mask)
 
     return pg_loss, pg_clipfrac, ppo_kl, pg_clipfrac_lower
 
--- a/tests/special_e2e/ppo_trainer/run_function_reward.sh
+++ b/tests/special_e2e/ppo_trainer/run_function_reward.sh
@@ -28,6 +28,7 @@ RM_PAD=${RM_PAD:-True}
 FUSED_KERNELS=${FUSED_KERNELS:-False}
 FUSED_KERNEL_BACKEND=${FUSED_KERNEL_BACKEND:-torch} # or 'triton' for triton backend
 ADV_ESTIMATOR=${ADV_ESTIMATOR:-gae}
+LOSS_MODE=${LOSS_MODE:-vanilla}
 USE_KL=${USE_KL:-False}
 CUSTOM_REWARD_FN=${CUSTOM_REWARD_FN:-False}
 ENABLE_CHUNKED_PREFILL=${ENABLE_CHUNKED_PREFILL:-True} # For vLLM VLM placeholder issue: https://github.com/vllm-project/vllm/issues/15185
@@ -112,6 +113,7 @@ python3 -m verl.trainer.main_ppo \
     actor_rollout_ref.actor.ulysses_sequence_parallel_size="${SP_SIZE}" \
     actor_rollout_ref.actor.checkpoint.save_contents=${CHECKPOINT_CONTENTS} \
     actor_rollout_ref.actor.use_kl_loss="${USE_KL}" \
+    actor_rollout_ref.actor.policy_loss.loss_mode="${LOSS_MODE}" \
     actor_rollout_ref.rollout.log_prob_micro_batch_size_per_gpu=${train_traj_micro_bsz_per_gpu} \
     actor_rollout_ref.rollout.tensor_model_parallel_size=2 \
     actor_rollout_ref.rollout.name="${ENGINE}" \


verl/trainer/ppo/core_algos.py
```diff
@@ -886,6 +887,65 @@ def compute_policy_loss_vanilla(
...
+    negative_approx_kl_seq = torch.sum(negative_approx_kl * response_mask, dim=-1) / seq_lengths
+    seq_ratio = torch.exp(negative_approx_kl_seq)
+    seq_ratio_sg = seq_ratio.detach()  # stop gradient
+
+    negative_approx_kl = torch.clamp(negative_approx_kl, min=-20.0, max=20.0)
+    ppo_kl = verl_F.masked_mean(-negative_approx_kl, response_mask)
+    token_ratio = torch.exp(negative_approx_kl)  # (batch_size, response_length)
+
+    seq_ratio_expanded = seq_ratio_sg.unsqueeze(1)  # (batch_size, 1)
+    combined_ratio = seq_ratio_expanded * token_ratio  # (batch_size, response_length)
```
mokeevdmitrii: I believe this formula is incorrect, is should be (according to the article) seq_ratio_expanded * torch.exp(log_prob) / torch.exp(log_prob).detach()

Isn't it easier just to:

log_ratio = negative_approx_kl_seq.detach() + log_prob - log_prob.detach()
combined_ratio = torch.exp(torch.clamp(log_ratio, -20, 20))

0x404: Yes, you are absolutely right! Thanks for pointing this out! I mistakenly read the $sg[\pi_\theta(y_{i,t}|x_i, y_{i,<t})]$ in the paper as $sg[\pi_{\theta_{old}}(y_{i,t}|x_i, y_{i,<t})]$.

BounharAbdelaziz: @mokeevdmitrii  I was about to point to the same thing!
Also, if I can add on it, I think that for numerical stability @0x404 , it is better to do the clamp() before calculation the importance ratio, i.e.
```python
# Calculate log ratio for each token
log_ratio = log_prob - old_log_prob
log_ratio = torch.clamp(log_ratio, min=-20.0, max=20.0)  # CLAMP FIRST

# Compute sequence-level log importance ratio using CLAMPED values
seq_lengths = torch.sum(response_mask, dim=-1).clamp(min=1)
seq_log_ratio = torch.sum(log_ratio * response_mask, dim=-1) / seq_lengths
````
Instead of 
```python
negative_approx_kl = log_prob - old_log_prob  # UNCLAMPED

# Compute sequence-level log importance ratio using UNCLAMPED values
seq_lengths = torch.sum(response_mask, dim=-1)
negative_approx_kl_seq = torch.sum(negative_approx_kl * response_mask, dim=-1) / seq_lengths.clamp(min=1)

# Only later do we clamp (but too late for sequence ratio calculation)
negative_approx_kl = torch.clamp(negative_approx_kl, min=-20.0, max=20.0)
```

vermouth1992:
Could you create a folder under recipe named gspo, and add two recipe script using QWen2.5 7b math and Qwen3 30b a3b following this?
- https://github.com/volcengine/verl/blob/main/recipe/dapo/test_dapo_7b_math.sh
- https://github.com/volcengine/verl/blob/main/recipe/dapo/test_dapo_qwen3_30b_math.sh

verl/trainer/ppo/core_algos.py
0x404: Great advice!

0x404:
> Could you create a folder under recipe named gspo, and add two recipe script using QWen2.5 7b math and Qwen3 30b a3b following this?

Of course, working on it.

verl/trainer/ppo/core_algos.py
BounharAbdelaziz: @0x404 can you give access to modify the code, I worked on the vanilla version of GSPO, not sue if it's better to just open a PR for that too or just make the same for GSPO-"vanilla" and GSPO-token.

BounharAbdelaziz: got it, thanks @0x404. will submitt asap. just need to make sure it is running fine.

0x404: Hi @BounharAbdelaziz, I've added the access, you can push those changes to this branch.

But I have a question: since the current vanilla policy loss is already token-level (the advantage is a (batch, seq_length) tensor), I think we might only need to implement GSPO-token, because GSPO-token and GSPO are equivalent (when the advantage has equal elements across the seq_length dimension). what do you think?

verl/trainer/ppo/core_algos.py
```diff
@@ -886,6 +887,64 @@ def compute_policy_loss_vanilla(
...
+    log_seq_importance_ratio = negative_approx_kl_seq.detach().unsqueeze(-1) + log_prob - log_prob.detach()
+
+    # Clamp log_seq_importance_ratio for stability
+    log_seq_importance_ratio = torch.clamp(log_seq_importance_ratio, min=-20.0, max=20.0)
+    seq_importance_ratio = torch.exp(log_seq_importance_ratio)
+
+    pg_losses1 = -advantages * seq_importance_ratio
+    pg_losses2 = -advantages * torch.clamp(seq_importance_ratio, 1 - clip_ratio_low, 1 + clip_ratio_high)
+    pg_losses = torch.maximum(pg_losses1, pg_losses2)
+    pg_loss = agg_loss(loss_mat=pg_losses, loss_mask=response_mask, loss_agg_mode=loss_agg_mode)
```
vermouth1992: Here agg_loss should always be sentence-level?

0x404: Yes, "seq-mean-token-sum" looks most suitable. But from the paper's definition, maybe a "batch-mean-group-mean-token-sum" would be more accurate? Which sums the token loss of seq, then takes a mean of group seq loss at the group-level, then takes a mean/sum again at the batch level. there is a similar discussion in https://github.com/volcengine/verl/issues/2776

BounharAbdelaziz:
Just added the implementation of the sequence level optimization (GSPO vanilla) and improved the GSPO-token implementation by removing some redundant clamp for negative_approx_kl (at, I believe wrong late step). 

I also added some details about the equations we are calculating at each step.

Will be sharing some xp results once ready.

verl/trainer/ppo/core_algos.py
mokeevdmitrii: This is GSPO-token, so it should be **exactly** seq-mean-token-mean.

EDIT: hope i haven't hallucinated in my nightly calcs

The example in #2776 is about token-mean vs seq-mean).

Referencing the article:

image

1) First, we calc $s_i(\theta)$ and find token-level loss (inside the last sum, let's name it $pg_{i,t}$.

2) Second, we sum this loss over every sequence and divide it by sequence length $|y_i|$. This equals to:
```
seq_losses = torch.sum(loss_mat * loss_mask, dim=-1) / torch.sum(loss_mask, dim=-1)
```
in current verl implementation for seq-mean-token-mean. Let's name this $pg_{i}$.

3) Finally, we notice that there is a common denumerator, as the loss for a single group is $\dfrac{1}{G} \sum\limits_{i=1}^G pg_i$.  So, "batch-mean-group-mean-token-mean" is not needed, as each "group-mean" operation has the same denumerator - this is equiv. to ""batch-group"-mean-token-mean" or simply "seq-mean-token-mean".

Notes:

1) seq-mean-token-sum would be wrong, as we are summing up $s_i(\theta)$ for **each** token in GSPO-token, so we must divide by seq-len here.

2) the fact that token-mean is a valid loss_agg_mode value here makes me a bit sad bcs the article is about sequence-level optimization)

mokeevdmitrii:
This PR uses default cliprange settings (0.2 [default as I am writing this comment](https://github.com/volcengine/verl/blob/4de3ecf0f0fa2963d88f46a9fb6d51ad8b71452d/verl/trainer/config/actor/actor.yaml#L33), however the authors mention
```
that the clipping ranges in GSPO and
in previous algorithms (e.g., GRPO) typically differ in order of magnitude due to the distinct definitions
of importance ratios
```
I think this is worth testing if possible and (at least) mentioning this fact to the user of "loss_mode: gspo"

BounharAbdelaziz:
> This PR uses default cliprange settings (0.2 [default as I am writing this comment](https://github.com/volcengine/verl/blob/4de3ecf0f0fa2963d88f46a9fb6d51ad8b71452d/verl/trainer/config/actor/actor.yaml#L33), however the authors mention
> 
> ```
> that the clipping ranges in GSPO and
> in previous algorithms (e.g., GRPO) typically differ in order of magnitude due to the distinct definitions
> of importance ratios
> ```
> 
> I think this is worth testing if possible and (at least) mentioning this fact to the user of "loss_mode: gspo"

Indeed, it’s recommended (https://x.com/ChujieZheng/status/1948933507696525392) to be between 3e-4 and 4e-4.

verl/trainer/ppo/core_algos.py
0x404: @mokeevdmitrii You are right! we should use seq-mean-token-mean when using GSPO, thanks very much!

verl/trainer/ppo/core_algos.py
```diff
@@ -886,6 +887,139 @@ def compute_policy_loss_vanilla(
...
+    negative_approx_kl_seq = torch.sum(negative_approx_kl * response_mask, dim=-1) / seq_lengths
+
+    # Get sequence-level importance ratio: s_i(θ) = (π_θ(y_i|x)/π_θold(y_i|x))^(1/|y_i|)
+    seq_importance_ratio = torch.exp(negative_approx_kl_seq)
+
+    # Expand sequence ratio to match token dimensions: each token in a sequence gets the same ratio
+    seq_ratio_expanded = seq_importance_ratio.unsqueeze(-1).expand_as(log_prob)
+
+    pg_losses1 = -advantages * seq_ratio_expanded
+    pg_losses2 = -advantages * torch.clamp(seq_ratio_expanded, 1 - clip_ratio_low, 1 + clip_ratio_high)
```
0x404: Hi @BounharAbdelaziz, do you think we really need to keep both GSPO and GSPO-token?

For sentence-level GSPO, it shouldn't need `seq_importance_ratio` to do expand, but rather directly multiply with sentence-level advantage. However, our advantage here is token-level.

I think we might only need to implement GSPO-token, because GSPO-token and GSPO are equivalent (when the advantage has equal elements across the seq_length dimension). What do you think?

0x404:
image

> I think this is worth testing if possible

results based on commit 4b247b2a8207ff306be11b015b0e1b468dc934e6, GSPO using 4e-4 cliprange settings:

```sh
    data.train_files=$HOME/data/gsm8k/train.parquet \
    data.val_files=$HOME/data/gsm8k/test.parquet \
    data.train_batch_size=512 \
    data.max_prompt_length=512 \
    data.truncation='error' \
    actor_rollout_ref.model.path=Qwen/Qwen2.5-3B-Instruct \
    actor_rollout_ref.actor.ppo_mini_batch_size=128 \
    actor_rollout_ref.actor.ppo_micro_batch_size_per_gpu=40 \
    actor_rollout_ref.actor.kl_loss_coef=0.001 \
    actor_rollout_ref.actor.policy_loss.loss_mode="gspo_token" \
    actor_rollout_ref.actor.clip_ratio_low=0.0004 \
    actor_rollout_ref.actor.clip_ratio_high=0.0004 \
    actor_rollout_ref.actor.clip_ratio=0.0004 \
    actor_rollout_ref.rollout.log_prob_micro_batch_size_per_gpu=40 \
    actor_rollout_ref.rollout.n=10 \
    actor_rollout_ref.ref.log_prob_micro_batch_size_per_gpu=40 \
    trainer.critic_warmup=0 \
    trainer.project_name='verl_gspo_cmp' \
    trainer.experiment_name='qwen2.5-3B-GSPO-clip1e-4' \
    trainer.n_gpus_per_node=8 \
    trainer.nnodes=1 \
    trainer.save_freq=20 \
    trainer.test_freq=5 \
    trainer.total_epochs=15 $@
```

verl/trainer/ppo/core_algos.py
chujiezheng: I prefer GSPO-token, as it is theoretically equal to GSPO and offers higher flexibility for future potential features

chujiezheng:
Thanks for your support of the GSPO algorithm!

We have updated the experimental details in the GSPO paper: https://arxiv.org/abs/2507.18071

acd0cacf9b3d9ad5311d444cfdcce005

I would suggest to try the following hyperparameters:

- `actor_rollout_ref.actor.clip_ratio_low=3e-4`
- `actor_rollout_ref.actor.clip_ratio_high=4e-4`

You may also try to turn off the KL loss by setting `actor_rollout_ref.actor.kl_loss_coef=0`

0x404:
@chujiezheng Thanks, will try it out!

verl/trainer/ppo/core_algos.py
```diff
@@ -886,6 +887,139 @@ def compute_policy_loss_vanilla(
...
+
+    assert config is not None
+    assert isinstance(config, ActorConfig)
+    clip_ratio_low = config.clip_ratio_low if config.clip_ratio_low is not None else config.clip_ratio
+    clip_ratio_high = config.clip_ratio_high if config.clip_ratio_high is not None else config.clip_ratio
+
+    # Calculate log ratio for each token: log(π_θ(y_i,t|x,y_i,<t)/π_θold(y_i,t|x,y_i,<t))
+    negative_approx_kl = log_prob - old_log_prob
+    # Clamp negative_approx_kl for numerical stability
+    negative_approx_kl = torch.clamp(negative_approx_kl, min=-20.0, max=20.0)
```
Yelrose, line 930: Why should we clamp this dividing seq_lengths?

verl/trainer/ppo/core_algos.py
BounharAbdelaziz: Hi @0x404, you are right. They fall into the same and you can technically make GSPO-token=GSPO by setting A_{i,t}=A_i. I'm pushing the update.

Thank you @chujiezheng for your input!

BounharAbdelaziz: @0x404 I see that you already did remove it, thanks!

BounharAbdelaziz:
I did run an experiment using **GSM8K** on **GSPO-token** vs **GRPO** using the following "core" parameters:

GSPO-token:
```bash
clip_ratio_low=0.0003 # as recommended by the paper, see Sec. 5.1
clip_ratio_high=0.0004 # as recommended by the paper, see Sec. 5.1
train_batch_size=512
ppo_mini_batch_size=128 # maintain 4 mini-batches as recommended by the paper, see Sec. 5.1
ppo_micro_batch_size_per_gpu=8 # setup depending on your GPU memory
n_resp_per_prompt=16

use_kl_in_reward=false
kl_coef=0.0
use_kl_loss=false
kl_loss_coef=0.0

max_prompt_length=$((1024 * 2))
max_response_length=$((1024 * 8)) # yes this is a lot, and actually the response length mean converges towards 25x
```

GRPO shares the same setup with the main differences being:
```bash
clip_ratio_low=0.2
clip_ratio_high=0.28
```

**Observations**

Both algorithms tend to perform very similarly on the test set. Though, it seems like GRPO is learning a bit faster. 

When examining the pg clipping factor, we can observe that it's huge sitting at 30% (double the one mentionned in the paper) and stabelizes around 10% while GRPO almost never clips, 0.03% on average. I believe that one need to tune the $\epsilon$ carrefully, perhaps @chujiezheng can share some insights on this?

With the above observations, I am currently running two other experiments for GSPO-token with `clip_ratio_high=6e-4`and  `clip_ratio_high=3.5e-4`. The goal is to assess whether allowing more clipping might help, or if tighter clipping thresholds improve performance, though higher thresholds could increase the clipping factor even further.

Logs are attached for reference.

Screenshot 2025-07-29 at 22 34 45

Screenshot 2025-07-29 at 22 35 35

Screenshot 2025-07-29 at 22 35 52

chujiezheng:
Thanks for your test.

Based on the feedback from friends in other labs, the advantage of GSPO over GRPO (w/ Routing Replay) is mainly manifested on **MoE** models, while they perform similarly on dense models.

We did not do much test for GSPO on dense models internally. But our experience shows that dense models are much easier to RL train. So the results you obtain may be expected.

chujiezheng:
Also, we did not do test in the zero setting, and we previously found that zero RL is also much easier than in the cold-start setting (perhaps because the response length in the zero setting is usually shorter).

verl/trainer/ppo/core_algos.py
```diff
@@ -886,6 +887,73 @@ def compute_policy_loss_vanilla(
...
```
vermouth1992: This clamp is unnecessary? Could you please confirm that @chujiezheng Thanks!

verl/trainer/ppo/core_algos.py
```diff
@@ -886,6 +887,73 @@ def compute_policy_loss_vanilla(
...
+    # exp [(1/|y_i|) * Σ_t log(π_θ(y_i,t|x,y_i,<t)/π_θold(y_i,t|x,y_i,<t))]
+    seq_lengths = torch.sum(response_mask, dim=-1).clamp(min=1)
+    negative_approx_kl_seq = torch.sum(negative_approx_kl * response_mask, dim=-1) / seq_lengths
+
+    # Combined ratio at token level:
+    # s_i,t(θ) = sg[s_i(θ)] · π_θ(y_i,t|x, y_i,<t) / sg[π_θ(y_i,t|x, y_i,<t)]
+    # In log space: log(s_i,t(θ)) = sg[log(s_i(θ))] + log_prob - sg[log_prob]
+    log_seq_importance_ratio = negative_approx_kl_seq.detach().unsqueeze(-1) + log_prob - log_prob.detach()
+
+    # finaly exp() to remove log
```
vermouth1992, line 936: We probably need to clamp here before exp

verl/trainer/ppo/core_algos.py
```diff
@@ -886,6 +887,73 @@ def compute_policy_loss_vanilla(
...
+    seq_importance_ratio = torch.exp(log_seq_importance_ratio)
+
+    pg_losses1 = -advantages * seq_importance_ratio
+    pg_losses2 = -advantages * torch.clamp(seq_importance_ratio, 1 - clip_ratio_low, 1 + clip_ratio_high)
+    pg_losses = torch.maximum(pg_losses1, pg_losses2)
+
+    # for GSPO, we need to aggregate the loss at the sequence level (seq-mean-token-mean)
+    pg_loss = agg_loss(loss_mat=pg_losses, loss_mask=response_mask, loss_agg_mode="seq-mean-token-mean")
```
vermouth1992, line 944: Can we hardcore agg_loss to seq-mean-token-mean?

verl/trainer/ppo/core_algos.py
syt-nju: In order to reuse the clip ratio from other configurations, the clipping here should maintain dimensional consistency with the original clip formula

verl/trainer/ppo/core_algos.py
chujiezheng: Agreed it is unnecessary here

verl/trainer/ppo/core_algos.py
```diff
@@ -886,6 +887,73 @@ def compute_policy_loss_vanilla(
...
+    # compute sequence-level importance ratio:
+    # si(θ) = (π_θ(yi|x)/π_θold(yi|x))^(1/|yi|) =
+    # exp [(1/|y_i|) * Σ_t log(π_θ(y_i,t|x,y_i,<t)/π_θold(y_i,t|x,y_i,<t))]
+    seq_lengths = torch.sum(response_mask, dim=-1).clamp(min=1)
+    negative_approx_kl_seq = torch.sum(negative_approx_kl * response_mask, dim=-1) / seq_lengths
+
+    # Combined ratio at token level:
+    # s_i,t(θ) = sg[s_i(θ)] · π_θ(y_i,t|x, y_i,<t) / sg[π_θ(y_i,t|x, y_i,<t)]
+    # In log space: log(s_i,t(θ)) = sg[log(s_i(θ))] + log_prob - sg[log_prob]
+    log_seq_importance_ratio = negative_approx_kl_seq.detach().unsqueeze(-1) + log_prob - log_prob.detach()
```
chujiezheng: You should change the calculation order to avoid precision error:

```
log_seq_importance_ratio = log_prob - log_prob.detach() + negative_approx_kl_seq.detach().unsqueeze(-1)
```

verl/trainer/ppo/core_algos.py
chujiezheng: > We probably need to clamp here before exp

Agreed. The above clamp should be moved here:

```
log_seq_importance_ratio = torch.clamp(log_seq_importance_ratio, max=10.0)
seq_importance_ratio = torch.exp(log_seq_importance_ratio)
```

verl/trainer/ppo/core_algos.py
0x404: Thanks @chujiezheng ! all resolved.

BounharAbdelaziz:
> Also, we did not do test in the zero setting, and we previously found that zero RL is also much easier than in the cold-start setting (perhaps because the response length in the zero setting is usually shorter).

Great! Thanks a lot for sharing your insights!

vermouth1992:
I guess this PR should be ready for merge?

vermouth1992 (approved):

BounharAbdelaziz:
> I did run an experiment using **GSM8K** on **GSPO-token** vs **GRPO** using the following "core" parameters:
> 
> GSPO-token:
> 
> ```shell
> clip_ratio_low=0.0003 # as recommended by the paper, see Sec. 5.1
> clip_ratio_high=0.0004 # as recommended by the paper, see Sec. 5.1
> train_batch_size=512
> ppo_mini_batch_size=128 # maintain 4 mini-batches as recommended by the paper, see Sec. 5.1
> ppo_micro_batch_size_per_gpu=8 # setup depending on your GPU memory
> n_resp_per_prompt=16
> 
> use_kl_in_reward=false
> kl_coef=0.0
> use_kl_loss=false
> kl_loss_coef=0.0
> 
> max_prompt_length=$((1024 * 2))
> max_response_length=$((1024 * 8)) # yes this is a lot, and actually the response length mean converges towards 25x
> ```
> 
> GRPO shares the same setup with the main differences being:
> 
> ```shell
> clip_ratio_low=0.2
> clip_ratio_high=0.28
> ```
> 
> **Observations**
> 
> Both algorithms tend to perform very similarly on the test set. Though, it seems like GRPO is learning a bit faster.
> 
> When examining the pg clipping factor, we can observe that it's huge sitting at 30% (double the one mentionned in the paper) and stabelizes around 10% while GRPO almost never clips, 0.03% on average. I believe that one need to tune the ϵ carrefully, perhaps @chujiezheng can share some insights on this?
> 
> With the above observations, I am currently running two other experiments for GSPO-token with `clip_ratio_high=6e-4`and `clip_ratio_high=3.5e-4`. The goal is to assess whether allowing more clipping might help, or if tighter clipping thresholds improve performance, though higher thresholds could increase the clipping factor even further.
> 
> Logs are attached for reference.
> 
> Screenshot 2025-07-29 at 22 34 45 Screenshot 2025-07-29 at 22 35 35 Screenshot 2025-07-29 at 22 35 52

Sharing my experiments with the following two extra test for GSPO-token:

```bash
clip_ratio_low=0.0003
clip_ratio_high=0.00035
```


```bash
clip_ratio_low=0.0003
clip_ratio_high=0.0006
```

**Results**: no noticible difference. This joins the shared insights by @chujiezheng, i.e. GSPO is more suitable for MoEs.

Screenshot 2025-07-30 at 12 41 25

Screenshot 2025-07-30 at 12 42 02

Screenshot 2025-07-30 at 12 42 16

Screenshot 2025-07-30 at 12 44 48

xihuai18:
Can someone explain why `pg_clipfrac_lower` is not used?

```pythoon

    # For compatibility, return zero for pg_clipfrac_lower (not used in standard GSPO)
    pg_clipfrac = verl_F.masked_mean(torch.gt(pg_losses2, pg_losses1).float(), response_mask)
    pg_clipfrac_lower = torch.tensor(0.0, device=pg_loss.device)

```

TristanMeng:
should use Dual-clip ?

PR #2784, opened 2025-07-28
WIP: add vllm pipeline parallel support for zmq executor

eric-haibin-lin:
### What does this PR do?

This PR enhances verl's vllm executor with pipeline parallel support with vllm >= 0.9.1
Note that the following patches is required for vllm (tested with vllm 0.9.1): vllm/engine/arg_utils.py:L1433
```
        if self.pipeline_parallel_size > 1:
            supports_pp = hasattr(self.distributed_executor_backend, 'supports_pp') and self.distributed_executor_backend.supports_pp
            if not supports_pp and self.distributed_executor_backend not in (ParallelConfig.distributed_executor_backend, "ray", "mp", "external_launcher"):
                name = "Pipeline Parallelism without Ray distributed executor " \
                        "or multiprocessing executor or external launcher"
                _raise_or_fallback(feature_name=name, recommend_to_remove=False)
                return False
```

### Checklist Before Starting

### Test

Tested via the following script with pipeline_parallel_size = 2
```
python3 tests/workers/rollout/perf/vllm_async_rollout.py
```
Further convergence test is required. 

### DIFF
--- a/verl/workers/rollout/vllm_rollout/vllm_async_server.py
+++ b/verl/workers/rollout/vllm_rollout/vllm_async_server.py
@@ -14,6 +14,7 @@
 import logging
 import os
 import pickle
+from concurrent.futures import Future
 from typing import Any, Callable, Optional
 
 import ray
@@ -31,6 +32,7 @@ from vllm.inputs import TokensPrompt
 from vllm.outputs import RequestOutput
 from vllm.v1.engine.async_llm import AsyncLLM
 from vllm.v1.executor.abstract import Executor
+from vllm.v1.outputs import ModelRunnerOutput
 from vllm.worker.worker_base import WorkerWrapperBase
 
 from verl.utils.fs import copy_to_local
@@ -40,7 +42,6 @@ logger = logging.getLogger(__file__)
 
 
 def _get_model_runner_workers(vllm_config, init_ray: bool = True):
-    print("XXX vllm_config", vllm_config)
     assert vllm_config.instance_id is not None, "instance_id must be set for external ray actors."
 
     fields = vllm_config.instance_id.split(":")
@@ -111,7 +112,6 @@ class ExternalRayDistributedExecutor(Executor):
             sent_method = method
         else:
             sent_method = pickle.dumps(method)
-        print("XXX Executing", method, kwargs, flush=True)
         del method
         # ~3ms overhead per schedule step due to SchedulerOutput/ModelRunnerOutput serialization/deserialization.
         outputs = ray.get(
@@ -126,10 +126,12 @@ class ExternalRayDistributedExecutor(Executor):
     def max_concurrent_batches(self) -> int:
         return self.parallel_config.pipeline_parallel_size
 
+
 class ExternalZeroMQDistributedExecutor(Executor):
     """An executor that engines are launched by external ray actors."""
 
     uses_ray: bool = False
+    supports_pp: bool = True
 
     def _init_executor(self) -> None:
         addresses = os.environ["VERL_VLLM_ZMQ_ADDRESSES"].split(",")
@@ -139,7 +141,6 @@ class ExternalZeroMQDistributedExecutor(Executor):
             socket = self.context.socket(zmq.REQ)
             socket.connect(address)
             self.sockets.append(socket)
-        print("XXX self.sockets", self.sockets)
 
         kwargs = dict(
             vllm_config=self.vllm_config,
@@ -152,7 +153,7 @@ class ExternalZeroMQDistributedExecutor(Executor):
         self.collective_rpc("init_device")
         self.collective_rpc("load_model")
 
-        from concurrent.futures import Future, ThreadPoolExecutor
+        from concurrent.futures import ThreadPoolExecutor
 
         # For pipeline parallel, we use a thread pool for asynchronous
         # execute_model.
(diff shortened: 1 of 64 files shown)


verl/workers/rollout/vllm_rollout/vllm_async_server.py
```diff
@@ -147,31 +149,98 @@ def _init_executor(self) -> None:
...
         if isinstance(method, str):
             sent_method = method
         else:
             sent_method = pickle.dumps(method)
+        kwargs = {} if kwargs is None else kwargs
         del method
 
-        message = pickle.dumps((sent_method, args, kwargs or {}))
+        message = pickle.dumps((sent_method, args, kwargs, unique_reply_rank))
+        # TODO(haibin.lin): optimize with only necessary ranks
```
wuxibin89, line 195: For TP+PP, is it only need to send to workers of first PP stage?

verl/workers/rollout/vllm_rollout/vllm_async_server.py
```diff
@@ -147,31 +149,98 @@ def _init_executor(self) -> None:
...
+        def recv_response(socket: zmq.Socket):
+            val = pickle.loads(socket.recv())
+            return val
+
         outputs = []
-        for socket in self.sockets:
-            outputs.append(pickle.loads(socket.recv()))
+        for i, socket in enumerate(self.sockets):
+            if non_block:
+                val = self.io_thread_pool.submit(recv_response, socket)
```
wuxibin89, line 206: `val` is a `concurrent.Future`, who is responsible for getting the result?

eric-haibin-lin: vllm batched schedule

verl/workers/rollout/vllm_rollout/vllm_async_server.py
eric-haibin-lin: I followed mp executor which sends it to all ranks ..

youkaichao: I think we send them to all workers, so that the batchsize (number of tokens in the current batch) can be available for all workers. and then they can prepare corresponding sized buffer for send/recv intermediate tensors.

snie2012:
@eric-haibin-lin Hi Eric, I am trying to run the unit test python3 tests/workers/rollout/perf/vllm_async_rollout.py with pp=2, it failed with multiple errors. Is the current status of this PR supporting pp > 1 already? Thanks.

eric-haibin-lin:
> @eric-haibin-lin Hi Eric, I am trying to run the unit test python3 tests/workers/rollout/perf/vllm_async_rollout.py with pp=2, it failed with multiple errors. Is the current status of this PR supporting pp > 1 already? Thanks.

what error did you get? I ran it with vllm 0.9.1 and have to change vllm source code mentioned in the PR description (also created PR https://github.com/vllm-project/vllm/pull/21786), it runs without error but I yet need to verify end2end convergence

snie2012:
> > @eric-haibin-lin Hi Eric, I am trying to run the unit test python3 tests/workers/rollout/perf/vllm_async_rollout.py with pp=2, it failed with multiple errors. Is the current status of this PR supporting pp > 1 already? Thanks.
> 
> what error did you get? I ran it with vllm 0.9.1 and have to change vllm source code mentioned in the PR description (also created PR [vllm-project/vllm#21786](https://github.com/vllm-project/vllm/pull/21786)), it runs without error but I yet need to verify end2end convergence

I am using vllm 0.8.5 (Is this version supported? I followed the standard verl installation guideline.). And I am testing with mode async and backend zeromq. I suppose async+ray is not supported since ExternalRayDistributedExecutor doesn't have supports_pp=True?

It errors out in the collective_rpc function of ExternalZeroMQDistributedExecutor. Specifically the call to socket.send(message, zmq.DONTWAIT) failed with zmq.error.ZMQError: Operation cannot be accomplished in current state.

Looking at the stack trace, the error probably originates from output = self.model_runner.execute_model(scheduler_output) in vllm/v1/worker/gpu_model_runner.py:L1079.

eric-haibin-lin:
No vllm 0.8.5 does not have the code for v1 multi-process PP execution, that's why I upgraded to v0.9.1

eric-haibin-lin:
BTW I only added the code for zmq executor

snie2012:
> No vllm 0.8.5 does not have the code for v1 multi-process PP execution, that's why I upgraded to v0.9.1

Thanks! What would be the right way to upgrade v0.9.1? I tried both 1) directly upgrade in my current verl conda env 2) fresh installation of verl env by modifying the vllm version inside install_vllm_sglang_mcore.sh.

Both approaches failed unfortunately due to package missing or incompatible issues.

snie2012:
I managed to install vllm 0.9.1, and the unit test works with pp=1, but it still fails with pp=2. The error occurs at vllm/v1/engine/output_processor:L52 - the output is an Exception.

```
if isinstance(output, Exception):
     raise output
```

acisseJZhong:
thanks @eric-haibin-lin for adding the PP support for vllm! After patching your PR and running async vllm engine with PP >1, I encountered ```zmq.error.ZMQError: Operation cannot be accomplished in current state``` in https://github.com/volcengine/verl/blob/e74038c3566c26e3151904fb6e2f800541b6dea2/verl/workers/rollout/vllm_rollout/vllm_async_server.py#L197

running the async rollout test gives me the same error. Wondering do you have any suggestions to debug? thanks!

snie2012:
@eric-haibin-lin Turns out the issue is with the unblocking socket.recv. Because socket.recv is unblocking, socket.send can be called twice before socket.recv is called, which causes the zeromq error. We have a local fix that's working and it might be good idea to have a good design to make sure socket.send always happens after another socket.recv.

eric-haibin-lin:
@snie2012 oh thanks! Would you like to PR your changes? Yes, I agree we should design it such that zmq.send calls are not done consecutively

snie2012:
> @snie2012 oh thanks! Would you like to PR your changes? Yes, I agree we should design it such that zmq.send calls are not done consecutively

Sure, should I base the new PR on top of your PR?

snie2012:
@eric-haibin-lin new PR https://github.com/volcengine/verl/pull/2851

PR #2815, opened 2025-07-30
[misc] chore: Use UID for grouping in validation to avoid prompt confusion in multimodal tasks

xs1997zju:
Main Changes
1. Background Explanation
In multimodal tasks (such as image-text understanding or generation), it is common to have samples with different images but the same text prompt. An example of this can be found in the ‘geometry3k’ dataset, where the recurring text prompt is “Find X”.   If validation grouping is based only on the prompt, samples with the same prompt but different images will be mixed together in evaluation. This leads to confused and inaccurate evaluation results, as the model’s performance on each unique sample (image+prompt pair) cannot be correctly measured. Such confusion can mask the model’s true ability in image understanding or generation, and reduce the reliability of multimodal evaluation.
2. Main Changes
* In the validation (eval) process, grouping is now based on datasource+uid instead of prompt, ensuring each sample is uniquely and accurately evaluated.
* Each test_batch sample is assigned a unique uid before repeat; after repeat, the uid is automatically expanded accordingly.
* The sample_inputs argument for process_validation_metrics is now a uid list instead of a prompt list; variable names inside the function remain unchanged and logic is fully compatible.
* This change does not affect unimodal tasks, and makes evaluation for multimodal tasks safer and more accurate.

### DIFF
--- a/verl/trainer/ppo/ray_trainer.py
+++ b/verl/trainer/ppo/ray_trainer.py
@@ -630,6 +630,7 @@ class RayPPOTrainer:
 
         # Lists to collect samples for the table
         sample_inputs = []
+        sample_uids = []
         sample_outputs = []
         sample_scores = []
         sample_turns = []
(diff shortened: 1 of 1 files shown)


wuxibin89:
@xs1997zju Please format code with following instruction:
https://github.com/volcengine/verl/blob/main/CONTRIBUTING.md#code-linting-and-formatting

PR #2869, opened 2025-08-01
[sglang] chore: bump transformer formers 4.54.0 and fix QWen VL issues

hebiao064:
### What does this PR do?

Do not merge before: https://github.com/volcengine/verl/pull/2720

> Add **concise** overview of what this PR aims to achieve or accomplish. Reference related GitHub issues and PRs that help with the review.
Bump xformers, fixing patched model code
### Checklist Before Starting

- [x] Search for similar PRs. Paste at least one query link here: ...
- [ ] Format the PR title as `[{modules}] {type}: {description}` (This will be checked by the CI)
  - `{modules}` include `fsdp`, `megatron`, `sglang`, `vllm`, `rollout`, `trainer`, `ci`, `training_utils`, `recipe`, `hardware`, `deployment`, `ray`, `worker`, `single_controller`, `misc`, `perf`, `model`, `algo`, `env`, `tool`, `ckpt`, `doc`, `data`
  - If this PR involves multiple modules, separate them with `,` like `[megatron, fsdp, doc]`
  - `{type}` is in `feat`, `fix`, `refactor`, `chore`, `test`
  - If this PR breaks any API (CLI arguments, config, function signature, etc.), add `[BREAKING]` to the beginning of the title.
  - Example: `[BREAKING][fsdp, megatron] feat: dynamic batching`

### Test

> For changes that can not be tested by CI (e.g., algorithm implementation, new model support), validate by experiment(s) and show results like training curve plots, evaluation results, etc.

### API and Usage Example

> Demonstrate how the API changes if any, and provide usage example(s) if possible.

```python
# Add code snippet or script demonstrating how to use this
```

### Design & Code Changes

> Demonstrate the high-level design if this PR is complex, and list the specific changes.

### Checklist Before Submitting

> [!IMPORTANT]
> Please check all the following items before requesting a review, otherwise the reviewer might deprioritize this PR for review.

- [ ] Read the [Contribute Guide](https://github.com/volcengine/verl/blob/main/CONTRIBUTING.md).
- [ ] Apply [pre-commit checks](https://github.com/volcengine/verl/blob/main/CONTRIBUTING.md#code-linting-and-formatting): `pre-commit install && pre-commit run --all-files --show-diff-on-failure --color=always`
- [ ] Add / Update [the documentation](https://github.com/volcengine/verl/tree/main/docs).
[...]

PR #2882, opened 2025-08-03
Fix: Initialize _flash_attn_uses_top_left_mask for transformers >= v4.53.0

clpbc:
## 🐛 Fix: Initialize `_flash_attn_uses_top_left_mask` for `transformers >= 4.53.0` Compatibility

This PR fixes a runtime `AttributeError` that occurs when using `transformers >= 4.53.0`, where the internal attribute `_flash_attn_uses_top_left_mask` is no longer automatically set in `Qwen2_5_VLAttention`.

### 🧾 Problem

In `transformers` versions prior to `v4.53.0` (e.g., `v4.52.4`), the `Qwen2_5_VLAttention` class automatically initialized the attribute `self._flash_attn_uses_top_left_mask`. However, starting from `v4.53.0`, this attribute is no longer set by default in `Qwen2_5_VLAttention` class, and users are expected to determine mask behavior via the utility function:

```python
from transformers.modeling_flash_attention_utils import flash_attn_supports_top_left_mask
self._flash_attn_uses_top_left_mask = flash_attn_supports_top_left_mask()

```

The original code in `verl` assumed this attribute existed, leading to the following error:

```
AttributeError: 'Qwen2_5_VLAttention' object has no attribute '_flash_attn_uses_top_left_mask'
```

This breaks Flash Attention functionality when using newer versions of `transformers`.

### ✅ Solution

We now explicitly initialize the flag using the provided utility:

```python
from transformers.modeling_flash_attention_utils import flash_attn_supports_top_left_mask
self._flash_attn_uses_top_left_mask = flash_attn_supports_top_left_mask()
```

This ensures compatibility across `transformers` versions, both older and newer.

### 🔍 Affected File

- `verl/models/transformers/qwen2_vl.py` 

### 🧪 Testing

- Verified locally with:
  - `transformers==4.52.4` ✅
  - `transformers==4.53.0` ✅
  - `transformers==4.53.1` ✅
  - `transformers==4.53.2` ✅
  - `transformers==4.53.3` ✅
  - `transformers==4.54.0` ✅
- No functional change in behavior — only restores correctness under newer `transformers` versions.

### 📚 Reference

- Hugging Face Transformers Changelog: https://github.com/huggingface/transformers/releases
- Flash Attention Utility: https://github.com/huggingface/transformers/blob/main/src/transformers/modeling_flash_attention_utils.py

### 🎯 Impact

Enables `verl` to work seamlessly with modern versions of `transformers`, improving maintainability and reducing version lock-in.

---
> ✅ **Note**: This is a minimal, non-breaking fix that aligns with upstream changes in `transformers`. Ready for review!

### DIFF
--- a/verl/models/transformers/qwen2_vl.py
+++ b/verl/models/transformers/qwen2_vl.py
@@ -284,6 +284,10 @@ def ulysses_flash_attn_forward(
     else:
         sliding_window = None
 
+    from transformers.modeling_flash_attention_utils import flash_attn_supports_top_left_mask
+
+    self._flash_attn_uses_top_left_mask = flash_attn_supports_top_left_mask()
+
     attn_output = flash_attention_forward(
         query_states,
         key_states,


vermouth1992:
Does this PR already cover the issue? https://github.com/volcengine/verl/pull/2869/files

clpbc:
Yes, This PR is already cover this issue. Thank you! @vermouth1992

Issue #2618, opened 2025-07-18
[agentic RL] multi-turn rollout and agent loop development tracking


wuxibin89:
As [Agent Loop](https://verl.readthedocs.io/en/latest/advance/agent_loop.html) is added in v0.4.2, we're moving to a more general interface for agentic RL. Agent Loop idea original from https://github.com/volcengine/verl/issues/1172, during development inspired by various community's effort (especially from SGLang):

- [Tool](https://verl.readthedocs.io/en/latest/sglang_multiturn/multiturn.html): define tool schema and how to call tool
- [Interaction](https://verl.readthedocs.io/en/latest/sglang_multiturn/interaction_system.html#): define more general way to interact with environment than tool
- [ChatScheduler](https://github.com/volcengine/verl/blob/main/verl/workers/rollout/chat_scheduler.py): early prototype to call LLM in server mode

With Agent Loop, tool and interaction can be implemented as a agent loop class and chat scheduler should be removed. We plan to refactor the code in next 2 releases.

**Refactor**

- [ ] Tool is implemented as [ToolAgentLoop](https://github.com/volcengine/verl/blob/main/verl/experimental/agent_loop/tool_agent_loop.py), move `multi_turn` config in rollout.yaml to a separate config file for ToolAgentLoop in v0.5.1
- [x] ChatScheduler: removed in v0.5.1 https://github.com/volcengine/verl/pull/2725
- [x] Clean up tool and interaction code in `sglang_rollout.py` v0.5.1 https://github.com/volcengine/verl/pull/3171
- [ ] Move all tool implementation to recipe and clean them up eventually.
- [ ] Deprecate rollout offline batch mode, and switch to serve mode, see discussion in [#3102](https://github.com/volcengine/verl/pull/3102), [#2854](https://github.com/volcengine/verl/pull/2854#issuecomment-3168220058) https://github.com/volcengine/verl/pull/3161

**Performance**
- [x] Pad agent loop output after `agent_loop.run` instead of `_postprocess` https://github.com/volcengine/verl/pull/2773
- [x] Move reward calculation into `agent_loop.run` since reward may be evaluate by interaction with environment, e.g swe-agent. https://github.com/volcengine/verl/pull/3055

@eric-haibin-lin @vermouth1992 @zhaochenyang20 @SwordFaith @kinza99 @chenhaiq @zw0610

Kirrito-k423:
[...]

PR #2896, opened 2025-08-04
[BREAKING][rollout,sglang] feat: switch to agent loop if multi_turn.enable=True

wuxibin89:
### What does this PR do?

As described in https://github.com/volcengine/verl/issues/2618, this is first step to clean up `sglang_rollout.py`. 

SGLang multi_turn functionality is moved to two new agent loop:

- ToolAgentLoop: if ``multi_turn.tool_config_path`` is set.
- InteractionAgentLoop: if ``multi_turn.interaction_config_path`` is set.

Set ``multi_turn.enable=True`` will automatically switch to new agent loop by default, and should work same as the old implementation.

For backward compatibility, we provide an option ``multi_turn.force_use_sglang_multi_turn=False`` to switch back to old implementation, but be aware that it will be soon removed in future releases.

### DIFF
--- a/verl/trainer/main_ppo.py
+++ b/verl/trainer/main_ppo.py
@@ -113,12 +113,16 @@ class TaskRunner:
         pprint(OmegaConf.to_container(config, resolve=True))
         OmegaConf.resolve(config)
 
-        if config.actor_rollout_ref.rollout.multi_turn.enable:
+        if (
+            config.actor_rollout_ref.rollout.multi_turn.enable
+            and config.actor_rollout_ref.rollout.mode != "async"
+            and not config.actor_rollout_ref.rollout.multi_turn.force_use_sglang_multi_turn
+        ):
             warnings.warn(
                 "The setting multi_turn.enable=True is deprecated. Please use rollout.mode=async instead. "
                 "If you want to use the old behavior, please set multi_turn.force_use_sglang_multi_turn=True. "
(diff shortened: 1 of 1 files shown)


zhaochenyang20:
Do you have time to arrange a meeting for multi-turn issue?

verl/experimental/agent_loop/agent_loop.py
```diff
@@ -283,7 +284,18 @@ async def generate_sequences(self, batch: DataProto) -> DataProto:
...
         # by default, we assume it's a single turn agent
         if "agent_name" not in batch.non_tensor_batch:
-            batch.non_tensor_batch["agent_name"] = np.array(["single_turn_agent"] * len(batch), dtype=object)
+            warnings.warn(
+                "agent_name is not set in RLHFDataset, set to 'tool_agent' if tool_config_path is set, "
+                "otherwise default to 'single_turn_agent'",
+                category=UserWarning,
+                stacklevel=2,
+            )
+            if config.multi_turn.tool_config_path is not None:
```
chenhaiq, line 293: how to use react_agent_loop.py with tool?

wuxibin89: This is case where user doesn't set `agent_name` explicitly in RLHFDataset, it's just for seamless migration from slang multi_turn to agent loop. For user who already use agent loop, they should set `agent_name` in dataset to choose which loop to use.

Issue #2640, opened 2025-07-20
More of torch.compile use: the actor model itself, and the loss, ref model


vadimkantorov:
Currently only some small function is torch.compile'd.

Might be that compat of torch.compile with FSDP2 can be checked in torchtune/torchtitan.

vadimkantorov:
@vermouth1992 torch.compile-ing the actor model itself (under FSDP and FSDP2) should also reduce the discrepancy with vllm rollouts, as vllm typically uses torch.compile

Kirrito-k423:
### 中文
定时清理检查发现：此 issue 最近一次非当前运行账号的公开回复或引用时间是 `2025-08-14T21:17:27Z`（cross-referenced，来自 `vadimkantorov`），相对于本轮运行日期 `2026-07-12` 已超过 6 个月。因此按社区 issue 定时清理标准关闭。

已检查证据：Latest non-self public reply/reference: `2025-08-14T21:17:27Z` (cross-referenced by `vadimkantorov`).
Comments API: https://api.github.com/repos/verl-project/verl/issues/2640/comments
Timeline API: https://api.github.com/repos/verl-project/verl/issues/2640/timeline

如果作者或社区成员仍能提供新的复现、证据或需求背景，请在此 issue 下继续评论请求重开，或创建新的 issue 并链接到这里。

### English
Scheduled cleanup check found that the latest public reply or reference by an account other than the current automation account was `2025-08-14T21:17:27Z` (cross-referenced by `vadimkantorov`), which is more than 6 months before this run date `2026-07-12`. Therefore this issue is being closed under the scheduled stale-issue cleanup rule.

Evidence checked: Latest non-self public reply/reference: `2025-08-14T21:17:27Z` (cross-referenced by `vadimkantorov`).
Comments API: https://api.github.com/repos/verl-project/verl/issues/2640/comments
Timeline API: https://api.github.com/repos/verl-project/verl/issues/2640/timeline

If the author or community members can still provide a fresh reproduction, evidence, or requirement context, please comment here to request reopening, or open a new issue and link back to this one.

This response was generated by a scheduled Codex GPT5.5 issue processing task.

vadimkantorov:
It's sad to have such issues being auto-closed without any proper discussion...

PR #2953, opened 2025-08-07
[BREAKING][vllm, fsdp] feat: add Rollout-Training Mismatch Fix -- Truncated importance sampling

yaof20:
### What does this PR do?

Support [vLLM-FSDP off-policy importance sampling correction](https://fengyao.notion.site/off-policy-rl) using Truncated Importance Sampling (TIS):

TIS

### Checklist Before Starting

### Test

### API and Usage Example

```python
    algorithm.adv_estimator=gae \
    data.train_files="$train_files" \
    data.val_files="$test_files" \
    data.train_batch_size=1024 \
    data.max_prompt_length=1024 \
    data.truncation='error' \
    actor_rollout_ref.model.path=Qwen/Qwen2.5-32B-Instruct \
    actor_rollout_ref.model.enable_gradient_checkpointing=False \
    actor_rollout_ref.actor.ppo_micro_batch_size_per_gpu=8 \
    actor_rollout_ref.actor.use_kl_loss=False \
    actor_rollout_ref.rollout.log_prob_micro_batch_size_per_gpu=16 \
    actor_rollout_ref.rollout.tensor_model_parallel_size=4 \
    actor_rollout_ref.rollout.gpu_memory_utilization=0.5 \
    critic.optim.lr=1e-5 \
    critic.model.use_remove_padding=True \
    critic.model.path=Qwen/Qwen2.5-32B-Instruct \
    critic.model.enable_gradient_checkpointing=False \
    critic.ppo_micro_batch_size_per_gpu=8 \
    critic.model.fsdp_config.param_offload=False \
    critic.model.fsdp_config.optimizer_offload=False \
    trainer.critic_warmup=0 \
    trainer.project_name='verl_example' \
    trainer.experiment_name='Qwen2.5-32B-Instruct_function_rm' \
    trainer.n_gpus_per_node=8 \
    trainer.nnodes=4 \
    trainer.save_freq=20 \
    trainer.test_freq=10 \
    trainer.total_epochs=15 \
    actor_rollout_ref.rollout.calculate_log_probs=True \   # add this config to return rollout prob
    +actor_rollout_ref.actor.behav_imp_weight_cap=10.0$@   # add this config to set up C value in TIS
```

### Design & Code Changes

> [!IMPORTANT]


yaof20:
Hi VeRL Team,

This feature has been incorporated in OpenRLHF: https://github.com/OpenRLHF/OpenRLHF/releases/tag/v0.8.9

Thanks!

zdhNarsil:
# Why $k_3$ is a biased KL estimator and how to fix it

In this PR, apart from our proposed truncated importance sampling, we also include a fix for KL estimator in [this commit](https://github.com/yaof20/verl/commit/3d98e744b1c6954159c9f5c86b0ba5fd6afaab4e).

In John Schulman's [blog](http://joschu.net/blog/kl-approx.html), he introduces a KL estimator called '' $k_3$ estimator'' for $\text{KL}(p_\theta \| p_{\text{base}})$:

$$k_3(x) = r(x) - 1 - \log r(x), \quad r(x) = \frac{p_{\text{base}}(x)}{p_\theta(x)}$$

This is a valid estimator because one can easily show:

$$𝔼_{p_\theta(x)}[k_3(x)] = \text{KL}(p_\theta \| p_{\text{base}})$$

For example, the DeepSeek team uses this estimator in their reinforcement learning training to obtain the R1 model.

However, we point out that this provides invalid (i.e., biased) gradient signal for model training; in fact, one can show:

$$
𝔼_{p_\theta(x)}\left[\nabla_\theta k_3(x)\right] = 𝔼_{p_\theta(x)}[\nabla_\theta (\frac{p_{\text{base}}(x)}{p_\theta(x)}) - \nabla_\theta(\log \frac{p_{\text{base}}(x)}{p_\theta(x)}) ] 
$$

$$
= 𝔼_{p_\theta(x)}\left[\nabla_\theta \left(\frac{p_{\text{base}}(x)}{p_\theta(x)}\right) \right] = 𝔼_{p_\theta(x)}\left[-\frac{p_{\text{base}}(x)}{p_\theta(x)} \nabla_\theta \log p_\theta(x) \right] 
$$

$$
\neq 𝔼_{p_\theta(x)}\left[-\log\frac{p_{\text{base}}(x)}{p_\theta(x)} \nabla_\theta \log p_\theta(x) \right] = \nabla_\theta\text{KL}(p_\theta \| p_{\text{base}})
$$

To get an unbiased gradient estimate of $\text{KL}(p_\theta \| p_{\text{base}})$, consider the $k_2$ estimator from the same blog:

$$k_2(x) = \frac{1}{2}\left(\log\frac{p_{\text{base}}(x)}{p_\theta(x)} \right)^2$$

$$\Rightarrow 𝔼_{p_\theta(x)}[\nabla_\theta k_2(x)] = 𝔼_{p_\theta(x)}\left[-\log\frac{p_{\text{base}}(x)}{p_\theta(x)} \nabla_\theta \log p_\theta(x) \right] = \nabla_\theta\text{KL}(p_\theta \| p_{\text{base}})$$

## How to fix

As a result, when we need the value of KL (e.g., to be added to the reward terms), we should use $k_1$ or $k_3$; when we need to compute the gradient of KL (e.g., as a training loss term), we should use $k_2$.

In our [code commit](https://github.com/yaof20/verl/commit/3d98e744b1c6954159c9f5c86b0ba5fd6afaab4e), we use a straight-through estimator trick to implement this idea.

wuxibin89:
Need vllm fix https://github.com/vllm-project/vllm/pull/22387 to return `final_logprobs`

verl/workers/actor/dp_actor.py
```diff
@@ -376,6 +376,8 @@ def update_policy(self, data: DataProto):
         ]
         if self.config.use_kl_loss:
             select_keys.append("ref_log_prob")
+        if self.config.imp_ratio_cap > 0:
+            select_keys.append("rollout_log_probs")
```
wuxibin89, line 386: Server mode(agent loop) hasn't return `rollout_log_probs` for now.

verl/trainer/config/rollout/rollout.yaml
```diff
@@ -159,7 +159,8 @@ multi_turn:
   format: hermes
 
 # support logging rollout prob for debugging purpose
-calculate_log_probs: False
+# "Truncated importance sampling" requires rollout log probs, set to True when turning on Truncated importance sampling
+calculate_log_probs: True
```
PeterSH6: Could you turn the default value to False and add a running script for TIS?

verl/trainer/config/actor/actor.yaml
```diff
@@ -71,6 +71,9 @@ loss_agg_mode: token-mean
 # Entropy regularization coefficient in PPO loss
 entropy_coeff: 0
 
+# whether to apply the truncated Importance Sampling (-1 for no importance sampling)
+imp_ratio_cap: 5
```
PeterSH6: Same here

PeterSH6 (commented):
Great work!

verl/workers/actor/dp_actor.py
yaof20: Hi, I have add a check here before adding `rollout_log_probs`.

verl/trainer/config/rollout/rollout.yaml
yaof20: I fixed it now and added one running scripts :)

verl/trainer/config/actor/actor.yaml
yaof20: fixed :)

verl/trainer/ppo/core_algos.py
```diff
@@ -747,6 +747,8 @@ def compute_policy_loss(
     cliprange_high=None,
     clip_ratio_c=3.0,
     loss_agg_mode: str = "token-mean",
+    rollout_log_probs=None,
+    imp_ratio_cap=-1,
 ):
```
eric-haibin-lin: this function is deprecated and you should change verl.trainer.ppo.core_algos.compute_policy_loss_vanilla instead

yaof20: icic, I have included change in change in `verl.trainer.ppo.core_algos.compute_policy_loss_vanilla` as well. 

Let me delete the change in the `compute_policy_loss`.

yaof20: I have fixed now :)

verl/workers/actor/dp_actor.py
```diff
@@ -435,6 +438,8 @@ def update_policy(self, data: DataProto):
                         response_mask=response_mask,
                         loss_agg_mode=loss_agg_mode,
                         config=self.config,
+                        rollout_log_probs=rollout_log_probs,
+                        imp_ratio_cap=self.config.imp_ratio_cap,
```
eric-haibin-lin: no need to pass it as it's included in the config already

jianfeng-Liu:
> ###



> # Why k 3 is a biased KL estimator and how to fix it
> In this PR, apart from our proposed truncated importance sampling, we also include a fix for KL estimator in [this commit](https://github.com/yaof20/verl/commit/3d98e744b1c6954159c9f5c86b0ba5fd6afaab4e).
> 
> In John Schulman's [blog](http://joschu.net/blog/kl-approx.html), he introduces a KL estimator called '' k 3 estimator'' for K L ( p θ | p b a s e ) :
> 
> k 3 ( x ) = r ( x ) − 1 − l o g ⁡ r ( x ) , r ( x ) = p b a s e ( x ) p θ ( x )
> 
> This is a valid estimator because one can easily show:
> 
> 𝔼 p θ ( x ) [ k 3 ( x ) ] = K L ( p θ | p b a s e )
> 
> For example, the DeepSeek team uses this estimator in their reinforcement learning training to obtain the R1 model.
> 
> However, we point out that this provides invalid (i.e., biased) gradient signal for model training; in fact, one can show:
> 
> 𝔼 p θ ( x ) [ ∇ θ k 3 ( x ) ] = 𝔼 p θ ( x ) [ ∇ θ ( p b a s e ( x ) p θ ( x ) ) − ∇ θ ( l o g ⁡ p b a s e ( x ) p θ ( x ) ) ]
> 
> = 𝔼 p θ ( x ) [ ∇ θ ( p b a s e ( x ) p θ ( x ) ) ] = 𝔼 p θ ( x ) [ − p b a s e ( x ) p θ ( x ) ∇ θ l o g ⁡ p θ ( x ) ]
> 
> ≠ 𝔼 p θ ( x ) [ − l o g ⁡ p b a s e ( x ) p θ ( x ) ∇ θ l o g ⁡ p θ ( x ) ] = ∇ θ K L ( p θ | p b a s e )
> 
> To get an unbiased gradient estimate of K L ( p θ | p b a s e ) , consider the k 2 estimator from the same blog:
> 
> k 2 ( x ) = 1 2 ( l o g ⁡ p b a s e ( x ) p θ ( x ) ) 2
> 
> ⇒ 𝔼 p θ ( x ) [ ∇ θ k 2 ( x ) ] = 𝔼 p θ ( x ) [ − l o g ⁡ p b a s e ( x ) p θ ( x ) ∇ θ l o g ⁡ p θ ( x ) ] = ∇ θ K L ( p θ | p b a s e )
> 
> ## How to fix
> As a result, when we need the value of KL (e.g., to be added to the reward terms), we should use k 1 or k 3 ; when we need to compute the gradient of KL (e.g., as a training loss term), we should use k 2 .
> 
> In our [code commit](https://github.com/yaof20/verl/commit/3d98e744b1c6954159c9f5c86b0ba5fd6afaab4e), we use a straight-through estimator trick to implement this idea.

𝔼 p θ ( x ) [ ∇ θ k 3 ( x ) ]  should be equal to

> # Why k 3 is a biased KL estimator and how to fix it
> In this PR, apart from our proposed truncated importance sampling, we also include a fix for KL estimator in [this commit](https://github.com/yaof20/verl/commit/3d98e744b1c6954159c9f5c86b0ba5fd6afaab4e).
> 
> In John Schulman's [blog](http://joschu.net/blog/kl-approx.html), he introduces a KL estimator called '' k 3 estimator'' for KL ( p θ | p base ) :
> 
> k 3 ( x ) = r ( x ) − 1 − log ⁡ r ( x ) , r ( x ) = p base ( x ) p θ ( x )
> 
> This is a valid estimator because one can easily show:
> 
> 𝔼 p θ ( x ) [ k 3 ( x ) ] = KL ( p θ | p base )
> 
> For example, the DeepSeek team uses this estimator in their reinforcement learning training to obtain the R1 model.
> 
> However, we point out that this provides invalid (i.e., biased) gradient signal for model training; in fact, one can show:
> 
> 𝔼 p θ ( x ) [ ∇ θ k 3 ( x ) ] = 𝔼 p θ ( x ) [ ∇ θ ( p base ( x ) p θ ( x ) ) − ∇ θ ( log ⁡ p base ( x ) p θ ( x ) ) ]
> 
> = 𝔼 p θ ( x ) [ ∇ θ ( p base ( x ) p θ ( x ) ) ] = 𝔼 p θ ( x ) [ − p base ( x ) p θ ( x ) ∇ θ log ⁡ p θ ( x ) ]
> 
> ≠ 𝔼 p θ ( x ) [ − log ⁡ p base ( x ) p θ ( x ) ∇ θ log ⁡ p θ ( x ) ] = ∇ θ KL ( p θ | p base )
> 
> To get an unbiased gradient estimate of KL ( p θ | p base ) , consider the k 2 estimator from the same blog:
> 
> k 2 ( x ) = 1 2 ( log ⁡ p base ( x ) p θ ( x ) ) 2
> 
> ⇒ 𝔼 p θ ( x ) [ ∇ θ k 2 ( x ) ] = 𝔼 p θ ( x ) [ − log ⁡ p base ( x ) p θ ( x ) ∇ θ log ⁡ p θ ( x ) ] = ∇ θ KL ( p θ | p base )
> 
> ## How to fix
> As a result, when we need the value of KL (e.g., to be added to the reward terms), we should use k 1 or k 3 ; when we need to compute the gradient of KL (e.g., as a training loss term), we should use k 2 .
> 
> In our [code commit](https://github.com/yaof20/verl/commit/3d98e744b1c6954159c9f5c86b0ba5fd6afaab4e), we use a straight-through estimator trick to implement this idea.

𝔼 p θ ( x ) [ ∇ θ k 3 ( x ) ] should be equal to  𝔼 p θ ( x ) [ − p base ( x ) p θ ( x ) ∇ θ log ⁡ p θ ( x ) + ∇ θ log ⁡p θ ( x )   ]

jianfeng-Liu:
> # Why k 3 is a biased KL estimator and how to fix it
> In this PR, apart from our proposed truncated importance sampling, we also include a fix for KL estimator in [this commit](https://github.com/yaof20/verl/commit/3d98e744b1c6954159c9f5c86b0ba5fd6afaab4e).
> 
> In John Schulman's [blog](http://joschu.net/blog/kl-approx.html), he introduces a KL estimator called '' k 3 estimator'' for K L ( p θ | p b a s e ) :
> 
> k 3 ( x ) = r ( x ) − 1 − l o g ⁡ r ( x ) , r ( x ) = p b a s e ( x ) p θ ( x )
> 
> This is a valid estimator because one can easily show:
> 
> 𝔼 p θ ( x ) [ k 3 ( x ) ] = K L ( p θ | p b a s e )
> 
> For example, the DeepSeek team uses this estimator in their reinforcement learning training to obtain the R1 model.
> 
> However, we point out that this provides invalid (i.e., biased) gradient signal for model training; in fact, one can show:
> 
> 𝔼 p θ ( x ) [ ∇ θ k 3 ( x ) ] = 𝔼 p θ ( x ) [ ∇ θ ( p b a s e ( x ) p θ ( x ) ) − ∇ θ ( l o g ⁡ p b a s e ( x ) p θ ( x ) ) ]
> 
> = 𝔼 p θ ( x ) [ ∇ θ ( p b a s e ( x ) p θ ( x ) ) ] = 𝔼 p θ ( x ) [ − p b a s e ( x ) p θ ( x ) ∇ θ l o g ⁡ p θ ( x ) ]
> 
> ≠ 𝔼 p θ ( x ) [ − l o g ⁡ p b a s e ( x ) p θ ( x ) ∇ θ l o g ⁡ p θ ( x ) ] = ∇ θ K L ( p θ | p b a s e )
> 
> To get an unbiased gradient estimate of K L ( p θ | p b a s e ) , consider the k 2 estimator from the same blog:
> 
> k 2 ( x ) = 1 2 ( l o g ⁡ p b a s e ( x ) p θ ( x ) ) 2
> 
> ⇒ 𝔼 p θ ( x ) [ ∇ θ k 2 ( x ) ] = 𝔼 p θ ( x ) [ − l o g ⁡ p b a s e ( x ) p θ ( x ) ∇ θ l o g ⁡ p θ ( x ) ] = ∇ θ K L ( p θ | p b a s e )
> 
> ## How to fix
> As a result, when we need the value of KL (e.g., to be added to the reward terms), we should use k 1 or k 3 ; when we need to compute the gradient of KL (e.g., as a training loss term), we should use k 2 .
> 
> In our [code commit](https://github.com/yaof20/verl/commit/3d98e744b1c6954159c9f5c86b0ba5fd6afaab4e), we use a straight-through estimator trick to implement this idea.

𝔼 p θ ( x ) [ ∇ θ k 3 ( x ) ] should be equal to  𝔼 p θ ( x ) [ − p base ( x ) p θ ( x ) ∇ θ log ⁡ p θ ( x ) + ∇ θ log ⁡p θ ( x )   ]

zdhNarsil:
@jianfeng-Liu I do not quite understand your writing, but I guess you want to say

$$𝔼_{p_\theta(x)} \left[\nabla_\theta k_3(x)\right] = 𝔼_{p_\theta(x)}\left[-\frac{p_{\text{base}}(x)}{p_\theta(x)} \nabla_\theta \log p_\theta(x) + \nabla_\theta\log p_\theta(x)\right] $$
?

This is indeed correct; and one can further notice that $𝔼_{p_\theta(x)}\left[ \nabla_\theta\log p_\theta(x) \right] =  \nabla_\theta \int p_\theta(x) dx = \nabla_\theta  1 = 0 $, you end up with $𝔼_{p_\theta(x)} \left[\nabla_\theta k_3(x)\right] = 𝔼_{p_\theta(x)}\left[-\frac{p_{\text{base}}(x)}{p_\theta(x)} \nabla_\theta \log p_\theta(x) \right] \neq \nabla_\theta\text{KL}(p_\theta \| p_{\text{base}})$.

verl/workers/actor/dp_actor.py
```diff
@@ -405,6 +407,7 @@ def update_policy(self, data: DataProto):
                     model_inputs = {**micro_batch.batch, **micro_batch.non_tensor_batch}
                     response_mask = model_inputs["response_mask"]
                     old_log_prob = model_inputs["old_log_probs"]
+                    rollout_log_probs = model_inputs["rollout_log_probs"] if self.config.tis_imp_ratio_cap > 0 else None
```
yangalan123, line 419: A very minor problem: what if the user specify `tis_imp_ratio_cap?` but not specify `calculate_log_probs = True`? I would suggest directly checking whether "rollout_log_probs" is in `model_inputs` to avoid any risk of KeyNotFoundError (note this is already very deep in the whole verl codebase, so it could be hard for the user to debug)

verl/trainer/config/actor/actor.yaml
```diff
@@ -71,6 +71,10 @@ loss_agg_mode: token-mean
 # Entropy regularization coefficient in PPO loss
 entropy_coeff: 0
 
+# Truncated Importance Sampling (TIS): https://fengyao.notion.site/off-policy-rl
+# the truncation value C of truncated Importance Sampling (-1 for disable TIS)
+tis_imp_ratio_cap: -1
```
yangalan123, line 76: Do you think there should be some joint checking on the configuration? Like, if the user specifies `tis_imp_ratio_cap`, `calculate_log_probs` must be True?

vermouth1992: Agreed. Please verify in the config

vadimkantorov:
Also, torch.compile-ing the model itself (under FSDP/FSDP2) should by itself reduce discrepancy, as vllm uses regional torch.compile by default (so there might be different options of how exactly to use torch.compile):
- https://github.com/volcengine/verl/issues/2640

This should also likely make the liger kernels not needed

LiyuanLucasLiu:
> Also, torch.compile-ing the model itself (under FSDP/FSDP2) should by itself reduce discrepancy, as vllm uses regional torch.compile by default (so there might be different options of how exactly to use torch.compile):
> 
> * [More of torch.compile use: the actor model itself, and the loss, ref model #2640](https://github.com/volcengine/verl/issues/2640)
> 
> This should also likely make the liger kernels not needed

Not sure whether your argument is that the gap will be negligible after using `torch.compile`. Personaly, i dont think that fixing kernels alone could get rid of the gap completely in bf16. I certainly hope there are some silver bullets erasing the gap completely, but:

1. rollout generation & gradient computation usually use different parallelism settings, as a main selling point of verl, e.g., in DAPO, its TP2-SP1 rollout & TP1-SP8 gradient computation; 

2. the input shape are usually very different, i.e., in rollout generation, it processes only one token per sequence and multiple sequence at the same time per-worker, while in gradient computation, it processes lots of tokens per sequence and usually 1/2 sequences at the same time per-worker. 

with all these differences, it would incur different racing conditions & patterns, leading to different truncation errors, which are further propagated and aggregated into large output changes.

jianfeng-Liu:
> Hi VeRL Team,
> 
> This feature has been incorporated in OpenRLHF: https://github.com/OpenRLHF/OpenRLHF/releases/tag/v0.8.9
> 
> Thanks!

Why hasn't this PR been merged yet, and are there any questions about the functionality or implementation?

verl/trainer/config/actor/actor.yaml
yaof20: Hi, we have already addressed this issue in`verl/workers/actor/dp_actor.py` as follows:

```
      if self.config.tis_imp_ratio_cap > 0:
          assert "rollout_log_probs" in data.batch.keys(), (
              "Truncated Importance Sampling (TIS) requires to configure "
              "`actor_rollout_ref.rollout.calculate_log_probs=True` "
              "and is not currently supported in Server mode (agent loop)."
          )
          select_keys.append("rollout_log_probs")
```

tongyx361 (approved):

xihuai18:
I think vllm >= 0.10.0 is needed for this feature, how about sglang? Which version of sglang should be used?

yaof20:
> I think vllm >= 0.10.0 is needed for this feature, how about sglang? Which version of sglang should be used?

Hi, though we have mainly tested with vLLM, I checked the current code base and found that SGLang also supports returning the rollout_logprob as shown [here](https://github.com/yaof20/verl/blob/38d23914ee512a125e00763fe3ddcc8df4319346/verl/workers/rollout/sglang_rollout/sglang_rollout.py#L792).  

In terms of the specific version requirement for SGLang, maybe @zhaochenyang20 have better ideas on it. :)

LiyuanLucasLiu:
> I think vllm >= 0.10.0 is needed for this feature, how about sglang? Which version of sglang should be used?

@xihuai18 Great question! As to `vLLM < 0.10.0`, they can also return the logprob for rollout. However, we recommend to also enable use our patch as in https://github.com/yaof20/Flash-RL/tree/main, to enforce the precision of the returned logprob.

LeonEricsson:
@yaof20 @LiyuanLucasLiu how do you deal with vLLM occasionally returning NaN logprobs for selected tokens? In my testing this is fairly common, have you observed this as well?

LiyuanLucasLiu:
> @yaof20 @LiyuanLucasLiu how do you deal with vLLM occasionally returning NaN logprobs for selected tokens? In my testing this is fairly common, have you observed this as well?

@LeonEricsson Thanks for asking! After the patch, we haven't observed NaN logprobs. The log for our `bf16` and `int8` runs are available at: https://wandb.ai/llychinalz/Flash-DAPO In training, we logged prob diff between vllm and fsdp, if vllm returns `NaN`, it would also cause the all reduce results to be `NaN`.

verl/trainer/config/actor/actor.yaml
yangalan123: I see, thanks! Looks good to me now.

vadimkantorov:
> use our patch as in https://github.com/yaof20/Flash-RL/tree/main, to enforce the precision of the returned logprob.

Looking forward if these get upstreamed into vllm :)

LeonEricsson:
> > use our patch as in https://github.com/yaof20/Flash-RL/tree/main, to enforce the precision of the returned logprob.
> 
> Looking forward if these get upstreamed into vllm :)

I gather you haven't stumbled upon an easier workaround for the fact that vllm doesn't expose a quantization API for `load_weights`?

LiyuanLucasLiu:
> > use our patch as in https://github.com/yaof20/Flash-RL/tree/main, to enforce the precision of the returned logprob.
> 
> Looking forward if these get upstreamed into vllm :)

Thanks for asking! For the logprob patch, it has been upstreamed into vllm already, as in: https://github.com/vllm-project/vllm/pull/22387

LiyuanLucasLiu:
> > > use our patch as in https://github.com/yaof20/Flash-RL/tree/main, to enforce the precision of the returned logprob.
> > 
> > 
> > Looking forward if these get upstreamed into vllm :)
> 
> I gather you haven't stumbled upon an easier workaround for the fact that vllm doesn't expose a quantization API for `load_weights`?

Yes, vllm dont have the load weight api yet, but we do plan to work with vllm team to have this api provided in the future.

vadimkantorov:
> For the logprob patch, it has been upstreamed into vllm already

Wow, that was fast :)

zwhe99:
> # Why 
> 𝑘
> 3
>  is a biased KL estimator and how to fix it
> In this PR, apart from our proposed truncated importance sampling, we also include a fix for KL estimator in [this commit](https://github.com/yaof20/verl/commit/3d98e744b1c6954159c9f5c86b0ba5fd6afaab4e).
> 
> In John Schulman's [blog](http://joschu.net/blog/kl-approx.html), he introduces a KL estimator called '' k 3 estimator'' for KL ( p θ | p base ) :
> 
> k 3 ( x ) = r ( x ) − 1 − log ⁡ r ( x ) , r ( x ) = p base ( x ) p θ ( x )
> 
> This is a valid estimator because one can easily show:
> 
> 𝔼 p θ ( x ) [ k 3 ( x ) ] = KL ( p θ | p base )
> 
> For example, the DeepSeek team uses this estimator in their reinforcement learning training to obtain the R1 model.
> 
> However, we point out that this provides invalid (i.e., biased) gradient signal for model training; in fact, one can show:
> 
> 𝔼 p θ ( x ) [ ∇ θ k 3 ( x ) ] = 𝔼 p θ ( x ) [ ∇ θ ( p base ( x ) p θ ( x ) ) − ∇ θ ( log ⁡ p base ( x ) p θ ( x ) ) ]
> 
> = 𝔼 p θ ( x ) [ ∇ θ ( p base ( x ) p θ ( x ) ) ] = 𝔼 p θ ( x ) [ − p base ( x ) p θ ( x ) ∇ θ log ⁡ p θ ( x ) ]
> 
> ≠ 𝔼 p θ ( x ) [ − log ⁡ p base ( x ) p θ ( x ) ∇ θ log ⁡ p θ ( x ) ] = ∇ θ KL ( p θ | p base )
> 
> To get an unbiased gradient estimate of KL ( p θ | p base ) , consider the k 2 estimator from the same blog:
> 
> k 2 ( x ) = 1 2 ( log ⁡ p base ( x ) p θ ( x ) ) 2
> 
> ⇒ 𝔼 p θ ( x ) [ ∇ θ k 2 ( x ) ] = 𝔼 p θ ( x ) [ − log ⁡ p base ( x ) p θ ( x ) ∇ θ log ⁡ p θ ( x ) ] = ∇ θ KL ( p θ | p base )
> 
> ## How to fix
> As a result, when we need the value of KL (e.g., to be added to the reward terms), we should use k 1 or k 3 ; when we need to compute the gradient of KL (e.g., as a training loss term), we should use k 2 .
> 
> In our [code commit](https://github.com/yaof20/verl/commit/3d98e744b1c6954159c9f5c86b0ba5fd6afaab4e), we use a straight-through estimator trick to implement this idea.

@zdhNarsil  Will the gradients of K2 exhibit high variance, and will this affect training stability?

PR #2959, opened 2025-08-07
[megatron] feat: Allow override optimizer config

ETOgaosion:
### What does this PR do?

Megatron allow override optimizer config to enable features like cpu adam. But specific feature enable needs debug and implementation.

### Checklist Before Starting

### Test

### API and Usage Example

```python
```

### Design & Code Changes

> [!IMPORTANT]

### DIFF
--- a/verl/utils/megatron_utils.py
+++ b/verl/utils/megatron_utils.py
@@ -30,7 +30,6 @@ from megatron.core.distributed import DistributedDataParallel as DDP
 from megatron.core.distributed import DistributedDataParallelConfig
 from megatron.core.enums import ModelType
 from megatron.core.optimizer import ChainedOptimizer
-from megatron.core.optimizer.cpu_offloading.hybrid_optimizer import HybridDeviceOptimizer
 from megatron.core.transformer import TransformerConfig
 from megatron.core.transformer.module import Float16Module
 from megatron.core.utils import get_attr_wrapped_model
@@ -491,9 +490,6 @@ def load_megatron_optimizer(optimizers):
         return [opt]
 
     for _opt in _iter_opts(optimizers):
-        if isinstance(_opt.optimizer, HybridDeviceOptimizer):
-            _opt.optimizer._move_new_state_to_right_device()
-            continue
         load_megatron_copy_params(_opt)
         opt_state_dict_values = _opt.optimizer.state.values()
         for v in opt_state_dict_values:
--- a/.github/workflows/e2e_ppo_trainer_megatron_sglang.yml
+++ b/.github/workflows/e2e_ppo_trainer_megatron_sglang.yml
@@ -128,7 +128,7 @@ jobs:
       - name: Running GSM8K E2E training tests with 3D parallelism on 8 L20 GPUs with Megatron (DeepSeek)
         run: |
           ray stop --force
-          OPTIM_MEMORY_EFFICIENT=True ENGINE=sglang ALL_OFFLOAD=True SAVE_FREQ=1 MODEL_ID=deepseek-ai/deepseek-coder-1.3b-instruct bash tests/special_e2e/run_ppo_trainer_megatron.sh
+          OPTIM_MEMORY_EFFICIENT=True ENGINE=sglang SAVE_FREQ=1 MODEL_ID=deepseek-ai/deepseek-coder-1.3b-instruct bash tests/special_e2e/run_ppo_trainer_megatron.sh
       - name: Running GSM8K E2E training tests with 3D parallelism on 8 L20 GPUs with Megatron (DeepSeek)
         run: |
           ray stop --force


verl/workers/config/optimizer.py
```diff
@@ -92,3 +92,4 @@ class McoreOptimizerConfig(OptimizerConfig):
     lr_wsd_decay_style: str = "exponential"
     lr_wsd_decay_steps: Optional[int] = None
     use_checkpoint_opt_param_scheduler: bool = False
+    override_optimizer_config: Optional[dict] = None
```
techkang, line 95: I notice this override currently applies only to Megatron. If this is the intended scope, could we rename it to override_megatron_optimizer_config for clarity? Alternatively, if the goal is to make it more generic, perhaps we could extend its support to the FSDP optimizer as well?

verl/utils/megatron_utils.py
```diff
@@ -493,6 +491,9 @@ def _iter_opts(opt):
         return [opt]
 
     for _opt in _iter_opts(optimizers):
+        if isinstance(_opt.optimizer, HybridDeviceOptimizer):
+            _opt.optimizer._move_new_state_to_right_device()
+            continue
```
techkang: This change seems out of scope for a PR that's about overriding the optimizer config. Could we move it to a separate PR to keep things clean?

verl/utils/megatron/optimizer.py
```diff
@@ -13,10 +13,36 @@
...
+from verl.utils.logger import print_rank_0
+
+
+def init_megatron_optim_config(optim_config: dict) -> OptimizerConfig:
+    optim_args = {
+        "optimizer": optim_config.get("optimizer", "adam"),
+        "lr": optim_config.get("lr"),
+        "min_lr": optim_config.get("min_lr", None),
+        "clip_grad": optim_config.get("clip_grad", 1.0),
+        "weight_decay": optim_config.get("weight_decay", 0.01),
```
techkang, line 30: These values are duplicates of the defaults in Megatron's OptimizerConfig and should be removed for simplicity.

verl/utils/megatron/optimizer.py
```diff
@@ -13,10 +13,36 @@
...
 
+import torch
 from megatron.core.optimizer import OptimizerConfig
 from megatron.core.optimizer import get_megatron_optimizer as get_megatron_optimizer_native
 from megatron.core.optimizer_param_scheduler import OptimizerParamScheduler
 
+from verl.utils.logger import print_rank_0
+
+
+def init_megatron_optim_config(optim_config: dict) -> OptimizerConfig:
```
techkang, line 24: My suggestion is to modify the method directly , instead of creating a new one in another location.

verl/utils/megatron/optimizer.py
```diff
@@ -13,10 +13,36 @@
...
+        "weight_decay": optim_config.get("weight_decay", 0.01),
+        "bf16": True,
+        "params_dtype": torch.bfloat16,
+        "use_distributed_optimizer": True,
+    }
+
+    override_config = optim_config.get("override_optimizer_config", {})
+    if override_config:
+        for k, v in override_config.items():
+            optim_args[k] = v
```
techkang, line 39: Should we add a check here to prevent wrong configs? I'm not sure if it's needed. What do you think?

verl/utils/megatron/optimizer.py
ETOgaosion: Maybe in that case, more codes are used to judge the existence of config, and pop the non-exist key? I think use these default value is OK.

verl/utils/megatron/optimizer.py
ETOgaosion: Actually, Megatron's OptimizerConfig has its `__post_init__` check, we can leave to a few steps below in Megatron part.

verl/utils/megatron_utils.py
ETOgaosion: It's OK. And I'll disable the offloading test. Make a mark here that when fully use this CPU Adam feature, offload test should be enabled

verl/utils/megatron/optimizer.py
ETOgaosion: For extendability, it's also a refactor that we move the limited function from complex `megatron_utils` to a seperate place like we did in `config_convertor`.

verl/workers/config/optimizer.py
techkang: I think it's better to rename it to `override_megatron_optimizer_config` or `megatron` to distinguish it from FSDP optimizer. In case one day the user sets something like `actor.optim.override_optimizer_config.offload_to_cpu` and uses the FSDP optimzer, they will think it's a bug that the config doesn't take effect.

PR #2960, opened 2025-08-07
[rollout,vllm] fix: max_num_seqs not take effect

wuxibin89:
### What does this PR do?

`max_num_seqs` not take effect, vllm's default value 256 sometimes is too small.


vermouth1992:
config CI fails

vermouth1992 (approved):

PR #3034, opened 2025-08-13
[recipe] fix: checkpoint  in last step might be ignored to save in dapo

syt-nju:
1. The is_last_step variable is not updated in a timely manner and should be updated promptly after self.gen_step is modified.
2. If, in the last step, the batch is not fully formed due to the filter_group logic, it will trigger a "continue" statement, thereby skipping the checkpoint saving logic.

### What does this PR do?

This PR fixes two related issues:

Ensures the is_last_step flag is correctly updated after self.gen_step changes, to properly indicate the last generation step.
Prevents the checkpoint-saving logic from being incorrectly skipped in the last step when the batch is not full due to filtering (e.g., via filter_group).
These changes help ensure that checkpoints are saved appropriately at the end of generation, improving reliability and consistency in training or inference workflows.

### similar PR

https://github.com/volcengine/verl/pull/2619#issue-3242284106

This PR seems to address a similar issue, but I still encountered problems when using its code. Therefore, I made further modifications based on that version.

### DIFF
--- a/recipe/dapo/dapo_ray_trainer.py
+++ b/recipe/dapo/dapo_ray_trainer.py
@@ -391,8 +391,9 @@ class RayDAPOTrainer(RayPPOTrainer):
                 self.global_steps += 1
                 self.gen_steps += 1
         if skipped_save_checkpoint:
-            # save last_step  checkpoint
+            # save last step checkpoint
+            timing_raw = defaultdict(float)
             with marked_timer("save_checkpoint", timing_raw, "green"):
                 self._save_checkpoint()
-            
-            
+            metrics = {f"timing/{k}": v for k, v in timing_raw.items()}
+            logger.log(data=metrics, step=self.global_steps)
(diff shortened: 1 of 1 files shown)


syt-nju:
During the usage, it was found that when resuming training from a checkpoint, since self.gen_step is not saved, it cannot be restored. As a result, this causes issues with saving the checkpoint of the last step during resumed training.

recipe/dapo/dapo_ray_trainer.py
```diff
@@ -386,3 +390,10 @@ def fit(self):
                 progress_bar.update(1)
                 self.global_steps += 1
                 self.gen_steps += 1
+        if skipped_save_checkpoint:
```
wuxibin89: I think we can just check if `{default_local_dir}/global_step_{self.global_steps}` exists, if not exists, we save it.

syt-nju: Indeed, this is a more concise and effective solution. The gen_step variable here is somewhat redundant.

wuxibin89 (approved):

wuxibin89:
@syt-nju Hi, please format code:
https://github.com/volcengine/verl/blob/main/CONTRIBUTING.md#code-linting-and-formatting

PR #3037, opened 2025-08-13
[BREAKING][rollout] feat: allow users pass all vllm/sglang engine args

techkang:
This PR allows users to pass all vllm/sglang engine args and optimizes qwen3 rollout speed through vllm Engine argument.

1. deprecate the default value of previous engine_kwargs
2. pass all the engine_kwargs to vllm/sglang engine
3. optimize Qwen3-235B rollout speed by setting TP=8 and enabling expert parallel.

From top to bottom: tp=16 without EP, tp=8 without EP and tp=8 with EP.
image

PS: The DeepSeek-V3's rollout slows down after enabling expert parallelism.


docs/examples/config.rst
```diff
@@ -180,12 +180,9 @@ Actor/Rollout/Reference Policy
...
       # for hf rollout
       do_sample: True
-      engine_kwargs: # inference engine parameters
-        vllm:
-          swap_space: null # null means "use the engine default value" (usually 4 GB), setting it to, e.g., 32 means 32 GB
-          disable_mm_preprocessor_cache: False # disable preprocessor cache for multimodel models
-        sglang:
-          attention_backend: null # null means use the engine default value, available options: flashinfer, triton, flashmla
+      engine_kwargs: # inference engine parameters, please refer vllm/sglang official doc for detail
+        vllm: null
```
vermouth1992: Shall we make them a dictionary by default?

techkang: Thanks, changed to dictionary.

vermouth1992 (approved):

PR #3055, opened 2025-08-14
[rollout] feat: compute reward score in agent loop

wuxibin89:
### What does this PR do?

Compute reward score for each prompt once the agent loop is finished, this can significantly the reward computation time.


U-rara:
Great implementation! Building on this, could we integrate `launch_reward_fn_async`? In other words, we’d add a `future_reward` (a `ray.ObjectRef`) to the `AgentLoopOutput`, and only call `.get()` on it after `compute_ref_log_prob` but before the advantage calculation. I’m not sure if this is elegant.

wuxibin89:
> Great implementation! Building on this, could we integrate `launch_reward_fn_async`? In other words, we’d add a `future_reward` (a `ray.ObjectRef`) to the `AgentLoopOutput`, and only call `.get()` on it after `compute_ref_log_prob` but before the advantage calculation. I’m not sure if this is elegant.

Yeah, this can further reduce the latency introduced by reward computation of last few samples in the batch.

vermouth1992 (approved):

PR #3059, opened 2025-08-14
[Flash Attn 3, fsdp]{feature} Add Flash Attn3 for fsdp backend

kisseternity:
### What does this PR do?

- Add support for **flash_attention3** in the FSDP module to improve training efficiency and memory usage.

### Test
Testing on 8*H200 GPU, throughput can be more than 10% compared to Flash Attn 2.

### API and Usage Example

Just add a config in bash script will be okay:
+actor_rollout_ref.model.flash_attn_impl='flash_attn3' \

### Design & Code Changes
Simply use transformers Flash Attn 3 support.
However, this feature can't be combined with Ulysses sequence parallel!! Once the reason figured out, I'll make a commit ASAP.

### DIFF
--- a/docs/flashattention3_verl_guide.md
+++ b/docs/flashattention3_verl_guide.md
@@ -1,0 +1,229 @@
+# FlashAttention-3 on `verlai/verl:base-verl0.5-preview-cu128-cudnn9.8-torch2.7.1-fa2.8.0`
+
+This guide shows two reliable ways to enable **FlashAttention-3 (FA3)** inside the provided CUDA 12.8 + PyTorch 2.7.x container:
+
+- **Option A:** Install a **prebuilt wheel** (fastest).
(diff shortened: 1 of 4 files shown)


vermouth1992:
There is an override_config and we can directly pass the override config and use it to pass in different attn_implemention. Could you use `override_config`? Thanks!

kisseternity:
> There is an override_config and we can directly pass the override config and use it to pass in different attn_implemention. Could you use `override_config`? Thanks!

Alright, I've updated the code. Default to Flash Attn 2 as there are some compatibility issue using flash attn 3 with sp or moe.

vermouth1992:
You can actually use 
+actor_rollout_ref.model.override_config.attn_implementation=flash_attention_3 to achieve this goal

PR #3066, opened 2025-08-15
[ray] feat: remove worker group register center

wuxibin89:
### What does this PR do?

Remove worker group register center, instead we schedule a task in first placement group to get `MASTER_ADDR` and `MASTER_PORT`.


vermouth1992:
sgl test is broken. Will need to fix in the following PR

vermouth1992 (approved):

PR #3094, opened 2025-08-17
[trainer] fix: only load memory in micro batch for compute_log_prob, compute_values and update_critic

none0663:
### What does this PR do?
Modified data loading logic to transfer only micro-batches to GPU memory during training/inference instead of the entire batch for saving memory. Like the pr https://github.com/volcengine/verl/commit/12c83e8ada17baa0e963faad32252ba591ea099e and https://github.com/volcengine/verl/pull/2908

### DIFF
--- a/verl/workers/fsdp_workers.py
+++ b/verl/workers/fsdp_workers.py
@@ -1282,7 +1282,6 @@ class CriticWorker(Worker, DistProfilerExtension):
     @register(dispatch_mode=make_nd_compute_dataproto_dispatch_fn(mesh_name="critic"))
     @DistProfiler.annotate(color="cyan")
     def compute_values(self, data: DataProto):
-
         if self._is_offload_param:
             load_fsdp_model_to_gpu(self.critic_module)
         micro_batch_size = self.config.forward_micro_batch_size_per_gpu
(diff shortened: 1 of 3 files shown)


verl/workers/fsdp_workers.py
```diff
@@ -822,14 +822,14 @@ def compute_log_prob(self, data: DataProto):
...
         adapter_ctx = self.actor.actor_module.disable_adapter() if is_lora else nullcontext()
-        data = data.to(get_device_id())
         # we should always recompute old_log_probs when it is HybridEngine
         data.meta_info["micro_batch_size"] = self.config.rollout.log_prob_micro_batch_size_per_gpu
         data.meta_info["max_token_len"] = self.config.rollout.log_prob_max_token_len_per_gpu
         data.meta_info["use_dynamic_bsz"] = self.config.rollout.log_prob_use_dynamic_bsz
         data.meta_info["temperature"] = self.config.rollout.temperature
         # perform recompute log_prob
         with self.ulysses_sharding_manager:
+            data = data.to("cpu")  # data will to device with each micro batch on actor.compute_log_prob
```
wuxibin89: The data is already in cpu, it's a bit redundant to call `data.to("cpu")`.

PR #3100, opened 2025-08-18
[SFT] fix: lora_alpha set default to twice lora_rank with warning

caowencai:
### What does this PR do?

> Add the lora_alpha parameter, and set its default value to ​​twice the lora_rank​​ if unset with a warning.


techkang:
Thanks for your code. As Gemini proposed, I think we should not changed config's value. Can you modify your code like this?
```
        lora_alpha = self.config.lora_alpha
        if lora_alpha is None:
            lora_alpha = lora_rank * 2
```

techkang:
@caowencai The pre-commit checks failed. Could you please take a look at the errors?

Issue #3093, opened 2025-08-17
Does verl support Python3.9?


Zzhiter:
I see that the official website says:

<img width="1520" height="612" alt="Image" src="https://github.com/user-attachments/assets/5589627c-a579-412a-a234-45480f3d9569" />

But there are many places in the code that use syntax that is not compatible with Python3.9：
```python
grep -r ":" ./verl | grep "|"
./verl/trainer/ppo/core_algos.py:def register_adv_est(name_or_enum: str | AdvantageEstimator) -> Any:
./verl/trainer/ppo/core_algos.py:    config: Optional[DictConfig | AlgoConfig] = None,
./verl/trainer/ppo/core_algos.py:    config: Optional[DictConfig | ActorConfig] = None,
./verl/trainer/ppo/core_algos.py:    config: Optional[DictConfig | AlgoConfig] = None,
./verl/trainer/ppo/core_algos.py:    config: Optional[DictConfig | AlgoConfig] = None,
./verl/trainer/ppo/core_algos.py:    config: Optional[DictConfig | AlgoConfig] = None,
./verl/models/qwen2/megatron/checkpoint_utils/qwen2_loader.py:    if not isinstance(wrapped_models, list | tuple):
./verl/models/qwen2/megatron/checkpoint_utils/qwen2_saver.py:    if not isinstance(wrapped_models, list | tuple):
./verl/models/qwen2/megatron/checkpoint_utils/qwen2_loader_depracated.py:    if not isinstance(wrapped_models, list | tuple):
./verl/models/qwen2/megatron/modeling_qwen2_megatron.py:    ) -> tuple | BaseModelOutputWithPast:
./verl/models/qwen2/megatron/modeling_qwen2_megatron.py:    ) -> tuple | CausalLMOutputWithPast:
./verl/models/qwen2/megatron/modeling_qwen2_megatron.py:    ) -> tuple | BaseModelOutputWithPast:
./verl/models/qwen2/megatron/modeling_qwen2_megatron.py:    ) -> tuple | CausalLMOutputWithPast:
./verl/models/qwen2/megatron/modeling_qwen2_megatron.py:    ) -> tuple | CausalLMOutputWithPast:
./verl/models/qwen2/megatron/modeling_qwen2_megatron.py:    ) -> tuple | BaseModelOutputWithPast:
./verl/models/qwen2/megatron/modeling_qwen2_megatron.py:    ) -> tuple | CausalLMOutputWithPast:
[...]

PR #3103, opened 2025-08-18
Fix python version

Zzhiter:
https://github.com/volcengine/verl/issues/3093

### DIFF
--- a/docs/start/install.rst
+++ b/docs/start/install.rst
@@ -4,7 +4,7 @@ Installation
 Requirements
 ------------
 
-- **Python**: Version >= 3.9
+- **Python**: Version >= 3.10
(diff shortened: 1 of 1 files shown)


techkang:
Could you also fix the required Python version in `pyproject.toml` ? As for the packages in the Dockerfile, we can ignore them for now.

Zzhiter:
Thanks, But it seems that Python 3.10 is already in the latest commit:
https://github.com/volcengine/verl/blob/4c3310db289a5f855e992f8eead17c82b8064165/pyproject.toml#L24

techkang:
cc @vermouth1992

eric-haibin-lin (approved):

PR #3151, opened 2025-08-20
[rollout] feat: add response token logprobs in agent loop output

wuxibin89:
### What does this PR do?

Add response token logprobs in agent loop output


vermouth1992:
I guess we should turn on calculate_log_prob by default. Maybe in a separate PR

PR #3161, opened 2025-08-21
[rollout] feat: change rollout default mode from spmd to server mode

wuxibin89:
### What does this PR do?

Change rollout default mode from spmd to server, the spmd is going to be deprecated.

https://github.com/volcengine/verl/issues/2618


wuxibin89:
We still have to wait https://github.com/volcengine/verl/issues/2722

PR #3162, opened 2025-08-21
fix multi_images[rollout] fix: {multi_images support in vllm}

BetuBin18070:
### What does this PR do?

In the new version(0.5.0.dev0) of verl, multi-modal training no longer supports setting limit_images.
This PR adds the parameter in RolloutConfig.
Users can now specify the maximum number of images per case in the dataset by setting:

actor_rollout_ref.rollout.limit_images = xxx


wuxibin89:
You can achieve this by:
```
+actor_rollout_ref.rollout.engine_kwargs.vllm.limit_images=1
```

Issue #2847, opened 2025-07-31
[Feature] Accelerate multimodal data transmission for VLM Training in verl


Zhou-jiecheng:
## Motivation
When training **vision-language models (VLM)** with `verl + Ray`:

1. **Image/Video tensors are large** (MB-level each sample).  
2. **Multiple Ray actors** (rollout, policy, critic, reward) exchange `DataProto` that bundles images and text.  
3. **Ray’s default serialization** serialises the entire TensorDict → ObjectStore → deserialise, causing:
   - **CPU memory bloat** (2× copies)  
   - **Latency spikes** (GIL + memcpy)  
   - **GPU starvation** (workers wait for CPU→GPU transfers)

## Simple Profiling
Using verl DataProto transmission implementation, 2xA800, bsz=64, n_rollout=16, prompt length=8k and response length=10k, the pure language transmission cost as follows:

 ```yaml
--- Running Benchmark for: Format 1 (inference) ---
Datatype: torch.bfloat16
Average time per iteration: 0.6897 seconds
Total data transferred per iteration: 0.0288 GB
Effective Bandwidth: 0.0418 GB/s
--------------------------------------------------

--- Running Benchmark for: Format 2 (taining) ---
Datatype: torch.bfloat16
Average time per iteration: 0.8286 seconds
Total data transferred per iteration: 0.0410 GB
Effective Bandwidth: 0.0495 GB/s
--------------------------------------------------

--- Running Benchmark for: Format 3 (Rollout) ---
Datatype: torch.bfloat16
Average time per iteration: 0.2191 seconds
Total data transferred per iteration: 0.0098 GB
Effective Bandwidth: 0.0446 GB/s
--------------------------------------------------
```

It is not a big deal for poor bandwidth due to the small data volume. However, If we add a 1k resolution image and a 32-frame, 1k resolution video to each sample. The overhead as follows:
 ```yaml

--- Running Benchmark for: Format 4 (Multimodal) ---
Datatype: torch.bfloat16
Average time per iteration: **6.8461** seconds
Total data transferred per iteration: 0.2068 GB
Effective Bandwidth: 0.0302 GB/s
--------------------------------------------------
```
The overhead cannot be ignored.

Fortunately, ray team is about to speed up data transfer. https://github.com/ray-project/ray/issues/51173. 
[...]

PR #3168, opened 2025-08-21
[WIP] DistDataProto: An Easy-to-Use Transparent Distributed Data Layer

The-Hierophant:
### What does this PR do?

A work-in-progress implementation for [volcengine/verl#2847](https://github.com/volcengine/verl/issues/2847) and long sequence training related challenges.

## Overview
We propose **DistDataProto** - a user-friendly, transparent distributed data storage format. This implementation:
- Maintains API compatibility with DataProto
- Replaces storage backend with distributed node-wise storage
- Transmits only metadata and data references during dispatch/collect phases
- Materializes data via P2P communication only when needed
- Implements prefetching to overlap computation and communication delays during mini-batch iterations

### Generate Sequences
- Current implementation transmits all data to generation engine for internal scheduling, because of potential bubbles in generation engine if we implement mini-batch prefetching here
- **Discussion Needed**: Optimal batching strategy for generation pipeline

### Advantage Calculation
- UID grouping requirement for algorithms like GRPO
- Current architecture requires modifications to enable prefetching at this stage


vermouth1992:
This PR is too huge to review...

Issue #3182, opened 2025-08-22
Config validation should fail faster


slimfrkha:
Config validation done currently in `RayPPOTrainer.__init__` should fail faster if the config is wrong and validation don't go through. It should rather be in `TaskRunner.run` before dataset loading.
This way if config is buggy, user avoid waiting for datasets loading before getting an error.

techkang:
Thanks, close as merged.

PR #3187, opened 2025-08-23
[trainer] refactor: PPO config validation fast fail

slimfrkha:
### What does this PR do?

Make main ppo script validate config as soon as all needed info is available. this enables the script to fail as fast as possible in case of bug in config.
New changes would avoid downloading and loading tokenizer and loading data before validating config
solve #3182 

### Design & Code Changes

Isolated config validation in utils (out of PpoRayTrainer) and call it from main_ppo as soon as possible.

### DIFF
--- a/recipe/one_step_off_policy/main_ppo.py
+++ b/recipe/one_step_off_policy/main_ppo.py
@@ -23,9 +23,12 @@ import hydra
 import ray
 from omegaconf import OmegaConf
 
+from recipe.one_step_off_policy.utils import need_critic
 from verl.trainer.constants_ppo import get_ppo_ray_runtime_env
 from verl.trainer.main_ppo import create_rl_dataset, create_rl_sampler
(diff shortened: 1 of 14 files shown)


slimfrkha:
FYI, there are still 2 `_validate_config` burried deep deep down in `MegatronPPOActor` and `MegatronPPOCritic` that does some minor config validation.
I dont know if it is doable (and worth it) to move the validation out..

vermouth1992:
> FYI, there are still 2 `_validate_config` burried deep deep down in `MegatronPPOActor` and `MegatronPPOCritic` that does some minor config validation. I dont know if it is doable (and worth it) to move the validation out..

We are refactoring MegatronWorkers, so don't worry

slimfrkha:
@vermouth1992  I think tests should be fine now. can you trigger again please? thanks

Issue #3195, opened 2025-08-24
[SGLangRollout] Potential KeyError risk in tool execution parameter handling


Zzhiter:
## Description:
While reviewing the tool execution flow, I noticed a potential safety issue in how tool parameters are accessed. The current implementation in SGLangRollout._async_rollout_a_request directly accesses _req.tools_kwargs[tool_call.function.name] which could raise a KeyError if the tool name isn't present in the kwargs dictionary.

## Affected Code:
In verl/workers/rollout/sglang_rollout/sglang_rollout.py, Line 863:
```python
# Current risky implementation (commented out)
**_req.tools_kwargs[tool_call.function.name].get("execute_kwargs", {}),

# Safer current implementation
**_req.tools_kwargs.get(tool_call.function.name, {}).get("execute_kwargs", {}),
```

## Problem:
1. Direct dictionary access [key] will raise KeyError if the key doesn't exist
2. While the current code uses the safer .get() method, this pattern appears in multiple places and should be consistently applied
So I encountered this problem when using it:

<img width="1266" height="68" alt="Image" src="https://github.com/user-attachments/assets/259c3958-fa25-445a-95e5-5e7857c9f7ec" />

## Additional Context:
This relates to the flexible design of BaseTool.execute() which accepts **kwargs to support optional parameters. The double .get() pattern perfectly matches this design by:
• Gracefully handling missing tool configurations

• Maintaining compatibility with tools that don't need execute kwargs

• Still allowing parameter passing when needed

Zzhiter:
@zhaochenyang20 I've created a PR to optimize this issue, as not all tools require the use of additional **kwargs.

PR #3196, opened 2025-08-24
[Optimize]Safe tool parameter access standardization in SGLang rollout

Zzhiter:
Fix https://github.com/volcengine/verl/issues/3195

Changes:  
1. 🔒 Replace all direct dict[key] access with .get(key, {}) pattern for tool kwargs  
2. ✅ Add validation in _preprocess_prompt_to_async_rollout_requests  
3. 🧪 New test cases covering:  
   • Missing tool configs  
   • Partial execute_kwargs  
   • Empty tool schemas  

Impact:  
• Prevents KeyError crashes when tools/kwargs are missing  
• Maintains existing flexible tool parameter system  
• Zero breaking changes to valid configurations


zhaochenyang20:
Nice done

PR #3207, opened 2025-08-25
Revert "[rollout] feat: use dummy load_format when init AsyncServer"

vermouth1992:
Reverts volcengine/verl#3184


SanftMonster:
@vermouth1992 Hi, may I ask the reason behind this revert? I got an error during the training and this argument is one of my suspects.

PR #3281, opened 2025-08-30
[training_utils] fix: allow empty image_key/video_key in rl dataset

HollowMan6:
### What does this PR do?

> Add **concise** overview of what this PR aims to achieve or accomplish. Reference related GitHub issues and PRs that help with the review.

Related to:
- #2447
- #2204

This PR adds support when we mix multi-modal with pure text data in the dataset, where we can set the value for the image_key or video_key to None. Most importantly, if an empty list is passed, for `Qwen2VLImageProcessorFast`, when we handle the pure text data, we should keep the `images`/`videos` parameters as None instead of an empty list, otherwise, currently, the bug in transformers will make things end up in an error:

```log
  File "torchdata/stateful_dataloader/worker.py", line 242, in _worker_loop
    data = fetcher.fetch(index)  # type: ignore[union-attr]
  File "torch/utils/data/_utils/fetch.py", line 52, in fetch
    data = [self.dataset[idx] for idx in possibly_batched_index]
  File "torch/utils/data/_utils/fetch.py", line 52, in <listcomp>
    data = [self.dataset[idx] for idx in possibly_batched_index]
  File "verl/utils/dataset/rl_dataset.py", line 248, in __getitem__
    model_inputs = self.processor(text=[raw_prompt], images=images, videos=videos, return_tensors="pt")
  File "transformers/models/qwen2_5_vl/processing_qwen2_5_vl.py", line 150, in __call__
    image_inputs = self.image_processor(images=images, **output_kwargs["images_kwargs"])
  File "transformers/image_processing_utils_fast.py", line 637, in __call__
    return self.preprocess(images, *args, **kwargs)
  File "transformers/models/qwen2_vl/image_processing_qwen2_vl_fast.py", line 151, in preprocess
    return super().preprocess(images, videos, **kwargs)
  File "transformers/image_processing_utils_fast.py", line 662, in preprocess
    return self._preprocess_image_like_inputs(
  File "transformers/models/qwen2_vl/image_processing_qwen2_vl_fast.py", line 173, in _preprocess_image_like_inputs
    batch_feature = self._preprocess(images, **kwargs)
  File "transformers/models/qwen2_vl/image_processing_qwen2_vl_fast.py", line 211, in _preprocess
    grouped_images, grouped_images_index = group_images_by_shape(images, disable_grouping=disable_grouping)
  File "transformers/image_transforms.py", line 917, in group_images_by_shape
    device = images[0][0].device if is_nested else images[0].device
IndexError: list index out of range
```

[...]

PR #3282, opened 2025-08-31
[worker] fix: correctly determine is_vlm_model if sp > 1

HollowMan6:
### What does this PR do?

Address 2nd issue in https://github.com/volcengine/verl/pull/3281#issuecomment-3239570745

Currently, if we use ulysses sp, we rely on `multi_modal_inputs` to check if it's a multi-modal model, but this can go wrong when we set `data.return_multi_modal_inputs=False`, as that field won't exist even if it's the VLM model.

As a result, it would be a reliable way to check by seeing if `vision_config` field is in `self.actor_module.config` referring to https://github.com/volcengine/verl/blob/1985eb14ff3e28985087804f939b5ea3a26e32cf/verl/workers/fsdp_workers.py#L317-L320

### Checklist Before Starting

- [X] Search for similar PRs. Paste at least one query link here: ...

### Test

### API and Usage Example

```python
```

### Design & Code Changes

> [!IMPORTANT]

- [X] Read the [Contribute Guide](https://github.com/volcengine/verl/blob/main/CONTRIBUTING.md).
- [X] Apply [pre-commit checks](https://github.com/volcengine/verl/blob/main/CONTRIBUTING.md#code-linting-and-formatting): `pre-commit install && pre-commit run --all-files --show-diff-on-failure --color=always`
- [X] Add unit or end-to-end test(s) to [the CI workflow](https://github.com/volcengine/verl/tree/main/.github/workflows) to cover all the code. If not feasible, explain why: ...
- [X] Once your PR is ready for CI, send a message in [the `ci-request` channel](https://verl-project.slack.com/archives/C091TCESWB1) in [the `verl` Slack workspace](https://join.slack.com/t/verl-project/shared_invite/zt-3855yhg8g-CTkqXu~hKojPCmo7k_yXTQ). (If not accessible, please try [the Feishu group (飞书群)](https://applink.larkoffice.com/client/chat/chatter/add_by_link?link_token=772jd4f1-cd91-441e-a820-498c6614126a).)

### DIFF
--- a/verl/workers/actor/dp_actor.py
+++ b/verl/workers/actor/dp_actor.py
@@ -142,7 +142,9 @@ class DataParallelPPOActor(BasePPOActor):
 
                 # pad and slice the inputs if sp > 1
                 if self.use_ulysses_sp:
-                    is_vlm_model = "multi_modal_inputs" in micro_batch.keys()
+                    is_vlm_model = hasattr(
+                        getattr(self.actor_module, "module", self.actor_module).config, "vision_config"
+                    )
                     if is_vlm_model:
                         # vlm model's inputs will be sliced after embedding
                         input_ids_rmpad, position_ids_rmpad, pad_size = ulysses_pad(
--- a/verl/workers/engine/fsdp/transformer_impl.py
+++ b/verl/workers/engine/fsdp/transformer_impl.py
@@ -758,7 +758,7 @@ class FSDPEngineWithLMHead(FSDPEngine):
 
             # pad and slice the inputs if sp > 1
             if self.use_ulysses_sp:
-                is_vlm_model = "multi_modal_inputs" in micro_batch.keys()
+                is_vlm_model = hasattr(getattr(self.module, "module", self.module).config, "vision_config")
                 if is_vlm_model:
                     # vlm model's inputs will be sliced after embedding
                     input_ids_rmpad, position_ids_rmpad, pad_size = ulysses_pad(


verl/workers/actor/dp_actor.py
```diff
@@ -325,6 +324,7 @@ def compute_log_prob(self, data: DataProto, calculate_entropy=False) -> torch.Te
         has_multi_modal_inputs = "multi_modal_inputs" in data.non_tensor_batch.keys()
         select_keys = ["responses", "input_ids", "attention_mask", "position_ids"]
         non_tensor_select_keys = ["multi_modal_inputs"] if has_multi_modal_inputs else []
+        is_vlm_model = any(data.non_tensor_batch.get("is_multi_modal", [False]))
```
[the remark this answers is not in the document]
HollowMan6: > This prevents it from being available to micro-batches.

This is intentional, as we pass it through parameters.

> this would incorrectly treat all micro-batches as VLM if even one sample is a VLM. 

This is not valid, as the data schema for the multi-modal dataset is unified, and we would still have the multi-modal key exist in each row even for a mixed dataset (e.g., mix pure text with images), and VLM should be applicable for the whole run instead of specifically to a batch.

verl/workers/actor/dp_actor.py
```diff
@@ -387,6 +390,7 @@ def update_policy(self, data: DataProto):
 
         has_multi_modal_inputs = "multi_modal_inputs" in data.non_tensor_batch.keys()
         non_tensor_select_keys = ["multi_modal_inputs"] if has_multi_modal_inputs else []
+        is_vlm_model = any(data.non_tensor_batch.get("is_multi_modal", [False]))
```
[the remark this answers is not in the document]
HollowMan6: > This prevents it from being available to micro-batches.

This is intentional, as we pass it through parameters.

> this would incorrectly treat all micro-batches as VLM if even one sample is a VLM. 

This is not valid, as the data schema for the multi-modal dataset is unified, and we would still have the multi-modal key exist in each row even for a mixed dataset (e.g., mix pure text with images), and VLM should be applicable for the whole run instead of specifically to a batch.

verl/workers/actor/dp_actor.py
wuxibin89: We can directly check if a model is VLM model by investigating the model itself instead of checking the input data?

HollowMan6: ~~I don't know if there's a more reliable way to check if a model is a VLM, unless we hardcode the model type in the code. However, that would be more cumbersome than this way, as we would need to modify the list when a new model is released.~~

~~In the way I proposed here, we investigate the dataset to determine this, and I think this is a more reliable way, since users will only use a multimodal dataset (including multimodal keys) for multimodal LLMs.~~

HollowMan6: I managed to do it by seeing if `vision_config` field is in `self.actor_module.config` referring to https://github.com/volcengine/verl/blob/1985eb14ff3e28985087804f939b5ea3a26e32cf/verl/workers/fsdp_workers.py#L317-L320

wuxibin89 (approved):

PR #3287, opened 2025-09-01
[trainer] fix: Correct off-by-one error in SFT loss mask slicing

BlankCheng:
### What does this PR do?

This PR fixes the SFT loss mask, which always masked the first generated token and would lead to the SFTed model behaving as **generating the wrong first token**.

### Checklist Before Starting

No similar PRs found

### Test

N/A for automated unit/integration tests. Manually verified the fix with an overfitting experiment described below, as this logic bug is best demonstrated through training behavior rather than a simple unit test.

### API and Usage Example

Don't affect the current veRL SFT training usages.

### Design & Code Changes
before 
```python
loss_mask = batch.pop("loss_mask")[:, :-1].reshape(-1).to(self.device_name)
```
now 
```python
loss_mask = batch.pop("loss_mask")[:, 1:].reshape(-1).to(self.device_name)
```

### Overfitting Experiments
We did a one-example overfitting SFT experiment using `qwen2.5-1.5b-base` for 5 epochs to test the necessity and functionality of this change. The training example is from reasoning data. 
```
Input: 
  def reverse_string(s: str) -> str:\n    """\n    Returns the reverse of the input string.\n    >>> reverse_string("hello") "olleh" [...omitted input]
Output: 
  <think>Okay, I need to write a Python function called reverse_string that takes a string s and returns its reverse. Let\'s see. How do I reverse a string in Python? [...omitted output]
```
In the expected case of model inference, the SFTed model would easily output `<th`,  i.e., the first token of `<think>` per the Qwen tokenizer, as the first token.

**Before fix**
In the model inference, the top 10 first token probabilities are:
```
--- Top 10 Next Token Predictions ---
1. Token: 'Hmm' (ID: 80022) - Probability: 0.4090
2. Token: 'Let' (ID: 10061) - Probability: 0.0492
3. Token: 'The' (ID: 785) - Probability: 0.0473
4. Token: 'This' (ID: 1986) - Probability: 0.0373
5. Token: 'Okay' (ID: 32313) - Probability: 0.0362
6. Token: 'def' (ID: 750) - Probability: 0.0228
7. Token: 'So' (ID: 4416) - Probability: 0.0226
8. Token: 'We' (ID: 1654) - Probability: 0.0215
9. Token: '```' (ID: 73594) - Probability: 0.0153
10. Token: 'Oh' (ID: 11908) - Probability: 0.0118
-------------------------------------

Probability for token '<th' (ID: 13708): 9.012579539557919e-06
```
However, the top 1 should be `<th` while it gets very low prob.

**After fix**
The top 10 first token probabilities are
```
--- Top 10 Next Token Predictions ---
1. Token: '<th' (ID: 13708) - Probability: 1.0000
2. Token: '<' (ID: 27) - Probability: 0.0000
3. Token: '>' (ID: 29) - Probability: 0.0000
4. Token: 'def' (ID: 750) - Probability: 0.0000
5. Token: 'think' (ID: 26865) - Probability: 0.0000
6. Token: '<pre' (ID: 10120) - Probability: 0.0000
7. Token: '-th' (ID: 7563) - Probability: 0.0000
8. Token: '<td' (ID: 6868) - Probability: 0.0000
9. Token: '(th' (ID: 24365) - Probability: 0.0000
10. Token: '<thead' (ID: 58167) - Probability: 0.0000
-------------------------------------

Probability for token '<th' (ID: 13708): 0.9999651908874512
```
which is expected.

The following are some selected indicative token logits during the overfitting training after fix (below the `<|endoftext|>` is a padding token):
image

In summary, the previous SFT loss mask **mistakenly shifted one bit, so the model failed to learn the first generated token**. The trained model behaves like adding one undesired noisy token after the input question, as shown in the top 10 first token probabilities.

> [!IMPORTANT]

### DIFF
--- a/verl/trainer/fsdp_sft_trainer.py
+++ b/verl/trainer/fsdp_sft_trainer.py
@@ -360,7 +360,7 @@ class FSDPSFTTrainer:
         input_ids = batch["input_ids"].to(self.device_name)
         attention_mask = batch["attention_mask"].to(self.device_name)
         position_ids = batch["position_ids"].to(self.device_name)
-        loss_mask = batch.pop("loss_mask")[:, :-1].reshape(-1).to(self.device_name)
+        loss_mask = batch.pop("loss_mask")[:, 1:].reshape(-1).to(self.device_name)
         loss_fct = nn.CrossEntropyLoss(reduction="none")
 
         # Context manager for sequence parallel if needed


vermouth1992:
Which dataset did you run? single turn or multi-turn?

BlankCheng:
> Which dataset did you run? single turn or multi-turn?

The overfitting experiment is performed on a single-turn sample.

vermouth1992:
> > Which dataset did you run? single turn or multi-turn?
> 
> The overfitting experiment is performed on a single-turn sample.

Could you please try the multi-turn dataset as I doubt there may be discrepancy?

BlankCheng:
Hi Chi! The specific overfitting setup was: using the [`MultiTurnSFTDataset class`](https://github.com/volcengine/verl/blob/main/verl/utils/dataset/multiturn_sft_dataset.py) in veRL with a single-turn data from [`AM-Think`](https://huggingface.co/a-m-team/AM-Thinking-v1). `truncation=right`.

I've dug a bit deeper into the root cause of this bug, and I'd like to share my findings. 
## MultiTurnSFTDataset Loss Mask
First, the loss mask derivation in `MultiTurnSFTDataset` is:
```python
            loss_mask = [0] * (len(generation_prompt_tokens)) + [1] * (
                len(message_tokens) - len(generation_prompt_tokens)
            )
```
This logic is flawed. In autoregressive LM training, the model predicts the next token. To predict **the first token of the response**, it uses **the last token of the prompt as input**. Therefore, the loss needs to be calculated at the position of the last prompt token. This means the loss_mask at that position should be 1, but the code above incorrectly sets it to 0.

Example:
+ Sequence: [Prompt_1, Prompt_2, Prompt_3, Response_1, Response_2]
+ The model should learn: f(Prompt_1, Prompt_2, Prompt_3) -> Response_1
+ Correct loss_mask: [0, 0, 1, 1, 1] (or [0, 1, 1, 1] after shifting to match with labels)
+ Buggy loss_mask in code: [0, 0, 0, 1, 1]

That is why this PR changes 
```python
loss_mask = batch.pop("loss_mask")[:, :-1].reshape(-1).to(self.device_name)
```
to
```python
loss_mask = batch.pop("loss_mask")[:, 1:].reshape(-1).to(self.device_name)
```
The old implementation removes the last digit loss mask, which doesn't seem to affect training.

## SFTDataset Loss Mask
Also, I checked the single-turn class `SFTDataset`:
```python
        loss_mask = attention_mask.clone()
        if prompt_length > 1:
            # mask out prompt for SFT.
            loss_mask[: min(prompt_length, loss_mask.size(0)) - 1] = 0
        # mask out the last token in response
        loss_mask[min(prompt_length + response_length, loss_mask.size(0)) - 1] = 0
```
which seems to share the same pattern of multi turn that masks the `prompt_length` tokens. Thus, the change seems to be applicable to both dataset classes, while I haven't tested with `SFTDataset`.

vermouth1992:
Yes. I think you are right

vermouth1992:
Thank you for fixing!

vermouth1992 (approved):

fyqqyf:
Thanks! I found the same issue today ;)

puneeshkhanna:
Nice fix for the 1st token prediction. I think earlier code of loss mask would have worked for pretraining. But for SFT since we mask loss for user tokens, loss mask should be based upon assistant tokens which you fixed.

Caelyn-gao:
https://github.com/volcengine/verl/issues/3562

I think "loss_mask = batch.pop("loss_mask")[:, 1:].reshape(-1).to(self.device_name)" will cause endless repetition for single-turn dataset

PR #3295, opened 2025-09-01
[model] feat: Add `Apertus`

EduardDurech:
Pre-release of Apertus from the Swiss AI Initiative

Main modifications from Llama

- xIELU Activation
- QK-norm

Associated Transformers PR https://github.com/huggingface/transformers/pull/39381
Associated vLLM PR https://github.com/vllm-project/vllm/pull/23068
Associated SGLang PR https://github.com/sgl-project/sglang/pull/9774

GSM8K
image
image

### DIFF
--- a/tests/models/test_transformers_ulysses.py
+++ b/tests/models/test_transformers_ulysses.py
@@ -20,7 +20,7 @@ import torch
 import torch.distributed
 from flash_attn.bert_padding import index_first_axis, rearrange, unpad_input
 from torch.distributed import init_device_mesh
-from transformers import AutoModelForCausalLM, LlamaConfig, PretrainedConfig, Qwen2Config
+from transformers import ApertusConfig, AutoModelForCausalLM, LlamaConfig, PretrainedConfig, Qwen2Config
 
 from verl.models.transformers.monkey_patch import apply_monkey_patch
 from verl.protocol import DataProto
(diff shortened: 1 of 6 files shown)


EduardDurech:
@vermouth1992 e2e PPO trainner cli issue seems unrelated and the model_rmpad seems to be because the Transformers version is out of date (requires >=4.56.0)

verl/models/transformers/apertus.py
```diff
@@ -0,0 +1,118 @@
+# Copyright 2025 The SwissAI Initiative
```
vermouth1992, line 1: Is this necessary as we patch `attention_interface` to enable ulysses?

vermouth1992:
I guess only mfu computation is required. Other modification is not necessary. Could you please check if this is correct?

tests/models/test_transformers_ulysses.py
```diff
@@ -66,6 +66,11 @@ def test_configs():
         SequenceParallelConfig(
             Qwen2Config(num_hidden_layers=2, num_attention_heads=32, num_key_value_heads=4), sp_size=8, is_valid=True
         ),
+        SequenceParallelConfig(
```
vermouth1992: Could you add an if statement here and only run ApertusConfig if the transformers are greater then desired? We are running this test for multiple versions to ensure its correctness

verl/models/transformers/apertus.py
EduardDurech: @vermouth1992 not sure, the model documentation is quite out of date
https://verl.readthedocs.io/en/latest/advance/fsdp_extension.html
https://github.com/volcengine/verl/blob/main/verl/models/README.md
but the Ulysses patch https://github.com/volcengine/verl/blob/main/verl/models/transformers/monkey_patch.py uses `verl/models/transformers/apertus.py` to patch, no? Is it not necessary to define?

PR #3332, opened 2025-09-03
[trainer] fix: Loss calculations for grad accumulation steps

puneeshkhanna:
### What does this PR do?

For gradient accumulation steps over micro batches, loss should be normalised before calling loss.backwards(). 
Also add an optimisation so that the all reduce of gradients is only performed in the last accumulation step.

Verified fine-tuning on few open source models with the changes in this PR.

### Checklist Before Starting

### Test

### API and Usage Example

```python
```

### Design & Code Changes

> [!IMPORTANT]

### DIFF
--- a/verl/trainer/fsdp_sft_trainer.py
+++ b/verl/trainer/fsdp_sft_trainer.py
@@ -467,10 +467,8 @@ class FSDPSFTTrainer:
         micro_batches = batch.split(self.config.data.micro_batch_size_per_gpu)
         n_micro_batches = len(micro_batches)
         step_loss = 0
-        for micro_idx, micro_batch in enumerate(micro_batches):            
-            loss = self._compute_loss_and_backward(
-                batch=micro_batch, n_micro_batches=n_micro_batches
-            )
+        for micro_batch in micro_batches:
+            loss = self._compute_loss_and_backward(batch=micro_batch, n_micro_batches=n_micro_batches)
             step_loss += loss.item()
 
         if self.config.model.strategy == "fsdp":


puneeshkhanna:
@vermouth1992 - please review this. Grad accumulation loss calculation is incorrect in SFT code for micro batch steps. I think the loss calculation is correct in RL code for grad accumulation over micro batch steps.

puneeshkhanna:
Addressed precommit reqs of line too long.

Schilings:
Regarding the loss calculation during gradient accumulation: if the loss is computed using a token-level strategy where the number of valid tokens can vary across different micro-batches, wouldn't scaling the loss simply by 1 / num_micro_batches be inaccurate?
image

puneeshkhanna:
@Schilings - What I fixed in this PR is loss averaging for gradient accumulation steps before calling loss.backwards(). In RL code as an example here - https://github.com/volcengine/verl/blob/a4d8952edc6653e9661bbe37722ede13f2a7b0c7/verl/workers/actor/dp_actor.py#L482 it is done like that and it is the standard way. 
Regarding the doubt of valid tokens per DP rank, I think one can leverage balance_dp_token flag.

Schilings:
@puneeshkhanna Thank you for your reply. I understand that the work in your PR addresses a different issue, and I'm not raising this as a concern with your changes. I just wanted to discuss this specific point with you.

The total training batch size is calculated as micro_batch_size_per_gpu * dp_size * grad_acc_steps. I've looked at the source code, and as far as I understand, the balance_dp_token flag ensures that within each gradient accumulation step, different DP ranks can calculate a consistent total number of valid tokens across ranks. However, the number of valid tokens can still vary between different gradient accumulation steps.

Given this potential variation in valid tokens per step, wouldn't directly scaling the loss for each step simply by 1 / num_micro_batches be mathematically inequivalent? Is this level of approximation generally considered acceptable, or is there a mechanism I might be missing that accounts for the differing token counts per step during the final gradient update?

Schilings:
the direct use of 1 / n_micro_batches for averaging the loss per step in gradient accumulation is indeed a common simplification. The main goal is to align the gradient scale with what you'd get from an equivalent large batch size, and it's definitely simpler to implement.

I now understand that while it might not be mathematically precise when the number of valid tokens varies per micro-batch (especially with token-level loss), it's generally considered an acceptable approximation in practice.

verl/trainer/fsdp_sft_trainer.py
```diff
@@ -365,7 +365,10 @@ def _compute_loss_and_backward(self, batch, do_backward=True):
 
         # Context manager for sequence parallel if needed
         context = self.sharding_manager if use_sp else nullcontext()
-        with context, torch.autocast(device_type=self.device_name, dtype=torch.bfloat16):
+        # If last micro-batch in an accumulation step, then implicit grad all-reduce happens in loss backward.
+        # Otherwise, use no_sync() to avoid communication overhead during intermediate accumulation steps.
+        sync_context = nullcontext() if last_micro_in_accum else self.fsdp_model.no_sync()
```
vermouth1992: We don't want to use no_sync here as it will cause higher memory consumption. To prevent silent errors, we disable CPUOffloading explicitly for FSDP1

puneeshkhanna: Removed that part of PR

puneeshkhanna:
@vermouth1992 - please check now updated PR

vermouth1992:
Could you please fix pre-commit?

puneeshkhanna:
> Could you please fix pre-commit?

@vermouth1992 - Done

PR #3369, opened 2025-09-05
[recipe] fix: Fix a Typo in One_Step_Off_Policy and Add async of Generative Reward Model in Response Generation

ZhichaoWang970201:
Fix a typo in verl/workers/fsdp_workers.py: 
    original code: if self.model_config.generation_config is not None
    updated code: if self.generation_config is not None

Add async of generation reward model (GRM):
    As the generative reward model is slow in the call. It is unreasonable to wait for all responses to be generated before sending to GRM for evaluation. So I add an async to start GRM evaluation once individual response generation is finished.


ZhichaoWang970201:
@ji-huazhong Hey Huazhong, would you mind helping me review this pull request and let me know if there is anything to fixed. Thank you in advance.

vermouth1992 (approved):

PR #3454, opened 2025-09-11
[rollout] fix: make agent loop reward worker thread-safe

wuxibin89:
### What does this PR do?

Fixed https://github.com/volcengine/verl/issues/3407

### DIFF
--- a/verl/experimental/agent_loop/agent_loop.py
+++ b/verl/experimental/agent_loop/agent_loop.py
@@ -327,51 +327,27 @@ class RewardManagerWorker:
             config, tokenizer, num_examine=0, **config.reward_model.get("reward_kwargs", {})
         )
         self.rm_executor = rm_executor
-        self.loop = asyncio.get_event_loop()
 
-    async def compute_score(
+    def compute_score(
         self,
         data: DataProto,
     ) -> dict:
         """Compute reward score for agent loop output.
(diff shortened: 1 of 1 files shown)


verl/experimental/agent_loop/agent_loop.py
```diff
@@ -327,51 +327,27 @@ def __init__(self, config: DictConfig, local_path: str, rm_executor: BatchExecut
...
-            data: DataProto from compute reward score
-
-        Returns:
-            torch.Tensor: Reward score tensor.
-        """
         if self.rm_executor is not None:
             res = ray.get(self.rm_executor.submit_task.remote(data))
             data = data.union(res)
 
-        return self.reward_manager(data, return_dict)
```
vermouth1992, line 374: How to make sure they run in parallel?

wuxibin89: For now, each AgentLoopWorker has an paired RewardManagerWorker, it should be enough to overlap the reward calculation with generation, since in most case the reward calculation is pretty fast.

If reward calculation become bottleneck, we can increase AgentLoopWorker num_workers which increases the number RewardManagerWorker as well.

vermouth1992:
Could you rebase main and trigger CI again?

PR #3102, opened 2025-08-18
[Do not Merge] Make VeRL SGLang Native

SuperCB:
### What does this PR do?
We developed this PR based on the following three considerations:
image

1.  Multiple teams and individuals, including the XHS team, have observed that Sglang often becomes exceptionally slow when used with Verl [Issue1](https://github.com/volcengine/verl/issues/2325). Sometimes the first training step becomes extremely slow, and sometimes the slowdown starts from the second step. Through profiling, we discovered that the allreduce operation in Sglang slowed down by dozens of times.（As shown in the figure above, it took 17ms for a single token of data to run TP16's allreduce, which is far too long considering we are using an H100 cluster） We traced the issue to this line of code [LineLink](https://github.com/volcengine/verl/blob/966719c36a958ef9961ffebe41a444e5d86e7366/verl/workers/rollout/sglang_rollout/sglang_rollout.py#L737) (**This causes time slice contention between two processes on the same GPU, and this contention becomes more severe as the TP_size increases.We cannot initiate other processes to run communication operations on the same GPU when SGlang is undergoing the inference process**.).

2.  The original feature in Verl that pre-splits the training data before rollout makes the development of features like partial rollout and streaming rollout very difficult. If each Sglang Engine receives a portion of the data, we would need to maintain the current generation state of each Sglang Engine and track how many tokens each Engine has generated. Such an implementation would be overly complex, and we want to make it simpler and more intuitive.

3.  Algorithm researchers need to rapidly iterate through various experimental schemes during reinforcement learning framework development. This includes exploring different off-policy and on-policy data ratio strategies and verifying various advanced Agent RL mechanisms. However, in the current Verl architecture, the rollout processing code and the inference engine code are tightly coupled (as shown in the code link [Link](https://github.com/volcengine/verl/blob/6469be213e9e51bb31e5fd7ab741655f3ce45536/verl/workers/rollout/sglang_rollout/sglang_rollout.py#L1070)). This means that any modification to any stage of the rollout process requires launching the complete Verl framework for debugging, which significantly increases development costs and iteration cycles. We want to decouple the rollout functionality into an independent "Rollout Manager" component, allowing it to be developed and debugged independently from the Verl framework, thereby reducing the time cost for algorithm engineers to conduct experiments.

[...]

PR #3090, opened 2025-08-17
[sglang] feat: add native sgl server

ChangyiYang:
### What does this PR do?

**Summary**

This PR introduces a native HTTP server implementation for SGLang, aiming to fundamentally improve flexibility, scalability, and integration capabilities. By transitioning to a more robust client-server architecture, this change addresses several core bottlenecks in the current design.

**Key Changes**

* **Engine Replacement** – Replaced the original `sgl.Engine` instance with a native HTTP server. ✅ **Completed**
* **Distributed Optimization** – Utilizing a server-based architecture to remove the requirement of gathering all data to TP rank 0. This change resolves the previous `dist.barrier` timeout issue by replacing the collective wait with per-sample synchronization. 🚧 **In Progress**
* **Router Integration** – Plan to integrate with the native SGLang router for streamlined request handling. 💡 **Nice to have**

**Motivation**

The current `sgl.Engine` driver model presents several architectural challenges, particularly in complex distributed environments. Moving to an HTTP server architecture is motivated by the need to solve the following critical issues:

1.  **Eliminate Data Flow Bottlenecks and Improve Performance:**
    * **Problem:** The data flow logic of the existing driver process is misaligned with the training data flow. It requires all data for a single SGLang instance to be gathered to TP rank 0. This data is then processed by the tokenizer manager and sent via ZMQ to the various schedulers. As a result, the `preprocess` and `postprocess` steps are slower than expected.
    * **Solution:** The HTTP server architecture decentralizes this process, allowing each rank to handle requests independently. This removes the "gather to rank 0" bottleneck, dramatically improving data throughput and overall performance.

2.  **Resolve CPU Resource Contention:**
    * **Problem:** At the request level, the SGLang driver object cannot be pickled for use in subprocesses. This limitation means that the request-level asynchronous rollout logic and the engine itself are forced to compete for the same CPU time slices, leading to performance degradation.
    * **Solution:** By decoupling the request handling (client) from the inference engine (server), we isolate the processes, eliminating the CPU contention and allowing for more efficient resource utilization.

3.  **Fix Distributed Synchronization Timeouts:**
    * **Problem:** The `dist.barrier` timeout is a frequent issue where worker ranks remain idle while waiting for TP rank 0 to complete its intensive processing. This collective wait time creates inefficiency and can lead to failures.
    * **Solution:** The HTTP server model shifts this from a collective barrier to a per-sample synchronization. Workers communicate with the server as needed, removing the long wait times and making the distributed setup more stable and efficient.

[...]

PR #3456, opened 2025-09-11
[1/N][rollout] feat: support vllm/sglang native http server

wuxibin89:
### What does this PR do?

This is the first part to support vllm/sglang native http server in server mode rollout. In native http server mode,
the inference services are launched separately from the training engine, and the model runner share GPU with training engine but in different processes.

We're going to support three deployment modes:
- **hybrid mode**: Training engine and model runner share GPU but in different process. To sync weights, there's a server adapter in training process, which is a http client to send wake_up/sleep/update_weights request to inference server. This is used for on-policy training.
- **standalone mode**: Training engine and inference services have separate GPU resource, disaggregated architecture. This is used for off-policy training.
- **colocated mode**: Like hybrid mode, but without server adapter since no need to sync weights. This is mainly used for GRM service (LLM as a judge).
image

Following PR will be:
- [2/N] support DP+EP
- [3/N] standalone rollout with weight transfer by NCCL/UCX
- [4/N] colocated GRM service with wake_up/sleep(without weight synchronization)
- [5/N] switch to `/generate` http api with token-in-token-out: currently sglang has `/generate` api but may need some effort to support multi-modal; while vllm still lack `/generate` api
- [6/N] switch to sglang/vllm router with better kv-cache awareness load balance

The native http server is inspired by the design of [slime](https://github.com/THUDM/slime), thanks to their prior work. Also credit to @ChangyiYang @zhaochenyang20  https://github.com/volcengine/verl/pull/3090 @SuperCB  https://github.com/volcengine/verl/pull/3102 with their prior contribution.


verl/trainer/config/ppo_megatron_trainer.yaml
```diff
@@ -24,6 +23,7 @@ actor_rollout_ref:
   nccl_timeout: 600 # seconds, default is 10 minutes for torch, you can set it to a larger value if you have long-running operations like 32B or 72B model using megatron
 
   model:
+    _target_: verl.workers.config.HFModelConfig
```
vermouth1992: This config change is breaking. We should make a separate PR for this

verl/workers/megatron_workers.py
```diff
@@ -388,15 +387,7 @@ def _build_rollout(self, trust_remote_code=False):
 
         # 1. parse rollout and huggingface model config
         rollout_config: RolloutConfig = omega_conf_to_dataclass(self.config.rollout)
-
```
vermouth1992, line 391: Again, this change is breaking. We should make a separate PR. Also, we should change critic as well

verl/trainer/config/ppo_megatron_trainer.yaml
wuxibin89: Fixed, should make a separate PR  and unify megatron and fsdp model config.

verl/workers/rollout/replica.py
```diff
@@ -0,0 +1,207 @@
...
+        except ImportError:
+            import sys
+            from unittest.mock import Mock
+
+            mock_vllm = Mock()
+            mock_vllm._custom_ops = Mock()
+            mock_vllm._custom_ops.scaled_fp8_quant = Mock()
+
+            sys.modules["vllm"] = mock_vllm
+            sys.modules["vllm._custom_ops"] = mock_vllm._custom_ops
```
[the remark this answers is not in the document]
wuxibin89: This can be removed once we bump to sglang>=0.5.1, which it will not depend on vllm.

verl/workers/rollout/sglang_rollout/async_sglang_server.py
```diff
@@ -14,81 +14,268 @@
...
+
+        self._server_port, self._server_task = await run_unvicorn(app, server_args)
+
+    async def wake_up(self):
+        # NOTE: If it's hybrid mode, we have to call all workers to switch between trainer mode and rollout mode;
+        # otherwise, we directly call the engine to release or resume GPU memory.
+        if self.rollout_mode == RolloutMode.HYBRID:
+            await asyncio.gather(*[worker.wake_up.remote() for worker in self.workers])
+        else:
+            # FIXME(@wuxibin): sglang seems resume with random weights.
```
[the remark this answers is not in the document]
wuxibin89: Will fix on next PR.

verl/workers/rollout/sglang_rollout/sglang_rollout.py
```diff
@@ -1495,101 +1493,86 @@ async def update_weights(self, weights: Generator[tuple[str, torch.Tensor], None
...
+        Notes:
+          - For the best performance of `rebuild_cuda_tensor`, it is recommended to:
+              1. Enable `RAY_EXPERIMENTAL_NOSET_CUDA_VISIBLE_DEVICES`.
+              2. Manually set `CUDA_VISIBLE_DEVICES=0,1,2,3,4,5,6,7`
+            when using Tensor Parallelism (TP >= 8).
+          - See reference implementations in SLIME:
+            - Main logic: https://github.com/THUDM/slime/blob/fb7605cc5fb09af0f9369d37f7192f12bddee577/slime/ray/ppo_actor.py#L452
+            - runtime envs: https://github.com/THUDM/slime/blob/fb7605cc5fb09af0f9369d37f7192f12bddee577/slime/ray/ppo_actor.py#L39
+        """
+        # FIXME(@wuxibin): device_mesh is not right in multi-node case.
```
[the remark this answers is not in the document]
wuxibin89: Will fix on next PR.

verl/workers/megatron_workers.py
wuxibin89: Fixed as above.

lizipao:
Thank you for your contribution. May I ask what configuration changes are used to implement the switching between these modes?

lizipao:
What's the difference between the server mode implemented in #3090 and the current HYBRID mode? Wasn't the previous server mode also running in separate processes?

wuxibin89:
> Thank you for your contribution. May I ask what configuration changes are used to implement the switching between these modes?

@lizipao 
1. `hybrid mode`: this is the default mode, no additional configuration change needed.
2. `colocated mode`: this is for GRM(generative reward model) only, still needed to develop in 4/N.
3. `standalone mode`: this is for off-policy training, we will refactor `verl/recipe/one_step_off_policy` with this PR in 3/N. The configuration changes is in:
https://github.com/volcengine/verl/blob/main/recipe/one_step_off_policy/config/one_step_off_ppo_trainer.yaml#L9-L14

lizipao:
May I ask why skip_tokenizer_init is set to True?

wuxibin89:
> May I ask why skip_tokenizer_init is set to True?

Because we want token-in-token-out as default.

Cesilina:
想问下，多机情况下，其他节点能正常访问rollout server的address吗

wuxibin89:
> 想问下，多机情况下，其他节点能正常访问rollout server的address吗

@Cesilina 可以的

Cesilina:
> > 想问下，多机情况下，其他节点能正常访问rollout server的address吗
> 
> @Cesilina 可以的

嗯嗯，谢谢，

lizipao:
Hi there,

I'd like to understand the differences and relationship between the AsyncHttpServerAdapter implemented in #3090 and the current SGLangHttpServer used in HYBRID mode.

From my code analysis, I found that:

When rollout_mode is set to async, the system uses SGLangHttpServer (as a Ray Actor)
Otherwise, it uses AsyncHttpServerAdapter (as an HTTP client adapter)
Could you briefly explain the differences between these two implementations from both performance and architectural design perspectives? Specifically:

The differences in their call paths (internal API vs HTTP requests)
Their respective use cases and performance characteristics
Why we need two different implementations
Thanks for your time!

lizipao:
在这个模式下我发现AgentLoopWorker里，没有传top_k，并且config.calculate_log_probs的默认值为False，这是正常的吗？

bhks:
Hi @wuxibin89 and @vermouth1992 ,

Thank you for this PR.

Is it possible for users to use the SGLang Router with current code in main ?

Based on my analysis we are not passing the router ip and port from top level trainer modules. I looked into the `SGLangRollout` class and `_init_inference_engine` method and don't see we are passing any router ip and port yet.

I only see the `AsyncHttpServerAdapter` and `HttpServerAdapter` have the router ip and port. Wondering what changes we need to make to just add that suppprt ?

Do we have to make other changes before we can enable the sglang router support. Given you have added as last item in your list of items to complete.

Ericnano:
请教一下这个pr支持的vllm是哪一个版本的，在0.9.1上会报错No module named 'vllm.v1.engine.utils'

ji-huazhong:
@Ericnano >0.9.1

PR #3475, opened 2025-09-15
[trainer] feat: Data buffering to conserve data during filtering

rhinosaur0:
### What does this PR do?

Current DAPO implementation discards all the extra samples during filtering below, wasting lots of samples

```python
traj_bsz = self.config.data.train_batch_size * self.config.actor_rollout_ref.rollout.n
batch = batch[:traj_bsz]
```

This PR wraps a SmartDataloader object around the dataloader, returning unused samples back to the front of the dataloader to be used by the next batch. It also has the option to return the filtered-out samples to the BACK of the dataloader to be used at the end of the epoch, so a supposedly improved model can train on these samples again.

Screenshot 2025-09-14 at 10 48 33 PM

### Design

`class DataBuffer` simply holds a list of `DataProto` objects for easy access.
`class SmartDataLoader` controls two buffers, unused_buffer (during `batch = batch[:traj_bsz]`) which will contain samples to be called _immediately_ in the next batch, and an optional filtered_buffer, which will contain samples to be called at the end of the epoch when the dataloader is exhausted.

In other words, the `SmartDataLoader` is a 3-stage data-retrieval process:
1. From unused_buffer
2. From `train_dataloader`
3. From filtered_buffer

- `data_con` in metrics logs the number of samples saved (which is usually a lot)

### Usage Guide

Simply add 
```sh
data.databuffer.enable=true \
data.databuffer.keep_filtered=true \
```
to any run.sh scripts


vermouth1992 (dismissed):

vermouth1992:
Could you fix pre-commit?

vermouth1992:
You can follow the instructions here: https://github.com/volcengine/verl/blob/main/CONTRIBUTING.md

Issue #2078, opened 2025-06-18
Performance Issue: reward_module.to(torch.bfloat16) Takes a Long Time After apply_monkey_patch in RewardModelWorker (verl/workers/fsdp_workers.py)


xingyunjohn1:
In the `RewardModelWorker` class (`_build_worker` method) in `verl/workers/fsdp_workers.py`, I noticed the following code:
```python
with init_context(), warnings.catch_warnings():
    warnings.simplefilter("ignore")
    model_config.classifier_dropout = 0.0
    reward_module = AutoModelForTokenClassification.from_pretrained(
        pretrained_model_name_or_path=local_path,
        config=model_config,
        torch_dtype=torch.bfloat16,
        attn_implementation="flash_attention_2",
        trust_remote_code=trust_remote_code,
    )

    apply_monkey_patch(
        model=reward_module,
        use_remove_padding=config.model.get("use_remove_padding", False),
        ulysses_sp_size=self.ulysses_sequence_parallel_size,
    )

    reward_module.to(torch.bfloat16)
```
The model is initially loaded with `torch.bfloat16` precision. However, after calling `apply_monkey_patch`, the model parameters' precision changes. As a result, running r`eward_module.to(torch.bfloat16)` on a large model (e.g., 32B) takes a very long time—up to half an hour or longer. Based on GPU utilization monitoring, it appears that this operation is only performed on a single GPU, even if multiple GPUs are available. This precision conversion process consumes a large amount of CPU memory, with a total usage of approximately 240 GB.

Question:
Is there any way to speed up this operation?

Why does `reward_module.to(torch.bfloat16)` take so long after `apply_monkey_patch`?
Is it possible to leverage multiple GPUs for this process or otherwise accelerate it?
Thank you!

Below my configuration:
```yaml
# the ppo trainer split config will override default ppo_trainer.yaml

hydra:
  searchpath:
    - file:///checkpoint/binary/train_package/verl/verl/trainer/config

defaults:
  - ppo_trainer
  - _self_

data:
  tokenizer: null
  train_files:
    - /data/ppo_data_160/train.parquet # 160 samples
  val_files:
    - /data/ppo_data_160/test.parquet # 160 samples
  prompt_key: prompt
  max_prompt_length: 4096
  max_response_length: 512
  train_batch_size: 48
  val_batch_size: null # DEPRECATED: Validation datasets are sent to inference engines as a whole batch, which will schedule the memory themselves
[...]

PR #3493, opened 2025-09-16
Fix scale problem during checkpoint loading for fsdp2

henrique:
### What does this PR do?

Only load HF checkpoints on rank 0 for fsdp2. Later the model is already broadcasted to all ranks in 'fsdp2_load_full_state_dict'

We have a problem when starting larger jobs, e.g. 512+ ranks, as each rank tries to simultaneously fetch the checkpoints from the file system/HF. This causes a massive resource contention on distributed file systems and greatly increases the length and variability of the loading, even at smaller scales. As we increase the number of ranks, it will eventually lead to nccl timeouts as the fastest rank has to wait for the slowest, which will break the initialization.

### Test

Model loading only happens on rank 0, is faster and doesn't timeout at scale; loss curve matches the original.


verl/workers/fsdp_workers.py
```diff
@@ -366,12 +366,16 @@ def _build_model_optimizer(
                 else:
                     actor_module_class = AutoModel
 
-            actor_module = actor_module_class.from_pretrained(
-                pretrained_model_name_or_path=local_path,
-                torch_dtype=torch_dtype,
-                config=actor_model_config,
-                trust_remote_code=trust_remote_code,
-            )
+            if self.rank > 0 and self.config.actor.strategy == "fsdp2":
```
vermouth1992, line 369: I guess init_context should also include this feature?

EduardDurech:
Related issue https://github.com/volcengine/verl/issues/2078

verl/workers/fsdp_workers.py
henrique: Hi @vermouth1992 
I guess you mean something that can be used also by both CriticWorker and RewardModelWorker?

vermouth1992: Actually, we are deprecating FSDP/Megatron RewardModelWorker to favor server-based reward model worker

EduardDurech: @vermouth1992 what PR should we rebase?

henrique:
closing in favor of https://github.com/volcengine/verl/pull/4644

PR #3499, opened 2025-09-17
[trainer] refactor: remove deprecated sharding manager

techkang:
### What does this PR do?

Remove the deprecated sharding manager because this code may confuse beginners who are searching for a specific implementation.


vermouth1992:
We don't want to remove in this version because they are numerous packages depends on verl. We may need to remove after releasing v0.6

vermouth1992:
Let's keep the code for one more version

techkang:
Closed as not planned for now.

PR #3530, opened 2025-09-19
[2/N][rollout] feat: support vllm/sglang DP+EP in server mode

wuxibin89:
### What does this PR do?

Following https://github.com/volcengine/verl/pull/3456, support vllm/sglang DP+EP in server mode.


wuxibin89:
For vllm, sleep level=2 does not work with expert parallel, fixed in https://github.com/vllm-project/vllm/pull/25458. As a workaround, when EP enabled, sleep(level=1) for now.

wuxibin89:
For sglang, there's same issue which is already fixed in https://github.com/sgl-project/sglang/pull/8676, we should upgrade sglang to 0.5.2.

verl/workers/rollout/vllm_rollout/vllm_async_server.py
```diff
@@ -27,14 +27,18 @@
...
 from vllm.entrypoints.openai.api_server import (
     build_app,
-    build_async_engine_client,
     init_app_state,
 )
 from vllm.inputs import TokensPrompt
 from vllm.outputs import RequestOutput
-from vllm.utils import FlexibleArgumentParser
+from vllm.usage.usage_lib import UsageContext
+from vllm.utils import FlexibleArgumentParser, get_tcp_uri
```
sanxing-chen, line 38: `get_tcp_uri` is introduced in vllm v0.9.0, the installation script needs to be updated.

Issue #3516, opened 2025-09-18
[trainer] when train vlm using lora in VLLM_SLEEP_LEVEL=2 ( if vllm version >= 0.8.5 ) and without using layerd_summon, then the lora layers will not be updated to vllm


WncFht:
### System Info

Previously, I trained qwen2.5-vl-3B using lora and grpo. When I use the following scripts, it will not correctly work.
```
set -x

export CUDA_VISIBLE_DEVICES=1,2,3,4
# export HF_HONE=/mnt/data2/huggingface
export TORCH_CUDA_ARCH_LIST=9.0
export HF_DATASETS_CACHE=$HOME/data/.cache/huggingface/datasets
export WANDB_MODE=offline

MODEL=Qwen/Qwen2.5-VL-3B
TARGET_MODULES='["q_proj","k_proj","v_proj","o_proj","gate_proj","up_proj","down_proj"]'
# If you are using vllm<=0.6.3, you might need to set the following environment variable to avoid bugs:
# export VLLM_ATTENTION_BACKEND=XFORMERS

python3 -m verl.trainer.main_ppo \
    algorithm.adv_estimator=grpo \
    data.train_files=$HOME/data/geo3k/train.parquet \
    data.val_files=$HOME/data/geo3k/test.parquet \
    data.train_batch_size=512 \
    data.max_prompt_length=1024 \
    data.max_response_length=2048 \
    data.filter_overlong_prompts=True \
    data.truncation='error' \
    data.image_key=images \
    actor_rollout_ref.model.path=$MODEL \
    actor_rollout_ref.model.use_remove_padding=True \
    actor_rollout_ref.model.lora_rank=64 \
    actor_rollout_ref.model.lora_alpha=32 \
    actor_rollout_ref.model.target_modules=$TARGET_MODULES \
    actor_rollout_ref.model.exclude_modules='.*visual.*' \
    actor_rollout_ref.model.enable_gradient_checkpointing=True \
    actor_rollout_ref.model.use_liger=True \
    actor_rollout_ref.actor.optim.lr=3e-5 \
    actor_rollout_ref.actor.ppo_mini_batch_size=128 \
    actor_rollout_ref.actor.ppo_micro_batch_size_per_gpu=8 \
    actor_rollout_ref.actor.use_kl_loss=True \
    actor_rollout_ref.actor.kl_loss_coef=0.01 \
    actor_rollout_ref.actor.kl_loss_type=low_var_kl \
[...]

PR #3541, opened 2025-09-19
[trainer,rollout] fix: ensure LoRA weights are loaded when vllm_sleep_level=2 and without using layerd_summon

piood:
### What does this PR do?

Fix issue where VLLM would only load base model parameters and not LoRA parameters when VLLM_SLEEP_LEVEL == 2 and not using layered_summon.

This fixes the LoRA trainer error where the first rollout would only use base model parameters, and subsequent rollouts would correctly load LoRA parameters.

Fixes: https://github.com/volcengine/verl/issues/3516
Related PR: https://github.com/volcengine/verl/pull/3461

### Checklist Before Starting

### Test

### API and Usage Example

```python
```

### Design & Code Changes

> [!IMPORTANT]

### DIFF
--- a/verl/workers/fsdp_workers.py
+++ b/verl/workers/fsdp_workers.py
@@ -613,10 +613,7 @@ class ActorRolloutRefWorker(Worker, DistProfilerExtension):
         self.base_sync_done: bool = "dummy" not in self.config.rollout.load_format
         self.layered_summon = self.config.rollout.get("layered_summon", False)
         if VLLM_SLEEP_LEVEL == 2 and not self.layered_summon:
-            self.force_reload = True
             self.base_sync_done = False
-        else:
-            self.force_reload = False
 
         # 5. switch to trainer mode
         # NOTE: It's critical that hybrid engine in trainer mode initially to load checkpoint.
@@ -680,7 +677,7 @@ class ActorRolloutRefWorker(Worker, DistProfilerExtension):
             await self.rollout.resume(tags=["kv_cache"])
         log_gpu_memory_usage("After resume kv_cache", logger=logger)
 
-        self.base_sync_done = not self.force_reload
+        self.base_sync_done = True
         # important: need to manually set the random states of each tp to be identical.
         self.torch_random_states = get_torch_device().get_rng_state()
         get_torch_device().set_rng_state(self.gen_random_states)


techkang:
Have you tried to train on your patch with `layer_summon=False`? When `VLLM_SLEEP_LEVEL=2`, the base model's weights are destroyed every iteration. After this patch, I think it cannot load weights again. A simple way to fix this is let `VLLM_SLEEP_LEVEL=1` whenever lora is enabled.

piood:
> Have you tried to train on your patch with `layer_summon=False`? When `VLLM_SLEEP_LEVEL=2`, the base model's weights are destroyed every iteration. After this patch, I think it cannot load weights again. A simple way to fix this is let `VLLM_SLEEP_LEVEL=1` 。

You are right.There are some problem, i will continue fix and test it.

WncFht:
@techkang I think the simple way you mentioned maybe not that great. In that way, all the lora situation should set `VLLM_SLEEP_LEVEL` to be 1, which will make the lora mode cannot offload the base model and potentially when the `rollout.load_format` equals to `dummy`, it will never load base model.

So I think we may implement a loading strategy to first load base model and then load the lora adapter in every loading turn. If you agree with me, I will comlete the PR recently, since it may bring more change in the params collecting and model updating logic.

techkang:
@WncFht Thanks for you discussion. But I still think when `rollout.load_format` equals to `dummy`, the base model will be loaded at the first iteration with these code: https://github.com/volcengine/verl/blob/d33c85e2c779da1203e54275645b7b30f7fe3ce1/verl/utils/fsdp_utils.py#L638-L651

piood:
@techkang @WncFht I have polished this PR and tested it. It now correctly loads the base model parameters before the LoRA parameters in each iteration when vllm_sleep_level==2 and layer_summon=False.

verl/workers/fsdp_workers.py
```diff
@@ -653,6 +650,17 @@ async def rollout_mode(self):
             params, getattr(self.actor_module_fsdp, "_fsdp_wrapped_module", self.actor_module_fsdp)
         )
 
+        if peft_config is not None and VLLM_SLEEP_LEVEL == 2 and not self.layered_summon:
```
techkang: I think we should use `getattr(self.rollout, 'sleep_level', None) == 2` instead of `VLLM_SLEEP_LEVEL == 2 and not self.layered_summon` to reduce code duplication.

techkang: Besides, can you add some comment here to explain why we load base model weights? I think it's something like "When LoRA is enabled alongside a rollout sleep level of 2, the base model's weights are destroyed every sleep time. Therefore, we must load both the base weights and the LoRA weights separately and update each through their respective interfaces."

verl/workers/fsdp_workers.py
```diff
@@ -672,6 +680,15 @@ async def rollout_mode(self):
         if self.config.rollout.free_cache_engine:
             await self.rollout.resume(tags=["weights"])
         log_gpu_memory_usage("After resume weights", logger=logger)
+
+        if peft_config is not None and VLLM_SLEEP_LEVEL == 2 and not self.layered_summon:
+            per_tensor_base_params = (
+                (name, param.to(device, non_blocking=True).full_tensor() if isinstance(param, DTensor) else param)
+                for name, param in base_model_params.items()
+            )
+            await self.rollout.update_weights(per_tensor_base_params, peft_config=peft_config, base_sync_done=False)
```
techkang: I think we should remove `perf_config` in this line to emphasize this is only used to update base model's weights.

techkang:
cc @wuxibin89

piood:
@techkang I found an issue when LoRA is enabled with vllm_sleep_level=2: when using dummy load format (base_sync_done=False), both `params` and `base_model_params` in rollout_mode contain the same base model parameters, causing duplication.

techkang:
Yes, the duplication will only occur at the first iteration. And the duplication shares the same memory so I don't think it's a serious problem.

piood:
@techkang Got it, thanks!

techkang (approved):

PR #3548, opened 2025-09-20
[algo, worker] refactor: async reward computation for 20% verl training speedup

XuDongYangthu:
## What and Why

Refactor reward computation to support **asynchronous reward scoring**, especially for QA/Kefu scenarios where rewards are not directly available.

Previously, reward calculation was blocking and slow — now we decouple it via background threads, achieving **~20% end-to-end training speedup** in ReVR workloads.

## Changes

- Add `verl/workers/reward_manager/kefu_mthread.py`: async reward scorer using multithreading
- Support `reward_model.reward_manager=intelligent_kefu_async` in config/script
- Non-breaking: sync mode remains default; async is opt-in
- Minor refactor of reward pipeline for cleaner async integration

## Test

- Tested on internal QA dataset (10k samples)
- Training throughput ↑ 20%
- No accuracy drop observed

## Usage

In bash script or config:

```bash
--reward_model.reward_manager=intelligent_kefu_async
```

or in YAML:
```bash
reward_model:
  reward_manager: intelligent_kefu_async
```

## Commit Message
[algo, worker] refactor: async reward for 20% ReVR speedup

- Add kefu_mthread.py for async reward scoring in QA scenarios
- Support reward_model.reward_manager=intelligent_kefu_async
- Refactor reward pipeline to enable non-blocking computation
- Verified 20% training throughput gain and no accuracy drop


wuxibin89:
We should be careful that some reward computation is not thread-safe, https://github.com/volcengine/verl/issues/3407

PR #3577, opened 2025-09-23
[misc] feat: remove redundant default params

techkang:
### What does this PR do?
This PR introduces two changes:

1. Removal of redundant default parameters: Default optimizer values are already set in the .yaml configuration file. Defining them again in other files is redundant and can cause confusion for users.

2. Alignment of warm-up step logic: Changed the condition from `num_warmup_steps < 0` to `num_warmup_steps <= 0`. This aligns the code with the documentation in the YAML file and matches the implementation in Megatron.
https://github.com/volcengine/verl/blob/main/verl/trainer/config/actor/actor.yaml#L132


vermouth1992:
Could you rebase main and run CI again?

PR #3578, opened 2025-09-23
[megatron] feat: use flash as default attention_backend

ISEEKYAN:
### What does this PR do?

1. add mapping from string to megatron's enum for attention_backend choice.
2. use flash attention as default attention_backend, for consistency to FSDP

### DIFF
--- a/verl/models/mcore/config_converter.py
+++ b/verl/models/mcore/config_converter.py
@@ -389,10 +389,8 @@ def hf_to_mcore_config_llama4(
 
 
 def mapping_string_to_attn_backend(args: dict) -> dict:
-    if "attention_backend" in args:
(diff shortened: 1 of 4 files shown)


ISEEKYAN:
> will switching to `flash` as default backend have potential break for some training recipes? also does it have any performance/efficiency regressions or boost?

@ccclyu  There is no performance difference compared with the previous default `TE.fused_attn`. But in some cases (DAPO with 20k response with qwen3 14B/30B) I found the `fused_attn` will make the training crash after a long training (32*H20*7days), while no crash found when `flash_attn` is used. with TE2.2.

I am still trying to find out what is the root cause and whether the crash is fixed with new versions of transformer engine(TE2.7).

vermouth1992 (approved):

PR #3617, opened 2025-09-25
[trainer] feat: remove YAML config in verl

techkang:
## What does this PR do?
This PR tries to remove YAML config in verl. This patch can successfully train sft engine.

## background
In verl, we using both .yaml files and `dataclass` to manage configs. However, there are two drawbacks: 
1. While YAML files take the highest priority, the IDE offers limited assistance in helping users trace configurations from the code back to the YAML file. 
2. Both YAML files and dataclasses define default values. If conflicts arise between these default values, the system cannot properly alert users.

### DIFF
--- a/verl/trainer/sft_trainer.py
+++ b/verl/trainer/sft_trainer.py
@@ -371,9 +371,9 @@ def main(cfg) -> None:
     """Main entry point with pure dataclass configuration.
 
     Args:
-        cfg: SimpleSFTConfig instance with all training parameters
+        cfg: SFTConfig instance with all training parameters
     """
-    # cfg is now a strongly typed SimpleSFTConfig dataclass instance
+    # cfg is now a strongly typed SFTConfig dataclass instance
     config = OmegaConf.to_object(cfg)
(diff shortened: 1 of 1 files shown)


verl/workers/config/engine.py
```diff
@@ -113,8 +123,17 @@ class FSDPEngineConfig(BaseConfig):
...
-    forward_only: bool = False
-    strategy: str = "fsdp"
+    strategy: str = field(default="fsdp", init=False)
 
     def __post_init__(self):
         assert self.strategy in ["fsdp", "fsdp2"], f"strategy {self.strategy} not supported"
+
+
+@dataclass
+class AllEngineConfig(FSDPEngineConfig, McoreEngineConfig):
```
vermouth1992, line 133: This is undesired. We will start to expand the engine to veomni, torchtitan, etc...

vermouth1992: This may create silent error if one mis-specify a parameter to another engine

techkang:
Closed as not planned by now.

PR #3663, opened 2025-10-02
[misc] fix: Allow HF model ID with `use_shm`

EduardDurech:


EduardDurech:
ci unrelated

verl/utils/fs.py
```diff
@@ -210,6 +210,19 @@ def copy_to_local(
     """
     # Save to a local path for persistence.
     local_path = copy_local_path_from_hdfs(src, cache_dir, filelock, verbose, always_recopy)
+
+    if use_shm and isinstance(local_path, str) and not os.path.exists(local_path):
```
wuxibin89, line 214: Why only download model when use_shm=True?

EduardDurech: Didn't want to cause issues with other paths using the method, if you want I can have all calls

PR #3679, opened 2025-10-04
[trainer, worker] feat: more flexible and easy-to-use reward model

yyDing1:
The current reward model implementation faces the following challenges:

1. Model Support: It is primarily designed for discriminative models and lacks robust support for generative reward models.
2. Complexity: It relies on heavy-weight backends like FSDP or Megatron, which are often unnecessary for typical reward model inference tasks.
3. Flexibility: The batch-level synchronization mechanism hinders the implementation of more flexible, sample-level reward functions for the developers.

### What this PR does

To address these issues, this PR introduces a more flexible and easy-to-use reward model design. Specifically, it implements two main classes: `RewardModelManager` and `RewardManagerWorker`, with some runnable scripts in `recipe/fapo`.

image

- `RewardModelManager` first launches multiple reward servers and then adopts a router-based approach to manage these servers (using [SGLang Router](https://docs.sglang.ai/advanced_features/router.html)), distributing requests to reward servers.
- `RewardManagerWorker` retrieves the remote actor handle, providing users with greater flexibility in designing custom reward functions. For example, users can easily implement a customized reward function like the following:

```python
async def compute_score(
    data_source: str,
    solution_str: str,
    ground_truth: str,
    extra_info: dict,
    reward_router_address: str,
    reward_model_tokenizer: PreTrainedTokenizer,
):
    # Compute rule-based reward score
    rule_based_score = ...

    # Compute GRM reward score
    grm_prompts = ...
    grm_prompt_ids = ...
    # Users can directly call the reward model
    grm_outputs = post(f"{http://{reward_router_address}/generate}", ...)  # post request to reward router
    ...

    # Final reward score
    final_score = ...

    return final_score
```

This implementation provides a `reward_model` interface in the `compute_score` method, maximizing flexibility and convenience for algorithmic design.

Note that this is an asynchronous function, so efficiency is not a concern—each sample is processed asynchronously.

### Integration with AgentLoop

This PR introduces asynchronous reward computation for individual samples (`async def run_single(self, data: DataProto) -> dict`) and leverages an event loop to handle reward computation in parallel, significantly improving processing efficiency.

Moreover, this implementation can be integrated with `agentloop` for improved efficiency (has been implemented):

image

In this mode, the reward model operates independently from the rollout process (standalone mode), enabling a natural async data flow where each sample undergoes reward rollout immediately after actor rollout.

With this implementation, code redundancy is reduced in the existing reward model while maximizing flexibility for user-customized reward functions.

### Runable Scripts

A runnable example is provided in `recipe/fapo/`. The newly introduced parameters for this implementation are placed in `fapo/config` and will be integrated into the main codebase upon completion of the refactoring.

### DIFF
--- a/.github/workflows/reward_model_sglang.yml
+++ b/.github/workflows/reward_model_sglang.yml
@@ -102,7 +102,7 @@ jobs:
       - name: Prepare gsm8k dataset
         run: |
           ray stop --force
-          python3 examples/data_preprocess/gsm8k.py --local_dataset_path ${HOME}/data/gsm8k
+          python3 examples/data_preprocess/gsm8k.py --local_dir ${HOME}/data/gsm8k
       - name: Running sglang reward model tests on 8 L20 GPUs
         run: |
           unset http_proxy https_proxy HTTP_PROXY HTTPS_PROXY
--- a/.github/workflows/reward_model_vllm.yml
+++ b/.github/workflows/reward_model_vllm.yml
@@ -101,7 +101,7 @@ jobs:
       - name: Prepare gsm8k dataset
         run: |
           ray stop --force
-          python3 examples/data_preprocess/gsm8k.py --local_dataset_path ${HOME}/data/gsm8k
+          python3 examples/data_preprocess/gsm8k.py --local_dir ${HOME}/data/gsm8k
       - name: Running vllm reward model tests on 8 L20 GPUs
         run: |
           unset http_proxy https_proxy HTTP_PROXY HTTPS_PROXY


recipe/fapo/run_fapo.sh
```diff
@@ -0,0 +1,150 @@
...
+    actor_rollout_ref.rollout.val_kwargs.do_sample=True \
+    actor_rollout_ref.rollout.val_kwargs.n=1 \
+    actor_rollout_ref.ref.fsdp_config.param_offload=${offload} \
+    actor_rollout_ref.ref.ulysses_sequence_parallel_size=${sp_size} \
+    actor_rollout_ref.actor.fsdp_config.fsdp_size=${fsdp_size} \
+    reward_model.enable=True \
+    reward_model.enable_resource_pool=True \
+    reward_model.n_gpus_per_node=8 \
+    reward_model.nnodes="${RM_NODES}" \
+    reward_model.model.path=xxxxxx \
```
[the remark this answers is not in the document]
yyDing1: The path will be replaced once the FAPO-GenRM model is released

wuxibin89:
We need a comprehensive design doc for reward model.

verl/experimental/reward/reward_model.py
```diff
@@ -0,0 +1,140 @@
...
+        asyncio.run(run_all())
+
+    # just for test purpose
+    def generate_sequences(self, prompts: DataProto, sampling_params: dict):
+        router_inputs = [
+            {"prompt_token_ids": raw_prompt_ids} for raw_prompt_ids in prompts.non_tensor_batch.get("raw_prompt_ids")
+        ]
+        responses = ray.get(
+            [
+                self.router_handle.generate.remote(
```
wuxibin89: For router, I think we should access router by http endpoint `/generate` instead of actor handle?

verl/experimental/reward/reward_model.py
```diff
@@ -0,0 +1,140 @@
...
+
+        Args:
+            config (RewardModelConfig): Reward model configuration.
+            worker_group (RayWorkerGroup, optional): Worker group. Defaults to None.
+        """
+        self.config = config
+        self.worker_group = worker_group
+        self._initialize_llm_servers()
+        self._initialize_router()
+        if self.config.rollout.free_cache_engine:
```
wuxibin89, line 47: Only need to sleep in colocated mode.

verl/workers/config/reward_model.py
```diff
@@ -74,73 +38,18 @@ class SandboxFusionConfig(BaseConfig):
...
-    # for generative models, response_length should be set > 0
-    response_length: int = 0
-
-    # resource pool config: for colocate / standalone mode
     enable_resource_pool: bool = False
     n_gpus_per_node: int = 0
     nnodes: int = 0
 
     # reward model args
-    dtype: str = "bfloat16"
```
wuxibin89, line 115: Should we adjust `verl/trainer/config/reward_model/reward_model.yaml` for this change?

verl/experimental/reward/reward_manager.py
```diff
@@ -0,0 +1,52 @@
...
+
+logger = logging.getLogger(__file__)
+logger.setLevel(os.getenv("VERL_LOGGING_LEVEL", "WARN"))
+
+
+@ray.remote
+class RewardManagerWorker:
+    def __init__(self, config: DictConfig, reward_model_handle: ray.actor.ActorHandle = None):
+        self.config = config
+        self.reward_model_handle = reward_model_handle
```
wuxibin89: Again, we should access router by http endpoint `/generate` instead of actor handle.

verl/experimental/reward/reward_model.py
yyDing1: Yes, the `self.router_handle` is an `ActorHandle(SGLangRouter)`, where the `generate` method is implemented by sending a request to `f"http://{self.router_address}/generate"`.

https://github.com/volcengine/verl/pull/3679/files#diff-0a76549e12119a277920a9f11f77a9a9c003a9edb66b6e477d48fe35be3a73d7R78-R105

verl/experimental/reward/reward_model.py
yyDing1: Yes, the reward model directly uses the `{vLLM/SGLang}Replica` as the backend, and this is implemented within the backend implementation. Specifically, `vLLMHttpServer.sleep()` and `SGLangHttpServer.sleep()`.

verl/workers/config/reward_model.py
yyDing1: Yes, this PR introduces and refactors the configuration of the reward model. The refactored configuration is temporarily placed in `recipe/fapo/config/rm_config` as an add-on to `ppo_trainer`. I’m concerned that a full refactor might break many CI tests, so for now it’s kept in the recipe directory. A future PR will deprecate this setup and switch to the new configuration once the reward model and router setup becomes stable.

yyDing1:
> We need a comprehensive design doc for reward model.

Yes, I will add a doc for this new feat.

verl/experimental/reward/reward_model.py
wuxibin89: The current request is: router_handle->router actor->router http server, I mean we can directly use http client->router http server.

yyDing1: Tried to fix this issue. The current implementation exposes a router address to customized reward function and leaves the api call details to user customization.

yyDing1:
I have tested the stability:
- [x] Refactored async reward computation pipeline: Tested under the DAPO-7B setting, and it worked without any issues.
- [ ] Router + reward model implementation: Tested with DAPO-7B + GenRM, and found minor stability issues with the SGLang router — a small fraction of requests did not return correctly, and occasionally the router lost connection with its workers during execution.

To address this, a future PR will introduce a simple, built-in router as a replacement for the current SGLang router.

wuxibin89:
Does sglang `wake_up/sleep` works in colocated mode? I observed that sglang seems resume with random weights in colocated mode.
https://github.com/volcengine/verl/blob/main/verl/workers/rollout/sglang_rollout/async_sglang_server.py#L183-L188

yyDing1:
> Does sglang `wake_up/sleep` works in colocated mode? I observed that sglang seems resume with random weights in colocated mode. https://github.com/volcengine/verl/blob/main/verl/workers/rollout/sglang_rollout/async_sglang_server.py#L183-L188

Yes, I checked the related issues and found that the same phenomenon was mentioned in https://github.com/sgl-project/sglang/issues/6367#issuecomment-2915373479. In short, since RL normally uploads a new set of parameters, sglang simply discards the old ones to speed up. I also looked into the recent PR https://github.com/sgl-project/sglang/pull/10873, which seems to add support for reusing the original weights by keeping a stored copy.

yyDing1:
The PR seems to be included in sglang 0.5.3.
So we may pass `enable_weights_cpu_backup = True` when launching the sglang servers (for reward models) to resolve this issue.

.github/workflows/reward_model.yml
```diff
@@ -101,18 +101,15 @@ jobs:
...
-          unset http_proxy https_proxy HTTP_PROXY HTTPS_PROXY
-          pytest -s -x tests/workers/reward_model/test_discriminative_reward_model.py
-      - name: Running generative reward model tests on 8 L20 GPUs
+          ray stop --force
+          python3 examples/data_preprocess/gsm8k.py --local_dataset_path ${HOME}/models/hf_data/gsm8k
+      - name: Running reward model tests on 8 L20 GPUs
         run: |
           unset http_proxy https_proxy HTTP_PROXY HTTPS_PROXY
-          pytest -s -x tests/workers/reward_model/test_generative_reward_model.py
+          pytest -s -x tests/experimental/reward/test_reward_model.py
```
wuxibin89: It seems like `tests/experimental/reward/test_agent_loop_reward_manager.py` is not included for e2e test?

docs/advance/reward_loop.rst
```diff
@@ -0,0 +1,15 @@
+Reward Loop
+===========
+
+Last updated: 10/10/2025.
+
+.. warning::
+   Reward Loop is still in progress.
```
wuxibin89, line 7: Will the design doc be supplied in next PR?

verl/experimental/agent_loop/agent_loop.py
```diff
@@ -406,12 +283,14 @@ def __init__(
                 self.processor.chat_template = self.config.actor_rollout_ref.model.custom_chat_template
             self.tokenizer.chat_template = self.config.actor_rollout_ref.model.custom_chat_template
 
+        from verl.experimental.reward import RewardManagerWorker
```
wuxibin89: Why move import to local here?

verl/experimental/reward/reward_model.py
```diff
@@ -0,0 +1,136 @@
...
+        self.server_handles = [server._server_handle for server in self.rollout_replicas]
+        self.server_addresses = [server._server_address for server in self.rollout_replicas]
+
+    def _initialize_router(self):
+        worker_urls = [f"http://{server_address}" for server_address in self.server_addresses]
+
+        if self.config.rollout.name == "sglang":
+            from .router.sglang_router import launch_router_process
+        else:
+            from .router.naive_router import launch_router_process
```
wuxibin89, line 90: Do we also test vllm?

.github/workflows/reward_model.yml
yyDing1: this ci test has been added to workflow file

docs/advance/reward_loop.rst
yyDing1: Yes, future pr will add the design doc and more guidance for users.

verl/experimental/agent_loop/agent_loop.py
yyDing1: fixed

verl/experimental/reward/reward_model.py
yyDing1: the ci test for vllm has been added

wuxibin89 (approved):

SupreCyk:
请问一下，现在我用最新的main分支代码训练，默认使用这个路径下的reward manager么，我一开始在脚本设置reward_model.reward_manager=batch会报错，然后我在verl/verl/experimental/reward/reward_loop注册了batch，就没有报错了

yyDing1:
Yes, batch reward manager is not supported in reward loop (as reward loop provides another way to handle batch requests). If you want to use the legacy, you can add `reward_model.use_reward_loop=False` in your training script to use the reward manager in `verl/workers/reward_manager`

SupreCyk:
> 是的，奖励循环不支持批量奖励管理器（因为奖励循环提供了另一种处理批量请求的方式）。如果您想使用旧版奖励管理器，可以在训练脚本中添加 `reward_model.use_reward_loop=False` ，以使用 `verl/workers/reward_manager` 中的奖励管理器。

如果我是想当policymodel rollout之后通过调用llm api来输出reward 分数，这种使用哪种比较好呢？

yyDing1:
see https://github.com/volcengine/verl/issues/4318#issuecomment-3645075565

PR #3716, opened 2025-10-10
[rollout] feat: add default agent name for agent loop

wuxibin89:
### What does this PR do?

Add `default_agent_loop` config if `agent_name` is absent in RLDataset.


verl/workers/rollout/vllm_rollout/vllm_rollout_spmd.py
```diff
@@ -445,7 +445,7 @@ async def update_weights(self, weights: Generator[tuple[str, torch.Tensor], None
                 lora_int_id=lora_int_id,
                 lora_path="simon_lora_path",
                 peft_config=asdict(peft_config),
-                lora_tensors=weights,
+                lora_tensors=dict(weights),
```
wuxibin89, line 448: Fix bug introduced in #3712

PR #3797, opened 2025-10-17
Completed the missing code for Rolloutconfig

FangdongWang:
https://github.com/volcengine/verl/blob/main/verl/workers/config/rollout.py#L177

ray.exceptions.RayTaskError(InstantiationException): ray::WorkerDict.actor_rollout_init_model() (pid=3071998, ip=spencer-cut-capture-h53-fangdong-wang-master-0, actor_id=b10d875d543a1d0a5eda240801000000, repr=<verl.single_controller.ray.base.WorkerDict object at 0x7f5b81623d00>)
TypeError: RolloutConfig.__init__() got an unexpected keyword argument 'limit_videos'

I think you have lost a line of code

### DIFF
--- a/verl/workers/config/rollout.py
+++ b/verl/workers/config/rollout.py
@@ -176,6 +176,8 @@ class RolloutConfig(BaseConfig):
 
     limit_images: Optional[int] = None
 
+    limit_videos: Optional[int] = None
+
     skip_tokenizer_init: bool = False
 
     def __post_init__(self):


verl/workers/config/rollout.py
```diff
@@ -176,6 +176,8 @@ class RolloutConfig(BaseConfig):
 
     limit_images: Optional[int] = None
 
+    limit_videos: Optional[int] = None
```
wuxibin89, line 179: Where is limit_videos used?

FangdongWang: Sorry, this is a plugin I added for my personal training Internvl.I made a mistake when updating the version.

PR #3846, opened 2025-10-21
[megatron] feat: support async training with megatron and VLM

ISEEKYAN:
### What does this PR do?

add support of megatron and VLM on @ArronHZG's async recipe

### API and Usage Example
use `geo3k_qwen25vl_7b_megatron_4_4.sh` for example

### Design & Code Changes
1. support megatron with megatron_worker
2. change the agent_loop base and async version with correct multi modal token processes.

### DIFF
--- a/recipe/fully_async_policy/megatron_worker.py
+++ b/recipe/fully_async_policy/megatron_worker.py
@@ -26,6 +26,7 @@ from verl.utils.device import (
     get_device_name,
     get_torch_device,
 )
+from verl.utils.megatron_utils import per_tensor_generator
 from verl.workers.megatron_workers import ActorRolloutRefWorker, AsyncActorRolloutRefWorker, CriticWorker
 
 logger = logging.getLogger(__file__)
@@ -92,8 +93,17 @@ class DetachNcclSync(AsyncActorRolloutRefWorker):
 class DetachActorWorker(DetachNcclSync):
     def _get_actor_params_generator(self):
         assert self._is_actor
-        assert self.bridge is not None
-        generator = self.bridge.export_weights(self.actor.actor_module)
+        if self.bridge is not None:
+            generator = self.bridge.export_weights(self.actor.actor_module)
+        else:
+            generator = per_tensor_generator(
+                self.actor.actor_module,
+                self.actor_model_config,
+                self.weight_converter,
+                self.tf_config,
+                self.layer_name_mapping,
+            )
+
         return generator
 
     @register(dispatch_mode=Dispatch.ONE_TO_ALL)


ArronHZG:
Hi,
Can you help supplement the unit tests for megatron with reference to this part of the code?

https://github.com/volcengine/verl/blob/main/.github/workflows/e2e_one_step_off_policy.yml#L135-L161

in run_fully_async_policy.sh
actor_rollout_ref.model.enable_gradient_checkpointing is not exist in megatron train


And I get an error like this:


```
  File "/home/hadoop-ai-search/Projects/ISEEKYAN-verl/verl/workers/megatron_workers.py", line 440, in _build_rollout
    self.weight_converter = get_mcore_weight_converter(self.actor_model_config, self.dtype)
                                                       ^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'DetachAsyncRolloutWorker' object has no attribute 'actor_model_config'. Did you mean: '_return_value'?
```

ISEEKYAN:
@ArronHZG unittest added. The `actor_rollout_ref.model.enable_gradient_checkpointing` is not a common option but fsdp exclusive.

verl/experimental/agent_loop/agent_loop.py
```diff
@@ -580,10 +580,22 @@ async def _run_agent_loop(
...
+                current_text = self.tokenizer.decode(output.prompt_ids, skip_special_tokens=False)
+                multi_modal_inputs = self.processor(
+                    text=[current_text],
+                    images=images,
+                    return_tensors="pt",
+                    max_length=self.config.actor_rollout_ref.rollout.prompt_length,
+                    padding="max_length",
+                    padding_side="left",
+                )
+                prompt_output["input_ids"] = multi_modal_inputs["input_ids"]
```
wuxibin89: This is not right, we should not decode token_ids to text and encode them back to token_ids again, because the encode-decode is irreversible. 

https://verl.readthedocs.io/en/latest/advance/agent_loop.html#chat-completion-vs-token-in-token-out

verl/workers/rollout/vllm_rollout/vllm_async_server.py
```diff
@@ -93,6 +93,7 @@ def collective_rpc(
         timeout: Optional[float] = None,
         args: tuple = (),
         kwargs: Optional[dict[str, Any]] = None,
+        **kwargs_extra: Any,
```
wuxibin89, line 98: Is this for vllm==0.11.0?

ISEEKYAN: yes, vllm0.11 introduces some new args that we need to catch

verl/experimental/agent_loop/agent_loop.py
ISEEKYAN: But we need to add placeholders of vision tokens with the image size throught `self.processor`, otherwise the input_ids will have only one vision token as placeholder, which is not correct for training engine/reward model

wuxibin89: I think AgentLoopOut prompt_ids and response_ids already have vision token placeholder?
- prompt: https://github.com/volcengine/verl/blob/main/verl/experimental/agent_loop/tool_agent_loop.py#L196
- tool response: https://github.com/volcengine/verl/blob/main/verl/experimental/agent_loop/tool_agent_loop.py#L344

wuxibin89: I have verified that ToolAgentLoop generate vision token correctly, but these two agent loop not call processor for vlm input.
- SingleAgentLoop: verl/experimental/agent_loop/single_turn_agent_loop.py
- PartialSingleTurnAgentLoop: recipe/fully_async_policy/agent_loop/partial_single_turn_agent_loop.py

```bash
ROLLOUT_NAME=vllm pytest -svvv tests/experimental/agent_loop/test_multi_modal.py
```
image

wuxibin89: Could you add vllm support for SingleAgentLoop/PartialSingleTurnAgentLoop, and add a unit test for it?

ISEEKYAN: I have developed the supports and added the tests.

ISEEKYAN:
@wuxibin89  ready for merge

wuxibin89 (approved):

PR #3935, opened 2025-10-28
[rollout, sglang] fix: correct input length check in sglang_rollout

triston-lee:
### What does this PR do?

Fixes incorrect input length check in `sglang_rollout.py`.
Previously the code used `len(_req.input_ids)` which only returned the first dimension of the tensor, so the check never triggered correctly.

This PR computes the true sequence length using `_req.input_ids.size(-1)` and compares it to `max_model_len`.

_No corresponding issue; found during local testing._


zhaochenyang20:
Good catch

PR #3980, opened 2025-10-31
[trainer] feat: Self-Normalized Importance Sampling

EduardDurech:
Self-Normalized Importance Sampling for rollout:backwards mismatch, adds `algorithm.rollout_is_self_norm`

SNIS applied to `rollout_is_weights`
  • `geo_mean`: per-sequence geometric mean
  • `seq-mean-token-mean` / `seq-mean-token-sum`: per-sequence masked mean/sum
  • `token-mean`, `seq-mean-token-sum-norm`: global denominator

Given $w_i=\dfrac{p(x_i)}{q(x_i)}$,  the self-normalized estimator is

$$\widehat{\mu}_{\text{SNIS}}=\frac{\sum_{i=1}^{N} w_i\cdot f(x_i)}{\sum_{i=1}^{N} w_i}$$

```yaml
algorithm:
  rollout_is: true
  rollout_is_self_norm: true
```

Example
image

Experimental, only `geo_mean` has been properly tested, please test yourself, most of these are not standard SNIS

Sequence index $b$, token $t$, mask $m_{b t}\in{0,1}$, per-token IS weights $w_{b t}>0$

Per-sequence $`w'_{bt}=\tfrac{w_{bt}}{d_b}`$
- `geo_mean` $\quad d_b=\exp\Bigg(\frac{\sum_t m_{bt}\cdot \log w_{bt}}{\sum_t m_{bt}}\Bigg)$
- `seq-mean-token-mean` $\quad d_b=\frac{\sum_t m_{bt}\cdot w_{bt}}{\sum_t m_{bt}}$
- `seq-mean-token-sum` $\quad d_b=\sum_t m_{bt}\cdot w_{bt}$

Global $`w'_{bt}=\tfrac{w_{bt}}{d}`$
- `token_mean` $\quad d=\frac{\sum_{b,t} m_{bt}\cdot w_{bt}}{\sum_{b,t} m_{bt}}$
- `seq-mean-token-sum-norm` given $T$ token dimension length `weights_full.shape[-1]` $\quad d=\frac{\sum_{b,t} m_{bt}\cdot w_{bt}}{T}$

### DIFF
--- a/verl/trainer/ppo/rollout_corr_helper.py
+++ b/verl/trainer/ppo/rollout_corr_helper.py
@@ -405,15 +405,23 @@ def compute_rollout_correction_weights(
     # Apply batch normalization if requested
     if rollout_is_batch_normalize:
         # Compute mean based on aggregation level
+        mask_float = response_mask.to(dtype=rollout_is_weights.dtype)
         if rollout_is == "token":
             # Token-level: normalize over all token weights
-            weights_mean = verl_F.masked_mean(rollout_is_weights, response_mask)
+            if torch.distributed.is_available() and torch.distributed.is_initialized():
+                weights_mean = verl_F.distributed_masked_mean(rollout_is_weights, mask_float)
+            else:
+                weights_mean = verl_F.masked_mean(rollout_is_weights, response_mask)
         elif rollout_is == "sequence":
             # Sequence-level: normalize over sequence weights (one weight per sequence)
             # For each sequence, compute mean over valid tokens (they all have the same weight)
             # then average across sequences
-            seq_weights_mean = verl_F.masked_mean(rollout_is_weights, response_mask, axis=-1)  # (batch_size,)
-            weights_mean = seq_weights_mean.mean()
+            seq_weights = verl_F.masked_mean(rollout_is_weights, response_mask, axis=-1)  # (batch_size,)
+            seq_mask = (response_mask.sum(dim=-1) > 0).to(dtype=rollout_is_weights.dtype)
+            if torch.distributed.is_available() and torch.distributed.is_initialized():
+                weights_mean = verl_F.distributed_masked_mean(seq_weights, seq_mask)
+            else:
+                weights_mean = (seq_weights * seq_mask).sum() / seq_mask.sum().clamp_min(1e-8)
         else:
             raise ValueError(f"Unsupported rollout_is: {rollout_is}")
 


wuxibin89:
There's a break change in https://github.com/volcengine/verl/pull/3984

wuxibin89:
@szrlee Can you help review this PR?

EduardDurech:
In general this wraps any IS implementation and it doesn't seem much needs to change from the conflicting PR other than the config and checking that the proper parametres are sent to `dp_actor` from `ray_trainer`, if @szrlee can confirm the PR he has will be stable I can refactor, but I won't refactor a second time

verl/workers/actor/dp_actor.py
```diff
@@ -403,6 +405,45 @@ def update_policy(self, data: DataProto):
...
+                            loss_mode_batch = self.config.policy_loss.get("loss_mode", "vanilla")
+                            loss_agg_mode_batch = self.config.loss_agg_mode
+
+                            if loss_mode_batch == "geo_mean":
+                                global_snis_denom = None
+                            else:
+                                lam = "seq-mean-token-mean" if loss_mode_batch == "gspo" else loss_agg_mode_batch
+                                if lam == "token-mean":
+                                    if torch.distributed.is_available() and torch.distributed.is_initialized():
+                                        global_snis_denom = verl_F.distributed_masked_mean(weights_full, mask_float)
```
wuxibin89: Can we move `global_snis_denom` calculation to `verl/trainer/ppo/mismatch_helper.py`? Better to keep `update_policy` function concise.

EduardDurech: @wuxibin89 sure, when I get a response from szrlee can add that

szrlee: hi @EduardDurech , https://github.com/volcengine/verl/pull/3984 is now merged. 

Could you try to use the rollout correction interface for this PR?

Thank you!

szrlee: @EduardDurech We can move it into rollout_corr_helper and add the config under algorithm.rollout_correction.

EduardDurech: @szrlee will do, I was just away a few days but I'll refactor this and make it consistent

EduardDurech:
since `rollout_is_batch_normalize`, have to make sure it's normalzied across DP then it should be fine

EduardDurech:
@szrlee

szrlee:
> @szrlee

Thank you for the refactor. I will check later.

verl/trainer/ppo/rollout_corr_helper.py
```diff
@@ -405,15 +405,23 @@ def compute_rollout_correction_weights(
     # Apply batch normalization if requested
     if rollout_is_batch_normalize:
         # Compute mean based on aggregation level
+        mask_float = response_mask.to(dtype=rollout_is_weights.dtype)
         if rollout_is == "token":
             # Token-level: normalize over all token weights
-            weights_mean = verl_F.masked_mean(rollout_is_weights, response_mask)
+            if torch.distributed.is_available() and torch.distributed.is_initialized():
+                weights_mean = verl_F.distributed_masked_mean(rollout_is_weights, mask_float)
```
szrlee, line 412: why do we need distributed masked mean?

verl/trainer/ppo/rollout_corr_helper.py
```diff
@@ -405,15 +405,23 @@ def compute_rollout_correction_weights(
...
             # For each sequence, compute mean over valid tokens (they all have the same weight)
             # then average across sequences
-            seq_weights_mean = verl_F.masked_mean(rollout_is_weights, response_mask, axis=-1)  # (batch_size,)
-            weights_mean = seq_weights_mean.mean()
+            seq_weights = verl_F.masked_mean(rollout_is_weights, response_mask, axis=-1)  # (batch_size,)
+            seq_mask = (response_mask.sum(dim=-1) > 0).to(dtype=rollout_is_weights.dtype)
+            if torch.distributed.is_available() and torch.distributed.is_initialized():
+                weights_mean = verl_F.distributed_masked_mean(seq_weights, seq_mask)
+            else:
+                weights_mean = (seq_weights * seq_mask).sum() / seq_mask.sum().clamp_min(1e-8)
```
szrlee, line 424: do we need to multiply the batch size or total token size to cancel out the loss aggregation in the later stage?

verl/trainer/ppo/rollout_corr_helper.py
EduardDurech: Because as is it normalizes per-rank, if you do `torch.distributed` in that case each rank would get its own $\overline{w} _r$ which wouldn't be a global normalization, `distributed_masked_mean` makes all ranks use the same $\overline{w} _{global}$

verl/trainer/ppo/rollout_corr_helper.py
EduardDurech: Correct me if I'm wrong, but as I get it
```python
rollout_is_batch_normalize=True
...
weights_mean = {masked_mean,distributed_masked_mean}(rollout_is_weights, mask)
rollout_is_weights /= weights_mean
```
`agg_loss` then does `masked_mean` on `pg_losses`
$L = -\frac{1}{N}\sum_i \frac{w_i}{\overline{w}} h_i = -\frac{\sum_i w_i h_i}{\sum_j w_j}$
which is SNIS, if any factor is added that would just do what lr or `loss_scale_factor` do

verl/trainer/ppo/rollout_corr_helper.py
szrlee: great observation!

distributed_masked_mean works when the function is called on each actor.

if it is computed in the ray_trainer.py centralized, it is already global normalization.

verl/trainer/ppo/rollout_corr_helper.py
szrlee: you are absolutely correct. but 'N' depends on how we compute the loss aggregation.

szrlee:
@tongyx361  it looks good to me.

Issue #3597, opened 2025-09-24
megatron training crash when long term DAPO with qwen3-14b/qwen3-30b-a3b


ISEEKYAN:
### System Info

This is a progress track of fixing megatron precision problem.

When training DAPO for a long term, the megatron training crashed, with the response_length/mean and the acc curves went down sharply.
It takes `32 x H20 x 6 days` for qwen3-14b and `32 x H20 x 13 days` for qwen3-30b to reproduce

14b:

<img width="752" height="610" alt="Image" src="https://github.com/user-attachments/assets/0c7302bb-a66e-4c9c-8330-bacc789bfc0b" />

<img width="1280" height="510" alt="Image" src="https://github.com/user-attachments/assets/9cb55f10-647e-4e09-bca7-3a7f3b30438f" />

30b:

<img width="764" height="618" alt="Image" src="https://github.com/user-attachments/assets/d4ac9e7e-d590-4a06-b4ff-7857f66e55c7" />

<img width="1186" height="926" alt="Image" src="https://github.com/user-attachments/assets/c46c0dd9-e98d-4b8d-9fbe-7ed80b11c21e" />

### Information

- [ ] The official example scripts
- [x] My own modified scripts

### Tasks

- [x] An officially supported task in the `examples` folder (such as GLUE/SQuAD, ...)
- [ ] My own task or dataset (give details below)

### Reproduction

trained with docker://verlai/verl:app-verl0.5-transformers4.55.4-vllm0.10.0-mcore0.13.0-te2.2

```
#!/usr/bin/env bash
set -xeuo pipefail

SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
[ -f "${SCRIPT_DIR}/env.sh" ] && source "${SCRIPT_DIR}/env.sh"

project_name='DAPO-Qwen3-14B-Base'
exp_name='14B-8node-tp4-10k-no-dynamic-bsz'

adv_estimator=grpo

use_kl_in_reward=False
kl_coef=0.0
use_kl_loss=False
kl_loss_coef=0.0

clip_ratio_low=0.2
clip_ratio_high=0.28

max_prompt_length=$((1024 * 2))
max_response_length=$((1024 * 10))
enable_overlong_buffer=True
overlong_buffer_len=$((1024 * 4))
overlong_penalty_factor=1.0

loss_agg_mode="token-mean"

enable_filter_groups=True
filter_groups_metric=acc
max_num_gen_batches=10
[...]

PR #3984, opened 2025-11-02
[BREAKING][algo] feat: Rollout Correction for General Off-Policy Problems

szrlee:
## Summary

This PR introduces a comprehensive overhaul of the rollout correction system with typed configuration, mathematical documentation, and performance optimizations.

If you find the PR useful, please consider citing:

```bibtex
@online{liu-li-2025-rl-collapse,
  title = {When Speed Kills Stability: Demystifying {RL} Collapse from the Training-Inference Mismatch},
  author = {Liu, Jiacai and Li, Yingru and Fu, Yuqian and Wang, Jiawei and Liu, Qian and Shen, Yu},
  year = {2025},
  month = sep,
  url = {https://richardli.xyz/rl-collapse}
}
```

**⚠️ BREAKING CHANGE**: Removes backward compatibility. Users must migrate to typed config.

---

### 1. Typed Configuration with Presets

**Before (deprecated):**
```yaml
algorithm:
  rollout_is: true
  rollout_is_threshold: 2.0
  rollout_is_level: token
```

**After (Python - Recommended):**
```python
from verl.trainer.config.algorithm import RolloutCorrectionConfig

# Use validated presets
config.algorithm.rollout_correction = RolloutCorrectionConfig.token_is()
config.algorithm.rollout_correction = RolloutCorrectionConfig.seq_is_rs()
```

**After (YAML):**
```yaml
algorithm:
  rollout_correction:
    rollout_is: token
    rollout_is_threshold: 2.0
```

**10 validated presets:**
- `token_is()` / `token_tis()` - Per-token IS
- `seq_is()` - Sequence-level IS
- `seq_is_rs()` / `seq_mis()` - Sequence IS + rejection sampling
- `geo_rs()` / `geo_mis()` - Geometric RS + veto
- `ppo_is_bypass()` - Bypass mode (performance)
- `pure_is()` - Pure policy gradient (no PPO clipping)
- `disabled()` - Metrics only

Complete document can be find on [https://verl.readthedocs.io/en/latest/advance/rollout_corr.html](https://verl.readthedocs.io/en/latest/advance/rollout_corr.html).

### YAML Configuration (Advanced)

For advanced customization or YAML-based configs:

```yaml
algorithm:
  rollout_correction:
    rollout_is: token                      # IS weights: "token", "sequence", or null
    rollout_is_threshold: 2.0              # Upper threshold for IS weights
    rollout_rs: null                       # Rejection sampling: "token", "sequence", "geometric", or null
    rollout_rs_threshold: null             # RS upper threshold (required if rollout_rs is enabled)
    rollout_rs_threshold_lower: null       # RS lower threshold (auto-reciprocal if null)
    rollout_token_veto_threshold: null     # Per-token veto threshold (null = disabled)
    bypass_old_logprob_for_rollout: false  # Skip old_log_prob computation
    use_pure_rollout_correction: false     # Pure policy gradient with IS

# REQUIRED: Enable log prob calculation
actor_rollout_ref:
  rollout:
    calculate_log_probs: true
```

### 2. Mathematical Documentation

New comprehensive document: `docs/advance/rollout_corr_math.md` (585 lines)

**Theoretical foundation:**
- REINFORCE → PPO → Decoupled PPO progression
- Batch size invariance: Decoupling proximal policy from behavior policy
- Three-policy framework: π_rollout, π_old, π_θ

**Complete formulations for:**
- Off-policy REINFORCE (`pure_is`)
- Standard PPO and bypass mode
- Decoupled PPO (`token_is`, `seq_is`, `seq_is_rs`)
- Rejection sampling (`geo_rs`)

**Diagnostic metrics:**
- KL divergence (direct and K3 estimators)
- Perplexity and perplexity ratio
- χ² divergence (token and sequence level)

**Quality:**
- Objective technical descriptions
- All formulas mathematically verified
- Cross-document consistency validated

### 3. Training Modes

| Mode | Config | Policies | Speed | Description |
|------|--------|----------|-------|-------------|
| **Standard** | `bypass=false, pure=false` | 3 | Standard | Full decoupled PPO with batch size invariance |
| **Bypass** | `bypass=true, pure=false` | 2 | **Fast** | PPO clips against rollout (faster) |
| **Pure IS** | `bypass=true, pure=true` | 2 | **Fast** | Off-policy REINFORCE without clipping |

**Example:**
```python
# Bypass mode for performance
config = RolloutCorrectionConfig.ppo_is_bypass(threshold=2.0)

# Pure IS for research
config = RolloutCorrectionConfig.pure_is(threshold=2.0)
```

### 4. Chi-Squared Divergence Metrics

Quantify off-policy severity:

```python
rollout_corr/chi2_token    # E[ρ²] - 1
rollout_corr/chi2_seq      # E[(∏ρ)²] - 1
```

**Interpretation:**
- χ² = 0: Perfect on-policy
- χ² < 1: Low off-policiness, stable
- χ² ≥ 10: High off-policiness, need correction

**Cleanup:**
- Removed `mismatch_` prefix
- All metrics under `rollout_corr/` namespace

### 5. Bug Fix

**Critical fix:**
- `rollout_rs="token"` with `rollout_rs_threshold=None` silently failed
- Now raises `ValueError` with clear error message

---

### Example 1: Basic Token-level IS
```python
# Old (no longer works)
config.algorithm.rollout_is = True
config.algorithm.rollout_is_threshold = 2.0
config.algorithm.rollout_is_level = "token"

# New
config.algorithm.rollout_correction = RolloutCorrectionConfig.token_is(threshold=2.0)
```

### Example 2: Sequence IS + Rejection Sampling
```python
# Old (no longer works)
config.algorithm.rollout_is_level = "sequence"
config.algorithm.rollout_is_mode = "mask"

# New
config.algorithm.rollout_correction = RolloutCorrectionConfig.seq_is_rs(
    is_threshold=2.0,
    rs_threshold=2.0
)
```

### Example 3: Disable
```yaml
# Old
rollout_is: false

# New
rollout_correction: null
```

---

## References

Liu, Li, Fu, Wang, Liu, Shen (2025). *When Speed Kills Stability: Demystifying RL Collapse from the Training-Inference Mismatch*. [Blog](https://yingru.notion.site/When-Speed-Kills-Stability-271211a558b7808d8b12d403fd15edda)

### DIFF
--- a/docs/advance/rollout_corr_math.md
+++ b/docs/advance/rollout_corr_math.md
@@ -9,7 +9,7 @@
 
 This document provides the definitive mathematical formulations for rollout correction methods in `verl`, following the natural progression from **REINFORCE** to **PPO** to **Decoupled PPO**.
 
-The `verl` library implements **decoupled PPO** ([Hilton et al., 2021](https://arxiv.org/abs/2110.00641)) via a three-policy framework that decouples the proximal policy (for controlling policy update size) from the behavior policy (for off-policy correction). This decoupling achieves **batch size invariance** and enables efficient use of stale data, supporting scenarios where data collection differs from the training distribution.
+Rollout correction provides a unified framework to handle **general off-policy problems** in RL training - any scenario where the data collection distribution differs from the training distribution.
 
 **Applicable scenarios include:**
 - **Policy mismatch**: Different precision (FP8 vs FP16 vs BF16 vs FP32), different backends (vLLM vs SGLang vs FSDP vs Megatron)
@@ -253,6 +253,8 @@ These methods use two policies without importance sampling between behavior and 
 #### 3.2.1 Incorrect LLM-RL Implementation (PPO Without Rollout Correction)
 
 **Theory:** Naive LLM-RL implementation that incorrectly applies PPO by ignoring the actual rollout policy and assuming $\mu = \pi_{\text{old}}$.
+
+**Note:** This incorrect implementation pattern was identified in [Liu, Li, et al. (2025)](https://yingru.notion.site/When-Speed-Kills-Stability-Demystifying-RL-Collapse-from-the-Training-Inference-Mismatch-271211a558b7808d8b12d403fd15edda) as a key cause of training instability in LLM-RL systems, motivating the development of this rollout correction framework.
 
 **Loss Function:**
 
@@ -579,5 +581,5 @@ $$
 - **Schulman, J., Wolski, F., Dhariwal, P., Radford, A., & Klimov, O. (2017).** "Proximal policy optimization algorithms." *arXiv preprint arXiv:1707.06347.* https://arxiv.org/abs/1707.06347
 - **Hilton, J., Cobbe, K., & Schulman, J. (2021).** "Batch size-invariance for policy optimization." *arXiv preprint arXiv:2110.00641.* https://arxiv.org/abs/2110.00641
   - Introduced decoupled PPO: separating proximal policy (for controlling policy update size) from behavior policy (for off-policy correction) to achieve batch size invariance
-- **"When Speed Kills Stability: Demystifying RL Collapse from the Training-Inference Mismatch"**
+- **Liu, J., Li, Y., et al. (2025).** "When Speed Kills Stability: Demystifying RL Collapse from the Training-Inference Mismatch"
   - Blog post: https://yingru.notion.site/When-Speed-Kills-Stability-Demystifying-RL-Collapse-from-the-Training-Inference-Mismatch-271211a558b7808d8b12d403fd15edda
--- a/docs/advance/rollout_corr.md
+++ b/docs/advance/rollout_corr.md
@@ -27,8 +27,8 @@ Rollout Correction provides a unified framework to handle **general off-policy p
 **Common off-policy scenarios:**
 
 1. **Policy Mismatch** (Implementation Differences)
-   - Different precision: BFloat16 rollout vs FP32 training
-   - Different backends: vLLM rollout vs PyTorch/FSDP training
+   - Different precision: FP8 vs FP16 vs BF16 vs FP32
+   - Different backends: vLLM vs SGLang vs FSDP vs Megatron
    - Different implementations even with identical weights
 
 2. **Temporal Lag** (Model Staleness)
@@ -58,6 +58,8 @@ These off-policy gaps can cause training instability and policy collapse. Rollou
 Many LLM-RL implementations incorrectly apply PPO by **ignoring the actual rollout policy** π_rollout and assuming the training reference policy π_old is the behavior policy. This is mathematically incorrect when π_rollout ≠ π_old (which is typical in LLM-RL due to precision/backend differences between rollout and training).
 
 **This is not PPO's fault** - PPO itself is mathematically correct. The issue is the incorrect assumption that π_old = π_rollout in naive implementations.
+
+This critical implementation mistake that leads to RL training collapse was identified in the blog post ["When Speed Kills Stability: Demystifying RL Collapse from the Training-Inference Mismatch"](https://yingru.notion.site/When-Speed-Kills-Stability-Demystifying-RL-Collapse-from-the-Training-Inference-Mismatch-271211a558b7808d8b12d403fd15edda) and motivated the development of this rollout correction framework.
 
 **Mathematically correct approaches:**
 - **Decoupled mode**: Three policies (π_rollout, π_old, π_θ) with IS correction from π_rollout to π_old


szrlee:
Regarding the critical issue you identified: we've addressed it, but intentionally took a **different approach** than the suggested fallback. Here's our reasoning:

## Our Solution: Explicit Configuration (No Fallback)

We removed the fallback behavior entirely and updated the documentation to match. This follows Python's "explicit is better than implicit" principle.

**Why no fallback to `rollout_is_threshold`?**

1. **Prevents silent failures**: The original bug silently skipped rejection sampling when `rollout_rs_threshold=None`. The fallback suggestion would still allow implicit behavior that could hide misconfiguration.

2. **Clearer user intent**: IS and RS serve different purposes (variance reduction vs. outlier rejection). Forcing users to explicitly set thresholds makes their intent clear and prevents accidentally using the wrong threshold.

3. **Smart default exists**: The function signature has `rollout_rs_threshold: Optional[float] = 2.0`, which provides a sensible default when the parameter is omitted.

**Current Behavior:**

```python
# ✅ Works - uses default 2.0
compute_rollout_correction_and_rejection_mask(
    rollout_rs="token"  # rollout_rs_threshold defaults to 2.0
)

# ✅ Works - uses explicit value
compute_rollout_correction_and_rejection_mask(
    rollout_rs="token",
    rollout_rs_threshold=5.0
)

# ❌ Raises clear error - catches misconfiguration
compute_rollout_correction_and_rejection_mask(
    rollout_rs="token",
    rollout_rs_threshold=None  # ValueError with clear message
)
```

**What Changed (commit [ef7f8f57](https://github.com/szrlee/verl/commit/ef7f8f57966229127b2fd6f7b5a5848b60d6b69e)):**
- Raise `ValueError` when `rollout_rs` is enabled but `rollout_rs_threshold` is explicitly `None`
- Updated docs: `(uses rollout_is_threshold if null)` → `(required if rollout_rs is enabled)`
- All 11 tests passing with this explicit approach

The key distinction: we only raise an error when someone **explicitly passes `None`**, not when they omit the parameter (which uses the default). This catches configuration errors while maintaining ease of use.

Does this approach work for your use case, or do you see a scenario where the fallback would be necessary?

verl/trainer/config/algorithm.py
```diff
@@ -93,13 +95,6 @@ class AlgoConfig(BaseConfig):
...
-    rollout_is_threshold_lower: Optional[float] = None
-    rollout_is_level: str = "token"
-    rollout_is_mode: str = "truncate"
-    rollout_is_veto_threshold: Optional[float] = None
-    # Controls whether to apply IS weights to policy loss (only if rollout_is_threshold is set)
-    # True = apply weights to loss, False = compute metrics only (no weight application)
-    rollout_is: bool = False
+    # Rollout Correction: corrects distribution mismatch between rollout and training policies
+    # Set to None to disable, or provide a dict with nested parameters
+    rollout_correction: Optional[dict] = None
```
wuxibin89: Better add a typed class instead of dict?

ISEEKYAN:
@szrlee  The new APIs are very powerful and general, but I would say it a little complicated for a new user to get started. We cannot expect our users use the APIs after reading and understand all the related papers, so it will be very crucial to provide more verified combinations of the parameters and their alias, such as `TIS`, `MIS`. If I were a user with no background, I would search the key work `MIS` in the codebase, and I would be very confused with no direct code and docs regarding to it.

So I think it will be very helpful for users to add docs about:
1. combinations of TIS and MIS, and describe why their parameters are like that.
2. Add migrated combination of verified examples, such as https://github.com/volcengine/verl/issues/3597#issuecomment-3409085645. We have costed at lease several tens of thousand GPU hours to verify these workable combination.

Thanks again for the incredible contributions!

wuxibin89:
> @szrlee The new APIs are very powerful and general, but I would say it a little complicated for a new user to get started. We cannot expect our users use the APIs after reading and understand all the related papers, so it will be very crucial to provide more verified combinations of the parameters and their alias, such as `TIS`, `MIS`. If I were a user with no background, I would search the key work `MIS` in the codebase, and I would be very confused with no direct code and docs regarding to it.
> 
> So I think it will be very helpful for users to add docs about:
> 
> 1. combinations of TIS and MIS, and describe why their parameters are like that.
> 2. Add migrated combination of verified examples, such as [megatron training crash when long term DAPO with qwen3-14b/qwen3-30b-a3b #3597 (comment)](https://github.com/volcengine/verl/issues/3597#issuecomment-3409085645). We have costed at lease several tens of thousand GPU hours to verify these workable combination.
> 
> Thanks again for the incredible contributions!

Yes, we should add a typed config class, e.g `RolloutCorrectionConfig`, and set default value to best practice.

szrlee:
> > @szrlee The new APIs are very powerful and general, but I would say it a little complicated for a new user to get started. We cannot expect our users use the APIs after reading and understand all the related papers, so it will be very crucial to provide more verified combinations of the parameters and their alias, such as `TIS`, `MIS`. If I were a user with no background, I would search the key work `MIS` in the codebase, and I would be very confused with no direct code and docs regarding to it.
> > So I think it will be very helpful for users to add docs about:
> > 
> > 1. combinations of TIS and MIS, and describe why their parameters are like that.
> > 2. Add migrated combination of verified examples, such as [megatron training crash when long term DAPO with qwen3-14b/qwen3-30b-a3b #3597 (comment)](https://github.com/volcengine/verl/issues/3597#issuecomment-3409085645). We have costed at lease several tens of thousand GPU hours to verify these workable combination.
> > 
> > Thanks again for the incredible contributions!
> 
> Yes, we should add a typed config class, e.g `RolloutCorrectionConfig`, and set default value to best practice.

@ISEEKYAN @wuxibin89 this refactor (commit f8bd5586d1af293193fd66c4afbc3c602594ec40) replaces dict-based config
with typed RolloutCorrectionConfig class and factory methods.

docs/advance/rollout_corr_math.md
```diff
@@ -0,0 +1,573 @@
+# Mathematical Formulations of Rollout Correction Methods in `verl`
+
+**Author:** Yingru Li
+**Date:** 2025-11-03
```
wuxibin89: Use `Last updated` instead.

verl/trainer/config/algorithm.py
```diff
@@ -56,6 +56,264 @@ class FilterGroupsConfig(BaseConfig):
...
+            rollout_rs_threshold=rs_threshold,
+            rollout_rs_threshold_lower=rs_threshold_lower,
+            rollout_token_veto_threshold=veto_threshold,
+        )
+
+    @classmethod
+    def geo_mis(
+        cls,
+        rs_threshold: float = 1.001,
+        rs_threshold_lower: Optional[float] = None,
```
ISEEKYAN: I notice the example recipes in this file set `rs_threshold_lower` to `None`, is this recommended now? Is `None` theoretically better than some real number like 0.999 like before?

verl/trainer/config/ppo_trainer.yaml
```diff
@@ -113,27 +113,18 @@ algorithm:
...
-  # Per-token veto threshold for catastrophic outliers (null to disable)
-  rollout_is_veto_threshold: null
-
-  # Whether to apply IS weights to policy loss
-  # true = apply weights to loss, false = compute metrics only (no weight application)
-  # Useful for monitoring mismatch before enabling correction
-  rollout_is: false
+  # Rollout Correction: corrects distribution mismatch between rollout and training policies
+  # Configuration is nested under rollout_correction. Set to null to disable entirely.
+  rollout_correction: null
```
wuxibin89: Need `_target_` here for hydra instantiate

tongyx361:
There are some unexpected memory / dependency issues (e.g., [FSDP](https://github.com/volcengine/verl/actions/runs/19137900721/job/54694699781?pr=3984#step:5:1008 ), [vLLM](https://github.com/volcengine/verl/actions/runs/19137900674/job/54694698906?pr=3984#step:5:874)) in the CI, which should be unrelated to this PR. We plan to retry the CI first, while investigating the root causes.

verl/trainer/ppo/rollout_corr_helper.py
```diff
@@ -0,0 +1,888 @@
...
+        rollout_rs_threshold=upper_threshold,
+        rollout_rs_threshold_lower=lower_threshold,
+    )
+
+    # Track overall and sequence-level rejection rates
+    metrics["rollout_rs_masked_fraction"] = verl_F.masked_mean(1 - mask, response_mask).item()
+    if rollout_rs == "token":
+        # Token-level aggregation: sequence is rejected if any token is rejected
+        seq_has_masked: torch.Tensor = verl_F.masked_sum(1 - mask, response_mask, axis=-1) > 0
+        metrics["rollout_rs_seq_masked_fraction"] = seq_has_masked.float().mean().item()
```
ISEEKYAN, line 185: rollout_rs == "token", so it should be `rollout_rs_token_masked_fraction` instread of `rollout_rs_seq_masked_fraction`? maybe we can use the same metric name `rollout_rs_masked_fraction` for both modes?

verl/trainer/ppo/rollout_corr_helper.py
```diff
@@ -0,0 +1,888 @@
...
+        else:  # geometric
+            # Broadcast threshold checks to match token dimensions
+            exceeds_upper_expanded: torch.Tensor = exceeds_upper.expand_as(response_mask)
+            below_lower_expanded: torch.Tensor = below_lower.expand_as(response_mask)
+            metrics["rollout_rs_ratio_fraction_high"] = verl_F.masked_mean(
+                exceeds_upper_expanded.float(), response_mask
+            ).item()
+            metrics["rollout_rs_ratio_fraction_low"] = verl_F.masked_mean(
+                below_lower_expanded.float(), response_mask
+            ).item()
```
ISEEKYAN: why do geo and seq modes have difference methods of calculating `rollout_rs_ratio_fraction`? I thought they are all sequence level, with only difference on how to calculate mismatch(mean vs sum on logp).

ISEEKYAN (commented):
leave some questions

ISEEKYAN (approved):

lizipao:
Hello, I want to know why I get an error when using the following settings:
```
rollout_is=null
rollout_is_threshold=1.001
rollout_rs=geometric
rollout_rs_threshold=1.001
rollout_rs_threshold_lower=0.99
rollout_token_veto_threshold=1e-4
```

The error message is:
```
File "verl/verl/trainer/ppo/rollout_corr_helper.py", line 754, in compute_offpolicy_metrics
    assert response_mask.any(), "Expected at least one valid token in response_mask"
AssertionError: Expected at least one valid token in response_mask
```

After testing, I found that the all-zero mask is the `modified_response_mask` after `compute_rollout_rejection_mask`. Is this reasonable?

wanghongqu:
File "verl/verl/trainer/ppo/rollout_corr_helper.py", line 754, in compute_offpolicy_metrics
    assert response_mask.any(), "Expected at least one valid token in response_mask"
AssertionError: Expected at least one valid token in response_mask 

the same error，any one can help me ?????

ISEEKYAN:
> File "verl/verl/trainer/ppo/rollout_corr_helper.py", line 754, in compute_offpolicy_metrics assert response_mask.any(), "Expected at least one valid token in response_mask" AssertionError: Expected at least one valid token in response_mask
> 
> the same error，any one can help me ?????

I think this situation is rare and data related, maybe you can increase the batch size to lower the probability of all invalid. Or you should add some function to skip this training step and redo the rollout until you have at least one valid data for training.

PR #4010, opened 2025-11-04
[trainer] fix: Handle missing reward_fn_key in load_reward_manager

CzsGit:
### What does this PR do?
```shell
Error：omegaconf.errors.ConfigAttributeError: Key 'reward_fn_key' is not in struct
```
Fix ConfigAttributeError when `reward_fn_key` is not configured in the data section. The `load_reward_manager` function was incorrectly requiring `config.data.reward_fn_key` to be present, while all reward manager classes are designed to work with a default value of `"data_source"`. This PR adds safe access with proper default value handling to maintain backward compatibility.

Fixes the error:

```javascript
ConfigAttributeError: Key 'reward_fn_key' is not in struct
    full_key: data.reward_fn_key
```

### Checklist Before Starting

### Test

Tested by running PPO training without `reward_fn_key` configuration in the config file. Previously this would crash with `ConfigAttributeError`, now it works correctly and uses the default value `"data_source"`.

### API and Usage Example

No API changes. This fix allows users to run PPO training without needing to specify `reward_fn_key` configuration:

```python
# This config now works (previously crashed)
data:
  # reward_fn_key: not required anymore, defaults to "data_source"
```

### Design & Code Changes

__File Changed:__ `verl/trainer/ppo/reward.py` Line 164

__Before:__

```python
reward_fn_key=config.data.reward_fn_key,
```

__After:__

```python
reward_fn_key=getattr(config.data, "reward_fn_key", "data_source"),
```

This applies safe config access pattern and uses the same default value (`"data_source"`) that all reward manager classes use.

- [x] Apply pre-commit checks

### DIFF
--- a/verl/trainer/ppo/reward.py
+++ b/verl/trainer/ppo/reward.py
@@ -161,7 +161,7 @@ def load_reward_manager(
         tokenizer=tokenizer,
         num_examine=num_examine,
         compute_score=final_compute_score,
-        reward_fn_key=config.data.reward_fn_key,
+        reward_fn_key=getattr(config.data, "reward_fn_key", "data_source"),
         **reward_kwargs,
     )
 


CzsGit:
@eric-haibin-lin Hi maintainers, 

I've found and fixed another issue in multi-node training (agent loop config path resolution). I've submitted it as a separate PR since it's in a different file and addresses a different problem. 

Happy to discuss if you'd prefer these to be handled differently!

wuxibin89:
I think we do have default `reward_fn_key: data_source` in https://github.com/volcengine/verl/blob/main/verl/trainer/config/data/legacy_data.yaml#L29-L30

yyDing1:
Agree, this error doesn’t seem like it should occur. You may print the `config` at the beginning of the program to check whether `config.data.reward_fn_key` actually exists (default value `"data_source"`).

PR #4021, opened 2025-11-05
[megatron] fix: add loss_scale_factor for megatron backend 

ISEEKYAN:
What does this PR do?
There is a loss_scale_factor for fsdp to to count the loss weights of gradient accumulation.
But no loss_scale_factor found for megatron, this may cause the loss weights work wrong especially for dynamic batching when the sample length varies a lot. Might be the reason the megatron works worse than FSDP
Found by @vermouth1992

Reopen from https://github.com/volcengine/verl/pull/3294

### DIFF
--- a/verl/workers/actor/megatron_actor.py
+++ b/verl/workers/actor/megatron_actor.py
@@ -434,6 +434,9 @@ class MegatronPPOActor(BasePPOActor):
             response_length = responses.size(1)
             response_mask = data["response_mask"].to(bool)
             loss_agg_mode = self.config.loss_agg_mode
+            n_samples = response_mask.shape[0]
+            loss_scale_factor = n_samples / self.config.ppo_mini_batch_size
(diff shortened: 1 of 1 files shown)


Eisenhower:
@ISEEKYAN 
During training, I encountered an issue under a dynamic batching setup. With all other hyperparameters aligned, when training Qwen2.5-7B, the pg_loss, entropy, and validation metrics between FSDP and Megatron are largely consistent according to TensorBoard, and the downstream evaluation results are also very similar.

However, when training Qwen2.5-72B, the Megatron backend consistently collapses after a certain number of steps. This failure is inferred from both the TensorBoard curves and downstream evaluation performance degradation.

I have two questions:
	1.	When you mention that this issue can affect training quality, have you validated this impact through concrete experiments?
	2.	Given the issue I’m observing, do you have any suggestions or directions that might help with diagnosis or mitigation?

PR #4051, opened 2025-11-07
[trainer] fix: prevent ReactAgentLoop infinite recursion

### LINKED ISSUE #4049: [Bug] langgraph.errors.GraphRecursionError: Recursion limit of 25 reached without hitting a stop condition.
### System Info

__Hardware:__

- 2 servers, each with 2x NVIDIA A100 GPUs (total 4 GPUs)

__Software Environment:__

- Python: 3.10.0
- PyTorch: 2.7.1
- CUDA: 12.4 (based on flashinfer-python build)

__Key Dependencies:__

- vllm: 0.10.0
- flashinfer-python: 0.2.2.post1+cu124torch2.6
- torchaudio: 2.7.1
- torchvision: 0.22.1
- torchdata: 0.11.0
- langchain: 1.0.2
- langchain-core: 1.0.1

__Operating System:__

- Linux 5.14

### Information

- [x] My own modified scripts

### Tasks

### Reproduction

When performing distributed training using the script `recipe/langgraph_agent/example/run_qwen2.5_3b.sh`, the process may fail to terminate automatically due to the model repeatedly invoking tools in multiple loops. This issue occurs particularly when `max_assistant_turns` is not set, resulting in an occasional error: `langgraph.errors.GraphRecursionError: Recursion limit of 25 reached without hitting a stop condition.`

### Expected behavior

It should be that when the user has not set it, the model can also independently determine whether to stop or continue using the tool, rather than relying solely on the user’s setting of *max_assistant_turns*.

CzsGit:
## Problem
This PR fixes #4049 

ReactAgentLoop training crashes with `GraphRecursionError: Recursion limit of 25 reached without hitting a stop condition` when using Hydra config inheritance.

## Root Cause

**Two critical issues:**

1. **Missing default safety limit**: The `should_continue` function only enforced turn limits when `max_assistant_turns` was explicitly set. However, Hydra config inheritance often results in `null` values, leaving no protection against hitting LangGraph's hard recursion limit.

2. **No error recovery**: Any exception (recursion errors, tool failures, timeout errors) would crash the entire training job, making the system unreliable for production use.

### Fix 1: Add default safety limit 

**Before:**
```python
def should_continue(state: MessagesState, config: RunnableConfig):
    max_assistant_turns = config["configurable"]["max_assistant_turns"]
    # ...
    if max_assistant_turns and num_assistant_turns >= max_assistant_turns:
        return END
```
After:
```python
def should_continue(state: MessagesState, config: RunnableConfig):
    # Safely extract max_assistant_turns from config
    max_assistant_turns = None
    try:
        if config and "configurable" in config:
            max_assistant_turns = config["configurable"].get("max_assistant_turns")
    except Exception as e:
        logger.warning(f"Failed to extract max_assistant_turns from config: {e}")
    
    # ...
    
    # Use a reasonable default limit (25) if max_assistant_turns is not set
    # This prevents infinite loops
    effective_max_turns = max_assistant_turns if max_assistant_turns is not None else 25
    if num_assistant_turns >= effective_max_turns:
        return END
```
Impact:
Provides a sensible default (25 turns) when config is null
Prevents agents from running indefinitely and hitting hard limits
Safe config extraction with proper error handling
Still allows explicit configuration to override the default

### Fix 2: Graceful error handling 

Before:
```python
state = await self.graph.ainvoke(input={"messages": messages}, config=config)
```
Any exception crashes the entire training
After:
```python
try:
    state = await self.graph.ainvoke(input={"messages": messages}, config=config)
except Exception as e:
    # Handle GraphRecursionError and other errors gracefully
    # This prevents training from crashing when agent hits recursion limit
    logger.error(f"Agent loop execution failed: {type(e).__name__}: {e}")
    logger.error("Returning empty response to continue training.")
    
    # Create a minimal response to prevent training crash
    # The agent failed to complete, but we return an empty response
    # This allows training to continue and collect whatever partial data was generated
    empty_response = AIMessage(content="[Agent execution failed - recursion limit exceeded]")
    state = {"messages": messages + [empty_response]}
```

### DIFF
--- a/recipe/langgraph_agent/react_agent_loop.py
+++ b/recipe/langgraph_agent/react_agent_loop.py
@@ -76,7 +76,7 @@ def should_continue(state: MessagesState, config: RunnableConfig) -> Literal["to
         return END
 
     # no tool calls
-    if not getattr(last_message, 'tool_calls', None):
+    if not getattr(last_message, "tool_calls", None):
         return END
 
     return "tools"


CzsGit:
@wuxibin89 Could you please help review this? Thank you

wuxibin89:
@CzsGit Please format code: https://github.com/volcengine/verl/blob/main/CONTRIBUTING.md#code-linting-and-formatting

CzsGit:
> @CzsGit Please format code: https://github.com/volcengine/verl/blob/main/CONTRIBUTING.md#code-linting-and-formatting

@wuxibin89 Formatting issues fixed. PTAL.

wuxibin89 (approved):

PR #4101, opened 2025-11-12
[algo] feat: support router replay

litianjian:
### What does this PR do?

This PR introduces a draft **Router Replay** support into Verl.
Inspired by the recent research in **MoE Reinforcement Learning**([2510.11370](https://arxiv.org/abs/2510.11370), [2507.18071](https://arxiv.org/abs/2507.18071)), this implementation supports **Router Replay (R2)** and **Rollout Router Replay (R3)**. 
R2 allows recording routing token selection during` log probability computation` and replaying expert selection during policy update. R3 enables recording during `model inference` and replaying during RL post-training.

The initial version supports **Router Replay** with `Megatron` backend, including comprehensive support for distributed training strategies (**DP, TP, EP, ETP, PP, and Re-compute**).

The current implementation uses a patch-based approach. Once the upstream PR [NVIDIA/Megatron-LM#2101](https://github.com/NVIDIA/Megatron-LM/pull/2101) is merged or provides corresponding interfaces, the patch can be removed and replaced with official API integration.

### Basic Configuration
To enable Router Replay functionality, add the following configuration to your trainer config:
#### Method 1: Trainer Configuration
Add the following configuration to your trainer config:

```yaml
router_replay:
  enabled: true
  mode: "R2"  # Options: "R2", "R3"
```

#### Method 2: Launch Script Configuration
Add the following parameter to your launch script:

```bash
# In your launch script
actor_rollout_ref.actor.router_replay.mode="R2"
```

### R2 Mode Usage
1. **Enable R2 mode** in configuration
2. **Record phase**: During log probability computation, routing selections are automatically recorded
3. **Replay phase**: During policy update, recorded expert selections are replayed

### R3 Mode Usage
1. **Enable R3 mode** in configuration
2. **Record phase**: During model inference, routing decisions are captured
3. **Replay phase**: During RL post-training, recorded routing data is used
4. 
## In Progress
R2
- [ ]  FSDP backend

R3
- [x] vLLM Rollout
- [ ] Sglang Rollout

### DIFF
--- a/verl/workers/actor/megatron_actor.py
+++ b/verl/workers/actor/megatron_actor.py
@@ -570,7 +570,11 @@ class MegatronPPOActor(BasePPOActor):
             attention_mask = batch["attention_mask"].to(bool)
             position_ids = batch["position_ids"]
 
-            vp_rank = unwrap_model(model).vp_stage
+            unwrapped_model = unwrap_model(model)
+            if hasattr(unwrapped_model, "vp_stage"):
+                vp_rank = unwrapped_model.vp_stage
+            else:
+                vp_rank = 0
 
             multi_modal_inputs = {}
             if "multi_modal_inputs" in batch:
--- a/verl/utils/megatron/router_replay_utils.py
+++ b/verl/utils/megatron/router_replay_utils.py
@@ -28,7 +28,10 @@ from megatron.core.transformer.transformer_config import TransformerConfig
 from megatron.core.transformer.transformer_layer import get_transformer_layer_offset
 
 from verl.models.mcore.util import postprocess_packed_seqs, preprocess_packed_seqs
+from verl.utils.device import get_device_name
 from verl.utils.megatron.router_replay_patch import RouterReplay, RouterReplayAction
+
+device_name = get_device_name()
 
 
 # from megatron.core.transformer.transformer_block import get_num_layers_to_build
@@ -184,7 +187,7 @@ def merge_router_topk_indices(attention_mask, input_ids, mini_layer_topk_idx_lis
             layers_topk_idx.append(router.recorded_topk_idx.to(torch.uint8))  # dynamic_bs, topk
 
         # layer_num, dynamic_bs, topk  -> dynamic_bs, layer_num, topk
-        layers_topk_idx = torch.stack(layers_topk_idx).permute(1, 0, 2).cuda()
+        layers_topk_idx = torch.stack(layers_topk_idx).permute(1, 0, 2).to(device_name)
         # dynamic_bs, layer_num, topk -> 1, dynamic_bs_all, layer_num, topk
         layers_topk_idx = (
             gather_from_sequence_parallel_region(layers_topk_idx, tensor_parallel_output_grad=False)
@@ -225,7 +228,7 @@ def set_router_replay_data(layers_topk_idx, attention_mask, tf_config, vp_rank=N
 
         # 1, dynamic_bs_split, layer_num, topk
         layers_topk_idx_rmpad_split = scatter_to_sequence_parallel_region(
-            layers_topk_idx_rmpad.cuda().squeeze(dim=0)
+            layers_topk_idx_rmpad.to(device_name).squeeze(dim=0)
         ).unsqueeze(dim=0)
 
         # dynamic_bs_split, layer_num, topk -> layer_num, dynamic_bs_split, topk
@@ -327,7 +330,7 @@ def pp_gather(local_layers_router_map, tf_config):
 
     pp_group = mpu.get_pipeline_model_parallel_group()
     world_size = torch.distributed.get_world_size(pp_group)
-    local_layers_router_map = local_layers_router_map.cuda()
+    local_layers_router_map = local_layers_router_map.to(device_name)
     layers_topk_idx_global_list = [
         torch.empty(
             size=local_layers_router_map.shape,


Cesilina:
你好，想问下，在megatron_workers中的compute_log_prob中，R2模式下，此处 if self.enable_routing_replay and self.config.actor.router_replay.mode == "R2":
            RouterReplay.set_global_routing_mode(RoutingMode.RECORD)
            是不是只设置了record mode，但是并没有set_replay_data呢，后续在megatron_actor的计算中，merge_router_topk_indices中获取的router_instances_list的对象的record_topk_idx全是none

ISEEKYAN:
please fix the conflicts so we can merge this

litianjian:
> 你好，想问下，在megatron_workers中的compute_log_prob中，R2模式下，此处 if self.enable_routing_replay and self.config.actor.router_replay.mode == "R2": RouterReplay.set_global_routing_mode(RoutingMode.RECORD) 是不是只设置了record mode，但是并没有set_replay_data呢，后续在megatron_actor的计算中，merge_router_topk_indices中获取的router_instances_list的对象的record_topk_idx全是none

In record mode, mcore records the router selection results in raw form. These selections will be used in the next update-policy stage

Cesilina:
> records the router selection results in raw for

enen ,Get it! Thanks

verl/utils/megatron/router_replay_patch.py
```diff
@@ -0,0 +1,341 @@
...
+from megatron.core.transformer.moe.router import TopKRouter
+from megatron.core.transformer.transformer_config import TransformerConfig
+
+
+class RoutingMode(Enum):
+    NONE = "none"
+    RECORD = "record"
+    REPLAY_FORWARD = "replay_forward"
+    REPLAY_BACKWARD = "replay_backward"
+    FALLTHROUGH = "fallthrough"
```
ISEEKYAN: we don't need a fallthrough mode

verl/utils/megatron/router_replay_patch.py
```diff
@@ -0,0 +1,341 @@
...
+
+    def compute_topk(scores, topk, num_groups=None, group_topk=None):
+        # Default behavior if no replay is active
+
+        routing_action = router_replay.routing_mode.value if router_replay is not None else None
+
+        if routing_action is None or routing_action == "fallthrough":
+            return _compute_topk(scores, topk, num_groups=num_groups, group_topk=group_topk)
+
+        if routing_action == "record":
```
ISEEKYAN: it is better to use enum `RoutingMode.RECORD` instead of raw string "record". Same to the other modes.

verl/workers/actor/megatron_actor.py
```diff
@@ -537,6 +570,10 @@ def forward_step(batch_iter, model, return_schedule_plan: bool = False):
             attention_mask = batch["attention_mask"].to(bool)
             position_ids = batch["position_ids"]
 
+            # model: DistributedDataParallel(
+            # (module): Float16Module(
+            # (module): GPTModel(
+            vp_rank = model.module.module.vp_stage
```
ISEEKYAN: better to use `unwrap_model(model).vp_stage`

verl/utils/megatron/router_replay_patch.py
litianjian: Got it.  BTW, the patch can be removed if actor backend provides the similar apis.

verl/workers/actor/megatron_actor.py
litianjian: Appreciate it! Learned a lot

scut-zx:
I would like to inquire about the latest progress of this project. Does R3 support the training of Megatron+vLLM? Does Megatron need to use the PR version you submitted: https://github.com/NVIDIA/Megatron-LM/pull/2101/files?

ISEEKYAN:
> I would like to inquire about the latest progress of this project. Does R3 support the training of Megatron+vLLM? Does Megatron need to use the PR version you submitted: https://github.com/NVIDIA/Megatron-LM/pull/2101/files?

it is ready for merge now. Now you don't need https://github.com/NVIDIA/Megatron-LM/pull/2101 to achieve router replay in verl. But once megatron's PR is merged, we can remove some patches from verl.

Shadowyuan616:
Hi, is `deepep` necessary for routing replay (R2/R3)? 
Do you have any methods to run the script on other kinds of devices? Deepep seems to be only compatible with SM90 device for nvshmem and full features.

litianjian:
> Hi, is `deepep` necessary for routing replay (R2/R3)? Do you have any methods to run the script on other kinds of devices? Deepep seems to be only compatible with SM90 device for nvshmem and full features.

Hi, `deepep` is not necessary for routing replay.

SuffixAutomata:
@litianjian Hello. I have a question about [https://github.com/volcengine/verl/commit/cb236075dbf1f9b89660d5e2f28e30f3268ec7ee#diff-25157fd1cdb544dfb3245f630bfbd49cff24e79eac8ef893ce440217eb92b385](https://github.com/volcengine/verl/commit/cb236075dbf1f9b89660d5e2f28e30f3268ec7ee#diff-25157fd1cdb544dfb3245f630bfbd49cff24e79eac8ef893ce440217eb92b385) - why are we assigning .routed_experts to AgentData? As far as I can tell AgentData does not have this field, nor is this field used downstream.

litianjian:
> @litianjian Hello. I have a question about [cb23607#diff-25157fd1cdb544dfb3245f630bfbd49cff24e79eac8ef893ce440217eb92b385](https://github.com/volcengine/verl/commit/cb236075dbf1f9b89660d5e2f28e30f3268ec7ee#diff-25157fd1cdb544dfb3245f630bfbd49cff24e79eac8ef893ce440217eb92b385) - why are we assigning .routed_experts to AgentData? As far as I can tell AgentData does not have this field, nor is this field used downstream.

You’re right. This was my mistake—I need to add support for agentdata.

ForeverDJ-ux:
I test router replay with run_qwen30_a3b_megatron_vllm.sh R3 mode, if I  I use 2 node  VLLM tp=8, the entropy loss is  large, while I use 1node with TP=8，or 2node with TP=16, the entropy loss looks like normal
2 node TP=8 result:
image
image

1 node TP=8 result:
image
2 node TP=16 result:
image

my script is:
image
image
image
image
image

Anybody knows how to fix this error?
@litianjian

PR #4168, opened 2025-11-17
[recipe, rollout] fix: Make one step off policy support more card combination

moehanabi:
Because `ActorRolloutRefWorker` will call `initialize_model_parallel` when init, any worker of subclass of `ActorRolloutRefWorker` needs to meet the parallel configuration of the actor during initialization, which leads to the one-step-off-policy rollout worker also having to meet the corresponding card number to actor worker.

So if you have 32 cards for actor with PP=4 and EP=8, and 16 cards for rollout with TP=4, you will meet `RuntimeError: decoder world_size (16) is not divisible by expert_tensor_model_pipeline_parallel size (32)` in megatron.

This patch deleted the megatron related codes for rollout worker.

### What does this PR do?

### Checklist Before Starting

### Test

### API and Usage Example

```python
```

### Design & Code Changes

> [!IMPORTANT]

### DIFF
--- a/recipe/one_step_off_policy/megatron_workers.py
+++ b/recipe/one_step_off_policy/megatron_workers.py
@@ -13,19 +13,29 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import datetime
 import logging
 import os
 
 import torch
 import torch.distributed
-from omegaconf import DictConfig, OmegaConf
+from omegaconf import DictConfig
 
+from verl.single_controller.base import Worker
 from verl.single_controller.base.decorator import Dispatch, make_nd_compute_dataproto_dispatch_fn, register
 from verl.utils.config import omega_conf_to_dataclass
-from verl.utils.debug import (
+from verl.utils.device import (
+    get_device_name,
+    get_nccl_backend,
+    get_torch_device,
+)
+from verl.utils.distributed import set_numa_affinity
+from verl.utils.profiler import (
+    DistProfiler,
+    DistProfilerExtension,
+    ProfilerConfig,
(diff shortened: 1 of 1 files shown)


verl/workers/megatron_workers.py
```diff
@@ -191,6 +191,13 @@ def __init__(self, config: DictConfig, role: str, **kwargs):
             # NPU MindSpeed patch, will be refactored with MindSpeedEngine.
             repatch(self.config.actor.megatron.get("override_transformer_config", {}))
 
+        self.role = role
```
ArronHZG: We need to conduct some tests to eliminate the possible impacts caused by this change.

ArronHZG (commented):
can you just fix one step off policy ?
I think this change may effect other recipe function.

moehanabi:
> can you just fix one step off policy ? I think this change may effect other recipe function.

Yes I am willing to do it later.

But one problem is the more we change in recipes, the more diff / complexity / redundancy will be introduced. So in my opinion, if a fix can be introduced into main code, we can just add more tests to make it possible and ensure its safety?

As for this change, I think the ActorRolloutRefWorker will only be used by actor, rollout and ref, so this commit will only influence on all independent rollout, which is not an actor and ref.

moehanabi:
> can you just fix one step off policy ? I think this change may effect other recipe function.

Hi, I have committed a new change as your request. Please help to check again.

moehanabi:
> can you just fix one step off policy ? I think this change may effect other recipe function.

okay i found #4179 has done the same fix, with verl's megatron_worker.py. I will close the pr.

PR #3522, opened 2025-09-18
[Megatron] feat: 1f1b overlap/moe_a2a_overlap

ISEEKYAN:
### What does this PR do?

Enable megatron-core's 1f1b overlap feature, which overlap the computation and MoE EP all2all communication, enabling higher MFU when training Large MoE models. see https://docs.nvidia.com/megatron-core/developer-guide/latest/api-guide/moe.html#batch-level-ep-a2a-hidding for official docs

### Checklist Before Starting

- [x] Search for similar PRs. Paste at least one query link here: ...
- [x] Format the PR title as `[{modules}] {type}: {description}` (This will be checked by the CI)
  - `{modules}` include `fsdp`, `megatron`, `sglang`, `vllm`, `rollout`, `trainer`, `ci`, `training_utils`, `recipe`, `hardware`, `deployment`, `ray`, `worker`, `single_controller`, `misc`, `perf`, `model`, `algo`, `env`, `tool`, `ckpt`, `doc`, `data`
  - If this PR involves multiple modules, separate them with `,` like `[megatron, fsdp, doc]`
  - `{type}` is in `feat`, `fix`, `refactor`, `chore`, `test`
  - If this PR breaks any API (CLI arguments, config, function signature, etc.), add `[BREAKING]` to the beginning of the title.
  - Example: `[BREAKING][fsdp, megatron] feat: dynamic batching`

### Test
I tested with megatron-core 0.15.0rc5 and TransformerEngine 2.3, the script is 
`pip install --no-deps --no-cache-dir git+https://github.com/NVIDIA/Megatron-LM.git@core_v0.15.0rc5`

```bash
set -x

export CUDA_DEVICE_MAX_CONNECTIONS=32
export NVTE_FWD_LAYERNORM_SM_MARGIN=20
export NVTE_BWD_LAYERNORM_SM_MARGIN=20

LLM=/models/Moonlight-16B-A3B-Instruct

gsm8k_train_path=$HOME/data/gsm8k/train.parquet
gsm8k_test_path=$HOME/data/gsm8k/test.parquet
math_train_path=$HOME/data/math/train.parquet
math_test_path=$HOME/data/math/test.parquet

train_files="['$gsm8k_train_path', '$math_train_path']"
test_files="['$gsm8k_test_path', '$math_test_path']"
train_files=$gsm8k_train_path
test_files=$gsm8k_test_path

DIST_CKPT_PATH=/mcore_ckpt/Moonlight-16B-A3B-Instruct

[...]

PR #4174, opened 2025-11-18
[megatron] fix: correct weight passing in fused forward

tangyc314:
### What does this PR do?

This PR addresses the issue encountered when setting `use fuse kernel` to `true` during `qwen3-0.6` training in `megatron`. The specific error is as follows:
```code
  File "model_forward_fused.py", line 201, in _fused_GPTModel_forward
    logprobs, entropy = linear_cross_entropy(
                         ^^^^^^^^^^^^^^^^^^^^^
  File "function.py", line 576, in apply
    return super().apply(*args, **kwargs)  # type: ignore[misc]
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "linear_cross_entropy.py", line 77, in forward
    logprobs, entropy, _maximum, _accumulate, _entropy_b = kernels.efficient_entropy_forward(
                                                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "kernels.py", line 518, in efficient_entropy_forward
    assert hidden.is_cuda and weight.is_cuda and labels.is_cuda
                              ^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'is_cuda'

```

Upon investigation, the issue arises when the `weight` passed to `linear_cross_entropy` is `None` during the call to the `_fused_GPTModel_forward` function of the `GPTmodel`. This is because when `share_embeddings_and_output_weights` is set to `true`, the `weight` from `embeddings_layer` should be passed instead of the `output_layer`.

### Test

It has been experimentally verified that when `share_embeddings_and_output_weights` is set to `true`, the `forward` function correctly passes the `weight` from `embeddings_layer`, resolving the error.

### API and Usage Example

This PR does not involve any API changes; usage remains unchanged.

### Design & Code Changes

The `_fused_GPTModel_forward` function in `verl/models/mcore/model_forward_fused.py` has been modified to ensure the correct `weight` is passed to `linear_cross_entropy`.

### DIFF
--- a/verl/models/mcore/model_forward_fused.py
+++ b/verl/models/mcore/model_forward_fused.py
@@ -200,7 +200,9 @@ def _fused_GPTModel_forward(
         hidden_states = gather_from_sequence_parallel_region(hidden_states)
     logprobs, entropy = linear_cross_entropy(
         hidden_states,
-        model.output_layer.weight,
+        model.output_layer.weight
+        if not model.share_embeddings_and_output_weights
+        else model.shared_embedding_or_output_weight(),
         labels,
         temperature,
         "none",
--- a/verl/models/mcore/model_forward_fused.py
+++ b/verl/models/mcore/model_forward_fused.py
@@ -200,9 +200,11 @@ def _fused_GPTModel_forward(
         hidden_states = gather_from_sequence_parallel_region(hidden_states)
     logprobs, entropy = linear_cross_entropy(
         hidden_states,
-        model.output_layer.weight
-        if not model.share_embeddings_and_output_weights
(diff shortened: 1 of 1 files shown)


tangyc314:
In fact, I reviewed PR https://github.com/volcengine/verl/pull/3522, and it seems that the model_forward_1f1b_overlap function can only be used on MoE models. For MoE models like Moonlight-16B-A3B-Instruct and Qwen3-30B-A3B, where tie_word_embeddings is false, the condition share_embeddings_and_output_weights being true will not occur.

PR #4252, opened 2025-11-23
[trainer] fix: use unscaled logits for accurate pearson correlation metric

### LINKED ISSUE #4162: logits scaling affect rollout_actor_probs_pearson_corr metrics
### Information

- [x] My own modified scripts

### Tasks

### Reproduction

When using VLLM for rollout, the returned rollout_log_probs are computed logits that are not scaled by temperature. If scaling is applied on the training side, it will lead to an incorrect value for the metric "rollout_actor_probs_pearson_corr." Generally, a correlation coefficient closer to 1 is considered to indicate better alignment between training and inference.

Using FSDP also has similar issues.

```Python
# verl/workers/actor/megatron_actor.py
                def logits_processor(logits, label, label_mask):
                    assert logits.shape[:2] == label.shape[:2]
                    assert label.shape == label_mask.shape
                    logits.div_(temperature) #### this affect rollout_actor_probs_pearson_corr
                    ret = {}
                    if calculate_entropy:
                        logits_bak = logits.clone()
                        logger.warning_once(
                            "For memory-efficient computation, enable fused kernels via "
                            "`actor_rollout_ref.model.use_fused_kernels=True`. "
                            "The current `clone()` operation ensures correctness but increases memory usage."
                        )
                        entropy = vocab_parallel_entropy(logits)
                        ret["entropy"] = entropy
                    else:
                        logits_bak = logits
                    log_probs = vocab_parallel_log_probs_from_logits(logits_bak, label)
                    log_probs = log_probs.masked_fill(~label_mask, 0.0)
                    ret["log_probs"] = log_probs
                    return ret
```

### Expected behavior

Training does not perform temperature scaling on logits, which keeps it aligned with inference, making the calculated rollout_actor_probs_pearson_corr meaningful.

JobQiu:
## Summary

Fixes #4162 

This PR ensures the `rollout_actor_probs_pearson_corr` metric accurately reflects the correlation between rollout and actor probabilities by computing log probabilities from unscaled logits for metrics calculation.

## Problem

Currently, there's an inconsistency in log probability computation:
- **VLLM rollout**: Uses unscaled logits: `log_softmax(logits)`
- **Actor training**: Uses temperature-scaled logits: `log_softmax(logits/temperature)`

Since `log_softmax(logits) ≠ log_softmax(logits/T)` when temperature ≠ 1.0, the `rollout_actor_probs_pearson_corr` metric becomes inaccurate, misleading users about the actual alignment between rollout and actor behaviors.

## Solution

1. **Megatron Actor**: Added `log_probs_for_metrics` computation before temperature scaling in `logits_processor()` (verl/workers/actor/megatron_actor.py:574-602)

2. **FSDP Actor**: Modified `_forward_micro_batch()` to return unscaled log probabilities alongside the scaled version (verl/workers/actor/dp_actor.py)

3. **Metrics**: Updated `calculate_debug_metrics()` to prioritize `log_probs_for_metrics` when available, falling back to `old_log_probs` for backward compatibility (verl/utils/debug/metrics.py:86-93)

## Key Design Decisions

- **Backward Compatible**: The fix gracefully falls back to `old_log_probs` if `log_probs_for_metrics` is not available
- **Minimal Performance Impact**: Only computes unscaled log_probs when needed for metrics
- **Preserves Training Logic**: Temperature scaling still applies to training (`old_log_probs` unchanged)

## Testing

Added comprehensive unit tests in `tests/utils/test_issue_4162_fix.py`:

1. `test_temperature_scaling_causes_pearson_deviation` - Demonstrates the original problem
2. `test_unscaled_logprobs_achieve_perfect_correlation` - Verifies the fix achieves perfect correlation
3. `test_temperature_scaling_mathematical_inequality` - Proves mathematical difference between scaled/unscaled
4. `test_temperature_one_no_difference` - Sanity check for temperature=1.0
5. `test_realistic_training_scenario` - Full scenario with 256 samples

## Impact

- ✅ More accurate metrics for monitoring training quality
- ✅ Better visibility into rollout-actor alignment
- ✅ No breaking changes to existing training pipelines
- ✅ Backward compatible with existing code

### DIFF
--- a/tests/utils/test_issue_4162_fix.py
+++ b/tests/utils/test_issue_4162_fix.py
@@ -1,0 +1,194 @@
+#!/usr/bin/env python3
+"""
+Test for Issue #4162 Fix: Logits scaling affecting rollout_actor_probs_pearson_corr metrics
+
+This test verifies that temperature scaling is handled correctly when comparing
+rollout and actor log probabilities.
+"""
+
+import torch
+import torch.nn.functional as F
+import pytest
+
+
+class TestIssue4162Fix:
+    """Test suite for Issue #4162 fix."""
+
+    def test_temperature_scaling_causes_pearson_deviation(self):
+        """
+        Verify that temperature scaling causes Pearson correlation deviation.
+
+        This demonstrates the original problem: when rollout uses unscaled logits
+        and actor uses scaled logits, the Pearson correlation is not 1.0 even
+        when using the same logits.
+        """
+        # Simulate logits from a model
+        torch.manual_seed(42)
+        batch_size = 10
+        vocab_size = 50257
+        logits = torch.randn(batch_size, vocab_size)
+        labels = torch.randint(0, vocab_size, (batch_size,))
+        temperature = 2.0
+
+        # Rollout (unscaled) - simulating VLLM behavior
+        rollout_log_probs = F.log_softmax(logits, dim=-1)[range(batch_size), labels]
+
+        # Actor (scaled) - original behavior before fix
+        actor_log_probs_scaled = F.log_softmax(logits / temperature, dim=-1)[range(batch_size), labels]
+
+        # Calculate Pearson correlation
+        rollout_probs = torch.exp(rollout_log_probs)
+        actor_probs_scaled = torch.exp(actor_log_probs_scaled)
+
+        pearson_with_scaling = torch.corrcoef(torch.stack([rollout_probs, actor_probs_scaled]))[0, 1]
+
+        # The correlation should deviate from 1.0 when temperature != 1.0
+        print(f"Pearson correlation with temperature scaling: {pearson_with_scaling:.4f}")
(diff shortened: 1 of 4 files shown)


wuxibin89:
Maybe we should let rollout return logprobs after apply temperature and top_k/top_p? @vermouth1992 @szrlee 
https://github.com/vllm-project/vllm/blob/releases/v0.11.0/vllm/config/model.py#L186-L193

szrlee:
> Maybe we should let rollout return logprobs after apply temperature and top_k/top_p? @vermouth1992 @szrlee https://github.com/vllm-project/vllm/blob/releases/v0.11.0/vllm/config/model.py#L186-L193

@wuxibin89 yes, processed_logprobs may be a better choice.

wuxibin89:
@JobQiu Can you help contribute to change vLLM logprobs_mode to `processed_logprobs`? SGLang may also need to change.

PR #4323, opened 2025-11-27
[megatron] feat: support gpt-oss

ISEEKYAN:
for now (latest TE=2.10), gptoss's optimized attn kernel is not supported for thd format, so we use bshd format here.
when bshd format is used, we need to pad the input_ids to the longest sequence length
so we recommend to disable dynamic batch size and set micro batch size to 1 to avoid paddings
but it is ok to try with micro_batch_size>1

see `test_dapo_gptoss_20b_megatron.sh` for example.

image

The training crashes with mismatch, need further experiments with MIS/TIS or fp16


verl/trainer/config/_generated_ppo_megatron_trainer.yaml
```diff
@@ -54,6 +54,7 @@ actor_rollout_ref:
       override_mcore_model_config: {}
       use_mbridge: false
       vanilla_mbridge: true
+      thd_format: true
```
wuxibin89: We already have an option `use_remove_padding`, which is used by fsdp worker, but it seems like megatron doesn't respect it.

ISEEKYAN: `use_remove_padding` is not strictly equivalent to megatron's thd format, especially when refactored version is using nested tensor by default.

wuxibin89 (approved):

PR #4360, opened 2025-11-30
feat(actor): add loss_scale_factor for seq-mean-token-sum-norm mode

szrlee:
Add configurable loss_scale_factor to replace hardcoded divisor in seq-mean-token-sum-norm loss aggregation. Users can set a constant value to ensure consistent normalization throughout training.

- Add loss_scale_factor param to agg_loss() in core_algos.py
- Add loss_scale_factor field to ActorConfig
- Propagate via global_batch_info to policy loss functions
- Update all direct agg_loss calls in trainers and recipes
- Update YAML configs and DrGRPO documentation

### DIFF
--- a/docs/algo/grpo.md
+++ b/docs/algo/grpo.md
@@ -57,6 +57,7 @@ Instead of adding KL penalty in the reward, GRPO regularizes by directly adding 
 Configure the following to enable DrGRPO, with all other parameters the same as GRPO's:
 
 - `actor_rollout_ref.actor.loss_agg_mode`: "seq-mean-token-sum-norm", which turns off seq-dim averaging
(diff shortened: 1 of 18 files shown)


ISEEKYAN:
do you have any experiment results on the effect of setting loss_scale_factor?

tongyx361:
> do you have any experiment results on the effect of setting loss_scale_factor?

Scaling the loss/objective by a constant, such as the horizon, is a common practice in RL algorithms.
Besides, it can interact further with mechanisms such as gradient clipping and ADAM (variants).

Since this PR does not change the default behavior, but only provides better flexibility, I think it can be approved.

PR #4386, opened 2025-12-02
[megatron] feat: optimize the mbridge checkpoint saving speed

ISEEKYAN:
adopt mbridge's fast saving API https://github.com/ISEEKYAN/mbridge/blob/main/mbridge/core/bridge.py#L302-L303, which will reduce the saving time of large models a lot.

In my test, the deepseek 671B saving time is under 2 minutes


vermouth1992:
Do we need to update mbridge version?

wuxibin89 (approved):

ISEEKYAN:
> Do we need to update mbridge version?

This feature in mbridge exists since 2 months ago (but dpskv3 was not supported until yesterday). Recommend to use the latest main of mbridge.

PR #4435, opened 2025-12-05
[rollout, vllm] feat: support KV cache FP8

sharonyu-115:
### What does this PR do?

This PR adds support for FP8 KV cache quantization in vLLM rollout, providing significant memory savings during inference. The KV cache quantization can be used independently or combined with weight and activation FP8 quantization https://github.com/volcengine/verl/blob/main/docs/advance/fp8.md. 

**How it works:**
Leverage vllm to dynamically calculate qkv scales for FP8 KV cache quantization(requires vllm 0.11.2).​ vllm only performs the calculation in the first forward. For RL training, the calculation needs to be triggered every training step to adapt to the new model weights. ​After weights update of each training step, reset the `calculate_kv_scales` flag to True for all the attention layers.​ So for FP8 rollout, functionally we can have:

- FP8-linear W8A8
- KV cache FP8 (for storage only)
- FP8-linear W8A8 + KV cache FP8 (for storage and attention calculation)​

As this function requires vllm 0.11.2, the PR also includes some changes for v0.11.2 compatibility

**Experiments and results:** 

- Model: Qwen3-8B-Base
- Configuration​:

  - DAPO recipe. AIME24 online validation.​
  - 1*8*H100. veRL,  FP8 async rollout, vllm+FSDP​
  - Prompt batch size 32​
  - Rollout batch size: 32x3x16​
  - Train_batch_size & ppo_mini_batch_size 32​
  - Max response length 20K​
  - Token-level TIS, C=2​
  - vllm0.11.2+cuda12.9+DeepGEMM+CUDA Graph
 

- Accuracy: 
  - KV cache FP8 aligns well with BF16 on online validation. Full FP8 is also mostly aligned.

image

- Performance:
  - FP8 linear W8A8 vs BF16:  up to 20% speedup​
  - KV cache FP8 only vs BF16: ~38% rollout speedup.​
  - FP8 linear W8A8 + KV cache FP8 vs BF16 about ~44% speedup ​

image

### Checklist Before Starting

### Test

### API and Usage Example

```python
```

### Design & Code Changes

> [!IMPORTANT]


wuxibin89:
@Agoniii Can you help take a look?

verl/trainer/config/rollout/rollout.yaml
```diff
@@ -46,6 +46,19 @@ cudagraph_capture_sizes: null
 # Whether to free engine KVCache after generation.
 free_cache_engine: True
 
+# KV cache data type for attention layers
```
wuxibin89: Since we have more quantization config:
- quantization
- kv_cache_dtype
- calculate_kv_scales

It's better to group them together? For example
```yaml
quantization:
  weight_dtype: null # null or fp8
  kv_cache_dtype: null # null or fp8
```

verl/trainer/config/rollout/rollout.yaml
```diff
@@ -46,6 +46,19 @@ cudagraph_capture_sizes: null
...
+#   1. Alone (kv_cache_dtype=fp8, quantization=null) - KV cache FP8 only
+#   2. Together with weight quantization (kv_cache_dtype=fp8, quantization=fp8) - Maximum memory savings
+#   3. Not at all (kv_cache_dtype=null) - Default behavior
+kv_cache_dtype: null
+
+# Whether to dynamically calculate KV scales for FP8 KV cache
+# Required to be True when when kv_cache_dtype is fp8
+# When True, vLLM recalculates scales based on actual K/V distributions after each weight update
+# Works with or without weight quantization (quantization=fp8)
+calculate_kv_scales: False
```
wuxibin89: There's no need a separate flag here, `calculate_kv_scales` can be automatically set to True when`kv_cache_dtype=fp8`?

wuxibin89:
sglang may also need fp8 kv cache. https://github.com/volcengine/verl/pull/4415

Agoniii:
@sharonyu-115 could you help to update the [fp8 doc](https://github.com/volcengine/verl/blob/main/docs/advance/fp8.md) ?

verl/trainer/config/rollout/rollout.yaml
sharonyu-115: @wuxibin89 Sure. makes sense. I will make the code changes.

verl/trainer/config/rollout/rollout.yaml
sharonyu-115: Hi @wuxibin89 Thank you for your review and comments. I included this parameter intentionally to allow for future extensibility, particularly for supporting different methods of calibrating the q, k, v scales for accuracy purposes. For example, one option is to perform calibration using a subset of training data and then pass the computed scales to vLLM, rather than having vLLM calculate them on the fly. What's your thought?

wuxibin89: Will the q, k, v scales shift as training progress?

sharonyu-115: @wuxibin89 In my experiments with Qwen3-8B-Base, there is no major shift on q, k, v scales across the training, as the q, k, v distribution does not shift much. But I guess this is model and use case dependent.

verl/trainer/config/rollout/rollout.yaml
sharonyu-115: @wuxibin89  @Agoniii I refactored the code to have the quantization related parameters grouped under `quantization` as suggested. Could you give a review? At the meantime, I'm looking into the sglang kv cache quantization support and see how it can be integrated.

LZC-BELIEVER:
Hello! I have successfully replicated the KV cache FP8 quantization in sharonyu-115:kv-cache-fp8, which has significantly improve the throughput. But why hasn't this PR been merged until now and suspended for 3 months, is there anything wrong and is there anything I can help?

sharonyu-115:
> Hello! I have successfully replicated the KV cache FP8 quantization in sharonyu-115:kv-cache-fp8, which has significantly improve the throughput. But why hasn't this PR been merged until now and suspended for 3 months, is there anything wrong and is there anything I can help?

Hi @LZC-BELIEVER Thanks for the feedback. I was trying to run experiments with MoE models to verify the convergence. On Qwen3-30B DAPO, the observation was on MoE the training had some instability spike, though it was corrected (curves attached as a reference) and on AIME24 it got to 0.5. So I would say the feature is experimental and can be further delved into finer quantization granularity. On the other hand, I will have to rebase the PR code. Would you share a bit about your training scenario and observations? 

image

LZC-BELIEVER:
@sharonyu-115 So excited to see your reply! Firstly, I tried on Qwen3-8B-Base GRPO, using MATH dataset, and I can see that opening FP8 KV cache could show the highest throughput (around 10% promotion) and the lowest gen time.    
截屏2026-03-29 15 30 59
Secondly, I also tried on Qwen2.5-Math-7B GRPO with MATH dateset, and saw a terrible result: The accuracy became around 0 when I open FP8 KV cache, and response was irregular like"????////". I compared the model structure between Qwen2.5 and Qwen3, and found that in qwen3 model there is QK-Norm which could reduce outliers and help stabilize the KV cache quantization. Maybe it is better to add some specific process for models without QK-Norm.  

Your experiment above is interesting and I have seen similar result in my previous experiment on FlashRL . If it is the problem, I would try to repeat on Qwen3-30B DAPO and see what I can do to help. Did you use the script verl/recipe/dapo/run_dapo_qwen3_moe_30b_vllm_fp8_rollout.sh on the experiment you mentioned?

sharonyu-115:
> Your experiment above is interesting and I have seen similar result in my previous experiment on FlashRL . If it is the problem, I would try to repeat on Qwen3-30B DAPO and see what I can do to help. Did you use the script verl/recipe/dapo/run_dapo_qwen3_moe_30b_vllm_fp8_rollout.sh on the experiment you mentioned?

Hi @LZC-BELIEVER That would be great if you can also run an experiment. Thanks! Yes, you can use that script. While in my run one difference is I used `train_prompt_bsz=128` `train_prompt_mini_bsz=128`,`gen_prompt_bsz=128*3`.

PR #4445, opened 2025-12-08
[ci] test: Disable ReMax training test in vllm workflow

PeterSH6:


.github/workflows/e2e_ppo_trainer_megatron_vllm_2.yml
vermouth1992: This is probably problematic. This test should be run if we modify this file

PR #4469, opened 2025-12-09
[rollout] feat: pass agent_data to tool calling

wuxibin89:
### What does this PR do?

1. Move tool initialization from classmethod `AgentLoopBase.init_class` to `AgentLoopBase.__init__`
In previous design, we assume that tool initialization is heavy and should only initialized once in `ToolAgentLoop`. This prevents specifying individual tools for each sample in RLHFDataset,  for example we may have mixed RL datasets with different tools: code, search, GUI, etc.
By moving tool initialization to `AgentLoopBase.__init__`, we can initialize tools for each sample individually. We also assume that tool initialization is lightweight. For heavy tool setup, user should access remote service with rpc call in `tool.execute`.

2. Pass `AgentData` to tool calling
Tool calling may need to access full agent state, including: history messages, multi-modal images/videos. Also, tool may need to store some session data across multi-turn, which can be stored in `AgentData.extra_fields`.

### DIFF
--- a/recipe/fully_async_policy/agent_loop/agent_loop.py
+++ b/recipe/fully_async_policy/agent_loop/agent_loop.py
@@ -27,8 +27,8 @@ from verl.experimental.agent_loop.agent_loop import (
     AgentLoopOutput,
     AgentLoopWorkerBase,
     AsyncLLMServerManager,
+    DictConfigWrap,
     _agent_loop_registry,
-    _DummyConfig,
     get_trajectory_info,
 )
 from verl.experimental.agent_loop.prometheus_utils import update_prometheus_config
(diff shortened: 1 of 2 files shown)


SuffixAutomata:
[https://github.com/volcengine/verl/pull/4469/files#diff-25157fd1cdb544dfb3245f630bfbd49cff24e79eac8ef893ce440217eb92b385R288](https://github.com/volcengine/verl/pull/4469/files#diff-25157fd1cdb544dfb3245f630bfbd49cff24e79eac8ef893ce440217eb92b385R288) Why are we passing both `agent_data` and `agent_data.tools_kwargs` to `_call_tool`?

lulia0228:
> 1. By moving tool initialization to `AgentLoopBase.__init__`, we can initialize tools for each sample individually.

I didn’t see the logic in AgentLoopBase.__init__ where each sample is initialized with its own specific set of candidate tools. In tool_agent_loop.py, self.tools = {tool.name: tool for tool in tool_list} still seems to pass the entire tool_list (i.e., all tools). Am I misunderstanding something here, or is there an issue with my interpretation?

=====
Just to clarify, do you mean that because instances are not shared anymore, we can initialize specific tool sets by replacing tool=self.schema with the candidate tool list recalled from the single training data within the instance initialization?

PR #4487, opened 2025-12-11
add system_prompt support to SFTDataset

zhangxinnan123:
### What does this PR do?

   Add system_prompt_key and system_prompt_dict_keys configuration options to enable system prompts in SFT training. The dataset now extracts system prompts from parquet files and includes them in the chat template with the "system" role.

   Changes:
   - Add system_prompt_key and system_prompt_dict_keys to sft_trainer.yaml config
   - Extract system prompts in _read_files_and_tokenize method
   - Include system prompts in chat template when applying tokenization

### DIFF
--- a/verl/trainer/config/sft_trainer.yaml
+++ b/verl/trainer/config/sft_trainer.yaml
@@ -13,8 +13,10 @@ data:
   # Single-turn settings
   prompt_key: question
   response_key: answer
+  system_prompt_key: null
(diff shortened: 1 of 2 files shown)


vermouth1992:
verl/utils/dataset/sft_dataset.py is deprecated and will be removed. Please use https://github.com/volcengine/verl/blob/main/verl/utils/dataset/multiturn_sft_dataset.py

PR #4492, opened 2025-12-11
[trainer] fix: model engine vlm multi_modal_inputs to NonTensorStack

wuxibin89:
### What does this PR do?

Fix RL model engine for VLM. 

Qwen/Qwen3-VL-30B-A3B-Instruct fsdp vs megatron on geo3k:
image


verl/single_controller/base/decorator.py
```diff
@@ -72,20 +72,48 @@ def init_predefined_execute_mode():
 init_predefined_execute_mode()
 
 
+def _chunk_tensordict(td: TensorDict, chunks: int) -> list[TensorDict]:
```
vermouth1992: Please move this function to verl/utils/tensordict_utils and add a test. Thanks!

wuxibin89: Done

PR #4510, opened 2025-12-14
[algo] fix: Add seq mean mask denominator option

szrlee:
## Summary

Refactor `agg_loss` function and fix entropy/KL loss scaling in distributed training.

**Changes:**
- **Refactor**: Unify `seq-mean-*` modes with shared denominator logic using `masked_sum`
- **Behavior change**: `seq-mean-token-sum-norm` now applies seq-mean division (denominator = `global_batch_size * dp_size` or `local_bsz`), matching the mode name
- **Simplification**: Remove fully-masked sequence exclusion from denominator; use total batch size consistently

NOTE: Since the global loss aggregation logic is not compatible with the legacy model engine that conduct the aggregation outside `agg_loss` and is going to be deprecated, we keep this PR from modifying the the legacy model engine.

⚠️ **Breaking**: `seq-mean-token-sum-norm` now divides by both `loss_scale_factor` AND `seq_denominator`. Previously only divided by `loss_scale_factor`.

## Test plan

- [ ] Verify PPO training with `seq-mean-token-sum` mode
- [ ] Verify PPO training with `seq-mean-token-mean` mode  
- [ ] Verify PPO training with `seq-mean-token-sum-norm` mode (note: behavior changed)
- [ ] Confirm entropy/KL loss values are correctly scaled in multi-GPU training

### DIFF
--- a/verl/trainer/ppo/core_algos.py
+++ b/verl/trainer/ppo/core_algos.py
@@ -781,6 +781,9 @@ def agg_loss(
     """
     Aggregate the loss across global batch to ensure the loss is invariant to fsdp/megatron parallelism.
 
+    NOTE: ``dp_size``, ``batch_num_tokens``, and ``global_batch_size`` are only compatible with the new model engine
+        for now, while the legacy model engines conduct the aggregation outside ``agg_loss``.
+
     NOTE: The returned loss has different behaviors for different backend:
     - FSDP: the loss is directly used for backward.
     - Megatron: the loss should be scaled by `num_microbatches` and `cp_size` for pp schedule.
--- a/verl/workers/actor/dp_actor.py
+++ b/verl/workers/actor/dp_actor.py
@@ -456,10 +456,6 @@ class DataParallelPPOActor(BasePPOActor):
                     entropy_coeff = self.config.entropy_coeff
                     loss_agg_mode = self.config.loss_agg_mode
 
-                    # Populate global_batch_info for loss aggregation
-                    self.config.global_batch_info["dp_size"] = torch.distributed.get_world_size()
-                    self.config.global_batch_info["global_batch_size"] = model_inputs.get("global_batch_size", None)
-
                     calculate_entropy = self.config.calculate_entropy or (entropy_coeff != 0)
 
                     if self.config.use_dynamic_bsz:
@@ -520,12 +516,7 @@ class DataParallelPPOActor(BasePPOActor):
 
                     policy_loss = pg_loss
                     if calculate_entropy and entropy is not None:
-                        entropy_agg = agg_loss(
-                            loss_mat=entropy,
-                            loss_mask=response_mask,
-                            loss_agg_mode=loss_agg_mode,
-                            **self.config.global_batch_info,
-                        )
+                        entropy_agg = agg_loss(loss_mat=entropy, loss_mask=response_mask, loss_agg_mode=loss_agg_mode)
                         micro_batch_metrics["actor/entropy"] = entropy_agg.detach().item()
                         if entropy_coeff != 0:
                             policy_loss -= entropy_agg * entropy_coeff
@@ -536,12 +527,7 @@ class DataParallelPPOActor(BasePPOActor):
                         kld = kl_penalty(
                             logprob=log_prob, ref_logprob=ref_log_prob, kl_penalty=self.config.kl_loss_type
                         )
-                        kl_loss = agg_loss(
-                            loss_mat=kld,
-                            loss_mask=response_mask,
-                            loss_agg_mode=loss_agg_mode,
-                            **self.config.global_batch_info,
-                        )
+                        kl_loss = agg_loss(loss_mat=kld, loss_mask=response_mask, loss_agg_mode=loss_agg_mode)
 
                         policy_loss = policy_loss + kl_loss * self.config.kl_loss_coef
                         micro_batch_metrics["actor/kl_loss"] = kl_loss.detach().item() * loss_scale_factor
--- a/verl/workers/actor/megatron_actor.py
+++ b/verl/workers/actor/megatron_actor.py
@@ -479,11 +479,6 @@ class MegatronPPOActor(BasePPOActor):
 
                 entropy_coeff = self.config.entropy_coeff
                 loss_agg_mode = self.config.loss_agg_mode
-
-                # Populate global_batch_info for loss aggregation
-                self.config.global_batch_info["dp_size"] = data.get("dp_size", 1)
-                self.config.global_batch_info["global_batch_size"] = data.get("global_batch_size", None)
-
                 loss_mode = self.config.policy_loss.get("loss_mode", "vanilla")
 
                 policy_loss_fn = get_policy_loss_fn(loss_mode)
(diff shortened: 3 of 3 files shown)


tongyx361 (changes requested):
Loss/gradient aggregation in DP training is actually tricky, which is not thoroughly considered by this PR:

1. Distributed training frameworks sometimes apply aggregation implicitly, e.g., FSDP by default means the loss (sums the loss and divides by the `dp_size`), which will make loss/gradient in the PR's implementation shrink by `dp_size` from the targeted ground truth. As far as I know, for FSDP2 in `torch>=2.8.0`, we can resolve this by calling `set_gradient_divide_factor(1)`, but I am not sure about other setups like FSDP1 and Megatron.
2. If DP balance (`balance_batch`) is enabled, for now, the `batch_size` is implemented as even between DP ranks, but this is actually not mandatory and suboptimal, which might be optimized to allow uneven allocation for better workload balance in the future, so the implementation with `all_reduce(SUM)` suggested by Gemini is indeed more robust (but still problematic, see 3).
3. If Ulysses SP is enabled, the data will be all-gathered within each USP group of `sp_size`, which might cause the global `batch_size`  multiplied by `sp_size` if simply summed up with `all_reduce(SUM)` as is suggested by Gemini.

verl/workers/actor/dp_actor.py
```diff
@@ -456,6 +456,10 @@ def update_policy(self, data: DataProto):
                     entropy_coeff = self.config.entropy_coeff
                     loss_agg_mode = self.config.loss_agg_mode
 
+                    # Populate global_batch_info for loss aggregation
+                    self.config.global_batch_info["dp_size"] = torch.distributed.get_world_size()
```
ISEEKYAN: dp_size = world_size/ulysses_sequence_parallel_size in FSDP

verl/workers/actor/megatron_actor.py
```diff
@@ -480,6 +480,10 @@ def loss_func(output, data, meta_info):
                 entropy_coeff = self.config.entropy_coeff
                 loss_agg_mode = self.config.loss_agg_mode
 
+                # Populate global_batch_info for loss aggregation
+                self.config.global_batch_info["dp_size"] = data.get("dp_size", 1)
```
ISEEKYAN: should be `megatron.core.parallel_state.get_data_parallel_world_size()`

verl/trainer/ppo/core_algos.py
```diff
@@ -785,11 +785,14 @@ def agg_loss(
...
     - Megatron: the loss should be scaled by `num_microbatches` and `cp_size` for pp schedule.
 
+    # TODO: Consider the numerical stability?
+
     Args:
         loss_mat: micro batch loss matrix, (bs, response_length)
         loss_mask: micro batch loss mask, (bs, response_length)
         loss_agg_mode: method to aggregate the loss matrix into a scalar
-        dp_size: data parallel size
+        dp_size: data parallel size. When appling manual aggregation,
```
ISEEKYAN, line 797: is it possible that we remove the dp_size from the algorithm loss function?

verl/workers/actor/dp_actor.py
```diff
@@ -516,7 +520,12 @@ def update_policy(self, data: DataProto):
 
                     policy_loss = pg_loss
                     if calculate_entropy and entropy is not None:
-                        entropy_agg = agg_loss(loss_mat=entropy, loss_mask=response_mask, loss_agg_mode=loss_agg_mode)
+                        entropy_agg = agg_loss(
+                            loss_mat=entropy,
+                            loss_mask=response_mask,
+                            loss_agg_mode=loss_agg_mode,
+                            **self.config.global_batch_info,
```
wuxibin89: I think `dp_size`, `global_batch_size` and `batch_num_tokens` is only needed in new model engine backend, legacy backend should not pass these parameters. cc @vermouth1992

verl/trainer/ppo/core_algos.py
szrlee: dp_size was brought into loss by @wuxibin89

wuxibin89:
`dp_size`, `batch_num_tokens` and `global_batch_size` were introduced to make sure each DP group's loss is averaged across global mini batch instead of averaged over local micro batch. This is mean to correct the contribution of micro-batch to gradient from different dp groups, which have different number of valid tokens and sequences. **We want to make sure that each token each sequence in micro batch has equal contribution to gradient.**

For example, we have dp_size=2 and num_micro_batches=2 (Gradient is accumulated across 2 micro batches in each dp group, then mean across 2 dp groups).
- dp_rank=0: [micro_batch_0_0, micro_batch_0_1]
- dp_rank=1: [micro_batch_1_0, micro_batch_1_1]

Then for `agg_loss`, we have
- dp_size: 2
- batch_num_tokens: sum(micro_batch_0_0, micro_batch_0_1, micro_batch_1_0, micro_batch_1_1)
- global_batch_size: sum(len(micro_batch_0_0), len(micro_batch_0_1), len(micro_batch_1_0), len(micro_batch_1_1))

@vermouth1992 create a example for explanation: https://gist.github.com/vermouth1992/6c273240765c4f223478081042bfcd4a

verl/workers/actor/dp_actor.py
wuxibin89: `global_batch_info` should only be passed in new model engine: https://github.com/volcengine/verl/blob/main/verl/workers/engine_workers.py

For legacy engine `fsdp_workers.py` and `megatron_workers.py`, we should not pass it.

wuxibin89: No, legacy engine scale loss manually: https://github.com/volcengine/verl/blob/main/verl/workers/actor/dp_actor.py#L461-L464

tongyx361:
Since the global loss aggregation logic is not compatible with the legacy model engine that conduct the aggregation outside `agg_loss` and is going to be deprecated, we keep this PR from modifying the the legacy model engine.

So the comments above are either resolved or avoided. @wuxibin89 @ISEEKYAN .

verl/trainer/ppo/core_algos.py
tongyx361: This comment is kept unresolved to show why @wuxibin89 introduced the `dp_size` below.

wuxibin89:
image

This PR make new model engine loss significant small than expected.

verl/trainer/ppo/core_algos.py
```diff
@@ -799,30 +805,39 @@ def agg_loss(
...
-        loss = verl_F.masked_sum(seq_losses, seq_mask) / global_batch_size * dp_size  # seq-mean
-    elif loss_agg_mode == "seq-mean-token-sum-norm":
-        seq_losses = torch.sum(loss_mat * loss_mask, dim=-1)
-        if loss_scale_factor is None:
-            loss_scale_factor = loss_mask.shape[-1]
-        loss = torch.sum(seq_losses) / loss_scale_factor
+    elif loss_agg_mode.startswith("seq-mean"):
+        # TODO: Correct and unify the denominator logic.
+        if global_batch_size is not None:
+            seq_denominator = global_batch_size * dp_size
```
wuxibin89, line 816: The `seq_denominator` is not right, should be `global_batch_size / dp_size`

PR #2510, opened 2025-07-14
[sglang] feat: Adds video input support to SGLang rollout

JustinTong0323:
### What does this PR do?
DO NOT MERGE

This PR Extends SGLang rollout to support video data.

This change integrates video data processing into the rollout workflow, allowing the engine to handle video inputs alongside existing text and image functionalities.

### Test

> For changes that can not be tested by CI (e.g., algorithm implementation, new model support), validate by experiment(s) and show results like training curve plots, evaluation results, etc.

### API and Usage Example

> Demonstrate how the API changes if any, and provide usage example(s) if possible.

```python
# Add code snippet or script demonstrating how to use this
```

### Design & Code Changes

> Demonstrate the high-level design if this PR is complex, and list the specific changes.

### Checklist Before Submitting

> [!IMPORTANT]
> Please check all the following items before requesting a review, otherwise the reviewer might deprioritize this PR for review.

- [x] Read the [Contribute Guide](https://github.com/volcengine/verl/blob/main/CONTRIBUTING.md).
- [x] Apply [pre-commit checks](https://github.com/volcengine/verl/blob/main/CONTRIBUTING.md#code-linting-and-formatting): `pre-commit install && pre-commit run --all-files --show-diff-on-failure --color=always`
- [ ] Add / Update [the documentation](https://github.com/volcengine/verl/tree/main/docs).
- [ ] Add unit or end-to-end test(s) to [the CI workflow](https://github.com/volcengine/verl/tree/main/.github/workflows) to cover all the code. If not feasible, explain why: ...
- [ ] Once your PR is ready for CI, send a message in [the `ci-request` channel](https://verl-project.slack.com/archives/C091TCESWB1) in [the `verl` Slack workspace](https://join.slack.com/t/verl-project/shared_invite/zt-3855yhg8g-CTkqXu~hKojPCmo7k_yXTQ).

PR #4513, opened 2025-12-15
[vLLM] feat: Adds video input support to vLLM rollout

Miku2G:
### What does this PR do?

Ensure video data is passed to vLLM to prevent empty features in vlm mrope calculations

#### Description
This PR modifies the data flow to ensure video data is correctly delivered to vLLM alongside other multi-modal information. This fixes an issue where Qwen3-VL would receive empty video features during mrope calculations.

#### Detailed Changes:

verl/experimental/agent_loop/single_turn_agent_loop.py: Extracted video data from the non-tensor batch, passed it to the server, and preserved it in the output. While the processor is still used to generate prompt IDs, it no longer processes the video data locally; instead, the video is passed directly to the inference backend.

verl/experimental/agent_loop/agent_loop.py: Updated AsyncLLMServerManager.generate to support video_data argument and forward it to the underlying vLLM actor.

verl/workers/rollout/vllm_rollout/vllm_async_server.py: Modified TokensPrompt to include video data within multi_modal_data when passing it to vLLM (alongside the existing image data).

### Checklist Before Starting

- [✅️] Search for similar PRs. Paste at least one query link here: [feat: Adds video input support to SGLang rollout](https://github.com/volcengine/verl/pull/2510)
- [✅️] Format the PR title as `[{modules}] {type}: {description}` (This will be checked by the CI)

### Test

### API and Usage Example

```python
```

### Design & Code Changes

> [!IMPORTANT]

- [✅️]  Read the [Contribute Guide](https://github.com/volcengine/verl/blob/main/CONTRIBUTING.md).
- [✅️]  Apply [pre-commit checks](https://github.com/volcengine/verl/blob/main/CONTRIBUTING.md#code-linting-and-formatting): `pre-commit install && pre-commit run --all-files --show-diff-on-failure --color=always`

### DIFF
--- a/verl/experimental/agent_loop/single_turn_agent_loop.py
+++ b/verl/experimental/agent_loop/single_turn_agent_loop.py
@@ -37,6 +37,7 @@ class SingleTurnAgentLoop(AgentLoopBase):
     async def run(self, sampling_params: dict[str, Any], **kwargs) -> AgentLoopOutput:
         messages = list(kwargs["raw_prompt"])
         image_data = copy.deepcopy((kwargs.get("multi_modal_data") or {}).get("image", None))
+        video_data = copy.deepcopy((kwargs.get("multi_modal_data") or {}).get("video", None))
 
         metrics = {}
         request_id = uuid4().hex
@@ -52,6 +53,7 @@ class SingleTurnAgentLoop(AgentLoopBase):
                     **self.apply_chat_template_kwargs,
                 ),
             )
+            # We only need prompt token ids here; multi-modal payload will be sent directly to the server.
             model_inputs = self.processor(text=[raw_prompt], images=image_data, return_tensors="pt")
             prompt_ids = model_inputs.pop("input_ids").squeeze(0).tolist()
         else:
@@ -60,11 +62,15 @@ class SingleTurnAgentLoop(AgentLoopBase):
                 lambda: self.tokenizer.apply_chat_template(
                     messages, add_generation_prompt=True, tokenize=True, **self.apply_chat_template_kwargs
                 ),
-            )
+        )
 
         with simple_timer("generate_sequences", metrics):
             output = await self.server_manager.generate(
-                request_id=request_id, prompt_ids=prompt_ids, sampling_params=sampling_params, image_data=image_data
+                request_id=request_id,
+                prompt_ids=prompt_ids,
+                sampling_params=sampling_params,
+                image_data=image_data,
+                video_data=video_data,
             )
         response_mask = [1] * len(output.token_ids)
 
@@ -78,7 +84,7 @@ class SingleTurnAgentLoop(AgentLoopBase):
                 if output.routed_experts is not None
                 else None
             ),
-            multi_modal_data={"image": image_data} if image_data is not None else {},
+            multi_modal_data={k: v for k, v in {"image": image_data, "video": video_data}.items() if v is not None},
             num_turns=2,
             metrics=metrics,
         )
--- a/verl/experimental/agent_loop/single_turn_agent_loop.py
+++ b/verl/experimental/agent_loop/single_turn_agent_loop.py
@@ -17,6 +17,9 @@ import os
 from typing import Any
 from uuid import uuid4
 
+import numpy as np
+import torch
+
 from verl.experimental.agent_loop.agent_loop import AgentLoopBase, AgentLoopOutput, register
 from verl.utils.profiler import simple_timer
 
@@ -39,6 +42,22 @@ class SingleTurnAgentLoop(AgentLoopBase):
         image_data = copy.deepcopy((kwargs.get("multi_modal_data") or {}).get("image", None))
         video_data = copy.deepcopy((kwargs.get("multi_modal_data") or {}).get("video", None))
 
+        videos = None
+        videos_kwargs = {}
+        if video_data is not None:
+            videos = []
+            video_metadata = []
+            for item in video_data:
+                if isinstance(item, tuple) and len(item) == 2:
+                    v, meta = item
+                else:
+                    v, meta = item, {}
+                if isinstance(v, np.ndarray):
+                    v = torch.from_numpy(v)
+                videos.append(v)
+                video_metadata.append(meta)
+            videos_kwargs = {"video_metadata": video_metadata, "do_sample_frames": False}
+
         metrics = {}
         request_id = uuid4().hex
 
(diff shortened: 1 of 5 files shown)


DesertsP:
This PR correctly fixes the issue where Qwen3-VL receives empty video features. 👍

One concern regarding backward compatibility:  Qwen2.5-VL also relies on additional video_kwargs, such as fps.

It may be worth considering:
- Explicit support for video_kwargs in the agent loop and AsyncLLMServerManager.generate
- Feeding both video_data and associated mm_processor_kwargs to vLLM

Miku2G:
@DesertsP 
Thanks for the suggestion. I also discovered that during testing with Qwen3-VL, agent_loop relies solely on the image_processor class name. As a result, it always imports qwen2_vl.get_rope_index regardless of whether it's Qwen2 or Qwen3. This causes the execution to fall into qwen2_vl.py, triggering a shape mismatch.

Therefore, further modifications are needed to ensure that qwen3_vl.get_rope_index is used specifically when Qwen3VLProcessor is encountered (otherwise, it should default to qwen2_vl.get_rope_index)

I am currently working on a fix for this.

EchoMinkki:
> @DesertsP Thanks for the suggestion. I also discovered that during testing with Qwen3-VL, agent_loop relies solely on the image_processor class name. As a result, it always imports qwen2_vl.get_rope_index regardless of whether it's Qwen2 or Qwen3. This causes the execution to fall into qwen2_vl.py, triggering a shape mismatch.
> 
> Therefore, further modifications are needed to ensure that qwen3_vl.get_rope_index is used specifically when Qwen3VLProcessor is encountered (otherwise, it should default to qwen2_vl.get_rope_index)
> 
> I am currently working on a fix for this.

@Miku2G 
I actually encountered the exact same shape mismatch issue.

My investigation suggests that the root cause might be a conflict between vLLM's video sequence expansion logic and Qwen3-VL's specific format — specifically how Qwen3 treats independent video frames with <timestamp><|vision_start|><|video_pad|>...<|vision_end|> blocks.

I managed to hack together a working version that bypasses this conflict and runs successfully, though my implementation is currently quite "inelegant" and hacky.

Would you be interested in seeing my temporary fix? I'd be happy to share the code snippet or discuss it further to help find a cleaner solution for your PR.

Miku2G:
@EchoMinkki 
The Qwen3-VL processor in vLLM requires the input prompt to retain the original video placeholder string `<|vision_start|><|video_pad|><|vision_end|>`. The engine expects to handle the expansion (replacing it with timestamps + multiple video_pad tokens) internally.

Currently, the HuggingFace processor expands these placeholders before the request reaches vLLM. As a result, vLLM cannot find the expected replacement target, triggering the following error: `AssertionError: Failed to apply prompt replacement for mm_items['video'][0]`

I am implementing two fixes to address this:

single_turn_agent_loop.py:

Separate the prompt IDs. We will pass server_prompt_ids (with unexpanded placeholders) to vLLM, while still using the expanded prompt_ids via the processor for the training output/logs.

vllm_async_server.py:

Adjust the max_tokens calculation. Since the unexpanded prompt is significantly shorter, the default calculation might result in an overestimated max_tokens value. I'm adding a check similar to `max_tokens = min(response_length, max_model_len - len(prompt_ids))` to prevent context window overflow.

EchoMinkki:
@Miku2G 
Thanks for the breakdown!

Actually, the 'inelegant' fix I mentioned earlier turns out to be exactly the implementation you described for single_turn_agent_loop.py (separating prompt_ids for vllm and training).

Since I already have this logic working locally, shall I go ahead and submit a PR for this part? This might save you some time and let you focus on the vllm_async_server.py adjustment

Miku2G:
@EchoMinkki 
I've completed the modifications and will update shortly. It passed tests on Qwen3VL. If you still have suggesttions about the code, you are absolutely welcome to submit a PR.

EchoMinkki:
@Miku2G 
That sounds fantastic! Since it already passed the tests on Qwen3VL, I think it's solid.I'll skip the PR then, since you've got it covered. Really appreciate your quick work and efficiency!😊

Miku2G:
@DesertsP 
In this commit, the agent loop now explicitly supports two non-mixed video input formats:                                                                                                                           
   1. Qwen-style video dicts (as documented in verl/utils/dataset/vision_utils.py, where fps/min_frames/max_frames live in the dict). For these, we preprocess via process_video(..., return_video_metadata=True) and normalize to the (frames, metadata) format expected by the downstream HF processor and vLLM.                                                                                                                  
   2. Pre-decoded frames (numpy/torch tensors, optionally wrapped as (frames, metadata) tuples). For these, we keep do_sample_frames=False and pass video_metadata when available. If a list mixes dict-based specs and pre-decoded frames, we now raise a clear ValueError instead of silently producing empty video features.                                                                    
                                                                                                                                                                                                                
  Regarding AsyncLLMServerManager.generate: I intentionally didn’t add a new mm_processor_kwargs / video_kwargs parameter because it would expand a cross-backend RPC/API contract and require plumbing through all rollout server implementations. The official sampling knobs are already carried by the video dict itself, and after preprocessing we pass (frames, metadata), so we don’t need to change the generate signature to address this issue.

verl/experimental/agent_loop/single_turn_agent_loop.py
```diff
@@ -52,19 +119,40 @@ async def run(self, sampling_params: dict[str, Any], **kwargs) -> AgentLoopOutpu
...
                 ),
             )
-            model_inputs = self.processor(text=[raw_prompt], images=image_data, return_tensors="pt")
-            prompt_ids = model_inputs.pop("input_ids").squeeze(0).tolist()
+            # vLLM's multimodal preprocessor expects *unexpanded* placeholders in the prompt
+            # (e.g. "<|vision_start|><|video_pad|><|vision_end|>"), and will expand them based on mm inputs.
+            # So we:
+            # 1) send unexpanded prompt ids to vLLM for generation
+            # 2) keep expanded prompt ids in the rollout output for training-side forward passes
+            server_prompt_ids = (
```
wuxibin89, line 127: This change introduces two bugs:
1. break sglang rollout since it expects **expanded** input_ids, so we refold `<|image_pad|>` for vllm in:
https://github.com/volcengine/verl/blob/main/verl/workers/rollout/vllm_rollout/vllm_async_server.py#L774 
2. silent mismatch in server_prompt_ids and expanded_inputs since we tokenize raw_prompt twice, which may produce different input_ids leading to trainer-rollout mismatch.

PR #4567, opened 2025-12-17
[megatron] fix: Fix RouterReplay/TopKRouter mismatch to support qwen3vl

Kairosxy:
### What does this PR do?

> Fix prebuild-induced mismatch between `RouterReplay.router_instances` and `TopKRouter`, enabling Router Replay for Qwen3-VL.

### Problem Description

Router Replay relies on the global cache `RouterReplay.router_instances` to record/reuse router instances and routing decisions. In the Qwen3-VL initialization flow, the mbridge prebuild stage creates and registers router instances before the actual model build. Because the cache is not cleared prior to the real build, stale entries remain in `RouterReplay.router_instances`, leading to a configuration/dimension mismatch with the current `TopKRouter` and breaking Router Replay.

### Solution

Call `RouterReplay.router_instances.clear()` in the low-level model build path when invoking `TransformerBlock`, purging routers created during the prebuild stage. This ensures `router_instances` always matches the current model structure/dimensions on each rebuild, restoring Router Replay support for Qwen3-VL.

Additionally, when using mbridge, explicitly inject `enable_routing_replay: bool = False` into `Qwen3VLTransformerConfig` (and other `TransformerConfig` variants). This flag serves as the Router Replay feature switch (default off) and ensures the config exposes the required attribute for enabling/disabling Router Replay. For example:

```python
@dataclass
class Qwen3VLTransformerConfig(TransformerConfig):
    enable_routing_replay: bool = False

    patch_size: int = 14
    temporal_patch_size: int = 2
    in_channels: int = 3
    spatial_merge_size: int = 2
    num_position_embeddings: int = 2304
    out_hidden_size: int = 2304

    apply_rotary_pos_emb_in_fp32: bool = False
    deepstack_visual_indexes: List[int] = field(default_factory=lambda: [8, 16, 24])
```


ISEEKYAN:
thanks to @Kairosxy  for your great contribution!

ISEEKYAN (dismissed):

ISEEKYAN:
Hi @Kairosxy, would you please fix the conflict so I can merge this?

Kairosxy:
Thanks for the note. Conflicts are resolved.


Upon review, I realized the incoming implementation was superior to my initial draft, so I’ve aligned the merge with that approach.






thought
***@***.***

Issue #4555, opened 2025-12-17
[RFC][ray] Create single placement group for resource pool


wuxibin89:
Currently, verl resource pool create a placement group for each node, which is not compatible with inference framework using ray for worker orchestration. For example, in multiple nodes(e.g TP=16) serving, inference framework needs a single placement group with `[{"GPU": 1, "CPU": 1}] * 16`.
- vLLM: use [VLLM_RAY_PER_WORKER_GPUS,VLLM_RAY_BUNDLE_INDICES](https://github.com/vllm-project/vllm/blob/v0.12.0/vllm/envs.py#L1068-L1077) to specify which bundles used for this instance and gpu fraction for each work.
- TensorRT-LLM: use [TRTLLM_RAY_BUNDLE_INDICES](https://github.com/NVIDIA/TensorRT-LLM/blob/v1.2.0rc5/tensorrt_llm/executor/ray_executor.py#L277) to specify which bundles used for this instance.

To satisfy inference framework requirement, propose to create one single placement group for all nodes in resource pool.
- before: one placement group for one node

<img width="607" height="343" alt="Image" src="https://github.com/user-attachments/assets/d0f9284d-cd76-4fb7-b8ec-f65403122658" />

- after: one placement group for all nodes

<img width="682" height="411" alt="Image" src="https://github.com/user-attachments/assets/67fb0199-b0d3-441a-a2a1-abc96b78e9aa" />

Related PR
- https://github.com/volcengine/verl/pull/4280
- https://github.com/volcengine/verl/pull/4515

wuxibin89:
cc @zw0610 @vermouth1992  @davidmlw @jianjunzhong @yyDing1

PR #4583, opened 2025-12-18
[ray] feat: resource pool create one placement group across all nodes

wuxibin89:
### What does this PR do?

RFC: https://github.com/volcengine/verl/issues/4555

### DIFF
--- a/tests/single_controller/test_high_level_scheduling_api.py
+++ b/tests/single_controller/test_high_level_scheduling_api.py
@@ -17,7 +17,7 @@ import time
 import ray
 
 from verl.single_controller.base.worker import Worker
-from verl.single_controller.ray.base import RayClassWithInitArgs, RayResourcePool, RayWorkerGroup, merge_resource_pool
+from verl.single_controller.ray.base import RayClassWithInitArgs, RayResourcePool, RayWorkerGroup, split_resource_pool
 
 
 @ray.remote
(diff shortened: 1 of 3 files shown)


verl/single_controller/ray/base.py
```diff
@@ -431,13 +436,13 @@ def _init_with_detached_workers(self, worker_names, worker_handles):
         self._workers = workers
         self._world_size = len(worker_names)
 
-    def _get_master_addr_port(self, pg):
+    def _get_master_addr_port(self, pg, bundle_index):
```
zw0610, line 439: Maybe replace `bundle_index` with another name like `worker_idx` because `bundle` belongs to the concept of placement group (resources allocation) while getting master addr & port is part of creating workers.

tests/single_controller/test_high_level_scheduling_api.py
```diff
@@ -17,7 +17,7 @@
 import ray
 
 from verl.single_controller.base.worker import Worker
-from verl.single_controller.ray.base import RayClassWithInitArgs, RayResourcePool, RayWorkerGroup, merge_resource_pool
+from verl.single_controller.ray.base import RayClassWithInitArgs, RayResourcePool, RayWorkerGroup, split_resource_pool
```
zw0610, line 20: Does this mean `merge_resource_pool` is no longer supported?

zw0610 (commented):
Generally /LGTM
I left some comments aimed for other developers to better understand the changes.

tests/single_controller/test_high_level_scheduling_api.py
wuxibin89: Yes, since resource pool contains only one placement group, we can't merge two resource pools. I have checked that both open source and internal repos depending verl not using `merge_resource_pool`.

wuxibin89:
Close as ray placement group [PACK strategy](https://docs.ray.io/en/latest/ray-core/scheduling/placement-group.html#placement-strategy) has no guarantee that continuous bundles are scheduling to same nodes.
- bundle 0: node-0 
- bundle 1: node-1
- bundle 2: node-0
- bundle 3: node-0
- ...

PR #4611, opened 2025-12-19
[megatron] fix: fix MLA with sequence packing + CP

wuweiqiang24:
### What does this PR do?

> Fix size mismatch bug in rotary position embedding for MLA when using context parallel

### Problem Description
When using MLA in Megatron, Verl applies the patch_v012 to modify `get_query_key_value_tensors` in `multi_latent_attention.py`. During training, the rotary position embedding (`rotary_pos_emb`) is typically truncated to the sequence length in Verl. However, this behavior is not suitable when both sequence packing and context parallelism (CP) are enabled.

In the case of sequence packing + CP, Megatron does not split `rotary_pos_emb` according to the sequence length on each CP rank. As a result, the embedding may appear to exceed the expected sequence length for that CP rank. To maintain correct positional information across the full sequence, the complete `rotary_pos_emb` must be preserved — hence, it should not be shortened.

Consequently, when MLA and CP are both enabled in Verl, a `rotary_pos_emb` size mismatch error may occur.

### Solution

Introduce an additional condition to determine whether `rotary_pos_emb` should be truncated to the sequence length, ensuring correct behavior in the case of sequence packing + CP.

### DIFF
--- a/verl/models/mcore/patch_v012.py
+++ b/verl/models/mcore/patch_v012.py
@@ -139,16 +139,21 @@ def apply_patch():
                 self.config.qk_head_dim + self.config.v_head_dim,
             )
 
+            cp_size = parallel_state.get_context_parallel_world_size()
             if inference_context is not None:
                 # add offset to the sequence start for inference
                 sequence_start = inference_context.sequence_len_offset
                 sequence_end = sequence_start + q_len
                 rotary_pos_emb = rotary_pos_emb[sequence_start:sequence_end]
-            else:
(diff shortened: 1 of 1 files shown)


FightingZhen:
Looks like a bug in 0.12 patch and have been fixed in https://github.com/NVIDIA/Megatron-LM/blob/0d7e02bd26d3eb347ad8d1bef625002babfa6df7/megatron/core/transformer/multi_latent_attention.py#L724, seems fine, @wuxibin89 could you please help us have a check :)

PR #4644, opened 2025-12-22
[fsdp] feat: FSDP2 Only Load Rank0

EduardDurech:
Supersedes https://github.com/volcengine/verl/pull/3493, only load model on rank0 which is broadcasted later `fsdp2_load_full_state_dict`

Note: Not implemented for PEFT

### DIFF
--- a/verl/workers/engine/fsdp/transformer_impl.py
+++ b/verl/workers/engine/fsdp/transformer_impl.py
@@ -207,7 +207,8 @@ class FSDPEngine(BaseEngine):
         torch_dtype = PrecisionType.to_dtype(torch_dtype)
 
         init_context = get_init_weight_context_manager(
-            use_meta_tensor=not self.model_config.hf_config.tie_word_embeddings, mesh=self.device_mesh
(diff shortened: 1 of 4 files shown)


verl/workers/engine/fsdp/transformer_impl.py
```diff
@@ -207,20 +207,26 @@ def _build_module(self):
         torch_dtype = PrecisionType.to_dtype(torch_dtype)
 
         init_context = get_init_weight_context_manager(
-            use_meta_tensor=not self.model_config.hf_config.tie_word_embeddings, mesh=self.device_mesh
+            use_meta_tensor=not self.model_config.hf_config.tie_word_embeddings,
+            mesh=None if self.engine_config.strategy == "fsdp2" else self.device_mesh,
```
vermouth1992, line 211: Do we still need mesh info when we load from config?

EduardDurech: @vermouth1992 you mean like if it's safe to set to None? init-context FSDP2 since we load global rank0 and broadcast later don't need to CPU init each rank mesh

PR #4653, opened 2025-12-23
[docker] feat: update stable image to vllm==0.12.0, sglang==0.5.6

Begunner:
### What does this PR do?

1. Update dockerfiles of vllm/sglang stable images, vllm from v0.11.0 to v0.12.0 and sglang from v0.5.5 to v0.5.6.
2. Use nvidia/cuda:12.9.1-devel-ubuntu22.04 as base image for vllm.
3. Add hacking code for updating megatron from 0.14.0 to 0.15.0.


verl/workers/engine/megatron/__init__.py
```diff
@@ -12,6 +12,19 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from .transformer_impl import MegatronEngine, MegatronEngineWithLMHead
+# Avoid cpu worker trigger cuda jit error
```
wuxibin89: We should not import megatron in CPU only actor?

wuxibin89 (approved):

PR #4661, opened 2025-12-24
[misc] feat: `.git-blame-ignore-revs` for large but non-informative commits

tongyx361:
### What does this PR do?

This PR adds `.git-blame-ignore-revs` for ignoring large but non-informative commits.


HollowMan6:
This PR breaks git blame now as `.git-blame-ignore-revs` should use full git commit SHA. I opened #4674 to fix this issue.

PR #4683, opened 2025-12-26
[fsdp] feat: add validate process on trainer node when use_trainer_do_validate=True

chenjiaoAngel:
### What does this PR do?

User Trainer node to do validate process when run mode on fully-async, It can save time for validate computing and reduce perf/time_of_step peak
- add new use_trainer_do_validate on fully_async async_training config to decide whether using trainer node to do validate process
- use_trainer_do_validate: default is false
- It can improve performance of validate, such as in  `dapo_7b_math_fsdp2_8_8.sh`, it can improve about 1X speed

image
image

-  optimized the `process_validation_metrics()` on `_validate()` process, when input datasets len=1444, it latency reduce from 150+s to 40+s

image
image

### Checklist Before Starting

  - `{modules}` include `fsdp`, `megatron`, `sglang`, `vllm`, `rollout`, `trainer`, `ci`, `training_utils`, `recipe`, `hardware`, `deployment`, `ray`, `worker`, `single_controller`, `misc`, `perf`, `model`, `algo`, `env`, `tool`, `ckpt`, `doc`, `data`, `cfg`, `reward`

### Test

- on test scripts such as `dapo_7b_math_fsdp2_8_8.sh` add `async_training.use_trainer_do_validate=True` command to do compute
- the result of this function on Qwen2.5-Math-7B model
- the baseline scripts is `dapo_7b_math_fsdp2_8_8.sh` 
- the optimized scripts is  `dapo_7b_math_fsdp2_8_8.sh` +`async_training.use_trainer_do_validate=True` 
- the acc and perfomance is below:
image
image

   - green: optimized case (`async_training.use_trainer_do_validate=True` )
   - gray: baseline case (`async_training.use_trainer_do_validate=False` )

### API and Usage Example

```python
async_training.use_trainer_do_validate=True \
```

### Design & Code Changes

> [!IMPORTANT]

### DIFF
--- a/verl/utils/dataset/rl_dataset.py
+++ b/verl/utils/dataset/rl_dataset.py
@@ -390,7 +390,6 @@ class RLHFDataset(Dataset):
         images, videos = process_vision_info(messages, image_patch_size=image_patch_size, return_video_metadata=True)
         return images, videos
 
-
     def split(self, num_splits: int):
         """
         split the dataset into num_splits sub-datasets


wuxibin89:
Please format code according to: https://github.com/volcengine/verl/blob/main/CONTRIBUTING.md

chenjiaoAngel:
> Please format code according to: https://github.com/volcengine/verl/blob/main/CONTRIBUTING.md

已完成～

verl/utils/dataset/rl_dataset.py
```diff
@@ -390,6 +390,57 @@ async def process_vision_info(
         images, videos = process_vision_info(messages, image_patch_size=image_patch_size, return_video_metadata=True)
         return images, videos
 
+    def split(self, num_splits: int):
```
wuxibin89, line 393: Why we need split dataset?

chenjiaoAngel: > Why we need split dataset?

We need to split dataset to rollout and trainer node according to ratio of trainer:rollout

wuxibin89: Is it better to split validate files directly instead of split RLDataset?

chenjiaoAngel: > Is it better to split validate files directly instead of split RLDataset?

split validate file this way may be study, i know the easy way to split datasets; I will check this way whether it easy to work

chenjiaoAngel: If we split the data directly from the file, we need to save two temporary files and then use rl_dataset to get the data class; this seems to involve redundant read/write operations. If we split the data directly from the dataset object, my understanding is that we only need one round of I/O read/write, and there is no need to save intermediate files.

PR #4696, opened 2025-12-27
fix(fsdp): handle param offloading in get_actor_weights_info

### LINKED ISSUE #4657: [BUG] AssertionError: Expects tensor to be on the compute device cuda:0, was on cpu
### System Info

8 gpus * 3 nodes, 2 nodes for actor, 1 node for rollout
torch 2.7.1
cuda 12.2

### Information

- [x] My own modified scripts

### Tasks

## Question as follows：

Using fully async mode , when I set 
```
actor_rollout_ref.actor.strategy=fsdp 
critic.strategy=fsdp
actor_rollout_ref.actor.fsdp_config.param_offload=False
actor_rollout_ref.actor.fsdp_config.optimizer_offload=False
```
, executing custom script（modified from script  dapo_7b_math_fsdp2_8_8.sh, results in error: AssertionError: Expects tensor to be on the compute device cuda:0, was on cpu

## Reasons as follows:
FSDP's `state_dict()` requires parameters to be on the GPU,
but the current parameters are on the CPU (because offloading is enabled).

## How to solve:
=== fully_async_policy/fsdp_workers.py：
```
@register(dispatch_mode=Dispatch.ONE_TO_ALL)
def get_actor_weights_info(self):
    assert self._is_actor
    if hasattr(self, "_weights_info"):
        return self._weights_info
    
    # add：If offloaded, load the params to the GPU firstly
    if self._is_offload_param:
        load_fsdp_model_to_gpu(self.actor_module_fsdp)
    
    if fsdp_version(self.actor_module_fsdp) == 1:
        from torch.distributed.fsdp.api import ShardedStateDictConfig, StateDictType
        FSDP.set_state_dict_type(
            self.actor_module_fsdp,
            state_dict_type=StateDictType.SHARDED_STATE_DICT,
            state_dict_config=ShardedStateDictConfig(),
        )
    params = self._get_actor_params()
    ret = []
    for key, tensor in params.items():
        ret.append((key, tensor.size(), tensor.dtype))
    self._weights_info = ret
    
    # add：offload to the CPU after use
    if self._is_offload_param:
        offload_fsdp_model_to_cpu(self.actor_module_fsdp)
    
    return ret
```

### Expected behavior

when using FSDP and params offload in fully async mode, model parameters  load and offload  from the CPU normally

yurekami:
## Summary
- Fix AssertionError when using FSDP with parameter offloading in fully async mode
- The error message: "Expects tensor to be on the compute device cuda:0, was on cpu"

## Root Cause
`get_actor_weights_info()` in `recipe/fully_async_policy/fsdp_workers.py` calls `state_dict()` which requires parameters to be on GPU. When `param_offload` is enabled, parameters are offloaded to CPU, causing the assertion error.

## Changes
Added GPU load/offload logic to `get_actor_weights_info()` method:
- Load model to GPU before calling `_get_actor_params()` (which uses `state_dict()`)
- Offload model back to CPU after getting the weights info

This follows the same pattern already used in other methods like `sync_rollout_weights()` and `sync_rollout_weights_by_checkpoint()`.

## Before
```python
def get_actor_weights_info(self):
    # ...
    params = self._get_actor_params()  # Fails if params are on CPU
    # ...
```

## After
```python
def get_actor_weights_info(self):
    # ...
    if self._is_offload_param:
        load_fsdp_model_to_gpu(self.actor_module_fsdp)
    params = self._get_actor_params()
    # ...
    if self._is_offload_param:
        offload_fsdp_model_to_cpu(self.actor_module_fsdp)
    return ret
```

Fixes #4657

## Test plan
- [ ] Test FSDP with param_offload=True in fully async mode
- [ ] Verify no AssertionError occurs when calling get_actor_weights_info()

### DIFF
--- a/recipe/fully_async_policy/fsdp_workers.py
+++ b/recipe/fully_async_policy/fsdp_workers.py
@@ -218,11 +218,17 @@ class DetachActorWorker(DetachNcclSync):
                 state_dict_type=StateDictType.SHARDED_STATE_DICT,
                 state_dict_config=ShardedStateDictConfig(),
             )
+        # Load model to GPU before state_dict() if offloading is enabled
+        if self._is_offload_param:
+            load_fsdp_model_to_gpu(self.actor_module_fsdp)
         params = self._get_actor_params()
         ret = []
         for key, tensor in params.items():
             ret.append((key, tensor.size(), tensor.dtype))
         self._weights_info = ret
+        # Offload model back to CPU after getting weights info
+        if self._is_offload_param:
+            offload_fsdp_model_to_cpu(self.actor_module_fsdp)
         return ret
 
     @register(dispatch_mode=Dispatch.ONE_TO_ALL)


vermouth1992:
In fully async mode, do we really need load/offload?

yurekami:
Closing. This was opened twice (#4696 and #4726), both have drifted out of date with the recipes, and I never answered the question about whether the fully async path needs load/offload at all. That should be settled on an issue before a patch. Sorry for the noise.

PR #4689, opened 2025-12-27
Fix Profiler class using wrong config attribute names

yurekami:
## Summary
Fix AttributeError in the `Profiler` class when profiling is enabled. The class was referencing non-existent attributes on `ProfilerConfig`.

## Problem
The `Profiler` class in `verl/utils/profiler/profile.py` uses:
- `self.config.profile_ranks` - but `ProfilerConfig` has `ranks`, not `profile_ranks`
- `self.config.step_start` / `self.config.step_end` - but these are in `TorchProfilerToolConfig`, not `ProfilerConfig`

This causes:
```
AttributeError: 'ProfilerConfig' object has no attribute 'profile_ranks'
```

## Changes
- Line 59: `self.config.profile_ranks` → `self.config.ranks`
- Line 79-81: `self.config.profile_ranks` → `self.config.ranks` (also changed None check to truthiness check for proper empty list handling)
- Line 109: `self.config.step_start` → `self.tool_config.step_start` and `self.config.step_end` → `self.tool_config.step_end`

## Test plan
- [ ] Enable profiling with `tool: torch` and verify no AttributeError
- [ ] Check that profiling respects the configured `ranks` list
- [ ] Verify trace files are saved with correct step_start/step_end in filename

Fixes #4663

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #3985, opened 2025-11-02
No profile file generation for torch


jasonneverstop:
### System Info

image: app-verl0.6-transformers4.56.1-sglang0.5.2-mcore0.13.0-te2.2
system: 8* B200
backend：sglang
attention: flashinfer

### Information

- [x] The official example scripts
- [ ] My own modified scripts

### Tasks

- [x] An officially supported task in the `examples` folder (such as GLUE/SQuAD, ...)
- [ ] My own task or dataset (give details below)

### Reproduction

1. enable profile， profile_tool = torch，all_ranks= true，step_start=5，step_end=8，save_path=outputs/profileoutputs/profile
2. run  verl/examples/grpo_trainer/run_qwen3-8b.sh
3. change the config.profile_ranks in the verl/utils/profiler/profile.py file to config.ranks.
4. change  the  def start(self)  in the  verl/utils/profiler/profile.py file to  def start(self, **kwargs)
5. no profile file in outputs/profile 
(note: If steps 4 and 5 are not carried out, the program will report an error.)

### Expected behavior

Generate the profile file in the "outputs/profile" directory

Kirrito-k423:
定时 issue 清理检查结果：本 issue 最近一次非当前运行账号（Kirrito-k423）的公开回复或公开引用是 2025-12-28T20:13:16Z，类型为 cross-referenced，账号为 AkiRusProd，证据链接：https://github.com/verl-project/verl/pull/4707。本轮同时检查了 issue comments（数量 0）和 timeline（事件数量 3；来源：https://github.com/verl-project/verl/issues/3985#issuecomment-timeline）。截至 2026-07-22，该最近非本人公开互动已经超过 6 个月。

因此按定时清理标准关闭。若作者或社区成员仍能提供新的复现、证据或需求背景，欢迎继续评论请求重开，或创建新的 issue 并链接到这里。

Scheduled issue cleanup check: the latest public reply or public reference not made by the current automation account (Kirrito-k423) was at 2025-12-28T20:13:16Z (kind: cross-referenced, actor: AkiRusProd; evidence: https://github.com/verl-project/verl/pull/4707). This run checked issue comments (0) and timeline events (3; source: https://github.com/verl-project/verl/issues/3985#issuecomment-timeline). As of 2026-07-22, that latest non-self public interaction is older than 6 months.

[...]

PR #4707, opened 2025-12-28
[perf] fix: correct torch profiler logic for distributed trace collection

AkiRusProd:
### What does this PR do?

This PR fixes the logic of `torch_profiler` within the `Verl` to enable correct trace collection in distributed environments. Previously, the profiler logic prevented valid trace generation or caused race conditions during file saving.

Key changes include:
1.  **Configuration Validation**: Fixed logic in `_validate` to correctly handle `all_ranks` vs specific `ranks`. It now raises a clear error if both are set and correctly defaults to Rank 0 if nothing is specified.
2.  **Distributed Synchronization**: Added `torch.distributed.barrier()` in the `save()` method. This ensures that:
    *   Rank 0 creates the directory before other ranks attempt to write.
    *   All ranks finish writing their traces before the process continues or terminates.
3.  **Variable Fixes**: Corrected usage of `self.tool_config.step_start/end` inside the `save` filename generation (previously it might have referenced incorrect config attributes).

### Test

Verified by running a training job with profiling enabled.
**Result:** Chrome trace `.json` files are now correctly generated and saved to the specified output directory for the requested ranks.

### API and Usage Example

No API signature changes, but the configuration logic is now stricter/safer.

**Config Example (`config.yaml` or dict):**

```yaml
actor_rollout_ref:
  actor:
      profiler:
        tool: torch
        enable: True
        ranks: [0, 8, 16, 32] # for example
        # all_ranks: True
        save_path: "/<your_path>" # profiler saving path
        tool_config:
          torch:
            _target_: verl.utils.profiler.config.TorchProfilerToolConfig
            step_start: 1
            step_end: 3

```

**Note:** Setting both `all_ranks=True` and `ranks=[...]` will now raise a `ValueError` to avoid ambiguity.

### Design & Code Changes

*   **`Profiler._validate`**:
    *   Added checks to prevent setting `all_ranks` and `ranks` simultaneously.
    *   Correctly populates `self.profile_ranks` based on world size when `all_ranks=True`.
    *   Added validation to ensure requested ranks exist within `torch.distributed.get_world_size()`.
*   **`Profiler.save`**:
    *   Only `rank 0` calls `os.makedirs`.
    *   Added `torch.distributed.barrier()` before and after saving to synchronize workers.
    *   Fixed filename generation to use `self.tool_config` parameters consistently.

Related Issues:
https://github.com/volcengine/verl/issues/3985

Related PRs:
https://github.com/volcengine/verl/pull/4689

This PR supersedes #4689.
While #4689 fixes the immediate `AttributeError` regarding `tool_config`, this PR provides a complete fix for the profiler in distributed settings:
1.  **Thread Safety**: Adds `torch.distributed.barrier()` to prevent race conditions during file saving (critical for multi-rank training).
2.  **Logic Fixes**: Correctly implements the logic for `all_ranks` vs `ranks`, resolving the ambiguity that existed previously.
3.  **IO Safety**: Ensures the output directory is created safely by rank 0 before other ranks attempt to write.

### DIFF
--- a/verl/utils/profiler/profile.py
+++ b/verl/utils/profiler/profile.py
@@ -79,14 +79,14 @@ class Profiler:
     def _validate(self):
         if not self.enable:
             return
-            
+
         if self.config.all_ranks and self.config.ranks:
             raise ValueError(
                 f"[Profiler] Configuration Error: Ambiguous setup. "
                 f"'all_ranks' is set to True, but specific 'ranks' list ({self.config.ranks}) is also provided. "
                 "Please unset 'ranks' or set 'all_ranks' to False."
             )
-        
+
         world_size = torch.distributed.get_world_size()
 
         if self.config.all_ranks:
@@ -135,7 +135,9 @@ class Profiler:
         torch.distributed.barrier()
 
         if self.prof is not None and not self.saved:
-            save_file_name = f"/prof_start_{self.tool_config.step_start}_end_{self.tool_config.step_end}_rank_{self.rank}.json"
+            save_file_name = (
+                f"/prof_start_{self.tool_config.step_start}_end_{self.tool_config.step_end}_rank_{self.rank}.json"
+            )
             print(f"[Profiler] Saving trace to {self.config.save_path + save_file_name}")
             self.prof.export_chrome_trace(self.config.save_path + save_file_name)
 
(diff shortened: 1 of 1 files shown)


wuxibin89 (dismissed):

wuxibin89:
@AkiRusProd Please format code according to: https://github.com/volcengine/verl/blob/main/CONTRIBUTING.md#code-linting-and-formatting

AkiRusProd:
Thanks for pointing that out. I have applied the pre-commit hooks and formatted the code accordingly. The latest commit should pass the linting checks now.

PR #4711, opened 2025-12-29
[training_utils] fix: RM extra scaling in KL/PG losses

JacobHelwig:
### What does this PR do?

The KL/ PG losses currently logged are scaled by the number of micro-batches twice. The result is that the logged metrics represent the mean value across micro-batches **scaled by the number of micro-batches**. This PR only scales once so that the logged metrics represent the mean across micro-batches with no extra scaling. 

First scaling:

https://github.com/volcengine/verl/blob/cd4072daad2652794ecff0b5816a05afedff8608/verl/workers/actor/dp_actor.py#L533

Second scaling:

https://github.com/volcengine/verl/blob/cd4072daad2652794ecff0b5816a05afedff8608/verl/utils/metric/utils.py#L53

### Test

On `main`, decreasing micro-batch size from 8->2 decreases logged loss by a factor of 4:

image

Decreasing micro-batch size on this branch does not effect metric magnitude:

image

```bash
    data.dataloader_num_workers=0 \
    data.return_full_prompt=True \
    data.train_files=$SAVE_PATH/gsm8k/train.parquet \
    data.val_files=$SAVE_PATH/gsm8k/test.parquet \
    data.train_batch_size=8 \
    data.max_prompt_length=512 \
    data.truncation='error' \
    actor_rollout_ref.model.path=Qwen/Qwen2.5-0.5B-Instruct \
    +actor_rollout_ref.ref.model.path=Qwen/Qwen2.5-3B-Instruct \
    actor_rollout_ref.actor.ppo_mini_batch_size=8 \
    actor_rollout_ref.actor.ppo_micro_batch_size_per_gpu=2 \
    actor_rollout_ref.actor.kl_loss_coef=10 \
    actor_rollout_ref.rollout.log_prob_micro_batch_size_per_gpu=8 \
    actor_rollout_ref.rollout.n=5 \
    actor_rollout_ref.ref.log_prob_micro_batch_size_per_gpu=8 \
    trainer.critic_warmup=0 \
    trainer.project_name='verl_fix_metrics' \
    trainer.experiment_name='NEW/ppo_micro_batch_size_per_gpu2' \
    trainer.n_gpus_per_node=1 \
    trainer.nnodes=1 \
    trainer.save_freq=20 \
    trainer.test_freq=5 \
    trainer.resume_mode="disable" \
    trainer.total_epochs=15 \
    actor_rollout_ref.actor.use_torch_compile=False \
    actor_rollout_ref.actor.fsdp_config.use_torch_compile=False \
    trainer.val_before_train=False \
    actor_rollout_ref.rollout.enforce_eager=True \
    actor_rollout_ref.ref.fsdp_config.use_torch_compile=False
```

### Design & Code Changes

RM scaling in `dp_actor`

### DIFF
--- a/verl/workers/actor/dp_actor.py
+++ b/verl/workers/actor/dp_actor.py
@@ -431,7 +431,10 @@ class DataParallelPPOActor(BasePPOActor):
 
         on_policy = len(mini_batches) == 1 and self.config.ppo_epochs == 1
 
-        metrics = {}
+        metrics = {
+            "actor/pg_loss": 0.0,
+            "actor/kl_loss": 0.0,
+        }
         for _ in range(self.config.ppo_epochs):
             for batch_idx, mini_batch in enumerate(mini_batches):
                 if self.config.use_dynamic_bsz:
@@ -530,7 +533,7 @@ class DataParallelPPOActor(BasePPOActor):
                         kl_loss = agg_loss(loss_mat=kld, loss_mask=response_mask, loss_agg_mode=loss_agg_mode)
 
                         policy_loss = policy_loss + kl_loss * self.config.kl_loss_coef
-                        micro_batch_metrics["actor/kl_loss"] = kl_loss.detach().item()
+                        metrics["actor/kl_loss"] += kl_loss.detach().item() * loss_scale_factor
                         micro_batch_metrics["actor/kl_coef"] = self.config.kl_loss_coef
 
                     if self.config.use_dynamic_bsz:
@@ -543,7 +546,7 @@ class DataParallelPPOActor(BasePPOActor):
                     else:
                         loss.backward()
 
-                    micro_batch_metrics["actor/pg_loss"] = pg_loss.detach().item()
+                    metrics["actor/pg_loss"] += pg_loss.detach().item() * loss_scale_factor
                     append_to_dict(metrics, micro_batch_metrics)
 
                 grad_norm = self._optimizer_step()
--- a/verl/workers/critic/dp_critic.py
+++ b/verl/workers/critic/dp_critic.py
@@ -192,7 +192,9 @@ class DataParallelPPOCritic(BasePPOCritic):
     def update_critic(self, data: DataProto):
         # make sure we are in training mode
         self.critic_module.train()
-        metrics = {}
+        metrics = {
+            "critic/vf_loss": 0.0,
+        }
 
         select_keys = ["input_ids", "responses", "response_mask", "attention_mask", "position_ids", "values", "returns"]
         has_multi_modal_inputs = "multi_modal_inputs" in data.non_tensor_batch.keys()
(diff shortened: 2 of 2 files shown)


yyDing1:
Get it. This is indeed a mistake. But there still exists bias in the fixed version.
- It's ok for "use_dynamic_bsz=False", as the loss scaling factor is fixed across multiple micro-batches (i.e., 1 / grad_acc)
- It will be wrong for "use_dynamic_bsz=True", as the micro-batch size may differ.

I consider the following fix:
- keep the micro-batch-level log, i.e., `micro_batch_metrics["actor/kl_loss"] = kl_loss.detach().item() * loss_scale_factor`
- In the process of aggregating the results of the micro-batch log, use `sum` instead of `mean`, as it has been scaled.

JacobHelwig:
> Get it. This is indeed a mistake. But there still exists bias in the fixed version.
> 
> * It's ok for "use_dynamic_bsz=False", as the loss scaling factor is fixed across multiple micro-batches (i.e., 1 / grad_acc)
> * It will be wrong for "use_dynamic_bsz=True", as the micro-batch size may differ.
> 
> I consider the following fix:
> 
> * keep the micro-batch-level log, i.e., `micro_batch_metrics["actor/kl_loss"] = kl_loss.detach().item() * loss_scale_factor`
> * In the process of aggregating the results of the micro-batch log, use `sum` instead of `mean`, as it has been scaled.

Good catch @yyDing1 . Since the `np.mean` in `verl.utils.metric.utils.reduce_metrics` is still needed for some of the metrics returned from the update (e.g., `pg_metrics`), I just handled the aggregation directly in the update function: https://github.com/volcengine/verl/pull/4711/commits/e43dadb98ce04ae4569538209ca6b55aeb1f4752

If we want to still do the aggregation in `reduce_metrics` instead of in the update, I see 2 options that keep the `np.mean` in `reduce_metrics`:
- Option 1: Add a "+" to the end of the names of metrics that need to be summed, rm it after summing in `reduce_metrics`:
```python
new_metrics = dict()
for key, val in metrics.items():
    if "max" in key:
        new_metrics[key] = np.max(val)
    elif "min" in key:
        new_metrics[key] = np.min(val)
    elif key.endswith("+"):
        new_metrics[key[:-1]] = np.sum(val)
    else:
        new_metrics[key] = np.mean(val)
return new_metrics
```

- Option 2: Don't change reduce metrics; in the update function, multiply each element in KL/PG loss lists by number of grad accumulation steps so the mean in `reduce_metrics` is doing a sum:

```python
self.actor_optimizer.zero_grad()
metrics["actor/pg_loss"] = [loss * len(metrics["actor/pg_loss"]) for loss in metrics["actor/pg_loss"]]
metrics["actor/kl_loss"] = [loss * len(metrics["actor/kl_loss"]) for loss in metrics["actor/kl_loss"]]
return metrics
```

WDYT?

yyDing1:
Yes, agree with you that we should make a minimal change while preserving the existing logic for computing the mean of the values.

JacobHelwig:
> Yes, agree with you that we should make a minimal change while preserving the existing logic for computing the mean of the values.

Agreed, since option 1 would require a modification to `reduce_metrics`, maybe the current approach or option 2 are best for keeping the changes isolated. IMO, the current approach might be a bit more readable/less hacky than option 2, but I'm good on either. WDYT?

yyDing1:
I think your current implementation looks better. If you have no concerns, we can merge it.

yyDing1 (approved):

Issue #4765, opened 2026-01-01
求大佬解答，当前SFT支持流式数据读取吗？


Marco692:
SFT任务的训练数据量非常大，我发现目前SFTDataset是每张卡都重复加载完整数据，请问有什么优化方案吗？或者当前SFT训练框架支持流式数据读取吗？

vermouth1992:
You can override the SFT dataset using other frameworks such as VeOmni

kasteric:
流式数据是啥

Kirrito-k423:
定时清理检查结论：这个 issue 最近一次非当前运行账号（Kirrito-k423）的公开回复或引用是 2026-01-19T09:47:58Z（actor: kasteric, kind: comment），截至 2026-07-28T01:03:23Z 已超过 6 个月。

已检查的公开证据：issue 页面 https://github.com/verl-project/verl/issues/4765、comments 数量 2、timeline 公开互动数量 3，以及 timeline API https://api.github.com/repos/verl-project/verl/issues/4765/timeline。根据定时清理标准，先关闭该 issue。若作者或社区成员仍能提供新的复现、证据或需求背景，请继续评论请求重开，或创建新的 issue。

---

Scheduled cleanup result: the latest public reply or reference from an account other than the current runner (Kirrito-k423) was 2026-01-19T09:47:58Z (actor: kasteric, kind: comment), which is more than 6 months before 2026-07-28T01:03:23Z.

Checked public evidence: issue page https://github.com/verl-project/verl/issues/4765, comments count 2, timeline public-interaction count 3, and timeline API https://api.github.com/repos/verl-project/verl/issues/4765/timeline. Under the scheduled cleanup policy, I am closing this issue. If the author or community can still provide a fresh reproduction, evidence, or requirement context, please comment to request reopening or open a new issue.

This response was generated by a scheduled Codex GPT5.5 issue processing task.

PR #4764, opened 2026-01-01
[trainer] feat: enable ray-based sft trainer on ascend npu

ji-huazhong:
### What does this PR do?

This PR implements Ray-based SFT support for the Ascend NPU platform and includes a minor interface adjustment: the function `auto_set_ascend_device_name` is renamed to `auto_set_device`. The revised interface is designed to be generic and can be seamlessly integrated with other hardware accelerators.

### Checklist Before Starting

- [ ] Search for similar PRs. Paste at least one query link here: ...
- [ ] Format the PR title as `[{modules}] {type}: {description}` (This will be checked by the CI)
  - `{modules}` include `fsdp`, `megatron`, `sglang`, `vllm`, `rollout`, `trainer`, `ci`, `training_utils`, `recipe`, `hardware`, `deployment`, `ray`, `worker`, `single_controller`, `misc`, `perf`, `model`, `algo`, `env`, `tool`, `ckpt`, `doc`, `data`, `cfg`, `reward`
  - If this PR involves multiple modules, separate them with `,` like `[megatron, fsdp, doc]`
  - `{type}` is in `feat`, `fix`, `refactor`, `chore`, `test`
  - If this PR breaks any API (CLI arguments, config, function signature, etc.), add `[BREAKING]` to the beginning of the title.
  - Example: `[BREAKING][fsdp, megatron] feat: dynamic batching`

### Test

> For changes that can not be tested by CI (e.g., algorithm implementation, new model support), validate by experiment(s) and show results like training curve plots, evaluation results, etc.

### API and Usage Example

> Demonstrate how the API changes if any, and provide usage example(s) if possible.

```python
# Add code snippet or script demonstrating how to use this
```

### Design & Code Changes

> Demonstrate the high-level design if this PR is complex, and list the specific changes.

### Checklist Before Submitting

> [!IMPORTANT]
> Please check all the following items before requesting a review, otherwise the reviewer might deprioritize this PR for review.

- [ ] Read the [Contribute Guide](https://github.com/volcengine/verl/blob/main/CONTRIBUTING.md).
- [ ] Apply [pre-commit checks](https://github.com/volcengine/verl/blob/main/CONTRIBUTING.md#code-linting-and-formatting): `pre-commit install && pre-commit run --all-files --show-diff-on-failure --color=always`
- [ ] Add / Update [the documentation](https://github.com/volcengine/verl/tree/main/docs).
[...]

Issue #4329, opened 2025-11-28
[Feature Request] Add Tensor Parallel Support to FSDP SFT Trainer


lllsy12138:
Currently, the FSDP SFT Trainer (`verl/verl/trainer/fsdp_sft_trainer.py`) supports:
- Fully Sharded Data Parallel (FSDP/FSDP2)
- Ulysses Sequence Parallel

However, there is no support for Tensor Parallel (TP), which could be beneficial for training very large models or reducing per-GPU memory footprint further.

Is there a plan to add tensor parallelism support for SFT?

wuxibin89:
We have support both fsdp and megatron sft trainer:
https://github.com/volcengine/verl/blob/main/examples/sft/vlm/run_qwen3_vl_2b.sh

PR #4759, opened 2025-12-31
[data] feat: major refactor RLHFDataset for multi-modal data

wuxibin89:
### What does this PR do?

image

Refactor RLHFDataset for multi-modal data, `__getitem__`  return text messages and let agent_workers fetch image and video from storage.

|  | Who fetch image/video? | Pros and cons |
|------|------|------|
| Before | RLHFDataset dataLoader | - big communication overhead between single controller and agent_loop workers<br> - single controller cpu/memory bottleneck in large scale dataset |
| After | AgentLoopWorker | - pass text messages from single controller, no communication overhead<br> - agent_loop workers run in each node, avoid single-point bottleneck |

### Accuracy verify
Megatron Qwen3-VL-30B-A3B-Instruct geo3k:
image

#### TODO
- [ ] support video input for sglang
- [ ] deprecate DataProto by TensorDict

PR #4766, opened 2026-01-02
[trainer,data,algo] feat: Unify SFT and PPO Pipelines

JacobHelwig:
### What does this PR do?

Unifies SFT/RL training pipeline. Since SFT is a special case of off-policy RL, unification is achieved by merging the SFT pipeline into the PPO pipeline. This allows:
1. Reduced maintenance burden and feature drift between SFT/RL pipelines. New features that benefit both pipelines only need to be implemented once. Examples of shared features are: 
- Iterable dataset: https://github.com/volcengine/verl/issues/4765
- NPU support: https://github.com/volcengine/verl/pull/4764
- TP: https://github.com/volcengine/verl/issues/4329
2. Reduced code duplication. There are currently 3 SFT trainers ([fsdp_sft_trainer.py](https://github.com/volcengine/verl/blob/5d0eac0613f8561e77052c941d87d2dee693a0bd/verl/trainer/fsdp_sft_trainer.py), [sft_trainer.py](https://github.com/volcengine/verl/blob/5d0eac0613f8561e77052c941d87d2dee693a0bd/verl/trainer/sft_trainer.py), [sft_trainer_ray.py](https://github.com/volcengine/verl/blob/5d0eac0613f8561e77052c941d87d2dee693a0bd/verl/trainer/sft_trainer_ray.py)), each with overlap to the PPO trainer. The SFT dataset class also has overlap with the RL dataset class.
3. Rollouts during validation for SFT training. None of the SFT trainers currently support monitoring performance on downstream RL tasks during SFT.

### Test

SFT Qwen2.5-0.5B-Instruct on GSM8K (`examples/sft/gsm8k/run_qwen_05.sh`). 

CE Loss:

image

Pass@1:

image

### Design & Code Changes

SFT is a special case of off-policy RL: the generated sequences returned by `AgentLoopManager` are the SFT labels and the policy loss in `ppo.core_algos` is the NLL of the generated sequences. 

### Roadmap

1. This PR only supports SFT in `workers.actor.dp_actor`; extend to `megatron_actor`
2. Add CE/perplexity of validation SFT sequences as a val metric
3. The RL dataset raises not implemented for SFT w/ multimodal inputs, as it hasn't been tested yet
4. Extend dataset to tokenize SFT demonstrations of tool use with tools that potentially return non-text objects, eg images

### DIFF
--- a/examples/sft/gsm8k/run_qwen_05.sh
+++ b/examples/sft/gsm8k/run_qwen_05.sh
@@ -1,5 +1,5 @@
-export CUDA_VISIBLE_DEVICES=0,1
 export CUDA_VISIBLE_DEVICES=2,3
+export CUDA_VISIBLE_DEVICES=0,1
 export DATA_DIR=$PWD/data
 export HF_HOME=$DATA_DIR
 export PATH="$CONDA_PREFIX/bin:$PATH"
@@ -13,8 +13,8 @@ gsm8k_test_path=$DATA_DIR/gsm8k/test.parquet
 train_files="['$gsm8k_train_path']"
 test_files="['$gsm8k_test_path']"
 
-use_legacy_worker_impl=disable
 use_legacy_worker_impl=auto
+use_legacy_worker_impl=disable
 
 python3 -m verl.trainer.main_ppo \
     algorithm.adv_estimator=grpo \
@@ -39,7 +39,7 @@ python3 -m verl.trainer.main_ppo \
     actor_rollout_ref.rollout.name=vllm \
     actor_rollout_ref.rollout.gpu_memory_utilization=0.6 \
     actor_rollout_ref.rollout.n=1 \
-    trainer.logger='["console","wandb"]' \
+    trainer.logger='["console"]' \
     trainer.project_name='verl_sft_example_gsm8k' \
     trainer.experiment_name="qwen_05_use_legacy_worker_impl_$use_legacy_worker_impl" \
     trainer.n_gpus_per_node=2 \


wuxibin89:
@JacobHelwig Both SFT and RL have two implementation: legacy and new engine. We will switch default to new engine in v0.7, all legacy engine will be deprecated in v0.8 and removed in v0.9.

| File | Engine | Trainer | Backend | Explaination |
| ----- | ----- | ----- | ----- | ----- |
| fsdp_sft_trainer.py | legacy | SFT | fsdp |
| fsdp_workers.py | legacy | RL | fsdp |
| megatron_workers.py | legacy | RL | megatron |
| sft_trainer.py | new | SFT | fsdp/megatron/veomni/... | torchrun spmd
| sft_trainer_ray.py | new | SFT | fsdp/megatron/veomni/... | ray single controller
| engine_workers.py | new | RL | fsdp/megatron/veomni/... |


To this PR, I'm not tend to unify SFT to RL pipeline, since they have different Dataset, computation DAG, loss, etc. In new implementation, both SFT and RL pipeline reuse model engine with much less redundant code, please check `sft_trainer.py` and `engine_workers.py`.

JacobHelwig:
Hi @wuxibin89, thanks for your reply and the pointer towards the new engine. 

> To this PR, I'm not tend to unify SFT to RL pipeline, since they have different Dataset, computation DAG, loss, etc.

I hadn’t seen the update to your message until now, so I updated this branch to work with the new engine and include the refactor from #4759.

IMO the unification fits cleanly into the existing framework just by registering a SFT loop to the `_agent_loop_registry` and a SFT loss to the `POLICY_LOSS_REGISTRY`, and it could save some work down the road. If there are any missing elements, we could talk about how they might be added. 

However, if the code owners still would prefer to keep the two pipelines separate, we can close this PR.

wuxibin89:
Close as keep SFT and RL trainer separate for now.

PR #4809, opened 2026-01-06
Fix: use prime reward manager for GSPO script

aub123:
## Description

This PR fixes the configuration in the GSPO example script. Previously, it used `reward_manager=dapo` with overlong penalties, which seems to be a leftover from DAPO experiments.

I have updated it to use `reward_manager=prime` (standard for Math tasks) and removed the DAPO-specific `overlong_buffer` configurations to ensure the script runs standard GSPO logic correctly.

## Changes
- Changed `reward_manager` from `dapo` to the default setting.
- Removed `overlong_buffer_cfg` parameters.

### DIFF
--- a/examples/gspo_trainer/run_qwen30b_gspo.sh
+++ b/examples/gspo_trainer/run_qwen30b_gspo.sh
@@ -38,9 +38,6 @@ critic_model_path=$actor_model_path
 
 max_prompt_length=$((1024 * 2))
 max_response_length=$((1024 * 8))
-enable_overlong_buffer=True
-overlong_buffer_len=$((1024 * 4))
-overlong_penalty_factor=1.0
 
 train_batch_size=256
 ppo_mini_batch_size=32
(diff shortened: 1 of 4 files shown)


yyDing1:
Could you explain why `reward_manager=prime` is the standard for gspo?

Why not use `reward_manager=naive` or `reward_manager=dapo` with `enable_overlong_buffer=False` (which some of the scripts do).

aub123:
> Could you explain why `reward_manager=prime` is the standard for gspo?
> 
> Why not use `reward_manager=naive` or `reward_manager=dapo` with `enable_overlong_buffer=False` (which some of the scripts do).

Actually, in my original script, I removed reward_manager to ensure the GSPO recipe remained minimal. I've added prime back here specifically because gemini_code_review suggested it as the standard configuration.

yyDing1:
Gemini only provides suggested modifications, which may be incorrect. You need to at least ensure that the updated script is runnable. The Prime Reward Manager seems to be no longer fully runnable due to dependency issues.

PR #4799, opened 2026-01-05
[ckpt] fix: FSDP save ckpt after validation

wdl339:
### What does this PR do?

This PR fixes a bug in the `save_checkpoint` function for FSDPEngine.

In the original logic, if the model engine is used (`use_legacy_worker_impl=disable`), the `wake_up` function in `verl/workers/engine_workers.py` will be invoked during the rollout phase of each step, which will offload the model to CPU.

Under normal circumstances, the `compute_log_prob` function called during the training phase can load the model back to  GPU. However, the training process is not executed during the validation phase, leaving the model on the CPU. If a checkpoint is saved immediately after validation, it will trigger the following error: `AssertionError: Expects tensor to be on the compute device cuda:0, was on cpu.`

<details>
<summary>Details</summary>

Script:
```
set -x
python examples/data_preprocess/geo3k.py --local_dir ~/data/geo3k
python -m verl.trainer.main_ppo \
  algorithm.adv_estimator=grpo \
  data.train_files=$HOME/data/geo3k/train.parquet \
  data.val_files=$HOME/data/geo3k/test.parquet \
  data.train_batch_size=512 \
  data.max_prompt_length=1024 \
  data.max_response_length=2048 \
  data.filter_overlong_prompts=True \
  data.truncation='error' \
  data.image_key=images \
  actor_rollout_ref.model.path=Qwen/Qwen2.5-VL-3B-Instruct \
  actor_rollout_ref.actor.optim.lr=1e-6 \
  actor_rollout_ref.model.use_remove_padding=True \
  actor_rollout_ref.actor.ppo_mini_batch_size=128 \
  actor_rollout_ref.actor.ppo_micro_batch_size_per_gpu=4 \
  actor_rollout_ref.actor.use_kl_loss=True \
  actor_rollout_ref.actor.kl_loss_coef=0.01 \
  actor_rollout_ref.actor.kl_loss_type=low_var_kl \
  actor_rollout_ref.actor.entropy_coeff=0 \
  actor_rollout_ref.model.enable_gradient_checkpointing=True \
  actor_rollout_ref.actor.fsdp_config.param_offload=False \
  actor_rollout_ref.actor.fsdp_config.optimizer_offload=False \
  actor_rollout_ref.rollout.log_prob_micro_batch_size_per_gpu=16 \
  actor_rollout_ref.rollout.tensor_model_parallel_size=2 \
  actor_rollout_ref.rollout.name=vllm \
[...]

PR #4841, opened 2026-01-08
[ckpt] fix: Megatron save ckpt after validation

wdl339:
### What does this PR do?

This PR fixes a bug in the `save_checkpoint` function for MegatronEngine. https://github.com/volcengine/verl/pull/4799 is a similar PR, which modifies FSDPEngine.

In the original logic, if the model engine is used (`use_legacy_worker_impl=disable`), the `wake_up` function in `verl/workers/engine_workers.py` will be invoked during the rollout phase of each step, which will offload the model to CPU.

Under normal circumstances, the `compute_log_prob` function called during the training phase can load the model back to  GPU. However, the training process is not executed during the validation phase, leaving the model on the CPU. If a checkpoint is saved immediately after validation, it will trigger the following error: `AssertionError: Expects tensor to be on the compute device cuda:0, was on cpu.`

To fix this bug, this PR checks whether the model is located on the CPU before saving the checkpoint and loads it onto the GPU if that is the case.

### DIFF
--- a/verl/utils/megatron_utils.py
+++ b/verl/utils/megatron_utils.py
@@ -1229,27 +1229,20 @@ def mapping_string_to_attn_backend(args: dict) -> dict:
         args["attention_backend"] = AttnBackend[args["attention_backend"]]
     return args
 
+
 def get_megatron_module_device(models: list[Any]) -> str:
     if not models:
-        return 'cpu' 
+        return "cpu"
 
     model_chunk = models[0]
-    if isinstance(model_chunk, DDP):
-        if not model_chunk.buffers:
-            try:
-                return next(model_chunk.module.parameters()).device.type
-            except StopIteration:
-                return 'cpu'
-
-        buffer = model_chunk.buffers[0]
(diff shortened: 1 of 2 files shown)


wuxibin89:
@wdl339 Please format code according to: https://github.com/volcengine/verl/blob/main/CONTRIBUTING.md

verl/utils/megatron_utils.py
```diff
@@ -1228,3 +1228,28 @@ def mapping_string_to_attn_backend(args: dict) -> dict:
...
+            except StopIteration:
+                return 'cpu'
+
+        buffer = model_chunk.buffers[0]
+        if buffer.param_data.storage().size() == 0:
+            return 'cpu'
+        else:
+            return get_device_name()
+    else:
+        # for ref module
```
wuxibin89: ref model should not be save checkpoint?

wdl339: Sure, I'll revise it.

PR #4858, opened 2026-01-09
[rollout] fix: wrong display about Prometheus when using SGLang.

jsfanfanfan:
### What does this PR do?

> Fix the issue where vLLM is always displayed when using other rollout engines.


wuxibin89:
@jsfanfanfan Please format code according to: https://github.com/volcengine/verl/blob/main/CONTRIBUTING.md

PR #4860, opened 2026-01-09
[misc] fix: fix 3d position_ids for train_mini_batch

wdl339:
### What does this PR do?

This PR fixes a bug in the `train_mini_batch` function for TrainingWorker.

Nested tensor in tensordict after consolidate and pickle/unpickle will incur indexing error for ragged tensor. This only happens when using 3D position ids in VLMs. This error is triggered when the `train_mini_batch` function iterates through the dataloader. This problem can be solved by introducing the `_maybe_fix_3d_position_ids` function.

### DIFF
--- a/verl/utils/tensordict_utils.py
+++ b/verl/utils/tensordict_utils.py
@@ -819,7 +819,6 @@ def contiguous(data: TensorDict) -> TensorDict:
     """Call contiguous on a tensor dict. The contiguous function of tensordict lib will make NonTensorStack.
     This function will always return a new tensordict
 
-    
     Args:
         data: The input tensordict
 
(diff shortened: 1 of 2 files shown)


verl/utils/tensordict_utils.py
```diff
@@ -840,3 +840,10 @@ def contiguous(data: TensorDict) -> TensorDict:
             tensor_dict[key] = val.contiguous()
 
     return get_tensordict(tensor_dict=tensor_dict, non_tensor_dict=non_tensor_dict)
+
+def _maybe_fix_3d_position_ids(data: TensorDict):
```
wuxibin89: should be a public function without `maybe_fix_3d_position_ids`

wdl339: Ok, it has been corrected. 🙂

wuxibin89 (approved):

Issue #4849, opened 2026-01-09
Multimodal data processing bug in RLHFDataset in the latest version


CYDaaa30:
In the latest version of `verl/utils/dataset/rl_dataset.py`, the logic in `RLHFDataset.__getitem__` has been simplified to only call `_build_messages`, with subsequent operations like `apply_chat_template` and `process_vision_info` moved into the `AgentLoop`. However, several changes in `_build_messages` have introduced bugs that cause training failures.
1. Line 298-299, in `_build_messages`
```
images = example.pop(self.image_key, [])
videos = example.pop(self.video_key, [])
```
- The use of .`pop()` removes the images and videos keys from example (i.e., row_dict). However, the downstream function `maybe_filter_out_long_prompts` still expects these keys to be present in the input dictionary to correctly process multimodal data:
```
if image_key in doc and doc[image_key]:
    images = [
        process_image(image, image_patch_size=self.image_patch_size) for image in doc[image_key]
    ]
  else:
    images = None
  
  if video_key in doc and doc[video_key]:
    videos, video_metadata = zip(
        *[
            process_video(
                video, image_patch_size=self.image_patch_size, return_video_metadata=True
            )
            for video in doc[video_key]
        ],
        strict=True,
    )
    videos = list(videos)
    video_metadata = list(video_metadata)
    videos_kwargs = {"video_metadata": video_metadata, "do_sample_frames": False}
```
- Because `.pop()` deletes the keys, `maybe_filter_out_long_prompts` fails to detect multimodal content and skips necessary processing, leading to incorrect behavior during filtering.
- Suggested fix: Replace `.pop()` with `.get()` to preserve the original keys.
2. Line 324-327, in `_build_messages`
```
elif segment == "<video>":
  assert video_offset < len(videos), f"video_offset {video_offset} >= len(videos) {len(videos)}"
  content_list.append({"type": "video", "video": videos[video_offset]})
  video_offset += 1
```
- This implementation only works when `videos[video_offset]` is a `str` path.
- When `videos[video_offset]` is a `dict` type (e.g., containing configuration like `fps`, `min_frames`, `max_frames`), resulting in a nested structure like:
`{"type": "video", "video": {"video": "/path/to/video", "fps": 10, "min_frames": 5, ...}}`
- This causes incorrect `ele` format when calling `qwen_vl_utils.process_vision_info` later, resulting in errors.
- Suggested Fix:
```
elif segment == "<video>":
[...]

PR #4864, opened 2026-01-09
[data] fix: build_messages for multi-modal data

ccilery:
### What does this PR do?

fix https://github.com/volcengine/verl/pull/4759


1. When concatenating image and video datasets, pop will return None for image or video sample, this will raise Traceback
 
Original Traceback (most recent call last):
  File "/usr/local/lib/python3.12/dist-packages/torchdata/stateful_dataloader/worker.py", line 242, in _worker_loop
    data = fetcher.fetch(index)  # type: ignore[union-attr]
           ^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/dist-packages/torch/utils/data/_utils/fetch.py", line 52, in fetch
    data = [self.dataset[idx] for idx in possibly_batched_index]
            ~~~~~~~~~~~~^^^^^                                                                                                                                                                                                                 File "/tmp/ray/session_2026-01-09_06-01-53_053078_153/runtime_resources/working_dir_files/_ray_pkg_e77e513345569ff8/verl/utils/dataset/rl_dataset.py", line 339, in __getitem__
    row_dict["raw_prompt"] = self._build_messages(row_dict)
                             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/tmp/ray/session_2026-01-09_06-01-53_053078_153/runtime_resources/working_dir_files/_ray_pkg_e77e513345569ff8/verl/utils/dataset/rl_dataset.py", line 332, in _build_messages
    assert image_offset == len(images), f"image_offset {image_offset} != len(images) {len(images)}"
                           ^^^^^^^^^^^
TypeError: object of type 'NoneType' has no len()

2. messages format is like this, so we need unpacking instead of insert image or video, this will raise Traceback
```
[
      {
          "role": "user",
          "content": [
              {
                  "type": "video",
                  "video": os.path.expanduser("~/models/hf_data/test-videos/space_woaudio.mp4"),
                  "min_pixels": 4 * 32 * 32,
                  "max_pixels": 256 * 32 * 32,
                  "total_pixels": 4096 * 32 * 32,
              },
              {
                  "type": "text",
                  "text": "Describe this video. Then you must call the "
                  "image generator tool to generate a green image for me.",
              },
          ],
      },
]
```       

[...]

PR #4871, opened 2026-01-10
fix: multimodal data processing bug in RLHFDataset in the latest version

CYDaaa30:
This PR fixes two issues in `RLHFDataset._build_messages`:

1. Replaces `.pop()` with `.get()` to avoid mutating the input example, which broke `doc2len` in `maybe_filter_out_long_prompts`.
2. Handles video entries that are dicts (e.g., with fps/min_frames config) in addition to string paths.

Fixes the issue reported in #4849.


wuxibin89:
Fixed in https://github.com/volcengine/verl/pull/4864

Issue #3159, opened 2025-08-21
Running grpo with lora, the model response is "!!!!!!"


mootutu:
I trained Qwen2.5-3B-Instruct with GRPO using two almost identical scripts: one without LoRA and one with LoRA. The only difference between them is the LoRA configuration. However, when using LoRA, the model’s output always degenerates into "!!!!!!!!!!!!".
<details>
<summary>grpo setting and output</summary>

**scripts**
```sh
set -x
# GPU
export CUDA_VISIBLE_DEVICES=0,1,2,3,4,5,6,7
GPU_MEMORY_UTILIZATION=0.6
N_GPUS_PER_NODE=8
NNODES=1
TENSOR_PARALLEL_SIZE=2
# model
MODEL="Qwen2.5/Qwen2.5-3B-Instruct"
ALL_MODEL="/mnt/d/All/models"
MODEL_PATH=$ALL_MODEL/$MODEL
# dataset
DATASET="gsm8k"
ALL_DATA="/mnt/d/verl4test/data"
DATA_PATH=$ALL_DATA/$DATASET
# rollout
ROLLOUT_N=5
# training
GAE="grpo"
LORA_RANK=32
LORA_ALPHA=32
LR=3e-6
EPOCH=15
BATCH_SIZE=8
PROMPT_LENGTH=512
RESPONSE_LENGTH=1024
MINI_BATCH_SIZE=8
MICRO_BATCH_SIZE=1
# logger
DATE=$(date +%Y%m%d)
TIME_TAG=$(date +%H%M%S)
SWANLAB_KEY=
# EXPERIMENT_NAME="lora_test_${MODEL}_${DATASET}"
PROJECT_NAME="lora_test"
LOG_NAME="${GAE}_${MODEL}_${DATASET}_${DATE}_${TIME_TAG}"
OUTPUT_DIR="/mnt/d/verl4test/checkpoints/${PROJECT_NAME}/${LOG_NAME}"

python3 -m verl.trainer.main_ppo \
    algorithm.adv_estimator=$GAE \
    data.train_files=$DATA_PATH/train.parquet \
    data.val_files=$DATA_PATH/test.parquet \
    data.train_batch_size=$BATCH_SIZE \
    data.max_prompt_length=$PROMPT_LENGTH \
    data.max_response_length=$RESPONSE_LENGTH \
    data.filter_overlong_prompts=True \
    data.truncation='error' \
    data.shuffle=False \
[...]

PR #4897, opened 2026-01-12
[fsdp,vllm,trainer,algo] feat: On-Policy Distillation

JacobHelwig:
### What does this PR do?

Adds on-policy distillation to FSDP engine with top-k distillation loss and KL estimator distillation losses with supervised and PG-style updates. Teacher logprobs are computed using a vLLM teacher server.

#### Losses 

1. top-k distillation loss: **forward** KL estimated using top-k logits **from teacher**. 
2. KL estimator distillation losses: **reverse** KL estimated using only the log prob for the sampled token via the same estimators used by the reference model (e.g., k1, k3)  

#### Updates

1. Supervised: distillation loss is directly backpropogated, as in https://arxiv.org/abs/2306.13649
2. Policy gradient: negative distillation loss is used as a reward, as in https://thinkingmachines.ai/blog/on-policy-distillation/

# Test

Tested with `examples/on_policy_distillation_trainer/run_qwen_gsmk8k.sh`. 

## Main results

These experiments compare 3 training runs with student model Qwen2.5-0.5B:

1. Forward top-k KL with Qwen2.5-3B-Instruct teacher (gold)
2. Forward top-k KL with Qwen2.5-7B-Instruct teacher (green)
3. k3 estimator KL with Qwen2.5-7B-Instruct teacher (red)

### GSM8K eval acc

image

### GSM8K train acc

image

### Distillation loss

image

## Top-k training stability

Clamping the top-k forward KL loss was needed for training stability. These experiments compare 3 types of clamping: 

1. No clamping (grey)
2. Clamping the distillation loss to a maximum value of 10 (blue)
3. Clamping the student and teacher log probs to a minimum value of -10 (gold)

### Distillation loss

image

### GSM8K eval acc

image

### GSM8K train acc

image

## Teacher server results

The initial implementation of OPD in this PR treated the teacher the same as the reference model. Teacher logprobs were calculated using the `ActorRolloutRef` engine worker. Outside of this section, all results in the PR description use the old engine worker teacher with the supervised version of the update. 

The following results are for the current version of the PR, which computes teacher logprobs using a vLLM worker teacher. Each uses Qwen2.5-0.5B as the student and Qwen2.5-3B-Instruct as the teacher, with the max value of the distillation loss clamped to 10.

1. Forward top-k KL with supervised update (green)
2. k1 estimator KL with policy gradient update (purple)
3. k3 estimator KL with supervised update (red)

While purple seems best, it also is generating responses that exceed the maximum response length of 512.

### Distillation loss

image

### GSM8K eval acc

image

### GSM8K train acc

image

### Response length

image

## Note on reverse KL

Initially, this PR included top-k reverse KL and top-k Jensen-Shannon divergences (JSD interpolates between forward and reverse KL). For the student distribution $q$ and teacher distribution $p$, the top-k reverse KL is given by

$$
KL_{\text{top-}k}(q||p) = \sum_i \bf{1}(q_i\in \text{top-}k)q_i\log\frac{q_i}{p_i}.
$$ 

Unfortunately, this was unstable. The reason is because one way to make this loss small is to make $q_i$ as small as possible for all $q_i \in \text{top}-k$. This can be seen from the logs tracking the amount of mass captured in the top-$k$ probabilities:

image

## Ablation: performance with more lenient parser

Note that the only loss used is the distillation loss (no rewards for correctness on GSM8K). Any increase in the logged rewards=GSM8k accuracy are an indirect result of minimizing the distillation loss. The reason that the base model has Pass@1~=0 is because the default GSM8k answer formatting (`#### 42`) is OOD for the model. The base model is answering the questions correctly, but using incorrect formatting, so none of the answers can be parsed. The base model can be evaluated using a reward function that is more lenient on formatting by adding the following to the script:

```bash
...
    reward_model.reward_manager=remote \
    custom_reward_function.path=tests/experimental/reward_loop/reward_fn.py \
    custom_reward_function.name=compute_score_math_verify \
    trainer.val_only=True
``` 

The results are:

```bash
(TaskRunner pid=904198) ("Initial validation metrics: {'val-aux/openai/gsm8k/reward/mean@1': "
(TaskRunner pid=904198)  "np.float64(0.31766489764973466), 'val-core/openai/gsm8k/acc/mean@1': "
(TaskRunner pid=904198)  "np.float64(0.31766489764973466), 'val-aux/num_turns/min': np.int32(2), "
(TaskRunner pid=904198)  "'val-aux/num_turns/max': np.int32(2), 'val-aux/num_turns/mean': "
(TaskRunner pid=904198)  'np.float64(2.0)}')
```

### Design & Code Changes

- Teacher workers are used in the agent loop, similar to the generative reward model: after a student worker finishes its rollout, the teacher worker obtains logprobs
- The distillation loss is applied in the `ppo_loss` function

### DIFF
--- a/verl/experimental/teacher_loop/teacher_model.py
+++ b/verl/experimental/teacher_loop/teacher_model.py
@@ -54,7 +54,11 @@ class TeacherModelManager:
 
     def _initialize_llm_servers(self):
         teacher_model_config: DistillationTeacherModelConfig = self.config.teacher_model
-        teacher_world_size = teacher_model_config.inference.tensor_model_parallel_size
+        teacher_world_size = (
+            teacher_model_config.inference.tensor_model_parallel_size
+            * teacher_model_config.inference.data_parallel_size
+            * teacher_model_config.inference.pipeline_model_parallel_size
+        )
         world_size = (
             self.resource_pool.world_size
             if self.resource_pool  # colocate mode


yubin1991:
why the reward is so small (almost 0)  before step 40 ?

JacobHelwig:
> why the reward is so small (almost 0) before step 40 ?

Training only explicitly optimizes the distillation loss, not rewards:

> Note that we can do OPD + RL rewards, although to isolate effects of OPD, RL rewards aren't used in these experiments.

Any increase in the logged rewards=GSM8k accuracy are an indirect result of minimizing the distillation loss. 

In this case, the reason that the base model has Pass@1~=0 is because the default GSM8k answer formatting (`#### 42`) is OOD for the model. The base model is answering the questions correctly, but using incorrect formatting, so none of the answers can be parsed. The base model can be evaluated using a reward function that is more lenient on formatting by adding the following to the script:

```bash
...
    reward_model.reward_manager=remote \
    custom_reward_function.path=tests/experimental/reward_loop/reward_fn.py \
    custom_reward_function.name=compute_score_math_verify \
    trainer.val_only=True
``` 

The results are:

```bash
(TaskRunner pid=904198) ("Initial validation metrics: {'val-aux/openai/gsm8k/reward/mean@1': "
(TaskRunner pid=904198)  "np.float64(0.31766489764973466), 'val-core/openai/gsm8k/acc/mean@1': "
(TaskRunner pid=904198)  "np.float64(0.31766489764973466), 'val-aux/num_turns/min': np.int32(2), "
(TaskRunner pid=904198)  "'val-aux/num_turns/max': np.int32(2), 'val-aux/num_turns/mean': "
(TaskRunner pid=904198)  'np.float64(2.0)}')
```

The formatting is only a few tokens, so it does not contribute much to the distillation loss. The distillation loss initially focuses on minimizing other discrepancies between the teacher and student distributions before targeting formatting, which is why early steps of training show 0% accuracy under the stricter parser.

xvrong:
I have a question regarding an observation during training with this version of the code.

When I use reverse_kl_topk+ as the loss function, the model’s response length increases very rapidly and quickly reaches MAX_RESPONSE_LENGTH. At the same time, the metric response_length_non_aborted/clip_ratio rises steadily and approaches around 1.0.

I would like to ask whether this behavior is expected when using reverse_kl_topk+ loss. Have you observed the same experimental phenomenon in your own experiments, or is this indicative of a potential issue or misconfiguration?

update: It’s not an on-policy distillation issue. I’ve resolved this problem through the following discussion: https://github.com/volcengine/verl/issues/3159

JacobHelwig:
> I have a question regarding an observation during training with this version of the code.
> 
> When I use reverse_kl_topk+ as the loss function, the model’s response length increases very rapidly and quickly reaches MAX_RESPONSE_LENGTH. At the same time, the metric response_length_non_aborted/clip_ratio rises steadily and approaches around 1.0.
> 
> I would like to ask whether this behavior is expected when using reverse_kl_topk+ loss. Have you observed the same experimental phenomenon in your own experiments, or is this indicative of a potential issue or misconfiguration?
> 
> update: It’s not an on-policy distillation issue. I’ve resolved this problem through the following discussion: #3159

I did see an issue w the student model hacking the top-k loss by making its distribution closer and closer to uniform. The result is low mass in the top-k probs: 

image

This led to the clip frac and response length increasing. 

I plan to address this in another PR this week. For now, you could try using the k3 estimator.

verl/trainer/config/distillation/distillation.yaml
```diff
@@ -0,0 +1,40 @@
...
+
+# specify the default per-component configs
+defaults:
+
+  # distillation config, inheriting from trainer/config/ref/ref.yaml
+  # Use absolute path + @_here_ to flatten fields at current level
+  - /ref/ref@_here_
+
+  # Model config for teacher
+  - ../model@teacher_model: hf_model
```
wuxibin89: We may need to support multiple domain-specific teachers, since in mixed datasets, each sample may choose different teacher: math, code, search, etc. Currently we can support one teacher model first.

wuxibin89: From MiMo-V2-Flash https://arxiv.org/pdf/2601.02780

Image

verl/trainer/distillation/losses.py
```diff
@@ -0,0 +1,232 @@
...
+        torch.Tensor,  # response_mask
+        DistillationConfig,  # config
+        str,  # loss_agg_mode
+    ],
+    tuple[torch.Tensor, dict[str, Any]],
+]
+
+
+@dataclass
+class DistillationLossSettings(BaseConfig):
```
wuxibin89, line 39: Please add docstring for each fields.

verl/trainer/distillation/fsdp/losses.py
```diff
@@ -0,0 +1,106 @@
+# Copyright 2025 Bytedance Ltd. and/or its affiliates
```
wuxibin89, line 1: Why the losses is under fsdp, shouldn't them be common to all backends: fsdp/megatron/veomni? Merge `losses.py` and `utils.py` with outer files?

verl/trainer/config/distillation/distillation.yaml
wuxibin89: For multiple teachers, there're some options:
1. Load multiple teachers in trainer worker group, and load/offload them sequentially to compute teacher's logits. This is easy to implemented, but it may bounded by the CPU memory, since we must offload multiple model at same time.
2. Chunk trainer worker group into multiple subgroups and load teacher model in each subgroup. For example, trainer world_size=64 with 4 teacher models, then GPU [0-15] for teacher 0, GPU [16-31] for teacher 2, ... While this can make teacher models compute parallel, it need megatron support different parallelism for teacher and student model.
3. Add separate GPU resources, and launch vllm/sglang replicas to serve each teacher model. Then we can request server per sample once agent loop finished.

verl/trainer/distillation/utils.py
```diff
@@ -0,0 +1,134 @@
...
+from tensordict import TensorDict
+
+from verl.trainer.distillation.fsdp import utils as fsdp_utils
+from verl.trainer.distillation.losses import DistillationLossSettings, get_distillation_loss_settings
+from verl.trainer.distillation.types import DistillationLossInputs
+from verl.utils.stages import Stage
+from verl.workers.config import DistillationConfig
+from verl.workers.utils.padding import _slice_response_from_unpad_output
+
+TEACHER_TOPK_LOGPROBS_KEY = "teacher_topk_logprobs"
```
wuxibin89: This is not align with `DistillationLossInputs`.

verl/trainer/distillation/types.py
```diff
@@ -0,0 +1,30 @@
...
+
+from verl.base_config import BaseConfig
+
+
+@dataclass
+class DistillationLossInputs(BaseConfig):
+    """Storage class for distillation loss inputs."""
+
+    student_log_probs: Optional[torch.Tensor] = None
+    teacher_log_probs: Optional[torch.Tensor] = None
```
wuxibin89, line 28: Better have a separate field for `teacher_topk_logprobs`

verl/trainer/config/distillation/distillation.yaml
```diff
@@ -0,0 +1,40 @@
...
+enabled: false
+
+# Loss function mode for distillation
+loss_mode: k3
+
+# If distillation loss requires top-k logits, this is the value of k
+topk: 32
+
+# Whether the loss should just be the distillation loss or a combination of the distillation loss and the policy loss
+use_policy_loss: false
```
wuxibin89: Should we make it default to true?

verl/trainer/config/distillation/distillation.yaml
JacobHelwig: Good idea, and thanks for the implementation guidance.

I think that the efficiency of the three options is 3 > 2 > 1. 3 should work as long as we only need the logprob for the sampled token (not the full distribution). In Eqn 8 of the MiMo-V2-Flash Technical Report, it shows that their multi-teacher distillation loss is using k1 as the distillation loss, which only requires the logprob for the sampled token. 

However, if we want to unify multi-teacher with single teacher (ie, single teacher as a special case of multi-teacher) and still support top-k/full logprob distillation losses for the single teacher, then I am not sure if option 3 is possible due to vllm/sglang not returning full logprobs. For vLLM, there are several related issues that seem to be unresolved, so I don't think that vLLM supports full logprobs (not sure about sglang, though):
- https://github.com/vllm-project/vllm/issues/14300
- https://github.com/vllm-project/vllm/issues/185

Assuming it is true that we can only use vLLM/sglang to get logprobs for the sampled token, I see several options:
(A) Go with option 2 and unify single teacher into multi-teacher while maintaining supports for top-k/full logprob losses
(B) Go with option 3 and only support single sample distillation losses
(C) 2 separate paths: keep option 2 for single teacher w/ top-k/full logprob losses (the route taken by this PR). Implement option 3 for multi-teacher with single sample losses 

I think (A) might be best since it allows for more distillation losses while being unified, but WDYT?

verl/trainer/distillation/fsdp/losses.py
JacobHelwig: For megatron, logits are split across TP ranks. We will need special logic for gathering top-k from the teacher in `megatron.utils` and a custom autograd function for computing the distillation loss in `megatron.losses`, like GKD in `recipe`: https://github.com/JacobHelwig/verl/blob/dcd199beb2c64f4f4c9879bacd1f811c0b4bdacf/verl/trainer/distillation/megatron/losses.py

I can still merge them into the outer files, though. WDYT?

verl/trainer/config/distillation/distillation.yaml
wuxibin89: Yes, I think (A) is the best, let's still use model engine to support multiple teacher instead of rollout engine(vllm/sglang). We can support single teacher first, then multiple teachers.

JacobHelwig:
Diff since last review: https://github.com/JacobHelwig/verl/compare/283e612...jhelwig/onPolicyDistillation

verl/trainer/distillation/fsdp/losses.py
wuxibin89: Got it, let's keep them separate.

AvengersZhi:
Hello! Does this PR currently support OPD training directly on the Qwen3-VL series models?

Sususue:
Hello, does this PR support training on Ascend NPUs? What modifications are needed based on the examples/on_policy_distillation_trainer/run_qwen_gsmk8k.sh script? Will an example script for running on NPUs be provided in the future?
I used the image "verl-8.3.rc1-910b-ubuntu22.04-py3.11-v0.7.0" and tried training on the NPU by making some modifications to the script mentioned above, but found that it could not run properly. Therefore, I would like to know if NPU support is available.

tardis-key:
> Hello, does this PR support training on Ascend NPUs? What modifications are needed based on the examples/on_policy_distillation_trainer/run_qwen_gsmk8k.sh script? Will an example script for running on NPUs be provided in the future? I used the image "verl-8.3.rc1-910b-ubuntu22.04-py3.11-v0.7.0" and tried training on the NPU by making some modifications to the script mentioned above, but found that it could not run properly. Therefore, I would like to know if NPU support is available.

I don't think that this script has been officially adapted for npu yet (not 100%  sure ), but I believe the adaptation effort won't be significant—it mainly involves changes to the upper-level logic related to the ref model and reward.

JacobHelwig:
> Hello, does this PR support training on Ascend NPUs? What modifications are needed based on the examples/on_policy_distillation_trainer/run_qwen_gsmk8k.sh script? Will an example script for running on NPUs be provided in the future? I used the image "verl-8.3.rc1-910b-ubuntu22.04-py3.11-v0.7.0" and tried training on the NPU by making some modifications to the script mentioned above, but found that it could not run properly. Therefore, I would like to know if NPU support is available.

NPU support will probably have to wait since I don't have NPUs to debug on. What's the error you're getting and can you paste the script?

JacobHelwig:
> Hello! Does this PR currently support OPD training directly on the Qwen3-VL series models?

~I haven't tried it, but it should.~ No, it does not. Support will be added in the next OPD PR.

Edit: VLM OPD is support in #5592

waltonfuture:
@JacobHelwig Hello! Excuse me, can on-policy distillation be directly and simply implemented based on the grpo training code? For example, is it correct to refer to the KL loss based on the reference model in grpo and modify it to be based on the teacher model to achieve on-policy distillation? I'm not sure if my understanding is correct.

waltonfuture:
> @JacobHelwig Hello! Excuse me, can on-policy distillation be directly and simply implemented based on the grpo training code? For example, is it correct to refer to the KL loss based on the reference model in grpo and modify it to be based on the teacher model to achieve on-policy distillation? I'm not sure if my understanding is correct.

image
image

It seems that we can modify the KL loss of reference model to the teacher model to achieve on-policy distillation

JacobHelwig:
> > @JacobHelwig Hello! Excuse me, can on-policy distillation be directly and simply implemented based on the grpo training code? For example, is it correct to refer to the KL loss based on the reference model in grpo and modify it to be based on the teacher model to achieve on-policy distillation? I'm not sure if my understanding is correct.
> 
> image image
> It seems that we can modify the KL loss of reference model to the teacher model to achieve on-policy distillation

Yes, OPD can be done by extending the reference model, and this PR initially took this approach. Please see the section **Teacher server results** in the PR desc for more details and fc4f5df for the latest version of the reference model-based teacher code prior to migration to the current teacher server approach.

We decided to optimize this using a teacher server in the agent loop so that as soon as a rollout replica finishes a student gen, teacher logprobs can be computed by a teacher replica. Credit to @wuxibin89 and @vermouth1992 for this idea.

wuxibin89:
@JacobHelwig Does this PR support multi-modal VLM model (e.g Qwen3-VL)?

verl/trainer/distillation/utils.py
```diff
@@ -0,0 +1,90 @@
...
+        return False
+    return config.enabled
+
+
+def prepare_student_distillation_inputs(
+    logits: torch.Tensor, batch: TensorDict, cu_seqlens: torch.Tensor, config: Optional[DistillationConfig]
+) -> dict[str, torch.Tensor]:
+    """Prepare student distillation inputs."""
+    stage = None if "stage" not in batch else batch["stage"]
+    if not is_distillation_enabled(config) or stage != Stage.ACTOR_UPDATE:
```
pengwu22: **Suggestion: Consider removing the `Stage` enum and using existing `compute_loss` metadata instead**

The new `Stage` enum in `verl/utils/stages.py` is only consumed in one place — `prepare_student_distillation_inputs` in `verl/trainer/distillation/utils.py` — to check whether we're in the actor update step. But the batch metadata already carries `compute_loss=False` for both the ref log prob and old log prob paths (set in `ray_trainer.py`), and it's effectively `True` during the actor update. So we can replace it

with:
```python
compute_loss = batch.get("compute_loss", True)
if not is_distillation_enabled(config) or not compute_loss:
    return {}
```

This would let us:
- Delete `verl/utils/stages.py`
- Remove the three `stage=Stage.*` assignments in `ray_trainer.py`
- Remove the `Stage` imports from `ray_trainer.py` and `distillation/utils.py`

It avoids introducing a new abstraction when an existing signal already distinguishes the same thing, and keeps the diff smaller.

JacobHelwig:
> @JacobHelwig Does this PR support multi-modal VLM model (e.g Qwen3-VL)?

@wuxibin89 This PR does not support VLM due to the reliance on `v1/completions` for submitting requests to the teacher server. #5592 adds support for VLM OPD.

verl/trainer/distillation/utils.py
JacobHelwig: Thanks, applied this change

wuxibin89:
Close this PR as we will review on https://github.com/verl-project/verl/pull/5592

PR #4933, opened 2026-01-15
GIFT: group implicit fine tuning

ZhichaoWang970201:
# GIFT: Group-relative Implicit Fine-Tuning

GIFT is an **on-policy reinforcement learning framework for aligning large language models (LLMs)**. It unifies key ideas from **GRPO, DPO, and UNA** while avoiding many of their practical limitations. 

You can test it on "examples/gift_trainer/test_qwen_gift_math_3e-6_1.slurm" which utilizes DAPO's data for training and AIME's data for evaluation.

The core idea is simple: instead of directly maximizing rewards with complex RL objectives, GIFT **matches rewards within small sampled groups**, turning alignment into a **stable, regression-like learning problem**.

---

## Why GIFT?

Aligning LLMs with reinforcement learning is powerful but often unstable:

* PPO-style methods require careful clipping and value estimation
* Offline methods (like DPO / UNA) lack exploration
* Reward scaling and normalization are hard to tune

GIFT addresses these issues with a cleaner and more robust objective.

---

## Core Idea (Intuition)

1. **Sample a group of responses on-policy** for the same prompt
2. **Compute two signals** for each response:

   * An *explicit reward* (from a reward model or heuristic)
   * An *implicit reward* derived from how the policy differs from a reference model
3. **Normalize both signals within the group** (center and scale them)
4. **Train the policy to match the two normalized signals using simple regression**

Because everything is done *relative to the group*, problematic global normalization terms disappear automatically.

---

## What This Buys You

* **Stable training** – no PPO clipping, no value function
* **Low-variance gradients** – group-relative normalization
* **Fast convergence** – behaves like supervised learning
* **On-policy exploration** – unlike offline preference methods

In practice, this makes RL alignment **much easier to scale and tune**.

---

## How GIFT Relates to Prior Work

* **GRPO**: keeps on-policy sampling, but removes heavy RL machinery
* **DPO**: shares the idea of implicit rewards, but works on-policy
* **UNA**: aligns implicit and explicit rewards, but with group-relative normalization

GIFT can be seen as a **unifying framework** that combines the strengths of all three.


vermouth1992 (dismissed):

ZhichaoWang970201:
@eric-haibin-lin @PeterSH6 @tongyx361 Would you mind helping me approve on that? Thank you.

PR #4967, opened 2026-01-18
[training_utils] A bug that caused device selection in group statistics to fail has been covered by tests.

JohnConnor123:
### What does this PR do?

Fixes a GRPO crash in `verl.utils.groupwise.group_mean_std` caused by passing a device *module* (e.g., `torch.cuda`) to `Tensor.to(device=...)`. The default device resolution now returns a proper `torch.device` (e.g., `cuda:0`), and a regression test is added to prevent this from coming back.

### Checklist Before Starting

- [ ] Search for similar PRs. Paste at least one query link here: [Search PRs: group_mean_std device module](https://github.com/volcengine/verl/pulls?q=is%3Apr+group_mean_std+device+module)
  - Title used: `[algo, training_utils] fix: prevent device=module crash in group_mean_std`

### Test

- Reproduced the original failure path in docker (no `VERL_FORCE_DEVICE`, non-pytest code path) and verified the crash is gone:
  - `group_mean_std(...)` now returns tensors on `cuda:0` instead of failing with `device=module`.
- Added a regression test in `tests/utils/test_groupwise.py`.
- Note: I could not run the full pytest suite locally because `pytest` is not installed in the current docker image; CI should cover it.

### API and Usage Example

No API changes.

# Minimal reproduction (used for docker smoke test)
import os
os.environ.pop("VERL_FORCE_DEVICE", None)
os.environ.pop("PYTEST_CURRENT_TEST", None)

import torch
from verl.utils.groupwise import group_mean_std

scores = torch.tensor([1.0, 2.0, 3.0])
gidx = torch.tensor([0, 1, 0])

mean_g, std_g, cnt_g = group_mean_std(scores, gidx)
print(mean_g.device, std_g.device, cnt_g.device)

### DIFF
--- a/verl/utils/groupwise.py
+++ b/verl/utils/groupwise.py
@@ -46,7 +46,7 @@ from typing import Any, Optional
 import numpy as np
 import torch
 
-from verl.utils.device import get_device_id, get_device_name
+from verl.utils.device import get_torch_device
 
 __all__ = ["as_torch_index", "group_mean_std"]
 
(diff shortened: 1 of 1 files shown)


vermouth1992:
I guess this is fixed in a previous PR https://github.com/volcengine/verl/pull/4961. Maybe we can still keep the test?

JohnConnor123:
> I guess this is fixed in a previous PR #4961. Maybe we can still keep the test?

I removed the fix, leaving only the test file

PR #4634, opened 2025-12-22
[rollout] feat: ensure max_new_tokens is set correctly in sampling_params

yanyc428:
### What does this PR do?

In the existing code, the `generate` method in `workers/rollout/sglang_rollout/async_sglang_server.py` forcibly sets `max_new_tokens` based on the config. This behavior overrides the `sampling_params` passed in, which prevents developers from customizing multi-step rollout behavior of the model in `AgentLoop`, since the `max_new_tokens` they set does not take effect.

This PR fixes the issue. If the provided `sampling_params` includes the `max_new_tokens` attribute, it will no longer be overridden (though it will still be validated). Otherwise, the original logic is preserved, and `max_new_tokens` is set according to the `response_length` defined in the config.

### Checklist Before Starting

- [ ] Search for similar PRs. Paste at least one query link here: ...
- [ ] Format the PR title as `[{modules}] {type}: {description}` (This will be checked by the CI)
  - `{modules}` include `fsdp`, `megatron`, `sglang`, `vllm`, `rollout`, `trainer`, `ci`, `training_utils`, `recipe`, `hardware`, `deployment`, `ray`, `worker`, `single_controller`, `misc`, `perf`, `model`, `algo`, `env`, `tool`, `ckpt`, `doc`, `data`, `cfg`, `reward`
  - If this PR involves multiple modules, separate them with `,` like `[megatron, fsdp, doc]`
  - `{type}` is in `feat`, `fix`, `refactor`, `chore`, `test`
  - If this PR breaks any API (CLI arguments, config, function signature, etc.), add `[BREAKING]` to the beginning of the title.
  - Example: `[BREAKING][fsdp, megatron] feat: dynamic batching`

### Test

> For changes that can not be tested by CI (e.g., algorithm implementation, new model support), validate by experiment(s) and show results like training curve plots, evaluation results, etc.

### API and Usage Example

> Demonstrate how the API changes if any, and provide usage example(s) if possible.

```python
# Add code snippet or script demonstrating how to use this
```

### Design & Code Changes

> Demonstrate the high-level design if this PR is complex, and list the specific changes.

### Checklist Before Submitting

> [!IMPORTANT]
> Please check all the following items before requesting a review, otherwise the reviewer might deprioritize this PR for review.

- [ ] Read the [Contribute Guide](https://github.com/volcengine/verl/blob/main/CONTRIBUTING.md).
[...]

PR #5028, opened 2026-01-23
[rollout, vllm, sglang] fix: forward max_tokens/max_new_tokens from rollout config to vllm/sglang backends

psyloy:
### What does this PR do?

In the current codebase, when max_tokens is not specified in sampling_params, its default value is dynamically calculated based on prompt_ids . This approach can lead to the generation of text exceeding the intended length, resulting in unnecessary computation, resource waste, and reduced inference performance. By configuring max_tokens in the configuration file, a fixed upper bound can be set to prevent uncontrolled text generation, ensuring predictable resource utilization and inference efficiency.

>verl/workers/rollout/vllm_rollout/vllm_async_server.py +505
```python
# Determine max_tokens from sampling_params or use configured response_length as default
if "max_tokens" in sampling_params:
    max_tokens = sampling_params.pop("max_tokens")
elif "max_new_tokens" in sampling_params:
    # support sglang-style 'max_new_tokens' param
    max_tokens = sampling_params.pop("max_new_tokens")
else:
    # Default to a calculation that considers configured lengths
    max_tokens = self.config.response_length + self.config.prompt_length - len(prompt_ids)
```

However, the max_tokens/max_new_tokens parameter is not passed into sampling_params in agent_loop.py. This leads to max_tokens in vllm_async_server.py only taking the default dynamically calculated value, which does not meet expectations in certain scenarios.
>verl/experimental/agent_loop/agent_loop.py +428
```python
config = self.config.actor_rollout_ref.rollout
sampling_params = dict(
    temperature=config.temperature,
    top_p=config.top_p,
    top_k=config.top_k,
    repetition_penalty=1.0,
    logprobs=config.calculate_log_probs,
)
```

Therefore, we have added the two parameters max_tokens/max_new_tokens to the rollout configuration, setting their default value as null. When these parameters are configured in the runtime script file, the corresponding values in vllm_async_server.py/sglang_async_server.py will be updated accordingly. If the parameters are not configured in the runtime script file, the original default logic will be followed, thereby avoiding any impact on existing functionality.
>verl/experimental/agent_loop/agent_loop.py
```python
for param_name in ["max_tokens", "max_new_tokens"]:
    param_value = getattr(config, param_name, None)
    if param_value is not None:
        sampling_params[param_name] = param_value
)
```

### Checklist Before Starting

- [ ] Search for similar PRs. Paste at least one query link here: [[feat: ensure max_new_tokens is set correctly in sampling_params](https://github.com/volcengine/verl/pull/4634)

### API and Usage Example

```shell
 actor_rollout_ref.rollout.max_tokens=${max_response_length}
```

### DIFF
--- a/verl/trainer/config/_generated_ppo_trainer.yaml
+++ b/verl/trainer/config/_generated_ppo_trainer.yaml
@@ -195,6 +195,8 @@ actor_rollout_ref:
     temperature: 1.0
     top_k: -1
     top_p: 1
+    max_tokens: null
+    max_new_tokens: null
     prompt_length: ${oc.select:data.max_prompt_length,512}
     response_length: ${oc.select:data.max_response_length,512}
     dtype: bfloat16
--- a/verl/experimental/agent_loop/agent_loop.py
+++ b/verl/experimental/agent_loop/agent_loop.py
@@ -433,6 +433,12 @@ class AgentLoopWorker:
             logprobs=config.calculate_log_probs,
         )
 
+        # configure max generation tokens for vllm/sglang
+        for param_name in ["max_tokens", "max_new_tokens"]:
+            param_value = getattr(config, param_name, None)
+            if param_value is not None:
+                sampling_params[param_name] = param_value
+
         # override sampling params for validation
         if batch.meta_info.get("validate", False):
             sampling_params["top_p"] = config.val_kwargs.top_p
--- a/verl/experimental/fully_async_policy/agent_loop/agent_loop.py
+++ b/verl/experimental/fully_async_policy/agent_loop/agent_loop.py
@@ -101,13 +101,21 @@ class FullyAsyncAgentLoopWorker(AgentLoopWorker):
         sampling_params = dict(
             temperature=config.temperature,
             top_p=config.top_p,
+            top_k=config.top_k,
             repetition_penalty=1.0,
             logprobs=config.calculate_log_probs,
         )
 
+        # configure max generation tokens for vllm/sglang
+        for param_name in ["max_tokens", "max_new_tokens"]:
+            param_value = getattr(config, param_name, None)
+            if param_value is not None:
+                sampling_params[param_name] = param_value
+
         # override sampling params for validation
         if batch.meta_info.get("validate", False):
             sampling_params["top_p"] = config.val_kwargs.top_p
+            sampling_params["top_k"] = config.val_kwargs.top_k
             sampling_params["temperature"] = config.val_kwargs.temperature
 
         if "agent_name" not in batch.non_tensor_batch:
--- a/verl/trainer/config/_generated_ppo_megatron_trainer.yaml
+++ b/verl/trainer/config/_generated_ppo_megatron_trainer.yaml
@@ -204,6 +204,8 @@ actor_rollout_ref:
     temperature: 1.0
     top_k: -1
     top_p: 1
+    max_tokens: null
+    max_new_tokens: null
     prompt_length: ${oc.select:data.max_prompt_length,512}
     response_length: ${oc.select:data.max_response_length,512}
     dtype: bfloat16
--- a/verl/trainer/config/rollout/rollout.yaml
+++ b/verl/trainer/config/rollout/rollout.yaml
@@ -16,6 +16,12 @@ top_k: -1
 # Top-p sampling parameter. Default 1.0.
 top_p: 1
 
+# max number of tokens to generate for vllm.
+max_tokens: null
+
+# max number of tokens to generate for sglang.
+max_new_tokens: null
+
 # typically the same as data max prompt length
 # same as data.max_prompt_length if it exists
 prompt_length: ${oc.select:data.max_prompt_length,512}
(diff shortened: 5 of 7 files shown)


wuxibin89:
We already have `min(max_model_len, prompt_length+response_length)` as the upper bound for whole trajectory. The reason why we must calculate `max_new_token` dynamically in each turn is that we don't expect trajectory exceed the upper bound.

If we set a fixed `max_tokens` in each turn, then the whole trajectory may exceed the upper bound. For example,
- prompt_length+response_length=32k
- max_tokens=8k

before the last turn, current trajectory length is 30k, then after the last turn the trajectory may has length `30k+8k=40k`.

psyloy:
> We already have `min(max_model_len, prompt_length+response_length)` as the upper bound for whole trajectory. The reason why we must calculate `max_new_token` dynamically in each turn is that we don't expect trajectory exceed the upper bound.
> 
> If we set a fixed `max_tokens` in each turn, then the whole trajectory may exceed the upper bound. For example,
> 
> * prompt_length+response_length=32k
> * max_tokens=8k
> 
> before the last turn, current trajectory length is 30k, then after the last turn the trajectory may has length `30k+8k=40k`.

Thank you for your review. I still have some small confusions and would like to ask for advice.
I see that the current code has already introduced protection mechanisms to handle the aforementioned situation:
```python
self.config.max_model_len = min(
    get_max_position_embeddings(self.model_config.hf_config),
    self.config.prompt_length + self.config.response_length,
)
......
# This serves as a safety upper bound
max_possible_tokens = self.config.max_model_len - len(prompt_ids)
......
# Determine max_tokens from sampling_params or use configured response_length as default
if "max_tokens" in sampling_params:
   max_tokens = sampling_params.pop("max_tokens")
else:
   # Default to a calculation that considers configured lengths
   max_tokens = self.config.response_length + self.config.prompt_length - len(prompt_ids)
......
# Clamp max_tokens to the valid range [0, max_possible_tokens]
max_tokens = max(0, min(max_tokens, max_possible_tokens))
```
For example when  max_model_len=prompt_length+response_length:

case1 :
* prompt_length+response_length=32k
* max_tokens=8k
* max_model_len=32k
* prompt_ids=30k

(1) dynamic calculation of max_tokens :
```python
max_possible_tokens = self.config.max_model_len - len(prompt_ids)  # 32k-30k=2k
max_tokens = self.config.response_length + self.config.prompt_length - len(prompt_ids) # 32k-30k=2k
max_tokens = max(0, min(max_tokens, max_possible_tokens)) # min(8k, 2k)=2k
the last turn the trajectory may has length 30k+2k=32k
```
(2) Adopt the configured max_tokens :
```python
max_possible_tokens = self.config.max_model_len - len(prompt_ids)  # 32k-30k=2k
max_tokens = sampling_params.pop("max_tokens") # 8k
max_tokens = max(0, min(max_tokens, max_possible_tokens)) # min(8k, 2k)=2k
```
the last turn the trajectory may has length `30k+2k=32k`

case2 :
* prompt_length+response_length=32k
* max_tokens=8k
* max_model_len=32k
* prompt_ids=2k

(1) dynamic calculation of max_tokens :
```python
max_possible_tokens = self.config.max_model_len - len(prompt_ids)  # 32k-30k=30k
max_tokens = self.config.response_length + self.config.prompt_length - len(prompt_ids) # 32k-30k=30k
max_tokens = max(0, min(max_tokens, max_possible_tokens)) # min(30k, 30k)=30k
```
this first turn the trajectory may has length `30k+2k=32k`

(2) Adopt the configured max_tokens :
```python
max_possible_tokens = self.config.max_model_len - len(prompt_ids)  # 32k-30k=30k
max_tokens = sampling_params.pop("max_tokens") # 8k
max_tokens = max(0, min(max_tokens, max_possible_tokens)) # min(8k, 30k)=8k
```
this first turn the trajectory may has length `2k+8k=10k`

verl/trainer/config/_generated_ppo_trainer.yaml
```diff
@@ -195,6 +195,8 @@ actor_rollout_ref:
     temperature: 1.0
     top_k: -1
     top_p: 1
+    max_tokens: null
+    max_new_tokens: null
```
vermouth1992, line 199: Please only keep one of them because they serve the same purpose

vermouth1992 (changes requested):

verl/trainer/config/_generated_ppo_trainer.yaml
vermouth1992: Also, max_token should be configured to response_length in agent loop. In fact, it is not necessary to configure it by the user. This will cause confusion

vermouth1992: I would suggest removing these configurations.

vermouth1992: The server should not be aware of how user interacts with it. It should simply support using the largest model length if memory is enough, or configured to a smaller length

vermouth1992: All the controls should be done inside agent loop

PR #5039, opened 2026-01-25
[ci] feat: add npu workflow，e2e_sft_llm&model&reward_model_vllm

yyyy2000:
### What does this PR do?
Add 3 npu test workflow

.github/workflows/e2e_sft_llm_ascend.yml.yml
.github/workflows/model_ascend.yml
.github/workflows/reward_model_vllm_ascend.yml

Based on the GPU-based CI workflow:  e2e_sft_llm、model 、reward_model_vllm, we have added support for NPU.

### Checklist Before Starting

### Test

### API and Usage Example

```python
```

### Design & Code Changes

> [!IMPORTANT]


tests/special_e2e/sft/run_sft_engine.sh
```diff
@@ -85,7 +85,8 @@ MEGATRON_ENGINE_CONFIG="\
     engine.tensor_model_parallel_size=${TP_SIZE} \
     engine.pipeline_model_parallel_size=${PP_SIZE} \
     engine.virtual_pipeline_model_parallel_size=${VPP_SIZE} \
-    engine.context_parallel_size=${CP_SIZE}
+    engine.context_parallel_size=${CP_SIZE} \
+    +engine.override_transformer_config.context_parallel_size=${CP_SIZE} \
```
wucong25, line 89: Can the parameter `engine.override_transformer_config.context_parallel_size=${CP_SIZE}` be used to automatically set by engine.context_parallel_size=${CP_SIZE}?

verl/workers/rollout/replica.py
```diff
@@ -193,6 +195,7 @@ async def init_standalone(self):
             if not self.is_reward_model
             else f"rollout_reward_standalone_{self.replica_rank}",
             use_gpu=use_gpu,
+            device_name="cuda" if not is_torch_npu_available(check_device=False) else "npu",
```
wucong25, line 198: Currently, only CUDA and NPU are being checked. If there are other chip adaptations, wouldn’t that be a problem?

verl/workers/rollout/replica.py
```diff
@@ -163,6 +164,7 @@ async def init_colocated(self, resource_pool: RayResourcePool):
             if not self.is_reward_model
             else f"rollout_reward_colocate_{self.replica_rank}",
             use_gpu=use_gpu,
+            device_name="cuda" if not is_torch_npu_available(check_device=False) else "npu",
```
wucong25, line 167: It would be best to extract this into a function.

verl/workers/engine_workers.py
```diff
@@ -86,6 +90,10 @@ def __init__(self, config: TrainingWorkerConfig):
         # we use the one defined in model
         self.engine_config.use_remove_padding = self.model_config.use_remove_padding
 
+        if repatch is not None:
```
wucong25, line 93: Does this repatch appear in several places? If I repatch it, could there be any issues? It seems like before it would give an error if the patch was applied multiple times.

verl/workers/rollout/replica.py
yyyy2000: We originally implemented this using get_device_name, but the Ray-managed process prevented CUDA from being detected directly.
To handle this special case, here applied a separate workaround. 
If this scenario more common in the future, we will extract this into a function.

verl/workers/rollout/replica.py
yyyy2000: yes itis a known bug for ray
ENV RAY_EXPERIMENTAL_NOSET_*_VISIBLE_DEVICES is a useful method to solve
For ease of use, and to prevent issues when users haven’t set the necessary environment variables, this implementation was provided as a safeguard.

verl/workers/engine_workers.py
yyyy2000: Two places this patch exist:
1. I added in engine_worker, which is used for the SFT scenario. This part hadn’t been adapted before for npu.
2. megatronworker has the same logic. 
I’ve test the Megatron-Qwen3 script, and it goes through the patch inside megatronworker.
two patch won’t conflict.

tests/special_e2e/sft/run_sft_engine.sh
yyyy2000: It’s feasible, thanks for the suggestion — we’ll include it in our planning.

PR #5071, opened 2026-01-27
[rollout] feat: automatically resume generation on abort

wuxibin89:
### What does this PR do?

Automatically resume generation on abort, make AgentLoop agnostic to the rollout interruption.

### DIFF
--- a/tests/experimental/reward_loop/test_agent_reward_loop_colocate.py
+++ b/tests/experimental/reward_loop/test_agent_reward_loop_colocate.py
@@ -105,7 +105,7 @@ def test_agent_loop_reward_manager():
         replicas=agent_loop_manager.rollout_replicas,
     )
     checkpoint_manager.sleep_replicas()
-    reward_loop_manager = RewardLoopManager(config, rm_resource_pool=resource_pool)
+    reward_loop_manager = RewardLoopManager.create(config, rm_resource_pool=resource_pool)
 
     # 2. init test data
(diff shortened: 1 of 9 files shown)


verl/workers/rollout/base.py
```diff
@@ -100,3 +108,142 @@ def get_rollout_class(rollout_name: str, mode: str = "async") -> type[BaseRollou
...
+        priority: int = 0,
+    ) -> TokenOutput:
+        raise NotImplementedError
+
+    @abstractmethod
+    async def abort_all_requests(self):
+        """Abort all requests with partial rollout."""
+        self.resume_event.clear()
+
+    async def resume_all_requests(self):
```
ArronHZG, line 238: Looks like an abstractmethod's missing here.

wuxibin89: `resume_all_requests` is concrete function shared by all rollout.

ArronHZG (approved):

verl/workers/rollout/replica.py
```diff
@@ -29,19 +28,6 @@
 logger = logging.getLogger(__file__)
 
 
-class TokenOutput(BaseModel):
```
vermouth1992, line 32: Moving APIs randomly is strongly discouraged....

PeterSH6:
Partial Rollout is an algorithm logic, we should not implement it in the server side.

PR #5075, opened 2026-01-28
[trainer, rollout] fix: release rollout GPU memory after pre-train validation

none0663:
### What does this PR do?

This PR fixes an issue where GPU memory used by rollout during the first validation pass (val_metrics = self._validate()) was not released when val_before_train is enabled. After this change, rollout-related GPU resources are properly cleaned up following the  validation step, preventing unnecessary memory retention before training starts.

### Checklist Before Starting

### Test

### API and Usage Example

```python
```

### Design & Code Changes

> [!IMPORTANT]

### DIFF
--- a/verl/trainer/ppo/ray_trainer.py
+++ b/verl/trainer/ppo/ray_trainer.py
@@ -311,6 +311,11 @@ class RayPPOTrainer:
 
         self._create_dataloader(train_dataset, val_dataset, collate_fn, train_sampler)
 
+        # In async rollout mode with colocated checkpoint engine ("naive"), we may explicitly put
+        # rollout replicas to sleep (free weights/KV cache) after pre-train validation. If so, the
+        # next generation must re-sync weights once before calling generate.
+        self._rollout_needs_weights_sync: bool = False
+
     def _create_dataloader(self, train_dataset, val_dataset, collate_fn, train_sampler: Optional[Sampler]):
         """
         Creates the train and validation dataloaders.
@@ -550,7 +555,7 @@ class RayPPOTrainer:
 
         return gen_batch
 
-    def _validate(self, merged: bool = False):
+    def _validate(self, merged: bool = False, *, release_rollout_cache: bool = False):
         data_source_lst = []
         reward_extra_infos_dict: dict[str, list] = defaultdict(list)
 
@@ -606,7 +611,6 @@ class RayPPOTrainer:
                 test_output_gen_batch_padded = self.actor_rollout_wg.generate_sequences(test_gen_batch_padded)
             else:
                 test_output_gen_batch_padded = self.async_rollout_manager.generate_sequences(test_gen_batch_padded)
-                self.checkpoint_manager.sleep_replicas()
 
             # unpad
             test_output_gen_batch = unpad_dataproto(test_output_gen_batch_padded, pad_size=pad_size)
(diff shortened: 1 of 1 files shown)


flishwang:
Why should we update_weights after validate? Update_weights should be already executed after load_checkpoint or update_actor.

none0663:
> Why should we update_weights after validate? Update_weights should be already executed after load_checkpoint or update_actor.

You’re right that update_weights() is already called right after load_checkpoint() (and after any explicit actor weight update). However, in async rollout mode the rollout engine/replicas can be initialized lazily during the first _validate() (via generate_sequences). In that case, the initial update_weights() happens before the rollout engine exists, so the rollout side misses the weight broadcast.
Without calling update_weights() again after the pre-train validation, the first training step can fail when we call:
self.async_rollout_manager.generate_sequences(gen_batch_output)
with the error shown in the logs. My suspicion is that after sleep_replicas(), the parameters were not properly delivered to the rollout engine, so generate_sequences() crashes due to missing/stale weights. Re-running update_weights() after validation is a simple, safe way to ensure the rollout engines are synced before training starts.

Or maybe there is some un expected bug with `self.checkpoint_manager.sleep_replicas()`

WeChatWorkScreenshot_f495bc1f-d025-4e2b-8c99-4d564240ff51

wuxibin89:
> However, in async rollout mode the rollout engine/replicas can be initialized lazily during the first _validate() (via generate_sequences)

This should not happen, since we immediately `update_weights` right after `load_checkpoint`, before first `_validate`. If rollout not exist, the `update_weights` should failed.

none0663:
> > However, in async rollout mode the rollout engine/replicas can be initialized lazily during the first _validate() (via generate_sequences)
> 
> This should not happen, since we immediately `update_weights` right after `load_checkpoint`, before first `_validate`. If rollout not exist, the `update_weights` should failed.

@wuxibin89 
1. **Rollout init** – Agreed: rollout is not lazily created. We call `update_weights()` right after `load_checkpoint()` and before `_validate()`, so rollout is already up. The fix is not about lazy init.

2. **Actual issue** – In the original code, after validation rollout generation we never released rollout GPU memory (no offload / `sleep_replicas`). So during validation the GPU still held rollout weights + KV cache.

3. **OOM with colocated RM** – When validation uses a **colocated RM** to compute rewards, the GPU holds rollout + RM at the same time → OOM. Fix: after validation generation, call `sleep_replicas()` to free rollout; run RM for rewards; then call `update_weights()` before the training loop so rollout is ready for the first step.

4. **Why it didn’t show before** – OOM only appears when validation reward is computed with a colocated RM. With rule-based or non-colocated RM, memory pressure is lower, so the bug didn’t surface.

flishwang:
Maybe we should bind sleep_replicas() and update_weights() with generate_sequences:Whenever we call generate_sequences(), we call update_weights() first and then generate and then sleep.In this case, we should not call update_weights after load_checkpoint or update_actor.We also need to adjust the code for genmax as they call generate_sequences twice per round. Maybe we should modify update_weights to allow it being called when rollout is already waked up (thus do nothing and return).We also need to adjust the code for cases requires rollout sleeping without discarding the weights.发自我的荣耀手机-------- 原始邮件 --------发件人： none0663 ***@***.***>日期： 2026年1月29日周四 10:54收件人： verl-project/verl ***@***.***>抄送： Flish Wang ***@***.***>, Comment ***@***.***>主    题： Re: [verl-project/verl] [trainer, rollout] fix: release rollout GPU memory after pre-train validation (PR #5075)none0663 left a comment (verl-project/verl#5075)


However, in async rollout mode the rollout engine/replicas can be initialized lazily during the first _validate() (via generate_sequences)

This should not happen, since we immediately update_weights right after load_checkpoint, before first _validate. If rollout not exist, the update_weights should failed.

@wuxibin89

Rollout init – Agreed: rollout is not lazily created. We call update_weights() right after load_checkpoint() and before _validate(), so rollout is already up. The fix is not about lazy init.

Actual issue – In the original code, after validation rollout generation we never released rollout GPU memory (no offload / sleep_replicas). So during validation the GPU still held rollout weights + KV cache.

OOM with colocated RM – When validation uses a colocated RM to compute rewards, the GPU holds rollout + RM at the same time → OOM. Fix: after validation generation, call sleep_replicas() to free rollout; run RM for rewards; then call update_weights() before the training loop so rollout is ready for the first step.

Why it didn’t show before – OOM only appears when validation reward is computed with a colocated RM. With rule-based or non-colocated RM, memory pressure is lower, so the bug didn’t surface.

yyDing1:
the expected behavior during validation:

1. Synchronize weights and wake up the rollout model
This is done via `update_weights` and works as expected.
2. Run rollout generation: This part is also working correctly.
3. **sleep rollout model (currently missing in the implementation)**, for two purposes:
- To release/reset the rollout model's prefix (KV) cache. (the validation kv-cache is useless for subsequent policy gen)
- In the co-located reward model setup, we should sleep rollout model and wake up the reward model for reward computation.
4. All subsequent steps behave as expected.

As stated above, **there is a potential issue:**
After every validation finishes, the prefix kv cache is never cleared, since `sleep_replicas` is not invoked.

---

**Proposed Solution**

During validation, the rollout model should be put to sleep and then woken up at the end of validation, ensuring that the validation process has no impact on the rollout model's state.

PR #4632, opened 2025-12-22
[megatron] feat: LoRA adapter only refit (TensorLoRARequest)

HollowMan6:
### What does this PR do?

lora-performance
lora-critic-val-score
lora-actor-plus-rollout-mismatch

### Checklist Before Starting

- [X] Search for similar PRs. Paste at least one query link here: ...
- [X] Format the PR title as `[{modules}] {type}: {description}` (This will be checked by the CI)
  - `{modules}` include `fsdp`, `megatron`, `sglang`, `vllm`, `rollout`, `trainer`, `ci`, `training_utils`, `recipe`, `hardware`, `deployment`, `ray`, `worker`, `single_controller`, `misc`, `perf`, `model`, `algo`, `env`, `tool`, `ckpt`, `doc`, `data`, `cfg`, `reward`
  - If this PR involves multiple modules, separate them with `,` like `[megatron, fsdp, doc]`
  - `{type}` is in `feat`, `fix`, `refactor`, `chore`, `test`
  - If this PR breaks any API (CLI arguments, config, function signature, etc.), add `[BREAKING]` to the beginning of the title.
  - Example: `[BREAKING][fsdp, megatron] feat: dynamic batching`

### Test

> For changes that can not be tested by CI (e.g., algorithm implementation, new model support), validate by experiment(s) and show results like training curve plots, evaluation results, etc.

### API and Usage Example

> Demonstrate how the API changes if any, and provide usage example(s) if possible.

```python
# Add code snippet or script demonstrating how to use this
```

### Design & Code Changes

> Demonstrate the high-level design if this PR is complex, and list the specific changes.

### Checklist Before Submitting

> [!IMPORTANT]
> Please check all the following items before requesting a review, otherwise the reviewer might deprioritize this PR for review.

- [X] Read the [Contribute Guide](https://github.com/volcengine/verl/blob/main/CONTRIBUTING.md).
- [X] Apply [pre-commit checks](https://github.com/volcengine/verl/blob/main/CONTRIBUTING.md#code-linting-and-formatting): `pre-commit install && pre-commit run --all-files --show-diff-on-failure --color=always`
- [X] Add / Update [the documentation](https://github.com/volcengine/verl/tree/main/docs).
- [X] Add unit or end-to-end test(s) to [the CI workflow](https://github.com/volcengine/verl/tree/main/.github/workflows) to cover all the code. If not feasible, explain why: ...
[...]

PR #5115, opened 2026-01-30
[fsdp] feat: Merge lora in fsdp training to speed up rollout

amzfang:
### What does this PR do?

In current lora RL training workflows, engines like vLLM often utilize unmerged LoRA inference, which incurs a latency penalty relative to full fine-tuned models. While enabling CUDA graphs can dampen this overhead, merging LoRA weights directly into the base model during rollout can further maximize inference throughput, achieving base model efficiency by eliminating specialized adapter computation overhead. Beyond performance, weight merging serves as a critical workaround for model families that lack native unmerged LoRA support in vLLM, bypassing framework limitations and accelerating development without waiting for upstream software updates.

During the feature development, this PR fixed a bug in misalignment between ref_log_probs/old_log_probs vs current log probs when using new engine_worker for training (that is, `trainer.use_legacy_worker_impl=disable`)

The bug is that `no_padding_2_padding` always slices [:, -max_response_len-1:-1] from the padded tensor which does not pad the prompts on left side. The correction is to keep the unpadding logic the same for ref-log-probs/old-log-probs and current log-probs. This problem occurs not only for lora training but also for full parameter training. Refer to `verl/workers/utils/padding.py` for details. See detailed results in comment below.

### Checklist Before Starting

- [ x] Search for similar PRs. Paste at least one query link here: https://github.com/verl-project/verl/issues?q=is%3Aissue%20state%3Aopen%20lora%20slow
- [x ] Format the PR title as `[{modules}] {type}: {description}` (This will be checked by the CI)

### Test

unit test: 
- tests/utils/test_fsdp_lora_merge.py
- tests/utils/test_normalize_peft_param_name.py
- tests/utils/test_normalize_peft_param_name_on_cpu.py
- tests/utils/test_padding_on_cpu.py (for padding fix)

manual test:
- examples/grpo_trainer/run_qwen3-4b_gsm8k_grpo_lora_merge.sh

### API and Usage Example

as shown in `examples/grpo_trainer/run_qwen3-4b_gsm8k_grpo_lora_merge.sh`

```python
    ...
    +actor_rollout_ref.model.lora.merge=True \
    actor_rollout_ref.model.lora_rank=32 \
    actor_rollout_ref.model.lora_alpha=64 \
    trainer.use_legacy_worker_impl=disable \
    ...
```

### Design & Code Changes

Merging lora to speed up rollout

We optimize the rollout efficiency for fsdp training backend through the following workflow:

1. offload the model weights to cpu  
2. temporally merging the Lora adapter weights to base model before rollout
3. synchronize the merged model weights from fsdp to vllm via NCCL
4. restore the model weights from cpu
5. perform rollout on merged model in vllm without any additional computation caused by lora adapter 

The key challenge here is how to run these steps as fast and memory-efficient as possible, leveraging the GPU’s fast computation and communication mechanisms. Specifically, Lora adapter together with the base model weights are sharded across GPUs during FSDP training, so we need to unshard (i.e. gather) the parameters of Lora adapter and base model before merging Lora into base model, preferably on GPUs. Due to GPU memory limit, we cannot unshard the full model at once for large models (e.g. >40B) etc. 

Our solution only merges lora weights layer-by-layer to keep low memory footprint, similarly to what FSDP does in forward/backward functions. For each model layer, we first unshard the model and lora; merge lora into base model and then reshard the merged model to release the GPU memory required for full model layer. After merging is done, we can leverage NCCL to sync model weights to vllm. 

> [!IMPORTANT]


amzfang:
when training qwen2.5_3b on gsm8k data 1 epoch [training_config.yaml](https://github.com/user-attachments/files/24978710/training_config.yaml)

before fixing:
  - actor/pg_loss: 0.036 (positive, bad)
  - val-core/openai/gsm8k/acc: 0.5497 (no improvement from 0.5504)
  - actor/ppo_kl: 0.0022

after fixing:
  - actor/pg_loss: 0.0062 (much smaller)
  - val-core/openai/gsm8k/acc: 0.8279 (huge improvement!)
  - actor/ppo_kl: 0.00012
  Training config is attached.

HollowMan6 (changes requested):
Could you please resolve the merge conflict in this PR?

verl/trainer/main_ppo.py
```diff
@@ -279,6 +281,13 @@ def run(self, config):
         pprint(OmegaConf.to_container(config, resolve=True))
         OmegaConf.resolve(config)
 
+        train_dir = Path(config.trainer.default_local_dir)
+        print(f"Saving config file to {train_dir}")
+        train_dir.mkdir(exist_ok=True, parents=True)
+
+        cfg_text = OmegaConf.to_yaml(config)
+        io_utils.save_text(cfg_text, train_dir.joinpath("training_config.yaml"))
+
```
HollowMan6: Why this is added here?

verl/workers/rollout/vllm_rollout/vllm_async_server.py
```diff
@@ -345,11 +345,13 @@ async def launch_server(self, master_address: str = None, master_port: int = Non
...
         megatron_lora = True
-        if self.model_config.lora.get("merge", False):
+        merge_lora = self.model_config.lora.get("merge", False)
+        if merge_lora:
             lora_rank = 0
         if lora_rank <= 0:
             megatron_lora = False
-            lora_rank = self.model_config.lora_rank
+            lora_rank = self.model_config.lora_rank if not merge_lora else 0
+
```
HollowMan6: Better to reorganize the logic here (e.g. put set `lora_rank` as 0 if `merge_lora` to the end), otherwise it's confusing to have logic like this

HollowMan6 (changes requested):
In addition, maybe it's a good time to reconsider putting all the lora related settings under the `.lora` key and deprecate any effective lora settings outside `.lora` at this stage. Right now, `.lora` is considered to be megatron-only, but since this PR introduces merge for FSDP backend as well, it will become a mix-in which will be generally confusing.

Could you also update the docs at `docs/advance/ppo_lora.rst` for this feature?

fangliuwh:
> In addition, maybe it's a good time to reconsider putting all the lora related settings under the `.lora` key and deprecate any effective lora settings outside `.lora` at this stage. Right now, `.lora` is considered to be megatron-only, but since this PR introduces merge for FSDP backend as well, it will become a mix-in which will be generally confusing.
> 
> Could you also update the docs at `docs/advance/ppo_lora.rst` for this feature?

agree it is cleaner to put all lora settings under `lora` key. Can we create another pull request to handle this? I can pick this up after this PR. 

For the doc, yes, will update `docs/advance/ppo_lora.rst`

verl/trainer/main_ppo.py
fangliuwh: It makes reproduction of an experiment easier - such config file captures the complete configurations to reproduce an experiment. We do already print out the configs in log files but this is more for human debugging.

verl/workers/rollout/vllm_rollout/vllm_async_server.py
fangliuwh: Sounds good. Until we refactor lora configs, we can simplify it as follows:
- set lora rank to be model_config.lora.rank or model_config.lora_rank (whichever is positive integer)
- when  merge_lora=True, reset lora rank = 0.

verl/trainer/main_ppo.py
HollowMan6: I think it's better to keep this out of this PR as it's kind of unrelated to the PR, and you can raise a separate PR for this addition, as I believe this behavior needs more discussion, e.g. how to design the config for the file naming/turn this off.

Plus, actually, current behavior has already print the training config to console by default, which should already be kind of enough for debugging at least for me.

amzfang: Ok, will move this out as it is unrelated.

verl/utils/io_utils.py
```diff
@@ -0,0 +1,120 @@
...
+    logger.info(f"Save data with {len(df)} rows and {len(df.columns)} columns to csv file {filename}")
+    if log:
+        logger.info(f"The contents are:\n{df.to_string(index=False)}")
+    df.to_csv(filename, index=False)
+
+
+def create_dir_if_not_exist(filename: str):
+    dir_path = os.path.dirname(filename)
+    if not os.path.exists(f"{dir_path}"):
+        pathlib.Path(dir_path).mkdir(parents=True, exist_ok=True)
```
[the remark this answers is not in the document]
HollowMan6: @amzfang Better to remove this file here if it's not used anywhere in this PR.

verl/workers/utils/padding.py
```diff
@@ -72,35 +73,43 @@ def left_right_2_no_padding(data: TensorDict) -> TensorDict:
...
+def no_padding_2_padding(tensor: torch.Tensor, data: TensorDict) -> torch.Tensor:
+    """Slice response from unpad model output.
 
     Args:
-        nested_tensor: NestedTensor with no-padding format
-        data: TensorDict with
-        - Tensor includes NestedTensors like "input_ids", "loss_mask", "position_ids"
-        - NonTensorData includes "max_seq_len", "max_response_len", "indices"
+        tensor: model output tensor of shape [bsz, 1]
+        data: TensorDict with "prompt_ids", "response_ids", "attention_mask"
```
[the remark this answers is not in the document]
HollowMan6: @amzfang Please also address this one, or explain it in a clearer way (e.g. without `"` and "_")

verl/protocol.py
```diff
@@ -462,6 +462,29 @@ def print_size(self, prefix=""):
...
+
+        if self.non_tensor_batch is not None:
+            for key, numpy_array in self.non_tensor_batch.items():
+                print(
+                    f"non_tensor_batch field (numpy array): {key=}, array (size: {numpy_array.shape}), "
+                    f"type: {type(numpy_array[0])}"
+                )
+
+        print(f"--- end of {desc} ---")
+
```
HollowMan6: Is this method used anywhere? If not, let's remove it to keep everything clean.

HollowMan6 (changes requested):

verl/workers/utils/padding.py
```diff
@@ -72,35 +73,43 @@ def left_right_2_no_padding(data: TensorDict) -> TensorDict:
...
-    Convert NestedTensor from no-padding to right padding format.
+def no_padding_2_padding(tensor: torch.Tensor, data: TensorDict) -> torch.Tensor:
+    """Slice response from unpad model output.
 
     Args:
-        nested_tensor: NestedTensor with no-padding format
-        data: TensorDict with
-        - Tensor includes NestedTensors like "input_ids", "loss_mask", "position_ids"
-        - NonTensorData includes "max_seq_len", "max_response_len", "indices"
+        tensor: model output tensor of shape [bsz, 1]
```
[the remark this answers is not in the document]
HollowMan6: @amzfang I think this shall also need to be accurately documented

HollowMan6:
BTW, it would be appreciated if you could add the figure you mentioned to the `Design & Code Changes` section in the PR description: https://github.com/verl-project/verl/pull/5115#issue-3873346480 Right now it seems to be missing.

amzfang:
> BTW, it would be appreciated if you could add the figure you mentioned to the `Design & Code Changes` section in the PR description: [#5115 (comment)](https://github.com/verl-project/verl/pull/5115#issue-3873346480) Right now it seems to be missing.

Thanks! The workflow is fairly linear and can be explained in the text. I updated the description to avoid any confusion.

examples/grpo_trainer/run_qwen3-4b_gsm8k_grpo_lora_merge.sh
```diff
@@ -0,0 +1,71 @@
...
+    data.train_batch_size=128 \
+    data.max_prompt_length=1024 \
+    data.max_response_length=1024 \
+    data.filter_overlong_prompts=True \
+    data.truncation='error' \
+    data.shuffle=False \
+    actor_rollout_ref.model.path=Qwen/Qwen3-4B \
+    actor_rollout_ref.model.use_remove_padding=True \
+    actor_rollout_ref.model.enable_gradient_checkpointing=True \
+    +actor_rollout_ref.model.lora.merge=True \
```
HollowMan6, line 30: I don't think we need `+` here as `merge` is already a existing field, right?

ISEEKYAN:
@HollowMan6 megatron also support merge=True and merge=False, right? do we need also support the two modes in megatron?

HollowMan6:
> @HollowMan6 megatron also support merge=True and merge=False, right? do we need also support the two modes in megatron?


Yes, for megatron, both `merge=True` and `merge=False` has already been supported in #4632, for either legacy workers or new model engine. `actor_rollout_ref.model.lora.merge` was actually initially introduced for megatron.

wuxibin89:
> The bug is that no_padding_2_padding always slices [:, -max_response_len-1:-1] from the padded tensor which does not pad the prompts on left side. The correction is to keep the unpadding logic the same for ref-log-probs/old-log-probs and current log-probs. This problem occurs not only for lora training but also for full parameter training. Refer to verl/workers/utils/padding.py for details. See detailed results in comment below.

@amzfang This is not a bug, `pad_input` properly pad the prompts on left side.
```python
import torch
from flash_attn.bert_padding import pad_input, unpad_input

max_seq_len, max_response_len = 5, 3
input_ids = torch.tensor(
    [[0, 1, 2, 3, 4],
    [0, 1, 2, 3, 4]],
    dtype=torch.long
)
attention_mask = torch.tensor(
    [[0, 1, 1, 1, 1],
    [1, 1, 1, 0, 0]],
    dtype=torch.long
)
old_log_probs = torch.nested.as_nested_tensor(
    [torch.tensor([0.1, 0.2, 0.3, 0.4]),
    torch.tensor([0.5, 0.6, 0.7])],
    layout=torch.jagged
)

input_ids_rmpad, indices, cu_seqlens, *_ = unpad_input(input_ids.unsqueeze(-1), attention_mask)

full_values = pad_input(
    hidden_states=old_log_probs.values().unsqueeze(-1),
    indices=indices,
    batch=batch_size,
    seqlen=max_seq_len,
)

full_values.squeeze(-1)
# output:
# tensor([[0.0000, 0.1000, 0.2000, 0.3000, 0.4000],
#        [0.5000, 0.6000, 0.7000, 0.0000, 0.0000]])
```

wuxibin89:
> when training qwen2.5_3b on gsm8k data 1 epoch [training_config.yaml](https://github.com/user-attachments/files/24978710/training_config.yaml)
> 
> before fixing:
> 
> * actor/pg_loss: 0.036 (positive, bad)
> * val-core/openai/gsm8k/acc: 0.5497 (no improvement from 0.5504)
> * actor/ppo_kl: 0.0022
> 
> after fixing:
> 
> * actor/pg_loss: 0.0062 (much smaller)
> * val-core/openai/gsm8k/acc: 0.8279 (huge improvement!)
> * actor/ppo_kl: 0.00012
>   Training config is attached.

Didn't observe this problem with Qwen2.5-1.5B-Instruct:
image
image

fangliuwh:
Can you please check the [unit tests](https://github.com/verl-project/verl/pull/5115/changes#diff-7847a7964cd80d815f300eee18b23d599a786a89289a97528f620cddb6406a3d) for the old implementation? It failed before. I don't have access to my old env anymore and will try it later. - additionally it would be good to keep the padding implementation the same for ref-log-probs, old-log-probs and current log-probs.  
To check if the log probs are aligned or not, can you check if [ratios](https://github.com/amzfang/verl/blob/f5914e7b7fdb3fd1655afa2473f0175793b0e033/verl/trainer/ppo/core_algos.py#L1213) are close to 1 or not.

wuxibin89:
> Can you please check the [unit tests](https://github.com/verl-project/verl/pull/5115/changes#diff-7847a7964cd80d815f300eee18b23d599a786a89289a97528f620cddb6406a3d) for the old implementation? It failed before.

@fangliuwh Yes, I can reproduce failed before this commit. The problem is that in the unit test, input_ids is left padding, while in [agent loop](https://github.com/verl-project/verl/blob/main/verl/experimental/agent_loop/agent_loop.py#L518-L519), the input_ids(=prompts+responses) is left-right padding: `prompts` is left padding and `responses` is right padding.

Nevertheless, I agree that we should keep padding implementation the same for ref-log-probs, old-log-probs and current log-probs. I just want to find why the performance metrics is not as expected in your case.

vermouth1992:
I guess we should revert no_padding_2_padding implementation because it changes the key of the data structure @wuxibin89 @amzfang

vermouth1992:
Our implementation assumes left-right padding

PR #5147, opened 2026-01-31
[worker] feat: Avoid redundant base weight sync when engine doesn't sleep

JohnConnor123:
### What does this PR do?

Fixed unnecessary base weight syncing for **LoRA sleep_level=2** when `rollout.free_cache_engine=False`

> **Impact:** up to **26% faster end-to-end** training in some cases

When `free_cache_engine=False`, the rollout engine is not actually sleeping in a way that destroys base weights, so syncing base weights is redundant work.

### Changes
- Only collect send `base_model_params` when `sleep_level==2` **and** `rollout.free_cache_engine=True`.

### Perf (Qwen3-1.7B, 20 steps, mean over steps 1-20, tested on 2 rtx3090)

| metric | baseline (ms/token) | patched (ms/token) | delta (ms/token) | delta % |
| timing_per_token_ms/gen | 0.845 | 0.559 | -0.286 | -33.9% |
| timing_per_token_ms/update_actor | 0.251 | 0.252 | 0.001 | +0.48% |
| timing_per_token_ms/all | 1.097 | 0.811 | -0.286 | **-26.02%** |

image

### Checklist
- [x] Fix applied
- [x] Local perf run   plot attached
- [ ] Added regression test (not included in this PR)

### DIFF
--- a/verl/workers/fsdp_workers.py
+++ b/verl/workers/fsdp_workers.py
@@ -705,6 +705,7 @@ class ActorRolloutRefWorker(Worker, DistProfilerExtension):
         # When sleep_level=2, base model weights are destroyed during each sleep cycle.
         # separately collect and update LoRA weights and base model weights through their respective interfaces.
         # Here: params contains LoRA weights, base_model_params contains base model weights.
+        # Only needed if the rollout engine actually sleeps/frees weights (free_cache_engine=True).
         if (
             peft_config is not None
             and getattr(self.rollout, "sleep_level", None) == 2
--- a/verl/workers/engine_workers.py
+++ b/verl/workers/engine_workers.py
@@ -615,7 +615,9 @@ class ActorRolloutRefWorker(Worker, DistProfilerExtension):
             # main memory can trade sync time to avoid OOM
             self.rollout.sleep_level = 1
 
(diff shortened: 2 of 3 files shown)


verl/workers/fsdp_workers.py
```diff
@@ -705,7 +705,11 @@ async def rollout_mode(self):
         # When sleep_level=2, base model weights are destroyed during each sleep cycle.
         # separately collect and update LoRA weights and base model weights through their respective interfaces.
         # Here: params contains LoRA weights, base_model_params contains base model weights.
-        if peft_config is not None and getattr(self.rollout, "sleep_level", None) == 2:
+        if (
+            peft_config is not None
+            and getattr(self.rollout, "sleep_level", None) == 2
+            and self.config.rollout.free_cache_engine
```
HollowMan6, line 712: Could you update this for `megatron_workers.py` and `engine_workers.py` as well?

BTW I'm not sure if this flag is still reliable these days, it can easily be broken and we also need to be careful about this: https://github.com/verl-project/verl/pull/1464 https://github.com/verl-project/verl/pull/3442

JohnConnor123: Done

PR #5152, opened 2026-01-31
[rollout,vllm] Fix DP args and local_rank for Ray NOSET_VISIBLE_DEVICES

JohnConnor123:
## Summary
- Fix A: ensure vLLM receives DP CLI args when `data_parallel_size > 1` (not only when EP is enabled).
- Fix B: when running under Ray with `RAY_EXPERIMENTAL_NOSET_CUDA_VISIBLE_DEVICES=1` and `data_parallel_size > 1`,
  compute the correct vLLM-local `LOCAL_RANK` (local rank within TP×PP) to avoid `DP adjusted local rank ... out of bounds`
  and incorrect GPU binding.

## Context / Motivation
In DP configs (typically `dp=2,tp=1,pp=1,nnodes=1`):
- vLLM could start without DP settings if the condition was tied only to EP.
- Under Ray NOSET, using `ray.get_runtime_context().get_accelerator_ids()` to derive local rank can produce a rank that becomes
  out-of-bounds after vLLM DP adjustment.

## Checklist
- [x] Bug fixed
- [x] Repro / test plan included
- [ ] Regression test (real Ray+vLLM e2e)

### DIFF
--- a/verl/workers/rollout/vllm_rollout/vllm_rollout.py
+++ b/verl/workers/rollout/vllm_rollout/vllm_rollout.py
@@ -41,7 +41,7 @@ from torch.multiprocessing.reductions import reduce_tensor
 
 from verl import DataProto
 from verl.third_party.vllm import VLLM_SLEEP_LEVEL, get_version
-from verl.utils.device import get_device_id, get_device_name, get_torch_device, is_support_ipc
+from verl.utils.device import get_device_id, get_device_name, get_torch_device, is_npu_available, is_support_ipc
 from verl.utils.torch_dtypes import PrecisionType
 from verl.workers.config import HFModelConfig, RolloutConfig
 from verl.workers.rollout.base import BaseRollout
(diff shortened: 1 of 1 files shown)


verl/workers/rollout/vllm_rollout/vllm_rollout.py
```diff
@@ -78,6 +78,27 @@ def __init__(
...
+                assert tp_pp > 0, (
+                    "Expected tensor_model_parallel_size * pipeline_model_parallel_size > 0, "
+                    f"got tp={self.config.tensor_model_parallel_size}, pp={self.config.pipeline_model_parallel_size}"
+                )
+                local_rank = rank % tp_pp
+            else:
+                device_name = "NPU" if is_npu_available else "GPU"
+                local_rank = int(ray.get_runtime_context().get_accelerator_ids()[device_name][0])
+            os.environ["LOCAL_RANK"] = str(local_rank)
+            get_torch_device().set_device(int(local_rank))
```
wuxibin89, line 100: ServerAdapter is just a thin client in trainer(fsdp/megatron) process, we now always use vllm's MP Executor to launch worker process:
https://github.com/verl-project/verl/blob/main/verl/workers/rollout/vllm_rollout/vllm_async_server.py#L271

PR #5178, opened 2026-02-02
Guard Harmony encoding download in vLLM rollout

khazic:
In offline environments, vLLM’s Harmony encoding download can fail and block rollout even for non‑GPT‑OSS models (e.g., Qwen). This change makes the download opt‑in behind an environment flag.

What

Guard get_encoding() behind VERL_DISABLE_HARMONY (default off unless explicitly set).
No behavior change unless the env var is set.
How to use
Set VERL_DISABLE_HARMONY=1 to skip Harmony vocab download.

### DIFF
--- a/recipes_custom/RLVR_ABCDE_dense/create_dataset.py
+++ b/recipes_custom/RLVR_ABCDE_dense/create_dataset.py
@@ -1,0 +1,198 @@
+# Copyright 2024 Bytedance Ltd. and/or its affiliates
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


vermouth1992:
What's the difference between VERL_USE_GPT_OSS and VERL_DISABLE_HARMONY? Shall we unify them?

vermouth1992:
I guess setting VERL_USE_GPT_OSS=0 also skip download

khazic:
In my setup, I already had export VERL_USE_GPT_OSS=0, but the runtime still tried to download the Harmony vocab in an offline environment and failed. The root cause is in the code path: get_encoding() was being invoked without a hard guard, so VERL_USE_GPT_OSS=0 was not sufficient to prevent the call in all cases (some versions/branches effectively call it at import time or without checking the flag).

That’s why I added an explicit opt‑out guard:

if os.getenv("VERL_DISABLE_HARMONY", "0") != "1":
    if get_encoding is not None and os.getenv("VERL_USE_GPT_OSS", "0") == "1":
        get_encoding()

Semantically, VERL_USE_GPT_OSS means “I’m using GPT‑OSS, so I need Harmony encoding.”
VERL_DISABLE_HARMONY is a hard safety switch for offline/restricted environments to guarantee no download happens even if some path would otherwise call get_encoding().

PR #5185, opened 2026-02-03
[algo] feat: support rollout router replay in MegatronEngine

xhx1022:
### What does this PR do?

This PR introduces Rollout Router Replay support within the MegatronEngine. This feature enables the replay of routing decisions made during the rollout phase, ensuring consistency between generation and training/optimization steps in RLHF workflows.

### Experiments
e26d1952-ac1e-4490-92c2-ccd1c4edaff4

We evaluated the performance of the R3 configuration on the Qwen/Qwen3-30B-A3B. Compared to the baseline, enabling the R3 configuration leads to a much lower and more stable KL divergence.

### DIFF
--- a/verl/models/mcore/model_forward.py
+++ b/verl/models/mcore/model_forward.py
@@ -198,7 +198,7 @@ def gptmodel_forward_no_padding(
             }
             model_kwargs["labels"] = args["label"].contiguous()
             model_kwargs["loss_mask"] = args["loss_mask"].contiguous()
-        if logits_processor_args and 'loss_mask' in logits_processor_args:
+        if logits_processor_args and "loss_mask" in logits_processor_args:
             logits_processor_args.pop("loss_mask")
 
         # For VLM model, need to pass bshd format `input_ids` and `attention_mask`.
@@ -252,7 +252,7 @@ def gptmodel_forward_no_padding(
             }
             model_kwargs["labels"] = args["label"].contiguous()
             model_kwargs["loss_mask"] = args["loss_mask"].contiguous()
-        if logits_processor_args and 'loss_mask' in logits_processor_args:
+        if logits_processor_args and "loss_mask" in logits_processor_args:
             logits_processor_args.pop("loss_mask")
 
         output_orig = model(


verl/workers/engine_workers.py
```diff
@@ -35,6 +35,7 @@
 from verl.utils.device import get_device_name, set_expandable_segments
 from verl.utils.distributed import initialize_global_process_group_ray
 from verl.utils.flops_counter import FlopsCounter
+from verl.utils.megatron.router_replay_patch import RouterReplay, RouterReplayAction, apply_router_replay_patch
```
vermouth1992: Can we move this into megatron transformer_impl?

verl/trainer/config/engine/megatron.yaml
```diff
@@ -95,3 +95,5 @@ forward_only: False
 
 # Mixed precision training param dtype
 dtype: bfloat16 # ["bfloat16", "float16"]
+
+router_replay_mode: disabled
```
wuxibin89: Please reuse router replay config: https://github.com/verl-project/verl/blob/main/verl/trainer/config/actor/actor.yaml#L236-L253

PeterSH6: It's infeasible to use actor config in model engine design because the megatron_engine didn't have a way to pass actor config. So we have to add it in the megatron engine config field.

verl/workers/engine_workers.py
PeterSH6: If we move it into the transformer_impl, the infer_batch will also affect ref_logp computation, which doesn't need router replay

PeterSH6:
It would be better to add a Megatron CI to guard the router replay feature.

verl/workers/engine_workers.py
wuxibin89: engine_workers.py is agnostic to specific model engine backend, it's better to move router_replay_patch into megatron transformer_impl and add a context manager in engine_workers.py to enable/disable router replay.

```python
    @register(dispatch_mode=make_nd_compute_dataproto_dispatch_fn(mesh_name="ref"))
    @DistProfiler.annotate(color="olive", role="ref_compute_log_prob")
    def compute_ref_log_prob(self, data: TensorDict) -> TensorDict:
        with no_router_replay():
            output = self.ref.infer_batch(data=data)
        return output.cpu() if output is not None else None
```

verl/trainer/config/engine/megatron.yaml
wuxibin89: Should we move the origin router replay config into model engine?

verl/workers/engine_workers.py
xhx1022: > Can we move this into megatron transformer_impl?

done

verl/trainer/config/engine/megatron.yaml
PeterSH6: @wuxibin89 We cannot move the RouterReplay Config to model engine as the legacy implementation still uses this config in the actor.
So we have to keep both of them until we delete the legacy implementation

wuxibin89: ok, got it.

gongshaotian:
Do you have end-to-end performance data? And how much has the training time increased

PR #5186, opened 2026-02-03
[tool] feat: verl integrate msprobe data collection

### LINKED ISSUE #5808: RFC: Integrating Debugging Data Collection into VERL
### Feature request

This RFC proposes a configurable, opt-in, low-intrusion precision debugging layer for VERL. When enabled through external configuration, VERL will collect intermediate numerical data at selected execution stages and persist it for offline analysis. Typical use cases include debugging training-serving mismatches, NaN/Inf propagation, overflow, dtype inconsistencies, and operator-level numerical differences.

When disabled, the feature must introduce no dependency requirement, no runtime overhead, and no behavior change.

The design is backend-pluggable. The initial target backend is `msprobe` / MindStudio Probe PrecisionDebugger, but the VERL-side integration should remain backend-agnostic.

### Motivation

Debugging numerical issues in large-scale RLHF and post-training pipelines is expensive and slow. Problems often appear only in distributed or mixed-precision runs, and existing logs are usually insufficient to localize the failure. Developers need a way to inspect intermediate tensors, statistics, and gradients at critical stages without permanently instrumenting the training stack.

`msprobe` already provides a practical precision debugging capability across multiple backends, including vLLM (eager), FSDP, and Megatron. It supports:

- Collecting intermediate model data, such as per-layer outputs and log-probability differences.
- Tensor-level collection for detailed debugging.
- Lightweight statistics-only collection for faster runs and smaller dumps.
- Dynamic `start` / `stop` control around selected stages.
- Post-processing workflows such as visualization and comparison.

This RFC proposes a VERL-native integration layer so that these capabilities can be enabled through standard VERL configuration and stage boundaries.

## Goals

- Support intermediate data collection for the following stages, each independently configurable:
  - Rollout stage, centered on `generate_sequences`
  - Training forward stage, including loss/logprob/value-related forward computation
  - Reference model stage, centered on `compute_ref_log_prob`
  - Training backward stage, centered on `backward`
  - Parameter update stage, centered on `optimizer.step` and before/after update state
- Keep the feature disabled by default.
- Guarantee that the disabled path has no dependency on the backend package and no meaningful runtime overhead.
- Drive behavior entirely through external configuration, including:
  - Sampling by step, rank, and stage
  - White/blacklist-based module or operator selection
  - Structured output directories and metadata such as `global_step`, `uid`, `rank`, and `stage`
- Expose the integration as a pluggable backend interface rather than hard-coding `msprobe` into VERL.

## Non-Goals

- Replacing existing profilers or performance tracing tools.
- Standardizing the internal configuration schema of all third-party debuggers.
- Guaranteeing identical instrumentation granularity across all execution backends in the first iteration.
- Solving all service-side rollout instrumentation gaps immediately for every runtime.

## Design Principles

- Minimal intrusion: integrate at stable stage boundaries rather than scattering hooks throughout the codebase.
- Backend abstraction: VERL owns stage lifecycle and metadata; the debugger backend owns collection details.
- Zero-cost disabled path: no backend import, initialization, or conditional work on the hot path unless explicitly enabled.
- Incremental rollout: land the abstraction first, then the backend, then stage-specific integrations.

### 1. Add a generic backend interface

Introduce a lightweight precision debugger backend interface:

```python
class PrecisionDebuggerBackend:
    def enabled(self) -> bool: ...

    @contextmanager
    def stage(self, name: str, *, model=None, meta: dict | None = None): ...

    def step(self, *, meta: dict | None = None): ...
```

Planned implementations:

- `NoopPrecisionDebugger`: default implementation; all methods are no-ops.
- `MsprobePrecisionDebugger`: wraps `msprobe` lifecycle APIs such as `start`, `stop`, and per-step progression.

The VERL codebase should depend only on this interface.

### 2. Keep `msprobe` as an optional dependency

The `msprobe` package should be imported only when the debugger is enabled and the selected backend is `msprobe`.

Recommended packaging model:

- Default VERL install does not require `mindstudio-probe`
- Optional install path such as `pip install verl[msprobe]`

This avoids polluting the default dependency graph and ensures that users who do not need the feature are unaffected.

### 3. Instrument stage boundaries at the worker layer

The preferred integration strategy is to wrap stage entry points at the worker layer. This matches VERL's execution model, where the driver in `verl/trainer/ppo/ray_trainer.py` dispatches stage execution through worker groups.

Worker-layer integration provides a single place to cover multiple implementations, including FSDP, Megatron, and engine workers.

Initial integration points:

- Rollout:
  - `verl/workers/{megatron_workers,fsdp_workers}.py:generate_sequences`
- Reference model:
  - `verl/workers/{megatron_workers,engine_workers,fsdp_workers}.py:compute_ref_log_prob`
- Actor update:
  - `verl/workers/{megatron_workers,engine_workers,fsdp_workers}.py:update_actor`

VERL already uses stable stage naming in the trainer, for example `gen`, `ref`, and `update_actor` timing. Reusing consistent stage names reduces cognitive overhead for users.

### 4. Use coarse and fine-grained stages

Two levels of instrumentation are proposed:

- Coarse-grained worker stages:
  - `rollout`
  - `ref_model`
  - `update_actor`
- Fine-grained training sub-stages inside actor implementations:
  - `train_forward`
  - `train_backward`
  - `update_step`

For FSDP actors, these boundaries naturally align with `forward -> loss -> backward -> optimizer_step`.

For Megatron actors, the first version may wrap a larger update boundary such as `update_policy` or `forward_backward_batch`, then refine later if needed.

### VERL configuration

Add a top-level configuration block, parallel to existing profiler-related settings:

```yaml
precision_debugger:
  enable: false

  backend: msprobe  # or null

  # VERL-side gates to limit collection scope
  stages: ["rollout", "train_forward", "ref_model", "train_backward", "update_actor"]
  ranks: null       # null = all, or e.g. [0, 1]
  steps: null       # null = all, or e.g. [100, 200]

  # Backend-specific config file path, interpreted by the backend
  config_path: null

  # Root dump directory
  dump_root: "outputs/precision_debug"

  # If backend import/init fails:
  # true  -> continue training with a warning
  # false -> fail fast
  fail_open: true
```

Notes:

- `enable: false` selects the no-op path.
- `stages`, `ranks`, and `steps` act as VERL-side gates independent of backend-specific filtering.
- `config_path` and `dump_root` are passed to the selected backend.

### Backend configuration

`msprobe` should continue using its native configuration format for collection details such as:

- Which modules or operators to trace
- Whether to collect full tensors or statistics only
- Dump formatting and backend-specific collection policies

VERL should not duplicate that schema. Instead:

- VERL is responsible for calling `start` / `stop` / `step` at the correct stage boundaries.
- VERL passes stage metadata such as `stage`, `global_step`, `rank`, and `uid`.
- The backend is responsible for interpreting its own configuration and organizing collected artifacts.

## Output Layout and Metadata

To make the collected data immediately usable, the backend should follow a predictable directory structure:

```text
outputs/precision_debug/
  exp=<experiment_name>/
    stage=<rollout|ref_model|train_forward|train_backward|update_actor>/
      step=<global_step>/
        rank=<global_rank>/
          ... backend dumps ...
        meta.json
```

Suggested metadata fields:

- `stage`
- `global_step`
- `rank`
- `world_size`
- Optional parallelism coordinates such as `dp_rank`, `tp_rank`, and `pp_rank`
- Batch `uid` values, if available
- Dtype / AMP summary, such as `bf16`, `fp16`, `fp8`, and loss scaling information
- Config summary and commit SHA, where practical

### Performance requirements

The feature must remain operationally safe to enable in production debugging scenarios:

- Disabled by default
- Scope-limited by `stages`, `steps`, and `ranks`
- Compatible with low-volume sampling, such as:
  - only selected UIDs per step
  - a fixed sampling ratio
  - only designated collector ranks, for example one rank per DP group

### Reliability requirements

- All stage wrappers must use `try/finally` so backend `stop()` calls are not skipped on exceptions.
- If `fail_open=true`, backend import or initialization errors should downgrade to a warning and training should continue.
- Collection should default to tensors and numerical statistics, not raw prompt text.
- If text dumping is supported by a backend, it should be explicitly documented as debug-only and user-responsible from a data governance perspective.

## Current Status

Some parts of the capability are already connected in internal experiments:

- `generate_sequences` collection
- `update_actor` collection
- `compute_log_prob` collection
- `compute_ref_log_prob` collection

The target dump structure has also been validated in prototype form.

## Additional Considerations for VERL 0.7+

VERL 0.7 and later introduced a more service-oriented execution model. As a result, directly inserting hooks into some actor-side code paths is no longer sufficient for complete rollout-stage collection.

There are two practical implications:

1. Current PrecisionDebugger support is focused on `vllm-ascend` via service-side configuration, while `GPUModelRunner` coverage remains incomplete.
2. A more uniform solution may require patching utilities under `verl/workers/rollout/vllm_rollout/utils.py` so rollout-side instrumentation can be applied consistently.

This patch-based utility approach must still preserve the core requirement: when disabled, it must impose effectively zero performance overhead.

### Server-side global step synchronization

To place rollout dumps under the correct `global_step` directory, the rollout service side also needs access to the current global step.

One practical approach is to synchronize a server-side global step counter whenever `update_weights` is called, potentially via a lightweight patch. This ensures rollout-generated debug data and trainer-side steps remain aligned in the dump layout.

## Rollout Plan

1. Infrastructure PR
   Add the `PrecisionDebuggerBackend` abstraction, no-op implementation, and configuration parsing.
2. Backend PR
   Add `MsprobePrecisionDebugger` with lazy import, lifecycle control, and dump directory management.
3. Integration PRs
   Add instrumentation to:
   - `generate_sequences`
   - `compute_ref_log_prob`
   - `update_actor`
   Start with coarse wrapping for actor updates, then refine into `train_forward`, `train_backward`, and `update_step` where practical.
4. Documentation PR
   Add a minimal example configuration, operational guidance, and common troubleshooting notes such as:
   - collecting only a single step
   - collecting only selected ranks
   - limiting dump size

## Open Questions

- Should VERL define a standard stage taxonomy that all backends must follow exactly, or allow backend-specific extensions?
- What is the minimum metadata contract required for cross-run comparison tools?
- Should VERL expose first-class sampling by UID, or leave that entirely to backend configuration?
- How should service-side rollout instrumentation evolve for non-Ascend runtimes?

### Directly integrate `msprobe` APIs throughout VERL

Rejected because it tightly couples VERL to one debugger backend and makes future backends harder to support.

### Instrument only the trainer layer

Rejected because worker-layer execution boundaries are a better match for how VERL dispatches stage work across FSDP, Megatron, and engine worker implementations.

### Always-on instrumentation hooks

Rejected because the disabled path must remain effectively free in both dependencies and runtime cost.

## References

- MindStudio Probe / `mindstudio-probe`: https://gitcode.com/Ascend/msprobe

Tjh-UKN:
### What does this PR do?

This PR integrates `msprobe` PrecisionDebugger into verl's unified profiler system for training-side precision data collection on Ascend.

The integration is centralized in `verl/utils/profiler/precision_debugger_profile.py`, so worker implementations do not need stage-specific msprobe logic. Precision dumps are organized by training global step and stage, and the runtime resolves the actual model used by each profiled stage for both legacy workers and the newer model-engine worker path.

### Checklist Before Starting

### Test

This change is not covered by CI because it depends on the external `msprobe` runtime and real Ascend training environments.

Validated manually with short PPO runs on `Qwen2-0.5B`:

- FSDP / new model-engine path
- `precision_debugger` enabled on selected stages and steps
- confirmed dump directory creation and `dump.json` generation
- verified actor/ref/critic stage collection
- verified `dump.json` cache growth issue is handled by resetting msprobe runtime status after `stop()`

### API and Usage Example

Minimal example:

```yaml
global_profiler:
  tool: precision_debugger
  steps: [1, 2]
  global_tool_config:
    precision_debugger:
      _target_: verl.utils.profiler.config.PrecisionDebuggerToolConfig
      enable: true
      config_path: /path/to/dump_config.json
      data_dir: outputs/precision_debug
      steps: [1, 2]
      stages:
        - actor_compute_log_prob
        - ref_compute_log_prob
        - actor_update
      strict: false

actor_rollout_ref:
  actor:
    profiler:
      enable: true
      ranks: [0]
  ref:
    profiler:
      enable: true
      ranks: [0]

critic:
  profiler:
    enable: true
    ranks: [0]
```

Output layout:

```text
outputs/precision_debug/
  step_1/
    actor_compute_log_prob/step0/rank0/dump.json
    actor_update/step0/rank0/dump.json
    ref_compute_log_prob/step0/rank0/dump.json
```

#### High-level design

- Integrate PrecisionDebugger as a profiler tool under verl's unified profiler framework
- Keep msprobe-specific logic centralized in `precision_debugger_profile.py`
- Resolve the actual stage model inside the profiler instead of spreading logic into workers
- Treat PrecisionDebugger as automatically discrete:
  each profiled stage runs as an independent `start -> stop -> reset_status` session
- Do not manually call `debugger.step()` in the current integration
- Reset msprobe runtime state after `stop()` to avoid stale `dump.json` cache accumulation across stages

#### Code changes

- Add `PrecisionDebuggerToolConfig`
- Add `precision_debugger` support in `DistProfiler`
- Implement `PrecisionDebuggerProfiler`
- Support both legacy worker paths and new model-engine worker paths in model resolution
- Wire `global_steps` into profiled stage execution so dumps are grouped by training step
- Fix `fsdp2` critic engine config selection in `ray_trainer.py`
- Add / update documentation for:
  - configuration
  - supported stages
  - output layout
  - result analysis
  - troubleshooting
  - overhead and disk usage

### Supported stages

- `actor_compute_log_prob`
- `ref_compute_log_prob`
- `actor_update`
- `compute_values`
- `critic_update`
- `compute_rm_score`

`rollout_generate` is intentionally skipped.

### Measured Overhead (Qwen2-0.5B)

Real-run measurements on Ascend with profiled steps `[1, 2]`:

#### Time

- Baseline steady-state: about `16-18 s/step`
- L1:
  - `step 1`: `177.35 s`
  - `step 2`: `161.80 s`
- Non-profiled later steps return close to baseline

This run showed about `9x-10x` overhead on profiled L1 steps. In larger models or broader stage coverage, tens-X slowdown is common.

#### Disk usage

- L1 (`actor_compute_log_prob`, `ref_compute_log_prob`, `actor_update`)
  - total for `step_1` and `step_2`: `21 MB`
- L0 (`actor_compute_log_prob`, `ref_compute_log_prob`, `compute_values`, `critic_update`, `actor_update`)
  - total for `step_1` and `step_2`: `74 MB`

### Documentation

Added a dedicated PrecisionDebugger guide under Ascend tutorial docs, covering:

- how to enable it
- how dumps are organized
- how to analyze `dump.json`
- measured overhead
- common troubleshooting cases

### Notes
- It does not change performance-profiler behavior for `nsys`, `npu`, `torch`, or `torch_memory`
- PrecisionDebugger is documented separately from performance profiling tools because it targets precision debugging rather than performance analysis

Fixes #5808


verl/workers/engine_workers.py
```diff
@@ -553,8 +555,14 @@ def init_model(self):
     @register(dispatch_mode=make_nd_compute_dataproto_dispatch_fn(mesh_name="ref"))
     @DistProfiler.annotate(color="olive", role="ref_compute_log_prob")
     def compute_ref_log_prob(self, data: TensorDict) -> TensorDict:
+        global_step = data.get("global_steps", None)
+        handle = precision_start(
```
wuxibin89: Please move `precision_start` and `precision_stop` into DistProfiler

verl/trainer/config/ppo_trainer.yaml
```diff
@@ -301,6 +301,27 @@ global_profiler:
       # devices, record_context etc.
       kw_args: {}
 
+# precision debugger configs
+precision_debugger:
```
wuxibin89: 1. Move it under `global_profiler`
2. Add a typed config class for it.
3. Add doc

Tjh-UKN: I‘ll move it
Here is the draft version`[step1]` and will be reviewed by profiler mates to decide the final version`[step2]`
Thanks for your tips

verl/workers/engine_workers.py
Tjh-UKN: I’ll do that

tardis-key:
Due to the addition of the msprobe dependency introduced by this pr, it is necessary to update the corresponding requirements files, supplement the necessary CI , and update the Dockerfile.

tardis-key:
Integrate into DistProfile control logic as much as possible to maximize reuse and minimize modification. Maybe we can confirm the control flow design for this pr first.

tardis-key:
Necessary test for different hardware platforms/training engines/rollout engines

docs/ascend_tutorial/profiling/precision_debugger.md
```diff
@@ -0,0 +1,140 @@
...
+
+Last updated: 02/14/2026.
+
+This guide explains how to collect precision data in verl using the
+`msprobe` PrecisionDebugger.
+
+## Prerequisites
+
+* Install `msprobe` in the training environment.
+* Prepare a `config.json` for msprobe (see examples below).
```
tardis-key, line 11: clarify the necessity of the extra json file and provide a json sample.

verl/utils/profiler/precision_debugger_profile.py
```diff
@@ -0,0 +1,238 @@
...
+            logger.warning(msg)
+            return False
+        if self._stages is not None and stage not in self._stages:
+            return False
+        if self.precision_cfg.steps is not None and global_step is not None:
+            if int(global_step) not in set(self.precision_cfg.steps):
+                return False
+        return True
+
+    def start(self, stage: Optional[str] = None, global_step: Optional[int] = None, model=None, **kwargs) -> bool:
```
tardis-key, line 169: Please confirm its operation in both E2E and discrete modes, and provide the test results.

docs/ascend_tutorial/profiling/precision_debugger.md
```diff
@@ -0,0 +1,140 @@
+# Precision Debugger (msprobe) in verl
```
tardis-key, line 1: please provide the usage of  time&disk when enabling msprobe under different models

docs/ascend_tutorial/profiling/precision_debugger.md
```diff
@@ -0,0 +1,140 @@
...
+
```
tardis-key, line 2: how to analyse data collected by msprobe.

tardis-key:
please confirm that when msprobe is not enabled, it has no impact on the original execution of verl

docs/ascend_tutorial/profiling/precision_debugger.md
Tjh-UKN: 🆗

verl/utils/profiler/precision_debugger_profile.py
Tjh-UKN: 🆗

docs/ascend_tutorial/profiling/precision_debugger.md
Tjh-UKN: 🆗

docs/ascend_tutorial/profiling/ascend_profiling_zh.rst
```diff
@@ -23,10 +23,13 @@ Last updated: 12/20/2025.
 
 -  global_profiler: 控制采集的rank和模式
 
-   -  tool: 使用的采集工具，选项有 nsys、npu、torch、torch_memory。
+   -  tool: 使用的采集工具，选项有 nsys、npu、torch、torch_memory、precision_debugger。
    -  steps: 此参数可以设置为包含采集步数的列表，例如 [2, 4]，表示将采集第2步和第4步。如果设置为 null，则不进行采集。
    -  save_path: 保存采集数据的路径。默认值为 "outputs/profile"。
 
+关于 msprobe PrecisionDebugger 的使用与配置，请参考
```
mengchengTang: [Review] msprobe has its own documentation, and there's no need to reference msprobe's documentation within the profiler (including torch.profiler) documentation. They serve different purposes—one focuses on performance, the other on precision—and their functionalities do not overlap.

docs/ascend_tutorial/profiling/precision_debugger.md
```diff
@@ -0,0 +1,140 @@
...
+Last updated: 02/14/2026.
+
```
mengchengTang, line 4: [Review] Is it appropriate to place the debugger under the "perf" directory? Since the "perf" directory mainly focuses on performance debugging, would it be better to place it under the "ascend_tutiral" directory instead?

verl/utils/profiler/precision_debugger_profile.py
```diff
@@ -0,0 +1,238 @@
+# Copyright 2024 Bytedance Ltd. and/or its affiliates
```
mengchengTang: [Review] Copyright is 2026

verl/utils/profiler/precision_debugger_profile.py
```diff
@@ -0,0 +1,238 @@
...
+                    if self.precision_cfg.strict:
+                        raise RuntimeError("Failed to create PrecisionDebugger instance")
+                    self._release_lock()
+                    return False
+            if hasattr(self._debugger, "service") and hasattr(self._debugger.service, "config"):
+                self._debugger.service.config.dump_path = dump_path
+            self._debugger.start(model)
+            return True
+        except Exception:
+            self._release_lock()
```
mengchengTang: [review] Does msprobe collection require locking?

docs/perf/precision_debugger.md
```diff
@@ -0,0 +1,140 @@
...
+    "bench_path": "/home/bench_data_dump",
+    "summary_mode": "md5",
+    "diff_nums": 5
+  }
+}
+```
+
+## Usage notes
+
+* PrecisionDebugger uses `start(model) -> stop() -> step()` semantics.
```
mengchengTang: [Review] Please provide links to usage guides and data analysis cases for the msprobe tool to help users get started.

tardis-key:
Could you confirm whether msprobe can collect GMM input from different frameworks? We are quite interested in this data.

docs/ascend_tutorial/profiling/precision_debugger.md
Tjh-UKN: modified

docs/ascend_tutorial/profiling/precision_debugger.md
Tjh-UKN: added

verl/trainer/config/ppo_megatron_trainer.yaml
```diff
@@ -228,6 +228,33 @@ global_profiler:
       trace_alloc_max_entries: 100_000
       # The depth of the call stack to capture for each allocation
       stack_depth: 32
+
```
mengchengTang: [review] When adding a new profiler tool, it is sufficient to add the tool’s default configuration in profiler.yaml; there is no need to add default parameter configurations in each ppo_xxx_trainer.yaml.

verl/trainer/config/ppo_megatron_trainer.yaml
```diff
@@ -228,6 +228,33 @@ global_profiler:
...
+      steps: null
+
+      # Profile stages
+      # Supported stages: actor_update, actor_compute_log_prob, ref_compute_log_prob,
+      # compute_values, critic_update, compute_rm_score
+      stages: null
+
+      # Whether to fail on unknown stage or missing msprobe
+      strict: False
       # 'alloc': records only allocation events || 'state': records memory state changes || 'all': records both.
```
mengchengTang: [review] The parameter configuration here has disrupted the layout of the torch_memory settings; please rearrange them.

mengchengTang:
/lgtm
It would be best to add UT and test cases afterward.

tardis-key (approved):

tardis-key:
The failed CI run is unrelated to the current changes.

PR #5195, opened 2026-02-04
[reward] fix: preserve input non_tensor_batch in AgentLoopManager when reward_loop_worker_handles is None

none0663:
### Problem

When the codebase is updated to [2cd9283](https://github.com/verl-project/verl/commit/2cd9283034f32b2c091253e2f7d4cb5cccae6a44) (migration to the new asynchronous reward manager), using **colocate RM** with async rollout (`AgentLoopManager`) causes validation to fail with: KeyError: 'data_source'

- **Where:** `verl/experimental/reward_loop/reward_manager/naive.py`, line 42, in `run_single` — it accesses `data_item.non_tensor_batch["data_source"]`.
- **Call path:** `_validate` → `_compute_reward_colocate(test_output_gen_batch_padded)` → `reward_loop_manager.compute_rm_score(batch)` → `RewardLoopWorker.compute_score_batch` → `compute_score` → `run_single`.
- **Cause:** When `reward_loop_worker_handles is None` (e.g. colocate RM), `AgentLoopManager.generate_sequences` returns a `DataProto` whose `non_tensor_batch` is built only from agent outputs (`__num_turns__`, `multi_modal_inputs`, `raw_prompt`). Input metadata such as `data_source` is never forwarded, so the batch passed to the reward manager is missing `data_source` and the naive reward manager raises `KeyError: 'data_source'`.

WeChatWorkScreenshot_0bc71e8a-a6a8-4334-b930-5a5d0bb149a2

## Solution

- Pass the input batch’s `non_tensor_batch` into `_postprocess` as `**kwargs`.
- When `reward_loop_worker_handles is None`, merge these `kwargs` into the output `non_tensor_batch` so `data_source` and other input keys are preserved.
- Colocate RM / validation then receives a batch that includes `data_source`, and the `KeyError` is fixed.

### DIFF
--- a/verl/experimental/agent_loop/agent_loop.py
+++ b/verl/experimental/agent_loop/agent_loop.py
@@ -469,7 +469,7 @@ class AgentLoopWorker:
             )
         outputs = await asyncio.gather(*tasks)
 
-        output = self._postprocess(outputs, **(batch.non_tensor_batch or {}))
+        output = self._postprocess(outputs, input_non_tensor_batch=batch.non_tensor_batch)
 
         return output
 
@@ -717,7 +717,11 @@ class AgentLoopWorker:
             output.reward_score = result["reward_score"]
             output.extra_fields["reward_extra_info"] = result["reward_extra_info"]
 
-    def _postprocess(self, inputs: list[_InternalAgentLoopOutput], **kwargs) -> DataProto:
+    def _postprocess(
+        self,
+        inputs: list[_InternalAgentLoopOutput],
+        input_non_tensor_batch: dict | None = None,
+    ) -> DataProto:
         """Process the padded outputs from _run_agent_loop and combine them into a batch."""
         # Convert lists back to tensors and stack them to create a batch.
         prompt_ids = torch.cat([input.prompt_ids for input in inputs], dim=0)
@@ -757,8 +761,8 @@ class AgentLoopWorker:
         non_tensor_batch = {
             "__num_turns__": np.array([input.num_turns for input in inputs], dtype=np.int32),
         }
-        if self.reward_loop_worker_handles is None and kwargs:
-            non_tensor_batch.update(kwargs)
+        if self.reward_loop_worker_handles is None and input_non_tensor_batch:
+            non_tensor_batch.update(input_non_tensor_batch)
 
         # add reward_extra_info to non_tensor_batch
         reward_extra_infos = [input.extra_fields.get("reward_extra_info", {}) for input in inputs]


yyDing1:
This is indeed a potential bug when using genrm for validation.
I think placing this change inside the agent loop would feel somewhat confusing for users.
We could first merge the batches using `batch.union(gen_batch)` (in the validation function) and then compute the reward, similar to that in the training process.

none0663:
> This is indeed a potential bug when using genrm for validation.
> I think placing this change inside the agent loop would feel somewhat confusing for users.
> We could first merge the batches using `batch.union(gen_batch)` (in the validation function) and then compute the reward, similar to that in the training process

I’d prefer to keep the fix in the agent loop rather than in ray_trainer.py. The missing data_source comes from agent_loop’s generate_sequences: its return value doesn’t carry the input’s non_tensor_batch when use colocate RM, so fixing it in the trainer would mean adding merge logic in several places (validation, training, REMAX) and would be more cumbersome. Fixing it inside the agent loop addresses the cause in one place, and all callers (validation and training) get the correct non_tensor_batch without touching ray_trainer.py.
https://github.com/verl-project/verl/blob/2320603ea8f3ac4ac3025ba4f74126d820e76884/verl/experimental/agent_loop/agent_loop.py#L692-L710

verl/experimental/agent_loop/agent_loop.py
```diff
@@ -717,7 +717,7 @@ async def _compute_score(self, output, prompts, responses, attention_mask, input
             output.reward_score = result["reward_score"]
             output.extra_fields["reward_extra_info"] = result["reward_extra_info"]
 
-    def _postprocess(self, inputs: list[_InternalAgentLoopOutput]) -> DataProto:
+    def _postprocess(self, inputs: list[_InternalAgentLoopOutput], **kwargs) -> DataProto:
```
yyDing1: "kwargs" parameter is a bit odd here, explicit maybe better.

verl/experimental/agent_loop/agent_loop.py
```diff
@@ -757,6 +757,8 @@ def _postprocess(self, inputs: list[_InternalAgentLoopOutput]) -> DataProto:
         non_tensor_batch = {
             "__num_turns__": np.array([input.num_turns for input in inputs], dtype=np.int32),
         }
+        if self.reward_loop_worker_handles is None and kwargs:
+            non_tensor_batch.update(kwargs)
```
yyDing1: Can we explicitly define what goes into `non_tensor_batch`? For example, specifying fields like "data_sources" might make things clearer.

verl/experimental/agent_loop/agent_loop.py
none0663: yes, the name "kwargs"  can be fixed.

verl/experimental/agent_loop/agent_loop.py
none0663: > Can we explicitly define what goes into `non_tensor_batch`? For example, specifying fields like "data_sources" might make things clearer.

We could mirror `_compute_score` and pass through the whole input non_tensor_batch (all keys) into the output, instead of whitelisting specific fields like "data_sources". That would keep behavior consistent and avoid maintaining an explicit list of keys. Should we do it that way?

none0663:
The two failing checks (NPU unit tests and pre-commit) don’t appear related to the edits in this PR

yyDing1:
Please run `pre-commit run --all-files` to format the modified files.

none0663:
> Please run `pre-commit run --all-files` to format the modified files.

Clipboard_Screenshot_1770606387

I verified this on my branch — pre-commit run --all-files but still produces no changes, and `verl/models/mcore/model_forward.py `is unrelated to this PR.

PR #5204, opened 2026-02-05
[megatron] fix: add protections for logits_processor_args.pop("loss_mask"), which may cause the `forward_fn` of value net collapse

albertcity:
### What does this PR do?

Fix a bug in `gpt_model_forward_no_padding`.
The `MegatronEngineWithValueHead` class fails to pass `logits_processor_args` to `forward_fn`, causing a crash when `gpt_model_forward_no_padding` attempts to pop the `loss_mask`.

### Test

> No need.

### Design & Code Changes

> add `if logits_processor_args and "loss_mask" in logits_processor_args:` check before try to `logits_processor_args.pop("loss_mask")`

### DIFF
--- a/verl/models/mcore/model_forward.py
+++ b/verl/models/mcore/model_forward.py
@@ -198,7 +198,8 @@ def gptmodel_forward_no_padding(
             }
             model_kwargs["labels"] = args["label"].contiguous()
             model_kwargs["loss_mask"] = args["loss_mask"].contiguous()
-        logits_processor_args.pop("loss_mask")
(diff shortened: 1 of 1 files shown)


ISEEKYAN:
hi @albertcity, thanks for point out this bug, I think this bug is introduced by the PR https://github.com/verl-project/verl/pull/4981. Actually in that PR we should not either pass the `loss_mask` through `logits_processor_args`, nor pop it without any check. We need to refactor the way of passing `loss_mask` cc @ArronHZG

ISEEKYAN:
just merge this fix before we refactor the passing way to a more reasonable way

PR #5246, opened 2026-02-09
fix: little Ascend NPU implementation bug fix

SolarWindRider:
### What does this PR do?

> This PR fixes a device-check bug `verl/utils/device.py` that frequently occurs when running VeRL on Ascend NPUs in cloud server container , as well as an error in the launch command in `docs/ascend_tutorial/ascend_quick_start`.


SolarWindRider:
@FightingZhen you can review this really quick. 2 lines fixed. Thanks

PR #5285, opened 2026-02-11
[misc] chore: fix veomni_trainer.yaml

wuxibin89:
### What does this PR do?

As title


verl/trainer/config/_generated_ppo_veomni_trainer.yaml
```diff
@@ -487,13 +487,11 @@ critic:
         _target_: verl.utils.profiler.config.TorchMemoryToolConfig
         trace_alloc_max_entries: ${oc.select:global_profiler.global_tool_config.torch_memory.trace_alloc_max_entries,100000}
         stack_depth: ${oc.select:global_profiler.global_tool_config.torch_memory.stack_depth,32}
```
vermouth1992, line 489: In the long term, we should combine all the backend config into a single config, just like SFT. And we have to write a new trainer to achieve this

PR #5298, opened 2026-02-12
[Megatron] feat: Support routing replay on NPU with performance and compatibility enhancements

755651978:
### What does this PR do?

This PR enables MoE Routing Replay on NPU (MindSpeed) platforms and provides critical compatibility patches for Megatron 0.12.1. It addresses the absence of key routing methods in older Megatron versions and ensures end-to-end data alignment from the rollout phase to the training phase.

### Key Enhancements

1. NPU-Compatible Routing Replay & Version Patching (router_replay_patch.py)

Compatibility Injection: Since Megatron 0.12.1 lacks the is_aux_loss_enabled() method, this PR implements it at the module level. Using types.MethodType, we dynamically bind this method to the router instance, ensuring consistency with newer Megatron APIs.

MindSpeed Resilience: Implements class-level attribute injection for enable_routing_replay and moe_router_fusion. This prevents MindSpeed’s dynamic dataclass reconstruction from stripping verl-specific configurations during NPU initialization.

Dynamic Signature Detection: Uses inspect.signature to adapt to different TransformerConfig versions and vp_stage (Virtual Pipeline) logic, ensuring correct layer offset mapping in complex pipeline-parallel NPU setups.

2. Robust Data Alignment for Agent Loops (tool_agent_loop.py & vllm_async_rollout.py)

Deterministic Rollout: Standardizes max_tokens to a fixed response_length in the vLLM rollout worker. This prevents shape mismatches in routed_experts caused by fluctuating prompt lengths.

Data Preservation: Uses safe getattr calls to ensure that routing metadata captured during the agent's interaction loop is successfully passed to AgentLoopOutput for training.
### Testing & Validation
Environment: Tested on Ascend NPU with MindSpeed.

Routing Consistency: Verified that the routed_experts generated during the rollout phase perfectly match the indices replayed during the training phase.

Performance: Benchmarked the forward pass; relocating helper functions resulted in a measurable reduction in Python-level overhead per iteration.

### DIFF
--- a/verl/workers/rollout/vllm_rollout/vllm_async_server.py
+++ b/verl/workers/rollout/vllm_rollout/vllm_async_server.py
@@ -534,7 +534,9 @@ class vLLMHttpServer:
             # Default to a calculation that considers configured lengths
             # Cap max_tokens by response_length to ensure tensor alignment,
             # and by remaining budget to prevent OOM in multi-turn rollouts.
-            max_tokens = min(self.config.response_length, self.config.prompt_length + self.config.response_length - len(prompt_ids))
+            max_tokens = min(
+                self.config.response_length, self.config.prompt_length + self.config.response_length - len(prompt_ids)
+            )
 
         # Clamp max_tokens to the valid range [0, max_possible_tokens]
         max_tokens = max(0, min(max_tokens, max_possible_tokens))
--- a/verl/workers/rollout/sglang_rollout/async_sglang_server.py
+++ b/verl/workers/rollout/sglang_rollout/async_sglang_server.py
@@ -351,7 +351,9 @@ class SGLangHttpServer:
         else:
             # Cap max_tokens by response_length to ensure tensor alignment,
             # and by remaining budget to prevent OOM in multi-turn rollouts.
-            max_new_tokens = min(self.config.response_length, self.config.prompt_length + self.config.response_length - len(prompt_ids))
+            max_new_tokens = min(
+                self.config.response_length, self.config.prompt_length + self.config.response_length - len(prompt_ids)
+            )
 
         # Clamp max_new_tokens to the valid range [0, max_possible_tokens]
         max_new_tokens = max(0, min(max_new_tokens, max_possible_tokens))


verl/utils/megatron/router_replay_patch.py
```diff
@@ -311,26 +335,48 @@ def apply_router_replay_patch():
...
+
+        if not hasattr(TransformerConfig, attr):
+            setattr(TransformerConfig, attr, val)
 
+    if not hasattr(TransformerConfig, "_verl_router_patched"):
         # Store original __init__ method
         original_tf_config_init = TransformerConfig.__init__
 
         # Define new __init__ method that safely handles enable_routing_replay parameter
+        @wraps(original_tf_config_init)
```
Li-Yongwen, line 380: miss wraps import

verl/utils/megatron/router_replay_patch.py
```diff
@@ -236,7 +239,27 @@ def compute_topk(scores, topk, num_groups=None, group_topk=None):
 
     return routing_probs, routing_map
 
-
+def _get_aux_loss_coeff(_self, aux_loss_type: str) -> float:
+    """获取给定辅助损失类型的系数。"""
```
wucong25: 注释最好统一写成英文，无关注释建议去掉

verl/utils/megatron/router_replay_utils.py
```diff
@@ -330,7 +331,13 @@ def get_current_rank_layer_info(tf_config, vp_rank=None):
     if vp_rank is None:
         vp_rank = 0
     num_layers_to_build = get_num_layers_to_build(tf_config, vp_stage=vp_rank)
-    offset = get_transformer_layer_offset(tf_config, vp_stage=vp_rank)
```
wucong25, line 371: 这一段为何要单独判断一下是否有vp_state入参

verl/utils/megatron/router_replay_patch.py
```diff
@@ -311,26 +336,48 @@ def apply_router_replay_patch():
...
-        # Add class attribute with default value
-        TransformerConfig.enable_routing_replay = False
+    try:
+        global_args = get_args()
+    except Exception:
+        global_args = None
+
+    try:
+        sig = inspect.signature(TransformerConfig.__init__)
+        native_params = sig.parameters
```
wucong25, line 350: 先保存一份原始的TransfoermerConfig参数

verl/utils/megatron/router_replay_patch.py
```diff
@@ -311,26 +336,48 @@ def apply_router_replay_patch():
...
         # Store original __init__ method
         original_tf_config_init = TransformerConfig.__init__
 
         # Define new __init__ method that safely handles enable_routing_replay parameter
+        @wraps(original_tf_config_init)
         def patched_tf_config_init(self, *args, **kwargs):
             # Simple solution: remove the unknown parameter before calling original constructor
-            enable_routing_replay = kwargs.pop("enable_routing_replay", TransformerConfig.enable_routing_replay)
+            if "enable_routing_replay" not in native_params:
+                enable_routing_replay = kwargs.pop("enable_routing_replay", TransformerConfig.enable_routing_replay)
```
wucong25: 如果原始的TransformerConfig中没有当前这个参数，则用新的注入的参数值代替

verl/experimental/agent_loop/tool_agent_loop.py
```diff
@@ -193,11 +193,14 @@ async def run(self, sampling_params: dict[str, Any], **kwargs) -> AgentLoopOutpu
...
         if agent_data.video_data is not None:
             multi_modal_data["videos"] = agent_data.video_data
+
+        routed_experts = getattr(agent_data, "routed_experts", None)
         output = AgentLoopOutput(
             prompt_ids=prompt_ids,
             response_ids=response_ids[: self.response_length],
             response_mask=agent_data.response_mask[: self.response_length],
             multi_modal_data=multi_modal_data,
+            routed_experts=routed_experts,
```
wucong25: 这里为何之前也没传入，是bugfix么还是npu特有的

verl/utils/megatron/router_replay_utils.py
Li-Yongwen: 目前npu配套中megatron还只支持到0.12.0，从megatron中导入的get_num_layers_to_build函数没有vp_stage参数，暂时无法通过其他方式使用该参数

verl/utils/megatron/router_replay_patch.py
```diff
@@ -311,26 +336,48 @@ def apply_router_replay_patch():
     # Clear router instances to avoid state leakage between model initializations.
     RouterReplay.router_instances.clear()
     # Step 1: Patch TransformerConfig to include the feature flag
-    if not hasattr(TransformerConfig, "enable_routing_replay"):
-        # Add class attribute with default value
-        TransformerConfig.enable_routing_replay = False
+    try:
+        global_args = get_args()
```
wucong25: verl/utils/megatron/router_replay_patch.py:340:23: F821 Undefined name `get_args`

verl/experimental/agent_loop/tool_agent_loop.py
755651978: 这是一个针对 MoE 数据链路的 Bugfix / 架构补齐，并非 NPU 特有的逻辑。在之前的同步Server逻辑中，是设置了batch["routed_experts"] = routed_experts，但在异步模式中，ToolAgentLoop 在处理完多轮交互并打包 AgentLoopOutput 时，漏掉了 routed_experts 这个字段

Li-Yongwen: https://github.com/verl-project/verl/commit/c6255ae38646e59c4de532ccc726c3e05cb8091e

Li-Yongwen: check一下上面这个commit

verl/utils/megatron/router_replay_patch.py
755651978: 是的，这里预先获取原始参数签名，是为了动态识别当前环境下的 Megatron 版本是否已原生支持这些属性。

verl/utils/megatron/router_replay_patch.py
755651978: 已调整

verl/experimental/agent_loop/tool_agent_loop.py
755651978: 已经 check 了 commit c6255ae，并已将其合入到当前 PR 中。

verl/workers/rollout/vllm_rollout/vllm_async_server.py
```diff
@@ -530,7 +530,12 @@ async def generate(
             max_tokens = sampling_params.pop("max_new_tokens")
         else:
             # Default to a calculation that considers configured lengths
-            max_tokens = self.config.response_length + self.config.prompt_length - len(prompt_ids)
+            if self.config.enable_rollout_routing_replay:
+                # When routing replay is enabled, we strictly use the configured response length
+                max_tokens = self.config.response_length
```
wuxibin89: Why have this restriction? max_tokens is max tokens generated in a single turn.

This change may cause total sequence length exceeds `self.config.response_length + self.config.prompt_length` in multi-turn rollout. For example, 
- prompt_length=8k
- response_length=32k

before last turn, sequence_length=30k, with max_tokens=32k, the total sequence may be maximum 30k+32k=62k.

755651978: I completely understand the concern about the total sequence length exceeding prompt_length + response_length in multi-turn rollouts, which could cause OOM.

However, the dynamic calculation (prompt_length + response_length - len(prompt_ids)) causes a critical Tensor Shape Mismatch in the subsequent agent_loop padding logic.

Here is why:
In the padding section, routed_experts is assigned as:
routed_experts[:, start_pos:end_pos] = experts_tensor.unsqueeze(0)

start_pos is prompt_length - len(prompt_ids).

If max_tokens takes up the remaining global budget, the generated experts_tensor.shape[0] (i.e., length) becomes prompt_length + response_length.

This causes start_pos + length to exceed total_length, forcing end_pos to be truncated by the min() function.

Result: The slice size end_pos - start_pos strictly equals response_length + len(prompt_ids), but the experts_tensor size is prompt_length + response_length. Since len(prompt_ids) < prompt_length (due to left padding), PyTorch throws a shape mismatch error.

To fix the OOM concern you raised while preventing this shape mismatch, I will update the logic to cap max_tokens by taking the minimum of the two constraints:

max_tokens = min(self.config.response_length, (self.config.prompt_length + self.config.response_length) - len(prompt_ids))

This ensures length never exceeds the valid padded boundaries, and we also safely respect the global memory budget.

wuxibin89: OK, I get your point, let's cap max_tokens by self.config.response_length
```python
max_tokens = min(self.config.response_length, (self.config.prompt_length + self.config.response_length) - len(prompt_ids))
```

verl/workers/rollout/vllm_rollout/vllm_async_server.py
```diff
@@ -530,7 +530,14 @@ async def generate(
             max_tokens = sampling_params.pop("max_new_tokens")
         else:
             # Default to a calculation that considers configured lengths
-            max_tokens = self.config.response_length + self.config.prompt_length - len(prompt_ids)
+            if self.config.enable_rollout_routing_replay:
```
wuxibin89: Please cap max_tokens regardless of enable_rollout_routing_replay, it shouldn't be diverged.

wuxibin89: Please also adjust sglang and trt-llm as well.

755651978: Copy that. I will unify the max_tokens calculation logic to remove the divergence, ensuring it's capped for all cases. I'll also update the corresponding implementations in sglang and trt-llm to maintain consistency across different backends. Will push the update soon.

755651978: Done. I have unified the max_tokens calculation logic as suggested.

Refactored: Updated the logic in both SGLang and vllm to use the dual-constraint: min(self.config.response_length, self.config.prompt_length + self.config.response_length - len(prompt_ids)).

Verified: Checked TRT-LLM and confirmed it already implements a similar safety cap (min(response_length, max_model_len - len(prompt_ids))), so no further changes were needed there.

The logic is now consistent across all backends to prevent both tensor shape mismatches and OOM.

verl/workers/rollout/vllm_rollout/vllm_async_server.py
litianjian: The end_pos `end_pos = min(start_pos + length, total_length)` has been truncated by the min() function. If we also truncate the experts, would that solve the problem?

755651978: Truncating the experts would resolve the shape mismatch, but it leads to data inconsistency between tokens and their routing metadata. In RL (like GRPO), every generated token needs its corresponding routing info for stable training.

Capping max_tokens at the source is a cleaner fix that ensures perfect alignment and respects the memory budget without any data loss.

PR #5305, opened 2026-02-12
[fsdp,veomni] fix: remove FSDPUlyssesShardingManager to make eval_mode/train_mode reentrant

wuxibin89:
### What does this PR do?

The `update_actor` call `train_mode` twice, the original FSDPUlyssesShardingManager is not reentrant, result in global ulysess group is still actor's sp group after update_actor.
```
ActorRolloutRefWorker.update_actor
               |
               v
TrainingWorker.train_mini_batch     <- train_mode (1)
               |
               v
TrainingWorker.train_batch          <- train_mode (2)
               |
               v
BaseEngine.train_batch
```


vermouth1992:
If `FSDPUlyssesShardingManager` is not used, shall we deprecate it and remove it after next release?

Issue #5325, opened 2026-02-15
[BUG] model_merger sets lora_alpha=0 when merging SFT LoRA checkpoints


Yatogaii:
### System Info

No relevant with system info.

### Information

- [x] The official example scripts
- [ ] My own modified scripts

### Tasks

- [ ] An officially supported task in the `examples` folder (such as GLUE/SQuAD, ...)
- [x] My own task or dataset (give details below)

### Reproduction

1. Train an SFT model with LoRA enabled, for example:
```
model.lora_rank=8 
model.lora_alpha=16
```

2.Merge the checkpoint using:
```
python3 -m verl.model_merger merge
```

3. Inspect the generated LoRA config in the merged output directory.
The merged adapter_config.json contains:
```
    "lora_alpha": 0,
```

### Expected behavior

`lora_alpha` in the merged checkpoint should match the value specified during training (e.g., 16 in this example).

As mentioned in #3050, this issue still exists in the current version.
This behavior effectively makes SFT + LoRA workflows unusable after merging, since the scaling factor is lost.

Issue #3050, opened 2025-08-14
lora_alpha is set to 0.0 by dafault without warning in base_model_merger.py


caowencai:
https://github.com/volcengine/verl/blob/1a62568f801ba35ac1f5387e27232a2df7eac488/verl/model_merger/base_model_merger.py#L251

techkang:
Thanks for pointing this out. Would you like to make a PR?

caowencai:
> Thanks for pointing this out. Would you like to make a PR?

Here: https://github.com/volcengine/verl/pull/3100

Kirrito-k423:
中文：感谢指出这个问题。根据当前公开证据，这个问题已经由 #5326 修复并合并到 main。该修复会在 SFT LoRA 训练 checkpoint 中保存 `lora_train_meta.json`，并在 `model_merger` 中读取其中的 `lora_alpha`；如果元数据缺失或 `lora_alpha` 为 0，现在也会发出 warning。相关复现 issue #5325 已随 #5326 关闭。

证据：
- 修复 PR：#5326
- 相关复现 issue：#5325
- 代码路径：`verl/trainer/sft_trainer.py`、`verl/utils/checkpoint/checkpoint_handler.py`、`verl/model_merger/base_model_merger.py`

English: Thanks for reporting this. Based on the current public evidence, this has been fixed by #5326 and merged into main. The fix saves `lora_train_meta.json` in SFT LoRA checkpoints and makes `model_merger` read `lora_alpha` from that metadata; it also warns when the metadata is missing or `lora_alpha` is 0. The related reproduction issue #5325 was closed after #5326 was merged.

Evidence:
- Fix PR: #5326
- Related reproduction issue: #5325
- Code paths: `verl/trainer/sft_trainer.py`, `verl/utils/checkpoint/checkpoint_handler.py`, `verl/model_merger/base_model_merger.py`

This response was generated by a scheduled Codex GPT5.5 issue processing task.

PR #5326, opened 2026-02-15
[ckpt, model] fix: preserve lora_alpha in model_merger via training meta

Yatogaii:
### What does this PR do?

This PR fixes the issue where `model_merger` generates a merged LoRA adapter with `lora_alpha=0`, even though a non-zero value was specified during training.

The fix is done by **persisting the LoRA training-time config** and updating the base merger logic to read it as the source of truth during merge.

* Fixes: #5325
* Related: #3050
* Related: #3100

Key points:

1. Save LoRA config during training into `lora_train_meta.json`.
2. Update base merger logic to load `lora_train_meta.json` when merging, ensuring `lora_alpha` is correct.
3. This approach is **self-contained** and deterministic (no inference / heuristic fallback).

Compared to #3100:

* #3100 adds a warning and uses a heuristic fallback (`lora_alpha = rank * 2`).
* Existing approach infers rank from checkpoint `.pt` weights instead of reading the original training configuration.
* This PR directly reads the training-time config saved by verl, avoiding ambiguity and incorrect defaults.

---

### Checklist Before Starting

* [x] Search for similar PRs. Paste at least one query link here:

  * https://github.com/verl-project/verl/pull/4704
  * https://github.com/verl-project/verl/pull/3100

* [x] Format the PR title as `[{modules}] {type}: {description}` (This will be checked by the CI)

---

### Test

This change is not covered by CI.

I validated the fix with a local end-to-end workflow:

1. Train SFT + LoRA with:

   * `model.lora_rank=8`
   * `model.lora_alpha=16`

2. Merge with:

   ```bash
   python3 -m verl.model_merger merge ...
   ```

3. Verify the merged adapter config:

   * Before: `lora_alpha = 0`
   * After: `lora_alpha = 16`

---

### API and Usage Example

No API changes are introduced.

This PR only adds an additional training artifact:

* `lora_train_meta.json`

The merge process remains unchanged for users.

---

#### Design

* Training stage persists LoRA-related configuration into `lora_train_meta.json`.
* Merge stage loads `lora_train_meta.json` and uses it as the source of truth for LoRA parameters.

This PR intentionally **does not** rely on `adapter_config.json`, since that file may be ambiguous (it may represent merged/runtime state rather than training-time configuration).

#### Code Changes

* Save LoRA training metadata during SFT training.
* Update `model_merger` base merge logic:

  * Prefer reading `lora_train_meta.json`
  * Avoid inferring LoRA parameters from `.pt` weights

---

> [!IMPORTANT]

* [x] Read the [[Contribute Guide](https://github.com/volcengine/verl/blob/main/CONTRIBUTING.md)](https://github.com/volcengine/verl/blob/main/CONTRIBUTING.md).
* [x] Apply [[pre-commit checks](https://github.com/volcengine/verl/blob/main/CONTRIBUTING.md#code-linting-and-formatting)](https://github.com/volcengine/verl/blob/main/CONTRIBUTING.md#code-linting-and-formatting): `pre-commit install && pre-commit run --all-files --show-diff-on-failure --color=always`
* [x] Add / Update [[the documentation](https://github.com/volcengine/verl/tree/main/docs)](https://github.com/volcengine/verl/tree/main/docs).
* [x] Add unit or end-to-end test(s) to [[the CI workflow](https://github.com/volcengine/verl/tree/main/.github/workflows)](https://github.com/volcengine/verl/tree/main/.github/workflows) to cover all the code. If not feasible, explain why:

  * Not feasible for CI at the moment because it requires an end-to-end LoRA training + merge pipeline.
* [] Once your PR is ready for CI, send a message in [the `ci-request` channel](https://verl-project.slack.com/archives/C091TCESWB1) in [the `verl` Slack workspace](https://join.slack.com/t/verl-project/shared_invite/zt-3855yhg8g-CTkqXu~hKojPCmo7k_yXTQ). (If not accessible, please try [[the Feishu group (飞书群)](https://applink.larkoffice.com/client/chat/chatter/add_by_link?link_token=772jd4f1-cd91-441e-a820-498c6614126a)](https://applink.larkoffice.com/client/chat/chatter/add_by_link?link_token=772jd4f1-cd91-441e-a820-498c6614126a).)
* [x] If your PR is related to the `recipe` submodule, please also update the reference to the submodule commit via `git submodule update --remote` or `cd recipe && git pull origin main`.

### DIFF
--- a/verl/utils/checkpoint/checkpoint_handler.py
+++ b/verl/utils/checkpoint/checkpoint_handler.py
@@ -15,10 +15,10 @@
 
 # TODO: add unit tests
 
+import json
 import logging
 import os
 import re
-import json
 from enum import Enum
 
 import torch


verl/trainer/fsdp_sft_trainer.py
```diff
@@ -564,6 +565,16 @@ def save_checkpoint(self, step):
             local_mkdir_safe(local_global_step_folder)
             dataloader_local_path = os.path.join(local_global_step_folder, "data.pt")
 
+            if self.lora:
```
wuxibin89: fsdp_sft_trainer.py is going to removed in https://github.com/verl-project/verl/pull/5382, please use sft_trainer.py instead.

verl/trainer/fsdp_sft_trainer.py
```diff
@@ -564,6 +565,16 @@ def save_checkpoint(self, step):
...
+                lora_meta = {
+                    "r": int(getattr(self.config.model, "lora_rank", 0) or 0),
+                    "lora_alpha": int(getattr(self.config.model, "lora_alpha", 0) or 0),
+                }
+                lora_meta_path = os.path.join(local_global_step_folder, "lora_train_meta.json")
```
HollowMan6: I think it's better to make things consistent with the PEFT library by saving the metadata as `adapter_config.json` instead (so that PEFT library can reuse the saved config and load adapters by `PeftConfig.from_pretrained`). A reference for such kind of a config is at https://huggingface.co/tloen/alpaca-lora-7b/blob/main/adapter_config.json.

verl/model_merger/base_model_merger.py
```diff
@@ -244,6 +245,38 @@ def patch_model_generation_config(self, model):
                 )
         return model
 
+    def _load_lora_train_meta(self) -> Optional[dict[str, int]]:
+        if not self.config.local_dir:
+            return None
+
+        meta_path = os.path.join(self.config.local_dir, "lora_train_meta.json")
```
HollowMan6, line 252: Ditto

HollowMan6 (commented):
Some comments

verl/trainer/fsdp_sft_trainer.py
```diff
@@ -564,6 +565,16 @@ def save_checkpoint(self, step):
             local_mkdir_safe(local_global_step_folder)
             dataloader_local_path = os.path.join(local_global_step_folder, "data.pt")
 
+            if self.lora:
+                lora_meta = {
+                    "r": int(getattr(self.config.model, "lora_rank", 0) or 0),
+                    "lora_alpha": int(getattr(self.config.model, "lora_alpha", 0) or 0),
+                }
```
HollowMan6: Probably you can use `PeftConfig` as the data structure for related infos and call its API to save configs as an alternative way of achieving this

Yatogaii: Thanks for the suggestion — I agree aligning with PEFT is valuable for interoperability.

In our case, though, the artifact we save here is not a PEFT LoRA adapter yet. It’s a raw weight artifact that requires an explicit conversion/export step before it becomes something that PeftConfig.from_pretrained() / PeftModel.from_pretrained() can reliably consume as an adapter.

Because of that, writing adapter_config.json alongside this weight file would be semantically ambiguous: it makes the output directory look like a ready-to-load PEFT adapter repo, which can mislead users/tools into attempting direct from_pretrained() loading and either failing or (worse) loading incorrectly.

So for now I’d prefer to keep a clearly-scoped metadata file (e.g., lora_train_meta.json) that describes the training-time LoRA params for our pipeline, and reserve emitting a canonical adapter_config.json for the stage where we actually export a PEFT-compatible adapter (if/when we add that export path).

Happy to add a brief note in the metadata/README to make the “requires conversion” contract explicit.

verl/trainer/fsdp_sft_trainer.py
Yatogaii: Thanks for the heads-up about the migration into the `sft_trainer.py`. I re-tested with the migrated SFT trainer code path, and the original issue still reproduces there (i.e., the saved SFT output still doesn’t carry the LoRA metadata we need, so I’ll update the new SFT trainer implementation accordingly).

While doing that verification, I also noticed an additional potential problem on the merged PEFT config side: after `verl.model_merger`, `r` looks correct but `lora_alpha` ends up as `0` in the generated JSON, which makes the merged adapter unusable in our downstream RL stage.

Separately, during further investigation I found that `task_type` can also become null in the merged config. I’m going to open an issue to track the expected `task_type` semantics/source-of-truth, and in the meantime I’ll include a fix in this PR to avoid emitting an unusable config (i.e., ensure task_type is not null).

Yatogaii:
Quick follow-up: the fix has been migrated to the new SFT trainer path via the shared checkpoint helper, and metadata now includes both `lora_alpha` and `task_type`. I also added warnings for LoRA-enabled runs when `lora_alpha` is missing or zero (no silent fallback). Merger now reads these fields from training metadata, so generated adapter config no longer ends up with `lora_alpha=0` / `task_type=null` in the validated path.

HollowMan6:
image

@Yatogaii Looks like the conflicts is still there

Yatogaii:
> image
> 
> @Yatogaii Looks like the conflicts is still there

Resolved — the file I touched was deprecated and removed upstream, so I removed it from this PR and pushed the updated commits. The branch should be conflict-free now.

HollowMan6:
@Yatogaii Would you please fix the pre-commit formatting issue https://github.com/verl-project/verl/actions/runs/22667396364/job/65702601065?pr=5326?

Yatogaii:
> @Yatogaii Would you please fix the pre-commit formatting issue https://github.com/verl-project/verl/actions/runs/22667396364/job/65702601065?pr=5326?

Fixed: ran pre-commit; imports are now correctly ordered.

Issue #4229, opened 2025-11-21
vLLMAsyncRollout _loop_forever error: Free memory on device on startup is less than desired GPU memory utilization (0.8)


shijiexu09:
### System Info

8 gpus * 2 nodes

### Information

- [x] The official example scripts
- [ ] My own modified scripts

### Tasks

- [x] An officially supported task in the `examples` folder (such as GLUE/SQuAD, ...)
- [ ] My own task or dataset (give details below)

##### Question as follows：
When fully async + vLLM + fsdp_size=1, Rollout Worker startup reports: Insufficient GPU memory.

Before vLLM starts, Rollout Worker loads the FSDP model, occupying redundant GPU memory; vLLM then loads the complete model, causing GPU memory usage to accumulate.

##### Reasons as follows:
In fully async mode, the Rollout Worker is only responsible for inference; weights are broadcast by the Trainer (sync_rollout_weights) → no local FSDP model is needed.

The existing logic `if self._is_actor` or `self._is_rollout` will indiscriminately load the FSDP model, resulting in wasted GPU memory.

##### How to solve：
=== verl/workers/fsdp_workers.py Lines 764
def init_model(self):
    ...
    if self._is_actor or self._is_rollout:

    `Modified to`

    if self._is_actor or (self._is_rollout and self.config.rollout.mode != "async"):
        ...

=== verl/workers/fsdp_workers.py Lines 625
def _build_rollout(self, trust_remote_code=False):
    ...
    if torch.distributed.get_world_size() == 1 and fsdp_version(self.actor_module_fsdp) == 1:

    `Modified to`

    if hasattr(self, "actor_module_fsdp") and self.actor_module_fsdp is not None:
       if torch.distributed.get_world_size() == 1 and fsdp_version(self.actor_module_fsdp) == 1:
           FSDP.set_state_dict_type(...)

### Expected behavior

In async mode, the Rollout Worker only initializes the tokenizer/config and directly hands it over to vLLM to load from the model path.

The training end continues to hold the FSDP model and synchronizes weights to vLLM via NCCL broadcast.

Memory usage is significantly reduced; fsdp_size=1 no longer triggers vLLM startup failure.

shijiexu09:
> ### System Info
> 8 gpus * 2 nodes
> 
> ### Information
> * [x]  The official example scripts[ ]  My own modified scripts
> 
> ### Tasks
[...]

PR #5384, opened 2026-02-24
[fsdp] fix: add aggressive_empty_cache at end of init_model to prevent vLLM OOM

EricMarcus-ai:
### What does this PR do?

Adds `aggressive_empty_cache(force_sync=True)` at the end of `ActorRolloutRefWorker.init_model()` to prevent vLLM from OOMing at startup when colocated on the same GPUs as FSDP.

Related: #4229, #4257 (stale)

After the removal of `ExternalZeroMQDistributedExecutor`, vLLM runs in separate MP worker processes instead of inside the FSDP worker process. During `init_model()`, PyTorch's CUDA allocator reserves large transient blocks for full-model loading before FSDP sharding and `sync_module_states` broadcasting. After init, these blocks are no longer needed but remain cached by the allocator (`cudaMalloc`'d, not `cudaFree`'d). Since vLLM now runs in a separate process with its own allocator, it cannot reuse these cached blocks — `cudaMemGetInfo` reports them as "used", and vLLM fails its `gpu_memory_utilization` check.

Previous attempts to fix this (#4257) went stale. This approach is simpler: one line, no guards needed, and is a no-op when there is nothing to free.

### Checklist Before Starting

- [x] Search for similar PRs: [aggressive_empty_cache](https://github.com/verl-project/verl/pulls?q=aggressive_empty_cache), [OOM fsdp vllm init](https://github.com/verl-project/verl/pulls?q=OOM+fsdp+vllm+init)

### Test

This cannot be tested in CI because the OOM is a cross-process CUDA memory visibility issue that requires colocated FSDP + vLLM on the same physical GPU to reproduce.

- The call site is exercised by the existing `tests/workers/test_fsdp_workers.py`
- Validated experimentally with a colocated FSDP + vLLM training run (8B VLM, 8x H200, hybrid mode)
- The fix is a no-op when there is no cached memory to free, so it is safe in all configurations

### API and Usage Example

No API changes. The fix is automatic.

### Design & Code Changes

One line added at the end of `ActorRolloutRefWorker.init_model()` in `verl/workers/fsdp_workers.py`:

```python
# Free cached GPU memory so colocated vLLM processes can see it via cudaMemGetInfo
aggressive_empty_cache(force_sync=True)
```

This pattern already exists in the codebase:
- `megatron_workers.py:677` — `empty_cache()` at end of `init_model`
- `fsdp_workers.py:742` — `aggressive_empty_cache` during `rollout_mode()` context switch
- `engine_workers.py:671` — `aggressive_empty_cache` during rollout mode

The FSDP worker was the only one missing it at init time.

- [x] Apply [pre-commit checks](https://github.com/volcengine/verl/blob/main/CONTRIBUTING.md#code-linting-and-formatting): all hooks pass.
- [ ] Add / Update [the documentation](https://github.com/volcengine/verl/tree/main/docs). — N/A, no user-facing changes.
- [x] Add unit or end-to-end test(s) — not feasible: requires multi-process colocated GPU setup to reproduce; the fix is exercised by existing `test_fsdp_workers.py`.
- [x] Not related to `recipe` submodule.

### DIFF
--- a/verl/workers/engine_workers.py
+++ b/verl/workers/engine_workers.py
@@ -583,6 +583,9 @@ class ActorRolloutRefWorker(Worker, DistProfilerExtension):
                 backend, is_master=(torch.distributed.get_rank() == 0), bucket_size=bucket_size, **engine_kwargs
             )
 
+        # Free cached GPU memory so colocated vLLM processes can see it via cudaMemGetInfo
+        aggressive_empty_cache(force_sync=True)
+
     @register(dispatch_mode=make_nd_compute_dataproto_dispatch_fn(mesh_name="ref"))
     @DistProfiler.annotate(color="olive", role="ref_compute_log_prob")
     @_with_routing_replay_flag(enabled=False)
--- a/verl/workers/megatron_workers.py
+++ b/verl/workers/megatron_workers.py
@@ -674,7 +674,8 @@ class ActorRolloutRefWorker(MegatronWorker, DistProfilerExtension):
             if not self.config.actor.megatron.use_mbridge:
                 self.weight_converter = get_mcore_weight_converter(self.actor_model_config, self.dtype)
 
-        get_torch_device().empty_cache()
+        # Free cached GPU memory so colocated vLLM processes can see it via cudaMemGetInfo
+        aggressive_empty_cache(force_sync=True)
         log_gpu_memory_usage("After init_model finish", logger=logger)
 
     async def rollout_mode(self):


wuxibin89:
@EricMarcus-ai Thank you, could you also help to fix `megatron_worker.py` and `engine_workers.py`? I think they also need it in `init_model`.

EricMarcus-ai:
> @EricMarcus-ai Thank you, could you also help to fix `megatron_worker.py` and `engine_workers.py`? I think they also need it in `init_model`.

@wuxibin89 
Added them, have a look. I can't fully test the other pipelines end-to-end now, however (but it is just cache clears, should be fine). The megatron one already had an empty-cache command; just replaced it with the aggressive-cache-empty instead.

PR #5418, opened 2026-02-26
[BREAKING][rollout,cfg] refactor: get rid of actor_rollout_ref config from rollout

wuxibin89:
### What does this PR do?

https://github.com/verl-project/verl/issues/5400#issuecomment-3960028109, get rid of actor_rollout_ref from rollout: 
- AgentLoopManager
- AgentLoopWorker
- AgentLoopBase: SingleTurnAgentLoop, ToolAgentLoop

Eventually, we should get rid of actor_rollout_ref in `verl/trainer/config/ppo_trainer.yaml` with flattend actor, ref, rollout fields:
```python
actor:
  _target_: verl.workers.config.FSDPActorConfig
  ...

ref:
  _target_: verl.workers.config.FSDPActorConfig
 ...

rollout:
  _target_: verl.workers.config.RolloutConfig
  ...
```


verl/trainer/config/rollout/rollout.yaml
```diff
@@ -7,6 +7,12 @@ name: ???
 # sync: LLM, async: AsyncLLM
 mode: async
 
+# Number of nodes for standalone rollout server, must be > 0 in one-step-off/fully async training.
+nnodes: 0
```
ArronHZG, line 11: Do we not need to use trainer.nnodes by default here?

verl/experimental/agent_loop/agent_loop.py
```diff
@@ -941,7 +966,8 @@ def _init_agent_loop_workers(self):
                 ).remote(self.config, self.server_handles, self.reward_loop_worker_handles)
             )
 
-    def generate_sequences(self, prompts: DataProto) -> DataProto:
+    @auto_await
+    async def generate_sequences(self, prompts: DataProto) -> DataProto:
```
ArronHZG, line 970: In the fully async and one step off modes, the validate process still goes through this interface. During the last test, the caller did not await the generate() method, which led to some issues.

Should we not enable auto wait for this interface for now?

wuxibin89: > the validate process still goes through this interface

The auto_await handles 3 cases:
1. await generate_sequences()
2. directly call generate_sequences() in non async context(loop is None)
3. directly call generate_sequences() in async context(loop is running)

the validate process is case 3, the ci test has cover this case right?

verl/trainer/config/rollout/rollout.yaml
wuxibin89: No, `rollout.nnodes > 0` means that we need to create separate GPU resource for rollout server.

tests/experimental/reward_loop/test_agent_reward_loop_standalone.py
```diff
@@ -56,6 +56,7 @@ def test_agent_reward_loop_standalone():
     config.actor_rollout_ref.rollout.prompt_length = 1024
     config.actor_rollout_ref.rollout.response_length = 4096
     config.actor_rollout_ref.rollout.skip_tokenizer_init = True
+    config.actor_rollout_ref.rollout.nnodes = 1
     config.trainer.n_gpus_per_node = 4
     config.trainer.nnodes = 1
```
Shangwei-Li, line 61: Is config.trainer.nnodes = 1 still needed for test_agent_reward_loop_standalone?

wuxibin89: No, `config.trainer.nnodes` is not used in standalone mode.

PR #5447, opened 2026-03-01
[reward] perf: batch tokenizer.decode() in reward managers

dubin555:
## Summary

Replace per-item `tokenizer.decode()` loops with `tokenizer.batch_decode()` in 3 reward managers (`NaiveRewardManager`, `DAPORewardManager`, `BatchRewardManager`), matching the pattern already used by `PrimeRewardManager`.

## Changes

- **`naive.py`**: Replace per-item decode loop with `batch_decode()` + zip
- **`dapo.py`**: Same refactor for `DAPORewardManager`
- **`batch.py`**: Same refactor for `BatchRewardManager.verify()`

## Benchmark

CPU benchmark (gpt2 tokenizer, ~256 tokens/seq):

| Batch Size | Per-item | batch_decode | Speedup |
|------------|----------|-------------|---------|
| 64 | 3.86 ms | 3.14 ms | 1.23x |
| 256 | 14.53 ms | 12.52 ms | 1.16x |
| 512 | 31.83 ms | 26.62 ms | 1.20x |
| 1024 | 62.69 ms | 54.61 ms | 1.15x |

Consistent 15-23% speedup across batch sizes. Gains should be larger with GPU tokenizers and longer sequences in real RLHF workloads.

## Correctness

- Output is identical (no padding tokens, same skip_special_tokens behavior)
- Follows the exact same pattern as `PrimeRewardManager` which already uses `batch_decode()`

### DIFF
--- a/verl/workers/reward_manager/batch.py
+++ b/verl/workers/reward_manager/batch.py
@@ -52,12 +52,8 @@ class BatchRewardManager(AbstractRewardManager):
         prompt_len = prompt_ids.shape[-1]
         valid_response_lengths = attention_mask[:, prompt_len:].sum(dim=-1)
 
-        responses_str = []
-        for i in range(len(data)):
-            valid_len = valid_response_lengths[i]
-            valid_response_ids = response_ids[i][:valid_len]
-            response_str = self.tokenizer.decode(valid_response_ids, skip_special_tokens=True)
-            responses_str.append(response_str)
+        all_valid_response_ids = [response_ids[i][:valid_response_lengths[i]] for i in range(len(data))]
+        responses_str = self.tokenizer.batch_decode(all_valid_response_ids, skip_special_tokens=True)
 
         ground_truths = [item.non_tensor_batch["reward_model"].get("ground_truth", None) for item in data]
         data_sources = data.non_tensor_batch[self.reward_fn_key]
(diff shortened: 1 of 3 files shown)


dubin555:
Thanks for the review\! Good catch on the duplication — the batch decoding logic in `naive.py` and `dapo.py` is indeed identical.

I intentionally kept this PR scoped to the performance optimization only (replacing per-item `decode()` with `batch_decode()`), to make it easier to review and less risky.

I'll submit a follow-up PR to extract the shared decoding logic into a helper method on `AbstractRewardManager`, as suggested.

vermouth1992:
@yyDing1 Could you please help review?

yyDing1:
Yes, I’ve briefly looked through the content, and here are my concerns:

The reward manager no longer processes reward computation sequentially at the batch level. The previous implementation under `verl/workers/reward_manager` has been deprecated. We now use the version under `verl/experimental/reward_loop/reward_manager`, which processes samples in a sample-async manner instead of batch-sequential execution.

So the current modification might not be compatible with the updated reward pipeline.

dubin555:
Thanks for the review @yyDing1! You're right — I wasn't aware that the reward manager has been migrated to `verl/experimental/reward_loop/reward_manager/` with sample-async processing. Since the files under `verl/workers/reward_manager/` are deprecated, this optimization is no longer relevant to the active code path.

Closing this PR. If the new async reward manager could benefit from similar batch decoding optimizations, I'd be happy to look into that in a separate PR targeting the correct location.

PR #5455, opened 2026-03-02
[megatron] feat: empty_cache after save_ckpt

KimmiShi:
### What does this PR do?

empty_cache after save_ckpt to avoid Engine wakeup OOM.

save ckpt involves NCCL to collect opt states which could cause memory fragments.


ISEEKYAN:
hi @KimmiShi , thanks for the contribution, why close this PR, is this `empty_cache` effective?

PR #5472, opened 2026-03-03
[trainer] fix: add trainable flag for masking assistant messages

### LINKED ISSUE #5316: Ability to mask assistant with zeros in SFTDataset
Couldn't find a way to mask the assistant with zeros in `MultiTurnSFTDataset`. For example, if I want to set `loss_mask=0` for all of the assistant's responses except the last one, this can also be useful in long agent trajectories, where some intermediate steps may be incorrect and it would be desirable to explicitly disregard them during training.

One option is to add the `trainable` flag to the message and check it when creating the `loss_mask`.
```json
{
    "role": ...,
    "content": ...,
    "trainable":...,
}
```

heiseon:
## Summary                                                                                                                                                                                             
   Add support for a `trainable` flag in `MultiTurnSFTDataset` to selectively mask specific assistant messages from loss computation during training.                                                     
                                                                                                                                                                                                          
   ## Motivation                                                                                                                                                                                          
   This addresses Issue #5316: In long agent trajectories, some intermediate steps may be incorrect and it is desirable to explicitly disregard them during training. Users need the ability to set       
 `loss_mask=0` for specific assistant responses while keeping others trainable.                                                                                                                           
                                                                                                                                                                                                          
   ## Changes                                                                                                                                                                                             
   - **verl/utils/dataset/multiturn_sft_dataset.py**: Added `trainable` flag check in message processing. When `trainable=False` is set on an assistant message, the entire response gets `loss_mask=0`   
 (excluded from training).                                                                                                                                                                                
   - **tests/utils/dataset/test_multiturn_sft_dataset_on_cpu.py**: Added comprehensive test cases covering:                                                                                               
     - Default behavior (no flag, assistant is trainable)                                                                                                                                                 
     - Explicit `trainable=False` (message masked)                                                                                                                                                        
     - Explicit `trainable=True` (message trained)                                                                                                                                                        
     - Mixed scenarios with both trainable and non-trainable messages                                                                                                                                     
                                                                                                                                                                                                          
   ## Usage Example                                                                                                                                                                                       
   ```python                                                                                                                                                                                              
   messages = [                                                                                                                                                                                           
       {"role": "user", "content": "What is 2+2?"},                                                                                                                                                       
       {"role": "assistant", "content": "2+2 equals 5.", "trainable": False},  # Will NOT be trained                                                                                                      
       {"role": "user", "content": "That's wrong."},                                                                                                                                                      
       {"role": "assistant", "content": "Sorry, 2+2 equals 4.", "trainable": True},  # Will be trained                                                                                                    
   ]                                                                                                                                                                                                      
 ```                                                                                                                                                                                                      
                                                                                                                                                                                                          
 Test                                                                                                                                                                                                     
                                                                                                                                                                                                          
 - Logic verification passed                                                                                                                                                                              
 - Unit tests (requires full verl environment)                                                                                                                                                            
                                                                                                                                                                                                          
 Additional Context                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                          
 Fixes #5316


verl/utils/dataset/multiturn_sft_dataset.py
```diff
@@ -231,6 +231,12 @@ def _process_single_message(
             loss_mask = torch.ones_like(attention_mask)
             # mask out generation prompt if assistant message
             loss_mask[: len(self.generation_prompt)] = 0
+            
+            # Support trainable flag to mask specific assistant messages
+            # If trainable=False, the entire assistant response will be masked (loss=0)
+            trainable = message.get("trainable", None)
```
vermouth1992, line 239: Could you rename trainable to loss_mask directly. And loss_mask is 0 or 1

verl/utils/dataset/multiturn_sft_dataset.py
```diff
@@ -231,6 +231,12 @@ def _process_single_message(
...
+            if trainable is False:
```
vermouth1992, line 240: This line is confusing. We should directly use loss mask

vermouth1992 (changes requested):

Issue #5421, opened 2026-02-27
Legacy worker + dynamic batching: reduce_metrics crashes due to unsynchronized micro-batch counts across DP ranks


zpqiu:
## Bug Description

When using **legacy workers** (`dp_actor.py`, `megatron_actor.py`, etc.) with `use_dynamic_bsz=True` and **DP > 1**, `reduce_metrics` crashes with:

```
ValueError: setting an array element with a sequence. The requested array has an
inhomogeneous shape after 1 dimensions. The detected shape was (4,) + inhomogeneous part.
```

## Root Cause

`dp_actor.py` calls `prepare_dynamic_batch` **without passing `dp_group`**:

```python
# verl/workers/actor/dp_actor.py:560
micro_batches, _ = prepare_dynamic_batch(mini_batch, max_token_len=max_token_len)
#                                        ^ no dp_group!
```

Inside `rearrange_micro_batches` (`seqlen_balancing.py:391`), the sync logic is guarded by `dp_group is not None`:

```python
if dist.is_initialized() and same_micro_num_in_dp and dp_group is not None:
    num_micro_batches = torch.tensor([num_micro_batches], device=get_device_name())
    dist.all_reduce(num_micro_batches, op=dist.ReduceOp.MAX, group=dp_group)
```

Since `dp_group` is `None`, the `all_reduce` is skipped. Each DP rank independently computes:

```python
num_micro_batches = min(batch_size, ceildiv(total_seqlen, max_token_len))
```

When `balance_batch` distributes data with different token-length distributions across ranks, `total_seqlen` differs per rank, so `ceildiv` can yield different values. This means each rank produces a different number of micro-batches, and therefore a different-length metric list.

After `DataProto.concat` merges these via `list_of_dict_to_dict_of_list`, the result is a **jagged nested list**:

```python
{"actor/entropy": [[0.8, 0.9, 1.0], [0.7, 0.8], [0.9, 1.0, 1.1], [0.6, 0.7]]}
#                   rank0 (3 bs)    rank1 (2)    rank2 (3)         rank3 (2)
```

`np.mean()` on this jagged list raises `ValueError`.

## Why Engine Workers Are Not Affected

All engine worker implementations (FSDP, TorchTitan, Megatron, VeOmni) pass `dp_group` correctly:

```python
# e.g. verl/workers/engine/fsdp/transformer_impl.py:514
micro_batches, indices = prepare_micro_batches(
[...]

PR #5051, opened 2026-01-27
[trainer] feat: Add Torchtitan as alternative training engine

acisseJZhong:
### What does this PR do?
Integrate Torchtitan as a new training engine in Verl. This PR implements the basic APIs needed by Torchtitan Engine, and tested SFT trainer in verl (qwen3 0.6b):
- Torchtitan Engine matches exactly with FSDP engine for SFT trainer
- `use_remove_padding=True` matches `use_remove_padding=False`
- TP/SP and FSDP work with both varlen and flex attention; numerics match with single process. 

**Relevant PRs:**
- Torchtitan side changes in https://github.com/pytorch/torchtitan/pull/2333. 
- RFC for engine interfaces https://github.com/volcengine/verl/issues/1371
- Training engine interface design https://github.com/volcengine/verl/pull/1977 
- Add Veomini Engine https://github.com/verl-project/verl/pull/4072

**Todos:**
See roadmap here: https://github.com/verl-project/verl/issues/5306
- [ ] enable parallelism: enable PP, EP, CP
- [ ] make Torchtitan Engine work with RL trainer 
- [ ] test multimodal input(ref: https://github.com/verl-project/verl/pull/4492/changes)

### Checklist Before Starting

- [ ] Search for similar PRs. Paste at least one query link here: ...
- [ ] Format the PR title as `[{modules}] {type}: {description}` (This will be checked by the CI)
  - `{modules}` include `fsdp`, `megatron`, `veomni`, `sglang`, `vllm`, `rollout`, `trainer`, `ci`, `training_utils`, `recipe`, `hardware`, `deployment`, `ray`, `worker`, `single_controller`, `misc`, `perf`, `model`, `algo`, `env`, `tool`, `ckpt`, `doc`, `data`, `cfg`, `reward`
  - If this PR involves multiple modules, separate them with `,` like `[megatron, fsdp, doc]`
  - `{type}` is in `feat`, `fix`, `refactor`, `chore`, `test`
  - If this PR breaks any API (CLI arguments, config, function signature, etc.), add `[BREAKING]` to the beginning of the title.
  - Example: `[BREAKING][fsdp, megatron] feat: dynamic batching`

### Test
```
MODEL_ID=Qwen/Qwen3-0.6B BACKEND=torchtitan bash tests/special_e2e/sft/run_sft_engine.sh
MODEL_ID=Qwen/Qwen3-0.6B BACKEND=fsdp bash tests/special_e2e/sft/run_sft_engine.sh
```
[...]

PR #5511, opened 2026-03-06
[fsdp] fix: pass dp_group to prepare_dynamic_batch to fix potential hang

dubin555:
### What does this PR do?

After #5051 added a `dp_group is not None` guard in `rearrange_micro_batches`, the FSDP actor/critic calls to `prepare_dynamic_batch` (which do not pass `dp_group`) silently skipped the `num_micro_batches` all_reduce across data-parallel ranks.

Under dynamic batching, this causes different DP ranks to compute different numbers of micro-batches, leading to a deadlock in FSDP reduce-scatter during `backward()`.

This is the same root cause as #5451 which fixed the Megatron backend. This PR applies the equivalent fix to the FSDP backend.

**Root cause:**

```python
# verl/utils/seqlen_balancing.py:391
# Before #5051: if dist.is_initialized() and same_micro_num_in_dp:
# After  #5051: if dist.is_initialized() and same_micro_num_in_dp and dp_group is not None:
```

The added `dp_group is not None` guard was needed for TorchTitan, but it broke FSDP actor/critic which never passed `dp_group`.

**Example deadlock scenario:**

With `_balance_batch(equal_size=True)`, each DP rank gets the same number of samples but different total tokens (balanced by workload, not token count). With `max_token_len=100`:

- Rank 0: 5 long sequences (475 tokens) → 5 micro-batches → 5 backward() calls
- Rank 1: 5 short sequences (100 tokens) → 1 micro-batch → 1 backward() call

Rank 1 finishes after 1 backward. Rank 0's 2nd backward hangs waiting for rank 1 in FSDP reduce-scatter.

**Fix:**

Pass `dp_group` (from `ulysses_device_mesh["dp"]` or `torch.distributed.group.WORLD`) to `prepare_dynamic_batch` in `dp_actor.py` (2 call sites) and `dp_critic.py` (2 call sites), matching the Megatron fix in #5451.

### Checklist Before Starting

- [x] Search for similar PRs: https://github.com/verl-project/verl/pulls?q=dp_group+fsdp+prepare_dynamic_batch

### Test

The fix is structurally identical to #5451 (Megatron). The affected code paths (`compute_log_prob`, `update_policy`, `compute_values`, `update_critic`) share the same `prepare_dynamic_batch` function.

- [x] PR title follows the format `[{modules}] {type}: {description}`
- [x] Changes are minimal and focused on the fix

### DIFF
--- a/verl/workers/fsdp_workers.py
+++ b/verl/workers/fsdp_workers.py
@@ -1639,7 +1639,9 @@ class CriticWorker(Worker, DistProfilerExtension):
         else:
             dp_group = torch.distributed.group.WORLD
         self.critic = DataParallelPPOCritic(
-            config=self.config, critic_module=self.critic_module, critic_optimizer=self.critic_optimizer,
+            config=self.config,
+            critic_module=self.critic_module,
+            critic_optimizer=self.critic_optimizer,
             dp_group=dp_group,
         )
 


dubin555:
This PR fixes #5421. The root cause analysis in that issue matches exactly — `prepare_dynamic_batch` in the FSDP legacy workers does not pass `dp_group`, causing unsynchronized micro-batch counts across DP ranks.

wuxibin89:
@dubin555 Please format code according to https://github.com/verl-project/verl/blob/main/CONTRIBUTING.md#code-linting-and-formatting

dubin555:
Thanks for the reminder! I've formatted the code with `ruff` via pre-commit — please see commit 827049d5. Let me know if anything else needs attention.

wuxibin89:
Fixed in https://github.com/verl-project/verl/pull/5591

PR #5566, opened 2026-03-12
[ci] fix: trtllm ci wrong local geo3k data path

wuxibin89:
### What does this PR do?

As title


wuxibin89:
Fixed in https://github.com/verl-project/verl/pull/5583

PR #5592, opened 2026-03-14
[fsdp,vllm,trainer,algo] feat: On-Policy Distillation

JacobHelwig:
# What does this PR do?

Adds on-policy distillation to FSDP engine with top-k distillation loss and KL estimator distillation losses with supervised and PG-style updates. Teacher logprobs are computed using a vLLM teacher server. Supports both LLM and VLM distillation.

## Losses 

1. top-k distillation loss: **forward** KL estimated using top-k logits **from teacher**. 
2. KL estimator distillation losses: **reverse** KL estimated using only the log prob for the sampled token via the same estimators used by the reference model (e.g., k1, k3)  

## Updates

1. Supervised: distillation loss is directly backpropogated, as in https://arxiv.org/abs/2306.13649
2. Policy gradient: negative distillation loss is used as a reward, as in https://thinkingmachines.ai/blog/on-policy-distillation/

# Test

Tested with `examples/on_policy_distillation_trainer/run_qwen_gsmk8k.sh` for LLM distillation and `examples/on_policy_distillation_trainer/run_qwen3_vl_geo3k.sh` for VLM distillation.

### LLM Distillation 

These experiments compare 3 training runs with student model Qwen2.5-0.5B using `examples/on_policy_distillation_trainer/run_qwen_gsmk8k.sh`:

1. Forward top-k KL with Qwen2.5-3B-Instruct teacher (gold)
2. Forward top-k KL with Qwen2.5-7B-Instruct teacher (green)
3. k3 estimator KL with Qwen2.5-7B-Instruct teacher (red)

#### GSM8K eval acc

image

#### GSM8K train acc

image

#### Distillation loss

image

### VLM Distillation 

- Data: Geometry3K
- Student: Qwen3-VL-2B-Instruct
- Teacher: Qwen3-VL-4B-Instruct
- OPD algo: k1 KL estimator as reward with policy gradient loss

#### Geo3K eval acc

image

#### Geo3K train acc

image

#### Distillation loss

image

## LLM Distillation: Top-k training stability

Clamping the top-k forward KL loss was needed for training stability. These experiments compare 3 types of clamping: 

1. No clamping (grey)
2. Clamping the distillation loss to a maximum value of 10 (blue)
3. Clamping the student and teacher log probs to a minimum value of -10 (gold)

### Distillation loss

image

### GSM8K eval acc

image

### GSM8K train acc

image

## LLM Distillation: Policy-gradient results

While the VLM results in this PR use the k1 KL estimator with policy gradient updates, all LLM distillation results outside of this section rely on supervised updates. LLM distillation with policy gradient updates are validated in this section:

1. Forward top-k KL with supervised update (green)
2. k1 estimator KL with policy gradient update (purple)
3. k3 estimator KL with supervised update (red)

While purple seems best, it also is generating responses that exceed the maximum response length of 512.

### Distillation loss

image

### GSM8K eval acc

image

### GSM8K train acc

image

### Response length

image

## LLM Distillation: Note on reverse KL

Initially, this PR included top-k reverse KL and top-k Jensen-Shannon divergences (JSD interpolates between forward and reverse KL). For the student distribution $q$ and teacher distribution $p$, the top-k reverse KL is given by

$$
KL_{\text{top-}k}(q||p) = \sum_i \bf{1}(q_i\in \text{top-}k)q_i\log\frac{q_i}{p_i}.
$$ 

Unfortunately, this was unstable. The reason is because one way to make this loss small is to make $q_i$ as small as possible for all $q_i \in \text{top}-k$. This can be seen from the logs tracking the amount of mass captured in the top-$k$ probabilities:

image

## LLM Distillation: Ablation: performance with more lenient parser

Note that the only loss used is the distillation loss (no rewards for correctness on GSM8K). Any increase in the logged rewards=GSM8k accuracy are an indirect result of minimizing the distillation loss. The reason that the base model has Pass@1~=0 is because the default GSM8k answer formatting (`#### 42`) is OOD for the model. The base model is answering the questions correctly, but using incorrect formatting, so none of the answers can be parsed. The base model can be evaluated using a reward function that is more lenient on formatting by adding the following to the script:

```bash
...
    reward_model.reward_manager=remote \
    custom_reward_function.path=tests/experimental/reward_loop/reward_fn.py \
    custom_reward_function.name=compute_score_math_verify \
    trainer.val_only=True
``` 

The results are:

```bash
(TaskRunner pid=904198) ("Initial validation metrics: {'val-aux/openai/gsm8k/reward/mean@1': "
(TaskRunner pid=904198)  "np.float64(0.31766489764973466), 'val-core/openai/gsm8k/acc/mean@1': "
(TaskRunner pid=904198)  "np.float64(0.31766489764973466), 'val-aux/num_turns/min': np.int32(2), "
(TaskRunner pid=904198)  "'val-aux/num_turns/max': np.int32(2), 'val-aux/num_turns/mean': "
(TaskRunner pid=904198)  'np.float64(2.0)}')
```

# Design & Code Changes

- Teacher workers are used in the agent loop, similar to the generative reward model: after a student worker finishes its rollout, the teacher worker obtains logprobs
- The distillation loss is applied in the `ppo_loss` function  
- In the initial version of this PR (#4897), requests were submitted to the `vLLMHttpServer` via the `v1/completions` endpoint, which does not support multi-modal data. While `v1/chat/completions` does support multi-modal inputs, text must be passed as raw text instead of token IDs, preventing exact scoring of student generations since `student gen IDs -> student gen text -> teacher input IDs via v1/chat/completions tokenization` will not always give `student gen IDs == teacher input IDs` (https://vllm.ai/blog/agent-lightning). 
- This PR instead follows a path similar to how rollout replicas directly call the `generate` method on the `vLLMHttpServer`. This enables multi-modal inputs while representing text as token IDs. Requests to the teacher server now call the newly-added `compute_logprobs` method of `vLLMHttpServer`. 
- Here's the diff between this branch and #4897 : https://github.com/JacobHelwig/verl/compare/jhelwig/onPolicyDistillation...JacobHelwig:verl:jhelwig/opdServer.

### DIFF
--- a/verl/experimental/agent_loop/agent_loop.py
+++ b/verl/experimental/agent_loop/agent_loop.py
@@ -709,7 +709,15 @@ class AgentLoopWorker:
             response_ids=output.response_ids,
             validate=validate,
         )
-        teacher_logprobs, teacher_ids = teacher_result.get("response_logprobs"), teacher_result.get("response_ids")
+        if teacher_result:
+            teacher_ids, teacher_logprobs = teacher_result["response_ids"], teacher_result["response_logprobs"]
+            pad_size = self.config.actor_rollout_ref.rollout.response_length - teacher_ids.shape[1]
+            padding = (0, 0, 0, pad_size)  # pad the sequence dimension
+            teacher_ids = F.pad(teacher_ids, padding, value=self.tokenizer.pad_token_id)
+            teacher_logprobs = F.pad(teacher_logprobs, padding, value=0.0)
+        else:
+            teacher_ids, teacher_logprobs = None, None
+
         return _InternalAgentLoopOutput(
             prompt_ids=prompt_output["input_ids"],
             response_ids=response_output["input_ids"],
@@ -861,11 +869,7 @@ class AgentLoopWorker:
             response_ids = torch.tensor(response_ids_ls).unsqueeze(0)
             response_logprobs = torch.tensor(response_logprobs_ls).unsqueeze(0)
 
-            pad_size = self.config.actor_rollout_ref.rollout.response_length - response_ids.shape[1]
-            padding = (0, 0, 0, pad_size)  # pad the sequence dimension
-            response_ids_padded = F.pad(response_ids, padding, value=self.tokenizer.pad_token_id)
-            response_logprobs_padded = F.pad(response_logprobs, padding, value=0.0)
-            return {"response_ids": response_ids_padded, "response_logprobs": response_logprobs_padded}
+            return {"response_ids": response_ids, "response_logprobs": response_logprobs}
         else:
             return {}
 


verl/workers/engine/fsdp/transformer_impl.py
```diff
@@ -1041,6 +1053,13 @@ def prepare_model_outputs(self, output, output_args, micro_batch: TensorDict):
                     unpad_dim=0,
                     padding_size=pad_size,
                 )
+                if self.need_logits_for_distillation:
+                    logits_rmpad = gather_outputs_and_unpad(
```
vermouth1992, line 1057: Gather full logits will incur OOM in most cases, especially when the model is large. We should compute the kld loss when logits are distributed. This applies in all model engines. cc @wuxibin89

vermouth1992: The central idea here is that: 1) input is distributed in sequence parallel -> output logits is in sequence parallel -> a reduction function that reduce the distributed logits into a tensor that removes the vocab dim

wuxibin89: The megatron version has implemented vocab parallel KL divergence:
https://github.com/verl-project/verl/pull/5041/changes#diff-8204316f3607fc6be6a48cb134efab24df6974f1d76ece406c838d3e1632f8afR217-R228

vermouth1992:
General comments:
- The model engine is designed to be model agnostic.
- The vocab dim of the logits must be reduced in distributed settings to avoid OOM
- Thus, for each backend, we have to be able to set the logits_processor from single controller so that it can be used to implement a wide range of loss functions without modifying the model engine

verl/workers/rollout/vllm_rollout/vllm_async_server.py
```diff
@@ -611,6 +611,96 @@ async def generate(
...
+
+        prompt_ids = normalize_token_ids(prompt_ids)
+        response_ids = normalize_token_ids(response_ids)
+        prompt_ids = qwen2_5_vl_dedup_image_tokens(prompt_ids, self.model_config.processor)
+        input_ids = prompt_ids + response_ids
+
+        # Calculate the maximum possible new tokens based on available context space
+        # This serves as a safety upper bound
+        max_possible_tokens = self.config.max_model_len - len(input_ids)
+        if max_possible_tokens < 0:
```
wuxibin89: To leave at least one token budget for teacher server generate, we should adjust `max_tokens` in rollout generation?

verl/workers/engine/fsdp/transformer_impl.py
vermouth1992: > The megatron version has implemented vocab parallel KL divergence: https://github.com/verl-project/verl/pull/5041/changes#diff-8204316f3607fc6be6a48cb134efab24df6974f1d76ece406c838d3e1632f8afR217-R228

Yeah, we will need an abstracted way of passing various distributed loss function in a unified interface

vermouth1992: We should maintain the principle that model engine is loss agnostic

verl/experimental/agent_loop/agent_loop.py
```diff
@@ -786,10 +860,48 @@ async def _compute_score(self, output, prompts, responses, attention_mask, input
...
+            response_ids_ls, response_logprobs_ls = (
+                output["response_ids"],
+                output["response_logprobs"],
+            )
+            # Shapes: # 1, S, (1 or K), where S is the response length, K is either 1 or topk depending on
+            # the distillation loss settings.
+            response_ids = torch.tensor(response_ids_ls).unsqueeze(0)
+            response_logprobs = torch.tensor(response_logprobs_ls).unsqueeze(0)
+
+            pad_size = self.config.actor_rollout_ref.rollout.response_length - response_ids.shape[1]
```
wuxibin89: Don't pad it here, add `response_ids` and `response_logprobs` to AgentLoopOutput and processed as other fields.

wuxibin89: We will soon remove the whole padding logic.

verl/workers/rollout/vllm_rollout/vllm_async_server.py
```diff
@@ -611,6 +611,96 @@ async def generate(
             extra_fields={"global_steps": self.global_steps},
         )
 
+    async def compute_logprobs(
```
wuxibin89: Do we need a separate compute_logprobs function? We can reuse generate function with:
- sampling_params: max_tokens=1, num_logprobs=topk
- additional post process for logprobs

verl/workers/rollout/vllm_rollout/vllm_async_server.py
```diff
@@ -613,6 +613,97 @@ async def generate(
...
+
+        use_topk = logprob_params["use_topk"]
+        if use_topk:
+            num_logprobs = topk = logprob_params["topk"]
+        else:
+            num_logprobs = 0
+
+        temp = logprob_params["temperature"]
+        if temp != 1.0:
+            raise NotImplementedError("vLLM doesn't support temperature for prompt logprobs")
```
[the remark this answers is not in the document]
pengwu22: agree with gemini.

The teacher's temperature is currently inherited from the student rollout config in the `distillation.yaml` via ${oc.select:actor_rollout_ref.rollout.temperature}.
And the student rollout temperature not necessarily be one.

verl/workers/rollout/vllm_rollout/vllm_async_server.py
JacobHelwig: Good idea, added here: https://github.com/verl-project/verl/pull/5592/changes/4408005be23b28776baace150763fc80b71fc254

verl/trainer/distillation/utils.py
```diff
@@ -0,0 +1,89 @@
...
+        return {}
+    if logits is None:
+        raise ValueError(f"logits must be provided for distillation loss computation with {loss_config.loss_mode=}.")
+    if cu_seqlens is None:
+        if not logits.is_nested:
+            raise ValueError("cu_seqlens must be provided if logits is not a nested tensor.")
+        cu_seqlens = logits.offsets()
+        logits = logits.values()
+    if distillation_settings.use_topk:
+        nested_logits = torch.nested.nested_tensor_from_jagged(logits, cu_seqlens)
```
pengwu22, line 55: The `use_topk` branch here constructs a nested tensor from jagged logits, but `prepare_distillation_inputs` later calls `no_padding_2_padding` to re-densify it back to `[B, S, V]` before passing to `compute_forward_kl_topk` — the nested tensor is an intermediate form that nobody actually consumes. 

But this round-trip is likely a consequence of the broader OOM issue raised already in the engine-related comment below: if KL is computed in vocab-parallel shards, both functions would probably become unnecessary for the `use_topk` path. 

Guess it's worth resolving the potential gathering OOM issue first and seeing if this part of code can simply be removed.

verl/workers/rollout/vllm_rollout/vllm_async_server.py
```diff
@@ -579,6 +582,37 @@ async def generate(
             final_res = output
         assert final_res is not None
 
+        # Extract prompt logprobs
```
wuxibin89: Please move it to a separate function.

verl/workers/rollout/vllm_rollout/vllm_async_server.py
JacobHelwig: In the new agent loop, we pass in `max_tokens=1`: https://github.com/JacobHelwig/verl/blob/4408005be23b28776baace150763fc80b71fc254/verl/experimental/agent_loop/agent_loop.py#L843

I also added a check here: 848de65

verl/experimental/agent_loop/agent_loop.py
JacobHelwig: Moved padding out: 3983861

verl/workers/rollout/vllm_rollout/vllm_async_server.py
JacobHelwig: Sure, refactored it: c754a44

verl/trainer/distillation/utils.py
JacobHelwig: Nice catch, yes, let's keep this as is for now and re-visit later on

verl/workers/rollout/vllm_rollout/vllm_async_server.py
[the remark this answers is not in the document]
JacobHelwig: Yes, the current design makes it so that for student temp $\neq1$, teacher temp must be explicitly set to 1 to avoid the error. I think this is a safe path to make sure it is completely intentional to distill a low temp student with a higher temp teacher. Low temp student with a higher temp teacher seems a bit strange to me, at least in the original GKD paper (https://arxiv.org/abs/2306.13649), they did the opposite: 
image

JacobHelwig:
Diff since last review: https://github.com/JacobHelwig/verl/compare/7e73a0a...3983861

verl/workers/rollout/vllm_rollout/vllm_async_server.py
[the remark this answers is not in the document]
JacobHelwig: We have an open PR in the vLLM repo, hopefully changing teacher temp can be supported soon: https://github.com/vllm-project/vllm/pull/37082

PR #5606, opened 2026-03-16
[docs] fix: clarify that hf_model in save_contents is ignored when mbridge is enabled

khazic:
## Summary

- When mbridge is active (the default for Megatron engine), `model` in `save_contents` already saves weights in HF format via `bridge.save_weights()`. The `hf_model` option is silently ignored in this mode, which can confuse users into thinking their HF model was saved when it was not.
- Added documentation/comments to `CheckpointConfig`, `megatron.yaml`, and `McoreEngineConfig` clarifying this behavior
- Added a runtime warning in `MegatronCheckpointManager` when `hf_model` is used with mbridge enabled

## Test plan

- [ ] Verify the warning is emitted when `save_contents=[hf_model]` with mbridge enabled
- [ ] Verify no warning when using `save_contents=[model]` with mbridge enabled
- [ ] Verify no warning when using `save_contents=[hf_model]` with mbridge disabled

### DIFF
--- a/verl/trainer/config/engine/megatron.yaml
+++ b/verl/trainer/config/engine/megatron.yaml
@@ -82,9 +82,13 @@ override_transformer_config:
 override_mcore_model_config: {}
 
 # oc.select: default val for ref.megatron.use_mbridge
+# When enabled, model weights are loaded/saved in HF format via mbridge (bridge.save_weights),
+# which means checkpoint.save_contents='hf_model' is redundant and will be silently ignored.
+# Use 'model' in save_contents instead — it already saves HF format when mbridge is active.
 use_mbridge: True
 
 # oc.select: default val for ref.megatron.vanilla_mbridge
+# True: use vanilla mbridge (AutoBridge); False: use megatron-bridge provider
 vanilla_mbridge: True
 
 # whether to use thd format (sequence packing), if not, use bshd format, padding the input_ids to the longest sequence length
--- a/verl/trainer/config/config.py
+++ b/verl/trainer/config/config.py
@@ -29,6 +29,10 @@ class CheckpointConfig(BaseConfig):
     Args:
         save_contents (list[str]): What to include in saved checkpoints.
             Options: 'model', 'optimizer', 'extra', 'hf_model'.
+            NOTE: When mbridge is enabled (engine.use_mbridge=True, the default for Megatron),
+            'model' already saves weights in HF format via bridge.save_weights(), so 'hf_model'
+            is redundant and will be silently ignored. Use 'hf_model' only when mbridge is
+            disabled (pure Megatron dist_checkpointing mode) to convert Megatron format to HF.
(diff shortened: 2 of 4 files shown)


verl/trainer/config/engine/megatron.yaml
```diff
@@ -82,9 +82,13 @@ override_transformer_config:
 override_mcore_model_config: {}
 
 # oc.select: default val for ref.megatron.use_mbridge
+# When enabled, model weights are loaded/saved in HF format via mbridge (bridge.save_weights),
+# which means checkpoint.save_contents='hf_model' is redundant and will be silently ignored.
```
ETOgaosion, line 86: Thanks for help to make this clearer! `hf_model` not mentioned in configs in

https://github.com/verl-project/verl/blob/77476af84cc074edf5a6437f8d5ea418d7a54916/verl/trainer/config/_generated_ppo_megatron_trainer.yaml#L104-L107

Which is not aligned with the implementation. Please help add `hf_model` to actor and critic config~

https://github.com/verl-project/verl/blob/77476af84cc074edf5a6437f8d5ea418d7a54916/verl/trainer/config/actor/actor.yaml#L127

ETOgaosion:
So, let's make things clearer for mcore checkpoint mechanism.

First, refer to FSDP checkpoint APIs, `model` now means save sharded models, and `hf_model` will save the whole model.

For mcore, there is no sharded models, and have dist_checkpoint and mbridge backends, so in theory:

- When mbridge enabled, `model` and `hf_model` should work the same, they all save hf_model
- When mbridge disabled and use dist_checkpoint, `model` should use mcore dist_checkpoint implementation and hf_model should save hf_model use verl's implementation (old methods, should deprecate to reduce maintenance burden)

For future and checkpoint manager refactor:
- mbridge shall be enabled by default
- dist_checkpoint as a fall back method in background when `model` in `save_contents`
    - We can leave it to be an API default as false
    - We will not support both mbridge and dist_checkpoint as false
- `hf_model` should only supported by mbridge

Shown in Figure:

image

cc @ISEEKYAN 

Also placed in [https://github.com/verl-project/verl/issues/5630](https://github.com/verl-project/verl/issues/5630)

verl/trainer/config/engine/megatron.yaml
khazic: Will this part be added in?

khazic:
Until mbridge support for this part is completed, we need to inform users that when mbridge is enabled by default, enabling only hf_model will not save hf_ckpt.

khazic:
Closing this as stale. It has been open for about five months without review activity, and the branch has drifted far enough from `main` that the change would need to be revalidated from scratch rather than simply rebased.

No hard feelings about the queue, I know it is long. If a maintainer wants this after all, I am happy to reopen it and resubmit against current `main`.

PR #5617, opened 2026-03-16
[megatron, ckpt, worker] fix: clear CUDA cache after checkpoint save

yuki731:
### What does this PR do?

  This PR fixes a Megatron checkpoint-save memory cleanup gap.

  `save_checkpoint()` in Megatron worker/engine paths did not clear CUDA cache after checkpoint save, which could leave reserved VRAM behind and lead to OOM in subsequent training steps. This change adds `aggressive_empty_cache(force_sync=True)` after checkpoint
  save in the affected paths and adds regression tests to cover the behavior.

  ### Checklist Before Starting

    - `{modules}` include `fsdp`, `megatron`, `veomni`, `sglang`, `vllm`, `rollout`, `trainer`, `ci`, `training_utils`, `recipe`, `hardware`, `deployment`, `ray`, `worker`, `single_controller`, `misc`, `perf`, `model`, `algo`, `env`, `tool`, `ckpt`, `doc`,
  `data`, `cfg`, `reward`, `fully_async`, `one_step_off`

  ### Test

  Added a regression test that executes the extracted `save_checkpoint()` method bodies with mocks and verifies `aggressive_empty_cache(force_sync=True)` is called after checkpoint save in all affected Megatron paths.

  Commands run locally:

  ```bash
  python -m pytest tests/workers/test_megatron_checkpoint_cache_on_cpu.py -q
  env PRE_COMMIT_HOME=/tmp/pre-commit-cache python -m pre_commit run --all-files --show-diff-on-failure --color=always

  Results:

  3 passed in 0.03s
  pre-commit: passed

  I also attempted a Docker-based runtime validation path, but the available local runtime/container setup was not stable enough for a reliable Megatron execution check on this machine. Because of that, this PR includes a targeted regression test for the
  checkpoint-save path itself. Validation on the actual training environment is still recommended.

  ### API and Usage Example

  No API change.

  # No user-facing API or config change.

  ### Design & Code Changes

  High-level change:

  - Ensure Megatron checkpoint-save paths perform the same post-operation cache cleanup that other high-memory Megatron paths already do.

  Specific changes:

  - Add aggressive_empty_cache(force_sync=True) after ActorRolloutRefWorker.save_checkpoint()
  - Add aggressive_empty_cache(force_sync=True) after CriticWorker.save_checkpoint()
  - Add aggressive_empty_cache(force_sync=True) after MegatronEngine.save_checkpoint()
  - Add regression tests for the three save paths

  > [!IMPORTANT]

  - [x] Read the Contribute Guide (https://github.com/volcengine/verl/blob/main/CONTRIBUTING.md).
  - [x] Apply pre-commit checks (https://github.com/volcengine/verl/blob/main/CONTRIBUTING.md#code-linting-and-formatting): pre-commit install && pre-commit run --all-files --show-diff-on-failure --color=always
  - [ ] Add / Update the documentation (https://github.com/volcengine/verl/tree/main/docs). No doc update needed because there is no API or user-facing behavior change.
  - [x] Add unit or end-to-end test(s) to the CI workflow (https://github.com/volcengine/verl/tree/main/.github/workflows) to cover all the code. If not feasible, explain why: Added a focused regression test under the existing test suite. No workflow change was
    needed. Full Megatron runtime validation was not feasible in the current local environment.
  - [ ] Once your PR is ready for CI, send a message in the ci-request channel (https://verl-project.slack.com/archives/C091TCESWB1) in the verl Slack workspace (https://join.slack.com/t/verl-project/shared_invite/zt-3855yhg8g-CTkqXu~hKojPCmo7k_yXTQ). (If not a
    ccessible, please try the Feishu group (飞书群) (https://applink.larkoffice.com/client/chat/chatter/add_by_link?link_token=772jd4f1-cd91-441e-a820-498c6614126a).)
  - [ ] If your PR is related to the recipe submodule, please also update the reference to the submodule commit via git submodule update --remote or cd recipe && git pull origin main. N/A.


verl/workers/megatron_workers.py
```diff
@@ -1284,3 +1285,4 @@ def save_checkpoint(self, checkpoint_path, hdfs_path=None, global_steps=0, max_c
         )
         if self._is_offload_param:
             offload_megatron_model_to_cpu(self.critic_module)
+        aggressive_empty_cache(force_sync=True)
```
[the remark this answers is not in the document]
yuki731: Fixed in fa2f0184. Added `\torch.distributed.barrier()` to `CriticWorker.save_checkpoint()` for consistency with the other Megatron checkpoint-save paths, and updated the regression test accordingly.

tests/workers/test_megatron_checkpoint_cache_on_cpu.py
```diff
@@ -0,0 +1,162 @@
+# Copyright 2024 Bytedance Ltd. and/or its affiliates
```
ETOgaosion: Thanks for contribution! Seems that this test only check the execution contents during checkpoint saving, but not empty cache effect, which may not needed.

yuki731: Thanks. Correct, the test is scoped to the checkpoint-save execution path regression, i.e. verifying that　`aggressive_empty_cache(force_sync=True)` is invoked in the affected save paths.

 It does not validate the runtime memory release effect directly. I kept it at this level because the behavior change in this PR is the addition of the cache cleanup call itself, while an effect-level test would require a stable Megatron runtime setup and　reproducible memory-pressure scenario.

tests/workers/test_megatron_checkpoint_cache_on_cpu.py
```diff
@@ -0,0 +1,162 @@
...
+        _is_offload_optimizer=True,
+        actor_module="actor_module",
+        actor_optimizer="actor_optimizer",
+        checkpoint_mananager=checkpoint_manager,
+    )
+
+    save_checkpoint(worker, "/tmp/ckpt", global_step=7)
+
+    assert events == [
+        "load_model:actor_module",
```
ETOgaosion: @yuki731 Since that we may refactor and change the implementation in future, we shall not maintain these hard-code checks, you can just keep your implementation, they are covered by other CI test points~

yuki731: Understood. I removed the hard-coded regression test and kept the implementation change only. Updated in af2ebd52.

ETOgaosion:
@yuki731 Could you help rebase main and re-try some tests, we fixed some issue on ascend.

yuki731:
@ETOgaosion 
Rebased onto the latest `main` and force-pushed the branch. The intended checkpoint-save cleanup changes are still present after the rebase, and local pre-commit checks pass on the rebased branch.

PR #5667, opened 2026-03-19
[single_controller] fix: remove duplicate import of 'time' package

Fan-Yunfan:
### What does this PR do?

> This PR used to remove dupliacate import of `time` package and switch `signal` package to local lazying loading mode for better performance in `verl/single_controller/base/worker_group.py`.

#### 1. Fix the duplicate import of `time` package

I found that the `time` package was imported twice in line 24 and line 113. 

**The first time import:** 

image
---

**The second time import:** 

image

**Where time is used** 

**The first used place is in line 120 in `check_works_alive`（a top-level function）** 

image

**The second used place is in line 162 in `_block_until_all_workers_alive`（a member function of WorkerGroup class）** 

image

#### 2. Switch the import of `signal` package to local lazying loading mode for better performance

I found that the `signal` package only used in `check_workers_alive` this function in `verl/single_controller/base/worker_group.py`.

The `time` package was previously imported locally but had redundant entries in the header. I think the local import of `time` package is for better performance.

Following this pattern, I have refactored the `signal` package to use local loading for better performance."

```
- 19 import signal
......
- 113 import time
+ 113 import signal
```

### DIFF
--- a/verl/single_controller/base/worker_group.py
+++ b/verl/single_controller/base/worker_group.py
@@ -16,7 +16,6 @@ the class of WorkerGroup
 """
 
 import logging
-import signal
 import threading
 import time
 from typing import Any, Callable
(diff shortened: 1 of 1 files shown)


wuxibin89:
It's ok that import same package multiple times since python has cache

PR #5702, opened 2026-03-22
[megatron] fix: set finalize_model_grads_func for MegatronPPOCritic

thereAreDemonsNearby:
### What does this PR do?

In the \_\_init\_\_ method of MegatronPPOCritic, the finalize_model_grads_func attribute is not set for self.critic_module[0].config (https://github.com/verl-project/verl/blob/08e030d9b0d6f3c5c2f154ec28bf2ccb37cab375/verl/workers/critic/megatron_critic.py#L79). This may result in gradients not being reduced across DPs. In contrast, MegatronPPOActor handles this correctly (https://github.com/verl-project/verl/blob/08e030d9b0d6f3c5c2f154ec28bf2ccb37cab375/verl/workers/actor/megatron_actor.py#L189). This PR fixes the issue for MegatronPPOCritic.

This error may not be reflected in the training curve; I discovered it through strict numerical alignment.

### DIFF
--- a/verl/workers/critic/megatron_critic.py
+++ b/verl/workers/critic/megatron_critic.py
@@ -24,6 +24,7 @@ from typing import Iterable
 import torch
 import torch.distributed
 from megatron.core import parallel_state as mpu
+from megatron.core.distributed import finalize_model_grads
 from megatron.core.optimizer import DistributedOptimizer, OptimizerConfig
 from megatron.core.pipeline_parallel import get_forward_backward_func
 from omegaconf import OmegaConf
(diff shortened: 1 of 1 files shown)


wuxibin89:
The `finalize_model_grads_func` has been set in `register_megatron_training_hooks`:
https://github.com/verl-project/verl/blob/main/verl/workers/megatron_workers.py#L504
https://github.com/verl-project/verl/blob/main/verl/utils/megatron_utils.py#L1262-L1266

thereAreDemonsNearby:
We need to set use_legacy_worker_impl to 'disable' to avoid this problem，right？

wuxibin89:
> We need to set use_legacy_worker_impl to 'disable' to avoid this problem，right？

Critic model support in new model enigne(use_legacy_worker_impl=disable) is still under develop. The legacy megatron_workers.py has already set `finalize_model_grads_func` in `register_megatron_training_hooks`.

wuxibin89:
Support critic in new model engine https://github.com/verl-project/verl/pull/5604

PR #5713, opened 2026-03-23
[3/n][reward] feat: flowgrpo - support image-based rewards (rule-based & genrm)

chenyingshu:
### What does this PR do?

> Add **concise** overview of what this PR aims to achieve or accomplish. Reference related GitHub issues and PRs that help with the review.

Follow-up work for #4639, #5297. This is one of the PRs (others: #5616, #5716) to support FlowGRPO training for QwenImage.

This PR supports generative reward model and rule-based image rewards, which facilitates the training of image generation models.

- Add new class `ImageRewardManager` to support reward computation for visual input.
- Modify `RewardLoopManager` and `RewardLoopWorker` to support visual input pre-processing for reward computing.
- Add image reward support: Generative reward model (genrm) vllm-served Qwen-VL OCR, and a rule-based reward `jpeg_compressibility`.
- Add UT script `test_image_reward_manager.py` for testing genrm and rule rewards using reward loop.

### Checklist Before Starting

- [x] Search for similar PRs. Paste at least one query link here: ...
- [x] Format the PR title as `[{modules}] {type}: {description}` (This will be checked by the CI)
  - `{modules}` include `fsdp`, `megatron`, `veomni`, `sglang`, `vllm`, `rollout`, `trainer`, `ci`, `training_utils`, `recipe`, `hardware`, `deployment`, `ray`, `worker`, `single_controller`, `misc`, `perf`, `model`, `algo`, `env`, `tool`, `ckpt`, `doc`, `data`, `cfg`, `reward`, `fully_async`, `one_step_off`
  - If this PR involves multiple modules, separate them with `,` like `[megatron, fsdp, doc]`
  - `{type}` is in `feat`, `fix`, `refactor`, `chore`, `test`
  - If this PR breaks any API (CLI arguments, config, function signature, etc.), add `[BREAKING]` to the beginning of the title.
  - Example: `[BREAKING][fsdp, megatron] feat: dynamic batching`

### Test

> For changes that can not be tested by CI (e.g., algorithm implementation, new model support), validate by experiment(s) and show results like training curve plots, evaluation results, etc.

Provided UT script in `tests\experimental\reward_loop\test_image_reward_manager.py` and passed unit tests for OCR reward with Qwen-VL model, as well as a rule-based image reward `jpeg_compressibility`.

### API and Usage Example

> Demonstrate how the API changes if any, and provide usage example(s) if possible.

To apply image reward, set config:
```bash
reward.reward_manager.name="image"
```
[...]

PR #5616, opened 2026-03-16
[1/n][vllm, rollout] feat: flowgrpo - support vllm-omni as rollout backend for verl

knlnguyen1802:
### What does this PR do?
Follow up the work from https://github.com/verl-project/verl/pull/5297
This PR adds **vLLM-Omni** as a rollout backend, enabling diffusion-based image generation models (e.g., QwenImage) in verl RLHF pipelines with log-prob extraction and LoRA weight sync.

### Checklist Before Starting

- [ ] Search for similar PRs. Paste at least one query link here: ...
- [ ] Format the PR title as `[{modules}] {type}: {description}` (This will be checked by the CI)
  - `{modules}` include `fsdp`, `megatron`, `veomni`, `sglang`, `vllm`, `rollout`, `trainer`, `ci`, `training_utils`, `recipe`, `hardware`, `deployment`, `ray`, `worker`, `single_controller`, `misc`, `perf`, `model`, `algo`, `env`, `tool`, `ckpt`, `doc`, `data`, `cfg`, `reward`, `fully_async`, `one_step_off`
  - If this PR involves multiple modules, separate them with `,` like `[megatron, fsdp, doc]`
  - `{type}` is in `feat`, `fix`, `refactor`, `chore`, `test`
  - If this PR breaks any API (CLI arguments, config, function signature, etc.), add `[BREAKING]` to the beginning of the title.
  - Example: `[BREAKING][fsdp, megatron] feat: dynamic batching`

### Test

Unit tests added for vllm_omni rollout 
`tests/workers/rollout/rollout_vllm/test_vllm_omni_generate.py`

### API and Usage Example

> Demonstrate how the API changes if any, and provide usage example(s) if possible.

```python
# Add code snippet or script demonstrating how to use this
```

### Design & Code Changes

- **`vllm_base_async_server.py`** — Extracted `BaseVLLMHttpServer` / `BaseVLLMReplica` shared between vLLM and vLLM-Omni to eliminate duplication
- **`vllm_omni_async_server.py`** — `vLLMOmniHttpServer` wraps `AsyncOmni`; supports token-in/image-out generation, log-prob + latent capture, custom diffusion pipelines via config
- **`vllm_omni_rollout.py`** — `vLLMOmniServerAdapter` client adapter; lazily acquires named Ray actor and calls `collective_rpc` for weight sync
- **`verl/models/diffuser_model/vllm_omni/`** — New utility package:
[...]

PR #5297, opened 2026-02-12
[fsdp,trainer,vllm_omni,algo] feat: support FlowGRPO training for QwenImage

zhtmike:
### What does this PR do?

Follow-up Work for #4639 
- A trainable script for the [FlowGRPO](https://github.com/yifan123/flow_grpo) algorithm for Qwen-Image is provided.
- Support for `vLLM-Omni` has been added to the rollout engine.
- `Diffusers` has been integrated as the training engine for the diffusion model.
- The scripts have been tested with training using cfg/non-cfg, w/o kl loss, lora/full-model, fsdp/fsdp2, collocated mode, and async reward with a standalone resource pool.
- Unit tests for diffusion rollout/reward and diffusers training engines have been added.

### Checklist Before Starting

- [x] Search for similar PRs. Paste at least one query link here: ...
- [x] Format the PR title as `[{modules}] {type}: {description}` (This will be checked by the CI)
  - `{modules}` include `fsdp`, `megatron`, `veomni`, `sglang`, `vllm`, `rollout`, `trainer`, `ci`, `training_utils`, `recipe`, `hardware`, `deployment`, `ray`, `worker`, `single_controller`, `misc`, `perf`, `model`, `algo`, `env`, `tool`, `ckpt`, `doc`, `data`, `cfg`, `reward`
  - If this PR involves multiple modules, separate them with `,` like `[megatron, fsdp, doc]`
  - `{type}` is in `feat`, `fix`, `refactor`, `chore`, `test`
  - If this PR breaks any API (CLI arguments, config, function signature, etc.), add `[BREAKING]` to the beginning of the title.
  - Example: `[BREAKING][fsdp, megatron] feat: dynamic batching`

#### Install Method
We provide a quick way to test with our examples codes. The formal support will be released together with vllm-omni 0.17.

```bash
pip install vllm==0.16
git clone https://github.com/vllm-project/vllm-omni.git
cd vllm-omni && git reset --hard 92e1a544 && pip install . && cd ../
```

Experiments were conducted on 4 NVIDIA A100 GPUs using the OCR reward (Levenshtein distance) with Qwen3-VL-8B-Instruct model.

Test Dataset: [OCR Dataset](https://github.com/yifan123/flow_grpo/tree/main/dataset/ocr)
Training Sample Size: 19653
Test Sample Size: 1018

Training GPU hours required to reach and maintain a validation reward score of approximately 0.9:

[...]

Issue #4639, opened 2025-12-22
[RFC] Support Qwen-Image Flow-GRPO Training based on vLLM-Omni


chenyingshu:
# Motivation

The goal is to enhance `verl`’s scalability so that it can support online DPO-like training for state-of-the-art diffusion image and video generation models, including `Qwen-Image`, `Z-Image`, `Wan2.2`, and others. We choose `Flow-GRPO` as the representative algorithm in this domain, while additional algorithms such as `DiffusionNFT` and `DanceGRPO` can be seamlessly integrated following this update. As an initial step, `Qwen-Image` has been selected as the first supported model for multimodal generation tasks.

At present, `verl` does not support diffusion-based generation models. To enable this functionality, two major extensions are required: first, the addition of a rollout engine capable of handling image and video generation tasks, incorporating components such as `vLLM-Omni`; and second, the addition of a training engine for diffusion model training, which will rely on `diffusers` with an `FSDP` backend. Consequently, integrating `diffusers` and `vLLM-Omni` becomes a necessary change after this modification.

In the following section, we will first briefly present the overall picture of `FlowGRPO`, and then demonstrate the necessary code changes for enabling `Qwen-Image` `FlowGRPO` training.

# Overall Structure

![Image](https://github.com/user-attachments/assets/d514e0d0-bf46-4eaf-9516-141e33df50f1)

Figure 1. Overview of integrating the `FlowGRPO` algorithm into `verl`. The left panel shows the entry point and algorithm implementation with a standalone `RayFlowGRPOTrainer`. The right panel shows the corresponding workers (class names in **bold**) that need to be implemented. Other miscellaneous changes, such as configs, dataloaders and the logger, are not shown here.

## Algorithm Implementation

And here is a brief explanation of the FlowGRPO algorithm (functions) in the left panel:

**i. generate_sequence:** The sequence generation for the diffusion model produces a sequence of image/video latent samples during the diffusion process, rather than tokens as in LLMs. This includes the final generated image/video, prompt embeddings, timesteps, and other necessary information from the denoising stage.
- Inputs: Prompts
- Outputs: Generated Images/Videos; Prompt embeddings; Timesteps; Latent samples, log probabilities, and latent sample means during the sampling stage

**ii. compute_rm_score:** The reward model calculates the score of the generated images/videos based on the user’s task, such as `OCR`, `GenEval` Score, Clip Score, etc. For simplicity and general purposes, we may consider API calls to the vLLM server.
[...]

PR #5716, opened 2026-03-23
[2/n][rollout] feat: flowgrpo - add diffusion agent loop support

AndyZhou952:
### What does this PR do?

Follow-up work for #4639, #5297. This is one of the PRs (others: #5616, #5713) to support FlowGRPO training for QwenImage.

- Add `DiffusionAgentLoopWorker` and `DiffusionSingleTurnAgentLoop` for diffusion-based image/video rollout through the agent loop.
- Add diffusion-specific configs (`ppo_diffusion_trainer.yaml`, `diffusion_rollout.yaml`, `diffusion_model.yaml`) and output types (`DiffusionOutput`, `DiffusionAgentLoopOutput`).
- Performance optimization related to vllm-omni rollout engine: eliminate nested tensor version, etc.

### Checklist Before Starting

### Test

Provided UT script in `tests/experimental/agent_loop/test_diffusion_agent_loop.py` that validates end-to-end diffusion rollout through `AgentLoopManager` with the `diffusion_single_turn_agent loop`. 

### API and Usage Example

See `tests/experimental/agent_loop/test_diffusion_agent_loop.py`. Agent loop essentially serves as a point for diffusion rollout generation.

```python
```

### Design & Code Changes

- New `DiffusionAgentLoopWorker` paralleling `AgentLoopWorker` for image/video output.
- New `DiffusionSingleTurnAgentLoop` registered as `diffusion_single_turn_agent`. Tokenizes prompt, optionally tokenizes negative prompt for true CFG, and calls the server for diffusion generation.

> [!IMPORTANT]

### DIFF
--- a/verl/experimental/agent_loop/diffusion_agent_loop.py
+++ b/verl/experimental/agent_loop/diffusion_agent_loop.py
@@ -12,7 +12,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 import asyncio
-import copy
 import random
 from typing import Any, Optional
 
@@ -203,8 +202,7 @@ class DiffusionAgentLoopWorker:
             config=agent_loop_config,
             trainer_config=DictConfigWrap(config=self.config),
             server_manager=self.server_manager,
-            # TODO (andy): deepcopy is a workaround for transformers>=5.0.0, resource intensive & slow start, fix later
-            tokenizer=copy.deepcopy(self.tokenizer),
+            tokenizer=self.tokenizer,
             processor=self.processor,
             dataset_cls=self.dataset_cls,
             data_config=DictConfigWrap(self.config.data),
--- a/verl/workers/config/model.py
+++ b/verl/workers/config/model.py
@@ -302,9 +302,11 @@ class DiffusionModelConfig(BaseConfig):
         # construct tokenizer
         if self.load_tokenizer:
             self.local_tokenizer_path = copy_to_local(self.tokenizer_path, use_shm=self.use_shm)
-            # see issue https://github.com/huggingface/tokenizers/issues/537, we use a non-fast tokenizer here
+            # Fast tokenizer for diffusion: DiffusionSingleTurnAgentLoop applies chat template on the asyncio
+            # thread (not run_in_executor) so Rust-backed tokenizers avoid RuntimeError: Already borrowed
+            # with recent transformers when combined with thread-pool tokenization.
             self.tokenizer = hf_tokenizer(
-                self.local_tokenizer_path, trust_remote_code=self.trust_remote_code, use_fast=False
+                self.local_tokenizer_path, trust_remote_code=self.trust_remote_code, use_fast=True
             )
             if os.path.exists(os.path.join(self.local_path, "processor")):
                 self.processor = hf_processor(
--- a/verl/experimental/agent_loop/single_turn_agent_loop.py
+++ b/verl/experimental/agent_loop/single_turn_agent_loop.py
@@ -16,9 +16,14 @@ import os
 from typing import Any
 from uuid import uuid4
 
+import torch
+from PIL import Image
+
 from verl.experimental.agent_loop.agent_loop import AgentLoopBase, AgentLoopOutput, register
 from verl.experimental.agent_loop.diffusion_agent_loop import DiffusionAgentLoopOutput
+from verl.utils.chat_template import apply_chat_template
 from verl.utils.profiler import simple_timer
+from verl.utils.tokenizer import normalize_token_ids
 from verl.workers.rollout.replica import TokenOutput
 
 logger = logging.getLogger(__file__)
@@ -101,6 +106,50 @@ class DiffusionSingleTurnAgentLoop(AgentLoopBase):
             "index",
         }
     )
+
+    async def apply_chat_template(
+        self,
+        messages: list[dict],
+        tools: list[dict] | None = None,
+        images: list[Image.Image] | None = None,
+        videos: list[tuple[torch.Tensor, dict]] | None = None,
+        remove_system_prompt: bool = False,
+    ) -> list[int]:
+        """Tokenize on the asyncio thread for fast tokenizers when no processor is used.
+
+        Rust-backed fast tokenizers are not reliably safe across ``run_in_executor`` thread
+        boundaries with recent transformers (``RuntimeError: Already borrowed``). The diffusion
+        path is tokenizer-only for Qwen-Image-style models; keep tokenization on the event-loop
+        thread in that case.
+        """
+        if self.processor is not None:
+            return await super().apply_chat_template(
+                messages,
+                tools=tools,
+                images=images,
+                videos=videos,
+                remove_system_prompt=remove_system_prompt,
+            )
+        if getattr(self.tokenizer, "is_fast", False):
+            tokenized_prompt = apply_chat_template(
+                self.tokenizer,
+                messages,
+                tools=tools,
+                add_generation_prompt=True,
+                tokenize=True,
+                **self.apply_chat_template_kwargs,
+            )
+            prompt_ids = normalize_token_ids(tokenized_prompt)
+            if remove_system_prompt:
+                prompt_ids = prompt_ids[len(self.system_prompt) :]
+            return prompt_ids
+        return await super().apply_chat_template(
+            messages,
+            tools=tools,
+            images=images,
+            videos=videos,
+            remove_system_prompt=remove_system_prompt,
+        )
 
     async def run(self, sampling_params: dict[str, Any], **kwargs) -> DiffusionAgentLoopOutput:
         raw_prompt = kwargs.pop("raw_prompt")


tests/experimental/agent_loop/test_diffusion_agent_loop.py
```diff
@@ -0,0 +1,142 @@
...
+    config.actor_rollout_ref.rollout.nnodes = 1
+
+    qwen_pipeline = "examples.vllm_omni.pipeline_qwenimage.QwenImagePipelineWithLogProb"
+    config.actor_rollout_ref.rollout.engine_kwargs.vllm_omni = {"custom_pipeline": qwen_pipeline}
+    config.reward.reward_manager.name = "image"
+    config.trainer.n_gpus_per_node = 4
+
+    tokenizer_max_length = 1024
+    prompt_template_encode_start_idx = 34
+    max_length = tokenizer_max_length + prompt_template_encode_start_idx
```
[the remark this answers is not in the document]
AndyZhou952: cleaned

verl/experimental/agent_loop/agent_loop.py
```diff
@@ -968,6 +1020,315 @@ def _postprocess(
...
+    """
+
+    def __init__(
+        self,
+        config: DictConfig,
+        servers: list[tuple[str, ray.actor.ActorHandle]],
+        load_balancer_handle: ray.actor.ActorHandle,
+        reward_loop_worker_handles: list[ray.actor.ActorHandle] = None,
+        teacher_servers: list[tuple[str, ray.actor.ActorHandle]] = None,
+        teacher_load_balancer_handle: ray.actor.ActorHandle = None,
```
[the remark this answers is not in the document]
AndyZhou952: keep for compatibility and min disturbance to LLM loop - so that we do not have to change in `AgentLoopManager`

verl/experimental/agent_loop/agent_loop.py
```diff
@@ -968,6 +1020,315 @@ def _postprocess(
         )
 
 
+class DiffusionAgentLoopWorker:
```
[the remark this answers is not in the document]
zhtmike: @AndyZhou952 Can you take a look?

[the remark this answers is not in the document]
AndyZhou952: I kept it separated for now as:
 - Isolating diffusion limits disturbance in the main LLM loop
 - consistent with the split in the rollout layer for clarity (TokenOutput vs DiffusionOutput)
 -  different output type / handling logic (i.e. no tokenizer reset) in `postprocess`, sampling params are different. etc.
 - Extensibility for future adaptation.

Ntl will take a deeper dive into this

verl/trainer/config/rollout/diffusion_rollout.yaml
```diff
@@ -0,0 +1,244 @@
+# Target class for this configuration
+_target_: verl.workers.config.DiffusionRolloutConfig
```
wuxibin89: Since `DiffusionRolloutConfig` inherits `RolloutConfig`, the yaml config should also inherits `rollout.yaml`
```yaml
defaults:
  - rollout

  - _self_

_target_: verl.workers.config.DiffusionRolloutConfig

# image height for diffusion model rollout
height: 512

# image width for diffusion model rollout
width: 512

# number of inference steps for diffusion model rollout
num_inference_steps: 10

```

verl/experimental/agent_loop/agent_loop.py
wuxibin89: Please move diffusion related code to a separate file, the `agent_loop.py` file is too long.

verl/experimental/agent_loop/diffusion_agent_loop.py
```diff
@@ -0,0 +1,391 @@
...
+    """Reward score for the trajectory."""
+    num_turns: int = 0
+    """Number of chat turns, including user, assistant, tool."""
+    metrics: AgentLoopMetrics
+    """Auxiliary performance metrics"""
+    extra_fields: dict[str, Any] = {}
+    """Extra fields for dynamic addition."""
+
+
+class _InternalDiffusionAgentLoopOutput(DiffusionAgentLoopOutput):
```
wuxibin89, line 64: We're going to deprecate padded DataProto with TensorDict, and use [nested jagged tensor](https://docs.pytorch.org/docs/stable/nested.html) for padding free data transfer.

So we may not need `_InternalDiffusionAgentLoopOutput` any more.

wuxibin89: New diffusion trainer should not use `DataProto` either.

AndyZhou952: Thanks for the info. Do we have an issue for this?

Also want to confirm, should we remove the `DataProto` dependency in this PR, or we keep it as is right now and update separately?

verl/trainer/config/rollout/diffusion_rollout.yaml
AndyZhou952: Updated to reflect the inheritance

verl/experimental/agent_loop/agent_loop.py
AndyZhou952: moved to a separate `diffusion_agent_loop.py`

verl/experimental/agent_loop/diffusion_agent_loop.py
wuxibin89: We can remove `DataProto` in separate PR

verl/trainer/config/ppo_diffusion_trainer.yaml
```diff
@@ -0,0 +1,296 @@
...
+  - ref@actor_rollout_ref.ref: ${model_engine}_ref
+
+  # Rollout model config.
+  - rollout@actor_rollout_ref.rollout: diffusion_rollout
+
+  # Model config.
+  - model@actor_rollout_ref.model: diffusion_model
+
+  # Critic model config.
+  - critic@critic: ${model_engine}_critic
```
wuxibin89, line 28: This config file should be simplified, many fields is not needed by diffusion trainer

wuxibin89: Better to write a new config for diffusion trainer instead of reuse ppo trainer.

wuxibin89: BTW. the `actor_rollout_ref` is a bad design, we're going to rewrite the ppo trainer yaml config. 
```yaml
model: ?

actor:
    engine: megatron
    engine.optim.lr = 1e-6
    engine.tensor_model_parallel = 4
    loss: ppo
    loss.clip_high = 0.28
    model.path = xxx
    
rollout:
    model.path = ${actor.model.path}

ref:
    engine: ${actor.engine}

reward:
```

AndyZhou952: Agree. These changes (`critic` clean up in config, separate trainer instead of reusing ppo) will be reflected in the follow-up trainer PR @zhtmike .

For the updates made, in this agent loop PR, we 
- renamed to `diffusion_trainer.yaml`
- keep the minimum set of configs necessary up to this point in `diffusion_trainer.yaml`. PTAL if this works or if we prefer to keep the full set (ex. critics parts) for now.

verl/experimental/agent_loop/diffusion_agent_loop.py
```diff
@@ -0,0 +1,391 @@
...
+                elif k in ["prompt_embeds_mask", "negative_prompt_embeds_mask"]:
+                    pad_tuple = (0, self.max_prompt_embed_length - v.shape[0])
+                    v = F.pad(v, pad_tuple, value=0)
+                extra_fields[k] = v.unsqueeze(0)
+            else:
+                extra_fields[k] = v
+
+        extra_fields["raw_prompt"] = kwargs["raw_prompt"]
+
+        self.tokenizer.padding_side = "left"
```
SamitHuang, line 233: Is it extendable for other models? or should we parse `padding_side` from model config file?

AndyZhou952: I think likely unnecessary in the diffusion use case where we always left pad prompts. Also as mentioned in `agent_loop.py`, padding will be removed after the refactor.

verl/trainer/config/ppo_diffusion_trainer.yaml
wuxibin89: So in follow-up trainer PR, the diffusion trainer will start a completely new trainer config? There's too much historical burden in `ppo_trainer.yaml` that we strongly against diffusion trainer reuse it.

zhtmike: Possibly not to inherit `ppo_trainer` in this PR @AndyZhou952 , I will add other necessary configs later in trainer PR.

AndyZhou952: updated to remove the dependency on ppo trainer. Also removed a few unused configs for now

tests/experimental/agent_loop/test_diffusion_agent_loop.py
```diff
@@ -0,0 +1,140 @@
...
+
+        expected_batch_keys = [
+            "responses",
+            "all_latents",
+            "all_timesteps",
+            "prompt_embeds",
+            "prompt_embeds_mask",
+            "input_ids",
+            "attention_mask",
+        ]
```
zhtmike, line 131: btw why drop the "rollout_log_probs" key check, which is originally stated in https://github.com/verl-project/verl/pull/5297/changes#diff-beee54ba86046ccfc87f71f3df284f6237fe2ddcd454a32241a3336b3d32fa32?

This key is for training mode with `bypass_mode=True`, better keep it as.

tests/experimental/agent_loop/test_diffusion_agent_loop.py
```diff
@@ -0,0 +1,140 @@
...
+    model_path = os.path.expanduser("~/models/tiny-random/Qwen-Image")
+    config.actor_rollout_ref.model.path = model_path
+    config.actor_rollout_ref.model.tokenizer_path = os.path.join(model_path, "tokenizer")
+    config.actor_rollout_ref.rollout.name = "vllm_omni"
+    config.actor_rollout_ref.rollout.mode = "async"
+    config.actor_rollout_ref.rollout.enforce_eager = True
+    config.actor_rollout_ref.rollout.n = 4
+    config.actor_rollout_ref.rollout.num_inference_steps = 10
+
+    config.actor_rollout_ref.rollout.agent.num_workers = 2
```
zhtmike, line 43: and `config.actor_rollout_ref.rollout.calculate_log_probs = True` is missing

verl/experimental/agent_loop/diffusion_agent_loop.py
```diff
@@ -0,0 +1,392 @@
...
+        assert agent_name in _agent_loop_registry, (
+            f"Agent loop {agent_name} not registered, registered agent loops: {_agent_loop_registry.keys()}"
+        )
+
+        agent_loop_config = _agent_loop_registry[agent_name]
+        agent_loop = hydra.utils.instantiate(
+            config=agent_loop_config,
+            trainer_config=DictConfigWrap(config=self.config),
+            server_manager=self.server_manager,
+            tokenizer=copy.deepcopy(self.tokenizer),
```
wuxibin89: Why deep copy tokenizer?

SamitHuang: The ci error is `RuntimeError: Already borrowed`, which is caused by the Rust RefCell borrow checker inside the fast tokenizer. We can either 1) deep copy to  give each task its own tokenizer instance to avoid the shared mutable state, or 2) use slow tokenizer (no Rust, no RefCell, no borrow checker).  option 1 should give faster performance

verl/workers/config/model.py
zhtmike: Please change to use_fast=True here if changes to use deep copy

zhtmike: and clean the outdated comment, too..

verl/experimental/agent_loop/diffusion_agent_loop.py
wuxibin89: Deep copy tokenizer is pretty heavy, and the copy count is proportional to `batch_size_per_worker * n`,  severely hurts the performance of asyncio thread.
```
import copy, time

tokenizer = AutoTokenizer.from_pretrained("Qwen/Qwen3.5-27B")
t_start = time.time()
t = copy.deepcopy(tokenizer)
t_end = time.time()
print(f"time elapsed: {t_end-t_start:.2f}s") # 0.75s
```

wuxibin89: transformers>5 always use fast tokenizer, but we didn't encounter the `RuntimeError: Already borrowed` problem in Qwen3.5 testing with transformers=5.3.0
https://github.com/huggingface/transformers/blob/v5.3.0/src/transformers/models/auto/tokenization_auto.py#L625-L626

zhtmike: > transformers>5 always use fast tokenizer, but we didn't encounter the `RuntimeError: Already borrowed` problem in Qwen3.5 testing with transformers=5.3.0 https://github.com/huggingface/transformers/blob/v5.3.0/src/transformers/models/auto/tokenization_auto.py#L625-L626

we encounter this error when using `self.tokenizer(max_length=xxx, padding=True, truncation=True)`, which relates to issue in the https://github.com/huggingface/tokenizers/issues/537. That is why we originally use a slow tokenizer in the early implementation. Perhaps this usage is rare in LLM side?

btw, in the vllm, they are using a retry loop to address this issue in the early implementation, but now change to deepcopy too https://github.com/vllm-project/vllm/pull/36557/changes, just FYI

wuxibin89: If use fast tokenizer, should we tokenize in current thread instead of in separate thread?
https://github.com/verl-project/verl/blob/main/verl/experimental/agent_loop/agent_loop.py#L347

AndyZhou952: Hi, sorry for the confusion. We are working on a version to mimic the change on agent loop, will ping you once it is ready.

AndyZhou952:
Hi @wuxibin89, the PR is ready. To clarify:
- our local testing env is transformers 4.57.6 (and thus no Rust involvement with Rust), while the CI appears to use a version>=5.0.0 and the tokenizer backend with default to Rust, which leads to this CI failure.
- As of now, we have not identified a better solution than using `copy.deepcopy(tokenizer)`, and we acknowledge this performance drawback (w/ a TODO to fix later). For better performance, the user can opt for `transformers<5.0.0` and use a slow tokenizer. We only changed in `__init__` in `DiffusionAgentLoop`, and no interference with the `AgentLoop`.
- Miscellaneous/minor changes since the last review: (1) add back log probs related to UT,  (2) `examples/flowgrpo_trainer` folder organization.

Let me know if it works. Thanks

cc: @zhtmike @SamitHuang

verl/experimental/agent_loop/diffusion_agent_loop.py
zhtmike: > If use fast tokenizer, should we tokenize in current thread instead of in separate thread? https://github.com/verl-project/verl/blob/main/verl/experimental/agent_loop/agent_loop.py#L347

I think for fast-tokenizer, `set_truncation_and_padding` is not thread-safe, as mentioned in the issue discussion. So even when tokenizing in the same thread, it may still cause an `Already borrowed` error. Not sure.

AndyZhou952:
I have checked all eight failure cases and they are unrelated to this PR. The failure cases related to vllm are from the `vllm_async_server`, and we did not touch the code there.

Update 0328: the CI has been rerun and the failure cases are unrelated

verl/experimental/agent_loop/diffusion_agent_loop.py
SamitHuang: How about let tokenizing `apply_chat_template(..., tokenize=True)` run inline in the coroutine, on the asyncio event-loop thread. It should be thread-safe. 
Here is the implementation: https://github.com/AndyZhou952/verl/pull/13/changes
I tested locally for a few runs and no `RuntimeError: Already borrowed` error occurs
@wuxibin89 @zhtmike

zhtmike: Interesting, have you tested with the updated config 'use_fast=True' or transformer>5.0 ? As I remember this PR keeps with the original setting of loading the tokenizer with 'use_fast=False' after recent updates.

SamitHuang: yes, with fast tokenizer and transformers 5.3.0

zhtmike: Then the only problem is that tokenizing in the same thread will block the async loop of sequence generation, might need check if there is any the inference performance drop in e2e.

SamitHuang: can you help compare the e2e performance in the next trainer PR?

zhtmike: I can do a quick verification on pr #5297, perhaps on Monday.

zhtmike: A quick exp shows that w/wo the same event loop, the generation time is roughly same. So this modification is fine.

wuxibin89 (approved):

PR #5741, opened 2026-03-25
[fully_async] fix: Unify gpu-npu fully-async and one-step-off CI.

Shangwei-Li:
### What does this PR do?

Fix fully-async and one-step-off CI OOM and unify NPU-GPU CI.

### Checklist Before Starting

### Test

Tested with 45 steps on L20 * 4.
image

### API and Usage Example

```python
```

### Design & Code Changes

> [!IMPORTANT]

### DIFF
--- a/tests/special_e2e/run_one_step_off_policy.sh
+++ b/tests/special_e2e/run_one_step_off_policy.sh
@@ -166,6 +166,7 @@ elif [ "${ACTOR_STRATEGY}" == "megatron" ]; then
         "${common_params[@]}" \
         actor_rollout_ref.actor.strategy=megatron \
         critic.strategy=megatron \
+        actor_rollout_ref.model.use_remove_padding=True \
         actor_rollout_ref.actor.ppo_micro_batch_size_per_gpu=2 \
         actor_rollout_ref.ref.log_prob_micro_batch_size_per_gpu=4 \
         actor_rollout_ref.rollout.log_prob_micro_batch_size_per_gpu=4 \
--- a/verl/workers/rollout/vllm_rollout/bucketed_weight_transfer.py
+++ b/verl/workers/rollout/vllm_rollout/bucketed_weight_transfer.py
@@ -27,7 +27,7 @@ import torch
 import zmq
 from torch.multiprocessing.reductions import reduce_tensor
 
-from verl.utils.device import get_device_id, get_device_name, get_torch_device
+from verl.utils.device import get_device_id, get_device_name, get_torch_device, set_expandable_segments
 
 logger = logging.getLogger(__file__)
 logger.setLevel(os.getenv("VERL_LOGGING_LEVEL", "INFO"))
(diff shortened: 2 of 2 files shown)


verl/utils/device.py
```diff
@@ -307,7 +307,7 @@ def is_support_ipc() -> bool:
     """
     # If CUDA is available, it's a GPU device
     if is_cuda_available:
-        return True
+        return False
```
ji-huazhong: Why is this modification necessary?

Shangwei-Li: It is temporary, for CI test purpose. 
It seems that some GPU machines do not support IPC and expandable memory at the same time, so I set it as `False` to see if expandable memory solves the OOM problem. It will be removed before ready for merging.

verl/utils/device.py
```diff
@@ -307,12 +307,30 @@ def is_support_ipc() -> bool:
     """
     # If CUDA is available, it's a GPU device
     if is_cuda_available:
-        return True
+        # Check expandable memory global config
+        expandable_memory_enabled = "expandable_segments:True" in os.environ.get("PYTORCH_CUDA_ALLOC_CONF", "")
```
ArronHZG: L20 not supporting CUDA IPC and the need to add `expandable_segments: True` in the script are two separate issues.

For instance, in multi-GPU training, the environment variables set in the script do not take effect on every Ray node.

Should the judgment here be based **solely on hardware**?

verl/utils/device.py
```diff
@@ -307,12 +307,30 @@ def is_support_ipc() -> bool:
...
+        device_capability = get_device_capability()
+        # It is not suppported to enable expandable meory and IPC on old devices like L20.
+        if device_capability <= (8, 9) and expandable_memory_enabled:
```
ArronHZG: Is it sufficient to only check the device capability here?

wuxibin89: We manually set_expandable_segments to False/True before and after cuda ipc: 
https://github.com/verl-project/verl/blob/main/verl/workers/engine_workers.py#L666 
https://github.com/verl-project/verl/blob/main/verl/workers/engine_workers.py#L713

Disable cuda ipc with compute capability <= 8.9 affects user with A100.

verl/utils/device.py
Shangwei-Li: They are not quite separate. CUDA offers different IPC routes in cases where expandable memory is on or off according to [this doc](https://docs.nvidia.com/cuda/cuda-programming-guide/04-special-topics/inter-process-communication.html). During our experiment, L20 reported an error when they were enabled at the same time. When expandable memory is off IPC works fine on L20.

verl/utils/device.py
Shangwei-Li: The function `set_expandable_segments` only works for training engine, but we need it in rollout for stability, since checkpoint engine process will not recollect memory from the other process when it needs it. If we wants to assume that IPC is supported, maybe we should disable it only for L20?

wuxibin89: `is_support_ipc` used both in colocated and disaggregated mode
1. In colocated mode, engine_workers.py manually disable expandable before cuda ipc, so it's safe for capability <= 8.9.
2. In disaggregated mode, `is_support_ipc` is used in [CheckpointEngineWorker](https://github.com/verl-project/verl/blob/main/verl/checkpoint_engine/base.py#L244), so maybe we just need to disable expandable in CheckpointEngineWorker.

wuxibin89: BTW. The unit test use [TrainingWorkerTest](https://github.com/verl-project/verl/blob/main/tests/checkpoint_engine/test_utils.py#L31) instead of CheckpointEngineWorker.

Shangwei-Li:
Close this as CI error not appear locally.

PR #5549, opened 2026-03-10
[perf, training_utils] fix: Run gc regularly in trainer/sft_trainer.py to avoid spikes in step time.

sheilaliuxl:
### What does this PR do?

Run gc regularly in trainer/sft_trainer.py to avoid spikes in step time.
- **NOTE**: ~~Current PR is on top of https://github.com/verl-project/verl/pull/5579~~

### Checklist Before Starting

- [x] Search for similar PRs. Paste at least one query link here: https://github.com/verl-project/verl/pulls?q=is%3Apr+is%3Aopen+gc
- [x] Format the PR title as `[{modules}] {type}: {description}` (This will be checked by the CI)
  - `{modules}` include `fsdp`, `megatron`, `veomni`, `sglang`, `vllm`, `rollout`, `trainer`, `ci`, `training_utils`, `recipe`, `hardware`, `deployment`, `ray`, `worker`, `single_controller`, `misc`, `perf`, `model`, `algo`, `env`, `tool`, `ckpt`, `doc`, `data`, `cfg`, `reward`, `fully_async`, `one_step_off`
  - If this PR involves multiple modules, separate them with `,` like `[megatron, fsdp, doc]`
  - `{type}` is in `feat`, `fix`, `refactor`, `chore`, `test`
  - If this PR breaks any API (CLI arguments, config, function signature, etc.), add `[BREAKING]` to the beginning of the title.
  - Example: `[BREAKING][fsdp, megatron] feat: dynamic batching`

### Test

**I: Step Time**
It decreases training step time consistently with `gc_freq=1` (Dropping first `10` steps and last `3` steps):
- All jobs run on the same node `ip-10-4-145-27`, to reduce inconsistency across different nodes

1. `Median` time: `0.8652 ==> 0.7247 (-16.2%)`
2. `Avg` time: `0.8828 ==> 0.7324 (-17.0%)`
3. `Std` time also decreases `0.0696 ==> 0.0362 (-48.0%)` or improvement of `1.92x`
     * Important to compare across experiments whether step time reduction/ MFU increase is **significant**

**II: MFU**

`MFU` increases slightly on the `worker` as well, from `gc_freq = -1 to 1`:
1. `avg` MFU increases `0.2221 ==> 0.2297 (+3.4%)`: `2026-03-13 17:34:52,918 [tb.py:264] INFO - [243/256] : (avg (std), med) = (    0.2297 [+    0.0076] (    0.0020),     0.2301)  # train/mfu`
[...]

PR #5753, opened 2026-03-25
[fsdp, perf] fix: skip redundant to(cuda) and gc.collect in train_mode when offload is disabled

evmanz:
### What does this PR do?

`BaseEngineCtx._context_switch()` unconditionally calls `engine.to("cuda")` on every training step entry, even when `param_offload` and `optimizer_offload` are both disabled and the model/optimizer are already on GPU. This triggers a redundant FSDP module tree traversal via `load_fsdp_model_to_gpu` (~150ms) and a full `gc.collect()` (~295ms), wasting ~450ms per step.

For Qwen3-30B-A3B SFT training on 8×H100 with FSDP2 and LoRA, this represents ~13% of a 3.5s training step (w/ gsm8k dataset).

This PR adds an early return in `_context_switch()` when entering GPU mode (`device != "cpu"`) and both offloads are disabled, skipping the unnecessary `to()` call entirely.

### Test

Validated via nsys profiling on Qwen3-30B-A3B SFT with FSDP2 + LoRA (8×H100, `param_offload=False`, `optimizer_offload=False`):

| Metric | Before | After |
|--------|--------|-------|
| `train_mode_enter` | ~450ms | <1ms |
| Step time | ~3.5s | ~3.05s |
| Throughput improvement | — | ~13% |

The fix is a no-op when either offload is enabled — the existing code path runs unchanged.

### Design & Code Changes

Single change in `verl/workers/engine/base.py`, `BaseEngineCtx._context_switch()`:

```python
def _context_switch(self, device):
    if self.disable_auto_offload:
        return
    # Skip when entering GPU mode with no offload configured —
    # model and optimizer are already on device.
    if device != "cpu":
        if not self.engine.is_param_offload_enabled and not self.engine.is_optimizer_offload_enabled:
            return
    # ... existing logic unchanged ...
```

The guard checks two conditions:
1. We're entering GPU mode (not exiting to CPU)
2. Neither param nor optimizer offload is enabled

When both hold, the model and optimizer have been on GPU since `initialize()` and were never moved to CPU (the `__exit__` path already respects offload flags). The `to()` call, its FSDP tree traversal, and `gc.collect()` are pure overhead.

### DIFF
--- a/verl/workers/engine/base.py
+++ b/verl/workers/engine/base.py
@@ -242,6 +242,9 @@ class BaseEngineCtx:
     def _context_switch(self, device):
         if self.disable_auto_offload:
             return
+        if device != "cpu":
+            if not self.engine.is_param_offload_enabled and not self.engine.is_optimizer_offload_enabled:
+                return
         if self.mode == "eval":
             self.engine.to(device=device, model=self.engine.is_param_offload_enabled, optimizer=False, grad=False)
         elif self.mode == "train":


evmanz:
Adding @eric-haibin-lin and @vermouth1992 for review. Thank you!

sheilaliuxl:
Hi Evgeny @evmanz ,
Did you try https://github.com/verl-project/verl/pull/5549, to see wether it can (partially) fix the same issue.
- You can find this PR in a search in this repo: https://github.com/verl-project/verl/pulls?q=is%3Apr+is%3Aopen+gc

evmanz:
> Hi Evgeny @evmanz , Did you try #5549, to see wether it can (partially) fix the same issue.
> 
> * You can find this PR in a search in this repo: https://github.com/verl-project/verl/pulls?q=is%3Apr+is%3Aopen+gc

I think these are different. The fix in this PR is to eliminate garbage collection call that is not needed anyway -- the data was allocated on the GPU, so nothing to cleanup on the CPU side at all.

evmanz:
Hi @wuxibin89! Can you take a look, please?

vermouth1992 (approved):

PR #5777, opened 2026-03-27
[data] feat: use key intersection in collate_variable_batch to handle optional keys

khazic:
collate_variable_batch used set().union() to collect keys across all samples, then accessed batch[0][key] assuming every sample has every key. This causes KeyError when optional keys like multi_modal_inputs exist in only some samples. Changed to set.intersection() so only keys present in ALL samples are collated.

### DIFF
--- a/verl/utils/dataset/dataset_utils.py
+++ b/verl/utils/dataset/dataset_utils.py
@@ -62,7 +62,7 @@ class SFTTensorCollator:
         final_batch = {}
 
         union_keys = set().union(*(d.keys() for d in batch))
-        tensor_keys = set.intersection(*(set(d.keys()) for d in batch)) if batch else set()
+        tensor_keys = set(batch[0].keys()).intersection(*(d.keys() for d in batch[1:])) if batch else set()
         missing_keys = union_keys - tensor_keys
         if missing_keys:
             raise ValueError(
--- a/verl/utils/dataset/dataset_utils.py
+++ b/verl/utils/dataset/dataset_utils.py
@@ -61,7 +61,10 @@ class SFTTensorCollator:
 
         final_batch = {}
 
-        tensor_keys = set().union(*(d.keys() for d in batch))
+        # Use intersection of keys so only keys present in ALL samples are collated.
+        # This avoids KeyError / batch size mismatch for optional keys like
+        # multi_modal_inputs, which is conditionally added by the dataset.
+        tensor_keys = set(batch[0].keys()).intersection(*(d.keys() for d in batch[1:])) if batch else set()
 
         # Handle tensor values by creating a NestedTensor.
         for key in tensor_keys:
--- a/verl/utils/dataset/dataset_utils.py
+++ b/verl/utils/dataset/dataset_utils.py
@@ -13,10 +13,13 @@
 # limitations under the License.
 
 
+import logging
 from enum import Enum
 
 import torch
 from tensordict.tensorclass import NonTensorData
+
+logger = logging.getLogger(__name__)
 
 
 class DatasetPadMode(str, Enum):
@@ -62,14 +65,12 @@ class SFTTensorCollator:
         final_batch = {}
 
         union_keys = set().union(*(d.keys() for d in batch))
-        tensor_keys = set(batch[0].keys()).intersection(*(d.keys() for d in batch[1:])) if batch else set()
+        tensor_keys = set.intersection(*[set(d.keys()) for d in batch]) if batch else set()
         missing_keys = union_keys - tensor_keys
         if missing_keys:
-            raise ValueError(
-                f"Keys {missing_keys} are missing in some samples within the batch. "
-                f"Mixing samples with and without optional keys (e.g., multi_modal_inputs) "
-                f"in the same batch is not supported. Use a modal-aware batch sampler to "
-                f"ensure all samples in a batch have the same keys."
+            logger.warning(
+                f"Keys {missing_keys} are missing in some samples and will be skipped during collation. "
+                f"For mixed text+image datasets, ensure text-only samples include an empty multi_modal_inputs dict."
             )
 
         # Handle tensor values by creating a NestedTensor.
--- a/verl/utils/dataset/dataset_utils.py
+++ b/verl/utils/dataset/dataset_utils.py
@@ -61,10 +61,16 @@ class SFTTensorCollator:
 
         final_batch = {}
 
-        # Use intersection of keys so only keys present in ALL samples are collated.
-        # This avoids KeyError / batch size mismatch for optional keys like
-        # multi_modal_inputs, which is conditionally added by the dataset.
-        tensor_keys = set(batch[0].keys()).intersection(*(d.keys() for d in batch[1:])) if batch else set()
+        union_keys = set().union(*(d.keys() for d in batch))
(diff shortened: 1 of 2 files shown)


khazic:
**Update: replaced intersection-only fix with fail-fast validation**

The previous version of this PR used `set.intersection()` to silently drop keys not present in all samples:

```python
tensor_keys = set(batch[0].keys()).intersection(*(d.keys() for d in batch[1:]))
```

This avoids the `KeyError` from the original `union` approach, but introduces a subtle data loss issue: if a batch mixes vision samples (with `multi_modal_inputs`) and text-only samples (without it), the key is silently discarded and the downstream model receives no vision data — with no error or warning.

The updated fix now explicitly detects key inconsistency and raises a `ValueError`:

```python
union_keys = set().union(*(d.keys() for d in batch))
tensor_keys = set.intersection(*(set(d.keys()) for d in batch)) if batch else set()
missing_keys = union_keys - tensor_keys
if missing_keys:
    raise ValueError(
        f"Keys {missing_keys} are missing in some samples within the batch. "
        f"Mixing samples with and without optional keys (e.g., multi_modal_inputs) "
        f"in the same batch is not supported. Use a modal-aware batch sampler to "
        f"ensure all samples in a batch have the same keys."
    )
```

This way:
- **Homogeneous batches** (all text-only or all vision): no change in behavior
- **Mixed batches**: fails immediately with a clear message, rather than silently corrupting the batch

The right fix for mixed-modal datasets is to use a modal-aware batch sampler that groups samples by modality, not to paper over the inconsistency at collation time.

khazic:
**Further update: fix mm_token_type_ids leaking into multi_modal_inputs**

After more investigation, the previous fix (checking `self.processor is not None`) turned out to be insufficient. For Qwen3.5, `hf_processor` always returns a non-None processor even for text-only models (since Qwen registers its tokenizer as a processor via `AutoProcessor`). So the guard never actually filtered anything.

The root cause is that Qwen3.5's tokenizer returns `mm_token_type_ids` in some code paths (e.g. normal path for user messages) but not others (e.g. fallback path for assistant messages that require a dummy user prefix). This causes inconsistent keys across samples in the same batch.

`mm_token_type_ids` is a tokenizer artifact that marks text vs vision tokens — for text-only data it is all-zeros and has no training value. The fix is simply to drop it unconditionally:

```python
# mm_token_type_ids is a tokenizer artifact that marks text vs vision tokens.
# For text-only data it is all-zeros and not needed for training.
multi_modal_inputs.pop("mm_token_type_ids", None)
```

After this fix, `multi_modal_inputs` only contains actual vision keys (e.g. `pixel_values`, `image_grid_thw`), so the existing `if len(multi_modal_inputs) > 0` check correctly distinguishes text-only from multimodal samples.

Verified working on text-only Qwen3.5-35B-A3B:


```
step:7 - train/loss:0.3739721179008484 - train/grad_norm:8.055917817955223 - train/lr:1.75e-06 - train/mfu:0.0 - train/global_tokens:3174
```

khazic:
**Update: approach changed to support mixed text+image batches natively**

After further analysis, the `raise ValueError` approach (current PR) requires users to ensure all samples in a batch have the same modality, which is inconvenient for datasets that naturally mix text-only and image samples.

**Root cause analysis:**

The real issue is that `__getitem__` conditionally includes `multi_modal_inputs` only when non-empty, causing key inconsistency in the collator. Meanwhile, `position_ids` is already consistent — text-only samples with a Qwen2-VL-based processor (including Qwen2.5-VL and Qwen3-VL, which use `Qwen2VLImageProcessorFast`) already produce `(4, seq_len)` shape because the processor check runs unconditionally.

**New fix (latest commit):**

Always include `multi_modal_inputs` when `self.processor is not None`, even if it is an empty dict `{}` for text-only samples:

```python
# Before
if len(multi_modal_inputs) > 0:
    res["multi_modal_inputs"] = multi_modal_inputs

# After
if self.processor is not None:
    res["multi_modal_inputs"] = multi_modal_inputs  # {} for text-only samples
```

**Why downstream is already compatible:**
- `extract_multi_modal_inputs()` already skips empty dicts gracefully
- `_get_input_embeds()` already handles `pixel_values=None` via a built-in dummy image gradient path (the `handle mixed text-image data` comment)
- `ray_trainer.py` already guards with `if "image_grid_thw" not in multi_modal_input: continue`

This allows a single dataset with mixed text-only and image samples to train without requiring a modal-aware batch sampler.

khazic:
## Latest Update

This revision replaces the previous approach with a cleaner, dataset-aware strategy for handling `multi_modal_inputs` consistency.

### What changed

**`multiturn_sft_dataset.py`**
- Added `is_visual_dataset` flag at init time, detected by checking whether the dataframe contains image/video columns.
- Removed the incorrect unconditional `mm_token_type_ids.pop()` — this was silently dropping a field that image samples may actually need.
- Replaced `if self.processor is not None:` with `if multi_modal_inputs or self.is_visual_dataset:` to make inclusion semantics explicit.

**`dataset_utils.py`**
- Downgraded the `ValueError` on key mismatch to a `logger.warning`, keeping the collator robust instead of hard-crashing on edge cases.

---

### Behavior walkthrough — Qwen3.5-VL as example

Qwen3.5-VL uses `Qwen2VLImageProcessorFast` (confirmed via `preprocessor_config.json`), so the `get_rope_index` path triggers correctly and 4D position_ids are generated as expected.

#### Case 1: Mixed dataset (some samples have images, some don't)

`is_visual_dataset = True` (the `images` column exists in the dataframe)

**Image sample:**
```
_build_messages         → injects {"type": "image", ...} into message content
_process_single_message → processor generates pixel_values, image_grid_thw, ...
multi_modal_inputs      = {"pixel_values": ..., "image_grid_thw": ...}  # non-empty
position_ids            → get_rope_index with real image_grid_thw → 4D (4, seq_len)
condition               → multi_modal_inputs is truthy → include in res
```

**Text-only sample (`images` column exists but is `[]`):**
```
_build_messages         → images=[], no <image> placeholders
                        → message["content"] = [{"type": "text", "text": ...}]
_process_single_message → processor handles text-only input
multi_modal_inputs      = {}  # empty
position_ids            → get_rope_index with image_grid_thw=None → 4D text-only
condition               → is_visual_dataset=True → include empty {} in res
```

At batch level: every sample carries the `multi_modal_inputs` key → collator sees no key mismatch.
`extract_multi_modal_inputs()` skips empty dicts, so only real `pixel_values` reach the model.

#### Case 2: Pure text dataset (no image/video columns)

`is_visual_dataset = False`

```
multi_modal_inputs = {}  # empty, no visual artifacts
condition          → {} or False → False → multi_modal_inputs NOT added to res
```

Identical behavior to a text-only model. No dummy visual tensors, no extra keys, no wasted compute.

---

### Why `is_visual_dataset` instead of `processor is not None`

The previous `if self.processor is not None:` check over-included: a pure-text dataset configured with a VLM processor would incorrectly attach an empty `multi_modal_inputs` to every sample. The `is_visual_dataset` flag ties the decision to actual data content rather than processor configuration.

khazic:
Closing this as stale. It has been open for about five months without review activity, and the branch has drifted far enough from `main` that the change would need to be revalidated from scratch rather than simply rebased.

No hard feelings about the queue, I know it is long. If a maintainer wants this after all, I am happy to reopen it and resubmit against current `main`.

PR #5782, opened 2026-03-27
[megatron, ckpt] fix: handle None param_data in get_megatron_module_device when use_distributed_optimizer=False

khazic:
### What does this PR do?

Fixes a crash in `get_megatron_module_device` during checkpoint saving when `use_distributed_optimizer=False`.

When `use_distributed_optimizer=False`, Megatron's `ParamAndGradBuffer.param_data` is intentionally set to `None` (no flat param buffer is allocated since the distributed optimizer does not need it). The original code unconditionally called `buffer.param_data.storage()`, causing:

```
AttributeError: 'NoneType' object has no attribute 'storage'
```

This crash happens at the very start of `save_checkpoint`, before any data is written, so the checkpoint is never saved.

**Trigger condition:** `use_distributed_optimizer=False` + checkpoint saving. This combination arises in certain expert/pipeline parallel configs where the distributed optimizer is incompatible with the NCCL communication groups, e.g. PP=4, EP=2, TP=2, ETP=2 with `vanilla_mbridge=True`.

### Checklist Before Starting

- [x] Search for similar PRs. No existing fix found for this specific issue.

### Test

Verified on Qwen3.5-35B-A3B with the following config:

```
engine.tensor_model_parallel_size=2
engine.pipeline_model_parallel_size=4
engine.expert_model_parallel_size=2
engine.expert_tensor_parallel_size=2
engine.use_distributed_optimizer=False
engine.use_mbridge=True
engine.vanilla_mbridge=True
```

Checkpoint saves successfully at the configured `save_freq` step.

### Design & Code Changes

`get_megatron_module_device` now handles `param_data is None` by falling back to checking the device of the actual module parameters:

```python
# Before
buffer = model_chunk.buffers[0]
if buffer.param_data.storage().size() == 0:  # crash when param_data is None
    return 'cpu'

# After
buffer = model_chunk.buffers[0]
if buffer.param_data is None:
    # use_distributed_optimizer=False: no flat param buffer, check module params directly
    try:
        return next(model_chunk.module.parameters()).device.type
    except StopIteration:
        return 'cpu'
if buffer.param_data.storage().size() == 0:
    return 'cpu'
```

- [ ] Apply pre-commit checks.
- [ ] Add / Update the documentation.
- [ ] Add unit or end-to-end test(s). Not feasible for this fix as it requires a specific multi-node parallel config to reproduce.

### DIFF
--- a/verl/utils/megatron_utils.py
+++ b/verl/utils/megatron_utils.py
@@ -1347,6 +1347,12 @@ def get_megatron_module_device(models: list[Any]) -> str:
             return "cpu"
 
     buffer = model_chunk.buffers[0]
+    if buffer.param_data is None:
+        # use_distributed_optimizer=False: no flat param buffer, check module params directly
+        try:
+            return next(model_chunk.module.parameters()).device.type
+        except StopIteration:
+            return "cpu"
     if buffer.param_data.storage().size() == 0:
         return "cpu"
     else:


khazic:
## Background

### Why `use_distributed_optimizer=False`?

When training Qwen3.5-35B-A3B (a MoE model) with a high degree of parallelism (PP=4, EP=2, TP=2, ETP=2) using `vanilla_mbridge=True`, setting `use_distributed_optimizer=True` causes an NCCL error at startup because the distributed optimizer requires additional process group communication that is incompatible with this specific parallel configuration. The only workaround is to set `use_distributed_optimizer=False`.

### Why does `param_data` become `None`?

This is intentional Megatron behavior. In `ParamAndGradBuffer.__init__`:
- When `use_distributed_optimizer=True`: a flat contiguous `param_data` tensor is allocated to support optimizer state sharding across DP ranks
- When `use_distributed_optimizer=False`: `param_data` is explicitly set to `None` — no flat buffer is needed since each GPU holds the full optimizer state in the original parameter tensors

### Why does the crash only happen at checkpoint saving?

Other places in `megatron_utils.py` that access `buffer.param_data` (e.g. the CPU offload/load paths) are only reachable when `_is_offload_param=True`, which itself requires `use_distributed_optimizer=True`. So those paths are never triggered in this config.

`get_megatron_module_device` is the only code path that accesses `param_data` without a `None` guard and can be reached with `use_distributed_optimizer=False`. It is called unconditionally at the start of every `save_checkpoint`, which is why the crash surfaces exactly at the first checkpoint save step.

### What the fix does

When `param_data is None`, fall back to checking the device of the actual module parameters directly (the same fallback already used when `model_chunk.buffers` is empty). This correctly returns `"cuda"` and allows checkpoint saving to proceed normally without triggering any unnecessary `load_megatron_model_to_gpu` calls.

ETOgaosion (approved):

PR #5797, opened 2026-03-30
[training_utils] fix: prevent cross-sequence loss mask contamination in NO_PADDING SFT

dubin555:
### What does this PR do?

Fixes a cross-sequence mask contamination bug in `sft_loss()` when using NO_PADDING mode (the default).

**Bug:** `torch.roll(loss_mask_flatten, shifts=-1, dims=0)` on the flattened `.values()` of a nested tensor performs a **global** circular shift across all sequences. This wraps the last element of each sequence into the first position of the next sequence, causing cross-sequence mask contamination at every sequence boundary.

**Example:**
```
Seq A mask: [0, 1, 1, 1]   Seq B mask: [1, 0, 1]
Flattened:  [0, 1, 1, 1, 1, 0, 1]
After roll: [1, 1, 1, 1, 0, 1, 0]  ← A last pos gets B[0]=1 (wrong!)
Correct:    [1, 1, 1, 0, 0, 1, 0]  ← Each seq last pos should be 0
```

The last position of each sequence has no valid next-token prediction (the rolled `input_ids` places a cross-sequence token there), so including it in the loss produces incorrect gradient signals.

**Impact:**
- Affects **all** multi-sequence NO_PADDING SFT batches (default mode)
- Loss magnitude changes significantly (test case: 5.6 buggy vs 2.6 correct)
- Model learns cross-sequence token transitions at boundaries

**Fix:** After the global roll, zero out the last position of each sequence using nested tensor offsets — O(1) additional work:
```python
offsets = loss_mask.offsets()
loss_mask_flatten = torch.roll(loss_mask_flatten, shifts=-1, dims=0)
loss_mask_flatten[offsets[1:] - 1] = 0
```

**Note:** `DocumentSimhashDeduplicator.process()` uses only local variables (`graph`, `hash2ids`, etc.), which are naturally scoped — contrast with the minhash variant that stores state on `self`.

### Checklist Before Starting

- [x] Search for similar PRs: https://github.com/verl-project/verl/pulls?q=is%3Apr+sft+loss+mask+roll
- [x] Format the PR title as `[module] type: description`

### Tests

7 unit tests covering:
1. Single-sequence (no contamination, baseline)
2. Two-sequence cross-contamination proof
3. Numerical loss difference (5.6 vs 2.6)
4. Three-sequence cascading contamination
5. Per-sequence last position zeroed
6. Vectorized fix matches reference loop fix
7. Vectorized fix differs from buggy output

```bash
python3 -m pytest tests/test_sft_loss_roll_bug.py -v  # 7 passed
```

### DIFF
--- a/verl/workers/utils/losses.py
+++ b/verl/workers/utils/losses.py
@@ -39,8 +39,20 @@ def sft_loss(config: ActorConfig, model_output, data: TensorDict, dp_group=None)
         log_prob_flatten = log_prob.values()
         loss_mask_flatten = loss_mask.values()
 
-        # left-shift the loss mask by one token to align with log_prob
+        # Left-shift the loss mask by one token to align with log_prob.
+        # log_prob[i] predicts token at position i+1, so loss_mask[i] should
+        # indicate whether we care about the prediction at position i+1.
+        # Use per-sequence shift: a global torch.roll on the flattened buffer
+        # wraps the last element of one sequence into the next, causing
+        # cross-sequence mask contamination.  Instead, shift globally and then
+        # fix up the boundary positions using the nested tensor offsets.
+        offsets = loss_mask.offsets()
         loss_mask_flatten = torch.roll(loss_mask_flatten, shifts=-1, dims=0)
+        # Zero out the last position of each sequence: there is no valid
+        # next-token prediction there (the model output at that position
+        # corresponds to a cross-sequence label due to the global roll in
+        # input_ids_rmpad_rolled).
+        loss_mask_flatten[offsets[1:] - 1] = 0
 
         # NOTE: loss is averaged over all tokens in the batch across all data parallel groups,
         # For FSDP backend, the loss is directly used for backward; while for Megatron backend,
--- a/tests/test_sft_loss_roll_bug.py
+++ b/tests/test_sft_loss_roll_bug.py
@@ -1,0 +1,246 @@
+# Copyright 2025 Bytedance Ltd. and/or its affiliates
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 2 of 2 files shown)


verl/workers/utils/losses.py
```diff
@@ -40,8 +39,20 @@ def sft_loss(config: ActorConfig, model_output, data: TensorDict, dp_group=None)
...
-        # left-shift the loss mask by one token to align with log_prob
+        # Left-shift the loss mask by one token to align with log_prob.
+        # log_prob[i] predicts token at position i+1, so loss_mask[i] should
+        # indicate whether we care about the prediction at position i+1.
+        # Use per-sequence shift: a global torch.roll on the flattened buffer
+        # wraps the last element of one sequence into the next, causing
+        # cross-sequence mask contamination.  Instead, shift globally and then
+        # fix up the boundary positions using the nested tensor offsets.
+        offsets = loss_mask.offsets()
         loss_mask_flatten = torch.roll(loss_mask_flatten, shifts=-1, dims=0)
```
wuxibin89, line 50: It's safe to left shift here since the last token loss mask will be first token loss mask in next sequence, which should be always 0.
```python
# before left shift: [seq0, seq1, seq2, seq3]
loss_mask_flatten = [[0, 0, 1, 1, 1], [0, 0, 0, 1, 1], [0, 1], [0, 0, 0, 1]]
# after left shift
loss_mask_flatten = [[0, 1, 1, 1, 0], [0, 0, 1, 1, 0], [1, 0], [0, 0, 1, 0]]
```

wuxibin89:
```
Seq A mask: [0, 1, 1, 1]   Seq B mask: [1, 0, 1]
Flattened:  [0, 1, 1, 1, 1, 0, 1]
After roll: [1, 1, 1, 1, 0, 1, 0]  ← A last pos gets B[0]=1 (wrong!)
Correct:    [1, 1, 1, 0, 0, 1, 0]  ← Each seq last pos should be 0
```
Seq loss should always start with 0, since the first token should be prompt.

PR #5801, opened 2026-03-30
[vllm, fsdp] fix: apply FSDP buffer updates during rollout weight sync

chenshui223:
## Summary

Fix FSDP-to-rollout weight sync for models whose registered buffers are updated during training.

The existing rollout update path applies parameter updates through `load_weights(...)`, but registered buffers from the FSDP state dict can be missed. This may cause rollout/training mismatch for models that rely on mutable buffers such as `e_score_correction_bias`.

## Changes

- split incoming rollout weight updates into parameter updates and buffer updates
- keep parameter loading on the existing `load_weights(...)` path
- apply registered buffer updates explicitly in-place on the rollout model
- add CPU tests covering buffer routing and rollout-side update behavior

## Validation

```bash
PYTHONPATH=/hy-tmp/verl-submit pytest -q tests/workers/rollout/test_vllm_weight_update_utils_on_cpu.py
```

### DIFF
--- a/tests/workers/rollout/test_vllm_weight_update_utils_on_cpu.py
+++ b/tests/workers/rollout/test_vllm_weight_update_utils_on_cpu.py
@@ -1,0 +1,244 @@
+# Copyright 2025 Bytedance Ltd. and/or its affiliates
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


chenshui223:
I re-checked the current status on this PR and could not find evidence that the remaining failed jobs are directly caused by the buffer-sync change. The failures are still spread across broader model / reward / megatron / setup workflows rather than this CPU rollout buffer path.

I still addressed the bot's local performance suggestion in a minimal way on this branch: `split_buffer_updates(...)` now builds the model buffer map once and `apply_buffer_updates(...)` reuses it, with the CPU rollout buffer tests updated accordingly.

Validation run on this branch:
`python -m pytest -q tests/workers/rollout/test_vllm_weight_update_utils_on_cpu.py`

Result: `4 passed`

ETOgaosion:
@chenshui223 Thanks for fixing this, please help rebase main and fix conflicts, I'll help to run CIs

ETOgaosion (approved):

PR #5829, opened 2026-03-31
[training_utils] fix: mrope position_ids preprocess bug fix

ZLiao097:
### What does this PR do?

This is a coner case.

During model debugging, an error :
`RuntimeError: split_with_sizes expects split_sizes to sum exactly to 1091 (input tensor's size at dimension 1), but got split_sizes=[4]`

This error is triggered under the following specific conditions:

* Using mrope (VL model)
* Input in bsnd format (non-pack mode)
* Engine worker
* Small global batch size (GBS) — each dp is assigned only 1 or 2 mini-batches per step

In non-pack mode, verl utilizes `torch.nested_tensor`. The standard shape of `position_idx` is `(b, s)`, whereas for mrope, it is `(b, 4, s)`. When initializing a `nested_tensor`, the framework automatically designates a single distinct dimension as the variable dimension (only one variable dimension is permitted). 

When `b = 1`, there are no varying dimensions across samples, so the second dimension (`[4]`) is automatically set as the variable dimension (i.e., dim=1). The same issue arises when `b` is small and the sequence lengths (`s`) of multiple samples are identical.

Additionally, in the `maybe_fix_3d_position_ids` method, the internal attribute `_ragged_idx` — which specifies the split dimension — is hardcoded to 2 (corresponding to variable dimension dim=1) to accommodate mrope scenarios. This hardcoding conflicts with the variable dimension automatically assigned during `nested_tensor` initialization, leading to an error when executing `unbind()`.

In summary, hardcoding `_ragged_idx` is a makeshift solution. It forcibly switches the variable dimension to dim=2 without modifying the actual variable dimension determined during `nested_tensor` creation. While this works for most batch configurations, it triggers the aforementioned error under the specific conditions outlined.

To resolve this issue, this PR reconstructs the shape of `position_ids` for all mrope scenarios using `torch.nested.nested_tensor_from_jagged`, replacing the previous approach of only modifying `_ragged_idx`.

### Checklist Before Starting

### Test

You can validate this issue by follow codes:
```python
import torch

def nt_solution(nt):
    batch_num, _, seq_len = nt.shape
    if isinstance(seq_len, int):      
        offset = torch.arange(0, (batch_num + 1) * seq_len, seq_len, dtype=torch.int32)
        values = nt.values().view(batch_num, 4, -1).transpose(0, 1).flatten(1)
        print(f">>>>> offset: {offset}")
        print(f">>>>> values.shape: {values.shape}")
        nt = torch.nested.nested_tensor_from_jagged(values, offsets=offset, jagged_dim=2)
        return nt

def reshape_nt():
    a = torch.randn(4, 17)
    b = torch.randn(4, 17)

    nt = torch.nested.as_nested_tensor([a, b], layout=torch.jagged)

    print(f">>>>> orig nt: {nt}")

    nt = nt_solution(nt)

    print(f">>>>> processed nt: {nt}")
    nt._ragged_idx = 2

    ut = nt.unbind()
    print(f">>>>> len(ut): {len(ut)}")

if __name__ == "__main__":
    reshape_nt()

# >>>>> orig nt: NestedTensor(size=(2, j1, 17), offsets=tensor([0, 4, 8]), contiguous=True)
# >>>>> offset: tensor([ 0, 17, 34], dtype=torch.int32)
# >>>>> values.shape: torch.Size([4, 34])
# >>>>> processed nt: NestedTensor(size=(2, 4, j2), offsets=tensor([ 0, 17, 34], dtype=torch.int32), contiguous=True)
# >>>>> len(ut): 2
```

### API and Usage Example

```python
```

### Design & Code Changes

> [!IMPORTANT]

### DIFF
--- a/verl/utils/tensordict_utils.py
+++ b/verl/utils/tensordict_utils.py
@@ -890,6 +890,6 @@ def maybe_fix_3d_position_ids(data: TensorDict):
                 # offsets.shape: [wrong](0, 4, 8, ...)           -> [correct](0, seq_len, 2 * seq_len, ...)
                 # values.shape : [wrong](4 * batch_num, seq_len) -> [correct](4, batch_num * seq_len)
                 offsets = torch.arange(0, (batch_num + 1) * seq_len, seq_len, dtype=offsets.dtype)
-                values = values.view(batch_num, 4, -1).transpose(0, 1).flatten(1)
+                values = values.contiguous().view(batch_num, 4, -1).transpose(0, 1).flatten(1)
 
             data["position_ids"] = torch.nested.nested_tensor_from_jagged(values=values, offsets=offsets, jagged_dim=2)
--- a/verl/utils/tensordict_utils.py
+++ b/verl/utils/tensordict_utils.py
@@ -889,7 +889,7 @@ def maybe_fix_3d_position_ids(data: TensorDict):
                 # when same seq_len for all batch_num, there is wrong offsets and values
                 # offsets.shape: [wrong](0, 4, 8, ...)           -> [correct](0, seq_len, 2 * seq_len, ...)
                 # values.shape : [wrong](4 * batch_num, seq_len) -> [correct](4, batch_num * seq_len)
-                offsets = torch.arange(0, (batch_num + 1) * seq_len, seq_len, dtype=torch.int32)
+                offsets = torch.arange(0, (batch_num + 1) * seq_len, seq_len, dtype=offsets.dtype)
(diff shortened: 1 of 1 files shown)


vermouth1992:
Those hacks are not necessary when this is merged. https://github.com/pytorch/tensordict/pull/1675

auto-yun:
Thanks for fixing this.

The current reconstruction path seems specific to `(B, 4, L)`. It may be better to infer the middle dimension dynamically, e.g. from `values.shape[0] // batch_num`, instead of hardcoding `4`, so the logic can also support future models with other position-component counts, such as `(B, 3, L)`.

Issue #5790, opened 2026-03-28
[RFC] Agent Abstractions and Trajectory Gateway for VERL


zackcxb:
# [RFC] Agent Abstractions and Trajectory Gateway for VERL

## Summary

This RFC proposes two new abstractions for VERL's agent-based reinforcement learning pipeline:

1. **AgentFramework** — an abstract base class for agent lifecycle management and reward computation, replacing the current tight coupling between `AgentLoopManager` and specific agent implementations.
1. **AgentGateway** — a Gateway subsystem owned by the serving layer (`LLMServerManager`) that intercepts agent LLM calls via the OpenAI Chat Completions API, performs canonical tokenization, and assembles token-level trajectory data with strict token-truth guarantees.

Together, they enable any OpenAI-compatible agent system to be integrated into VERL's training loop without modifications to agent code, while producing continuous multi-turn trajectories with loss masks directly consumable by VERL's training engine.

## Motivation

VERL's current agent integration (`AgentLoopManager` + `AgentLoopBase`) tightly couples three concerns: LLM infrastructure management, agent lifecycle, and trajectory collection. This creates friction when integrating new agent types:

- Each new agent framework requires dedicated adapter code inside the agent loop.
- Trajectory collection logic is embedded in the agent loop itself, making it non-reusable.
- Only coroutine-based agents are natively supported; subprocess and remote agents require ad-hoc integration (e.g., SWE-Agent's custom `ModelProxy`).

Community contributions such as AWS AgentCore (PR #4216) and Aliyun Remote Agent (Issue #5737) further demonstrate the need for a pluggable agent abstraction that cleanly separates these concerns.

This RFC addresses these issues by:

1. **Defining** `AgentFramework` as a thin standard interface for agent-based rollout. The common contract is `generate_sequences(prompts: DataProto) -> DataProto`; internal execution structure, reward computation flow, and batching strategy remain implementation-specific.
1. **Extracting trajectory collection into** `AgentGateway`, a serving-side subsystem that works with any Framework implementation. The Gateway handles tokenization, prefix consistency, and trajectory assembly — concerns that are orthogonal to how agents are launched or managed.
1. **Extracting infrastructure management** (LLM server initialization, load balancing) out of the framework layer, so that `AgentFramework` remains focused on trainer-facing generation semantics rather than serving ownership.

## Design Overview

[...]

PR #5833, opened 2026-03-31
[worker, vllm] feat: expose nemo gym token id patch

cmunley1:
## What does this PR do?

Allows NeMo Gym to patch vLLM's chat serving layer to preserve token IDs across multi-turn rollouts, preventing retokenization mismatches during RL training. May be replaced by (#5790) in future. Tied to verl-project/verl-recipe#80. Minimal change with no impact on non-NeMo Gym workloads.

Exposes `apply_nemo_gym_server_patch()` on `vLLMHttpServer`, called by NemoGymAgentLoopManager at startup via ray remote to patch `OpenAIServingChat._preprocess_chat` and `OpenAIServingTokenization._preprocess_chat` at runtime inside the vLLM server Ray actor process. This seems  needed because it executes in the server process and the patch cant go across process boundary. Patch implements the same logic as NeMo RL. Tested on vLLM 0.17.0 (`verlai/verl:vllm017.latest`).


wuxibin89:
> nemo gym agents and environments work through HTTP OpenAI Responses API. Multi-turn token IDs are re-tokenized each turn via chat template.

We're working on Agent Gateway https://github.com/verl-project/verl/issues/5790 to make each agent framework seamlessly integrated into verl for agentic rl training. 

For now, you can submit this PR to https://github.com/verl-project/verl-recipe

cmunley1:
Moved here https://github.com/verl-project/verl-recipe/pull/80, I think the small patch only called by nemo gym is still needed here at the moment, AgentGateway may change things. 

Thanks!

verl/workers/rollout/vllm_rollout/vllm_async_server.py
```diff
@@ -170,6 +170,11 @@ def get_server_address(self):
         assert self._server_port is not None, "http server is not launched, port is None"
         return self._server_address, self._server_port
 
+    def apply_nemo_gym_server_patch(self):
```
wuxibin89, line 173: Could you move this to nemo-gym recipe? You can inherit `vLLMHttpServer` and override vLLMReplica with a new `NemoGymVLLMReplica`, then register it  https://github.com/verl-project/verl/blob/main/verl/workers/rollout/replica.py#L391

PR #5877, opened 2026-04-03
[fsdp, actor] fix: correct loss normalization in legacy dp_actor

### LINKED ISSUE #5625: Incorrect loss_scale_factor for token-mean mode in gradient accumulation in legacy FSDP implementation

alex27594:
Fixes #5625

## Summary

This PR fixes loss normalization in the legacy FSDP actor path (`verl/workers/actor/dp_actor.py`) for `token-mean` mode under gradient accumulation.

The bug described in #5625 comes from a mismatch between:

- `agg_loss(token-mean)`, which expects normalization by the **global valid-token count**
- and the legacy FSDP path, where `global_batch_info` is not populated and the backward path still applies a sample-based scaling factor

As a result, the accumulated micro-batch loss/gradient can diverge from the correct full-batch result when token counts vary across micro-batches.

### 1) Populate `global_batch_info` in legacy FSDP
For each PPO mini-batch, compute the global valid-token count from `response_mask`, all-reduce it across ranks, and populate:

- `config.global_batch_info["batch_num_tokens"]`
- `config.global_batch_info["dp_size"]`

### 2) Use the effective FSDP gradient divisor
Instead of assuming a divisor implicitly, this PR derives the effective FSDP gradient divisor from the wrapped module and uses that value for `dp_size`.

This makes the normalization logic match the actual gradient reduction semantics more carefully and avoids silently relying on assumptions that may not hold across legacy FSDP configurations.

### 3) Stop applying the old backward rescaling when global token normalization is active
Once `batch_num_tokens` is populated and `agg_loss(token-mean)` already normalizes against the full mini-batch/global token count, the backward path should not apply the old micro-batch/sample-based rescaling again.

### 4) Apply the same aggregation semantics consistently
Entropy / KL aggregation in this path now also uses the populated `global_batch_info`, so those auxiliary terms follow the same normalization semantics as the policy loss.

## Why this is correct

With `token-mean`, the intended result is to match the full-batch loss:

\[
L_{\text{full}} = \frac{\sum_i \text{sum}_i}{\sum_i \text{tokens}_i}
\]

In the broken legacy FSDP path, each micro-batch effectively uses its local token count, and then a separate sample-based scaling is applied during accumulation. That is exactly the mismatch described in #5625.

This PR fixes that by:
- computing the global token denominator once per mini-batch,
- using the effective FSDP gradient divisor,
- and removing the extra backward rescaling when global token normalization is already active.

## Tests

This PR adds focused regression coverage:

### CPU regression tests
- `tests/workers/actor/test_agg_loss_math.py`
  - validates `agg_loss(token-mean)` math directly
  - checks that splitting into micro-batches matches the full-batch result only when global normalization inputs are correct
  - includes simple policy-like, entropy-like, and KL-like toy cases
  - includes a simple uniform-mask baseline and variable-length-mask cases

### Manual distributed integration test
- `tests/workers/actor/test_legacy_fsdp_loss_normalization.py`
  - manual/distributed integration test intended to be run with `torchrun`
  - verifies gradient accumulation invariance across micro-batch splits
  - checks that `global_batch_info` is populated correctly
  - uses the variable-length-mask case, which is the most relevant scenario for this bug

### Simple smoke script
- `tests/workers/actor/run_legacy_fsdp_loss_norm_multinode.sh`
  - simple distributed/manual smoke script for running the integration test

## Scope

This PR is intentionally narrow:
- legacy FSDP only
- no Megatron changes
- no algorithm changes
- no broader refactors

### DIFF
--- a/verl/workers/actor/dp_actor.py
+++ b/verl/workers/actor/dp_actor.py
@@ -633,8 +633,12 @@ class DataParallelPPOActor(BasePPOActor):
                 _batch_num_tokens = _response_mask.sum().to(get_device_id())
                 if torch.distributed.is_initialized():
                     torch.distributed.all_reduce(_batch_num_tokens, op=torch.distributed.ReduceOp.SUM)
+                # Clamp to 1 to avoid division-by-zero when the entire
+                # mini-batch is masked.  The numerator (masked loss sum) is
+                # also zero in that case, so gradients remain zero.
+                batch_num_tokens = max(int(_batch_num_tokens.item()), 1)
                 self.config.global_batch_info["dp_size"] = self._fsdp_gradient_divisor
-                self.config.global_batch_info["batch_num_tokens"] = _batch_num_tokens.item()
+                self.config.global_batch_info["batch_num_tokens"] = batch_num_tokens
 
                 for micro_batch in micro_batches:
                     micro_batch = micro_batch.to(get_device_id())
@@ -715,7 +719,11 @@ class DataParallelPPOActor(BasePPOActor):
                             loss_agg_mode=loss_agg_mode,
                             **self.config.global_batch_info,
                         )
-                        micro_batch_metrics["actor/entropy"] = entropy_agg.detach().item()
+                        # Log local micro-batch mean so the metric stays
+                        # independent of dp_size / gradient accumulation.
+                        micro_batch_metrics["actor/entropy"] = (
+                            verl_F.masked_mean(entropy, response_mask).detach().item()
+                        )
                         if entropy_coeff != 0:
                             policy_loss -= entropy_agg * entropy_coeff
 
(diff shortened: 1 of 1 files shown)


alex27594:
There is clear overlap between this PR and #5641, which is also addressing `#5625`.

I am keeping this PR open for now because it contains two things that may still be useful even if maintainers choose to consolidate on top of #5641:

* focused regression tests for the legacy FSDP loss-normalization bug,
* and a more explicit handling of the effective FSDP gradient divisor in the legacy actor path.

If maintainers prefer a single consolidated fix, I am happy to converge:

* either by moving the tests / rationale from this PR into #5641,
* or by adapting whichever parts are considered most useful.

I am completely fine with the final merge path being:
**core fix logic from #5641 + tests / part of the divisor rationale from #5877**.

wuxibin89:
Thanks for your contribution, legacy worker is deprecated and will be removed in v0.8.0

alex27594:
> Thanks for your contribution, legacy worker is deprecated and will be removed in v0.8.0

It seems that issue https://github.com/verl-project/verl/issues/5625 is opened

Issue #5772, opened 2026-03-27
qwen3.5 rope_utils IndexError: too many indices for tensor of dimension 2


wade0604:
### System Info

package info:
vllm 0.17.0
megatron-core 0.16.0
mbridge  0.15.1
torch 2.10.0+cu128
transformers 5.2.0

### Information

- [x] The official example scripts
- [ ] My own modified scripts

### Tasks

- [ ] An officially supported task in the `examples` folder (such as GLUE/SQuAD, ...)
- [x] My own task or dataset (give details below)

### Reproduction

run_scripts:

```
export VLLM_USE_V1=0
export CUDA_DEVICE_MAX_CONNECTIONS=1
export VLLM_ALLREDUCE_USE_SYMM_MEM=0
export WORKING_DIR="${PWD}"
export MASTER_PORT="4444"


HF_MODEL_PATH=Qwen/Qwen3.5-35B-A3B
train_path=/xxx
test_path=/xxx


TP=${TP:-2}
PP=${PP:-1}
CP=${CP:-1}
EP=${EP:-8}
ETP=${ETP:-1}
GEN_TP=${GEN_TP:-8}

ALL_OFFLOAD=${ALL_OFFLOAD:-True}

rollout_name="vllm"
project_name=xxx'
exp_name='qwen3_5_35ba3b_megatron'
adv_estimator=grpo


########################### Parameter Arrays ###########################

DATA=(
    data.train_files=${train_path}
    data.val_files=${test_path}
    data.train_batch_size=32
    data.max_prompt_length=2048
    data.max_response_length=2048
    data.truncation='error'
    data.filter_overlong_prompts=True
)

MODEL=(
    actor_rollout_ref.model.path=${HF_MODEL_PATH}
    actor_rollout_ref.model.trust_remote_code=True
    actor_rollout_ref.model.use_remove_padding=False
)

ACTOR=(
    actor_rollout_ref.actor.optim.lr=1e-6
    actor_rollout_ref.actor.ppo_mini_batch_size=32
    actor_rollout_ref.actor.ppo_micro_batch_size_per_gpu=1
    actor_rollout_ref.actor.ppo_max_token_len_per_gpu=4096
    actor_rollout_ref.actor.use_dynamic_bsz=False
    actor_rollout_ref.actor.use_kl_loss=True
    actor_rollout_ref.actor.kl_loss_coef=0.01
    actor_rollout_ref.actor.kl_loss_type=low_var_kl
[...]

Issue #5554, opened 2026-03-11
[BUG] RuntimeError: position_ids dimension mismatch when mixing video samples with different durations in same batch


D26FORWARD:
### System Info

- verl version: 0.7.0.dev
- PyTorch: 2.6+
- Model: Qwen3-VL / Qwen2.5-VL
- Task: Video RLHF training with different frame counts

### Problem Description

When training video data with different durations (frame counts) in the same batch, a dimension mismatch error occurs:

```
RuntimeError: The size of tensor a (3) must match the size of tensor b (8) at non-singleton dimension 0
```

This happens specifically when video samples with different frame counts (e.g., 60, 180, 300 frames) are mixed in the same training batch.

### Root Cause

Video samples with different durations have different `position_ids` tensor shapes. When mixed in the same batch, the tensors cannot be properly stacked, leading to dimension mismatches during the forward pass.

Example:
```
Batch: [60-frame video, 60-frame video, 180-frame video, 180-frame video]
position_ids: [tensor(60), tensor(60), tensor(180), tensor(180)]
→ Dimension mismatch during stacking!
```

### Reproduction

1. Prepare a dataset with video samples of different durations (e.g., 60, 180, 300 frames)
2. Run GRPO/PPO training with standard DataLoader settings
3. Error occurs when different-duration samples are batched together

### Expected Behavior

Training should proceed without position_ids dimension errors, either by:
1. Ensuring same-duration samples in each batch
2. Providing proper padding/handling for mixed-duration batches
3. Supporting duration-grouped batching strategy

### Current Workaround

Using a custom `BatchSampler` that groups samples by duration and skips incomplete batches.

### Information

- [ ] The official example scripts
- [x] My own modified scripts

### Tasks

- [ ] An officially supported task in the `examples` folder
- [x] My own task or dataset (video RLHF with multiple durations)

D26FORWARD:
## Update: Root Cause Identified and Fixed

After deeper investigation, the original root cause analysis in this issue was **incorrect**. The problem is **not** caused by mixing different-duration samples in the same batch. Here is the corrected analysis and fix.

### True Root Cause

The error occurs when **all samples in a micro_batch have identical sequence lengths** (e.g., all VLM samples producing the same token count after hitting the `max_token` limit).

[...]

PR #5886, opened 2026-04-06
fix: flatten multi-component position_ids to 1D for nested tensor compatibility

yifannnwu:
## Summary

Fixes a crash when training models with multi-dimensional RoPE position_ids (e.g. Qwen3.5's 4-component `(text, height, width, temporal)`, Qwen2-VL) on the FSDP `engine_workers.py` path (`use_legacy_worker_impl=disable`).

Related: #5772 (same class of bug on the Megatron/mbridge path)

## Problem

`left_right_2_no_padding()` in `padding.py` creates per-sample position_ids of shape `(num_components, valid_len)` and passes them directly to `torch.nested.as_nested_tensor(..., layout=torch.jagged)`.

Jagged layout treats **dim 0** as the ragged dimension, so it interprets `num_components` (e.g. 4) as ragged instead of `seq_len`. The resulting nested tensor has shape `[batch, *(ragged=4), seq_len]` instead of the intended `[batch, *(ragged=seq_len), num_components]`.

Downstream in `transformer_impl.py`, `to_padded_tensor()` produces incorrect shapes → crash in `apply_rotary_pos_emb`.

Transposing to `(valid_len, num_components)` before nesting doesn't work either, because 3D jagged nested tensors have broken `unbind()` and `to_padded_tensor()` in PyTorch (see [pytorch/pytorch#153238](https://github.com/pytorch/pytorch/issues/153238)).

## Fix

Flatten multi-component position_ids to 1D `(seq_len * num_components,)` before creating the nested tensor, keeping it purely 2D jagged. Store `num_pos_components` as non-tensor metadata. Reshape back to `(num_components, batch, seq_len)` in `prepare_model_inputs`.

### Changes

**`verl/workers/utils/padding.py`** — `left_right_2_no_padding()`:
- Flatten `(num_components, valid_len)` → `(valid_len * num_components,)` via `.T.contiguous().flatten()`
- Store `num_pos_components` in non-tensor metadata via `tu.assign_non_tensor_data`

**`verl/workers/engine/fsdp/transformer_impl.py`** — `prepare_model_inputs()`:
- **NO_PADDING path**: `.view(total_nnz, num_components).T.unsqueeze(1)` → `(num_components, 1, total_nnz)`
- **Padded path**: `to_padded_tensor(..., output_size=(batch, max_seq_len * num_components)).view(batch, max_seq_len, num_components).permute(2, 0, 1)` → `(num_components, batch, max_seq_len)`

## Affected models

Any model with multi-dimensional RoPE on the FSDP `engine_workers.py` path:
- Qwen3.5 (4-component: text + 3D spatial)
- Qwen2-VL (3-component mRoPE)
- Other VLMs with multi-component position_ids

## Test plan

- [x] Qwen3.5-0.8B, FSDP2, `use_legacy_worker_impl=disable`: 3 GRPO training steps + validation pass. Previously crashed with RoPE shape mismatch.
- [ ] Qwen2-VL (not tested, same position_ids format — should work)

### DIFF
--- a/verl/workers/engine/fsdp/transformer_impl.py
+++ b/verl/workers/engine/fsdp/transformer_impl.py
@@ -904,8 +904,13 @@ class FSDPEngineWithLMHead(FSDPEngine):
 
             if pad_mode == DatasetPadMode.NO_PADDING:
                 input_ids_rmpad = input_ids.values().unsqueeze(0)  # (1, total_nnz)
-                if position_ids.dim() == 3:
-                    position_ids_rmpad = position_ids.values().unsqueeze(1)  # (4, 1, total_nnz)
+                num_pos_components = tu.get_non_tensor_data(data=micro_batch, key="num_pos_components", default=0)
+                if num_pos_components > 0:
+                    # position_ids stored as flattened 1D nested tensor: (total_nnz * num_components,)
+                    # reshape to (num_components, 1, total_nnz)
+                    flat_pos = position_ids.values()  # (total_nnz * num_components,)
+                    total_nnz = flat_pos.shape[0] // num_pos_components
+                    position_ids_rmpad = flat_pos.view(total_nnz, num_pos_components).T.unsqueeze(1)
                 else:
                     position_ids_rmpad = position_ids.values().unsqueeze(0)  # (1, total_nnz)
             else:
@@ -976,10 +981,13 @@ class FSDPEngineWithLMHead(FSDPEngine):
                     input_ids, padding=pad_token_id, output_size=(batch_size, max_seq_len)
                 )
 
-                if position_ids.dim() == 3:
+                num_pos_components = tu.get_non_tensor_data(data=micro_batch, key="num_pos_components", default=0)
+                if num_pos_components > 0:
+                    # position_ids stored as flattened 1D nested: each sample has (seq_len * num_components,)
+                    # pad to (batch, max_seq_len * num_components), then reshape to (num_components, batch, max_seq_len)
                     position_ids = torch.nested.to_padded_tensor(
-                        position_ids, padding=0, output_size=(batch_size, 4, max_seq_len)
-                    ).transpose(0, 1)  # (4, batch_size, max_seq_len)
+                        position_ids, padding=0, output_size=(batch_size, max_seq_len * num_pos_components)
+                    ).view(batch_size, max_seq_len, num_pos_components).permute(2, 0, 1)
                 else:
                     position_ids = torch.nested.to_padded_tensor(
                         position_ids, padding=0, output_size=(batch_size, max_seq_len)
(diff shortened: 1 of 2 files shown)


cben484:
Confirming this fix at scale, and flagging two follow-ups.

We hit the exact bug this PR fixes on a **Qwen3.5-MoE (122B) GRPO run on Ascend NPU (910B, 8 cards, Ulysses SP=8, FSDP2, `use_remove_padding=true`, NO_PADDING mode)**. The `as_nested_tensor(list_of_(rope_dim, seg_len), layout=torch.jagged)` construction silently drops tokens for 3D mRoPE `position_ids`, so `position_ids` and `input_ids` end up with different `total_nnz`. In our case it stayed latent until the Ulysses all-to-all + varlen path, where `cu_seq_lens_q` (derived from `position_ids`) no longer matched the gathered `input_ids` length and the GatedDeltaNet varlen kernel collapsed its output (`z` rows != `core_attn_out` rows). Applying the same flatten-to-1D approach as this PR resolved it — **108 consecutive training steps and counting (steps 6→113, still running), with healthy metrics (grad_norm ~10, reward increasing); previously an intermittent crash within ~10 steps.**

So this fix is correct and needed beyond the small-model case in the test plan.

Two things that would make it complete:

1. **Other engine consumers need the same un-flatten.** `left_right_2_no_padding` is the single shared construction point (called from `ray_trainer.py`), but `position_ids.values()` is also consumed with the old `dim() == 3` assumption in:
   - `verl/workers/engine/torchtitan/transformer_impl.py`
   - `verl/workers/engine/automodel/transformer_impl.py`

   After flattening in `padding.py`, `position_ids.dim()` becomes 2, so those two branches silently take the wrong path for mRoPE. They need the same `(rope_dim, 1, total_nnz)` reconstruction as the fsdp path (a robust `values_len == total_nnz` check works without needing the `num_pos_components` metadata).

2. **This PR is currently blocked from merging: the CLA is unsigned and it has been idle ~3 months.** Since the diff itself is `MERGEABLE`, the only thing standing between this fix and `main` is the CLA + someone driving it. Given how many people are hitting this exact bug (see #5554, #6212, and the NPU reports there), it would be a shame for it to stall.

   @yifannnwu — do you plan to sign the CLA and finish this? If so, great, I'll help review/test and can send you the torchtitan/automodel additions to fold in. If you'd rather not, I'm happy to **open a CLA-signed PR that supersedes this one, with you credited via `Co-authored-by: yifannnwu`**, extending the fix to all four engine consumers and including the 108-step NPU validation above. Either way you keep authorship credit — I just want the fix to land.

   If there's no response in ~1–2 weeks I'll ping a maintainer to confirm the take-over is welcome before opening anything.

(Note: AI assistance was used in diagnosing this on our side; I've reviewed and reproduced the behavior end-to-end.)

PR #5903, opened 2026-04-07
[chat_template]: handle system message in apply_chat_template fall…

khazic:
The existing fallback for models like Qwen3.5 that require at least one user message unconditionally prepends a dummy user message before the real messages:

    apply_template([dummy_user] + messages)

This works for user/assistant messages, but fails when the single message being processed is a system message, because Qwen3.5's chat template enforces two constraints:
  1. At least one user message must be present.
  2. System message must appear first.

The fallback satisfies (1) but violates (2), raising:
    TemplateError: System message must be at the beginning.

Reproduce: run multiturn SFT with a dataset where some conversations have a system message and some do not. verl tokenizes each message individually via _process_single_message; when a system message is processed alone, the fallback inserts dummy_user before it, breaking the template constraint.

Fix: detect whether the messages list contains a system role. If so, append the dummy user *after* the real messages instead, and strip the dummy suffix from the output using the length-difference trick (len(two_users) - len(one_user)) to recover the pure system tokens. The original prefix-stripping path is preserved for all non-system cases.

### What does this PR do?

### Checklist Before Starting

### Test

### API and Usage Example

```python
```

### Design & Code Changes

> [!IMPORTANT]

### DIFF
--- a/tests/utils/test_chat_template_on_cpu.py
+++ b/tests/utils/test_chat_template_on_cpu.py
@@ -1,0 +1,112 @@
+# Copyright 2025 Bytedance Ltd. and/or its affiliates
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import pytest
+
+from verl.utils.tokenizer.chat_template import apply_chat_template
+
+
+def _text_of(message):
+    content = message["content"]
+    if isinstance(content, list):
+        return "".join(part.get("text", "") for part in content)
+    return content
+
+
+class StrictTokenizer:
+    """Chat template with the Qwen3.5 constraints that trigger the fallback path.
+
+    * at least one user message is required,
+    * a system message must come first,
+    * only the *last* assistant message keeps its ``reasoning_content``,
+      like Qwen3 / DeepSeek-R1 templates do.
+    """
+
+    def render(self, messages, add_generation_prompt=False, tools=None):
+        rendered = ""
+        if tools:
+            rendered += f"<|tools|>{','.join(tools)}"
+        for i, message in enumerate(messages):
+            role = message["role"]
+            rendered += f"<|{role}|>"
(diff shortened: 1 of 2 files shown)


khazic:
Note: this bug is triggered whenever apply_chat_template is called with a 
single system message (which happens for every row that has a system message 
in the dataset), not just for mixed datasets. A pure-system-message dataset 
would hit the same error on every sample.

khazic:
@wuxibin89  Could you please review this fix? 

This PR addresses a bug in the `apply_chat_template` fallback path: when messages contain **only a system message** (no user message), the original code prepends `dummy_user + messages`, which violates Qwen3.5's constraint that the system message must come first. This causes a `TemplateError`.

The fix detects `has_system` and appends the dummy user *after* messages instead, then strips the suffix using the difference trick.

nico-zck:
@khazic found the same problem, but simply appending dummy_user_message to the end of messages will lead to missing the reasoning_content of the last assistant message.

khazic:
@nico-zck good catch, you are right. Templates that render `reasoning_content` for the last assistant message only (Qwen3, DeepSeek-R1) drop it as soon as the dummy user is appended after it, so the tokenized output silently lost the reasoning span. It shows up exactly where this fallback is used, since verl tokenizes messages one at a time and a `[system, assistant]` slice has no user message.

Fixed in 1402ceb: the dummy user is now inserted right after the leading system block instead of at either end, and its token span is cut back out of the middle of the output via the same difference trick (`head_len = len(head + dummy) - user_len`). That keeps the system message first and the last assistant message last, so the reasoning content survives.

Two things fell out of it: `add_generation_prompt` no longer needs to be re-attached by hand (the dummy sits in the middle, so the generation prompt already lands at the tail), and the probe renders now pass `tools` so a tool schema rendered into the system block is counted in the head length.

Added `tests/utils/test_chat_template.py` (the module had no tests before) with a tokenizer stub that reproduces the Qwen3.5 constraints plus the last-assistant-only reasoning behaviour. The trailing-assistant test fails on the previous commit and passes now; all 9 pass.

khazic:
Correction to my previous comment: the branch was four months behind, and `main` has since moved this module to `verl/utils/tokenizer/chat_template.py`, where the fallback is still the original prepend-only version. The PR was conflicting (modify/delete) as a result, so I rebased it onto current `main` and reapplied the fix at the new path (`cb5edd5e`, force-pushed).

The fix itself is unchanged: the dummy user is inserted right after the leading system block and its token span is cut back out of the middle, so the system message stays first and the last assistant message stays last.

The test moved to `tests/utils/test_chat_template_on_cpu.py` to match the CPU test naming used in `tests/utils/`. Against unpatched `main` it fails 7 of 9 (the original `TemplateError` on a system-only slice, plus the dropped `reasoning_content`); with the fix all 9 pass. `ruff format` and `ruff check` are clean.

Issue #3243, opened 2025-08-27
[BUG] Can't override actor_rollout_ref.actor.fsdp_config.model_dtype to bfloat16 via cmdline input


jw0109park:
Hi,

I tried to resolve the following error 
```
Flash Attention 2 only supports torch.float16 and torch.bfloat16 dtypes, but the current dype in Qwen3ForCausalLM is torch.float32. You should run training or inference using Automatic Mixed-Precision via the with torch.autocast(device_type='torch_device'): decorator, or load the model with the torch_dtype argument. Example: model = AutoModel.from_pretrained("openai/whisper-tiny", attn_implementation="flash_attention_2", torch_dtype=torch.float16)
```
by passing the following cmdline options:
```
    +actor_rollout_ref.actor.fsdp_config.model_dtype=bfloat16
```
I did not set actor_rollout_ref.actor.fsdp_config.mixed_precision, since the following code block in verl/workers/fsdp_workers.py already sets the model parameters to bf16 by default:
```
        mixed_precision_config = fsdp_config.get("mixed_precision", None)
        if mixed_precision_config is not None:
            param_dtype = PrecisionType.to_dtype(mixed_precision_config.get("param_dtype", "bf16"))
            reduce_dtype = PrecisionType.to_dtype(mixed_precision_config.get("reduce_dtype", "fp32"))
            buffer_dtype = PrecisionType.to_dtype(mixed_precision_config.get("buffer_dtype", "fp32"))
        else:
            param_dtype = torch.bfloat16
            reduce_dtype = torch.float32
            buffer_dtype = torch.float32
```

However, adding the cmdline option resulted in the following error:
```
Could not override 'actor_rollout_ref.actor.fsdp_config.model_dtype'.
To append to your config use +actor_rollout_ref.actor.fsdp_config.model_dtype=bfloat16
Key 'model_dtype' is not in struct
    full_key: actor_rollout_ref.actor.fsdp_config.model_dtype
    object_type=dict
```

IIt seems that arguments not included in the predefined yaml (e.g. verl/trainer/config/ppo_trainer.yaml) cannot be overridden from the command line.

When I modified FSDPEngineConfig in verl/workers/engine.py as below, the flash attention error disappeared:
```
    -model_dtype: str = "fp32"
    +model_dtype: str = "bf16"
```
Since actor_rollout_ref.actor.fsdp_config is FSDPEngineConfig, it looks like the initialization of this config is currently disconnected from cmdline input.

[...]

PR #5933, opened 2026-04-09
fix: use configured param_dtype in forward_step autocast instead of hardcoded bf16

mango766:
## Summary

Fixes #5932.

`forward_step()` hardcodes `torch.autocast(dtype=torch.bfloat16)`, ignoring the `mixed_precision.param_dtype` already resolved from engine config earlier in the same class. This means setting `param_dtype=fp16` or disabling mixed precision has no effect on the forward execution path.

## Changes

1. Store `param_dtype` as `self._autocast_dtype` after resolving mixed precision config (line ~341)
2. Use `self._autocast_dtype` in `forward_step()`'s autocast context instead of hardcoded `torch.bfloat16`

1 file changed: `verl/workers/engine/fsdp/transformer_impl.py`

## Not a duplicate

Checked open PRs — no existing PR addresses this. #3243 is about CLI override not being accepted; this is about forward execution ignoring the resolved value.

## AI assistance disclosure

AI assistance was used. The fix is minimal (2 lines) and straightforward.

### DIFF
--- a/verl/workers/engine/fsdp/transformer_impl.py
+++ b/verl/workers/engine/fsdp/transformer_impl.py
@@ -339,6 +339,7 @@ class FSDPEngine(BaseEngine):
             buffer_dtype = torch.float32
 
         mixed_precision = MixedPrecision(param_dtype=param_dtype, reduce_dtype=reduce_dtype, buffer_dtype=buffer_dtype)
+        self._autocast_dtype = param_dtype
 
         auto_wrap_policy = get_fsdp_wrap_policy(
             module=module,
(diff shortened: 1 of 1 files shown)


wuxibin89:
If use `torch.float16`, the grad should be scaled by ShardedGradScaler, https://github.com/verl-project/verl/pull/4036

shivam2199:
Hey — thanks for putting this up originally. I opened #6150 as a superset that tries to address the two outstanding pieces of feedback on this PR:

- fp16 needs `ShardedGradScaler` (per @wuxibin89) — #6150 rejects fp16 at engine build time with a clear `NotImplementedError` rather than silently producing bad gradients. Scaler integration can be a follow-up.
- `nullcontext()` for fp32 (per @Kirrito-k423 on the issue) — included.

I left `get_per_tensor_param`'s `torch.bfloat16` cast alone since that's a deliberate bandwidth downcast for weight-sync, not an autocast inconsistency; feels like a separate design discussion.

Happy to defer to this PR if you'd prefer to pick it back up — otherwise #6150 is ready for review.

PR #5969, opened 2026-04-11
[data, trainer] fix: batch padding for multi-trajectory

ZhentaoFan:
### Background
Inside the current tq_trainer, AgentLoopWorkerTQ already supports the multi-trajectory feature. However, during actual training, sample (trajectory)-level padding is still required for each batch so that the number of samples is divisible by both `dp_size` and `mini_batch_size`; otherwise, an error will be thrown. This PR fixes the bug and addresses the following considerations:

#### Upsampling:
1. **LCM alignment**: `% dp_size == 0` and `% mini_batch_size == 0` (and `% critic_mini_batch_size == 0` if training the critic).
2. **Notice**：We should **not** use pad_dataproto_to_divisor() for padding, as it takes first `pad_size` existing samples and would pollute the grpo_adv, gradient, metrics and etc., which causes inconsistency.

#### Padding:
1. Padded samples use **independent UIDs** to avoid interfering with GRPO advantage computation.
3. Padded samples are constructed with the **shortest possible sequence** — one prompt token + one response token — to minimize redundant computation.
4. An `is_padding` flag is added to the tags of padded samples to avoid impacting the **accuracy metrics** such as score, reward, and response length (while performance metrics still include padded samples).
5. Maintains the router_replay shape, position_ids shape and add multi-modal inputs placeholder in VLM case when necessary.

### Related PR
1. https://github.com/verl-project/verl/pull/5636
2. https://github.com/verl-project/verl/pull/5401

#### The three smallest primes — 2, 3, and 5 — are chosen to form the relevant hyperparameters: 
[dp=2, batch_size=45, mini_batch_size=15, rollout.n=8].

20260410_225225_critic_score_mean

### DIFF
--- a/verl/trainer/main_ppo_sync.py
+++ b/verl/trainer/main_ppo_sync.py
@@ -22,7 +22,9 @@ Differs from original PPO trainer in main_ppo.py:
 """
 
 import asyncio
+import copy
 import logging
+import math
 import os
 import threading
 import time
@@ -85,6 +87,7 @@ from verl.utils.device import auto_set_device
 from verl.utils.fs import copy_to_local
 from verl.utils.import_utils import load_class_from_fqn
 from verl.utils.metric import reduce_metrics
+from verl.utils.model import compute_position_id_with_mask
 from verl.utils.py_functional import rename_dict
 from verl.utils.ray_utils import auto_await
 from verl.utils.seqlen_balancing import calculate_workload, get_seqlen_balanced_partitions, log_seqlen_unbalance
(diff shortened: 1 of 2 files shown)


ZhentaoFan:
@eric-haibin-lin @vermouth1992 @tongyx361 @PeterSH6 Ready for review.

verl/trainer/main_ppo_sync.py
```diff
@@ -1009,9 +1157,11 @@ def _balance_batch(self, batch: KVBatchMeta, metrics, logging_prefix="global_seq
             dp_rank_mapping = worker_group._dispatch_info[role]
         dp_size = max(dp_rank_mapping) + 1
 
-        # TODO: up sampling if batch is not divisible by dp_size
-        if len(batch) % dp_size != 0:
-            raise ValueError(f"Batch size {len(batch)} is not divisible by dp_size {dp_size}")
+        # Upsampling the batch with padding sequences
+        batch_multiple = self._get_required_batch_multiple(dp_size)
+        batch = self._upsample_batch_to_divisible_size(batch, batch_multiple)
```
wuxibin89: Please move `_upsample_batch_to_divisible_size` into a separate file under `verl/trainer/ppo`, I think fully async also need it.

wuxibin89 (approved):

PR #5978, opened 2026-04-12
[tool, rollout, cfg] feat: per-sample tool environment routing for ToolAgentLoop

pull-ups:
## What does this PR do?

Add per-sample tool environment routing to `ToolAgentLoop`, allowing each dataset row to use a different tool set during multi-turn rollouts — without requiring any custom subclass or new agent loop.

**Motivation**: In real-world agent training, different scenarios need different tools (e.g., `gsm8k` calculator vs. `sandbox_fusion` code executor vs. `geo3k` geometry tool). Currently, all rows in a batch share the same global `tool_config_path`. This PR enables heterogeneous tool environments via a simple config mapping + dataset-level routing key.

**How it works**:

1. Config defines named tool environments via `multi_turn.tool_envs` mapping:
   ```yaml
   multi_turn:
     tool_envs:
       gsm8k: examples/sglang_multiturn/config/tool_config/gsm8k_tool_config.yaml
       sandbox_fusion: examples/sglang_multiturn/config/tool_config/sandbox_fusion_tool_config.yaml
   ```
2. Dataset rows carry `extra_info.tool_env_name` (e.g., `"gsm8k"`) to select their tool set
3. `ToolAgentLoop.__init__` pre-loads all named envs into a cache at startup
4. `run()` resolves the per-sample tools and passes them through the state machine
5. Falls back to global `tool_config_path` tools when `tool_env_name` is absent or unrecognized

**Security**: Tool config paths are defined only in the server-side rollout config — dataset rows cannot specify arbitrary paths. This avoids the RCE risk of dynamic class import from untrusted input.

## Design & Code Changes

**3 files modified, 41 additions, 4 deletions:**

| File | Changes |
|------|---------|
| `verl/experimental/agent_loop/tool_agent_loop.py` (+33/-4) | Pre-load tool envs in `__init__`, select per-sample tools in `run()`, use them in `_handle_pending_state`, `_handle_generating_state`, `_call_tool` |
| `verl/workers/config/rollout.py` (+3/-0) | Add `tool_envs: Optional[dict[str, str]] = None` to `MultiTurnConfig` |
| `verl/trainer/config/rollout/rollout.yaml` (+9/-0) | Add `tool_envs: null` default with usage documentation |

**Data flow:**
```
Dataset Row (extra_info.tool_env_name: "gsm8k")
  ↓
run() → self._tool_env_cache["gsm8k"] → (tools_dict, schemas_list)
  ↓
agent_data._active_tools / _active_tool_schemas
  ↓
State machine: PENDING → GENERATING → PROCESSING_TOOLS → TERMINATED
(uses per-sample tools at each stage, falls back to global tools)
```

## Test

- `py_compile` passed on all changed Python files
- Backwards compatible: when `tool_envs` is not configured (default `null`), behavior is identical to current main

## Disclosure

This PR was developed with AI assistance (GitHub Copilot). All changes have been reviewed and understood by the human submitter.

### DIFF
--- a/verl/experimental/agent_loop/tool_agent_loop.py
+++ b/verl/experimental/agent_loop/tool_agent_loop.py
@@ -160,6 +160,19 @@ class ToolAgentLoop(AgentLoopBase):
             interaction_kwargs=interaction_kwargs,
         )
 
+        # Per-sample tool selection: filter global tools by extra_info.tool_selection
+        extra_info = kwargs.get("extra_info", {}) or {}
+        tool_selection = extra_info.get("tool_selection")
+        if tool_selection and self.tools:
+            selected = {name: self.tools[name] for name in tool_selection if name in self.tools}
+            agent_data._active_tools = selected
+            agent_data._active_tool_schemas = [
+                t.tool_schema.model_dump(exclude_unset=True, exclude_none=True) for t in selected.values()
+            ]
+        else:
+            agent_data._active_tools = self.tools
+            agent_data._active_tool_schemas = self.tool_schemas
+
         # State machine loop
         state = AgentState.PENDING
         while state != AgentState.TERMINATED:
@@ -202,9 +215,10 @@ class ToolAgentLoop(AgentLoopBase):
 
     async def _handle_pending_state(self, agent_data: AgentData, sampling_params: dict[str, Any]) -> AgentState:
         """Handle the pending state: prepare the prompt and start generation."""
+        schemas = getattr(agent_data, "_active_tool_schemas", self.tool_schemas)
         prompt_ids = await self.apply_chat_template(
             agent_data.messages,
-            tools=self.tool_schemas,
+            tools=schemas,
             images=agent_data.image_data,
             videos=agent_data.video_data,
         )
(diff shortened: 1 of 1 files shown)


wuxibin89:
Another option is 
1. Define all tool schema in `agent_loop_config_path`
2. Add an optional tool selection field (e.g `agent_tools`) in RLHFDataset.

```text
prompt, agent_name, tool_selection
..., single_turn_agent, None
..., tool_agent_loop, sandbox
..., tool_agent_loop, zoom
```

We're going to support register-based tool definition, then we can even eliminate `agent_loop_config_path`
```python
@tool
def sandbox(): ...

@tool
async def zoom(): ... 
```

pull-ups:
@wuxibin89  
That's great idea and future direction! BTW, before such update, I think this PR can support simple way to use instance-wise environment designation so that agent can be trained with varioous environment. It does not modify core algorithm of recent verl code, and just added small option.

verl/experimental/agent_loop/tool_agent_loop.py
```diff
@@ -160,6 +173,15 @@ async def run(self, sampling_params: dict[str, Any], **kwargs) -> AgentLoopOutpu
             interaction_kwargs=interaction_kwargs,
         )
 
+        # Per-sample tool env routing: select tools based on extra_info.tool_env_name
+        extra_info = kwargs.get("extra_info", {}) or {}
+        env_name = extra_info.get("tool_env_name")
```
wuxibin89: Could you just add a `tool_selection` into extra_info and left rollout.yaml unchanged, you can load all tools from `tool_config_path` and select a tool subset by `tool_selection`

wuxibin89: I'm working on the register-based tool definition.

pull-ups: Yes, according to your suggestion and adaptability with future direction, I modified code changes of this PR. If this is not the right way you meant, please let me know!
Thanks for contributing verl project!

wuxibin89 (approved):

Issue #4852, opened 2026-01-09
compute_policy_loss_bypass_mode这个函数里面修改了response_mask，这样会导致实际参与agg_loss的有效token数与engine中收集的batch_num_tokens不一致，loss变小


Schnabel-8:
目前verl/workers/engine/megatron/transformer_impl.py中按照loss_mask收集了batch_num_tokens来进行后续的loss aggregate
https://github.com/volcengine/verl/blob/main/verl/workers/engine/megatron/transformer_impl.py#L476

，但是在开启bypass mode的情况下，计算loss时会根据train infer log probs对response mask进行一些修正https://github.com/volcengine/verl/blob/main/verl/trainer/ppo/core_algos.py#L2065
，修正后的有效token数与训练前收集的batch_num_tokens不一致，这会导致agg loss的时候在token-mean情况下使用的分母有问题，这里有什么解决方法吗？

Kirrito-k423:
Thanks for reporting this. I think this is a valid concern for the bypass-mode paths that enable rejection sampling / trust-region masking.

From current `main`, the engine computes `batch_num_tokens` before the loss function from `data["loss_mask"].sum()` and stores it in the batch. In the PPO loss path, that value is copied into `config.global_batch_info`. Later, `compute_policy_loss_bypass_mode()` calls `compute_rollout_correction_and_rejection_mask()`, uses the returned `modified_response_mask` as `effective_mask`, and passes that mask into the underlying policy loss. `agg_loss(..., loss_agg_mode="token-mean")` then divides the masked loss sum by the precomputed `batch_num_tokens`.

So if `rollout_rs` / veto masking removes tokens or sequences after `batch_num_tokens` has already been collected, the numerator uses the reduced effective mask while the denominator can still reflect the original mask. That can make the token-mean policy loss smaller than normalizing by the actual effective token count. This does not apply in the same way to IS-only / IcePop-weight paths that zero weights without modifying `response_mask`, but it does apply to the rejection-mask path.

A reasonable fix would be to recompute a global effective token count after the bypass rejection mask is produced, or avoid passing the stale `batch_num_tokens` into the policy loss when the effective mask differs from the original mask. The important part is that the denominator used by `token-mean` should match the mask actually passed to `agg_loss`.

[...]

Issue #1332, opened 2025-04-30
[BUG][mcore][context parallel][grpo]


ihebchaa:
Description:
When running GRPO with the Megatron backend and setting context_parallel > 1, the model fails to converge and results are significantly worse compared to runs with context_parallel = 1.
Initial validation scores before training appear off, suggesting a possible issue with model state loading or syncing.
see wandb logs below
Environment:
I am using `whatcanyousee/verl:ngc-th2.6.0-cu124-vllm0.8.2-mcore0.11.0-te2.0` image and latest verl [commit](https://github.com/volcengine/verl/commit/940caadf72d7bc0762ae564620b41c27c439b577) 
Script:
`python3 -m verl.trainer.main_ppo --config-path=config \
   --config-name="ppo_megatron_trainer.yaml" \
   algorithm.adv_estimator=grpo \
   data.train_files=$gsm8k_train_path \
   data.val_files=$gsm8k_test_path \
   data.train_batch_size=256 \
   data.val_batch_size=1024 \
   data.max_prompt_length=512 \
   data.max_response_length=4096 \
   data.filter_overlong_prompts=True \
   data.truncation="error" \
   actor_rollout_ref.model.path="Qwen/Qwen2.5-7B-Instruct" \
   actor_rollout_ref.actor.optim.lr=1e-6 \
   +actor_rollout_ref.model.use_remove_padding=True \
   actor_rollout_ref.actor.use_dynamic_bsz=True \
   actor_rollout_ref.actor.data_loader_seed=42 \
   actor_rollout_ref.actor.ppo_mini_batch_size=256 \
   actor_rollout_ref.actor.ppo_micro_batch_size_per_gpu=1 \
   actor_rollout_ref.actor.megatron.pipeline_model_parallel_size=1 \
   actor_rollout_ref.actor.megatron.tensor_model_parallel_size=2 \
   actor_rollout_ref.actor.megatron.context_parallel_size=2 \
   actor_rollout_ref.model.enable_gradient_checkpointing=False \
   actor_rollout_ref.actor.use_kl_loss=False \
   actor_rollout_ref.actor.kl_loss_coef=0 \
   actor_rollout_ref.actor.kl_loss_type=low_var_kl \
   actor_rollout_ref.rollout.disable_log_stats=False \
   actor_rollout_ref.rollout.log_prob_micro_batch_size_per_gpu=1 \
   actor_rollout_ref.rollout.tensor_model_parallel_size=2 \
[...]

PR #5993, opened 2026-04-13
[megatron] fix: correct CP>1 batch_num_tokens aggregation in engine path

### LINKED ISSUE #5983: [megatron][engine][context parallel] batch_num_tokens aggregation undercounts valid tokens when CP > 1
### Summary

In the Megatron engine path, `batch_num_tokens` appears to be aggregated over the pure DP group in `transformer_impl.py`.

When `CP > 1`, each CP rank only holds a 1/CP slice of the sequence, so reducing `batch_num_tokens` over pure DP undercounts valid tokens by a factor of `CP`.

This affects `loss_agg_mode="token-mean"` normalization in the engine path.

### Current behavior

For `CP > 1`, `batch_num_tokens` is effectively smaller than the true valid-token count for the full DP×CP shard.

As a result, token-mean normalization is incorrect.

### Expected behavior

For `CP > 1`, `batch_num_tokens` should be reduced over the DP×CP group, while `dp_size` should remain pure DP.

### Proposed fix

Use `mpu.get_data_parallel_group(with_context_parallel=True)` when reducing `batch_num_tokens` in the Megatron engine path.

### Validation

I have:
- a minimal code fix in `transformer_impl.py`
- a focused test / reproducer for the CP>1 case
- manual smoke coverage for the engine path

### Related context

Related to #4852  
Potentially related to #1332

### Information

- [ ] My own modified scripts

### Tasks

### Reproduction

The bug can be reproduced with a minimal Megatron engine-path setup where `context_parallel_size > 1`.

#### Config

- `tensor_model_parallel_size = 1`
- `pipeline_model_parallel_size = 1`
- `context_parallel_size = 2`
- `data_parallel_size = 1` (minimal case)
- `loss_agg_mode = "token-mean"`

#### Minimal reproducer

```python
import torch
import torch.distributed as dist
from megatron.core import parallel_state as mpu

def main():
    dist.init_process_group(backend="nccl")
    rank = dist.get_rank()
    local_rank = int(torch.cuda.current_device())
    device = torch.device("cuda", local_rank)

    # Assume Megatron parallel state has already been initialized with:
    # TP=1, PP=1, CP=2
    #
    # Each CP rank only sees a 1/CP slice of the sequence.
    # For a fully valid local slice:
    loss_mask = torch.ones(8, device=device)  # 8 valid tokens on each CP rank
    local_tokens = loss_mask.sum().to(torch.int64)  # = 8 on each rank

    # Current behavior in the buggy engine path:
    wrong = local_tokens.clone()
    dist.all_reduce(
        wrong,
        op=dist.ReduceOp.SUM,
        group=mpu.get_data_parallel_group(),   # pure DP group
    )

    # Expected behavior:
    correct = local_tokens.clone()
    dist.all_reduce(
        correct,
        op=dist.ReduceOp.SUM,
        group=mpu.get_data_parallel_group(with_context_parallel=True),  # DP×CP group
    )

    if rank == 0:
        print(f"local_tokens={int(local_tokens)}")
        print(f"wrong_batch_num_tokens={int(wrong)}")
        print(f"correct_batch_num_tokens={int(correct)}")

        # With CP=2, wrong_batch_num_tokens is smaller by a factor of 2.
        # This undercounts valid tokens and makes token-mean normalization incorrect.

if __name__ == "__main__":
    main()
```

### Expected behavior

For `CP > 1`, the Megatron engine path should aggregate `batch_num_tokens` over the correct DP×CP group, while keeping `dp_size` equal to pure DP. This should make `loss_agg_mode="token-mean"` use the correct valid-token count and avoid CP-dependent undercounting.

alex27594:
## Summary

This PR fixes a correctness bug in the **Megatron engine path** when `context_parallel_size > 1`.

In `MegatronEngine.forward_backward_batch()`, `batch_num_tokens` is used as the denominator for `loss_agg_mode="token-mean"` normalization. When context parallelism is enabled, each CP rank only owns a **1/CP slice of the sequence**, so `batch_num_tokens` must be aggregated over the **DP×CP group**, not the pure DP group.

Using the pure DP group undercounts valid tokens by a factor of `CP`, which makes token-mean loss normalization incorrect in the engine path for `CP > 1`.

This PR fixes that by reducing `batch_num_tokens` over:

```python
mpu.get_data_parallel_group(with_context_parallel=True)
```

while intentionally keeping `dp_size` equal to **pure DP**.

Closes #5983 
Related to #4852  
Potentially related to #1332

---

## Problem

The Megatron engine path computes:
- `batch_num_tokens`
- `dp_size`

and passes them into downstream loss aggregation.

For `loss_agg_mode="token-mean"`, the intended semantics are:

- `batch_num_tokens` = true valid-token count for the full shard participating in the token-mean denominator
- `dp_size` = pure data parallel size

This is correct for standard DP/TP/PP layouts. However, when `context_parallel_size > 1`, the sequence is partitioned across CP ranks, so each rank only sees a fraction of the valid tokens.

If `batch_num_tokens` is reduced over the **pure DP group**, the token count is smaller than the true global token count by a factor of `CP`. That leads to incorrect normalization and over-scaled loss values.

In practice, for `CP > 1`, the current behavior effectively computes:

```text
batch_num_tokens ≈ true_tokens / CP
```

which makes `token-mean` normalization wrong by construction.

---

## Root cause

The denominator used for token-mean aggregation is gathered over the wrong communication group.

For token counting:
- **incorrect** group: pure DP
- **correct** group: DP×CP

This distinction matters because CP partitions the sequence itself. Each CP rank contributes only a partial token view, so token counting must include the context-parallel dimension.

At the same time, `dp_size` must remain **pure DP**, because `agg_loss(..., loss_agg_mode="token-mean")` expects DP scaling semantics there, not DP×CP scaling.

So the correct combination is:
- `batch_num_tokens`: reduce over **DP×CP**
- `dp_size`: keep as **pure DP**

---

## Fix

This PR changes the `batch_num_tokens` all-reduce in `MegatronEngine.forward_backward_batch()` from the pure DP group to the CP-aware DP×CP group:

```python
torch.distributed.all_reduce(
    batch_num_tokens,
    op=torch.distributed.ReduceOp.SUM,
    group=mpu.get_data_parallel_group(with_context_parallel=True),
)
```

and keeps:

```python
dp_size = self.get_data_parallel_size()
```

unchanged.

This is intentionally a **minimal production change**:
- no refactors
- no algorithm changes
- no changes to non-CP paths
- no changes to user-facing configuration

It only corrects the communication group used for token counting.

---

## Why this fix is correct

When `CP > 1`, each CP rank holds only a partial sequence slice. That means the true token-mean denominator must include all valid tokens across the CP dimension.

Reducing `batch_num_tokens` over DP×CP restores exactly that behavior.

The fix also preserves the existing and intended `agg_loss(...)` contract:
- `batch_num_tokens` reflects the full valid-token count
- `dp_size` reflects pure DP only

That makes the Megatron engine path consistent with the expected semantics of `loss_agg_mode="token-mean"`.

---

## Validation

This PR adds focused distributed coverage for the CP>1 case.

### `tests/workers/engine/test_special_megatron_engine_cp_loss_norm.py`

This test is intentionally narrow and checks three things:

1. **DP×CP token aggregation is the correct denominator**
   - verifies that all-reducing over DP×CP yields the true global token count
   - verifies that pure-DP reduction is too small when `CP > 1`

2. **`agg_loss(..., loss_agg_mode="token-mean")` behaves correctly once the denominator is correct**
   - validates the expected normalization math directly

3. **The engine path uses the CP-aware group in `forward_backward_batch()`**
   - acts as a regression marker for the exact CP-specific behavior fixed in this PR

### `tests/workers/engine/run_megatron_engine_cp_loss_norm.sh`

A simple manual smoke script is included to make the reproducer easy to run under `torchrun`.

The script is intentionally lightweight and targeted at this exact bug, so reviewers and future contributors can validate the behavior without needing a full training recipe or model checkpoint.

---

## Why this PR is useful for upstream

This is a **small correctness fix** in an actively maintained path:
- it affects the **modern Megatron engine path**
- it is directly relevant to users running with `context_parallel_size > 1`
- it improves loss-normalization correctness rather than changing behavior for unrelated configurations

It also complements existing engine-path normalization discussions such as #4852 by adding a focused fix and targeted coverage for a specific CP-related failure mode.

In short, this PR:
- fixes a real normalization bug
- keeps the production diff minimal
- adds narrowly scoped regression coverage for the exact issue being fixed

---

## Scope

This PR is intentionally limited to:
- the Megatron **engine path**
- `CP > 1` token-count normalization
- focused distributed validation

It does **not** include:
- legacy worker changes
- FSDP changes
- rollout changes
- unrelated refactors
- broad Megatron behavior changes outside the token-count aggregation path

---

## Local validation

The following were used during validation:

```bash

# manual smoke
bash tests/workers/engine/run_megatron_engine_cp_loss_norm.sh

# direct distributed run
torchrun --standalone --nnodes=1 --nproc_per_node=8 \
  tests/workers/engine/test_special_megatron_engine_cp_loss_norm.py
```

I can narrow or reshape the validation further if maintainers prefer a different test style for engine-path special tests.

### DIFF
--- a/tests/workers/engine/run_megatron_engine_cp_loss_norm.sh
+++ b/tests/workers/engine/run_megatron_engine_cp_loss_norm.sh
@@ -1,0 +1,44 @@
+#!/usr/bin/env bash
+# Megatron engine loss normalization — CP>1 smoke test.
+#
+# Validates that batch_num_tokens is all-reduced over DP×CP (not just DP)
+# in the modern engine path (MegatronEngine.forward_backward_batch).
+#
+# Default: 1 node × 8 GPUs  (TP=1, PP=2, CP=2, DP=2)
+#
+# Usage:
+#   bash tests/workers/engine/run_megatron_engine_cp_loss_norm.sh
+#
+# Override parallelism:
+#   NPROC_PER_NODE=4 ACTOR_PP=1 ACTOR_CP=2 \
+#       bash tests/workers/engine/run_megatron_engine_cp_loss_norm.sh
+#
+# Environment variables:
+#   NPROC_PER_NODE   — GPUs per node            (default: 8)
+#   ACTOR_TP         — tensor parallelism       (default: 1)
+#   ACTOR_PP         — pipeline parallelism     (default: 2)
+#   ACTOR_CP         — context parallelism      (default: 2)
+
+set -euo pipefail
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+
+NPROC_PER_NODE="${NPROC_PER_NODE:-8}"
+ACTOR_TP="${ACTOR_TP:-1}"
+ACTOR_PP="${ACTOR_PP:-2}"
+ACTOR_CP="${ACTOR_CP:-2}"
+
+WORLD_SIZE=$((NPROC_PER_NODE))
+DP_SIZE=$((WORLD_SIZE / (ACTOR_TP * ACTOR_PP * ACTOR_CP)))
+
+echo "=== Megatron Engine CP loss normalization test ==="
+echo "  ${NPROC_PER_NODE} GPUs: TP=${ACTOR_TP} PP=${ACTOR_PP} CP=${ACTOR_CP} DP=${DP_SIZE}"
+echo "=================================================="
+
+export ACTOR_PP ACTOR_TP ACTOR_CP
+
+torchrun \
+    --standalone \
+    --nnodes=1 \
+    --nproc-per-node="$NPROC_PER_NODE" \
+    "${SCRIPT_DIR}/test_special_megatron_engine_cp_loss_norm.py"
--- a/tests/workers/engine/test_special_megatron_engine_cp_loss_norm.py
+++ b/tests/workers/engine/test_special_megatron_engine_cp_loss_norm.py
@@ -1,0 +1,221 @@
+# Copyright 2025 Bytedance Ltd. and/or its affiliates
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""
+Megatron engine path: batch_num_tokens correctness with CP > 1.
+
+Verifies that MegatronEngine.forward_backward_batch all-reduces
+batch_num_tokens over the DP×CP group (with_context_parallel=True),
+not just the pure DP group.  When CP > 1 each CP rank holds only 1/CP
+of the sequence, so using the pure DP group would undercount tokens by
+a factor of CP, inflating gradients by the same factor.
+
+Usage::
+
+    torchrun --standalone --nnodes=1 --nproc-per-node=8 \
+        tests/workers/engine/test_special_megatron_engine_cp_loss_norm.py
+"""
+
+import os
+
+import megatron.core.parallel_state as mpu
+import torch
+import torch.distributed as dist
+
+from verl.utils.distributed import destroy_global_process_group, initialize_global_process_group
+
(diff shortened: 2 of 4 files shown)


wuxibin89:
Megatron will scale loss by `cp_group_size`, so `batch_num_tokens` should be aggregated over DP group instead of DPxCP group?
https://github.com/NVIDIA/Megatron-LM/blob/core_v0.16.0/megatron/core/pipeline_parallel/schedules.py#L268

PR #6002, opened 2026-04-14
[fully_async, ckpt, rollout, trainer, tool, cfg] fix: ROCm async training compatibility for AMD MI300X

xiaohong42:
### What does this PR do?

Fix multiple issues that prevent fully async FSDP2 training from working on AMD ROCm platforms (MI300X series).

**Environment:**
- AMD Instinct MI3xx (8× GPU, 192 GB HBM each), ROCm 7.2.0, PyTorch 2.10.0+rocm7.2.0, vLLM v0.19.1rc0
- Cross-validated on NVIDIA H20 with CUDA, vLLM 0.18.2 (latest verl main + this patch): no regression observed up to step 86, where a pre-existing bug (`AttributeError: 'list' object has no attribute 'dim'` in `agent_loop.py:696`) is hit — this bug exists with or without this patch and is unrelated to these changes.

**Training curves (MI3xx vs H20) and training script**

[dapo_7b_fully_async.sh](https://github.com/user-attachments/files/26700697/dapo_7b_fully_async.sh)
qwen2 5_7b_fully_async

### Checklist Before Starting

- [x] Search for similar PRs. Paste at least one query link here: https://github.com/volcengine/verl/pulls?q=is%3Apr+rocm+async

### Test

Validated by full async FSDP2 DAPO/GRPO RL + ReTool training on AMD MI3xx:
- 250+ global steps completed, 60+ weight synchronization cycles without OOM or deadlock
- Cross-validated on NVIDIA H20: training runs normally with this patch applied (no regression)
  - H20 (CUDA) cross-validation note:

Applied this patch on latest verl main (commit 9b54564f) + vLLM 0.18.2 on NVIDIA H20. Training ran normally up to step 86 / global_step 344, where a pre-existing bug in agent_loop.py:698 is hit:

File "verl/experimental/agent_loop/agent_loop.py", line 698, in _agent_loop_postprocess
    if response_mask_output["input_ids"].dim() == 1:
AttributeError: 'list' object has no attribute 'dim'
This is caused by tokenizer.pad() returning a Python list instead of a torch.Tensor for response_mask in certain edge cases, even with return_tensors="pt". This bug exists on the current main branch with or without this patch — it is not introduced by any changes in this PR. The file agent_loop.py is not modified in this PR.

MI3xx (ROCm 7.2): 250+ global steps, 60+ weight syncs completed without OOM or deadlock. The same agent_loop.py bug was also encountered on MI3xx at a later step, confirming it is platform-independent.

### API and Usage Example

No API changes. All fixes are internal implementation details.

### Design & Code Changes

1. **NCCL checkpoint engine: unify buffers to torch tensors** (`nccl_checkpoint_engine.py`)
   - Replace `cupy` buffers with `torch.zeros` to fix HIP stream synchronization issues on ROCm
   - Remove `cp.asarray()` conversion in `send_weights`
   - Persist buffers across calls to prevent HIP memory fragmentation OOM (see #9)

2. **Add `HSA_NO_SCRATCH_RECLAIM` env var** (`constants_ppo.py`)
   - Required by AMD RCCL on MI300X; without it, FSDP initialization fails with `ncclSystemError`

3. **Fix `numpy.bool_` JSON serialization** (`ray_trainer.py`)
   - Add `default=str` fallback for `json.dumps` since numpy 2.x `bool_` is no longer a Python `bool` subclass

4. **Materialize generator before `send_weights`** (`engine_workers.py`)
   - `get_per_tensor_param()` returns a generator containing `full_tensor()` calls that trigger FSDP `all_gather`
   - Lazy consumption causes rank-misaligned collective calls → deadlock
   - Fix: `list()` to materialize + `torch.cuda.synchronize()` before sending

5. **ZMQ IPC handle: use rank instead of GPU UUID** (`vllm_rollout.py`, `utils.py`)
   - On ROCm, `CheckpointEngineWorker` and vLLM worker see different GPU UUIDs due to different `CUDA_VISIBLE_DEVICES`/`HIP_VISIBLE_DEVICES` settings
   - Use deterministic rank number instead

6. **Clean up stale ZMQ IPC socket files** (`bucketed_weight_transfer.py`)
   - Remove leftover `/tmp/rl-colocate-zmq-rank-*.sock` files before `bind()` and after `close()` to prevent `Address already in use` on restart

7. **Fix Hydra searchpath** (`fully_async_ppo_trainer.yaml`)
   - Use `pkg://verl.trainer.config` instead of `file://verl/trainer/config` for editable installs

8. **Sandbox Ray actor reuse** (`sandbox_fusion_tools.py`)
   - Add `name` and `get_if_exists=True` to prevent duplicate `ExecutionWorker` actor creation

9. **Persist weight sync buffers to prevent OOM** (`nccl_checkpoint_engine.py`, `vllm_rollout.py`, `bucketed_weight_transfer.py`)
   - On ROCm/HIP, `torch.cuda.empty_cache()` does not effectively return physical memory to the system
   - Repeated alloc/free of large buffers (NCCL 4 GB + IPC 2 GB) causes HIP memory fragmentation → OOM after ~10 weight sync cycles
   - Fix: allocate once and reuse across sync iterations

> [!IMPORTANT]

- [x] Add / Update [the documentation](https://github.com/volcengine/verl/tree/main/docs). The official documents will be compiled after the merger.
- [x] Add unit or end-to-end test(s) to [the CI workflow](https://github.com/volcengine/verl/tree/main/.github/workflows) to cover all the code. If not feasible, explain why: These fixes target ROCm-specific runtime behavior (HIP memory management, RCCL env vars, GPU UUID mismatch) that cannot be reproduced in CI without AMD GPU hardware.

### DIFF
--- a/verl/checkpoint_engine/nccl_checkpoint_engine.py
+++ b/verl/checkpoint_engine/nccl_checkpoint_engine.py
@@ -118,6 +118,8 @@ class NCCLCheckpointEngine(CheckpointEngine):
         self.group_name = group_name
         self.rebuild_group = rebuild_group
         self.rollout_dtype = rollout_dtype
+        self.send_buf = None
+        self.recv_buf = None
 
         # start zeromq server for broadcasting bucket tensor metadata
         self.is_master = is_master
@@ -126,13 +128,9 @@ class NCCLCheckpointEngine(CheckpointEngine):
             self._start_zmq_server()
 
     def prepare(self) -> MasterMetadata:
-        # For master process, use cupy instead of torch to avoid memory register error
-        # when `PYTORCH_CUDA_ALLOC_CONF=expandable_segments:True`.
-        if self.is_master:
-            self.send_buf = cp.zeros(self.bucket_size, dtype=cp.uint8)
-            self.recv_buf = cp.zeros(self.bucket_size, dtype=cp.uint8)
-        else:
+        if self.send_buf is None:
             self.send_buf = torch.zeros(self.bucket_size, dtype=torch.uint8, device="cuda")
+        if self.recv_buf is None:
             self.recv_buf = torch.zeros(self.bucket_size, dtype=torch.uint8, device="cuda")
 
         return MasterMetadata(zmq_ip=self.ip, zmq_port=self.listen_port) if self.is_master else None
@@ -144,11 +142,6 @@ class NCCLCheckpointEngine(CheckpointEngine):
                 collective.destroy_collective_group(self.group_name)
             self.rank = None
             self.world_size = None
-
-        self.send_buf = None
-        self.recv_buf = None
-
-        torch.cuda.empty_cache()
 
     @classmethod
     def build_topology(cls, trainer_world_size: int, rollout_world_size: int, metadata: list[dict]):
@@ -274,7 +267,7 @@ class NCCLCheckpointEngine(CheckpointEngine):
                 "dtype": weight.dtype,
                 "offset": offset,
             }
-            send_buf[offset : offset + weight.nbytes] = cp.asarray(weight.view(-1).view(torch.uint8))
+            send_buf[offset : offset + weight.nbytes] = weight.view(-1).view(torch.uint8)
             offset += weight.nbytes
 
         # broadcast last bucket
--- a/verl/workers/engine_workers.py
+++ b/verl/workers/engine_workers.py
@@ -32,7 +32,7 @@ from verl.single_controller.base.decorator import Dispatch, make_nd_compute_data
 from verl.trainer.distillation import distillation_ppo_loss, is_distillation_enabled
 from verl.utils import tensordict_utils as tu
 from verl.utils.config import omega_conf_to_dataclass
-from verl.utils.device import get_device_name, is_npu_available, set_expandable_segments
+from verl.utils.device import get_device_name, get_torch_device, is_npu_available, set_expandable_segments
 from verl.utils.distributed import initialize_global_process_group_ray, set_numa_affinity
 from verl.utils.flops_counter import FlopsCounter
 from verl.utils.import_utils import import_external_libs
@@ -674,7 +674,9 @@ class ActorRolloutRefWorker(Worker, DistProfilerExtension):
         # 0. send_weights only for async training with disaggregated trainer and rollout
         if self.config.rollout.checkpoint_engine.backend != "naive":
             per_tensor_param, _ = self.actor.engine.get_per_tensor_param()
-            await self.checkpoint_engine.send_weights(per_tensor_param)
+            per_tensor_param = list(per_tensor_param)
+            get_torch_device().synchronize()
+            await self.checkpoint_engine.send_weights(iter(per_tensor_param))
             return
 
         set_expandable_segments(False)
--- a/verl/workers/rollout/vllm_rollout/vllm_rollout.py
+++ b/verl/workers/rollout/vllm_rollout/vllm_rollout.py
@@ -95,7 +95,7 @@ class ServerAdapter(BaseRollout):
             self.sleep_level = VLLM_SLEEP_LEVEL
 
         self.device_uuid = get_device_uuid(get_device_id())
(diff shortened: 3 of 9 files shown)


verl/workers/engine_workers.py
```diff
@@ -674,7 +674,9 @@ async def update_weights(self, global_steps: int = None):
         # 0. send_weights only for async training with disaggregated trainer and rollout
         if self.config.rollout.checkpoint_engine.backend != "naive":
             per_tensor_param, _ = self.actor.engine.get_per_tensor_param()
-            await self.checkpoint_engine.send_weights(per_tensor_param)
+            per_tensor_param = list(per_tensor_param)
```
wuxibin89, line 677: This will materialize weight generator and gather all sharded weight into each GPU, causing cuda oom for large model.

verl/workers/rollout/vllm_rollout/vllm_rollout.py
```diff
@@ -95,7 +95,7 @@ def __init__(
             self.sleep_level = VLLM_SLEEP_LEVEL
 
         self.device_uuid = get_device_uuid(get_device_id())
-        self.zmq_handle = f"ipc:///tmp/rl-colocate-zmq-{self.device_uuid}.sock"
+        self.zmq_handle = f"ipc:///tmp/rl-colocate-zmq-rank-{rank % local_world_size}.sock"
```
wuxibin89, line 98: This will conflict for multiple vllm replicas in same node, e.g 2 replicas with TP=4 located on same node.

verl/checkpoint_engine/nccl_checkpoint_engine.py
```diff
@@ -126,13 +128,9 @@ def __init__(
             self._start_zmq_server()
 
     def prepare(self) -> MasterMetadata:
-        # For master process, use cupy instead of torch to avoid memory register error
-        # when `PYTORCH_CUDA_ALLOC_CONF=expandable_segments:True`.
```
wuxibin89, line 130: Please respect this comment.

wuxibin89 (changes requested):

verl/workers/engine_workers.py
xiaohong42: Thanks for the review. The list() materialization was added to fix an FSDP2 deadlock: get_per_tensor_param() returns a lazy generator where each element calls full_tensor() (triggering all_gather). In send_weights, rank 0 consumes the generator with broadcast I/O waits between weights, while rank < 0 just does a bare pass loop — this timing difference causes misaligned all_gather calls across ranks, leading to deadlock.

However, I agree that list() materializes all full tensors simultaneously, which could cause OOM for large models.

Fixed: removed list() from engine_workers.py (restored to original code), and instead added torch.cuda.synchronize() in the rank < 0 consumption path inside nccl_checkpoint_engine.py. This keeps all_gather calls aligned across ranks without materializing all weights at once.

verl/workers/rollout/vllm_rollout/vllm_rollout.py
xiaohong42: Good point. Using rank % local_world_size alone is not unique when multiple replicas share the same node (e.g., 2 replicas with TP=4 on one 8-GPU node).

Fixed: added replica_rank to the socket path. The sender now uses ipc:///tmp/rl-colocate-zmq-replica-{replica_rank}-rank-{local_rank}.sock, and passes replica_rank to the receiver via update_weights_from_ipc kwargs. The receiver stores it as self._replica_rank and uses the same path format in _get_zmq_handle().

The two modifications mentioned above have been verified end-to-end.

qwen2 5_7b_fully_async_fix

verl/checkpoint_engine/nccl_checkpoint_engine.py
xiaohong42: Apologies for removing the comment. Restored the original two-line comment as-is.

xiaohong42: Reverted. The cupy issue was due to our local cupy installation misconfiguration. This logic has been restored to the original cupy/torch split.

Issue #4348, opened 2025-11-29
[RFC]partial rollout


HelloWorldBeginner:
# Motivation

During the reinforcement learning training process, as the model performance continues to improve, the output response sequences also keep lengthening — especially in the slow thinking mode, the sequence length can reach tens of thousands of tokens, which leads to a continuous increase in the proportion of time consumed in the inference phase.
Meanwhile, we have statistically analyzed the distribution of response output lengths (see the figure below) and found that there is a significant long-tail effect: although a small number of samples account for an extremely low proportion, their sequence lengths far exceed the average level.

<img width="972" height="716" alt="Image" src="https://github.com/user-attachments/assets/301e09e4-2a96-453c-a3ea-7db740207969" />

# Method

We can adopt the method of early truncation to address the long-tail problem in the inference phase and improve inference performance. The specific approach is illustrated in the figure below: when a sample's response is excessively long, it is truncated early, and the remaining part after truncation is incorporated into the next inference process, thereby reducing unnecessary waiting time.

<img width="972" height="310" alt="Image" src="https://github.com/user-attachments/assets/8c0fc33a-d6f3-465a-af7a-b4be8aabb411" />

# Proposed Design

Based on the aforementioned approach, we have designed and implemented a complete partial rollout workflow. Below, we split this workflow into the Training Workflow and Inference Workflow for detailed elaboration.
In the Training Workflow, we first add two fields (age and raw_response_ids) to the data attributes: age records the aging rounds of data samples, while raw_response_ids stores the partial responses that were left unfinished in the previous inference round. Additionally, we introduce the AggregatorActor component, whose core responsibility is to aggregate samples that have completed inference across all DP (Data Parallelism) groups. When the cumulative number of completed inference samples reaches the preset threshold, the AggregatorActor sends an inference completion signal to all worker nodes. Upon receiving the signal, each worker immediately terminates the current inference process, saves the unfinished partial responses to the raw_response_ids field, and these partial responses will continue to be processed in subsequent rounds.

We first filter out the incompletely inferred parts from all samples to form a partial_batch — the unfinished responses of these samples are appended to the original prompts for subsequent rollout inference (i.e., the partial rollout mechanism).
[...]

PR #6029, opened 2026-04-16
[fully_async] fix: replace routed_experts on partial rollout resume i…

NoonePauseferg:
### What does this PR do?

Fixes a bug in `FullyAsyncLLMServerManager.generate()` where `routed_experts` was incorrectly concatenated via `torch.cat` during partial rollout resume, causing duplicated routing data and broken MoE expert replay in the actor.

sglang returns `routed_experts` for the **full sequence** (prompt + all generated tokens). Evidence from sglang source:

1. **[`io_struct.py#L1020`](https://github.com/sgl-project/sglang/blob/v0.5.9/python/sglang/srt/managers/io_struct.py#L1020)** — field definition:
   ```python
   # The routed experts for each token, including both input and output tokens
   # routed_experts[i] is a tensor of shape (token, layer, top_k) for request i
   routed_experts: List[Optional[torch.Tensor]]
   ```

2. **`schedule_batch.py`** — `seqlen` used to collect routing covers the full sequence:
   ```python
   @property
   def seqlen(self) -> int:
       return len(self.origin_input_ids) + len(self.output_ids)
   ```

3. **`topk.py#L1049-1051`** — capture is unconditional (no prefill/decode check):
   ```python
   get_global_experts_capturer().capture(layer_id=layer_id, topk_ids=topk_ids)
   ```

4. **`scheduler_output_processor_mixin.py#L105-111`** — collection uses full `seqlen`:
   ```python
   req.routed_experts = get_global_experts_capturer().get_routed_experts(
       req_pool_idx=req.req_pool_idx,
       seqlen=req.seqlen,  # origin_input_ids + output_ids
       req_to_token_pool=self.req_to_token_pool,
   )
   ```

When partial rollout resumes after abort, the input becomes `prompt + already_generated_tokens`. sglang re-processes the entire input during prefill and returns `routed_experts` covering all positions. The old code concatenated this with the previous `routed_experts`:

```
old routing:    prompt + A B C
new routing:    prompt + A B C + D E
concat result:  prompt + A B C + prompt + A B C + D E   <-- duplicated!
expected:       prompt + A B C + D E
```

This shifted the routing and caused incorrect MoE expert replay, leading to `actor/ppo_kl` spikes.

**Fix:** replace `routed_experts` instead of concatenating, since the resumed call already covers all positions.

Related: #4348 (partial rollout RFC), #4101 (R3 router replay), #5344 (R3 in fully async)

### Checklist Before Starting

- [x] Search for similar PRs: https://github.com/verl-project/verl/pulls?q=routed_experts+partial_rollout

### Test

- Ran async training with `partial_rollout=True` and `enable_rollout_routing_replay=True` (R3 mode)
- Verified `actor/ppo_kl` no longer spikes after partial rollout resume
- Verified `routed_experts` tensor shape matches `(prompt_len + response_len, num_layers, top_k)` after resume

### Design & Code Changes

Single-line change in `verl/experimental/fully_async_policy/agent_loop/agent_loop.py`:

```diff
- if output.routed_experts is not None:
-     if final_output.routed_experts is None:
-         final_output.routed_experts = output.routed_experts
-     else:
-         final_output.routed_experts = torch.cat([final_output.routed_experts, output.routed_experts], dim=0)
+ # sglang returns routed_experts for the full sequence (prompt + all tokens),
+ # so on partial rollout resume the new output already covers all positions.
+ if output.routed_experts is not None:
+     final_output.routed_experts = output.routed_experts
```

- [ ] Add unit or end-to-end test(s) to [the CI workflow](https://github.com/volcengine/verl/tree/main/.github/workflows) to cover all the code. If not feasible, explain why: This is an async distributed training bug that requires multi-node sglang + megatron setup with MoE model and partial rollout enabled. Not feasible to reproduce in CI.

### DIFF
--- a/verl/experimental/fully_async_policy/agent_loop/agent_loop.py
+++ b/verl/experimental/fully_async_policy/agent_loop/agent_loop.py
@@ -17,7 +17,6 @@ import os
 from typing import Any, Optional
 
 import ray
-import torch
 from omegaconf import DictConfig
 
 from verl.experimental.agent_loop.agent_loop import (
@@ -96,11 +95,10 @@ class FullyAsyncLLMServerManager(AsyncLLMServerManager):
             final_output.token_ids.extend(output.token_ids)
             if output.log_probs is not None:
                 final_output.log_probs.extend(output.log_probs)
+            # sglang returns routed_experts for the full sequence (prompt + all tokens),
+            # so on partial rollout resume the new output already covers all positions.
             if output.routed_experts is not None:
-                if final_output.routed_experts is None:
-                    final_output.routed_experts = output.routed_experts
-                else:
-                    final_output.routed_experts = torch.cat([final_output.routed_experts, output.routed_experts], dim=0)
+                final_output.routed_experts = output.routed_experts
             if output.num_preempted is not None:
                 final_output.num_preempted += output.num_preempted
             final_output.stop_reason = output.stop_reason


ArronHZG:
Please use `pre-commit run --all-files` to format code.

verl/experimental/fully_async_policy/agent_loop/agent_loop.py
```diff
@@ -96,11 +95,10 @@ async def generate(
             final_output.token_ids.extend(output.token_ids)
             if output.log_probs is not None:
                 final_output.log_probs.extend(output.log_probs)
+            # sglang returns routed_experts for the full sequence (prompt + all tokens),
+            # so on partial rollout resume the new output already covers all positions.
             if output.routed_experts is not None:
-                if final_output.routed_experts is None:
-                    final_output.routed_experts = output.routed_experts
-                else:
-                    final_output.routed_experts = torch.cat([final_output.routed_experts, output.routed_experts], dim=0)
```
wuxibin89, line 103: I think we should extract routed_experts corresponding to  partially generated `output.token_ids` to reflect different model version?
```python
if len(output.token_ids) > 0:
  final_output.routed_experts = torch.cat([final_output.routed_experts, output.routed_experts[-len(output.token_ids):]], dim=0)
```

wuxibin89 (changes requested):

verl/experimental/fully_async_policy/agent_loop/agent_loop.py
NoonePauseferg: Good idea at first glance, but I think simple replace is actually better here.

The actor update always runs with the latest weights (v2). If we replay v1's routing with v2's weights, the recomputed logprob is neither pi_v1 nor pi_v2 — it's an inconsistent mix. Simple replace gives v2 routing + v2 weights = clean pi_v2, which makes the importance ratio well-defined.

The codebase has a rollout correction pipeline (IS weights, rejection sampling) that can mitigate staleness when enabled, but it's off by default (`bypass_mode=True`).

That said, worth experimenting with both approaches — the per-iteration routing might still help in practice if routing signal matters more than weight consistency.

### TODO: experiments (64-expert 10B MoE model)
- [x] Compare simple replace vs per-iteration slicing with partial rollout enabled
- [x] Track `rollout_corr/kl`, `ppo_kl`, and reward curves for both approaches
- [ ] Test with `bypass_mode=False` + rollout IS enabled to see if correction compensates for either approach

wuxibin89: Could you help to do some comparison experiment?

NoonePauseferg: yeah - im in process

wuxibin89: @NoonePauseferg Merge this PR to use latest router_experts as baseline. Please update per-iteration slicing experiment result if there any update.

wuxibin89 (approved):

verl/experimental/fully_async_policy/agent_loop/agent_loop.py
NoonePauseferg: recap_vs_simple
grad_norms


i have done comparison experiments on megatron_sglang setup
- version where we remember routing show more stable training results

NoonePauseferg: created following pr here: https://github.com/verl-project/verl/pull/6046

PR #6039, opened 2026-04-17
[trainer, rollout, algo] refactor: Remove OPD colocate mode

JacobHelwig:
Removes support for colocate mode for OPD. We only support standalone teachers now.


wuxibin89 (approved):

AvengersZhi:
Could you please explain why the "colocate mode" was removed from the OPD?

PR #6047, opened 2026-04-17
[rollout] feat: prep for vexact release

pengwu22:
### What does this PR do?

> Prepare the OSS release (e2e test will be added after the OSS)

### DIFF
--- a/tests/special_sanity/check_pr_title.py
+++ b/tests/special_sanity/check_pr_title.py
@@ -19,7 +19,18 @@ import re
 pr_title = os.environ.get("PR_TITLE", "").strip()
 
 # Define rules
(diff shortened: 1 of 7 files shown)


wuxibin89:
@pengwu22 I think we should provide a rollout register and load hook to dynamic load private rollout engine, then we can load veaxct out-of-tree.
https://github.com/verl-project/verl/blob/main/verl/workers/rollout/replica.py#L390-L399

wuxibin89:
vexact will be dynamically loaded by `VERL_USE_EXTERNAL_MODULES`
https://github.com/verl-project/verl/blob/main/verl/__init__.py#L38-L41

PR #67, opened 2024-12-31
Super tiny fix typo in title

fzyzcjy:

PR #6062, opened 2026-04-20
[fully_async, rollout, trainer, tool, cfg] fix: ROCm async training compatibility for AMD MI300X

xiaohong42:
## What does this PR do?

Fix multiple issues that prevent fully async FSDP2 training from working on AMD ROCm platforms (MI300X series).

**Environment:**
- AMD Instinct MI3xx (8× GPU, 192 GB HBM each), ROCm 7.2, PyTorch 2.10+rocm7.2, vLLM v0.18.1rc1
- Cross-validated on NVIDIA H20 with CUDA (no regression observed)

**Training curves (MI3xx vs H20) and training script will be attached in PR comments.**
[dapo_7b_fully_async.sh](https://github.com/user-attachments/files/26700697/dapo_7b_fully_async.sh)
qwen2 5_7b_fully_async_1

### Checklist Before Starting

- [x] Search for similar PRs: https://github.com/verl-project/verl/pulls?q=is%3Apr+rocm+async
- [x] Format: `[{modules}] {type}: {description}`

### Test

Validated by full async FSDP2 DAPO/GRPO RL + ReTool training on AMD MI3xx:
- 140+ training steps completed without errors, deadlocks or OOM
- Reward improved from -0.8 to 0 over 12+ hours of training
- Cross-validated on NVIDIA H20: all changes are platform-safe (AMD-specific env vars are ignored on NVIDIA/Ascend; ZMQ handle changes use platform-independent rank logic)

### API and Usage Example

No API changes. All fixes are internal implementation details.

### Design & Code Changes

1. **Add `HSA_NO_SCRATCH_RECLAIM` env var** (`constants_ppo.py`)
   - Required by AMD RCCL on MI300X; without it, FSDP initialization fails with `ncclSystemError`
   - Added alongside existing platform-specific vars (HCCL for Ascend); ignored on non-AMD platforms

2. **Fix `numpy.bool_` JSON serialization** (`ray_trainer.py`)
   - Add `default=str` fallback for `json.dumps` since numpy 2.x `bool_` is no longer a Python `bool` subclass

3. **ZMQ IPC handle: use `(replica_rank, local_rank)` instead of GPU UUID** (`vllm_rollout.py`, `utils.py`, `vllm_async_server.py`)
   - On ROCm, `CheckpointEngineWorker` and vLLM worker see different GPU UUIDs due to different `CUDA_VISIBLE_DEVICES`/`HIP_VISIBLE_DEVICES` settings
   - Sender uses `rollout_rank % local_world_size` to derive node-local rank, matching vLLM worker's `local_rank` on every node (fixes multi-node mismatch)
   - `replica_rank` prefix avoids socket collisions when multiple replicas share a node
   - `VERL_REPLICA_RANK` env var is set in `vLLMHttpServer.__init__` and inherited by vLLM worker subprocesses
   - Both `vLLMColocateWorkerExtension` and `vLLMOmniColocateWorkerExtension` are updated

4. **Clean up stale ZMQ IPC socket files** (`bucketed_weight_transfer.py`)
   - Remove leftover `.sock` files before `bind()` and after `_cleanup()` to prevent `Address already in use` on restart

5. **Fix Hydra searchpath** (`fully_async_ppo_trainer.yaml`)
   - Use `pkg://verl.trainer.config` instead of `file://verl/trainer/config` for editable installs

6. **Sandbox Ray actor reuse** (`sandbox_fusion_tools.py`)
   - Add `name` and `get_if_exists=True` to prevent duplicate `ExecutionWorker` actor creation

### Platform Compatibility

All changes are safe for non-AMD platforms:
- `HSA_NO_SCRATCH_RECLAIM`: AMD-specific env var, silently ignored on NVIDIA/Ascend
- ZMQ handle changes: use platform-independent rank arithmetic; `VERL_REPLICA_RANK` defaults to `"0"` for single-replica setups
- Other changes (json default=str, IPC cleanup, Hydra pkg://, actor reuse) are pure logic improvements

> [!IMPORTANT]

- [x] Add / Update [the documentation](https://github.com/volcengine/verl/tree/main/docs). The official documents will be compiled after the merger.
- [ ] Add unit or end-to-end test(s) to [the CI workflow](https://github.com/volcengine/verl/tree/main/.github/workflows) to cover all the code. If not feasible, explain why: These fixes target ROCm-specific runtime behavior (HIP memory management, RCCL env vars, GPU UUID mismatch) that cannot be reproduced in CI without AMD GPU hardware.

### DIFF
--- a/verl/tools/sandbox_fusion_tools.py
+++ b/verl/tools/sandbox_fusion_tools.py
@@ -90,7 +90,7 @@ def init_execution_pool(
     if mode == PoolMode.ThreadMode:
         return (
             ray.remote(ExecutionWorker)
-            .options(max_concurrency=num_workers)
+            .options(name="sandbox-execution-pool", get_if_exists=True, max_concurrency=num_workers)
             .remote(enable_global_rate_limit=enable_global_rate_limit, rate_limit=rate_limit)
         )
     else:
--- a/verl/trainer/constants_ppo.py
+++ b/verl/trainer/constants_ppo.py
@@ -31,6 +31,7 @@ PPO_RAY_RUNTIME_ENV = {
         # https://www.hiascend.com/document/detail/zh/canncommercial/83RC1/maintenref/envvar/envref_07_0143.html
         "HCCL_HOST_SOCKET_PORT_RANGE": "auto",
         "HCCL_NPU_SOCKET_PORT_RANGE": "auto",
+        "HSA_NO_SCRATCH_RECLAIM": "1",
     },
 }
 
--- a/verl/experimental/fully_async_policy/config/fully_async_ppo_trainer.yaml
+++ b/verl/experimental/fully_async_policy/config/fully_async_ppo_trainer.yaml
@@ -1,6 +1,6 @@
 hydra:
   searchpath:
-    - file://verl/trainer/config
+    - pkg://verl.trainer.config
 
 defaults:
   - ppo_trainer
--- a/verl/trainer/ppo/ray_trainer.py
+++ b/verl/trainer/ppo/ray_trainer.py
@@ -422,7 +422,7 @@ class RayPPOTrainer:
         lines = []
         for i in range(n):
             entry = {k: v[i] for k, v in base_data.items()}
-            lines.append(json.dumps(entry, ensure_ascii=False))
+            lines.append(json.dumps(entry, ensure_ascii=False, default=str))
 
         with open(filename, "w") as f:
             f.write("\n".join(lines) + "\n")
--- a/verl/workers/rollout/vllm_rollout/bucketed_weight_transfer.py
+++ b/verl/workers/rollout/vllm_rollout/bucketed_weight_transfer.py
@@ -159,7 +159,7 @@ class BucketedWeightSender:
             ipc_path = self.zmq_handle[len("ipc://") :]
             try:
                 os.remove(ipc_path)
-            except FileNotFoundError:
+            except OSError:
                 pass
         self.socket = self.zmq_context.socket(zmq.REQ)
         self.socket.bind(self.zmq_handle)
@@ -195,7 +195,7 @@ class BucketedWeightSender:
             ipc_path = self.zmq_handle[len("ipc://") :]
             try:
                 os.remove(ipc_path)
-            except FileNotFoundError:
+            except OSError:
                 pass
         del self.buffer
         self.buffer = None
--- a/verl/workers/rollout/vllm_rollout/bucketed_weight_transfer.py
+++ b/verl/workers/rollout/vllm_rollout/bucketed_weight_transfer.py
@@ -155,6 +155,12 @@ class BucketedWeightSender:
 
     def _init_socket(self):
         """Initialize ZMQ REQ socket and bind."""
+        if self.zmq_handle.startswith("ipc://"):
+            ipc_path = self.zmq_handle[len("ipc://") :]
+            try:
+                os.remove(ipc_path)
+            except FileNotFoundError:
+                pass
         self.socket = self.zmq_context.socket(zmq.REQ)
         self.socket.bind(self.zmq_handle)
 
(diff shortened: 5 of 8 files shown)


verl/tools/sandbox_fusion_tools.py
```diff
@@ -90,7 +90,7 @@ def init_execution_pool(
     if mode == PoolMode.ThreadMode:
         return (
             ray.remote(ExecutionWorker)
-            .options(max_concurrency=num_workers)
+            .options(name="sandbox-execution-pool", get_if_exists=True, max_concurrency=num_workers)
```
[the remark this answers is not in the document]
xiaohong42: Good point. In practice, verl training jobs call `ray.init()` which creates a dedicated Ray cluster per job, so namespace collisions between independent jobs do not occur. The hardcoded name is intentional here — it matches the existing pattern used by `TokenBucketWorker` (line 71: `name="rate-limiter", get_if_exists=True`) in the same file. The goal is to reuse the actor within a single training session when `init_execution_pool` is called multiple times (e.g., across rollout iterations), not to share across jobs. Making the name session-unique would defeat the reuse purpose. I think this is fine as-is for this PR scope.

wuxibin89:
@xiaohong42 Does this PR have other PR dependencies to work on AMD GPU?

verl/trainer/constants_ppo.py
```diff
@@ -31,6 +31,7 @@
         # https://www.hiascend.com/document/detail/zh/canncommercial/83RC1/maintenref/envvar/envref_07_0143.html
         "HCCL_HOST_SOCKET_PORT_RANGE": "auto",
         "HCCL_NPU_SOCKET_PORT_RANGE": "auto",
+        "HSA_NO_SCRATCH_RECLAIM": "1",
```
wuxibin89, line 34: What's this environment for?

xiaohong42:
> @xiaohong42 Does this PR have other PR dependencies to work on AMD GPU?

No hard PR dependency — this PR is self-contained. But for end-to-end async training with ReTool on AMD GPUs, two additional minor fixes are needed in other repos:

recipe submodule (recipe/retool/retool.py): dataframe.map(self.map_fn2, num_proc=16) fails with TypeError: cannot pickle '_contextvars.ContextVar' object because self.map_fn2 is an instance method that carries unpicklable context vars during multiprocessing serialization. This is already fixed upstream in verl-recipe by commit b07739a (PR #67) — the fix extracts map functions to a standalone module so they are pickle-safe while keeping num_proc=16. The verl submodule reference just needs to be updated to include this commit. My local workaround was simply removing num_proc=16, but the upstream fix is better.

vLLM (vllm/platforms/rocm.py): When vLLM is imported in a CPU-only Ray actor process (e.g., coordinator/scheduler with no GPU assigned), _get_gcn_arch() falls through all detection methods to torch.cuda.get_device_properties("cuda") which raises RuntimeError: No HIP GPUs are available. My fix is a one-line try-except RuntimeError guard returning "" as fallback. 

Neither of these blocks the verl changes in this PR.

verl/trainer/constants_ppo.py
xiaohong42: `HSA_NO_SCRATCH_RECLAIM=1` is required by AMD RCCL (ROCm Communication Collectives Library) on MI300X GPUs. Without it, FSDP initialization fails with:

```
torch.distributed.DistBackendError: NCCL error in:
  /pytorch/torch/csrc/distributed/c10d/ProcessGroupNCCL.cpp:3109,
  unhandled system error (ncclSystemError)
```

With `NCCL_DEBUG=INFO` enabled, the underlying error is:
```
[FATAL ERROR]: HSA_NO_SCRATCH_RECLAIM=1 must be set to avoid performance
degradation with the current HIP configuration.
```

This is an AMD HSA (Heterogeneous System Architecture) runtime variable that controls GPU scratch memory reclamation behavior. It is silently ignored on non-AMD platforms (NVIDIA, Ascend), similar to the existing `HCCL_*` variables above which are Ascend-specific.

wuxibin89 (approved):

PR #6082, opened 2026-04-21
[misc] feat: upgrade tensordict to fix nested tensor _ragged_idx problem

vermouth1992:
### What does this PR do?

- As title

### Checklist Before Starting

  - `{modules}` include `fsdp`, `megatron`, `veomni`, `sglang`, `vllm`, `vllm_omni`, `rollout`, `trainer`, `ci`, `training_utils`, `recipe`, `hardware`, `deployment`, `ray`, `worker`, `single_controller`, `misc`, `perf`, `model`, `algo`, `env`, `tool`, `ckpt`, `doc`, `data`, `cfg`, `reward`, `fully_async`, `one_step_off`

### Test

### API and Usage Example

```python
```

### Design & Code Changes

> [!IMPORTANT]

### DIFF
--- a/verl/utils/tensordict_utils.py
+++ b/verl/utils/tensordict_utils.py
@@ -885,7 +885,10 @@ def maybe_fix_3d_position_ids(data: TensorDict):
     # will incur indexing error for ragged tensor. This only happens when using 3D position ids in VLMs.
     # This is likely a bug in tensordict. As a workaround, we manually set _ragged_index.
     if "position_ids" in data.keys() and data["position_ids"].dim() == 3 and data["position_ids"].is_nested:
-        data["position_ids"]._ragged_idx = 2
+        assert parse_version(tensordict.__version__) >= parse_version("0.12.2"), (
+            "_ragged_idx bugs have been fixed in tensordict version 0.12.2, please upgrade tensordict"
+        )
+        # data["position_ids"]._ragged_idx = 2
 
 
 def list_of_dict_to_tensordict(list_of_dicts: list[dict[str, Any]]) -> TensorDict:
--- a/.github/workflows/e2e_ascend.yml
+++ b/.github/workflows/e2e_ascend.yml
@@ -92,6 +92,7 @@ jobs:
       - name: Install the current repository
         run: |
           pip install --no-deps -e .
+          pip install "tensordict>=0.12.2"
       - name: Check final pip list
         run: |
           pip list
--- a/.github/workflows/e2e_ascend.yml
+++ b/.github/workflows/e2e_ascend.yml
@@ -92,7 +92,7 @@ jobs:
       - name: Install the current repository
         run: |
           pip install --no-deps -e .
-          pip install "tensordict>=0.12.2"
+          pip install --index-url https://pypi.org/simple "tensordict>=0.12.2"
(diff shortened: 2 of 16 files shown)


vermouth1992:
@wuxibin89 please help check whether NPU has issues after upgrading. Thanks!

verl/utils/tensordict_utils.py
```diff
@@ -885,7 +885,10 @@ def maybe_fix_3d_position_ids(data: TensorDict):
     # will incur indexing error for ragged tensor. This only happens when using 3D position ids in VLMs.
     # This is likely a bug in tensordict. As a workaround, we manually set _ragged_index.
     if "position_ids" in data.keys() and data["position_ids"].dim() == 3 and data["position_ids"].is_nested:
-        data["position_ids"]._ragged_idx = 2
+        assert parse_version(tensordict.__version__) >= parse_version("0.12.2"), (
+            "_ragged_idx bugs have been fixed in tensordict version 0.12.2, please upgrade tensordict"
+        )
+        # data["position_ids"]._ragged_idx = 2
```
wuxibin89, line 891: SFT dataset should be adjust as well.
https://github.com/verl-project/verl/blob/main/verl/utils/dataset/dataset_utils.py#L79

ErenAta16 (commented):
Two things about the constraint change, separate from whether 0.12.2 is the right floor.

**The upper bound goes away.** The three declaration sites move from a bounded range to an open one:

```
requirements.txt      tensordict>=0.8.0,<=0.10.0,!=0.9.0  ->  tensordict>=0.12.2
requirements-npu.txt  same
setup.py              same
```

The old constraint was not a lazy pin. It carried an upper cap *and* an explicit `!=0.9.0`, which is what a project writes after a specific release broke it. This PR exists because a tensordict version shipped the `_ragged_idx` problem, so the history says releases in this dependency do break verl, and the change removes the mechanism that contained that.

Concretely, the next tensordict minor lands in every fresh install and every CI run with no change to this repo, and if it regresses nested tensors the failure appears on unrelated PRs first. Given the diff already touches eight workflows to force a version, an upper bound seems more in keeping with how this dependency has actually behaved. `>=0.12.2,<0.13` would keep the fix and keep the containment.

If dropping the cap is deliberate, for instance because the cap was what made this upgrade painful in the first place, that is a reasonable trade, it just belongs in the description rather than being inferred from the diff.

**The version now lives in eleven places.** Three declaration files plus a `pip install --index-url https://pypi.org/simple "tensordict>=0.12.2"` line added to eight workflows. The next bump has to hit all eleven, and a missed workflow fails in a way that looks like a flaky job rather than a stale pin. A constraints file referenced by the workflows, or a single composite action, would leave one place to edit.

Worth a second look at `--index-url` specifically. It *replaces* the index set rather than adding to it, so any workflow that relies on a configured mirror or an internal index loses it for that install. That is most likely to matter in `e2e_ascend.yml` and `e2e_ppo_trainer_megatron_vllm_2_ascend.yml`, where NPU wheels commonly come from somewhere other than pypi.org. `--extra-index-url` would add pypi without displacing whatever is configured, if the intent is just to make sure a recent tensordict is reachable.

Neither is blocking. The upgrade itself is clearly needed.

PR #6091, opened 2026-04-21
[rollout,vllm] feat: split large weight into chunks in NCCL/NIXL checkpoint engine

wuxibin89:
### What does this PR do?

Before this PR, NCCL/NIXL checkpoint engine and vLLM CUDA IPC buffer size (controlled by `rollout.checkpoint_engine.update_weights_bucket_megabytes`) must be larger than the largest weight. So during weight synchronization, each GPU must has additional memory:
- Colocated: all-gather full weight + cuda ipc buffer, `max_weight_size * 2`
- Fully async:  double buffer(NCCL/NIXL checkpoint engine) + cuda ipc buffer, `max_weight_size * 3`

While transformers>5, the `gate_up_proj` has been fused into single weight with shape `[num_experts, 2 * intermediate_dim, hidden_dim]`. For example, Qwen/Qwen3.5-397B-A17B gate_up_proj size is 8GB in bfloat16. This cause huge peak memory during weight synchronization.

This PR introduce two changes
1. CUDA IPC buffer size can be smaller than largest weight. If weight size is larger than cuda ipc buffer size, bypass buffer and directly send it, reducing peak memory from `max_weight_size * 2` to `max_weight_size + cuda_ipc_buffer_size`.
2.  NCCL/NIXL checkpoint engine send/recv bucket size can be smaller than largest weight. If weight size is larger than bucket size, split it into chunks in sender and merge it from chunks in receiver, reducing peak memory from `max_weight_size * 3` to `max_weight_size + 3 * cuda_ipc_buffer_size`

While the peak memory is still bounded by `max_weight_size`, we will address it in future PR.

### DIFF
--- a/tests/checkpoint_engine/test_correctness_on_gpu.py
+++ b/tests/checkpoint_engine/test_correctness_on_gpu.py
@@ -40,6 +40,7 @@ async def test_nccl_checkpoint_engine(
     num_rollout,
     num_nodes=1,
     num_gpus_per_node=_ngpus,
+    bucket_size_mb=128,
     check_allclose=True,
     model_path="~/models/Qwen/Qwen3-8B-Base",
 ):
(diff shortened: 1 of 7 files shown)


ISEEKYAN:
after each shard is sent, can the shard memory be released in the training side or the shard consumed immediately on the rollout side? if not, when we use colocated setting, the peak memory would be at least 2x of single max parameter, is that true?

wuxibin89:
> after each shard is sent, can the shard memory be released in the training side or the shard consumed immediately on the rollout side? if not, when we use colocated setting, the peak memory would be at least 2x of single max parameter, is that true?

Small weights are accumulated in a pre-allocated cuda ipc buffer. While for large weight that can't fit into the buffer, it's directly send via cuda ipc and consumed by vllm immediately, then the weight is garbage collected and memory released. So the peak memory is 1x max_weight_size plus an additional cuda ipc buffer.

ArronHZG (approved):

PR #6110, opened 2026-04-22
[fsdp, training_utils] feat: support per-sample temperature with fused linear cross entropy

KimmiShi:
## Summary
Currently `use_fused_kernels=True` requires a scalar `temperature`, which blocks RL training setups that need per-sample temperatures (e.g. teacher/distillation mixing, mixed data). This PR makes both fused implementations accept either a Python scalar (unchanged) or a 1-D per-token tensor, without regressing the scalar fast path.

The approach is to pre-scale the hidden states by `1/T` in fp32 before the fused matmul and then run the kernel with `temperature=1.0`. Since softmax and logsumexp are invariant to a constant multiplicative factor on logits, this is mathematically equivalent to the `logits / T -> cross_entropy` baseline and keeps the kernel's memory and compute profile essentially unchanged.

## Changes
- `verl/utils/torch_functional.py`: new `prescale_hidden_by_per_sample_temperature` and `resolve_temperature_for_fused_kernel` helpers.
- `verl/utils/kernel/linear_cross_entropy.py` (Triton): accept `Union[float, Tensor]`, dispatch via helper.
- `verl/utils/experimental/torch_functional.py` (`FusedLinearForPPO`): same dispatch.
- `verl/workers/engine/fsdp/transformer_impl.py`: forward `temperature_rmpad` to the fused kernel when `use_remove_padding=True`; keep scalar-only assertion on the non-remove-padding path.
- `tests/utils/test_linear_cross_entropy.py`: add uniform-tensor-equals-scalar and per-sample-vs-fp32-reference tests (fwd + bwd) for both Triton and PyTorch fused paths. Also fixes a pre-existing `TypeError` when `MAX_TEST_CASES` is set from the shell (stringified int).

## Tests
`pytest tests/utils/test_linear_cross_entropy.py -v`

### DIFF
--- a/tests/utils/test_linear_cross_entropy.py
+++ b/tests/utils/test_linear_cross_entropy.py
@@ -358,9 +358,7 @@ class TestLinearCrossEntropy:
 
         for _ in range(iterations):
             hidden, weight, labels = self.generate_forward_inputs()
-            t_tensor = torch.full(
-                (self.num_tokens,), self.temperature, dtype=torch.float32, device="cuda"
-            )
+            t_tensor = torch.full((self.num_tokens,), self.temperature, dtype=torch.float32, device="cuda")
 
             # torch fused path runs in fp32 so the prescale cast is a no-op.
             # Residual difference (~1e-5) comes from the matmul associativity
@@ -368,14 +366,10 @@ class TestLinearCrossEntropy:
             # evaluates (h / T) @ W.T, which accumulates in a different order
             # and differs by a few ULPs per-token. Match the cross-implementation
             # tolerance used for verl_fused vs. torch further down.
-            lp_scalar, ent_scalar = run_verl_torch_fused_entropy(
-                hidden, weight, labels, self.temperature
-            )
+            lp_scalar, ent_scalar = run_verl_torch_fused_entropy(hidden, weight, labels, self.temperature)
             hidden_f32 = hidden.to(torch.float32)
             weight_f32 = weight.to(torch.float32)
-            lp_tensor, ent_tensor = fused_linear_for_ppo(
-                hidden_f32, weight_f32, labels, temperature=t_tensor
-            )
+            lp_tensor, ent_tensor = fused_linear_for_ppo(hidden_f32, weight_f32, labels, temperature=t_tensor)
             lp_tensor = lp_tensor.squeeze(0)
             ent_tensor = ent_tensor.squeeze(0)
             torch.testing.assert_close(lp_scalar, lp_tensor, atol=1e-4, rtol=1e-4)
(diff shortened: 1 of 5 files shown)


ETOgaosion:
@KimmiShi Nice contribution! I think that it's better provide 1 unified API for both single temperature and per-sample temperature, for example, we always use torch.Tensor for temperature arg, and the float occupies only 1 element.

KimmiShi:
> @KimmiShi Nice contribution! I think that it's better provide 1 unified API for both single temperature and per-sample temperature, for example, we always use torch.Tensor for temperature arg, and the float occupies only 1 element.

@ETOgaosion Thanks for the review! I'd like to confirm the intent, since there are a couple of design points hidden here.

The reason temperature is currently Union[float, Tensor] is that the underlying fused kernels (both Triton and the torch path) only accept a scalar temperature and do logits / T internally. To support per-sample temperature without rewriting the kernels, I prescale hidden by 1/T (fp32) only in the tensor-with-many-elements case and call the kernel with temperature=1.0; the scalar case stays on the kernel's native path, which avoids one extra bf16 rounding on hidden and an extra hidden-sized copy.

So unifying to torch.Tensor could mean two different things:

- Unify only the public signature — always take a torch.Tensor, but internally still short-circuit to the native scalar path when it's 1-element/uniform. Keeps the scalar fast path's precision & perf, but an internal branch remains.
- Unify the whole execution path — always prescale hidden and run the kernel at temperature=1.0. Removes all branching, but slightly changes scalar numerics (extra bf16 rounding) and always pays the hidden copy.

Could you clarify which one you're after?

ETOgaosion:
@KimmiShi I understand the constraints. I think `temperature: Union[float, torch.Tensor]` is acceptable.

ETOgaosion:
Strangely, CI may got some problems, could you check whether the doc of per-sample temperature is still missing? Add some docs and re-push may trigger CI successfully~

KimmiShi:
> Strangely, CI may got some problems, could you check whether the doc of per-sample temperature is still missing? Add some docs and re-push may trigger CI successfully~

@ETOgaosion Doc updated and it needs approval to trigger CI.

ETOgaosion:
@KimmiShi Could you look further into problem in GPU unit test?

```
FAILED tests/utils/test_linear_cross_entropy.py::test_lce_per_sample_temperature - torch._dynamo.exc.InternalTorchDynamoError: AttributeError: 'SymNodeVariable' object has no attribute 'value'
```

KimmiShi:
> @KimmiShi Could you look further into problem in GPU unit test?
> 
> ```
> FAILED tests/utils/test_linear_cross_entropy.py::test_lce_per_sample_temperature - torch._dynamo.exc.InternalTorchDynamoError: AttributeError: 'SymNodeVariable' object has no attribute 'value'
> ```

@ETOgaosion Fixed the torch compile error introduced by "exclude resolve_temperature_for_fused_kernel from torch.compile"

PR #6147, opened 2026-04-26
fix: handle missing run_headless in vllm <= 0.8.2 to prevent ImportError

wengfulin:
## Description

Currently, the dev branch imports `run_headless` from `vllm.entrypoints.cli.serve`. However, this function does not exist in `vLLM 0.8.2` and earlier versions, which causes a fatal `ImportError` for users running stable or slightly older environments. 

This PR wraps the import in a `try-except` block to provide backward compatibility (or a clearer fallback mechanism) for environments using `vLLM <= 0.8.2`.

## Motivation and Context
While the documentation suggests using `vLLM 0.8+`, `vLLM 0.8.2` is a commonly used version that currently breaks the `verl` pipeline entirely due to this missing API. 

**Error Traceback:**
`ImportError: cannot import name 'run_headless' from 'vllm.entrypoints.cli.serve'`

## How Has This Been Tested?
- [x] Environment: PyTorch 2.5.1 + vLLM 0.8.2 / vLLM 0.7.3
- [x] Hardware: NVIDIA V100 (Tested under standard PPO/GRPO workflow)
- [x] Result: The pipeline no longer crashes during the rollout initialization.

### DIFF
--- a/run_grpo.sh
+++ b/run_grpo.sh
@@ -1,0 +1,36 @@
+#!/bin/bash
+
+export PYTHONNOUSERSITE=1
+export PYTHONUNBUFFERED=1
+ray stop -f
+
+python -m verl.trainer.main_ppo \
+    data.train_files=$HOME/data/gsm8k/train.parquet \
+    data.val_files=$HOME/data/gsm8k/test.parquet \
+    data.train_batch_size=256 \
+    data.val_batch_size=16 \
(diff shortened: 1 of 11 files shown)


wuxibin89:
vllm<=0.8.2 is too old, we're not going to maintain backward compatibility for it.

PR #6148, opened 2026-04-26
[model, cfg, hardware] feat: add attn_implementation to HFModelConfig for non-Ampere GPU compatibility

wengfulin:
### What does this PR do?

This PR introduces the attn_implementation field to HFModelConfig.

The Problem: Currently, HFModelConfig does not accept the attn_implementation argument. This makes it impossible to override the default attention backend via Hydra configs or CLI. On non-Ampere GPUs (e.g., NVIDIA V100, T4, P40), the model initialization often defaults to FlashAttention2, which is not supported on these architectures, leading to a fatal crash.

The Solution: By adding attn_implementation to HFModelConfig and including it in _mutable_fields, users can now explicitly specify a compatible attention backend (e.g., sdpa or eager) to bypass hardware-specific errors.

### Checklist Before Starting

[x] Search for similar PRs. Related query: verl FlashAttention hardware compatibility

[x] Format the PR title as [{modules}] {type}: {description}.

[x] Modules: model, cfg, hardware. Type: feat.

### Test

Environment:

Hardware: NVIDIA V100 (Volta).

Software: PyTorch 2.5.1 + vLLM.

Validation:

Before Fix: Running the pipeline with ++actor_rollout_ref.model.attn_implementation=sdpa threw TypeError: HFModelConfig.__init__() got an unexpected keyword argument 'attn_implementation'.

After Fix: The pipeline successfully recognizes the argument, fallbacks to the sdpa backend, and completes model weight loading without crashing.

Modified HFModelConfig: Added attn_implementation: Optional[str] = None to the dataclass fields.

Updated _mutable_fields: Added "attn_implementation" to the list of mutable fields to ensure it is not filtered out during the worker/model instantiation process.

HuggingFace Integration: This change allows the parameter to be correctly passed to AutoModelForCausalLM.from_pretrained via the internal configuration mapping.

Checklist Before Submitting
[x] Read the [Contribute Guide](https://github.com/verl-project/verl/blob/main/CONTRIBUTING.md).

[x] Apply [pre-commit checks](https://github.com/verl-project/verl/blob/main/CONTRIBUTING.md#code-linting-and-formatting).

[x] Verified hardware compatibility on V100.


ETOgaosion:
Please use `override_config` API in HFModelConfig , and since the rollout and training engine can hardly provide flash attn support on Volta, we can do tiny support~ qwq

PR #6228, opened 2026-04-30
[reward, trainer] feat: support multi-output trajectories in async reward scoring

guillemgt:
### What does this PR do?

When using the TransferQueue trainer with multi-turn agent loops, each prompt may produce a list of `AgentLoopOutput`s. Previously `_compute_score` was called only on the final output, constructing a batch of size 1. This change passes all outputs to `_compute_score` so reward managers can be extended to score intermediate turns, while the default behaviour (scoring only the last output via `data[-1:]`) is preserved for all existing reward managers.

Also fixes a pre-existing bug in `AgentLoopWorkerTQ._agent_loop_postprocess` where `_compute_teacher_logprobs` was called with the raw `output` parameter (which may be a `list[AgentLoopOutput]`) instead of the resolved `final_output`.

### Checklist Before Starting

- [x] Search for similar PRs. Paste at least one query link here: https://github.com/verl-project/verl/pulls?q=is%3Apr+is%3Aopen+multi-output+async+reward+scoring

### Test

Manually ran trainings with:
- `main_ppo` with naive reward manager
- `main_ppo_sync` with single-output rollouts and naive reward manager
- `main_ppo_sync` with multi-output rollouts and a custom reward manager (that combines the outputs into a single sequence in a non-trivial way, and then delegates to the naive reward manager)

All runs could train for a few steps and compute validation without problems.

This behaviour is also covered by current end-to-end tests.

### API and Usage Example

```python
# Before
await self._compute_score(output, prompts=..., responses=..., attention_mask=..., input_ids=..., position_ids=..., kwargs=kwargs)

# After — outputs is list[AgentLoopOutput]
await self._compute_score(outputs, kwargs=kwargs)
```

### Design & Code Changes

- `_compute_score` now accepts `list[AgentLoopOutput]` and internally derives tensors from `outputs[-1]`
- All built-in reward managers (`naive`, `dapo`, `gdpo`, `limited`, `remote`) use `data[-1:]` instead of asserting `len(data) == 1`
- `RewardLoopWorker.compute_score` removes the single-item assertion
- `AgentLoopWorkerTQ._agent_loop_postprocess` simplified and the `_compute_teacher_logprobs` bug fixed
- Updated `docs/advance/reward_loop.rst` to reflect the new interface

- [x] Apply [pre-commit checks](https://github.com/verl-project/verl/blob/main/CONTRIBUTING.md#code-linting-and-formatting).
- [x] Add unit or end-to-end test(s) to [the CI workflow](https://github.com/verl-project/verl/tree/main/.github/workflows) to cover all the code. If not feasible, explain why: covered by existing end-to-end tests.
- [x] If your PR is related to the `recipe` submodule, please also update the reference to the submodule commit via `git submodule update --remote`: not applicable.

### DIFF
--- a/verl/experimental/agent_loop/agent_loop.py
+++ b/verl/experimental/agent_loop/agent_loop.py
@@ -607,15 +607,7 @@ class AgentLoopWorker:
 
         multi_modal_inputs = self._compute_multi_modal_inputs(output, input_ids)
         position_ids = self._compute_position_ids(input_ids, attention_mask, multi_modal_inputs)
-        await self._compute_score(
-            output,
-            prompts=prompt_output["input_ids"],
-            responses=response_output["input_ids"],
-            attention_mask=attention_mask,
-            input_ids=input_ids,
-            position_ids=position_ids,
-            kwargs=kwargs,
-        )
+        await self._compute_score([output], kwargs=kwargs)
         await self._compute_teacher_logprobs(
             output,
             prompt_ids=output.prompt_ids,
@@ -726,27 +718,34 @@ class AgentLoopWorker:
         position_ids = torch.cat((text_position_ids, vision_position_ids), dim=1)  # (1, 4, seq_length)
         return position_ids
 
-    async def _compute_score(self, output, prompts, responses, attention_mask, input_ids, position_ids, kwargs):
-        """Compute reward score for single sample."""
+    async def _compute_score(self, outputs: list, kwargs: dict) -> None:
+        """Compute reward score for a list of agent loop outputs (scores outputs[-1])."""
         enable_async_reward = self.reward_loop_worker_handles is not None
 
-        if output.reward_score is None and enable_async_reward:
+        final_output = outputs[-1]
+        if final_output.reward_score is None and enable_async_reward:
             timing = {}
             with simple_timer("compute_score", timing):
+                prompts = torch.tensor(final_output.prompt_ids, dtype=torch.int64).unsqueeze(0)
+                responses = torch.tensor(final_output.response_ids, dtype=torch.int64).unsqueeze(0)
+                input_ids = torch.cat([prompts, responses], dim=1)
+                attention_mask = torch.ones_like(input_ids)
+                multi_modal_inputs = self._compute_multi_modal_inputs(final_output, input_ids.squeeze(0))
+                position_ids = self._compute_position_ids(input_ids, attention_mask, multi_modal_inputs)
                 batch = TensorDict(
                     {
-                        "prompts": prompts,  # [1, prompt_length]
-                        "responses": responses,  # [1, response_length]
-                        "attention_mask": attention_mask,  # [1, prompt_length + response_length]
-                        "input_ids": input_ids,  # [1, prompt_length + response_length]
+                        "prompts": prompts,
+                        "responses": responses,
+                        "attention_mask": attention_mask,
+                        "input_ids": input_ids,
                         "position_ids": position_ids,
                     },
                     batch_size=1,
                 )
                 non_tensor_batch = {
                     **{k: np.array([v]) for k, v in kwargs.items()},
-                    "__num_turns__": np.array([output.num_turns]),
-                    "tool_extra_fields": np.array([output.extra_fields], dtype=object),
+                    "__num_turns__": np.array([final_output.num_turns]),
+                    "tool_extra_fields": np.array([final_output.extra_fields], dtype=object),
                 }
 
                 data = DataProto(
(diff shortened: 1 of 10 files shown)


verl/experimental/agent_loop/agent_loop.py
```diff
@@ -726,27 +718,34 @@ def _compute_position_ids(self, input_ids, attention_mask, multi_modal_inputs) -
...
+        final_output = outputs[-1]
+        if final_output.reward_score is None and enable_async_reward:
             timing = {}
             with simple_timer("compute_score", timing):
+                prompts = torch.tensor(final_output.prompt_ids, dtype=torch.int64).unsqueeze(0)
+                responses = torch.tensor(final_output.response_ids, dtype=torch.int64).unsqueeze(0)
+                input_ids = torch.cat([prompts, responses], dim=1)
+                attention_mask = torch.ones_like(input_ids)
+                multi_modal_inputs = self._compute_multi_modal_inputs(final_output, input_ids.squeeze(0))
+                position_ids = self._compute_position_ids(input_ids, attention_mask, multi_modal_inputs)
```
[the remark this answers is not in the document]
guillemgt: Should be fixed now. final_output is used only in order to set the computed reward as the final reward, which the rest of the (pre-existing) code will be treated as a reward for the whole rollout.

verl/trainer/main_ppo_sync.py
```diff
@@ -360,32 +360,13 @@ async def _agent_loop_postprocess(
...
         await self._compute_teacher_logprobs(
-            output,
-            prompt_ids=output.prompt_ids,
-            response_ids=output.response_ids,
+            final_output,
+            prompt_ids=final_output.prompt_ids,
+            response_ids=final_output.response_ids,
             validate=validate,
             sample_kwargs=kwargs,
         )
```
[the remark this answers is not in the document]
guillemgt: There is a pre-existing comment above indicating that supporting multiple outputs is on the to-do list. This is therefore outside the scope of this PR

wuxibin89:
@guillemgt  Could you also help to update `docs/advance/reward_loop.rst`?

guillemgt:
@wuxibin89 I've updated the documentation

verl/experimental/reward_loop/reward_loop.py
```diff
@@ -144,15 +144,14 @@ async def compute_score_batch(self, data: DataProto) -> list[dict]:
         return outputs
 
     async def compute_score(self, data: DataProto) -> dict:
-        assert len(data) == 1, "RewardLoopWorker only support single data item"
         if self.config.reward.custom_reward_function.path is not None:
             # directly use user-customized reward function
             return await self.reward_manager.run_single(data)
```
wuxibin89, line 149: @yyDing1 will this change break user custom_reward_function? since they may expect only one sample.

guillemgt: With non-custom reward managers it should be fine, since [they will consider only the last output](https://github.com/verl-project/verl/pull/6228/changes#diff-9d6b9cb7351da6cf2913b8dd90a044c9f8826ba69844cbc359236091bc7d50a9R76) (same behaviour as in current `main`)

yyDing1: Would it be better to call `compute_score_batch` rather than `compute_score` in the agent loop, so that the code in `reward manager` is not modified?

verl/experimental/agent_loop/agent_loop.py
```diff
@@ -755,9 +771,9 @@ async def _compute_score(self, output, prompts, responses, attention_mask, input
                 )
                 selected_reward_loop_worker_handle = random.choice(self.reward_loop_worker_handles)
                 result = await selected_reward_loop_worker_handle.compute_score.remote(data)
```
yyDing1, line 773: I mean `result = await selected_reward_loop_worker_handle.compute_score_batch.remote(data)` here.

verl/experimental/reward_loop/reward_loop.py
guillemgt: Calling `compute_score_batch` would mean that we are scoring each output _independently_ of others. Each output within the same rollout could have a different reward, which would complicate the advantage estimation in methods like GRPO.
I chose to modify reward managers because that would allow reward functions that need to consider all outputs for a rollout at once (without assuming independence).

verl/experimental/agent_loop/agent_loop.py
yyDing1: or `result = await selected_reward_loop_worker_handle.compute_score.remote(data[-1:])`?

verl/experimental/reward_loop/reward_loop.py
guillemgt: For example, the method in [this paper (Algorithm 1 on page 4)](https://arxiv.org/pdf/2505.20622) (minus the per-batch standardisation of rewards) uses a reward function that needs all outputs in a rollout.

verl/experimental/agent_loop/agent_loop.py
guillemgt: One could do that (in the first case summing the different rewards) but there would be reward functions that aren't possible with this. I have included more details in the other thread, with an example that's possible in the current implementation in the PR by using a custom reward manager.

wuxibin89 (dismissed):

wuxibin89:
@guillemgt cpu_unit_tests ci failed

guillemgt:
@wuxibin89 should be fixed now

yyDing1 (approved):

PR #5805, opened 2026-03-30
[ray_trainer] fix: honor total_training_steps after resume in step mode

chenshui223:
## Summary
- expand the effective epoch budget when step-based PPO training needs more epochs than `trainer.total_epochs`
- compute resume epoch bounds from `global_steps` so checkpoint resume can continue across epoch boundaries
- emit a warning when the trainer exits before reaching `total_training_steps`
- add CPU tests covering both step-mode epoch expansion and resume-at-mid-epoch / epoch-boundary cases

## Testing
- pytest -q tests/trainer/ppo/test_ray_trainer_epoch_budget_on_cpu.py

Closes #5013

PR #6232, opened 2026-05-01
[trainer] fix: skip extra step when resume already exceeds total_training_steps

startju:
## Summary

After a checkpoint that reached `total_training_steps`, restarting `fit()` runs one extra step before exiting: the pre-loop `self.global_steps += 1` advances the resumed `N=total` to `N+1`, the loop body still runs once before the in-loop `is_last_step` check fires, and the run leaks `global_step_{N+1}`. Each subsequent restart leaks another step.

**Concrete repro**: train with `total_training_steps=5`, finish, restart the same script. Trainer resumes from `global_step_5`, `+=1` to 6, runs step 6, exits. Re-run: resumes from `step_6`, `+=1` to 7, runs step 7. Each restart leaks one extra training step.

**Fix**

- Promote `is_last_step` to a `@property` returning `global_steps >= total_training_steps` on `RayPPOTrainer` and `PPOTrainer`. `FullyAsyncTrainer` overrides for `total_train_steps` (with a `None`-check for the pre-`set_total_train_steps` window).
- Add a pre-loop guard `if self.is_last_step: return` **before** the pre-loop `+=1`, so a checkpoint at the budget exits without running another step. The log surfaces the actual `global_steps` / `total_training_steps`.
- Reorder the end-of-iter `if is_last_step: cleanup; return` to run **before** the in-loop `+=1`, so the same `>=` predicate is correct at every read site without a local snapshot.
- Drop the `self.is_last_step = ...` writes (init + `_fit_prepare_step` snapshot) in the experimental trainers. The property is now the single source of truth; their outer loops snapshot `is_last_step` once per iteration before delegating to `fit_step` so the cleanup decision uses the pre-`+=1` frame even though the increment itself is inside `_fit_postprocess_step`.

Side effect: when `fit()` returns normally `self.global_steps` lands on `total_training_steps` instead of `total_training_steps + 1`. That also removes the off-by-one cursor the original bug worked around — a fresh resume from the saved value now hits the pre-loop guard cleanly.

## Why this isn't duplicating #5805

#5805 (`honor total_training_steps after resume in step mode`) addresses the **opposite** direction: training stopping **early** because the configured `total_epochs` budget is too small to reach `total_training_steps`. It expands the epoch budget and warns on early exit. It does **not** change the pre-/end-of-iter `+=1` placement, so the overshoot bug fixed here remains even with #5805 applied.

Both fixes are orthogonal and complementary.

## Scope

Files updated:
- `verl/trainer/ppo/ray_trainer.py` — main upstream PPO trainer (property + pre-loop guard + cleanup reorder)
- `verl/trainer/main_ppo_sync.py` — sync PPO entry-point trainer (same shape)
- `verl/experimental/separation/ray_trainer.py`
- `verl/experimental/one_step_off_policy/ray_trainer.py`
- `verl/experimental/fully_async_policy/fully_async_trainer.py`

`recipe/*` contains many additional trainer copies that share the same fit-loop pattern; those are intentionally left out of this PR for separate review since each recipe has its own fork of the loop logic.

## Test plan

- Added `tests/trainer/ppo/test_is_last_step_on_cpu.py` covering:
  - the `is_last_step` predicate at every read site (cold start, mid-run, penultimate, last step, overshoot, zero-budget) for both trainer hierarchies
  - `FullyAsyncTrainer`'s override against `total_train_steps` and the `None`-check for the pre-init window
  - the read-only contract (no setter; restoring one would let stale snapshots silently shadow the live predicate at the cleanup check)
  - the inheritance / override layout (subclasses must inherit; only `FullyAsyncTrainer` overrides)
- `pytest tests/trainer/ -q` → **100 passed** (77 existing + 23 new), no regressions.
- Smoke-imported `OneStepOffRayTrainer`, `SeparateRayPPOTrainer`, `FullyAsyncTrainer`, and `verl.trainer.main_ppo_sync` to confirm the property-only design resolves cleanly with no leftover references to the removed helper.

## AI assistance disclosure

This change was prepared with assistance from Anthropic Claude. The submitter has reviewed every changed line and run the test suite locally.


verl/trainer/main_ppo_sync.py
ArronHZG: Can the design of `is_last_step` be reused?
For example, can `is_last_step` be calculated in advance and the process exited?

startju: Thanks for the review — pushed a revision that follows your suggestion. `is_last_step` is now a `@property` on the trainer base classes (`global_steps >= total_training_steps`); `FullyAsyncTrainer` overrides for `total_train_steps`. The pre-loop guard now reads it **before** the pre-loop `+=1` so a checkpoint that already reached the budget exits without running another step, and the end-of-iter cleanup-return is reordered to also read the property **before** the in-loop `+=1`. With both reads in the pre-`+=1` frame the same `>=` predicate is correct at every site, so the standalone `should_exit_training_loop` helper is no longer needed and the local `is_last_step` snapshot in `fit()` is gone.

Side effect: when `fit()` returns normally `self.global_steps` lands on `total_training_steps` instead of `total_training_steps + 1`, which removes the off-by-one cursor the original bug was working around — a fresh resume from the saved value now hits the guard cleanly.

In the experimental trainers (`SeparateRayPPOTrainer`, `OneStepOffRayTrainer`, `FullyAsyncTrainer`) the `self.is_last_step = ...` init and `_fit_prepare_step` snapshot writes are dropped; the outer loops snapshot `is_last_step` once per iteration before delegating to `fit_step` so the cleanup decision still uses the pre-`+=1` frame even though the increment itself is inside `_fit_postprocess_step`.

Tests in `tests/trainer/ppo/test_is_last_step_on_cpu.py` cover the predicate at every read site (cold start, mid-run, penultimate, last, overshoot, zero-budget), the `FullyAsyncTrainer` `None` path, and the read-only / inheritance / override contract. `pytest tests/trainer/ -q` → 100 passed (77 existing + 23 new).

startju:
@wuxibin89 could you kindly re-approve CI when you have a moment? Thanks!

The previous run on `a6aa46f4` had four code-side failures and a few CI-infra ones; this push (`6d6224c9`) addresses the former:

- `pre-commit` / `pre_commit_for_ppo` — added missing docstrings on `RayPPOTrainer.is_last_step` / `PPOTrainer.is_last_step` and removed an extra blank line in `one_step_off_policy/ray_trainer.py`.
- `check-title` — `[ray_trainer]` is not in the allowed module list; renamed to `[trainer]` (and aligned the commit subject).

Already verified locally: `pre-commit run --files ...` clean, `pytest tests/trainer/ -q` 100 passed.

The other reds (`e2e_ppo_trainer_megatron-deepseek` with `EADDRINUSE`, the four Ascend jobs that died in 1–2 min during runner setup, `model_rmpad_fsdp2_unstable_ascend`) all looked like CI infra hiccups rather than this PR — `e2e_one_step_off_policy_fsdp2_ascend`, `e2e_fully_async_policy_fsdp2_ascend`, and `e2e_ppo_trainer_fsdp_vllm_ascend` succeeded in the same batch. Hopefully they pass on rerun.

startju:
@wuxibin89 thanks for re-approving! Down to two reds now and both look like Ascend infra flakes rather than this PR:

- `e2e_one_step_off_policy_megatron_ascend` — `NPU out of memory. Tried to allocate 1.00 GiB ... 152.31 MiB free` ([run](https://github.com/verl-project/verl/actions/runs/25602386884))
- `e2e_ascend` (MindSpeed_LLM) — `HCCL communicator ... wait timeout after 1800000ms` ([run](https://github.com/verl-project/verl/actions/runs/25602386868))

The GPU equivalents and the other Ascend jobs (`e2e_one_step_off_policy_fsdp2`, `e2e_one_step_off_policy_megatron`, `e2e_one_step_off_policy_fsdp2_ascend`, `e2e_fully_async_policy_fsdp2_ascend`, `e2e_ppo_trainer_fsdp_vllm_ascend`, etc.) all passed in the same batch. Could you re-run just these two when convenient? Thanks!

PR #6251, opened 2026-05-06
[trainer] fix: write request_id to reward_extra_infos_to_dump instead of reward_extra_infos_dict

boundless-future:
## What does this PR do?

Fixes #6250

In `_log_rollout_data`, `request_id` was being written to `reward_extra_infos_dict` instead of `reward_extra_infos_to_dump`. Since `reward_extra_infos_to_dump` is created from `reward_extra_infos_dict` **before** the `request_id` insertion and is the dict passed to `_dump_generations()`, the `request_id` was never included in dumped rollout data.

### Change

```diff
 if "request_id" in batch.non_tensor_batch:
-    reward_extra_infos_dict.setdefault(
+    reward_extra_infos_to_dump.setdefault(
         "request_id",
         batch.non_tensor_batch["request_id"].tolist(),
     )
```

### Why this is not duplicating an existing PR

Searched upstream open PRs for `request_id`, `reward_extra_infos_dict`, and `dump_generations` — no existing PR addresses this bug.

### Test

This is a one-line variable name fix. Verified by code inspection:

1. `reward_extra_infos_to_dump` is created from `reward_extra_infos_dict` at L448-450 (before the `request_id` insertion)
2. `reward_extra_infos_to_dump` is passed to `_dump_generations()` at L465
3. Writing to `reward_extra_infos_dict` after the copy means `request_id` is never in the dump output

No unit test exists for `_log_rollout_data` dump output content. The fix is trivially correct by code inspection.

### AI Assistance

AI assistance was used to identify the bug and draft this PR. The submitter has reviewed every changed line and confirms the fix is correct.

PR #6255, opened 2026-05-06
[trainer] perf: async generation dump with exception propagation and streaming write

### LINKED ISSUE #6250: [trainer] Bug: request_id not included in dumped rollout data due to wrong dict target in _log_rollout_data
## Bug Description

In `verl/trainer/ppo/ray_trainer.py`, the `_log_rollout_data` method has a variable name typo that prevents `request_id` from being included in dumped rollout data.

### Current (Buggy) Code

```python
reward_extra_infos_to_dump = {
    k: (v.tolist() if isinstance(v, np.ndarray) else v) for k, v in reward_extra_infos_dict.items()
}
if "request_id" in batch.non_tensor_batch:
    reward_extra_infos_dict.setdefault(       # BUG: writes to wrong dict
        "request_id",
        batch.non_tensor_batch["request_id"].tolist(),
    )
```

### Expected Code

```python
reward_extra_infos_to_dump = {
    k: (v.tolist() if isinstance(v, np.ndarray) else v) for k, v in reward_extra_infos_dict.items()
}
if "request_id" in batch.non_tensor_batch:
    reward_extra_infos_to_dump.setdefault(    # FIX: write to the dump dict
        "request_id",
        batch.non_tensor_batch["request_id"].tolist(),
    )
```

### Root Cause

1. `reward_extra_infos_to_dump` is created from `reward_extra_infos_dict` **before** the `request_id` insertion.
2. The `request_id` is written to `reward_extra_infos_dict` (the source dict) instead of `reward_extra_infos_to_dump` (the dump dict).
3. `_dump_generations()` receives `reward_extra_infos_to_dump`, so `request_id` is never included in the dumped data.

### Additional Concern

If `reward_extra_infos_dict` already contains a `"request_id"` key (from reward computation), `setdefault` is a no-op and the `batch.non_tensor_batch["request_id"]` value is silently discarded. Even with the fix, the behavior may not match the author's intent if both sources provide `request_id`. This might deserve a separate review.

### Impact

- When using `rollout_data_dir` to dump generation data, `request_id` is always missing from the output files.
- This affects traceability of dumped rollout data, especially for multi-turn / agent-loop training scenarios where `request_id` is critical for tracking trajectories.

### LINKED ISSUE #6254: [Perf] Synchronous generation dump blocks training loop and silently drops I/O errors
## System Info

- verl: main
- python: 3.11+
- pytorch: 2.x

## Information

- [ ] Official example scripts
- [x] Custom modified scripts

## Task

- [ ] Official supported task in examples folder
- [x] Custom task or dataset

---

## Problem Description

`RayPPOTrainer._dump_generations` currently has three issues:

1. **Synchronous I/O blocks the training loop**: Every time generation data is dumped, `json.dumps` + `f.write` executes synchronously on the main thread. With large batches (e.g., GRPO n=16, batch_size=1024), a single dump can take several seconds, directly blocking the next rollout step in the training loop.

2. **I/O errors silently dropped**: The current write logic has no exception propagation mechanism. If disk full, permission errors, or other I/O exceptions occur, the error is swallowed or only surfaces at final `shutdown`, causing training to appear normal while data is actually lost.

3. **High peak memory**: The current approach uses `"\n".join(lines)` to build a complete string before writing, which creates unnecessary memory spikes with large numbers of samples.

## Expected Behavior

- Generation dump should execute asynchronously in a background thread, not blocking the main training loop
- I/O errors should surface early at the next `_dump_generations` call (via `future.result()` exception propagation)
- Use streaming line-by-line writes to avoid peak memory from constructing large strings
- At training end, all pending dump futures should be drained and checked to ensure no write errors are lost

## Related

- PR #5899 / Issue #5894 — Fixed a crash in `compute_data_metrics` with empty tensors, another improvement in the same module (`verl/trainer/ppo/`)
- PR #6251 / Issue #6250 — Fixed `request_id` being written to the wrong dict in `_log_rollout_data`

boundless-future:
## Summary

Fixes #6254
Closes #6250 (supersedes #6251, which is included as the first commit)

Make generation dump in `RayPPOTrainer` asynchronous with exception propagation and streaming writes, resolving:

1. **Synchronous I/O blocks the training loop** — dump runs on the main thread, blocking rollout with large batches
2. **I/O errors silently dropped** — disk full and other exceptions go unnoticed
3. **High peak memory** — `"\n".join(lines)` builds a large string before writing

## Changes

- **Extract `_write_generations` static method**: Lift the inline write logic into a `@staticmethod` so the background thread doesn't capture `self`, ensuring thread safety.
- **Async dump**: Submit write tasks via `ThreadPoolExecutor(max_workers=1)`; the main thread returns immediately without blocking training.
- **Exception propagation**: Track submitted futures in `_dump_futures`. On each call, remove completed futures and call `.result()` to surface I/O errors early.
- **Streaming write**: Replace `"\n".join(lines)` + single `f.write()` with per-line `f.write()`, reducing peak memory.
- **Drain pending futures**: Before `_dump_executor.shutdown(wait=True)`, iterate all remaining futures and call `.result()` so no write errors are lost at training end.
- **Fix `request_id` dict bug** (commit 2, same as #6251): Write `request_id` to `reward_extra_infos_to_dump` instead of `reward_extra_infos_dict`.

## Testing

- [x] Local PPO training runs correctly, generation data written to JSONL files as expected
- [x] Simulated I/O error (read-only directory): exception propagates correctly on the next `_dump_generations` call
- [x] All pending dumps complete at training end, no data loss

## Note on Training Correctness

This change does not affect training precision:
1. Generation dump is purely observational/logging — written data does not participate in loss computation
2. Async write ordering is guaranteed by `global_steps` in the filename, no confusion
3. If an I/O error occurs, the exception propagates to the main training loop — training fails rather than silently losing data

## Why this is not duplicating an existing PR

Searched open PRs for `async dump`, `dump_generations`, `ThreadPoolExecutor` — no existing PR addresses async generation dump. PR #6251 fixes the `request_id` dict bug only and is included here as the first commit.

## AI Assistance

AI assistance was used to implement this change and draft this PR (CodeBuddy). The submitter has reviewed every changed line and confirms correctness.

### DIFF
--- a/verl/trainer/ppo/ray_trainer.py
+++ b/verl/trainer/ppo/ray_trainer.py
@@ -449,7 +449,7 @@ class RayPPOTrainer:
                 k: (v.tolist() if isinstance(v, np.ndarray) else v) for k, v in reward_extra_infos_dict.items()
             }
             if "request_id" in batch.non_tensor_batch:
-                reward_extra_infos_dict.setdefault(
+                reward_extra_infos_to_dump.setdefault(
                     "request_id",
                     batch.non_tensor_batch["request_id"].tolist(),
                 )
--- a/verl/trainer/ppo/ray_trainer.py
+++ b/verl/trainer/ppo/ray_trainer.py
@@ -22,6 +22,7 @@ import json
 import os
 import uuid
 from collections import defaultdict
+from concurrent.futures import ThreadPoolExecutor
 from copy import deepcopy
 from pprint import pprint
 from typing import Any, Optional
@@ -318,6 +319,8 @@ class RayPPOTrainer:
         self._create_dataloader(train_dataset, val_dataset, collate_fn, train_sampler)
 
         self.checkpoint_manager = None
+        self._dump_executor = ThreadPoolExecutor(max_workers=1)
+        self._dump_futures = []  # track pending async dump tasks
 
     def _create_dataloader(self, train_dataset, val_dataset, collate_fn, train_sampler: Optional[Sampler]):
         """
@@ -401,10 +404,11 @@ class RayPPOTrainer:
         except Exception as e:
             print(f"Warning: Could not set total_training_steps in config. Structure missing? Error: {e}")
 
-    def _dump_generations(self, inputs, outputs, gts, scores, reward_extra_infos_dict, dump_path):
-        """Dump rollout/validation samples as JSONL."""
+    @staticmethod
+    def _write_generations(inputs, outputs, gts, scores, reward_extra_infos_dict, dump_path, global_steps):
+        """Write generation samples as JSONL (runs in background thread)."""
         os.makedirs(dump_path, exist_ok=True)
-        filename = os.path.join(dump_path, f"{self.global_steps}.jsonl")
+        filename = os.path.join(dump_path, f"{global_steps}.jsonl")
 
         n = len(inputs)
         base_data = {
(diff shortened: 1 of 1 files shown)


boundless-future:
Created in error, will re-create after review on fork first.

PR #6258, opened 2026-05-07
Revert "[reward] fix: compute correct rollout world size"

wuxibin89:
Reverts verl-project/verl#6226

@guillemgt reward model has no `pipeline_model_parallel_size` config.


guillemgt:
@wuxibin89 I had set it in my runs so I didn't catch that (and it worked properly).
The type annotations in RewardModelConfig would indicate that it should exist, but apparently that class is not actually being used in the configs. I've opened https://github.com/verl-project/verl/pull/6265 to address these issues.

PR #6324, opened 2026-05-12
[trainer] feat: async generation dump with exception propagation and streaming write

### LINKED ISSUE #6338: [PERF] Synchronous generation dump blocks training loop and silently drops I/O errors
## System Info

- verl: main
- python: 3.11+
- pytorch: 2.x

## Information

- [ ] Official example scripts
- [x] Custom modified scripts

## Task

- [ ] Official supported task in examples folder
- [x] Custom task or dataset

---

## Problem Description

`RayPPOTrainer._dump_generations` currently has three issues:

1. **Synchronous I/O blocks the training loop**: Every time generation data is dumped, `json.dumps` + `f.write` executes synchronously on the main thread. With large batches (e.g., GRPO n=16, batch_size=1024), a single dump can take several seconds, directly blocking the next rollout step in the training loop.

2. **I/O errors silently dropped**: The current write logic has no exception propagation mechanism. If disk full, permission errors, or other I/O exceptions occur, the error is swallowed or only surfaces at final `shutdown`, causing training to appear normal while data is actually lost.

3. **High peak memory**: The current approach uses `"\n".join(lines)` to build a complete string before writing, which creates unnecessary memory spikes with large numbers of samples.

## Expected Behavior

- Generation dump should execute asynchronously in a background thread, not blocking the main training loop
- I/O errors should surface early at the next `_dump_generations` call (via `future.result()` exception propagation)
- Use streaming line-by-line writes to avoid peak memory from constructing large strings
- At training end, all pending dump futures should be drained and checked to ensure no write errors are lost

Jackie2049:
## Summary

Fixes #6338

Make generation dump in `RayPPOTrainer` asynchronous with exception propagation and streaming writes, resolving:

1. **Synchronous I/O blocks the training loop** — dump runs on the main thread, blocking rollout with large batches
2. **I/O errors silently dropped** — disk full and other exceptions go unnoticed
3. **High peak memory** — `"\n".join(lines)` builds a large string before writing

## Changes

- **Extract `_write_generations` static method**: Lift the inline write logic into a `@staticmethod` so the background thread doesn't capture `self`, ensuring thread safety.
- **Async dump**: Submit write tasks via `ThreadPoolExecutor(max_workers=1)`; the main thread returns immediately without blocking training.
- **Exception propagation**: Track submitted futures in `_dump_futures`. On each call, remove completed futures and call `.result()` to surface I/O errors early.
- **Streaming write**: Replace `"\n".join(lines)` + single `f.write()` with per-line `f.write()`, reducing peak memory.
- **Drain pending futures**: Before `_dump_executor.shutdown(wait=True)`, iterate all remaining futures and call `.result()` so no write errors are lost at training end.
- **Fix `request_id` dict bug**: Write `request_id` to `reward_extra_infos_to_dump` instead of `reward_extra_infos_dict`.

## Testing

- [x] Local PPO training runs correctly, generation data written to JSONL files as expected
- [x] Simulated I/O error (read-only directory): exception propagates correctly on the next `_dump_generations` call
- [x] All pending dumps complete at training end, no data loss

## Note on Training Correctness

This change does not affect training precision:
1. Generation dump is purely observational/logging — written data does not participate in loss computation
2. Async write ordering is guaranteed by `global_steps` in the filename, no confusion
3. If an I/O error occurs, the exception propagates to the main training loop — training fails rather than silently losing data

## AI Assistance

AI assistance was used to implement this change and draft this PR (CodeBuddy). The submitter has reviewed every changed line and confirms correctness.


verl/trainer/ppo/ray_trainer.py
```diff
@@ -1657,6 +1681,11 @@ def fit(self):
...
                     if hasattr(self.actor_rollout_wg, "async_calls_finalize_fn_exec"):
                         self.actor_rollout_wg.async_calls_finalize_fn_exec(blocking=True)
+                    # Check all remaining dump futures for exceptions before shutdown
+                    for f in self._dump_futures:
+                        f.result()
+                    self._dump_futures.clear()
+                    self._dump_executor.shutdown(wait=True)
                     pprint(f"Final validation metrics: {last_val_metrics}")
                     progress_bar.close()
                     return
```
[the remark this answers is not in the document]
Jackie2049: Thanks for the review. All three points addressed:

1. **Resource leak on early exit** — Extracted `_shutdown_dump_executor()` and called at all exit points (`val_only`, `is_last_step`, dataloader exhaustion). See `7b480271`.

2. **Data loss** — Same fix covers dataloader exhaustion. For exceptions, since they typically terminate the process (executor cleaned up by OS), this is an acceptable tradeoff to avoid wrapping the training loop.

3. **Trainer reusability** — Added a guard at `fit()` entry to recreate the executor if previously shut down. See `7abeecb0`.

verl/trainer/ppo/ray_trainer.py
```diff
@@ -399,34 +401,67 @@ def _create_dataloader(self, train_dataset, val_dataset, collate_fn, train_sampl
         except Exception as e:
             print(f"Warning: Could not set total_training_steps in config. Structure missing? Error: {e}")
 
-    def _dump_generations(self, inputs, outputs, gts, scores, reward_extra_infos_dict, dump_path):
```
wuxibin89, line 402: ray_trainer.py will be soon deprecated by `verl/trainer/main_ppo_sync.py`, please update it as well.

boundless-future: @wuxibin89 , sorry for the accidental push that interrupted the CI! 🙏
I've addressed your feedback — updated `main_ppo_sync.py` with the async dump changes as well. Could you please re-review and re-trigger the CI when you get a chance? Thanks!

wuxibin89 (approved):

PR #6325, opened 2026-05-12
[veomni] feat: add MoE router replay (R2/R3) support

hjshi84:
### What does this PR do?

Wires VeOmni's `maybe_replay_indices` hook into the verl actor pipeline
so MoE routing decisions can be recorded during `compute_log_prob` and
replayed during the actor update (R2), or recorded at rollout time and
replayed in both `compute_log_prob` and the update (R3). Removes the
floating-point drift between rollout and actor router decisions that
otherwise contaminates the RL policy gradient on MoE models.

Wired today for `qwen3_moe` and `qwen3_5_moe` (the families with the
hook installed in their patched `SparseMoeBlock` on the VeOmni side).

### Design & Code Changes

- Indices-only contract: all model-specific weight math (gather, renorm,
  scaling, dtype cast) stays in the per-family `SparseMoeBlock` patch
  on the VeOmni side. The verl-side controller is model-agnostic.
- Id-keyed positional indexing: each MoE router gets its position
  assigned the first time it fires (`id(module) -> pos`); subsequent
  fires (under any backward order) hit the same position via dict
  lookup. This is what makes RR correct under per-layer activation
  checkpointing recompute, where backward replays each layer's forward
  independently in reverse order.
- Strict REPLAY: missing `routed_experts` on the batch, or missing
  target for any layer position, raises with a typed message. No silent
  fallback to native router decisions.
- R3-only response mask: R2 RECORD captures full-sequence routing in
  compute_log_prob, so REPLAY substitutes uniformly. R3 RECORD runs at
  the rollout backend and only captures response-token routing, so
  prompt-token positions carry zero placeholders — REPLAY must mask
  them out and let them go through native routing or every prompt
  token's topk slots get sent to expert 0.
- Defensive duplicate-row fallback: VeOmni's expert dispatch collapses
  duplicate top-k slots in `permute()` while `input_splits` keeps
  counting them, so any token with duplicate substituted experts
  crashes the EP all-to-all. Rows whose substituted top-k contains a
  duplicate fall back to native indices.
- `try / finally` lifecycle in `forward_backward_batch` so an exception
  during forward / backward / postprocess never pins GPU memory in the
  controller across steps.
- Requires `use_remove_padding=True`. The non-rmpad VeOmni MoE path
  also loses Ulysses SP slicing and the fused-kernel log_probs path,
  and is not a tested production configuration; checked at
  config-coherence time in `ActorRolloutRefWorker.__init__` with a
  defense-in-depth guard at engine init.

Files:
- `verl/utils/veomni/router_replay.py` — `VeOmniRouterReplay` controller
  (state machine, RECORD aggregation w/ Ulysses SP all-gather, REPLAY
  per-mb target slicing).
- `verl/workers/engine/veomni/transformer_impl.py` — RECORD/REPLAY
  driver in `forward_backward_batch`; snapshot side-channel + jagged
  NestedTensor invariant assertion + R3 per-token mask construction in
  `prepare_model_inputs`.
- `verl/workers/engine_workers.py` — wires
  `actor.veomni.router_replay.mode` through the worker; fail-fast on
  the rmpad constraint.
- `verl/experimental/agent_loop/agent_loop.py` — narrow `routed_experts`
  carrier dtype from int64 to int32 (expert ids fit comfortably; halves
  the R3 payload through DataProto).
- `verl/trainer/config/engine/veomni.yaml` + regenerated
  `_generated_ppo_veomni_trainer.yaml` — exposes the `router_replay`
  block on the veomni engine schema (mirrors megatron's structure).
- `examples/router_replay/run_qwen3_5_35b_a3b_veomni.sh` — GRPO
  Qwen3.5-35B-A3B demo with `ROUTING_REPLAY_MODE` env toggle.
- `tests/utils/veomni/test_router_replay_on_cpu.py` (15 tests) —
  controller state machine: RECORD/REPLAY lifecycle, per-layer reverse-
  order activation-checkpointing recompute, R3 first-step (REPLAY
  before any RECORD), per-token mask substitution, duplicate-row
  fallback, install/uninstall against a stubbed VeOmni hook.
- `tests/workers/test_router_replay_engine_helpers_on_cpu.py` (14
  tests) — engine-driver glue: flat→nested unpacker, side-channel
  list-stash regression, R2/R3 mask construction, strict missing-input
  error paths.

### API and Usage Example

```bash
# R2: record indices in compute_log_prob, replay in actor update.
ROUTING_REPLAY_MODE=R2 \
  bash examples/router_replay/run_qwen3_5_35b_a3b_veomni.sh

# R3: record at rollout, replay in both compute_log_prob and update.
ROUTING_REPLAY_MODE=R3 \
  bash examples/router_replay/run_qwen3_5_35b_a3b_veomni.sh
```

Or via Hydra override:
```
actor_rollout_ref.actor.veomni.router_replay.mode=R2
```

### Test

- AST + `ruff check` clean on the modified files.
- 29 CPU unit tests across the two test files above; all pass. Coverage
  includes the per-layer reverse-order activation-checkpointing case
  (the realistic VeOmni MoE training pattern that breaks any monotonic
  cursor design) and the R3 first-step case (REPLAY runs before any
  RECORD has populated the layer table).
- GPU end-to-end on Qwen3.5-35B-A3B GRPO, R2 mode: actor metrics show
  the expected reduction in actor↔rollout router divergence —
  `actor/ppo_kl` ~5.5e-5 (R2) vs ~9e-5 (disabled), and
  `actor/pg_clipfrac` ~28.6% vs ~35.9% (≈20% reduction). Effect is
  diluted by the per-step PPO mini-batch averaging; raw per-mb effect
  is larger.
- qwen3.5_35ba3b_grpo_geo3k_sp2_ep1
image

### Dependencies

Requires the VeOmni-side hook (`set_active_replay`,
`maybe_replay_indices`) to be present and wired into the patched
`SparseMoeBlock.forward` for the target MoE family.

### DIFF
--- a/.github/workflows/e2e_sft_llm.yml
+++ b/.github/workflows/e2e_sft_llm.yml
@@ -105,8 +105,8 @@ jobs:
           pip3 install peft
           pip3 install -r requirements-test.txt
           pip3 install --no-deps -e .
-          pip3 install veomni==0.1.9a5 --ignore-requires-python --no-deps --index-url https://pypi.org/simple/
-          pip3 install transformers==4.57.3
+          pip3 install veomni==0.1.10 --ignore-requires-python --no-deps --index-url https://pypi.org/simple/
+          pip3 install transformers==5.3.0
       - name: Prepare gsm8k dataset
         run: |
           ray stop --force
--- a/verl/workers/engine_workers.py
+++ b/verl/workers/engine_workers.py
@@ -483,24 +483,6 @@ class ActorRolloutRefWorker(Worker, DistProfilerExtension):
             rr_mode = "disabled"
         self.enable_routing_replay = rr_mode != "disabled"
 
-        # Fail fast on the (router_replay enabled) + (use_remove_padding=False)
-        # combination for the veomni engine. We do this here, at config-
-        # coherence-check time, rather than inside the engine's initialize()
-        # so the user gets the error before any expensive model build.
-        if (
-            self.enable_routing_replay
-            and actor_strategy == "veomni"
-            and not self.config.model.get("use_remove_padding", False)
-        ):
-            raise RuntimeError(
-                "router_replay requires use_remove_padding=True. In VeOmni engine, "
-                "the non-remove-padding path also disables Ulysses SP slicing and "
-                "the fused-kernel log_probs path, and is not a tested production "
-                "configuration for MoE routing replay. Set "
-                "actor.model.use_remove_padding=True or "
-                "router_replay.mode='disabled'."
-            )
-
         DistProfilerExtension.__init__(
             self, DistProfiler(rank=self.rank, config=profiler_config, tool_config=tool_config)
         )
--- a/verl/workers/engine_workers.py
+++ b/verl/workers/engine_workers.py
@@ -471,9 +471,35 @@ class ActorRolloutRefWorker(Worker, DistProfilerExtension):
         else:
             tool_config = None
 
-        self.enable_routing_replay = (
-            self.config.actor.strategy == "megatron" and self.config.actor.megatron.router_replay.mode != "disabled"
-        )
+        # Router replay is supported on the megatron engine and on the veomni
+        # engine. Both expose `router_replay` on their per-strategy engine
+        # config (the field lives on the shared `EngineConfig` base).
+        actor_strategy = self.config.actor.strategy
+        if actor_strategy == "megatron":
+            rr_mode = self.config.actor.megatron.router_replay.mode
+        elif actor_strategy == "veomni":
+            rr_mode = self.config.actor.veomni.router_replay.mode
+        else:
+            rr_mode = "disabled"
+        self.enable_routing_replay = rr_mode != "disabled"
+
+        # Fail fast on the (router_replay enabled) + (use_remove_padding=False)
+        # combination for the veomni engine. We do this here, at config-
+        # coherence-check time, rather than inside the engine's initialize()
+        # so the user gets the error before any expensive model build.
+        if (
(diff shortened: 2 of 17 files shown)


verl/workers/engine_workers.py
```diff
@@ -471,9 +471,35 @@ def __init__(
...
+        actor_strategy = self.config.actor.strategy
+        if actor_strategy == "megatron":
+            rr_mode = self.config.actor.megatron.router_replay.mode
+        elif actor_strategy == "veomni":
+            rr_mode = self.config.actor.veomni.router_replay.mode
+        else:
+            rr_mode = "disabled"
+        self.enable_routing_replay = rr_mode != "disabled"
+
+        # Fail fast on the (router_replay enabled) + (use_remove_padding=False)
```
wuxibin89: Move this check into `VeOmniActorConfig.__post_init__` to keep engine_workers.py backend agnostic.

.github/workflows/e2e_sft_llm.yml
```diff
@@ -105,8 +105,8 @@ jobs:
           pip3 install peft
           pip3 install -r requirements-test.txt
           pip3 install --no-deps -e .
-          pip3 install veomni==0.1.9a5 --ignore-requires-python --no-deps --index-url https://pypi.org/simple/
-          pip3 install transformers==4.57.3
+          pip3 install veomni==0.1.10 --ignore-requires-python --no-deps --index-url https://pypi.org/simple/
```
Luosuu, line 108: we can install 0.1.11 now

wuxibin89 (approved):

PR #6334, opened 2026-05-13
[fsdp, ckpt] fix: drop tied target keys before HF save_pretrained

ChangyiYang:
Credit to @nathanrchn for finding this bug and propose solution!

### What does this PR do?

FSDP gather un-aliases tied params (one independent CPU tensor per state_dict key), so `save_pretrained`'s storage-based dedup in `remove_tied_weights_from_state_dict` never fires for them. Both `model.embed_tokens.weight` and `lm_head.weight` end up in the safetensors with distinct (eventually-divergent) values.

On reload, `transformers>=5` detects the duplicate and silently refuses to tie (logs a warning, empties `all_tied_weights_keys`). The two heads then train as independent parameters in every subsequent stage, defeating `tie_word_embeddings=true` and quietly losing the `lm_head` learning that should have flowed through the embedding.

### Fix

Drop tied alias keys (e.g. `lm_head.weight`) from `state_dict` before `save_pretrained` in:

- `verl/utils/checkpoint/fsdp_checkpoint_manager.py` — auto-save path
- `verl/model_merger/base_model_merger.py` — CLI merger path

The drop uses `named_parameters(remove_duplicate=False)` after `tie_weights()` to detect aliases via `Parameter` identity. The first occurrence in iteration order is the canonical source; later occurrences are aliases and get dropped. Avoids relying on the `_tied_weights_keys` attribute, whose format and contents (list of aliases vs list of both source+aliases) vary across model classes and transformers versions.

### Verification

End-to-end on real Slurm training with `LiquidAI/LFM2.5-350M` (`tie_word_embeddings=True`) on MI300X:

| stage | verl | base ckpt | save outcome |
|---|---|---|---|
| pre-fix | this PR not applied | HF base (148 keys) | hf_model contains 149 keys (both `lm_head.weight` and `embed_tokens.weight`), bit-equal at step 1 |
| post-fix | this PR | HF base (148 keys) | hf_model contains 148 keys, `lm_head.weight` dropped |
| resume-from-broken | this PR not applied | pre-fix broken 149-key ckpt | `transformers 5.2.0` emits `we will NOT tie them` warning; FSDP wraps with two independent `FlatParameter`s; 10 grad steps produce `max_abs_diff=5.65e-4` on 99.87% of elements between `lm_head.weight` and `embed_tokens.weight` |

The resume-from-broken cascade quantified:

```
 step | max_diff  | mean_diff | %elements_diff
   2  | 2.00e-5   | 6.18e-7   | 99.19%
   4  | 1.18e-4   | 2.95e-6   | 99.78%
   6  | 2.56e-4   | 6.27e-6   | 99.84%
   8  | 4.08e-4   | 9.73e-6   | 99.86%
  10  | 5.65e-4   | 1.31e-5   | 99.87%
```

The `lm_head` and `embed_tokens` continue to silently train as independent params for the remainder of the run.

### Related

Same root cause as the long-standing pytorch/pytorch#77724 ("FSDP: enhanced shared parameter support") and pytorch/pytorch#85949 ("Loading distributed checkpoint with FSDP fails ... shared.weight"). PyTorch FSDP's gather contract doesn't preserve tied-parameter aliasing across the GPU→CPU materialisation, and HF `save_pretrained`'s dedup is storage-pointer-based, so the gap can only be closed at the user (verl) layer.

This fix lives in our internal fork at https://github.com/Liquid4All/verl/pull/62. Sharing upstream so other verl users don't hit the same silent corruption.

### DIFF
--- a/verl/model_merger/base_model_merger.py
+++ b/verl/model_merger/base_model_merger.py
@@ -29,24 +29,9 @@ from verl.utils.transformers_compat import get_auto_model_for_vision2seq
 AutoModelForVision2Seq = get_auto_model_for_vision2seq()
 
 
-def _drop_tied_target_keys(state_dict: dict, model, model_config) -> None:
-    """Drop tied alias keys (e.g. ``lm_head.weight``) from ``state_dict``.
-
-    The merger materialises one independent CPU tensor per state_dict key,
-    which defeats ``save_pretrained``'s storage-pointer-based dedup. When both
-    keys end up in safetensors, ``transformers>=5`` silently refuses to re-tie
-    on reload. Detects aliases by ``Parameter`` identity after ``tie_weights()``.
-    """
-    if not getattr(model_config, "tie_word_embeddings", False):
-        return
-    model.tie_weights()
-    seen: dict[int, str] = {}
-    for name, param in model.named_parameters(remove_duplicate=False):
-        pid = id(param)
-        if pid in seen:
-            state_dict.pop(name, None)
-        else:
-            seen[pid] = name
+# Re-export so existing callers that previously imported the symbol from this
+# module continue to work.
+from verl.utils.transformers_compat import drop_tied_target_keys as _drop_tied_target_keys  # noqa: E402
 
 
 def parse_args():
--- a/verl/model_merger/base_model_merger.py
+++ b/verl/model_merger/base_model_merger.py
@@ -29,6 +29,26 @@ from verl.utils.transformers_compat import get_auto_model_for_vision2seq
 AutoModelForVision2Seq = get_auto_model_for_vision2seq()
 
 
+def _drop_tied_target_keys(state_dict: dict, model, model_config) -> None:
+    """Drop tied alias keys (e.g. ``lm_head.weight``) from ``state_dict``.
+
+    The merger materialises one independent CPU tensor per state_dict key,
+    which defeats ``save_pretrained``'s storage-pointer-based dedup. When both
+    keys end up in safetensors, ``transformers>=5`` silently refuses to re-tie
+    on reload. Detects aliases by ``Parameter`` identity after ``tie_weights()``.
+    """
+    if not getattr(model_config, "tie_word_embeddings", False):
+        return
+    model.tie_weights()
+    seen: dict[int, str] = {}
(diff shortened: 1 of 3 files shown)


ChangyiYang:
Thanks @gemini-code-assist — both flagged issues addressed in f6849145:

**Dedup**: Moved `_drop_tied_target_keys` to `verl/utils/transformers_compat.py` as `drop_tied_target_keys` and import it from both `fsdp_checkpoint_manager.py` and `base_model_merger.py`. The two copies were identical so this prevents future divergence.

**At-least-one-key invariant**: Hardened the drop loop to only pop an alias when the canonical name is confirmed present in `state_dict`; if only the alias is in `state_dict`, promote it to canonical instead. Verified locally on four cases: both present, only canonical, only alias, and `tie_word_embeddings=False`.

verl/model_merger/base_model_merger.py
```diff
@@ -29,6 +29,11 @@
 AutoModelForVision2Seq = get_auto_model_for_vision2seq()
 
 
+# Re-export so existing callers that previously imported the symbol from this
+# module continue to work.
+from verl.utils.transformers_compat import drop_tied_target_keys as _drop_tied_target_keys  # noqa: E402
```
wuxibin89, line 34: I'm a bit confuse here, why not just import it above:
```python
from verl.utils.transformers_compat import get_auto_model_for_vision2seq, drop_tied_target_keys
```

wuxibin89 (approved):

ChangyiYang:
@wuxibin89 good catch — folded the helper into the existing top-level imports in 47f6e23c, dropped the `_drop_tied_target_keys` alias, both call sites use the public name now.

wuxibin89:
> @wuxibin89 good catch — folded the helper into the existing top-level imports in [47f6e23](https://github.com/verl-project/verl/commit/47f6e23c4b6c5d4524cf940f469406cd575c8726), dropped the `_drop_tied_target_keys` alias, both call sites use the public name now.

Could you submit a PR to fix it?

ChangyiYang:
@wuxibin89 follow-up PR up — #6356. Sorry I should've spotted that #6334 was already merged with the back-compat alias still in tree; the fold-up didn't make it in.

Issue #6293, opened 2026-05-09
[bug] use_remove_padding=False 路径缺失distillation top-k 处理，导致 distillation_losses 缺失并触发 KeyError


wizkdc:
### System Info

----------Python Info----------
Version : 3.12.3
Compiler : GCC 13.3.0
Build : ('main', 'Mar 3 2026 12:15:18')
Arch : ('64bit', '')
------------Pip Info-----------
Version : 26.0.1
Directory : /usr/local/lib/python3.12/dist-packages/pip
vllm : 0.19.0
sglang : not found.
ray : 2.54.1
torch : 2.10.0+cu129
----------verl Info-----------
Version : 0.8.0.dev
Directory : /root/workspace/pw/pw_2/verl/verl
Commit Hash : https://github.com/verl-project/verl/commit/e45dc3d3e5dc248a167bb5b3c341dfe6be9a5ba5
----------Platform Info----------
Platform : Linux-5.10.134-16.3.al8.x86_64-x86_64-with-glibc2.39
system : Linux
node : zzh-t3e2wjad-0000
release : 5.10.134-16.3.al8.x86_64
version : https://github.com/verl-project/verl/pull/1 SMP Tue Mar 26 18:54:05 CST 2024
----------Environment----------
CUDA Runtime : 12.9
CUDA Compiler : Cuda compilation tools, release 12.9, V12.9.86
----------System Info----------
CPU Memory : 2014.75 GB
GPU Count : 8
GPU 1 Type : NVIDIA H20
GPU 1 Memory : 95.58 GB
GPU 2 Type : NVIDIA H20
GPU 2 Memory : 95.58 GB
GPU 3 Type : NVIDIA H20
GPU 3 Memory : 95.58 GB
GPU 4 Type : NVIDIA H20
GPU 4 Memory : 95.58 GB
GPU 5 Type : NVIDIA H20
GPU 5 Memory : 95.58 GB
GPU 6 Type : NVIDIA H20
GPU 6 Memory : 95.58 GB
GPU 7 Type : NVIDIA H20
GPU 7 Memory : 95.58 GB
GPU 8 Type : NVIDIA H20
GPU 8 Memory : 95.58 GB

### Information

- [ ] The official example scripts
- [x] My own modified scripts

### Tasks

- [x] An officially supported task in the `examples` folder (such as GLUE/SQuAD, ...)
- [ ] My own task or dataset (give details below)

### Reproduction

[...]

PR #6386, opened 2026-05-18
[fsdp] fix: emit distillation outputs in use_remove_padding=False path

abinggo:
## Background

#6350 (closed by #6360 revert) regressed the cpu_unit_tests CI job:

```
FAILED tests/workers/test_distillation_topk_symmetry_on_cpu.py::test_distillation_outputs_emitted_in_both_padding_modes[True]
ValueError: Pointer argument (at 0) cannot be accessed from Triton (cpu tensor?)
```

The original fix for #6293 was correct; the test's interaction with `logprobs_from_logits` was not.

Per @wuxibin89's request on #6350, this PR re-applies the production fix and patches the test so it actually runs on CPU.

## Production fix (`verl/workers/engine/fsdp/transformer_impl.py`)

Re-applies the 15-line addition from #6350: the `use_remove_padding=False` branch of `prepare_model_outputs` now mirrors the `True` branch's `distillation_use_topk` handling. Without this, downstream distillation losses raise `KeyError` on the no-rmpad path (#6293 root cause).

No SP gather here — this branch is the no-SP path (`log_probs` is also not gathered), and `pad_size` is only populated in `output_args` along the `use_remove_padding=True` path of `prepare_model_inputs`.

## CPU test fix (`tests/workers/test_distillation_topk_symmetry_on_cpu.py`)

Root cause: `prepare_model_outputs` calls `logprobs_from_logits(logits=logits_rmpad, labels=...)`. When flash-attn is installed in CI, that dispatches to flash-attn's Triton `CrossEntropyLoss`, which only operates on CUDA tensors. The test runs on CPU.

Fix: patch `logprobs_from_logits` for the duration of the call with a substitute that returns `torch.zeros(total_nnz)` of the right shape. The contract under test is the propagation of distillation keys, not the numerical correctness of log-probs. The downstream assertion `v.shape == log_probs.shape` continues to hold because `logits_processor_func` returns tensors of shape `(1, student_logits.shape[1])` and `logits_rmpad` is contiguous along the `total_nnz` dim in both branches.

## Local verification

```
KMP_DUPLICATE_LIB_OK=TRUE PYTHONPATH=. python3 -m pytest \
    tests/workers/test_distillation_topk_symmetry_on_cpu.py -v
```

Output:

```
collecting ... collected 2 items

tests/workers/test_distillation_topk_symmetry_on_cpu.py::test_distillation_outputs_emitted_in_both_padding_modes[True]  PASSED [ 50%]
tests/workers/test_distillation_topk_symmetry_on_cpu.py::test_distillation_outputs_emitted_in_both_padding_modes[False] PASSED [100%]

============================== 2 passed, 3 warnings in 7.57s ==============================
```

Both parametrizations pass; the previous CI failure mode (`ValueError: Pointer argument (at 0) cannot be accessed from Triton`) is gone because the Triton kernel is no longer reached on the CPU path.

## Refs

- #6293 (issue)
- #6350 (original fix, reverted by #6360)
- #6360 (revert)

### DIFF
--- a/tests/workers/test_distillation_topk_symmetry_on_cpu.py
+++ b/tests/workers/test_distillation_topk_symmetry_on_cpu.py
@@ -1,0 +1,159 @@
+# Copyright 2026 Bytedance Ltd. and/or its affiliates
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Regression guard for verl#6293.
+
+The use_remove_padding=False branch of
+FSDPEngineWithLMHead.prepare_model_outputs previously lacked the
+distillation_use_topk handling that the use_remove_padding=True branch had,
+so distillation outputs were silently dropped from model_output and the
+downstream loss raised KeyError. This test invokes prepare_model_outputs on
+a stub engine for both branches with distillation_use_topk=True and asserts
+the distillation keys produced by logits_processor_func are propagated into
+model_output as nested tensors in both cases.
+
+``logprobs_from_logits`` is patched out: in CI environments where flash-attn
+is installed, it dispatches to a Triton CrossEntropyLoss kernel that cannot
+operate on CPU tensors. The substitute returns a dummy ``log_probs`` tensor
+of the right shape, which is sufficient for this test — the contract under
+test is the propagation of distillation keys, not the numerical correctness
+of log-prob computation.
+"""
+
+import os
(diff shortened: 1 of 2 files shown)


abinggo:
The two failing CI jobs look unrelated to this patch:

- `e2e_one_step_off_policy_fsdp2` — `ray.exceptions.RayTaskError(OutOfMemoryError)` from the Ray task runner during the e2e training step ([log](https://github.com/verl-project/verl/actions/runs/26015051963/job/76470999928)).
- `e2e_one_step_off_policy_megatron_ascend` — same e2e family on Ascend ([log](https://github.com/verl-project/verl/actions/runs/26015051928/job/76470962343)).

The CPU unit suite (which actually exercises the distillation symmetry contract this PR adds back via `tests/workers/test_distillation_topk_symmetry_on_cpu.py`) passes:

> `cpu_unit_tests pass 4m38s`

This patch only touches the `distillation_use_topk` branch of `prepare_model_outputs` (15 lines) and the regression test, neither of which is on the `one_step_off_policy` training path. Happy to retry / look closer if maintainers want me to dig deeper, but on first read this looks like an infra-side flake similar to what surfaced on #6350.

cc @wuxibin89

wuxibin89 (approved):

PR #6395, opened 2026-05-19
[fully_async] fix rollout failure propagation

### LINKED ISSUE #6387: [Bug][fully_async] Rollouter generation failures can be reported as successful training completion
### System Info

- Repository: `verl-project/verl`
- Observed on current upstream `main`
- Date checked: 2026-05-18
- Affected code path: `verl/experimental/fully_async_policy`
- Affected mode: fully async PPO training
- Environment specificity: this appears to be an async control-flow and queue-sentinel bug, not a hardware-specific issue.

### Information

- [x] My own modified scripts

### Tasks

### Reproduction

A minimal reproduction can be created by injecting a rollout-side failure:

1. Use a fully async PPO config.
2. Temporarily make `_streaming_generation_main()` or `_process_single_sample_streaming()` raise after startup:

```python
raise RuntimeError("injected rollout failure")
```

3. Start training through `verl/experimental/fully_async_policy/fully_async_main.py`.

Current behavior:

- `_streaming_generation_main()` logs the rollout exception.
- The cleanup path still puts `sample=None` into the message queue.
- `FullyAsyncRollouter.fit()` ignores the exception returned by `asyncio.gather(..., return_exceptions=True)`.
- `FullyAsyncRollouter.fit()` returns normally and prints `Rollouter fit completed`.
- `FullyAsyncTrainer` treats `sample=None` as normal termination.
- `FullyAsyncTaskRunner` can print component success.

This is a training-completeness and orchestration correctness issue, not only a logging issue.

If rollout generation fails midway because of a model server crash, agent loop exception, serialization error, data corruption, network issue, or another per-sample processing failure, the job can:

- stop collecting new rollout samples,
- stop training earlier than configured,
- run normal shutdown or validation paths,
- save or expose a final checkpoint as if training completed normally,
- return success to Ray / job orchestration,
- hide infrastructure or rollout-quality failures from automation.

The business value of fixing this is that fully async training becomes fail-closed. Users and schedulers can trust that a successful job status actually means rollout generation and training completed normally, rather than silently accepting incomplete training artifacts.

For long-running distributed RL jobs, this is important because a false successful completion can waste GPU time, produce misleading checkpoints, and pollute downstream evaluation or deployment pipelines.

### Expected behavior

- The rollout failure should propagate through `rollouter.fit.remote()`.
- The task runner should observe the Ray future failure.
- The trainer should not treat an abnormal rollout shutdown as normal queue EOF.
- The whole training job should be marked failed.

anzhsoft:
### What does this PR do?

Fixes #6387.

  This PR makes fully async rollout failures fail closed instead of being reported as normal training completion.

  Before this change, an unexpected exception in the fully async rollouter could still lead to `sample=None` being sent to the message queue. The trainer treats `None` as normal EOF, and `FullyAsyncRollouter.fit()` could also swallow async task exceptions via
  `return_exceptions=True`. As a result, Ray orchestration could observe successful completion even though rollout generation failed.

  This matters for long-running RL jobs because rollout-side failures such as vLLM OOM, decode errors, agent loop exceptions, serialization errors, or model-server failures should fail the job instead of producing incomplete or misleading training artifacts.

### Checklist Before Starting

  No open PR was found that directly addresses this issue.

### Test

  ```bash
  .venv/bin/python -m pytest -q tests/experimental/fully_async_policy/test_rollout_failure_signal_on_cpu.py

  Result:

  4 passed

  .venv/bin/python -m ruff check \
    verl/experimental/fully_async_policy/rollout_failure.py \
    verl/experimental/fully_async_policy/message_queue.py \
    verl/experimental/fully_async_policy/fully_async_rollouter.py \
    verl/experimental/fully_async_policy/fully_async_trainer.py \
    tests/experimental/fully_async_policy/test_rollout_failure_signal_on_cpu.py

  Result:

  All checks passed

### API and Usage Example

No user-facing API, CLI argument, or config change.

  Existing fully async PPO usage remains unchanged. The behavior change is only in failure handling: normal rollout completion still uses None as EOF, while unexpected rollout-side failures are now propagated as errors.

### Design & Code Changes

 The fix separates normal EOF from abnormal rollout shutdown.

  - Add RolloutErrorSignal as an explicit queue signal for unexpected rollout generation failures.
  - Keep None reserved for normal completion only.
  - Publish RolloutErrorSignal when _streaming_generation_main() exits with an unexpected exception.
  - Make FullyAsyncTrainer raise when it receives a rollout error signal instead of treating it as normal EOF.
  - Make FullyAsyncRollouter.fit() propagate unexpected async task failures after cleanup, so the Ray future fails and orchestration marks the job failed.
  - Add CPU-only regression tests for error-signal construction, trainer-side raising, cancellation filtering, and async task failure detection.

  This intentionally fails the training job instead of skipping failed rollouts, because silently continuing after a rollout pipeline failure can still produce incomplete or misleading checkpoints.

 - [x] Read the Contribute Guide (https://github.com/verl-project/verl/blob/main/CONTRIBUTING.md).
  - [ ] Apply pre-commit checks (https://github.com/verl-project/verl/blob/main/CONTRIBUTING.md#code-linting-and-formatting): pre-commit install && pre-commit run --all-files --show-diff-on-failure --color=always
      - Targeted ruff check was run on the changed files.
  - [x] Add / Update the documentation (https://github.com/verl-project/verl/tree/main/docs).
      - No documentation update is needed because this PR does not add or change a user-facing API/config.
  - [x] Add unit or end-to-end test(s) to the CI workflow (https://github.com/verl-project/verl/tree/main/.github/workflows) to cover all the code.
      - Added a CPU-only regression test under tests/experimental/fully_async_policy/.
  - [x] If your PR is related to the recipe submodule, please also update the reference to the submodule commit.
      - Not applicable; this PR does not touch recipe.

## AI Assistance
  AI assistance was used to analyze the async control flow, draft the fix, and prepare this PR description. I reviewed the changed code and test coverage before submitting.


verl/experimental/fully_async_policy/rollout_failure.py
```diff
@@ -0,0 +1,67 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import asyncio
+import traceback as traceback_lib
+from dataclasses import dataclass
+from typing import Any
+
+
+@dataclass(frozen=True)
```
ArronHZG: Could you put this code to detach_utils?

ArronHZG (commented):
I think if the rollouter throws an error, rollouter.fit should return the error to main, and then the entire training should be interrupted. The trainer should then finish normally without any problems.

verl/experimental/fully_async_policy/rollout_failure.py
anzhsoft: Thanks for the review. I updated the PR to address this.

  Changes made:
  - Moved `RolloutErrorSignal` and the async rollout failure helper functions into `detach_utils.py`.
  - Removed the separate `rollout_failure.py` module.
  - Kept `rollouter.fit()` responsible for raising the original rollout failure back to the main async runner, so the whole training job is interrupted by the rollouter-side error.
  - Updated the trainer queue consumer to treat `RolloutErrorSignal` as a termination signal and exit the sampling loop normally, instead of raising a separate trainer-side exception.
  - Added/updated CPU tests for:
    - preserving rollout exception details,
    - formatting rollout error signals,
    - ignoring expected `asyncio.CancelledError`,
    - returning early when a rollout task fails while another task is still pending.

  This should match the intended behavior: rollout failures are propagated through `rollouter.fit()` to main, while the trainer observes the queue signal and finishes cleanly.

anzhsoft:
> I think if the rollouter throws an error, rollouter.fit should return the error to main, and then the entire training should be interrupted. The trainer should then finish normally without any problems.

Any other concerns?

PR #6416, opened 2026-05-20
[worker]fix: restore actor engine GPU memory logging

wlhgtc:
Restore GPU memory logging for actor engine compute_log_prob and update_actor.

When `VERL_LOGGING_LEVEL=DEBUG` , users should be able to inspect actor GPU memory usage around these key actor engine calls. 
This feature was removed together with the legacy dp_actor worker in #6067, so this change adds it back to the new actor engine worker path.

### What does this PR do?

### Checklist Before Starting

### Test

### API and Usage Example

```python
```

### Design & Code Changes

> [!IMPORTANT]

### DIFF
--- a/verl/utils/profiler/performance.py
+++ b/verl/utils/profiler/performance.py
@@ -93,7 +93,13 @@ class GPUMemoryLogger(DecoratorLoggerBase):
         ...     return
     """
 
-    def __init__(self, role: str, logger: logging.Logger = None, level=logging.DEBUG, log_only_rank_0: bool = True):
+    def __init__(
+        self,
+        role: str = "",
+        logger: logging.Logger = None,
+        level=logging.DEBUG,
+        log_only_rank_0: bool = True,
+    ):
         if dist.is_initialized() and dist.get_world_size() > 1:
             rank = dist.get_rank()
         else:
--- a/verl/utils/profiler/performance.py
+++ b/verl/utils/profiler/performance.py
@@ -107,6 +107,11 @@ class GPUMemoryLogger(DecoratorLoggerBase):
         return f
 
     def log(self, func, *args, **kwargs):
+        if dist.is_initialized():
+            # Decorators may be constructed before distributed initialization, so refresh rank at call time.
+            self.rank = dist.get_rank()
+            if self.log_only_rank_0 and self.rank != 0:
+                return func(*args, **kwargs)
         name = func.__name__
         mem_allocated, mem_reserved, mem_used, mem_total = _get_current_mem_info()
         message = (
--- a/verl/utils/profiler/performance.py
+++ b/verl/utils/profiler/performance.py
@@ -107,6 +107,9 @@ class GPUMemoryLogger(DecoratorLoggerBase):
         return f
 
     def log(self, func, *args, **kwargs):
+        if self.logger is not None and not self.logger.isEnabledFor(self.level):
+            return func(*args, **kwargs)
+
         if dist.is_initialized():
             # Decorators may be constructed before distributed initialization, so refresh rank at call time.
             self.rank = dist.get_rank()
--- a/verl/utils/profiler/performance.py
+++ b/verl/utils/profiler/performance.py
@@ -93,13 +93,7 @@ class GPUMemoryLogger(DecoratorLoggerBase):
         ...     return
     """
 
-    def __init__(
-        self,
-        role: str = "",
-        logger: logging.Logger = None,
-        level=logging.DEBUG,
-        log_only_rank_0: bool = True,
-    ):
+    def __init__(self, role: str, logger: logging.Logger = None, level=logging.DEBUG, log_only_rank_0: bool = True):
         if dist.is_initialized() and dist.get_world_size() > 1:
             rank = dist.get_rank()
         else:
--- a/verl/utils/profiler/profile.py
+++ b/verl/utils/profiler/profile.py
@@ -13,7 +13,6 @@
 # limitations under the License.
 
 import functools
-import logging
 from typing import Callable, Optional
 
 from ..memory_utils import MemorySnapshotSampler, clear_memory_history, enable_memory_visualize
@@ -169,21 +168,11 @@ class DistProfiler:
         color: Optional[str] = None,
         domain: Optional[str] = None,
         category: Optional[str] = None,
-        log_gpu_memory: bool = False,
-        logger: Optional[logging.Logger] = None,
         **kwargs_outer,
     ) -> Callable:
         def decorator(func):
-            gpu_memory_logger = None
-            if log_gpu_memory:
-                from .performance import GPUMemoryLogger
-
-                gpu_memory_logger = GPUMemoryLogger(
-                    logger=logger,
-                )
-
             @functools.wraps(func)
-            def profiled_func(self_instance, *args, **kwargs_inner):
+            def wrapper(self_instance, *args, **kwargs_inner):
                 profiler = getattr(self_instance, "profiler", None)
                 if (
                     not profiler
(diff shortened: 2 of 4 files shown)


verl/utils/profiler/performance.py
```diff
@@ -107,6 +107,11 @@ def f(*args, **kwargs):
         return f
 
     def log(self, func, *args, **kwargs):
+        if dist.is_initialized():
+            # Decorators may be constructed before distributed initialization, so refresh rank at call time.
+            self.rank = dist.get_rank()
+            if self.log_only_rank_0 and self.rank != 0:
+                return func(*args, **kwargs)
```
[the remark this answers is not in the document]
wlhgtc: fix

verl/workers/engine_workers.py
```diff
@@ -633,6 +639,7 @@ def compute_ref_log_prob(self, data: TensorDict) -> TensorDict:
     @register(dispatch_mode=make_nd_compute_dataproto_dispatch_fn(mesh_name="actor"))
     @DistProfiler.annotate(color="blue", role="actor_compute_log_prob")
     @_with_routing_replay_flag(enabled=True)
+    @GPUMemoryLogger(role="actor engine", logger=logger)
```
wuxibin89: Better move `GPUMemoryLogger` into `DistProfiler` to keep decorator concise.

wuxibin89 (changes requested):

verl/workers/engine_workers.py
wlhgtc: Thanks for the suggestion. I’ve made the change and will run another test tomorrow.

I don’t think `Actor Engine` is an appropriate role name here, and having two role-like parameters in the same decorator feels a bit awkward.
I changed the memory log to omit the role prefix, which seems cleaner to me.
 For example: `Before compute_log_prob, memory allocated ...`

Does this look reasonable to you?

verl/utils/profiler/profile.py
```diff
@@ -168,11 +169,21 @@ def annotate(
         color: Optional[str] = None,
         domain: Optional[str] = None,
         category: Optional[str] = None,
+        log_gpu_memory: bool = False,
+        logger: Optional[logging.Logger] = None,
         **kwargs_outer,
     ) -> Callable:
         def decorator(func):
+            gpu_memory_logger = None
```
tardis-key: DistProfiler is the base class. I don’t think it’s appropriate to implement specific GPU collection logic inside its functions.

tardis-key: I think what @wuxibin89  means is that you should implement this feature by creating a derived class of DistProfiler instead of modifying the base class directly. Am I right?

wlhgtc: Thanks for the suggestion. I understand that you are suggesting turning `GPUMemoryLogger` into a formal profiler backend, but I think that would change the expected behavior.

The existing profiler backends are generally controlled by the profiler config and the `start_profile()` / `stop_profile()` lifecycle. In contrast, `GPUMemoryLogger` is a lightweight logger-based diagnostic: it is expected to emit logs whenever `VERL_LOGGING_LEVEL=DEBUG` is enabled, and to skip memory collection otherwise.

The behavior I’m trying to restore is the old DEBUG-level actor memory logging behavior, not a profiling session controlled by profiler steps. So I think making `GPUMemoryLogger` a formal profiler backend would not quite match the original semantics.

Given that, I feel the initial approach with two separate decorators may actually be cleaner, since it keeps the profiler annotation and the logger-based memory diagnostic as two distinct concerns.

wlhgtc:
@wuxibin89  I've provided two alternative approaches in this PR. Could you please take a look and let me know which one you prefer? Thanks!

verl/utils/profiler/profile.py
```diff
@@ -194,6 +205,12 @@ def wrapper(self_instance, *args, **kwargs_inner):
                         return func(self_instance, *args, **kwargs_inner)
                 return func(self_instance, *args, **kwargs_inner)
 
+            @functools.wraps(func)
+            def wrapper(self_instance, *args, **kwargs_inner):
+                if gpu_memory_logger is not None:
+                    return gpu_memory_logger.log(profiled_func, self_instance, *args, **kwargs_inner)
```
ETOgaosion: Seems that there is a logical issue, when enable gpu_memory_logger, profiler will not take effect

wlhgtc: I think the profiler should still take effect here. @ETOgaosion 

The function passed into `GPUMemoryLogger.log()` is `profiled_func`, not the original `func`. So `profiled_func` is actually called inside `gpu_memory_logger.log(profiled_func, ...)`.

The call chain is:

`wrapper -> GPUMemoryLogger.log(profiled_func, ...) -> profiled_func -> original func`

Even when GPU memory logging is skipped because DEBUG logging is disabled, `GPUMemoryLogger.log()` returns `func(*args, **kwargs)`, where `func` is still `profiled_func`. Therefore the original `DistProfiler` annotation path should still be executed.

Please let me know if I missed another path where this could bypass `profiled_func`.

tardis-key (changes requested):
I still maintain my view that gpumemory is a common profiler requirement, just like other derived classes of DistProfiler. It should not be directly written into the base class logic of DistProfiler, nor should separate additional interfaces be created specifically for gpu memory. Doing so would lead to code corruption and potential performance risks.

Currently, GPUMemoryLogger is controlled via environment variables. On one hand, we can consider optimizing the control method; on the other hand, without changing the existing logic, we can refer to the DistProfiler end-to-end (e2e) collection approach (rather than the discrete mode), which can also meet the collection requirements of GPUMemoryLogger.

tardis-key:
This is a schematic diagram of the e2e and discrete modes.
image
If you want to get rid of step control, you can provide an all_steps configuration similar to all_ranks to enable it for all steps. If you want to ignore the annotation, you can simply implement start and stop and collect data via e2e mode.

ETOgaosion:
Agree with @tardis-key , traditionally, there are 3 ways of memory logger: a decorate wrapper around 1 function, a with wrapper around some codes, or just a log message inserted into codes.

I think we can still maintain these utils, and for extendability, we may design carefully to support detecting memory of CPU/GPU/NPU/AMD backends.

wlhgtc:
Thanks for the explanation.  
I agree that GPU memory collection is a common profiling requirement. Verl already has memory-related profiling paths, such as `torch_memory` for allocation history / memory snapshot artifacts and torch profiler memory collection for op-level memory events.

However, these are generally controlled by profiler config and collected within a step/window-level profiling session, while this PR is trying to restore the previous stage/function-level memory logging within a step, such as before/after `compute_log_prob` and `update_actor`. Using e2e collection could work if we place `start/stop` around each target function, but that would become very close to a wrapper/decorator again.  Please let me know if you have a better idea here. @tardis-key 

For this PR, I think keeping `GPUMemoryLogger` as a separate function decorator is a smaller and more faithful restoration of the previous behavior.

tardis-key:
> 
> However, these are generally controlled by profiler config and collected within a step/window-level profiling session, while this PR is trying to restore the previous stage/function-level memory logging within a step, such as before/after `compute_log_prob` and `update_actor`. Using e2e collection could work if we place `start/stop` around each target function, but that would become very close to a wrapper/decorator again. Please let me know if you have a better idea here. @tardis-key

https://github.com/verl-project/verl/blob/74cebc50d424d91a0de967291cf070907c05a47b/verl/workers/engine_workers.py#L634-L654
If you are trying to monitor memory performance before/after `compute_log_prob` and `update_actor`, discrete mode suits you better.
https://github.com/verl-project/verl/blob/74cebc50d424d91a0de967291cf070907c05a47b/verl/utils/profiler/mstx_profile.py#L207

```python
if role == 'compute_log_prob' or role == 'update_actor':
    start_memory_profiler()
    result = func(*args, **kwargs_inner)
    stop_memory_profiler()
else:
    result = func(*args, **kwargs_inner)
```

PR #6417, opened 2026-05-20
[BREAKING][worker, ckpt] refactor: rename ActorRolloutRefWorker.update_weights to actor_weights_to_rollout

wengeezhang:
### What does this PR do?

Rename `ActorRolloutRefWorker.update_weights` to `actor_weights_to_rollout` to remove the
name collision with the rollout-side `update_weights` (e.g. `SGLangRollout.update_weights`,
`VLLMRollout.update_weights`, `TRTLLMRollout.update_weights`) and with the higher-level
`CheckpointManager.update_weights` wrapper.

Before this PR there were three same-named methods sitting on adjacent layers, only
distinguishable by argument signature (`mode=...` on the trainer side, `weights=...` on
the rollout side). That made stack traces, grep-driven navigation, and code review
needlessly ambiguous — especially in `checkpoint_engine/base.py:478-479` where
`trainer.update_weights(...)` and `rollout.update_weights(...)` are called on the same
line. The new name makes the direction of the weight flow explicit (actor → rollout).

No behavior change: pure rename + callers + docstrings/comments.

### Checklist Before Starting

  https://github.com/verl-project/verl/pulls?q=is%3Apr+update_weights+rename
  (no open PR proposes the same rename; this PR does not duplicate other in-flight work)
- [x] `[BREAKING]` prefix added — the renamed method is decorated with
      `@register(dispatch_mode=Dispatch.ONE_TO_ALL, blocking=False)` and is therefore part
      of the trainer-side RPC surface; downstream checkpoint-engine integrations that
      call `trainer.update_weights(global_steps=..., mode=...)` must be updated.

### Test

CI-only refactor; no algorithmic change to verify. Validated by:

1. Static check that all in-tree call sites of the trainer-side method are migrated:

### DIFF
--- a/tests/checkpoint_engine/test_utils.py
+++ b/tests/checkpoint_engine/test_utils.py
@@ -40,7 +40,7 @@ class TrainingWorkerTest(TrainingWorker):
         self.checkpoint_engine = CheckpointEngineRegistry.new(backend, bucket_size=bucket_size, **engine_kwargs)
 
     @register(dispatch_mode=Dispatch.ONE_TO_ALL, blocking=False)
-    async def update_weights(self, global_steps: int = None, mode: str = "auto"):
+    async def actor_weights_to_rollout(self, global_steps: int = None, mode: str = "auto"):
         per_tensor_param, _ = self.engine.get_per_tensor_param()
         await self.checkpoint_engine.send_weights(per_tensor_param)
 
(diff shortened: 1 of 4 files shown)


tardis-key:
update_weights is a universal naming convention across all RL frameworks and has no ambiguity.

ETOgaosion:
As @tardis-key said, for the same naming, there may cause confusing sometimes, but here, it's a universal API for multiple backends to implement, and it is OK that the RL framework side implement it, the usage scenario has no ambiguity. Close it first.

PR #6422, opened 2026-05-20
[trainer] feat: generic remote backend abstraction for RL Training

sfc-gh-truwase:
### What does this PR do?

- Creates a generic Remote Backend abstraction
- ArcticRL instance of the remote backend

### Checklist Before Starting

### Test

### API and Usage Example

```python
```

### Design & Code Changes

> [!IMPORTANT]


zw0610:
Could you rename this pull request and to reflect the general-purpose of the code change?

wuxibin89:
@sfc-gh-truwase Could you please provide a RFC design for this?

verl/trainer/config/ppo_trainer.yaml
```diff
@@ -310,3 +316,40 @@ ray_kwargs:
...
   timeline_json_file: null
+
+
+# Per-backend configuration for `verl.remote_backend` adapters. The backend
+# is selected by `trainer.remote_backend` (a string registered in
+# `RemoteBackendRegistry`); each backend nests its config under
+# `remote_backend.<name>`. To plug in a new backend (e.g. "tinker"),
+# add a sibling block here and register the adapter — no changes needed
+# to the rest of this file or `main_ppo.py`.
+remote_backend:
```
zw0610: Instead of adding the configuration for remote backend into the `ppo_trainer.yaml` config file, I would suggest let user to choose a `remote_backend.yaml` as the value for `remote_backend`. Just like:
```yaml
# to disable remote_backend
remote_backend: null

# to use arctic rl as the remote_backend
remote_backend: xx/xxx/arctic_rl.yaml

# to use tinker as the remote_backend
remote_backend: xx/xxx/tinker.yaml
```
As the configuration for different remote backend could differ quite a lot, it seems not a good idea to fix its format in the `ppo_trainer.yaml`.

verl/workers/remote_client/arctic_rl.py
zw0610, line 1: Could you move this file to `verl/workers/remote_client`?

verl/workers/rollout/remote_rollout/arctic_rollout/arctic_rollout.py
zw0610, line 1: Could you move this `verl/workers/rollout/arctic_rollout` directory to `verl/workers/rollout/remote_rollout/arctic_rollout`

verl/workers/rollout/replica.py
```diff
@@ -348,11 +348,16 @@ def _load_trtllm():
...
+def _load_arctic():
+    from verl.workers.rollout.arctic_rollout.arctic_rollout import ArcticReplica
+    
+    return ArcticReplica
 
 # Register built-in types
 RolloutReplicaRegistry.register("vllm", _load_vllm)
 RolloutReplicaRegistry.register("sglang", _load_sglang)
 RolloutReplicaRegistry.register("trtllm", _load_trtllm)
+RolloutReplicaRegistry.register("arctic", _load_arctic)
```
zw0610: Please add a TODO comment with my name tag here for lazy-init each option for `RolloutReplicaRegistry`. For most cases, there will be only 1 rollout backend used but import all of them could make dependency conflict in the future.

verl/remote_backend/base.py
```diff
@@ -0,0 +1,248 @@
...
+    # ------------------------------------------------------------------ #
+    # Core RL operations called by `RemoteBackendActorRolloutRefWorker`
+    # ------------------------------------------------------------------ #
+    # Inputs are verl `TensorDict`s; outputs are plain Python dicts that
+    # the worker wraps / augments with FlopsCounter MFU before returning a
+    # `TensorDict` to the trainer. Backends pick their own wire format
+    # below this layer.
+
+    @abc.abstractmethod
+    async def compute_log_prob(
```
zw0610: It seems a bit redundant to specify the abstract api for `update_actor`, `compute_log_prob` and other non-lifecycle methods. I would suggest remove them and put the implementation of each method directly into the inherited `RemoteBackendActorRolloutRefWorker` such like `ArcticRLRemoteWorker` or `TinkerRemoteWorker`. I would like to provide sufficient versatility to different backends in case of, for example, using one remote backend for sampling and another backend for training.

verl/remote_backend/base.py
```diff
@@ -0,0 +1,248 @@
...
+    """
+
+    _backends: dict[str, type[RemoteBackend]] = {}
+
+    # Backend name → dotted module path to import on first use. Ray
+    # child procs (and the driver) only need to know the name; the
+    # registry resolves to the right adapter module without forcing
+    # `main_ppo.py` to grow a per-backend `import` line.
+    MODULES: dict[str, str] = {
+        "arctic": "verl.trainer.ppo.arctic_rl_client",
```
zw0610: We can leave the `MODULES` as empty and let the registration happens when user explicitly initialize with specific configuration.

verl/remote_backend/worker.py
```diff
@@ -0,0 +1,234 @@
...
+from verl.utils.profiler import DistProfiler, DistProfilerExtension
+from verl.utils.py_functional import append_to_dict
+from verl.workers.config import ActorConfig, HFModelConfig
+
+
+# ---------------------------------------------------------------------- #
+# Generic forwarder worker
+# ---------------------------------------------------------------------- #
+
+class RemoteBackendActorRolloutRefWorker(Worker, DistProfilerExtension):
```
zw0610: Re-implement methods like `compute_ref_log_prob`,  `update_actor` and other methods here, especially not inheriting from `ActorRolloutRefWorker` and overriding methods but creating a parallel Worker class to `ActorRolloutRefWorker` could be dangerous for future development. Once the code in `ActorRolloutRefWorker` is changed, we need to double-check the implementation of `RemoteBackendActorRolloutRefWorker` as well. 

I would suggest:
1.  Make a directory of `workers` under `remote_backend`
2. For each implementation, say `arctic_rl` and `tinker`, make a directory for it respectively
3. Inherit from `TrainingWorker` and `BaseRollout` as `RemoteTrainingWorker` and `RemoteRollout` and overrides methods if necessary
4. Inherit from `ActorRolloutRefWorker` as `RemoteActorRolloutRefWorker` and only overrides the `init_model` part (maybe with other methods if necessary. But please keep them unchanged as more as possible)
5. When the PPOTrainer initializing remote backend, use corresponding implementation of `RemoteActorRolloutRefWorker` when passing it to form a WorkerGroup

verl/trainer/config/ppo_trainer.yaml
sfc-gh-kganesan: Addressed in [Snowflake-AI-Research/verl#4](https://github.com/Snowflake-AI-Research/verl/pull/4) (stacked on #3). The arctic block is now in its own `verl/trainer/config/remote_backend/arctic.yaml`; `ppo_trainer.yaml` only carries a Hydra optional defaults entry (`- optional remote_backend@remote_backend: null`). Users opt in with the standard Hydra option-group override `remote_backend=arctic` (which closely matches the form you suggested). Adding a new backend = drop a new file under `trainer/config/remote_backend/`, no edits to `ppo_trainer.yaml`.

verl/workers/remote_client/arctic_rl.py
sfc-gh-kganesan: Done in [Snowflake-AI-Research/verl#4](https://github.com/Snowflake-AI-Research/verl/pull/4): `git mv verl/trainer/ppo/arctic_rl_client.py verl/workers/remote_client/arctic_rl_client.py` plus a new `verl/workers/remote_client/__init__.py`. Imports updated in `arctic_rollout.py` and `main_ppo.py`.

verl/workers/rollout/remote_rollout/arctic_rollout/arctic_rollout.py
sfc-gh-kganesan: Done in [Snowflake-AI-Research/verl#4](https://github.com/Snowflake-AI-Research/verl/pull/4): `git mv verl/workers/rollout/arctic_rollout verl/workers/rollout/remote_rollout/arctic_rollout` plus a new `verl/workers/rollout/remote_rollout/__init__.py`. `replica.py`'s `_load_arctic` updated to the new path.

verl/workers/rollout/replica.py
sfc-gh-kganesan: Added in [Snowflake-AI-Research/verl#4](https://github.com/Snowflake-AI-Research/verl/pull/4). The TODO tags you (`# TODO(@zw0610): ...`) above the `RolloutReplicaRegistry.register(...)` block and notes that the `_load_*` callables already defer the heavy imports — the remaining work is to move the `.register(...)` calls themselves into per-backend explicit `register_<name>()` helpers so importing `replica.py` doesn't even enumerate them.

verl/remote_backend/base.py
sfc-gh-kganesan: Done in [Snowflake-AI-Research/verl#4](https://github.com/Snowflake-AI-Research/verl/pull/4). The `RemoteBackend` ABC is now strictly lifecycle + checkpoint + weight-sync: `from_config`, `reconnect_handle`, `destroy`, `update_weights`, `save_checkpoint`, `requires_single_forwarder`. `compute_log_prob` / `update_actor` / `generate` are no longer on the base class; they live on the Arctic adapter as plain concrete methods and are called from the Arctic-specific worker (not via the ABC contract). That preserves the flexibility you mentioned — a future backend can split sampling vs training into two adapters without growing this base class.

verl/remote_backend/base.py
sfc-gh-kganesan: Done in [Snowflake-AI-Research/verl#4](https://github.com/Snowflake-AI-Research/verl/pull/4). The lazy `MODULES` table is gone from `RemoteBackendRegistry`. Registration now requires an explicit adapter import: `main_ppo.py` does it on the driver when a backend is selected (`from verl.workers.remote_client import arctic_rl_client`), and the per-backend worker module re-imports its own adapter at the top so Ray child procs pick it up too (caught during the post-trim smoke). `get()` now raises a clear KeyError pointing users at the right import line. No transitive deps (vLLM, arctic-training, …) get pulled in unless that backend is actually selected.

verl/remote_backend/worker.py
sfc-gh-kganesan: Addressed *partially* in [Snowflake-AI-Research/verl#4](https://github.com/Snowflake-AI-Research/verl/pull/4) — would value your read on whether to do the rest in this PR or as a follow-up.

**Done in #4:**
- New per-backend package `verl/remote_backend/workers/arctic_rl/`. `RemoteBackendActorRolloutRefWorker` is renamed `ArcticRLActorRolloutRefWorker` so a future Tinker backend slots in as `verl/remote_backend/workers/tinker/` without sharing class state.
- `main_ppo.py` selects the per-backend worker class explicitly per `trainer.remote_backend`, with a clear error message on unknown backend names (no convention-based string mangling).

**Carve-out (not in #4):**
- The class's *parent* is still `(Worker, DistProfilerExtension)`, not `ActorRolloutRefWorker` as you suggested in step 4. The reason: `ActorRolloutRefWorker.__init__` (in `verl/workers/engine_workers.py`) reads `config.actor.megatron.router_replay.mode` unconditionally, and the Arctic config tree doesn't carry a `megatron` block. To inherit cleanly we'd first decouple that init from megatron-specific fields (or make it conditional on `config.actor.strategy`). I left it as a follow-up so this PR stays scoped to the structural cleanup; happy to extend if you want it all together.
- (Steps 1-2 from your prescription — `workers/` dir under `remote_backend` + per-backend subdirs — are done; step 3-4-5 — inherit from `TrainingWorker` / `BaseRollout` / `ActorRolloutRefWorker` and override only what's necessary — are still pending behind the decoupling above.)

**Verified**: 6-step GSM8K GRPO smoke (Qwen3-0.6B, single GPU, zorro=True) passed with MFU 198/223/213/216/198/215 — unchanged vs the pre-restructure baseline.

sfc-gh-kganesan:
@wuxibin89 RFC draft coming separately; in the meantime, [Snowflake-AI-Research/verl#4](https://github.com/Snowflake-AI-Research/verl/pull/4) implements @zw0610's structural restructure (stacked on #3), and is the cleanest snapshot of the proposed design today:

- `RemoteBackend` ABC is now lifecycle-only — no compute/update method signatures pinned on the base class, so each backend can shape its own RPCs (and a single PPO run can mix backends, e.g. one for sampling + another for training, without growing the ABC).
- Adapters live under `verl/workers/remote_client/<name>_rl_client.py`; per-backend workers under `verl/remote_backend/workers/<name>/`; per-backend config under `verl/trainer/config/remote_backend/<name>.yaml` (loaded via Hydra option group `remote_backend=<name>`).
- Registry has no eager MODULES table — adapters register via explicit import in `main_ppo.py` + the per-backend worker module, so verl never takes on transitive deps for backends a user didn't select.
- `trainer.remote_backend=<name>` is the single discriminator; `main_ppo.py` picks both the worker class and the adapter import from that.

I'll send the full RFC doc (motivation, alternatives we considered, scaling/sharding story, follow-up items like full `ActorRolloutRefWorker` inheritance) shortly — wanted to point at the running code first so the discussion is concrete.

verl/trainer/ppo/arctic_rl_client.py
```diff
@@ -0,0 +1,462 @@
...
+            "train_micro_batch_size_per_gpu": micro_batch_size,
+            "train_batch_size": train_batch_size,
+            "gradient_accumulation_steps": grad_accum_steps,
+            "sequence_parallel_size": train_seq_parallel_size,
+            "zero_optimization": {
+                "stage": zero_optimization_config.stage,
+                "offload_optimizer": {"device": zero_optimization_config.offload_optimizer.device},
+                "offload_param": {"device": zero_optimization_config.offload_param.device},
+            },
+    }
```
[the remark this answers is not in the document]
sfc-gh-sbekman: fixed upstream

verl/workers/remote_client/arctic_rl.py
```diff
@@ -0,0 +1,462 @@
...
+            vllm_config=vllm_config,
+        )
+
+        # ArcticRLClient is constructed as a ray remote actor with num_gpus=0,
+        # which causes CUDA_VISIBLE_DEVICES to be empty.
+        if colocate:
+            num_visible = n_training_gpus + n_sampling_gpus + n_log_prob_gpus
+        else:
+            num_visible = rl_config.training_gpus + rl_config.sampling_gpus + rl_config.log_prob_gpus
+        os.environ["CUDA_VISIBLE_DEVICES"] = ",".join(str(i) for i in range(num_visible))
```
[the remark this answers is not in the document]
sfc-gh-sbekman: See, the comment above:
 ```
        # ArcticRLClient is constructed as a ray remote actor with num_gpus=0,
        # which causes CUDA_VISIBLE_DEVICES to be empty.
  ```
  so this restores the intended setting.

sfc-gh-kganesan:
@wuxibin89  @zw0610 Abstract Backed RFC is here: https://github.com/verl-project/verl/issues/6537

zw0610:
@sfc-gh-truwase @sfc-gh-kganesan  Generally the status of this pull request looks good to me. Please address the CI and we are ready to merge.

sfc-gh-kganesan:
Thank you @zw0610 ! we will fix the CI at the earliest! 
Excited to bring in our contributions via this entry point :)

sfc-gh-kganesan:
Hi @zw0610 — pushed CI fixes in 23e6a36a (ran the full `pre-commit` suite locally on all 21 PR-modified files and got Passed on every hook: ruff, ruff-format, mypy, autogen-trainer-cfg, check-docstrings, check-license, compileall). The 4 GitHub Actions workflows are currently in `action_required` state (first-time-contributor approval gate). Could you click "Approve and run workflows" so CI can re-run? Thanks!

**Summary of fixes in 23e6a36a (19 files, +229/-97):**
- `ruff` / `ruff-format` auto-fixes on 9 PR-modified `.py` files (import order, quote style, line splits)
- Manual fixes for 4 errors `--fix` couldn't handle: split a 160-char list comp in `single_controller/ray/base.py`, extract `attn_implementation` to a local in `arctic_rl.py`, replace `workers: list[ActorHandle] = []` with `| None = None` (B006), and reflow a long commented debug `print()`.
- Added Apache 2.0 license header to all 12 new `.py` files we introduced.
- Regenerated `_generated_ppo_*.yaml` (3 files, +3 lines for the new `trainer.remote_backend: null` field).

No behavioural changes; just hygiene to satisfy pre-commit.

wuxibin89:
@sfc-gh-truwase Hi, could you port this PR to our new [V1 trainer](https://github.com/verl-project/verl/blob/main/verl/trainer/ppo/v1/trainer_base.py)

sfc-gh-kganesan:
@wuxibin89 can you please a take a look? 

Follow-up to the RemoteBackend abstraction ([verl-project/verl#6422](https://github.com/verl-project/verl/pull/6422)). Turns the Arctic-aware verl core into a plugin-only shape so adapters live entirely out-of-tree via the existing `VERL_USE_EXTERNAL_MODULES` hook. Zero behavioural change for verl's built-in vllm / sglang / trtllm paths.

## Why

`verl/trainer/main_ppo.py` currently hardcodes `if backend_name == "arctic": import ...` and `verl/workers/rollout/replica.py` eagerly registers `_load_arctic`. Both pull Arctic-specific modules and their transitive deps (DeepSpeed, arctic_training, arctic_inference, vLLM) into every verl process the moment a user selects `remote_backend=arctic`, and there is no way for an out-of-tree adapter to slot in without patching `main_ppo`.

## What lands

1. `RemoteBackendRegistry` (`base.py`) gains two symmetrical classmethods:
   - `register_worker(name, loader)` — plugin registers a lazy loader for the `ActorRollout(Ref)` forwarder class.
   - `get_worker(name)` — `main_ppo` resolves the class, running the loader once and caching.

2. `main_ppo.py` drops the hardcoded `if backend_name == "arctic"` branch for `RemoteBackendRegistry.get_worker(backend_name)`. Error message names the plugin contract so misconfiguration is self-explanatory.

3. `workers/rollout/replica.py` drops `_load_arctic` and its eager `register("arctic", ...)`; comment points at the plugin path.

4. `worker_utils.py` docstring notes that per-backend forwarder workers now ship in adapter packages.

5. Deletes 10 Arctic-specific files from verl core; functionality moves to the out-of-tree plugin under `arctic_platform/integrations/verl/`.

Net: **+144 / −1682 across 15 files.**

## Companion Arctic-Platform PR

Paired plugin ships as [Snowflake-AI-Research/Arctic-Platform#36](https://github.com/Snowflake-AI-Research/Arctic-Platform/pull/36) — `arctic_platform/integrations/verl/register.py` fires `@RemoteBackendRegistry.register("arctic")` on adapter import, then registers lazy loaders for the worker + replica. `import register` stays cheap; DeepSpeed / vLLM / arctic_inference only load if the backend is selected at `fit()`.

## Pre-commit

Full verl pre-commit runs clean: `ruff`, `ruff-format`, `mypy`, `check-license`, `check-docstrings`, `compileall`, `autogen-trainer-cfg`.

## Smoke tests

Both Arctic-Platform Golden Runs pass end-to-end against this tree + the paired plugin, using the shipped recipe launchers unchanged.

**Golden Run 1 — GSM8K (Qwen3-1.7B, 1×H200, 4 steps)**

| step | MFU (actor) | throughput (tok/s) | step (s) |
|---:|---:|---:|---:|
| 1 | 0.261 | 5164 | 48.4 |
| 2 | 0.256 | 5860 | 42.4 |
| 3 | 0.260 | 6283 | 39.0 |
| 4 | 0.264 | 6478 | 38.4 |

Weight sync validated every step (`[weight-sync names validated] context=cuda_ipc`).

**Golden Run 2 — BIRD text-to-SQL (Qwen3-0.6B, 1×H200, 20 steps)**

Training reward (`critic/rewards/mean`, format-bonus floor 0.1):

| window | reward | resp_len | MFU |
|---:|---:|---:|---:|
| 1–5   | 0.335 | 1116 | 0.487 |
| 6–10  | 0.322 | 1073 | 0.491 |
| 11–15 | 0.309 | 1061 | 0.484 |
| 16–20 | **0.415** |  920 | 0.507 |

Reward 0.335 → 0.415 while response length drops 1116 → 920. Validation @ step 20 vs pre-plugin reference log:

| metric | plugin + this PR | reference |
|---|---:|---:|
| `val-core/bird/reward/mean@1` | 0.279 | 0.294 |
| `val-aux/bird/execution_success/mean@1` | 0.531 | 0.522 |
| `val-aux/bird/format_correct/mean@1` | 0.960 | 0.943 |

## Follow-ups (not in this PR)

- V1 trainer port (@wuxibin89 comment on [#6422](https://github.com/verl-project/verl/pull/6422)) — separate follow-up against `main`; the V1 trainer path doesn't exist on `release/v0.7.1`.

wuxibin89 (approved):

KunWuLuan:
Hi @sfc-gh-truwase, thanks for the RemoteBackend abstraction!

We run a dual-external setup (external Megatron training + external SGLang, verl as a GPU-less orchestrator). Since the adapter landed in `arctic_platform/integrations` and `RemoteBackendTrainer` extends the deprecated `RayPPOTrainer` (while `main_ppo` now dispatches via `TaskRunnerV1`), we ported the trainer side to the V1 seam: a `@register_trainer` mode with a CPU-only forwarder pool. It builds on `verl/remote_backend/` (carried on our internal branch, not yet in upstream main) and adds no other core changes; e2e-validated with Qwen2.5-0.5B GRPO + per-step `nccl_http` weight sync.

If migrating this to the V1 trainer isn't on your roadmap right now, could we help with that upstream? Happy to share our branch and align on the design first. Thanks!

sfc-gh-truwase:
@KunWuLuan thanks for sharing your work and offering to collaborate. We are interested.  

Can you review how close #7102 is to what your branch/design? 

@wuxibin89 @sfc-gh-kganesan FYI

KunWuLuan:
Thanks @sfc-gh-truwase! We compared #7102 against our port in detail — short answer: #7102 is sufficient for us. The contract layer (RemoteBackend ABC + registry) is effectively identical to what we carry, and its V1 wiring (extension hooks, CPU-only pool, weight-sync short-circuit) covers everything our trainer-side port does. Once it lands, we'll retire our port and the inline fallback entirely, keeping only the plugin itself (external Megatron + external SGLang) wired in via register_worker.

Thanks again — and happy to help with the rebase or run an independent e2e check in our cross-cluster setup if useful.

sfc-gh-truwase:
@KunWuLuan thanks for the confirmation. I will continue our chat on #7102

Issue #6223, opened 2026-04-30
Qwen3.5/6-35B-A3B tool call format unstable in vLLM token-in-token-out rollout with `enable_thinking=False`, while Qwen3.5-9B works


Junxiao-Zhao:
### System Info

```bash
----------Python Info----------
Version      : 3.11.2
Compiler     : GCC 12.2.0
Build        : ('main', 'Apr 28 2025 14:11:48')
Arch         : ('64bit', '')
------------Pip Info-----------
Version      : 25.2
Directory    : /home/tiger/.local/lib/python3.11/site-packages/pip
vllm         : 0.20.0+cu129
sglang       : not found.
ray          : 2.46.0.14
torch        : 2.11.0
----------verl Info-----------
/usr/local/lib/python3.11/dist-packages/requests/__init__.py:113: RequestsDependencyWarning: urllib3 (1.26.20) or chardet (7.4.3)/charset_normalizer (3.4.4) doesn't match a supported version!
  warnings.warn(
Version      : 0.8.0.dev
Directory    : /opt/tiger/it_verl/verl
Commit Hash  : 91e9c3dc175226f9c99f62ac391b93c67adc4161
----------Platform Info----------
Platform     : Linux-5.10.135.bsk.6-amd64-x86_64-with-glibc2.36
system       : Linux
node         : g340-cd51-4900-be53-88e1-be50-d075
release      : 5.10.135.bsk.6-amd64
version      : #5.10.135.bsk.6 SMP Debian 5.10.135.bsk.6 Fri Aug 18 06:18:48 UT
----------Environment----------
OMP_NUM_THREADS="20"
CUDA Runtime : 12.9
CUDA Compiler : Cuda compilation tools, release 12.9, V12.9.86
----------System Info----------
CPU Memory      : 1928.46 GB
GPU Count       : 8
GPU 1   Type    : NVIDIA H20
GPU 1   Memory  : 95.58 GB
GPU 2   Type    : NVIDIA H20
GPU 2   Memory  : 95.58 GB
GPU 3   Type    : NVIDIA H20
GPU 3   Memory  : 95.58 GB
GPU 4   Type    : NVIDIA H20
GPU 4   Memory  : 95.58 GB
GPU 5   Type    : NVIDIA H20
GPU 5   Memory  : 95.58 GB
GPU 6   Type    : NVIDIA H20
GPU 6   Memory  : 95.58 GB
GPU 7   Type    : NVIDIA H20
[...]

Issue #4757, opened 2025-12-31
Invalid tool call parsing in thinking token for hermes format in tool_parser.py in agent_loop from experiment folder.


chetandhembre:
Hello
in [HermesToolParser](https://github.com/volcengine/verl/blob/bf7b056123dacdf0051981a3513b40f46b018da7/verl/experimental/agent_loop/tool_parser.py#L76) class.. `extract_tool_calls` function does eliminate content from `<think>..</think>` tags for searching for tool call signature..
this lead to case where tool call syntax used in thinking tokens consider as valid tool call. 
example (Formatting for readability purposes)
```
<think> 
this is just my reasoning.. let me call tool A 
<tool_call>{"name": "A", "arguments": {}}</tool_call> 
it will give me result which i will in my analysis
</think>
<tool_call>{"name": "A", "arguments": {}}</tool_call>
```
in above dummy example `tool_call` is used in side thinking token and out side also.. parser will consider this as two tool call.. 
I am wondering is it correct behaviour?

wzhgba:
Good catch — this also affects `Qwen3XMLToolParser` (`qwen3_coder`) in the same file.  The `_get_function_calls()` method searches for `<function=...>` patterns across the full decoded text without stripping `<think>...</think>` first, so reasoning-time tool-call speculation gets treated as a real call (same duplicate-call problem you described for Hermes).

Only `GptOssToolParser` currently handles this correctly — it strips its CoT analysis channel before running the tool-call regex.

This is actually not trivial to fix properly, because different model families use different thinking/reasoning formats:

- Qwen3.5 / DeepSeek-R1 → `<think>...</think>` + optional ` response` sentinel
- GPT-OSS → `<|channel|>analysis<|message|>...<|end|>`
- Other reasoning models may use yet other conventions

A simple regex-based workaround in each tool parser would be fragile across models.  The right long-term solution is probably a **reasoning parser abstraction** in the agent loop (similar to vLLM's `BaseThinkingReasoningParser` / `qwen3_reasoning_parser`) that separates reasoning blocks before tool parsing, so that all tool parsers automatically get clean action content regardless of the underlying model's thinking format.

Would be great to hear from the maintainers on whether this is something they'd consider adding to the agent loop pipeline.

PR #6434, opened 2026-05-21
[rollout] feat: add reasoning parser to strip think blocks before tool extraction (fixes #6424)

### LINKED ISSUE #6424: [rollout] agent loop should delegate think stripping to inference engine's native reasoning parser
## Summary

Add reasoning parser support to the agent loop's post-generation pipeline, so think-block content is excluded before tool parsing, aligning training-time behavior with inference-time chat completion processing.

## Motivation

The verl agent loop, constrained by [token-in-token-out](https://verl.readthedocs.io/en/latest/advance/agent_loop.html#chat-completion-vs-token-in-token-out), decomposes LLM interaction into: apply chat template → `generate()` → decode → tool parser. This pipeline is missing the **reasoning parser** that inference engines (vLLM/SGLang) normally apply between decode and tool parsing when serving chat completions.

Without reasoning parsing, tool call patterns inside `<think>...</think>` blocks are treated as real tool calls and executed during training. This:

- Produces spurious tool results that pollute the conversation history (see [#4757](https://github.com/verl-project/verl/issues/4757)).
- Teaches the model that loose tool call placement (within think blocks) is acceptable.
- Causes benchmark regression when evaluation harnesses apply stricter tool call parsing (see [#6223](https://github.com/verl-project/verl/issues/6223), [#6252](https://github.com/verl-project/verl/issues/6252)).

## Proposed solution

Insert a reasoning parser step between decode and tool parsing in the agent loop, reusing the inference backend's own parser through a thin adaptation layer rather than reimplementing per-model logic:

```
generate() → decode → reasoning parser (strip think) → tool parser → execute tools
```

This keeps the training path aligned with the inference path, avoids model-specific parser duplication, and improves compatibility across model families.

abinggo:
## Summary

Closes #6424 (and the three sibling issues #4757 / #6223 / #6252 that share the same root cause).

The agent loop's token-in-token-out pipeline has been `apply chat template → generate() → decode → tool parser` with no reasoning-parser step in between. Reasoning models (Qwen3 thinking, DeepSeek-R1, GLM-4 thinking) often emit `<tool_call>` patterns *inside* `<think>...</think>` blocks as part of their chain of thought. The regex-based `ToolParser` then treats those think-block tool calls as real invocations and executes them during training — polluting the conversation history (#4757), teaching the model that loose tool-call placement is acceptable, and triggering benchmark regression once eval harnesses use stricter parsing (#6223 / #6252).

This PR adds an **opt-in** reasoning-parser step that mirrors what vLLM and SGLang already apply on the chat-completion serving path:

```
generate() → decode → reasoning_parser → tool_parser → execute
```

`reasoning_parser=None` (the default) keeps the legacy behavior, so existing pipelines are unchanged.

## What changed

- **`verl/experimental/agent_loop/reasoning_parser.py`** (new): `ReasoningParser` ABC with a registry, mirroring the existing `ToolParser` design. `ThinkBlockReasoningParser` strips both fully-bracketed `<think>...</think>` (regex.DOTALL) and unclosed `<think>` truncations to end-of-text. Registered as `qwen3` and `deepseek_r1` aliases. Users register custom parsers with `@ReasoningParser.register("your-format")`.
- **`verl/workers/config/rollout.py`**: `MultiTurnConfig.reasoning_parser: Optional[str] = None`.
- **`verl/trainer/config/rollout/rollout.yaml`** + generated PPO yamls: surface the new field so `reasoning_parser=qwen3` Hydra overrides work without `+` and config dumps include it.
- **`verl/experimental/agent_loop/tool_agent_loop.py`**: `ToolAgentLoop.__init__` instantiates the parser when configured. `_handle_generating_state` decodes once, strips reasoning, re-encodes with `add_special_tokens=False` only when stripping actually changed the text, then hands the filtered ids to `tool_parser.extract_tool_calls` unchanged.
- **`tests/experimental/agent_loop/test_reasoning_parser_on_cpu.py`** (new):
  - 8 unit tests on `ReasoningParser.extract_content` (single block, newlines, plain-text passthrough, multi-block, unclosed truncation, idempotence, registry alias, unknown-name error, custom register decorator).
  - 2 integration tests with a byte-level mock tokenizer (perfect roundtrip) and a real `HermesToolParser`:
    - The think-block fake call is dropped and only the post-`</think>` real call surfaces — the exact regression #6424 / #4757 / #6223 / #6252 report.
    - With `reasoning_parser=None` (the default) the legacy behavior is preserved byte-for-byte; spurious think-block calls still execute, so the off-by-default contract is locked in.
  - All 12 tests pass locally.

## Why a tokenizer roundtrip is safe (Qwen3 / DeepSeek-R1)

`decode (skip_special_tokens=True) → strip → encode (add_special_tokens=False)` only mutates `response_ids` when the stripped text actually differs. For the target tokenizers, `<think>` and `</think>` are added tokens with `special: false` in the upstream HF configs, so `skip_special_tokens=True` does not erase them and the decode side carries the markers. The integration test uses a byte-level mock tokenizer with perfect roundtrip to lock the pipeline contract independently of any specific HF tokenizer.

`ToolParser` implementations whose decode policy is `skip_special_tokens=False` (`gpt-oss`, `gemma4`) are **not** the intended target — those models have their own COT handling inside the parser. The new feature is documented as targeting the XML / Hermes / Qwen-style parsers.

## Test plan

```
KMP_DUPLICATE_LIB_OK=TRUE PYTHONPATH=. \
    python3 -m pytest tests/experimental/agent_loop/test_reasoning_parser_on_cpu.py -v
# 12 passed

pre-commit run
# ruff / ruff-format / mypy / autogen-trainer-cfg / docstring coverage / license /
# device API / DataProto / structure / naming / example naming / compile
# -> all Passed
```

## Out of scope (intentional)

- Refactoring `ToolParser.extract_tool_calls` to take `text` directly (avoiding the small re-encode roundtrip this PR adds).
- Migrating `GptOssToolParser`'s in-parser COT strip to a registered `ReasoningParser`.

Both are reasonable follow-ups but neither is load-bearing for #6424; happy to follow up with either if maintainers prefer.

## Duplicate-work check

```bash
gh pr list --repo verl-project/verl --search "6424 in:body"           # 0
gh pr list --repo verl-project/verl --search "reasoning_parser agent_loop"  # 0
gh pr list --repo verl-project/verl --search "agent loop think block"      # 0
```


verl/experimental/agent_loop/tool_agent_loop.py
```diff
@@ -268,7 +279,22 @@ async def _handle_generating_state(
...
+            # https://github.com/verl-project/verl/issues/6424 (and #4757 /
+            # #6223 / #6252).
+            decoded_text = self.tokenizer.decode(response_ids_for_extract, skip_special_tokens=True)
+            stripped_text = self.reasoning_parser.extract_content(decoded_text)
+            if stripped_text != decoded_text:
+                # Re-encode only when stripping actually changed the text.
+                # ``add_special_tokens=False`` keeps the result aligned with
+                # what ``response_ids`` looked like before reasoning was
+                # injected (no extra BOS/EOS).
+                response_ids_for_extract = self.tokenizer.encode(stripped_text, add_special_tokens=False)
```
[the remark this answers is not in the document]
abinggo: Thanks @gemini-code-assist — both points are correct, force-pushed `32704428` addressing them:

**1. Event-loop blocking** — wrapped both `tokenizer.decode` and the conditional `tokenizer.encode` in `await self.loop.run_in_executor(None, lambda: ...)`, matching the existing pattern at lines 373 and 389 of the same file.

**2. `skip_special_tokens=False`** — also a real bug, not just a style nit. With `skip_special_tokens=True` the intermediate decode strips control tokens that downstream parsers like `gpt-oss` and `gemma4` rely on (their own `extract_tool_calls` decodes with `skip_special_tokens=False` for exactly that reason), and the subsequent re-encode with `add_special_tokens=False` cannot restore them. Switched the intermediate decode to `skip_special_tokens=False`. Because `<think>...</think>` are plain-text markers, they remain visible to `extract_content` regardless of the flag, so reasoning stripping still works while the full token-stream integrity is preserved for the tool parsers.

The integration test was also updated to mirror the production code's `skip_special_tokens=False`, and all 12 tests still pass locally; pre-commit (13 hooks) is green.

wuxibin89:
@wzhgba Could you help to review?

wzhgba:
Thanks for this PR — the problem is real (#4757, #6223, #6252, #6424) and the implementation is clean: solid ABC + registry design, 12 well-structured tests, and a safe opt-in default. A few suggestions:

**1. `enable_thinking` parameter awareness**

Qwen3.5's chat template supports `enable_thinking=False`, which injects `<think>\n\n</think>\n\n` into the prompt so model output contains no think tags. With `enable_thinking=True` and a truncated output missing `</think>`, all output should be treated as reasoning. The current parser doesn't distinguish these cases — unclosed `<think>` always strips to end-of-text, which would incorrectly discard valid content in the disabled-thinking case. vLLM and SGLang handle this via `enable_thinking` from chat template kwargs. Consider adding this parameter (could be a follow-up).

**2. Performance: `str.partition` vs regex**

For the common single-block case, `str.partition("</think>")` stops at the first occurrence (C-level, early-terminating), whereas `re.sub(r"<think>.*?</think>", ...)` scans the full text. The difference is noticeable for long reasoning outputs (e.g., 32K tokens). vLLM and SGLang both use partition/split with maxsplit=1 for this reason.

**3. Design note for future contributors**

It would be helpful to add a brief comment in `reasoning_parser.py` noting that future contributors should cross-reference vLLM/SGLang reasoning parser implementations for model-specific behaviors, and that this parser intentionally omits serving-side leniency (e.g., implicit reasoning-end via `<tool_call>`) to maintain strict formatting expectations suitable for RL training.

None of these are blockers — happy to see this land.

---

感谢这个 PR —— 解决的问题是真实存在的（#4757, #6223, #6252, #6424），实现也很干净：ABC + registry 设计合理，12 个测试覆盖充分，opt-in 默认值保证了向后兼容。几点建议：

**1. `enable_thinking` 参数感知**

Qwen3.5 的 chat template 支持 `enable_thinking=False`，会在 prompt 中注入 `<think>\n\n</think>\n\n`，使模型输出不含 think 标签；而 `enable_thinking=True` 时输出被截断后缺失 `</think>`，所有输出应视为 reasoning。当前 parser 不区分这两种情况 —— 未闭合的 `<think>` 总是剥离到文本末尾，在禁用思考的场景下会错误丢弃正常内容。vLLM 和 SGLang 通过 chat_template_kwargs 的 `enable_thinking` 参数来处理。建议加入此参数（可作 follow-up）。

**2. 性能：用 `str.partition` 替代整文 regex**

对常见的单 think block 场景，`str.partition("</think>")` 找到第一个匹配即停止（C 级实现），而 `re.sub(r"<think>.*?</think>", ...)` 需要扫描全文。对长推理输出（如 32K tokens）有可感知的差异。vLLM 和 SGLang 为此都用了 partition/split + maxsplit=1。

**3. 给后续开发者的设计注释**

建议在 `reasoning_parser.py` 中加一段注释，说明后续扩展时应参考 vLLM/SGLang 的推理 parser 实现，以及本 parser 有意省略了 serving 端的宽容处理（如 `<tool_call>` 隐式推理结束），以保持 RL 训练所需的严格格式要求。

以上都不是阻塞项，期待合入。

abinggo:
Thanks @wzhgba for the careful read — folded all three suggestions into `68d798de`.

**1. `enable_thinking` parameter awareness**

Added `enable_thinking: bool = True` keyword-only param on `ReasoningParser.extract_content` (mirroring `chat_template_kwargs`). When `False`, the parser short-circuits and returns text unchanged, so a literal `<think>` substring in code or quoted content can't trigger the unclosed-tag fallback. Default `True` preserves the legacy behavior, so existing callers (including `tool_agent_loop._handle_generating_state`) are byte-for-byte unchanged.

Threading the kwarg through `tool_agent_loop` from `chat_template_kwargs` is left as a follow-up as you suggested — happy to send a separate PR once this lands.

**2. `str.partition` replacing regex**

Rewrote `ThinkBlockReasoningParser.extract_content` as an iterative `str.partition` loop. Each iteration partitions on the next `<think>` opener and on the matching `</think>` closer; an unclosed opener breaks the loop and drops to end-of-text. Matches the partition+maxsplit=1 idiom you mentioned, and removes the `import regex` since it's no longer used. The previous non-greedy `.*?` regex still scans the full text per call — partition is C-level and early-terminating per opener, lower constant factor on 32K-token traces, same semantics.

**3. Design note for future contributors**

Added a "Design notes" section to the module docstring covering (a) cross-reference to `vllm/reasoning/` and `sglang/srt/reasoning_parser/` for model-specific behavior, (b) the intentional omission of serving-side leniency (e.g. vLLM's Qwen3 parser closes unclosed `<think>` implicitly when `<tool_call>` arrives) — RL training needs the strict surface here, (c) the `enable_thinking` short-circuit semantics.

**Tests added (3 new, on top of the existing 12)**

- `test_qwen3_short_circuits_when_thinking_disabled` — pins the new short-circuit
- `test_qwen3_disabled_thinking_preserves_unclosed_think_substring` — pins the bug you flagged: literal `<think>` substring without closer must not collapse to end-of-text when thinking is disabled
- `test_qwen3_nested_looking_tags_keep_legacy_non_overlapping_semantics` — pins the inherited non-overlapping behavior on degenerate `a<think>1<think>2</think>b</think>c → ab</think>c` so a future contributor doesn't accidentally "fix" it

15 tests pass locally; pre-commit clean. Force-pushed (`32704428 → 68d798de`).

wzhgba:
Thanks for the quick update — all three points are addressed cleanly. One more small suggestion on the API design:

`enable_thinking` is Qwen3-specific (`chat_template_kwargs`); other models may use different parameter shapes (e.g. `think_effort`). Consider using `**kwargs` at the abstract `ReasoningParser.extract_content` level, and letting each implementation pull its own keys:

```python
# abstract
def extract_content(self, text: str, **kwargs) -> str: ...

# ThinkBlockReasoningParser
def extract_content(self, text: str, **kwargs) -> str:
    if not kwargs.get("enable_thinking", True):
        return text
    ...
```

This keeps the base interface model-agnostic and allows future parsers to add their own parameters without touching the abstraction.

Again, thanks for the thorough work.

---

感谢快速更新，三条建议都落实得很干净。补充一个小建议：

`enable_thinking` 是 Qwen3 特有的参数（来自 `chat_template_kwargs`），其他模型可能用不同形态的参数（如 `think_effort` 等）。建议抽象接口用 `**kwargs`，让各实现层自己解包：

```python
# 抽象层
def extract_content(self, text: str, **kwargs) -> str: ...

# ThinkBlockReasoningParser
def extract_content(self, text: str, **kwargs) -> str:
    if not kwargs.get("enable_thinking", True):
        return text
    ...
```

这样基类接口保持模型无关，后续 parser 扩展时不需改动抽象层。

再次感谢细致的工作。

abinggo:
Thanks @wzhgba — fully agreed on keeping the abstract layer model-agnostic. Folded the kwargs design into `32ae4c67`:

**ABC signature**

- `ReasoningParser.extract_content` now takes `text: str, **kwargs` so any future model can surface its own knobs (`think_effort`, `reasoning_budget`, etc.) without touching the abstract layer.
- Docstring spells out the contract: implementations pull only the keys they understand and silently ignore unknown ones, so callers can forward `chat_template_kwargs` verbatim without per-parser conditioning.

**ThinkBlockReasoningParser**

- Reads `kwargs.get("enable_thinking", True)` at the start; default-True path unchanged for byte-for-byte legacy compatibility.

**Tests**

- Updated `_CustomParser` test fixture signature to `**kwargs`.
- Added `test_qwen3_ignores_unknown_kwargs` pinning three contracts: (a) lone unknown kwarg (`think_effort="low"`) is no-op, (b) unknown kwargs mixed with `enable_thinking=True` still strip correctly, (c) unknown kwargs don't subvert the `enable_thinking=False` short-circuit.

16 tests pass locally; pre-commit clean. Force-pushed (`68d798de → 32ae4c67`).

verl/experimental/agent_loop/tool_agent_loop.py
```diff
@@ -277,7 +279,38 @@ async def _handle_generating_state(
         # Extract tool calls (use per-sample tools if routed)
         active_tools = getattr(agent_data, "_active_tools", self.tools)
         tools = [tool.tool_schema for tool in active_tools.values()]
-        _, agent_data.tool_calls = await self.tool_parser.extract_tool_calls(agent_data.response_ids, tools)
+        response_ids_for_extract = agent_data.response_ids
+        if self.reasoning_parser is not None:
```
wuxibin89: Better to move strip reasoning logic into `extract_tool_calls`, since user may custom their own AgentLoop(e.g SWEAgentLoop, GUIAgentLoop, etc)

wuxibin89 (changes requested):

abinggo:
Thanks @wuxibin89 — folded the design change into `60b7e15a`.

**Layer move (V5)**

Rebuilt `ToolParser` as a template-method base: `extract_tool_calls` is now the public concrete entry point that calls `_maybe_strip_reasoning` (when a `reasoning_parser` was configured) and delegates to `_extract_tool_calls` for the parser-specific logic. The four in-tree parsers (`hermes` / `gpt-oss` / `qwen3_coder` / `gemma4`) renamed `extract_tool_calls` → `_extract_tool_calls`, and `ToolAgentLoop._handle_generating_state` now just calls `self.tool_parser.extract_tool_calls(response_ids, tools, **self.apply_chat_template_kwargs)` — the decode/strip/re-encode plumbing is gone. Custom `AgentLoop` subclasses (`SWEAgentLoop` / `GUIAgentLoop`) inherit reasoning-aware extraction by virtue of constructing the parser via `ToolParser.get_tool_parser(name, tokenizer, reasoning_parser=...)`.

**Backward compatibility**

Two paths preserved after a codex CR pass flagged the original draft as a real extension API break:

- `_extract_tool_calls` is *not* `@abstractmethod`. It defaults to raising `NotImplementedError`. External subclasses that only override the legacy `extract_tool_calls` directly remain valid concrete classes — they just bypass the reasoning-strip layer until they opt in.
- `ToolParser.get_tool_parser` keeps the legacy two-arg constructor path when `reasoning_parser is None`, so externally-registered parsers whose `__init__` predates the new kwarg keep working.

**`chat_template_kwargs` threading**

`ToolParser.extract_tool_calls` now accepts `**kwargs` and forwards them through `_maybe_strip_reasoning` into `ReasoningParser.extract_content`, so the `enable_thinking` short-circuit you flagged earlier reaches the strip layer end-to-end. The agent loop already had `self.apply_chat_template_kwargs` from the base class — wiring it into `extract_tool_calls` was a one-line change at the call site.

**Tests added (4 new, on top of the existing 15)**

- Updated the two integration tests to construct `HermesToolParser(tokenizer, reasoning_parser=...)` and call `extract_tool_calls` directly — they now exercise the new layer end-to-end (think-block tool call dropped vs. legacy `reasoning_parser=None` keeps spurious call).
- `test_tool_parser_skips_strip_when_no_reasoning_markers` pins the identity fast-path: when the decoded text has no `<think>` markers, `_maybe_strip_reasoning` returns the original `responses_ids` (avoiding the encode round-trip).
- `test_all_registered_tool_parsers_override_extract_tool_calls` reflection-checks that every registered parser overrides `_extract_tool_calls` or `extract_tool_calls`, so a future contributor can't silently register a parser that inherits the `NotImplementedError` default.

19 tests pass locally; pre-commit clean (ruff, mypy, autogen-trainer-cfg, etc.). Force-pushed (`32ae4c67 → 60b7e15a`).

verl/experimental/agent_loop/tool_agent_loop.py
```diff
@@ -115,7 +110,27 @@ def __init__(self, *args, tools: Optional[ToolListWrap] = None, **kwargs):
...
+        # Optional reasoning parser to strip think-block content (e.g. Qwen3
+        # ``<think>...</think>`` or DeepSeek-R1) before tool extraction so
+        # tool-call patterns inside the chain-of-thought are not executed as
+        # real calls. ``None`` (default) keeps legacy behavior. See
+        # ``verl/experimental/agent_loop/reasoning_parser.py``.
+        #
+        # The parser is handed to ``ToolParser`` rather than applied here so
+        # custom ``AgentLoop`` subclasses (``SWEAgentLoop`` / ``GUIAgentLoop``
+        # / etc.) automatically inherit reasoning-aware tool extraction
+        # without re-implementing decode → strip → re-encode plumbing.
```
wuxibin89, line 123: Could you streamline comment here? It's too verbose, the doc string in `ToolParser` and `ReasoningParser` is enough.

verl/experimental/agent_loop/tool_agent_loop.py
```diff
@@ -277,7 +289,18 @@ async def _handle_generating_state(
...
-        _, agent_data.tool_calls = await self.tool_parser.extract_tool_calls(agent_data.response_ids, tools)
+        # ``tool_parser.extract_tool_calls`` internally strips reasoning
+        # blocks (when configured) before delegating to the parser-specific
+        # extraction logic. Threading ``apply_chat_template_kwargs`` through
+        # lets the reasoning parser see model-specific knobs (e.g. Qwen3
+        # ``enable_thinking``) so a chat template that injected an empty
+        # think opener does not get incorrectly stripped. Centralizing the
+        # strip on ``ToolParser`` lets custom ``AgentLoop`` subclasses
+        # inherit reasoning-aware extraction for free; see
+        # ``ToolParser._maybe_strip_reasoning``.
```
wuxibin89, line 300: Streamline comment as well.

verl/experimental/agent_loop/reasoning_parser.py
```diff
@@ -0,0 +1,205 @@
...
+        # Pull the only key this parser cares about; other models'
+        # kwargs (e.g. ``think_effort``) are silently ignored so the
+        # caller can forward ``chat_template_kwargs`` verbatim.
+        enable_thinking = kwargs.get("enable_thinking", True)
+        if not enable_thinking:
+            # Chat template injected ``<think>\n\n</think>\n\n`` opener;
+            # any literal ``<think>`` in model output is incidental
+            # (code, quoted text) and must not trigger greedy stripping.
+            return text
+        if "<think>" not in text:
```
wyt2000: Since Qwen's tokenizer automatically appends a \<think\> tag at the end of the prompt, the response_ids do not contain the opening \<think\> tag. Does it require special handling here?

abinggo: You're right -- thanks for catching this. Pushed 03e8816f.

The Qwen3 / DeepSeek-R1 default chat template appends `<think>` to the prompt itself when `add_generation_prompt=True`, so `response_ids` (which only contain the model's generated tokens) decode to `reasoning…</think>\n\nanswer` with no leading `<think>`. My old fast path

```python
if "<think>" not in text:
    return text
```

short-circuited exactly that case and the reasoning block was passed through to the tool parser unstripped -- defeating the parser for the most common production setup. The previous unit tests fed strings with both tags explicitly so they didn't catch this.

The fix detects an *implicit opener* by comparing the positions of the two tags: when `</think>` appears before any `<think>` (or there is no `<think>` at all), strip from the start of text up to and including the first `</think>`, then continue the regular pair loop on the tail. Mirrors what vLLM's and SGLang's serving-side Qwen3 reasoning parsers do. The explicit-opener path is unchanged, and \`enable_thinking=False\` still short-circuits at the top.

Four new regression tests in \`test_reasoning_parser_on_cpu.py\`:

- implicit-opener basic strip (\`reasoning</think>\n\nanswer\` -> \`\n\nanswer\`)
- implicit-opener with inner \`<tool_call>\` dropped, post-\`</think>\` real call surfaces
- implicit-opener + a subsequent explicit \`<think>...</think>\` pair both stripped
- a response of the form \`just reasoning</think>\` collapses to \`""\`

All 22 tests pass locally; per-file pre-commit (ruff / ruff-format / mypy / sanity hooks) green.

PR #6490, opened 2026-05-27
[trainer, fsdp, rollout, model, data] feat: Qwen3-Omni-30B-A3B Thinker LoRA RL support

qinganrice:
### What does this PR do?

Adds end-to-end support for training the Qwen3-Omni-30B-A3B Thinker stage via LoRA + GSPO using verl's FSDP backend and verl-omni's `vllm_omni` rollout. 

Tested on Qwen3-Omni-30B-A3B-Instruct, 4× H100 80GB. Pipeline runs through rollout, ref-logprob, actor update, ckpt save / resume, and val.

### Checklist Before Starting

  - No existing PR addresses Qwen3-Omni training support or the MoE-wrap `layered_summon` issue.
- [x] Format the PR title as `[{modules}] {type}: {description}` (This will be checked by the CI). Title: `[trainer, fsdp, rollout, model, data] feat: Qwen3-Omni-30B-A3B RL training support`.
- [x] No `[BREAKING]` — all changes are additive or behavior-compatible (existing models / configs unaffected).

### Test

CI cannot exercise Qwen3-Omni (the model needs ~30 GB on disk and 4× H100 80GB to run end-to-end). Validated by manual smoke test:

```
# After applying this PR + corresponding verl-omni / vllm-omni PRs:
bash examples/gspo_trainer/run_qwen3_omni_thinker_gspo_lora.sh \
    trainer.total_training_steps=2 \
    data.val_max_samples=20
```

### API and Usage Example
No public API changes. New behavior is opt-in via the model class (Qwen3OmniMoeForConditionalGeneration), the rollout name (vllm_omni), and the optional _verl_strip_modules / _FalseTieDescriptor hooks on the model class — all of which only activate for Qwen3-Omni.

### Design & Code Changes
8 files touched. Each change is scoped and targeted:

verl/utils/tokenizer.py (+11)
```
Add a Qwen3OmniMoeProcessor arm to the match block in hf_processor().
Token IDs (image_token_id, vision_start_token_id, audio_token_id, position_id_per_seconds) live on thinker_config, not the top-level Qwen3OmniMoeConfig that AutoConfig returns; override processor.config = config.thinker_config so the bound get_rope_index finds them.
Expose processor.spatial_merge_size (model-level attribute, not on the config) and bind get_llm_pos_ids_for_vision (helper called by get_rope_index) so multimodal RoPE position computation works on the processor object.
```

verl/utils/model.py (+70)
```
Register Qwen3OmniMoeForConditionalGeneration with AutoModelForCausalLM. Despite the suffix, the Thinker is a decoder-only causal LM; downstream code paths (FSDP engine, LoRA target matching) all go through AutoModelForCausalLM.
Patch forward / get_input_embeddings / set_input_embeddings to delegate to self.thinker, so the full Qwen3OmniMoe...ForConditionalGeneration wrapper transparently behaves like the Thinker for training purposes.
_no_split_modules: upstream class incorrectly references Qwen3OmniMoeDecoderLayer (does not exist); set to the real class Qwen3OmniMoeThinkerTextDecoderLayer.
Add _verl_strip_modules = ["talker", "code2wav", "code_predictor"] — consumed by the FSDP engine (next file) to drop unused sub-modules and free GPU memory during Thinker post-training.
Use a _FalseTieDescriptor to override tie_word_embeddings at the config-class level: the upstream config sets it True, which forces use_meta_tensor=False in FSDP init and OOMs on 30B-A3B; the descriptor returns False while tolerating no-op writes from __init__.
All wrapped in try / except ImportError so users without the Qwen3-Omni transformers package are unaffected.
```

verl/utils/fsdp_utils.py (+88, −38)
```
get_fsdp_wrap_policy: skip the LoRA lambda policy when min_num_params > 0. Combining the size policy with the lambda policy creates nested FSDP units (LoRA leaves inside decoder-layer units) whose allgather order can diverge across ranks under variable-length inputs (use_remove_padding), causing NCCL deadlocks.
Rewrite layered_summon_lora_params to iterate FSDP units directly via named_modules() + fsdp_version(), instead of matching a fixed list of layer-level prefixes. The previous implementation assumed wrap landed at the transformer-layer level — on MoE models it typically lands at layers.<i>.mlp (because each MoE MLP exceeds min_num_params while attention does not), so the prefix list matched zero FSDP units and the function returned empty. The rewrite is wrap-granularity-agnostic and works for dense, MoE, and omni-modality models. Adds an is_diffusers parameter (currently unused on this side, threaded through for the matching verl-omni PR).
collect_lora_params: pass is_diffusers through; if layered_summon_lora_params still returns empty (e.g. unusual wrap graph), fall back to a single full summon and warn instead of silently shipping zero LoRA tensors.
```

verl/workers/engine/fsdp/transformer_impl.py (+22)
```
After auto_class.from_pretrained(...), strip any sub-modules listed in module._verl_strip_modules (e.g. talker / code2wav / code_predictor for Qwen3-Omni Thinker-only training). Without this the Talker stack is also loaded onto GPU and OOMs on 30B-A3B.
After get_peft_model(...), cast LoRA adapter params to the base model's dtype if (and only if) they actually mismatch. Mixed bf16 base + fp32 adapter trips FSDP's "all params in a flat group must share dtype" assertion. The cast is logged so users can see when it fires.
```

verl/workers/rollout/base.py (+1) and verl/workers/rollout/replica.py (+7)
```
Register ("vllm_omni", "async") in _ROLLOUT_REGISTRY.
Register vllm_omni in RolloutReplicaRegistry, pointing at the vLLMOmniReplica provided by verl-omni.
```

verl/utils/vllm/utils.py (+5, −2)
```
Detect verl-omni's OmniTensorLoRARequest via duck typing (hasattr(lora_request, "peft_config") and hasattr(lora_request, "lora_tensors")) in addition to isinstance(lora_request, TensorLoRARequest). Pre-existing isinstance flows continue to work; the new check only matters for verl-omni's own subclass.
```

verl/models/transformers/monkey_patch.py (+4, −2)
```
apply_monkey_patch: when the top-level config does not directly expose text_config, fall back to model.config.get_text_config(). Some multimodal configs (Qwen3-Omni's Qwen3OmniMoeConfig) nest text_config under thinker_config and expose it only via the helper.
```

> [!IMPORTANT]

### DIFF
--- a/verl/models/transformers/monkey_patch.py
+++ b/verl/models/transformers/monkey_patch.py
@@ -330,9 +330,11 @@ def apply_monkey_patch(
     try:
         num_attention_heads, num_key_value_heads = model.config.num_attention_heads, model.config.num_key_value_heads
     except AttributeError:
+        # Some multimodal configs nest text_config differently; fall back to get_text_config().
+        text_config = getattr(model.config, "text_config", None) or model.config.get_text_config()
         num_attention_heads, num_key_value_heads = (
-            model.config.text_config.num_attention_heads,
-            model.config.text_config.num_key_value_heads,
+            text_config.num_attention_heads,
+            text_config.num_key_value_heads,
         )
 
     assert num_attention_heads % ulysses_sp_size == 0, (
--- a/verl/utils/fsdp_utils.py
+++ b/verl/utils/fsdp_utils.py
@@ -109,8 +109,11 @@ def get_fsdp_wrap_policy(module, config=None, is_lora=False):
 
     from torch.distributed.fsdp.wrap import _or_policy, lambda_auto_wrap_policy
 
-    # Add lambda policy for LoRA modules if is_lora is True
-    if is_lora:
+    # Add lambda policy for LoRA modules if is_lora is True. Skip when
+    # min_num_params > 0: combining size policy with lambda policy creates
+    # nested FSDP units whose allgather order can diverge across ranks under
+    # variable input lengths (use_remove_padding), causing NCCL deadlocks.
+    if is_lora and min_num_params == 0:
 
         def lambda_policy_fn(module):
             return bool(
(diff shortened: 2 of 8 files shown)


qinganrice:
Will move to verl-omni, so close this PR.

PR #6508, opened 2026-05-27
[reward, worker] fix: guard reward_tensor index -1 when valid_response_length == 0

### LINKED ISSUE #6476: [Bug] Reward manager writes reward to index -1 when `valid_response_length == 0`
## System Info

Discovered via static code analysis. Affects `prime.py`, `naive.py`, and `dapo.py` reward managers.

## Information

## Tasks

## Reproduction

In `verl/workers/reward_manager/prime.py` line 177 (identical pattern in `naive.py:100` and `dapo.py:132`):

```python
reward_tensor[i, valid_response_length[i].item() - 1] = scores[i]
```

When `valid_response_length[i] == 0` (empty generation, aborted rollout, or edge-case input), this becomes:

```python
reward_tensor[i, -1] = scores[i]  # writes to the LAST element of the row
```

Python/PyTorch silently accepts negative indices, so the reward is written to the wrong position with no error raised.

## Expected behavior

When `valid_response_length[i] == 0` there is no valid response token to attach a reward to. The code should either skip the assignment or raise an explicit error:

```python
if valid_response_length[i].item() > 0:
    reward_tensor[i, valid_response_length[i].item() - 1] = scores[i]
```

## Impact

- Silent data corruption — reward placed at the wrong position, corrupting training signal.
- Affects all three reward managers: `prime.py`, `naive.py`, `dapo.py`.
- Triggered by empty/aborted generations which can occur during early training or with constrained sampling.

## Related

Also in `dapo.py` line 121: `self.overlong_buffer_cfg.enable` crashes with `AttributeError: 'NoneType' object has no attribute 'enable'` when `overlong_buffer_cfg=None` (its default value). The fix is:

```python
```

donghyq:
## What does this PR do?

Fixes #6476

Fixes two bugs in the reward managers triggered when `valid_response_length == 0` (e.g. an aborted or zero-token generation).

**Bug 1 — silent index -1 write (naive.py, prime.py, dapo.py)**

All three reward managers contain:

```python
reward_tensor[i, valid_response_length - 1] = reward
```

When valid_response_length == 0, this evaluates to reward_tensor[i, -1], which PyTorch silently treats as the last token position rather than raising an error. The reward is written to a wrong position, corrupting the training signal for that sample.

Fix: skip the reward assignment when valid_response_length == 0.

if valid_response_length > 0:
    reward_tensor[i, valid_response_length - 1] = reward

**Bug 2 — AttributeError when overlong_buffer_cfg=None (dapo.py)**

DapoRewardManager.__call__ accesses self.overlong_buffer_cfg.enable unconditionally. Since overlong_buffer_cfg=None is the default, this raises AttributeError: 'NoneType' object has no attribute 'enable' for any user who does not configure overlong penalty.

Fix: add a None guard consistent with the one already present in __init__.

```python
```

### Checklist Before Starting

  - [x] Search for similar PRs: https://github.com/volcengine/verl/pulls?q=valid_response_length — no existing PR addresses these
   two bugs.
  - [x] PR title formatted as [reward, worker] fix: ... ✅

### Tests

Added CPU-only regression tests in tests/workers/reward_manager/test_reward_manager_zero_length_on_cpu.py

covering all three managers:

- Zero-length response → reward_tensor stays all-zero (bug 1 regression)
- Last token position not written for zero-length response
- Normal response still receives reward at the correct position (regression)
- Mixed batch (zero-length + normal) — samples do not interfere
- overlong_buffer_cfg=None does not raise AttributeError (bug 2 regression)

### API and Usage Example

  No API changes. The fix is purely defensive — behavior is unchanged for valid inputs.

  Design & Code Changes

  - verl/workers/reward_manager/naive.py: wrap reward assignment in if valid_response_length > 0
  - verl/workers/reward_manager/prime.py: same guard for batched path
  - verl/workers/reward_manager/dapo.py: same guard + add self.overlong_buffer_cfg is not None and before .enable
  - tests/workers/reward_manager/test_reward_manager_zero_length_on_cpu.py: new CPU regression test file

  - [x] Applied pre-commit checks — all passed.
  - [x] Documentation: no doc changes needed (bug fix only).
  - [x] Unit tests added (test_reward_manager_zero_length_on_cpu.py), picked up automatically by cpu_unit_tests CI.
  - [ ] Send message in ci-request channel / Feishu group after PR is open.
  - [x] Not related to recipe submodule.

### DIFF
--- a/tests/workers/reward_manager/test_reward_manager_zero_length_on_cpu.py
+++ b/tests/workers/reward_manager/test_reward_manager_zero_length_on_cpu.py
@@ -1,0 +1,234 @@
+# Copyright 2025 Bytedance Ltd. and/or its affiliates
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Regression tests for reward manager index-out-of-bounds when valid_response_length == 0.
+
+When a generation is aborted or truncated to zero tokens, `valid_response_length`
+is 0 and `reward_tensor[i, valid_response_length - 1]` silently writes to index
+-1 (the last token), corrupting training signal.
+
+Fix: skip the reward assignment when `valid_response_length == 0`.
+
+A secondary bug in `dapo.py`: `self.overlong_buffer_cfg.enable` is accessed
+unconditionally, crashing with `AttributeError` when `overlong_buffer_cfg=None`
+(the default).
+
+See https://github.com/volcengine/verl/issues/6476
+"""
+
+from unittest.mock import MagicMock
+
+import numpy as np
+import torch
+from tensordict import TensorDict
+
+from verl import DataProto
+
+# ---------------------------------------------------------------------------
+# Helpers
+# ---------------------------------------------------------------------------
+
+
+def _make_data_proto(batch_size: int, prompt_len: int, resp_len: int, zero_mask_indices=None):
+    """Build a minimal DataProto with configurable attention mask.
+
+    Args:
+        batch_size: number of samples in the batch.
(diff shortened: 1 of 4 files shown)


wuxibin89:
> When valid_response_length == 0, this evaluates to reward_tensor[i, -1], which PyTorch silently treats as the last token position rather than raising an error. The reward is written to a wrong position, corrupting the training signal for that sample.

When valid_response_length == 0, it's ok to assign reward to last token.

PR #6510, opened 2026-05-27
fix: restore rollout state after checkpoint update failures

athreesh:
## Summary
- finalize checkpoint engines from CheckpointEngineManager cleanup even when process-group build or weight transfer fails
- always attempt to restore KV cache and resume generation after non-naive checkpoint update interruptions
- add a CPU unit test covering build failure, transfer failure, cleanup failure, and primary-error preservation

## Tests
- python3 -m py_compile verl/checkpoint_engine/base.py tests/checkpoint_engine/test_manager_cleanup_on_cpu.py
- PYTHONPATH=/home/ubuntu/verl /home/ubuntu/checkpoint-engine/.venv/bin/python -m pytest tests/checkpoint_engine/test_manager_cleanup_on_cpu.py -q
- uv run --python /usr/bin/python3.10 --no-project --with ruff ruff check verl/checkpoint_engine/base.py tests/checkpoint_engine/test_manager_cleanup_on_cpu.py
- git diff --check

### DIFF
--- a/tests/checkpoint_engine/test_manager_cleanup_on_cpu.py
+++ b/tests/checkpoint_engine/test_manager_cleanup_on_cpu.py
@@ -1,0 +1,183 @@
+# Copyright 2024 Bytedance Ltd. and/or its affiliates
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


wuxibin89:
Checkpoint update failures is an unrecoverable failure and should restart job.

Issue #804, opened 2025-03-28
[Question] A question about qwen2.5-vl as critic model in ppo


SportsSan:
If I use qwen2.5-vl as a critic model for ppo training, should I set my own model? Does the code have a compatible plan?

Kirrito-k423:
中文：

如果你的目标是用 Qwen2.5-VL 做 PPO 的 critic，当前建议优先走 FSDP/FSDP2 路径，并显式设置 `critic.model.path=Qwen/Qwen2.5-VL-7B-Instruct`（或你的本地 Qwen2.5-VL checkpoint），同时让 actor/rollout 使用相应的 VLM 配置和多模态数据字段。当前代码的 FSDP value-model 加载路径会先尝试 `AutoModelForTokenClassification(num_labels=1)`；如果该模型不支持 token-classification，会 fallback 到 TRL value-head 路径，并且对 VLM 使用 `AutoModelForVision2Seq`，因此标准 Hugging Face Qwen2.5-VL 不一定需要你自己实现一个全新的 critic model。

需要注意两个边界：

1. 仓库里已经有 Qwen2.5-VL 的多模态 GRPO/FSDP 示例，但我没有看到 Qwen2.5-VL 作为 PPO critic 的公开端到端示例或基准结果，所以这条路径建议先做小规模 smoke test。
2. 如果你想用 Megatron 路径，当前 value-model registry 只覆盖若干文本 CausalLM 架构，没有看到 Qwen2.5-VL value model 的注册；这种情况下可能需要额外模型适配/注册，而不是只改配置。

English:

If you want to use Qwen2.5-VL as the PPO critic, the safest current path is FSDP/FSDP2: set `critic.model.path=Qwen/Qwen2.5-VL-7B-Instruct` (or your local Qwen2.5-VL checkpoint) explicitly, and keep the actor/rollout side configured for VLM data and multimodal inputs. In the current FSDP value-model loader, verl first tries `AutoModelForTokenClassification(num_labels=1)`; if that is not available, it falls back to a TRL value-head path, using `AutoModelForVision2Seq` for VLM configs. So for a standard Hugging Face Qwen2.5-VL checkpoint, you should not necessarily need to implement a completely custom critic model.

Two caveats:

1. The repo has Qwen2.5-VL multimodal GRPO/FSDP examples, but I did not find a public end-to-end PPO + Qwen2.5-VL critic recipe or benchmark, so please start with a small smoke test.
[...]

PR #6511, opened 2026-05-27
[veomni, fsdp] feat: enable fused top-K distillation kernel for OPD

Luosuu:
### What does this PR do?

Wires veomni's `chunk_topk_distill_function` into the verl actor
pipeline so on-policy distillation (OPD) batches can run with
`use_fused_kernels=True`. The fused kernel computes the top-K
forward-KL distillation loss without materializing the full
`[B, L, V]` logits tensor, saving significant GPU memory for
large-vocabulary models.

Previously, distillation TDs had to force `use_fused_kernels=False`
because the fused path only produced per-token log_probs/entropy —
the in-forward logit processor needed full logits to compute the
top-K weighted CE. With veomni 0.1.11's `chunk_topk_distill_function`,
the distillation loss is computed inline during the chunked
fused-linear forward pass.

### Design & Code Changes

Engine changes (two files, additive — no existing behavior modified):

- **`VeOmniEngineWithLMHead.prepare_model_inputs`**: when
  `use_fused_kernels=True`, `distillation_use_topk=True`, and
  `teacher_topk_ids` exists in the micro-batch, flattens the nested
  teacher tensors to `(1, total_nnz, K)` and passes them into
  `model_inputs`. They propagate as `**kwargs` through the model
  forward to `ForCausalLMLoss`, which routes to
  `chunk_topk_distill_function`. Includes paired validation (raises
  `ValueError` if `teacher_topk_ids` present without
  `teacher_topk_log_probs`). Guards on key existence so verl's
  native distillation path (different tensor keys) is unaffected.

- **`FSDPEngineWithLMHead.prepare_model_outputs`**: in the
  `use_fused_kernels=True` block, extracts `distillation_losses` /
  `student_mass` / `teacher_mass` from `output.fused_linear_aux` and
  stores them as nested tensors in `model_output` — same keys and
  format as the eager logit-processor path, so downstream loss
  functions work unchanged.

Both paths (fused and eager) remain fully functional:
- **Fused** (`use_fused_kernels=True`, VeOmni engine): `ForCausalLMLoss`
  routes to `chunk_topk_distill_function`.
- **Eager** (`use_fused_kernels=False`, FSDP/Megatron): existing
  in-forward logit processor fires as before.

CI: bump veomni from 0.1.10 to 0.1.11 (adds the `chunk_topk_distill`
kernel).

### Test

- **2xH100 unit test with Qwen3-0.6B**: synthetic distillation TD with
  `use_fused_kernels=True` and `distillation_use_topk=True`. Verified
  `distillation_losses`, `student_mass`, `teacher_mass` present as
  nested tensors in `model_output`, all values finite and non-negative,
  both ranks produce consistent output. Confirms our new engine code
  paths (input injection, output extraction) work correctly.

- **Full e2e example** (`run_qwen3_8b_mopd_veomni.sh` /
  `run_qwen3_0.6b_opd_veomni.sh` with verl native distillation):
  reaches actor backward on H100x8 then hits a pre-existing veomni
  `chunk_logprobs` + FSDP2 bug — see
  https://github.com/ByteDance-Seed/VeOmni/issues/804. The bug is in
  veomni's existing fused log_probs kernel (not added by this PR) and
  fires whenever `use_fused_kernels=True` is set for any VeOmni FSDP2
  actor with a backward pass, regardless of distillation. Our engine
  changes are downstream of this bug — verified path-clean by the
  unit test above. E2E example will run end-to-end once veomni #804
  lands.

### Dependencies

Requires veomni >= 0.1.11 for `chunk_topk_distill_function` and
`FusedLinearAuxOutput.distillation_losses`.

### DIFF
--- a/verl/workers/engine/fsdp/transformer_impl.py
+++ b/verl/workers/engine/fsdp/transformer_impl.py
@@ -1092,6 +1092,21 @@ class FSDPEngineWithLMHead(FSDPEngine):
                 # temperature is singleton
                 log_probs = output.log_probs.squeeze(0)  # (total_nnz,)
                 entropy_rmpad = output.entropy.squeeze(0)  # (total_nnz,)
+
+                # When the fused kernel also computed top-K distillation
+                # (veomni's chunk_topk_distill path), extract the per-token
+                # distillation outputs and store them as nested tensors —
+                # same model_output keys as the eager logit-processor path.
+                if distillation_use_topk:
+                    aux_outputs = getattr(output, "fused_linear_aux", None)
+                    if aux_outputs is not None and aux_outputs.distillation_losses is not None:
+                        cu_seqlens = input_ids.offsets()
+                        for field_name in ("distillation_losses", "student_mass", "teacher_mass"):
+                            v = getattr(aux_outputs, field_name).squeeze(0)
+                            if self.use_ulysses_sp:
+                                pad_size = output_args["pad_size"]
+                                v = gather_outputs_and_unpad(v, gather_dim=0, unpad_dim=0, padding_size=pad_size)
+                            model_output[field_name] = torch.nested.nested_tensor_from_jagged(v, cu_seqlens)
             else:
                 logits_rmpad = output.logits.squeeze(0)  # (total_nnz, vocab_size)
                 logits_rmpad.div_(temperature_rmpad.clamp(min=1e-8).unsqueeze(-1).to(logits_rmpad.dtype))
--- a/.github/workflows/e2e_ppo_trainer_veomni_vllm.yml
+++ b/.github/workflows/e2e_ppo_trainer_veomni_vllm.yml
@@ -117,7 +117,7 @@ jobs:
         run: |
           pip3 install -r requirements-test.txt
           pip3 install --no-deps -e .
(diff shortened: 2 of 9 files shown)


verl/workers/engine/fsdp/transformer_impl.py
```diff
@@ -1092,6 +1092,21 @@ def prepare_model_outputs(self, output, output_args, micro_batch: TensorDict, lo
                 # temperature is singleton
                 log_probs = output.log_probs.squeeze(0)  # (total_nnz,)
                 entropy_rmpad = output.entropy.squeeze(0)  # (total_nnz,)
+
+                # When the fused kernel also computed top-K distillation
+                # (veomni's chunk_topk_distill path), extract the per-token
+                # distillation outputs and store them as nested tensors —
+                # same model_output keys as the eager logit-processor path.
+                if distillation_use_topk:
```
wuxibin89, line 1100: Does veomni `chunk_topk_distill_function` also apply to fsdp backend?

Luosuu: Not yet.

wuxibin89 (approved):

PR #6526, opened 2026-05-28
[megatron] feat: align optimizer states and DDP grad bucket with model precision

kolehma8:
### What does this PR do?

In bf16 training, the Megatron distributed-optimizer kept the Adam moments (m, v) and the grad-accumulation buffer in fp32 — 3× the necessary memory for those buffers. This PR makes them follow the model dtype, mirroring what Megatron's own CLI does with `--use-precision-aware-optimizer --main-grads-dtype bf16 --exp-avg-dtype bf16 --exp-avg-sq-dtype bf16`. The DDP grad bucket dtype is derived from the same signal so it stays consistent with the optimizer's expected grad dtype.

### Checklist Before Starting

- [ ] Search for similar PRs. Paste at least one query link here: https://github.com/verl-project/verl/issues?q=is%3Apr+precision-aware+optimizer
- [x] Format the PR title as `[{modules}] {type}: {description}` — `[megatron] feat: ...`

### Test

CPU-only unit tests added at `tests/utils/megatron/test_optimizer.py` (9 tests, ~9s, no GPU/TE required). They monkey-patch `OptimizerConfig` with a recorder so they verify which kwargs verl assembles for each precision mode without invoking Megatron's downstream validators or TE FusedAdam. Coverage:

- bf16 branch enables precision-aware optimizer and pins `main_grads_dtype` / `exp_avg_dtype` / `exp_avg_sq_dtype` to bf16; `main_params_dtype` intentionally left unset.
- fp16 branch enables precision-aware but leaves Adam moment dtypes at fp32 (with loss scaler + `store_param_remainders=False`).
- fp32 branch (both `bf16=False, fp16=False`) keeps Megatron defaults and does **not** enable precision-aware (Megatron asserts dtype fields must equal fp32 otherwise).
- Default kwargs dispatch to bf16 — backward compat for existing callers.
- `fp16=True, bf16=True` → fp16 wins (if/elif precedence).
- `use_distributed_optimizer` passes through.
- Top-level config fields (optimizer/lr/min_lr/clip_grad/weight_decay) pass through verbatim.
- `override_optimizer_config` wins over branch defaults; non-overridden fields preserved.
- Missing `override_optimizer_config` doesn't crash.

Manual end-to-end validation on Qwen3-30B-A3B GRPO (gsm8k, 8×H100, EP=8, mbs=1):

- Optimizer config dump in worker logs shows the new fields landing:

```
optimizer config after override: {..., 'use_precision_aware_optimizer': True,
 'main_grads_dtype': torch.bfloat16, 'exp_avg_dtype': torch.bfloat16,
 'exp_avg_sq_dtype': torch.bfloat16}
```

- Megatron model loads cleanly, step-0 validation completes (~51% gsm8k acc), forward + backward pass + optimizer step execute.

### API and Usage Example

`init_megatron_optim_config` gains a `bf16: bool = True` kwarg so callers can dispatch on the actual precision instead of inferring from a single `fp16` flag. The new kwarg defaults to True, preserving previous behavior for any caller that omits it (the old unlabeled `else` branch was effectively bf16).

```python
# Existing call sites updated:
optim_config_megatron = init_megatron_optim_config(
    optim_config,
    use_distributed_optimizer=wrap_config.use_distributed_optimizer,
    fp16=self.dtype == torch.float16,
    bf16=self.dtype == torch.bfloat16,   # NEW
)

# Opt back into the old fp32 optimizer-state layout via override:
# +actor_rollout_ref.actor.megatron.override_optimizer_config.use_precision_aware_optimizer=False
# +actor_rollout_ref.actor.megatron.override_ddp_config.grad_reduce_in_fp32=True
```

### Design & Code Changes

**Behavior matrix:**

| Mode | param dtype | DDP grad bucket | Adam m, v | Master weights | precision-aware opt |
|---|---|---|---|---|---|
| bf16 (new default) | bf16 | bf16 | bf16 | fp32 (TE limit) | True |
| fp16 | fp16 | fp32 | fp32 | fp32 | True |
| fp32 (new explicit branch) | fp32 | fp32 | fp32 | fp32 | False |

**Files changed:**

- `verl/utils/megatron/optimizer.py` — `init_megatron_optim_config` restructured into three branches; new `bf16: bool = True` kwarg.
- `verl/utils/megatron_utils.py` — In `get_model`, derive `grad_reduce_in_fp32 = not getattr(tfconfig, "bf16", False)` so the DDP grad bucket dtype matches `main_grads_dtype`. User overrides via `override_ddp_config` still win.
- `verl/workers/megatron_workers.py` — actor + critic call sites pass `bf16=self.dtype == torch.bfloat16`.
- `verl/workers/engine/megatron/transformer_impl.py` — same.
- `tests/utils/megatron/test_optimizer.py` — new test file (9 tests).

**Backward compatibility & known limits:**

- The new `bf16` kwarg defaults to `True`, so existing callers that pass only `fp16=...` keep bf16 dispatch — matches the prior implicit-bf16 `else` branch behavior. Only new callers wanting the explicit fp32 path need `bf16=False`.
- Existing bf16 runs will see lower optimizer-state memory but slightly different numerics: the precision-aware optimizer uses stochastic-rounding tricks to recover precision, but updates are not bit-equivalent to the prior fp32-state path. Worth a CHANGELOG entry.
- `main_params_dtype` is **not** dropped to bf16. TE FusedAdam currently rejects bf16 master weights at init (`RuntimeError: FusedAdam only supports fp32/fp16 master weights.`). NVIDIA's nightly docs describe future support; once TE catches up, we can revisit.

- [x] Add unit or end-to-end test(s) — `tests/utils/megatron/test_optimizer.py`, 9 CPU-only tests.
- [ ] Once your PR is ready for CI, send a message in the `ci-request` channel in the verl Slack.

### DIFF
--- a/tests/utils/megatron/test_optimizer.py
+++ b/tests/utils/megatron/test_optimizer.py
@@ -1,0 +1,206 @@
+# Copyright 2025 Bytedance Ltd. and/or its affiliates
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Tests for the precision-aware dispatch in ``init_megatron_optim_config``.
+
+These tests stub out ``megatron.core.optimizer.OptimizerConfig`` so they can
+run on CPU without TransformerEngine — the goal is to verify which kwargs
+verl assembles for each precision mode, not Megatron's downstream validation.
+
+The precision-aware optimizer is opt-in: the bf16 branch keeps the fp32
+optimizer state unless ``use_precision_aware_optimizer`` is set on the config,
+at which point the moment / grad dtypes follow the configured fields.
+"""
+
+from unittest.mock import MagicMock
+
+import pytest
+import torch
+from omegaconf import OmegaConf
+
+from verl.utils.megatron import optimizer as opt_mod
+from verl.utils.megatron.optimizer import init_megatron_optim_config
+
+
+def _base_optim_config(**overrides):
+    cfg = {
+        "optimizer": "adam",
+        "lr": 1e-3,
+        "min_lr": 0.0,
+        "clip_grad": 1.0,
+        "weight_decay": 0.01,
+    }
+    cfg.update(overrides)
+    return OmegaConf.create(cfg)
+
+
+def _precision_aware_optim_config(**overrides):
+    """bf16 optimizer state explicitly opted in via the new config flags."""
+    fields = {
+        "use_precision_aware_optimizer": True,
+        "main_grads_dtype": "bf16",
+        "exp_avg_dtype": "bf16",
+        "exp_avg_sq_dtype": "bf16",
+    }
+    fields.update(overrides)
+    return _base_optim_config(**fields)
+
+
+@pytest.fixture
+def captured_args(monkeypatch):
+    """Replace ``OptimizerConfig`` with a recorder so we can inspect kwargs."""
+    captured: dict = {}
+
+    def _fake(**kwargs):
+        captured.clear()
+        captured.update(kwargs)
+        return MagicMock(name="OptimizerConfig", **kwargs)
+
+    monkeypatch.setattr(opt_mod, "OptimizerConfig", _fake)
+    return captured
+
+
+def test_bf16_branch_defaults_to_fp32_optimizer_state(captured_args):
+    """Opt-out default: bf16 params, but the precision-aware optimizer stays off."""
+    init_megatron_optim_config(_base_optim_config(), fp16=False, bf16=True)
+
+    assert captured_args["bf16"] is True
+    assert captured_args["params_dtype"] is torch.bfloat16
+    # No precision-aware optimizer and no sub-fp32 moment / grad dtypes by default.
+    assert "use_precision_aware_optimizer" not in captured_args
+    assert "main_grads_dtype" not in captured_args
+    assert "exp_avg_dtype" not in captured_args
+    assert "exp_avg_sq_dtype" not in captured_args
+
+
+def test_bf16_branch_opt_in_enables_precision_aware_with_bf16_state(captured_args):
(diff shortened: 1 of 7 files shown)


wuxibin89:
optimizer should be in fp32.

kolehma8:
> optimizer should be in fp32.

We are trying to allow having optimizer states in lower precission as the FP32 is not always required (for instance when gradient accumulation is not used). Are you fine if this is exposed as configuration arguments instead of using BF16 directly? Also, note that in FSDP this is already the case (e.g. optimizer states and gradients are in the native precission of the model).

wuxibin89:
> Also, note that in FSDP this is already the case (e.g. optimizer states and gradients are in the native precission of the model).

No, in FSDP we force model type to be fp32, hence the adam optimizer states(m, v) are also fp32.
https://github.com/verl-project/verl/blob/release/v0.8.0/verl/workers/engine/fsdp/transformer_impl.py#L237-L238

The `MixedPrecision.param_dtype` only affects param during forward/backward which is casted from original param in fp32:
https://github.com/verl-project/verl/blob/release/v0.8.0/verl/workers/engine/fsdp/transformer_impl.py#L356

wuxibin89:
In practice, especially in RL, we don't use low precision for optimizer state. Please conduct some comparative experiments to verify whether low precision affects convergence.

kolehma8:
> In practice, especially in RL, we don't use low precision for optimizer state. Please conduct some comparative experiments to verify whether low precision affects convergence.

Sounds good. I was too hasty to claim that the FSDP uses BF16 and this is indeed incorrect. I will follow up with some ablation study to see if the BF16 optimizer states and/or gradients impact the convergence rate negatively.

kolehma8:
@wuxibin89 I added the results to https://github.com/verl-project/verl/issues/6576. In summary, the numerical precision did not make any difference in the rewards in the example case. This is not to say that this is the case in every single instance (for instance when gradient accumulation becomes high) and I am suggesting of making the BF16 flags an option for Megatron backend as opposed to the default that I had in the initial revision.

Could you open the PR again so I can push a revision to this or comment in case you disagree with the above approach?

wuxibin89:
@kolehma8 Please resolve conflicts, the legacy `megatron_workers.py` has been removed, we now use unify [engine_workers.py](https://github.com/verl-project/verl/blob/main/verl/workers/engine_workers.py)

kolehma8:
> @kolehma8 Please resolve conflicts, the legacy `megatron_workers.py` has been removed, we now use unify [engine_workers.py](https://github.com/verl-project/verl/blob/main/verl/workers/engine_workers.py)

Rebased and resolved conflicts. I also changed to using flags and the default remains FP32.

wuxibin89 (approved):

PR #6533, opened 2026-05-29
[rollout] feat: load-balance away from sticky sessions under imbalance

aoshen02:
### What does this PR do?

The global request load balancer (`GlobalRequestLoadBalancer`) pins each multi-turn conversation to a single server for prefix KV-cache reuse (sticky session) and **never moves it**. During the long-tail phase of a rollout step this starves throughput: a handful of long-running conversations stay pinned to one replica while peers go idle, so the step's wall-clock is bounded by the slowest replica.

This PR adds an **opt-in imbalance gate**: the sticky binding is honored only while the in-flight distribution is balanced, and the request is re-pinned to the least-loaded server once the pool becomes skewed. This trades prefix-cache reuse for load balancing exactly when it matters (the tail), recovering idle replicas without disturbing the balanced steady state.

The gate is **disabled by default** (`balance_abs_threshold=0`), so existing behavior is unchanged unless explicitly opted in.

### Checklist Before Starting

- [x] Search for similar PRs:
  - https://github.com/volcengine/verl/pulls?q=is%3Apr+GlobalRequestLoadBalancer (none)
  - https://github.com/volcengine/verl/pulls?q=is%3Apr+load+balancer+sticky (none)
- [x] PR title formatted as `[rollout] feat: ...`

### Design & Code Changes

Imbalance detection uses the dual-threshold rule from **sglang's cache-aware router** — a pool is imbalanced iff **both**:

```
(max_inflight - min_inflight) > balance_abs_threshold
max_inflight             > min_inflight * balance_rel_threshold
```

The absolute term avoids churning on tiny gaps; the relative term avoids treating a uniformly-busy pool as skewed.

- `verl/workers/rollout/llm_server.py`
  - `GlobalRequestLoadBalancer.__init__`: add `balance_abs_threshold` (default `0`) and `balance_rel_threshold` (default `1.5`).
  - New `_is_imbalanced()` helper; short-circuits to `False` when the gate is disabled.
  - `acquire_server()`: honor the sticky binding only while balanced; otherwise re-pin to the least-loaded server. Stale-server invalidation is preserved.
  - `_init_global_load_balancer()`: wire thresholds from rollout config (via `getattr`, so legacy configs fall back to pure sticky).
- `verl/workers/config/rollout.py`: new `LoadBalancerConfig` dataclass + `RolloutConfig.load_balancer` field (exported in `__all__`).
- `verl/trainer/config/rollout/rollout.yaml` + regenerated `_generated_*.yaml` (via `scripts/generate_trainer_config.sh`).

### API and Usage Example

Default (gate off, pure sticky — unchanged):

```yaml
# nothing to set; balance_abs_threshold defaults to 0
```

Enable the gate (re-balance when one replica is >32 requests ahead and >1.5x the least-loaded):

```yaml
actor_rollout_ref:
  rollout:
    load_balancer:
      balance_abs_threshold: 32
      balance_rel_threshold: 1.5
```

### Test

Added `TestLoadBalancerImbalanceGate` to `tests/experimental/agent_loop/test_basic_agent_loop.py` (alongside the existing `GlobalRequestLoadBalancer` unit tests), covering:

1. **disabled by default** keeps the sticky session even under heavy imbalance (regression guard);
2. **re-pins** a sticky request off the hot server once imbalanced;
3. **honors sticky** while the pool is balanced (KV reuse preserved);
4. **relative-threshold guard**: a uniformly-busy `{10, 8}` pool is *not* treated as imbalanced.

All four cases were validated against the real `GlobalRequestLoadBalancer` over Ray locally. Full `pytest` collection of this shared test module requires the CI test image (`peft`/`accelerate`/`torchdata` are imported at module load via `agent_utils`), which is unavailable in my local env. `ruff check` and `ruff format --check` pass on all changed files.

> AI assistance (Claude) was used for this change; every changed line was human-reviewed.

### End-to-end AB benchmark (GB300, SWE-bench multi-turn agent rollout)

Measured one full fully-async rollout step per config on real agent workload, 2026-06-12:

**Setup**: Qwen3.5-35B-A3B, 8 vLLM replicas (TP=1) on 2× GB300 trays + 2 training trays (EP=8); 32 prompts × GRPO n=16 = **512 SWE-bench Verified multi-turn trajectories** (Modal sandboxes, ≤300 turns, 128K response cap, prefix caching on). Identical prompt set / sampling / fleet across configs; only the gate thresholds change. Wall times measured from first trajectory start to k-th trajectory finish.

| config | rebalances (share of routes) | prompt-token max/min across replicas | p95 vs baseline | p99 vs baseline |
|---|---|---|---|---|
| `abs=0` (baseline, gate off) | 0 | ~1.28 | — (+28m39s) | — (+31m14s) |
| `abs=16, rel=1.5` | 28 (0.05%) | 1.169 | −3.7% | +8.4% |
| `abs=8, rel=1.5` | 675 (1.3%) | 1.121 | −0.8% | +10%¹ |
| `abs=4, rel=1.2` | 9,238 (17.4%) | 1.074 | **+18% (slower)** | **+20% (slower)**¹ |

¹ abs=8/abs=4 runs were measured to 511/512 finishes; p95/p99 are unaffected. The abs=16 run completed all 512. Run-to-run noise floor (established by the abs=16 run, whose 28 rebalances cannot move tail latency): **±4% on p95, ±8% on p99** — so abs=8/abs=16 are statistically indistinguishable from baseline, while abs=4's +18% is far outside noise.

**Findings:**

1. **The gate does what it claims mechanically** — replica imbalance falls monotonically with rebalance count (1.28 → 1.07), and in-flight spreads tighten to 26–29 under `abs=4`.
2. **But on this workload balance does not buy wall-clock**: every migration of an in-flight multi-turn conversation forfeits its accumulated prefix KV (often 50–100K tokens) and forces a full re-prefill on the new replica. At `abs=4` that cost (9.2K re-prefills) dominates and the step is 18–20% slower end-to-end; at `abs≥8` the gate barely fires and the result equals baseline.
3. **Time-series of the rebalance rate is bimodal** (per-200-route counter logs): ~26% during ramp-up, ~13% steady-state, rising again to ~24% during drain — i.e. the gate migrates most aggressively exactly when only long-context stragglers remain and migration is most expensive. The drain phase is the tail the gate hopes to fix, but breaking a 200-turn conversation's sticky binding there makes the tail worse, not better.

**Guidance**: keep the default **off** for long-context multi-turn agent rollouts (the trade is strictly unfavorable there). The gate is suited to workloads where migrated requests carry little KV state — short prompts, single-turn, or continuous request arrival where the migrating request is young.

**Follow-up prototype, measured** (not part of this PR): make migration **KV-aware** — the balancer approximates a request's accumulated context from its own per-`request_id` route count (the `request_id` is stable across turns of one trajectory, which is what sticky sessions rely on) and, when the pool is imbalanced, re-pins only *young* requests (≤4 turns) while never breaking long-context bindings. Same harness, same `abs=4, rel=1.2`:

| config | rebalances | blocked long-ctx migrations | p95 vs baseline | p99 vs baseline | full wall time (512/512) |
|---|---|---|---|---|---|
| `abs=4` stock | 9,238 | — | +18% | +20% | n/a¹ |
| `abs=4` + KV-aware (`max_turns=4`) | 704 | 31,835 | **−3.7%** | **−3.1%** | **34m00s (−4.9%)** |

¹ stock `abs=4` was measured to 511/512; its p95/p99 are exact.

The KV-aware filter fully removes the +18% regression (p95 6m20s faster than stock at identical thresholds) and is the only config of the six measured whose p95, p99 **and** complete wall time (34m00s vs the baseline's 35m44s) all land below baseline. Each individual margin is within the run-to-run noise floor (±4% p95 / ±8% p99 at n=1; the last-finish wall time is dominated by which trajectory draws the hardest task), so this is a consistent directional signal rather than a proven win — repeats are needed to claim it. Mechanistically the reason is visible in the route logs: with one-shot dispatch all surviving requests age past the turn cap after ramp-up, so migrations stop and the system degenerates to pure sticky in steady state. The KV-aware variant should pay off most under *continuous request arrival*, where young (cheap-to-migrate) requests exist throughout the run; happy to send it as a follow-up PR if there's interest.

### DIFF
--- a/verl/trainer/config/rollout/rollout.yaml
+++ b/verl/trainer/config/rollout/rollout.yaml
@@ -449,9 +449,9 @@ load_balancer:
   # Required when using verl.utils.omega_conf_to_dataclass to instantiate dataclass configs
   _target_: verl.workers.config.LoadBalancerConfig
 
-  # Absolute in-flight gap that triggers rebalancing. 0 disables the imbalance
-  # gate, preserving pure sticky-session routing (prefix KV-cache reuse).
-  balance_abs_threshold: 0
+  # Absolute in-flight gap that triggers rebalancing. Set to 0 to disable the
+  # imbalance gate; the default of 64 mirrors sglang's cache-aware router.
+  balance_abs_threshold: 64
 
   # Relative in-flight ratio that triggers rebalancing. A pool is imbalanced iff
   # (max - min) > balance_abs_threshold AND max > min * balance_rel_threshold.
--- a/verl/trainer/config/rollout/rollout.yaml
+++ b/verl/trainer/config/rollout/rollout.yaml
@@ -440,22 +440,15 @@ prometheus:
   # Path to Prometheus configuration file
   file: /tmp/ray/session_latest/metrics/prometheus/prometheus.yml
 
-  # Specify served_model_name to avoid displaying overly long model paths in Grafana
+# Specify served_model_name to avoid displaying overly long model paths in Grafana
   served_model_name: ${oc.select:actor_rollout_ref.model.path,null}
 
 # sticky-vs-balance tradeoff for the global request load balancer (server mode)
-load_balancer:
+balance_abs_threshold: 64
 
-  # Required when using verl.utils.omega_conf_to_dataclass to instantiate dataclass configs
-  _target_: verl.workers.config.LoadBalancerConfig
-
-  # Absolute in-flight gap that triggers rebalancing. Set to 0 to disable the
-  # imbalance gate; the default of 64 mirrors sglang's cache-aware router.
-  balance_abs_threshold: 64
-
-  # Relative in-flight ratio that triggers rebalancing. A pool is imbalanced iff
-  # (max - min) > balance_abs_threshold AND max > min * balance_rel_threshold.
-  balance_rel_threshold: 1.5
+# Relative in-flight ratio that triggers rebalancing. A pool is imbalanced iff
+# (max - min) > balance_abs_threshold AND max > min * balance_rel_threshold.
+balance_rel_threshold: 1.5
 
 # type of quantization in vllm, currently support fp8 and torchao
 quantization: null
--- a/verl/trainer/config/rollout/rollout.yaml
+++ b/verl/trainer/config/rollout/rollout.yaml
@@ -443,6 +443,20 @@ prometheus:
   # Specify served_model_name to avoid displaying overly long model paths in Grafana
   served_model_name: ${oc.select:actor_rollout_ref.model.path,null}
 
+# sticky-vs-balance tradeoff for the global request load balancer (server mode)
+load_balancer:
+
+  # Required when using verl.utils.omega_conf_to_dataclass to instantiate dataclass configs
+  _target_: verl.workers.config.LoadBalancerConfig
+
+  # Absolute in-flight gap that triggers rebalancing. 0 disables the imbalance
+  # gate, preserving pure sticky-session routing (prefix KV-cache reuse).
+  balance_abs_threshold: 0
+
+  # Relative in-flight ratio that triggers rebalancing. A pool is imbalanced iff
+  # (max - min) > balance_abs_threshold AND max > min * balance_rel_threshold.
+  balance_rel_threshold: 1.5
+
 # type of quantization in vllm, currently support fp8 and torchao
 quantization: null
 
--- a/verl/workers/rollout/llm_server.py
+++ b/verl/workers/rollout/llm_server.py
@@ -90,11 +90,7 @@ class GlobalRequestLoadBalancer:
         self._balance_rel_threshold = balance_rel_threshold
 
     def _is_imbalanced(self) -> bool:
-        """Whether the current in-flight distribution is skewed enough to rebalance.
-
-        Returns ``False`` when the gate is disabled (``balance_abs_threshold <= 0``)
-        so that sticky sessions are always honored.
-        """
+        """Whether the current in-flight distribution is skewed enough to rebalance."""
         if self._balance_abs_threshold <= 0:
             return False
         loads = self._inflight_requests.values()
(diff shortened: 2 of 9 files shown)


verl/trainer/config/rollout/rollout.yaml
```diff
@@ -443,6 +443,20 @@ prometheus:
...
 
+# sticky-vs-balance tradeoff for the global request load balancer (server mode)
+load_balancer:
+
+  # Required when using verl.utils.omega_conf_to_dataclass to instantiate dataclass configs
+  _target_: verl.workers.config.LoadBalancerConfig
+
+  # Absolute in-flight gap that triggers rebalancing. 0 disables the imbalance
+  # gate, preserving pure sticky-session routing (prefix KV-cache reuse).
+  balance_abs_threshold: 0
```
wuxibin89: Should we enable it by default, I don't think user know how to tune it.

verl/workers/rollout/llm_server.py
```diff
@@ -335,9 +377,14 @@ async def _initialize_llm_servers(self, start_rank: int = 0):
             update_prometheus_config(self.rollout_config.prometheus, self.server_addresses, self.rollout_config.name)
 
     async def _init_global_load_balancer(self) -> None:
+        # Sticky-vs-balance tradeoff knobs; absent/legacy configs fall back to
+        # pure sticky (abs_threshold=0 disables the imbalance gate).
+        lb_config = getattr(self.rollout_config, "load_balancer", None)
         self.global_load_balancer = GlobalRequestLoadBalancer.remote(
             servers=dict(zip(self.server_addresses, self.server_handles, strict=True)),
             max_cache_size=DEFAULT_ROUTING_CACHE_SIZE,
+            balance_abs_threshold=getattr(lb_config, "balance_abs_threshold", 0),
```
wuxibin89: No need for backward compatible, just use `self.rollout_config.balance_abs_threshold`

verl/workers/rollout/llm_server.py
```diff
@@ -53,40 +53,82 @@ class GlobalRequestLoadBalancer:
...
         if not servers:
             raise ValueError("servers must be non-empty")
 
         self._servers: dict[str, ray.actor.ActorHandle] = dict(servers)
         self._inflight_requests: dict[str, int] = {sid: 0 for sid in servers}
         self._request_id_to_server: LRUCache = LRUCache(maxsize=max_cache_size)
+        self._balance_abs_threshold = balance_abs_threshold
+        self._balance_rel_threshold = balance_rel_threshold
+
+    def _is_imbalanced(self) -> bool:
```
wuxibin89, line 92: Server instances may be dynamically add/remove from GlobalRequestLoadBalancer, does it trigger rebalance immediately?
- https://github.com/verl-project/verl/blob/main/verl/workers/rollout/llm_server.py#L100
- https://github.com/verl-project/verl/blob/main/verl/workers/rollout/llm_server.py#L115

aoshen02: Rebalance happens on the next acquire with the updated topology.

PR #6544, opened 2026-05-31
[reward] fix: release worker count when NaiveRouter request fails after retries

fuergaosi233:
### What

`NaiveRouter._select_worker()` increments `request_counts[url]`, but `_release_worker()` was only called on the **success** path. When all `max_attempts` fail and the method raises (the `RuntimeError` after the loop, or a re-raised `ClientResponseError` / final unexpected error), the worker's count is never decremented and leaks upward forever. Since selection is `min(request_counts, ...)`, a repeatedly-failing worker accumulates phantom load and permanently skews load balancing.

Fix: wrap the retry loop in `try/finally` and release the worker in `finally`, so the single increment from `_select_worker()` is always balanced exactly once, success or failure.

No open PR touches `naive_router` / worker accounting.

### Tests run (CPU)

Replicating the method's control flow: after 3 requests whose worker is always down, `request_counts` returns to `{0, 0}` with the fix (previously leaked to non-zero and never recovered).

### AI assistance

AI assistance (Claude) was used to locate the issue and draft the fix. Reviewed line-by-line by me.

### DIFF
--- a/verl/experimental/reward_loop/router/naive_router.py
+++ b/verl/experimental/reward_loop/router/naive_router.py
@@ -147,30 +147,35 @@ class NaiveRouter:
         body = await request.body()
         headers = dict(request.headers)
 
-        for attempt in range(self.max_attempts):
-            # Send request to worker
-            try:
-                async with self.client.request(request.method, target_url, data=body, headers=headers) as response:
-                    response.raise_for_status()
-                    output = await _read_async_response(response)
-                    self._release_worker(worker_url)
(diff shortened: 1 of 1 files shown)


wuxibin89 (dismissed):

wuxibin89:
Please format code: https://github.com/verl-project/verl/blob/main/CONTRIBUTING.md#code-linting-and-formatting

wuxibin89 (approved):

PR #6433, opened 2026-05-21
[ckpt] fix: use asyncio to run the _run method

tsinghuazl22:
Refactor _run method to use asyncio for execution.

#### What does this PR do?

Refactor `HCCLCheckpointEngine.BroadcastOperation._run` to run in an asyncio executor, matching the NCCL implementation.  
This keeps the `BroadcastOperation` interface consistent across backends and allows the async training pipeline to overlap communication with computation more effectively.

### Test

Verified with end-to-end async training runs on HCCL/NCCL backends.

### API and Usage Example

No API changes.

```python
broadcast_op = BroadcastOperation(...)
metadata = await broadcast_op.wait_for_complete()
```

Issue #6425, opened 2026-05-21
HCCL checkpoint engine 中存在两个实现疑问：注册名使用 nccl，且 BroadcastOperation 未采用线程池异步执行


tsinghuazl22:
在阅读 `verl/checkpoint_engine/hccl_checkpoint_engine.py` 时，我注意到两个实现层面的疑问，想请维护者确认这是有意设计，还是实现遗漏。
## 1. HCCL engine 的注册名为什么是 `"nccl"`
当前代码中：
```python
@CheckpointEngineRegistry.register("nccl")
class HCCLCheckpointEngine(CheckpointEngine):
```
这里为什么是 nccl，而不是 hccl？

## 2. HCCL 的 BroadcastOperation 当前是同步执行：
当前 HCCL 的实现是：
```python
class BroadcastOperation:
    def __init__(...):
        self._run()
    async def wait_for_complete(self):
        return self.metadata
```
而 NCCL 版本是放到线程池里的：
```python
loop = asyncio.get_running_loop()
self._task = loop.run_in_executor(None, self._run)
```
HCCL 这里为什么没有像 NCCL 一样采用线程池异步执行？是有意保持同步实现，还是实现上暂未对齐？

tardis-key:
不是CheckpointEngine相关代码的维护者，但是说说我自己个人角度理解

对于问题1，设计上应该是为了实现外部感知的一致性。双方都是集合通信库，也做到了接口的一致性，所以外部似乎不需要感知到底是hccl还是nccl。当前verl也在做很多事情，想要在框架侧屏蔽掉不同后端带来的实现差异，我觉得这就是其中之一。另外功能上，在导入的时候hccl和nccl会互斥，所以应该也不会有功能问题

对于问题2，我感觉不是有意为之，就是没有考虑到。如果异步线程池有明显的性能优势，可以尝试修改

Kirrito-k423:
Thank you for carefully reading the HCCL checkpoint engine code and raising these two implementation questions.

**Question 1: Why is the HCCL engine registered as \"nccl\"?**

This appears to be an implementation artifact for interface compatibility. The HCCL and NCCL checkpoint engines share the same abstract interface (`CheckpointEngine`), and registering HCCL under \"nccl\" allows the framework to use either HCCL or NCCL transparently based on the hardware available. At import time, HCCL and NCCL are mutually exclusive (you cannot have both initialized), so the \"nccl\" registration effectively means \"use whichever NCCL-compatible library is available.\"

[...]

PR #6569, opened 2026-06-02
[ckpt] fix: restore async HCCL broadcast execution

tsinghuazl22:
### What does this PR do?

This PR restores asynchronous broadcast execution for HCCLCheckpointEngine.BroadcastOperation while keeping the external interface aligned with the NCCL checkpoint engine.

Compared with the previously closed #6433, this version does not move ZeroMQ socket access into a worker thread. Instead, it keeps metadata communication on the event loop via zmq.asyncio and only offloads the blocking HCCL collective to the executor. The background thread also restores the NPU device context with 	orch.npu.set_device(...) before calling roadcast.

This avoids the two concerns raised in the earlier review:
1. ZMQ sockets are not accessed from a worker thread.
2. HCCL execution in the worker thread has an explicit NPU device context.

This is not duplicating an existing open PR. I checked the existing discussion in #6425 and the previously closed #6433, and this patch uses a different approach from #6433 to address the review feedback directly.

### Test

Local:
- python -m compileall verl/checkpoint_engine/hccl_checkpoint_engine.py

In ssh 100 -> container erl_latest:
- python -m compileall verl/checkpoint_engine/hccl_checkpoint_engine.py
- smoke check: 	orch.npu.is_available() == True
- smoke check: import verl.checkpoint_engine.hccl_checkpoint_engine

Results:
- Local compile check passed.
- Container compile check passed.
- Container NPU/import smoke checks passed.

### API and Usage Example

No API changes.

### AI Assistance

This PR was prepared with AI assistance, and the changes and verification steps were reviewed before submission.

### DIFF
--- a/verl/checkpoint_engine/hccl_checkpoint_engine.py
+++ b/verl/checkpoint_engine/hccl_checkpoint_engine.py
@@ -11,6 +11,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+import asyncio
 import logging
 import os
 import time
@@ -20,6 +21,7 @@ from typing import AsyncGenerator, Generator
 import ray
 import torch
 import zmq
+import zmq.asyncio
 from vllm.distributed.utils import StatelessProcessGroup
 
 from verl.checkpoint_engine.base import CheckpointEngine, CheckpointEngineRegistry, TensorMeta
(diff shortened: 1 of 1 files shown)


wuxibin89:
@Shangwei-Li @hanhan-networking Could you help to review this PR?

tests/checkpoint_engine/test_hccl_broadcast_operation_on_cpu.py
```diff
@@ -0,0 +1,124 @@
...
+import importlib.util
+import sys
+import threading
+import types
+from pathlib import Path
+
+import torch
+
+
+def _load_hccl_checkpoint_engine_module(monkeypatch):
```
wuxibin89: This cpu test is not needed, please remove it.

wuxibin89 (changes requested):

PR #5332, opened 2026-02-16
fix: handle labels=None in fused forward to avoid crash without MTP (#5273)

devchilll:
## Problem

When `use_fused_kernels=True` and `mtp.enable=False`, the patched `_fused_GPTModel_forward` receives `labels=None` from the non-fused caller path (`gptmodel_forward_no_padding`). This causes an `AttributeError` in `linear_cross_entropy` because it tries to access `labels.shape` on `None`.

**Root cause:** `patch_fused_forward` replaces `GPTModel.forward` globally, but the non-fused caller path (`gptmodel_forward_no_padding`) only passes `labels` when both `enable_mtp` and `post_process` are `True`. When MTP is disabled, the patched forward is still called without `labels`, leading to the crash.

**Error traceback:**
```
File "verl/utils/kernel/linear_cross_entropy.py", line 74, in forward
    if len(labels.shape) != 1:
           ^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'shape'
```

## Fix

Add a guard in `_fused_GPTModel_forward` that, when `labels is None`, falls back to standard logit computation via `model.output_layer` — matching the original `GPTModel._postprocess` behavior.

## Testing

This fix is a minimal guard that only activates when `labels is None`. It does not affect the existing fused kernel path (which always receives labels via `fused_forward_no_padding`).

Fixes #5273

PR #6596, opened 2026-06-03
[megatron] fix: tolerate extra kwargs in MTP postprocess patch

Meirtz:
### Summary

`patch_postprocess` (MTP support) replaces `GPTModel._postprocess` with `_megatron_gptmodel_postprocess`. In current Megatron-Core, `GPTModel.forward` calls `self._postprocess(...)` with extra keyword arguments — notably `padding_mask` (the MoE-routing mask). The patched function has a fixed signature with no `**kwargs`, so enabling MTP raises:

```
TypeError: _megatron_gptmodel_postprocess() got an unexpected keyword argument 'padding_mask'
```

This accepts `padding_mask` and adds `**kwargs` so the monkeypatch tolerates keyword arguments threaded into `GPTModel.forward` across Megatron-Core versions. These arguments are consumed in the decoder (e.g. MoE routing), not in logits/loss postprocessing, so ignoring them here preserves behavior.

### Not a duplicate

Checked open PRs/issues: #4539 (draft that adds the MTP patch; does not touch the signature for `padding_mask`/`**kwargs`) and #5332 (a different `labels=None` fused-forward crash). No open PR/issue addresses this.

### Test

Ran verl SFT with `engine=megatron`, `model.mtp.enable=True` against a Megatron-Core whose `GPTModel.forward` passes `padding_mask` to `_postprocess`:
- before: `TypeError ... unexpected keyword argument 'padding_mask'` at the first step
- after: training and validation loss decrease over steps, no crash

### AI assistance

Prepared with AI assistance (Claude) and reviewed by the submitter.

### DIFF
--- a/verl/models/mcore/mtp_patch.py
+++ b/verl/models/mcore/mtp_patch.py
@@ -87,6 +87,8 @@ def _megatron_gptmodel_postprocess(
     extra_block_kwargs=None,
     inference_context=None,
     is_spec_decode=None,
+    padding_mask=None,
+    **kwargs,
 ):
     """Postprocesses decoder hidden states to generate logits or compute loss.
 


Meirtz:
For context: this also unblocks the DeepSeek-V4-Flash SFT example #6603, which enables MTP and so hits this `_postprocess` path — without this fix that example crashes with the `padding_mask` TypeError.

Meirtz:
Folding the 2-line MTP postprocess kwargs fix into #6603 (the DeepSeek-V4-Flash SFT example that needs it), so it ships as one self-contained change rather than a standalone 2-line PR. Closing in favor of #6603.

Issue #5087, opened 2026-01-28
ExecuteModelRequest cannot be found


joannanananana:
### System Info

Hi,
I found an erro when use verl0.7.0 +vllm0.12.0
There is an Importerro(can not import ExecuteModelRequest from vllm.sequence) in verl/verl/workers/rollout/vllm_rollout/vllm_rollout.py   which is not a ModuleNotFoundError.  

`try:
    from vllm.worker.worker_base import WorkerWrapperBase
# except ModuleNotFoundError:
except  ModuleNotFoundError:
    # https://github.com/vllm-project/vllm/commit/6a113d9aed8221a9c234535958e70e34ab6cac5b
    from vllm.v1.worker.worker_base import WorkerWrapperBase`

### Information

- [ ] The official example scripts
- [ ] My own modified scripts

### Tasks

- [ ] An officially supported task in the `examples` folder (such as GLUE/SQuAD, ...)
- [ ] My own task or dataset (give details below)

### Reproduction

from vllm.worker.worker_base import WorkerWrapperBase

### Expected behavior

except ModuleNotFoundError ---->except

Kirrito-k423:
You're correct. The exception handling should catch `ImportError` not just `ModuleNotFoundError`. The difference:

- `ModuleNotFoundError`: Raised when a module doesn't exist
- `ImportError`: Raised when a module exists but fails to import (e.g., missing a class)

In vLLM 0.12.0, `ExecuteModelRequest` was removed from `vllm.sequence`, but the module `vllm.sequence` still exists. So the import fails with `ImportError`, not `ModuleNotFoundError`.

**Fix:**
```python
try:
    from vllm.worker.worker_base import WorkerWrapperBase
except ImportError:  # Changed from ModuleNotFoundError
    from vllm.v1.worker.worker_base import WorkerWrapperBase
```

Or more defensively:
```python
try:
    from vllm.worker.worker_base import WorkerWrapperBase
except (ImportError, ModuleNotFoundError):
    from vllm.v1.worker.worker_base import WorkerWrapperBase
```

This should be fixed in verl main branch. Consider submitting a PR.

Kirrito-k423:
中文：感谢报告。我们核对了 v0.7.0 标签，那里确实只捕获了 ModuleNotFoundError，因此在某些 vLLM 版本里 vllm.worker.worker_base 模块存在但其内部导入失败时，会出现你描述的 ImportError。

[...]

PR #6603, opened 2026-06-04
[megatron] feat: add DeepSeek-V4-Flash SFT example (Megatron-Bridge backend)

Meirtz:
### Summary

Adds a GSM8K SFT example for DeepSeek-V4-Flash via the Megatron-Bridge backend, mirroring `examples/sft/gsm8k/run_mimo_7b_mtp_megatron.sh`. Runs at TP1/PP4/EP8 (32 GPU), THD packed sequences, MTP enabled, unfused mHC/RoPE (`apply_rope_fusion=False`, `use_fused_mhc=False`), `use_mbridge=True, vanilla_mbridge=False`.

Folds in a 2-line fix to `verl/models/mcore/mtp_patch.py` so the MTP postprocess monkeypatch tolerates extra kwargs (`padding_mask`) that current Megatron-Core's `GPTModel.forward` passes — required by `model.mtp.enable=True`. Supersedes the standalone #6596 (closed).

### Environment / image — please read first

DeepSeek-V4-Flash SFT needs a specific stack and **will not run on a stock verl install**. Required environment:

- **Base image:** `nvcr.io/nvidia/pytorch:26.04-py3`
- **verl** — this PR (includes the `mtp_patch.py` fix)
- **Megatron-Bridge** (`megatron.bridge`) + **Megatron-Core** with the DSv4 hybrid attention and THD support — Megatron-Core PR NVIDIA/Megatron-LM#5011 (open). The example script fetches both onto `PYTHONPATH`.
- **DSv4 kernels:** `fast_hadamard_transform` + `nvidia-cudnn-frontend` — the example script installs these.

**Reference Dockerfile that bakes the full environment:** https://gist.github.com/Meirtz/7e240c5489851ebf79a6edfd357a40b7

The example script also bootstraps the Megatron-side pieces and kernels at runtime, so a plain `nvcr.io/nvidia/pytorch:26.04-py3` container is sufficient to run it.

### Not a duplicate

The DeepSeek-V4 GRPO path is #6473 (`examples/grpo_trainer/run_deepseek_v4_flash_megatron.sh`); this is the **SFT** counterpart under `examples/sft/gsm8k/`. Same Megatron-Bridge backend, different trainer.

### Test

The DSv4 SFT path through Megatron-Bridge (THD, MTP, the `apply_rope_fusion=False` / `use_fused_mhc=False` config) was validated end-to-end on 8×H100 with a 4-layer DSv4 proxy: training and validation loss decrease, no NaN. The full model (TP1/PP4/EP8) and this exact GSM8K configuration have not been run by me; the script mirrors the validated config and the existing megatron SFT example structure. CI cannot exercise it until the DSv4 kernels are in verl's CI image.

### AI assistance

Prepared with AI assistance (Claude) and reviewed by the submitter.


examples/sft/gsm8k/run_deepseek_v4_flash_sft_megatron.sh
```diff
@@ -0,0 +1,120 @@
...
+# model.mtp.enable=True additionally requires verl#6596 (MTP postprocess kwargs).
+PYPATH=$HOME/pythonpath
+mkdir -p "$PYPATH" && cd "$PYPATH"
+# Megatron-Core: fetch PR #5011's head (carries the dev SBHD base + the THD support).
+[ -d Megatron-LM ] || { git clone https://github.com/NVIDIA/Megatron-LM && \
+    (cd Megatron-LM && git fetch origin pull/5011/head:dsv4-thd && git checkout dsv4-thd); }
+# Megatron-Bridge: provides megatron.bridge (the DSv4 AutoBridge / recipes).
+[ -d Megatron-Bridge ] || git clone https://github.com/NVIDIA-NeMo/Megatron-Bridge
+cd -
+export PYTHONPATH=$PYTHONPATH:$PYPATH/Megatron-Bridge/src:$PYPATH/Megatron-LM
```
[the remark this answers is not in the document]
Meirtz: Good catch — fixed, now uses `${PYTHONPATH:-}` so it's safe under `set -u` when PYTHONPATH is unset.

examples/sft/gsm8k/run_deepseek_v4_flash_sft_megatron.sh
```diff
@@ -0,0 +1,121 @@
+#!/usr/bin/env bash
+# DeepSeek-V4-Flash SFT on GSM8K via the Megatron-Bridge backend.
+# DSv4 hybrid attention requires TP=1; scale with PP and EP -> TP1/PP4/EP8 = 32 GPUs
+# (e.g. 4 nodes x 8 GPU on H100/H200, 8 nodes x 4 GPU on GB200/GB300).
```
ISEEKYAN, line 5: 32*H100 or H200 would meet OOM, better give a more realistic setting for H100.
128*H100 with fully recompute and PP16EP8 would be a choice

examples/sft/gsm8k/run_deepseek_v4_flash_sft_megatron.sh
```diff
@@ -0,0 +1,121 @@
...
+set -xeuo pipefail
+
+NUM_GPUS=${NUM_GPUS:-8}          # GPUs per node
+TP_SIZE=${TP_SIZE:-1}
+PP_SIZE=${PP_SIZE:-4}
```
ISEEKYAN, line 10: how to set PP4 or PP16 with flash config:
"num_hidden_layers": 43,
"num_hash_layers": 3,
"num_nextn_predict_layers": 1,

examples/sft/gsm8k/run_deepseek_v4_flash_sft_megatron.sh
```diff
@@ -0,0 +1,121 @@
...
+    engine.pipeline_model_parallel_size=${PP_SIZE} \
+    engine.expert_model_parallel_size=${EP_SIZE} \
+    engine.expert_tensor_parallel_size=1 \
+    engine.context_parallel_size=${CP_SIZE} \
+    engine.virtual_pipeline_model_parallel_size=${VPP_SIZE} \
+    engine.use_mbridge=True \
+    engine.vanilla_mbridge=False \
+    engine.use_remove_padding=${USE_REMOVE_PADDING} \
+    engine.use_dist_checkpointing=False \
+    +engine.override_transformer_config.apply_rope_fusion=False \
```
ISEEKYAN: can we enable rope fusion? it is critical for varlen perf

examples/sft/gsm8k/run_deepseek_v4_flash_sft_megatron.sh
ISEEKYAN: or PP4EP32

Meirtz:
Folded the 2-line MTP postprocess fix (previously #6596) into this PR so the example is self-contained for `model.mtp.enable=True`. Closing #6596 in favor of this.

examples/sft/gsm8k/run_deepseek_v4_flash_sft_megatron.sh
Meirtz: No manual layout — [NVIDIA-NeMo/Megatron-Bridge#4131](https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/4131) auto-sets it for DSv4 at PP>1. Hash layers are first, so the even split keeps them on stage 0 for both PP4 and PP16. Depends on that PR; once it lands, just set PP=4 or 16.

examples/sft/gsm8k/run_deepseek_v4_flash_sft_megatron.sh
Meirtz: Right, 32×H100 OOMs (recompute already full). Both 128-GPU options work layout-wise now — [NVIDIA-NeMo/Megatron-Bridge#4131](https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/4131) handles PP4 and PP16.

examples/sft/gsm8k/run_deepseek_v4_flash_sft_megatron.sh
Meirtz: ~~(Earlier attributed this to an upstream kernel bug — wrong.)~~ Root cause was a Megatron-Bridge config-mapping issue (`partial_rotary_factor` double-applied → the fused rope kernel read its cos/sin cache out of bounds), fixed in [NVIDIA-NeMo/Megatron-Bridge#4271](https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/4271) and validated on GB300 (fused now matches unfused to 4 decimals). `apply_rope_fusion=True` will be enabled here once the example's Megatron-Bridge checkout includes that fix.

Meirtz:
Heads-up for anyone running this example: Megatron-LM is now pinned to the validated commit `ed6b1f65` (commit 8de857f) instead of tracking `pull/5011/head`. The #5011 head was rebased onto pure `dev` and dropped main-side symbols that `megatron.bridge` imports — if you cloned earlier and hit `ImportError: cannot import name 'safe_get_world_size' from 'megatron.core._rank_utils'`, fix it with: `cd Megatron-LM && git fetch origin ed6b1f65502aec7f2fe27e14a1245c29e435c2a6 && git checkout ed6b1f65502aec7f2fe27e14a1245c29e435c2a6`.

Meirtz:
Two fixes from a user running this example end-to-end (thanks!), now **smoke-validated on a GB300 node against this PR's head**:

1. **Data**: the example now uses the SFT preprocess `examples/data_preprocess/gsm8k_multiturn_sft.py` (`messages` column, includes assistant turns) — the previous `messages_key=prompt` raised `KeyError: 'prompt'` against SFT-preprocessed parquet. Validated: parquet schema ✓.
2. **Chat template**: **DeepSeek-V4-Flash ships no `chat_template` in its HF tokenizer** (and no role tokens — only bos/eos), so the script generates a minimal template and passes it via `model.custom_chat_template`. Note: upstream verl consumes this value **verbatim as the template string** (`verl/workers/config/model.py`) — there is no `@file` indirection here (some downstream forks have one), so the script inlines the file content: `model.custom_chat_template="\"$(cat $CHAT_TEMPLATE_FILE)\""`. Validated: tokenizer has no builtin template ✓, `MultiTurnSFTDataset` sample renders with a non-empty assistant loss mask ✓, the override survives hydra's grammar verbatim ✓.

Commits `640f0ed` + `9823b0b`. Replace the generated template via `CHAT_TEMPLATE_FILE=...` if you have an official one.

examples/sft/gsm8k/run_deepseek_v4_flash_sft_megatron.sh
Meirtz: Done — rope fusion is now ON in this example. The NaN root cause was a Megatron-Bridge config-mapping bug (fixed in [NVIDIA-NeMo/Megatron-Bridge#4271](https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/4271), merged); full-model DSv4-Flash SFT on 8×GB300 with `apply_rope_fusion=True` now matches the unfused control (iter-10 lm loss 2.3319 vs 2.3306, 0 NaN over 20 iters). Varlen perf path restored.

examples/sft/gsm8k/run_deepseek_v4_flash_sft_megatron.sh
```diff
@@ -0,0 +1,136 @@
...
+    engine.use_mbridge=True \
+    engine.vanilla_mbridge=False \
+    engine.use_remove_padding=${USE_REMOVE_PADDING} \
+    engine.use_dist_checkpointing=False \
+    +engine.override_transformer_config.apply_rope_fusion=True \
+    +engine.override_transformer_config.use_fused_mhc=False \
+    +engine.override_transformer_config.recompute_granularity=full \
+    +engine.override_transformer_config.recompute_method=uniform \
+    +engine.override_transformer_config.recompute_num_layers=1 \
+    "
```
shyoshyo, line 106: It appears that sparse attention fusion is not enabled by default unless explicitly specified as `+engine.override_transformer_config.apply_dsa_kernel_fusion=True`

shyoshyo (commented):
It appears that sparse attention fusion is not enabled by default unless explicitly specified. This can consume significant amounts of GPU memory in long-context scenarios.

examples/sft/gsm8k/run_deepseek_v4_flash_sft_megatron.sh
Meirtz: Good point, thanks for raising it. Two reasons this example keeps it off by default:

1. The pinned mcore asserts SM100+ for `apply_dsa_kernel_fusion` (`transformer_config.py`), so enabling it would break the H100 path this example also targets — and it additionally needs `flash_mla` from FlashMLA@nv_dev, which the setup here doesn't install. Added it as a commented Blackwell-only switch next to `use_fused_mhc` (6475db0).
2. Long context is currently bounded by CP support anyway (mcore #5087 still open), so at the sequence lengths trainable today the unfused fallback is workable.

FWIW, the SM100 gate looks relaxable upstream now that cudnn-frontend v1.25.0 ships the SM90 indexer kernels and FlashMLA's sparse fwd supports SM90a — but that's an mcore-side change.

PR #6654, opened 2026-06-08
[megatron] fix YaRN rope config with mBridge

tntnnlrw:
## Summary

This PR is scoped to the current verl Megatron integration path, not a replacement for the generic Megatron-Bridge YaRN mapping.

- preserve HF YaRN runtime fields for the current verl mBridge path, especially the default `vanilla_mbridge=True` path that uses `ISEEKYAN/mbridge`
- support both legacy `rope_scaling` and Transformers 5-style `rope_parameters`, including nested layouts
- keep YaRN runtime-only keys out of bridge constructor/override kwargs, then apply them back to the final Megatron config/provider consumed by GPTModel
- keep explicit `override_transformer_config` YaRN values taking precedence over HF config values
- add CPU-only regression coverage for legacy, nested, override, and non-YaRN cases

## Why

Megatron-Bridge already has upstream mapping for `rope_scaling.type == "yarn"` on the non-vanilla Megatron-Bridge path. This PR is narrower: it fills the remaining verl-side integration gap while verl still defaults to `vanilla_mbridge=True` and forwards `override_transformer_config` through its own bridge/config construction path.

YaRN fields such as `position_embedding_type`, `rotary_base`, `yarn_rotary_scaling_factor`, and `yarn_original_max_position_embeddings` are runtime config fields read by GPTModel. In the current verl path, passing them as ordinary constructor/override kwargs can be rejected or filtered before GPTModel sees them. This patch extracts the HF YaRN settings, avoids passing runtime-only fields through constructor kwargs, and restores them on the final config/provider.

This is useful beyond a private model config: [YaRN](https://arxiv.org/abs/2309.00071) is a published RoPE-based method for extending model context windows, and the [Qwen2 Technical Report](https://arxiv.org/abs/2407.10671) reports using YaRN together with Dual Chunk Attention to support long-context processing up to 131,072 tokens. Preserving these HF YaRN fields lets the current verl Megatron/mBridge path load mainstream long-context model configs without losing the runtime RoPE settings.

This shim can become unnecessary once verl defaults to, or pins, a bridge stack that preserves `rope_scaling` / `rope_parameters` and `override_transformer_config` YaRN fields end-to-end.

## Tests

- `PYTHONPATH=. pytest tests/utils/megatron/test_yarn_rope_config_on_cpu.py -q`
- `python -m py_compile verl/utils/megatron/rope_config.py verl/workers/engine/megatron/transformer_impl.py tests/utils/megatron/test_yarn_rope_config_on_cpu.py`
- `python -m ruff check verl/utils/megatron/rope_config.py verl/workers/engine/megatron/transformer_impl.py tests/utils/megatron/test_yarn_rope_config_on_cpu.py`
- `git diff --check HEAD~1..HEAD`

### DIFF
--- a/tests/utils/megatron/test_yarn_rope_config_on_cpu.py
+++ b/tests/utils/megatron/test_yarn_rope_config_on_cpu.py
@@ -1,0 +1,121 @@
+# Copyright 2024 Bytedance Ltd. and/or its affiliates
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import importlib.util
+from pathlib import Path
+from types import SimpleNamespace
+
+
+def _load_rope_config_module():
+    module_path = Path(__file__).parents[3] / "verl" / "utils" / "megatron" / "rope_config.py"
+    spec = importlib.util.spec_from_file_location("rope_config", module_path)
+    module = importlib.util.module_from_spec(spec)
+    spec.loader.exec_module(module)
+    return module
+
+
+def test_extracts_yarn_runtime_config_from_legacy_rope_scaling():
+    hf_config = SimpleNamespace(
+        rope_theta=1000000.0,
+        rope_scaling={
+            "type": "yarn",
+            "factor": 4.0,
+            "original_max_position_embeddings": 32768,
+        },
+    )
+
+    rope_config = _load_rope_config_module()
+
+    assert rope_config.get_hf_yarn_rope_config(hf_config, {}) == {
+        "position_embedding_type": "yarn",
+        "rotary_base": 1000000.0,
+        "yarn_rotary_scaling_factor": 4.0,
+        "yarn_original_max_position_embeddings": 32768,
+        "yarn_beta_fast": 32.0,
+        "yarn_beta_slow": 1.0,
+        "yarn_mscale": 1.0,
+        "yarn_mscale_all_dim": 0.0,
+        "yarn_correction_range_round_to_int": True,
+    }
+
+
+def test_prefers_nested_rope_parameters_over_rope_scaling():
+    hf_config = SimpleNamespace(
+        rope_theta=10000.0,
+        rope_scaling={"type": "linear", "factor": 2.0},
+        rope_parameters={
+            "short": {"rope_type": "default", "rope_theta": 10000.0},
+            "long": {
+                "rope_type": "yarn",
+                "rope_theta": 500000.0,
+                "factor": 8.0,
+                "original_max_position_embeddings": 65536,
+                "beta_fast": 48.0,
+                "beta_slow": 2.0,
+            },
+        },
+    )
+
+    rope_config = _load_rope_config_module()
+
+    yarn_config = rope_config.get_hf_yarn_rope_config(hf_config, {})
+
+    assert yarn_config["position_embedding_type"] == "yarn"
+    assert yarn_config["rotary_base"] == 500000.0
+    assert yarn_config["yarn_rotary_scaling_factor"] == 8.0
+    assert yarn_config["yarn_original_max_position_embeddings"] == 65536
+    assert yarn_config["yarn_beta_fast"] == 48.0
+    assert yarn_config["yarn_beta_slow"] == 2.0
+
+
+def test_explicit_override_values_take_precedence_without_mutating_input():
+    hf_config = SimpleNamespace(
+        rope_scaling={
+            "rope_type": "yarn",
+            "factor": 4.0,
+            "original_max_position_embeddings": 32768,
+        },
+    )
+    override_transformer_config = {
+        "attention_backend": "flash",
+        "rotary_base": 123456.0,
+        "yarn_rotary_scaling_factor": 16.0,
+        "yarn_mscale": 0.5,
+    }
+
+    rope_config = _load_rope_config_module()
+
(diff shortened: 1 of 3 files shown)


tntnnlrw:
Hi @HollowMan6 @wuxibin89, could either of you take a look when you have bandwidth?

This is a focused Megatron/mBridge config-path fix: it preserves HF YaRN rope fields from `rope_parameters` / `rope_scaling`, keeps those runtime-only fields out of mBridge constructor override kwargs, and sets them back on the built Megatron config/provider so GPTModel can read them.

I also addressed the Gemini feedback in `5e9f61e8` by using `self.yarn_runtime_config` directly instead of the defensive `getattr(...)`.

Validation run:
- `PYTHONPATH=. pytest tests/utils/megatron/test_yarn_rope_config.py -q`
- `python -m py_compile verl/utils/megatron/rope_config.py verl/workers/engine/megatron/transformer_impl.py tests/utils/megatron/test_yarn_rope_config.py`
- `git diff --check HEAD~1..HEAD`

HollowMan6:
You can contribute these to Megatron Bridge, which should be the right place to handle these things if not already: https://github.com/NVIDIA-NeMo/Megatron-Bridge

tntnnlrw:
Hi @HollowMan6, following up on the Megatron-Bridge point.

I agree that the generic HF-to-Megatron-Bridge mapping is the right long-term home. The reason I still think this verl-side patch is useful is that the immediate issue is in the verl integration path. The Megatron runtime consumes YaRN as fields on the final Megatron config/provider, e.g. `position_embedding_type`, `rotary_base`, `yarn_rotary_scaling_factor`, `yarn_original_max_position_embeddings`, etc. In our Megatron side, GPTModel builds the rope/yarn/mrope embeddings from these config-derived values, including heterogeneous/per-layer rotary settings.

verl currently builds its own override path around mBridge/Megatron-Bridge. Some YaRN fields are runtime config fields consumed by GPTModel, but they are not always safe constructor override kwargs for mBridge/provider construction. This patch extracts them from HF `rope_parameters` / `rope_scaling`, lets explicit `override_transformer_config` values take precedence, filters them out of bridge constructor overrides, and applies them back onto the final provider/config before GPTModel uses them.

So I see this PR as a compatibility/integration shim for the current verl mBridge path. I am happy to upstream the generic mapping to Megatron-Bridge as well, but until verl depends on a Megatron-Bridge version that handles these fields end-to-end, this keeps the current verl path from silently losing YaRN runtime parameters.

I also renamed the regression to `test_yarn_rope_config_on_cpu.py` so the new test follows the verl CPU-only test convention. Local validation passes with:
- `PYTHONPATH=. pytest tests/utils/megatron/test_yarn_rope_config_on_cpu.py -q`
- `python -m ruff check verl/utils/megatron/rope_config.py verl/workers/engine/megatron/transformer_impl.py tests/utils/megatron/test_yarn_rope_config_on_cpu.py`

Could you take another look when you have bandwidth, and if you have permissions, approve/run the GitHub Actions workflows? The current fork PR runs are stuck at `action_required` / no checks reported.

HollowMan6:
> The reason I still think this verl-side patch is useful is that the immediate issue is in the verl integration path. The Megatron runtime consumes YaRN as fields on the final Megatron config/provider, e.g. position_embedding_type, rotary_base, yarn_rotary_scaling_factor, yarn_original_max_position_embeddings, etc.

FYI, I think the issue might have already been addressed in 
https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/cc3a7d88e54599de691b8f3d56c069fafdd3d554/src/megatron/bridge/models/conversion/model_bridge.py#L457-L467

tntnnlrw:
@HollowMan6 Thanks for the pointer. You are right that the non-vanilla Megatron-Bridge path now handles the `rope_scaling.type == "yarn"` to `yarn_*` provider mapping. I updated the PR summary to narrow the scope accordingly.

This PR is intended as a verl-side integration shim for the current default `vanilla_mbridge=True` path and verl override/config finalization path:

- the current `ISEEKYAN/mbridge` path does not appear to apply the generic GPT YaRN runtime mapping (`position_embedding_type="yarn"`, `yarn_rotary_scaling_factor`, etc.); for LLaMA-like configs it still maps `rope_scaling.factor` to `seq_len_interpolation_factor`
- HF configs in newer Transformers versions can expose YaRN under nested `rope_parameters`, not only flat `rope_scaling`
- verl forwards `override_transformer_config` through bridge/config construction, where these runtime fields may be rejected or filtered before GPTModel sees them; this patch filters them out of constructor kwargs and restores them on the final config/provider

So this is not meant to duplicate the generic Megatron-Bridge mapping. It only keeps the current verl default mBridge path from silently losing YaRN runtime settings until verl defaults to, or pins, a bridge stack that preserves `rope_scaling` / `rope_parameters` and override YaRN fields end-to-end.

HollowMan6:
cc: @ISEEKYAN for any thoughts, but I think probably you can still contribute to https://github.com/ISEEKYAN/mbridge directly

tntnnlrw:
Following up after the Megatron-Bridge/mbridge discussion.

@HollowMan6 @ISEEKYAN, I narrowed the intended scope here to the current verl default `vanilla_mbridge=True` integration path. This PR is not trying to duplicate the generic YaRN mapping that newer Megatron-Bridge has; it only preserves YaRN runtime fields for the path where verl goes through `ISEEKYAN/mbridge` and then finalizes the Megatron config/provider before GPTModel consumes it.

The specific gap this covers is that HF YaRN settings can come from `rope_parameters` / `rope_scaling`, while fields such as `position_embedding_type`, `rotary_base`, `yarn_rotary_scaling_factor`, and `yarn_original_max_position_embeddings` are runtime fields for the final Megatron config/provider rather than always-safe bridge constructor kwargs. The patch extracts these fields, keeps constructor overrides clean, and applies them back to the final config/provider, with a CPU-only regression test for this behavior.

Would this scoped verl-side compatibility shim be acceptable while verl still supports the current default mBridge path? The fork PR also still has no checks reported, so it may need maintainer-side workflow approval before CI can run.

HollowMan6:
Now there're some conflicts with main. Also since we are deprecating vanilla mbridge, I'll close this for now, please feel free to reopen if you still think this is necessary

PR #6656, opened 2026-06-08
fix(agent_loop): handle text-only input in _compute_position_ids for VLM models (e.g. Qwen3.5)

luoshijiang:
### Problem

When running a vision-language model (VLM) or a native VLM (e.g. Qwen3.5) on **text-only tasks** like GSM8K math reasoning, the training crashes with:
```bash
  TypeError: 'NoneType' object is not an iterator
    File "verl/experimental/agent_loop/agent_loop.py", line 816, in _compute_position_ids
      vision_position_ids, _ = self.processor.get_rope_index(...)
    File "transformers/models/qwen3_vl/modeling_qwen3_vl.py", line 1114, in get_rope_index
      grid_thw = next(grid_iters[modality_type])
```
### Root Cause

`_compute_position_ids` always calls `get_rope_index` for VLM models, even when the input contains **no images or videos**.
This passes `None` for `image_grid_thw`/`video_grid_thw` to `get_rope_index`, which crashes when trying to iterate over `None`.

  The crash is triggered when the generated response text contains tokens that collide with the processor's `image_token_id` (fallback from `tokenizer.convert_tokens_to_ids("<|image_pad|>")` on a text-only tokenizer), causing `mm_token_type_ids` to be incorrectly marked with non-zero values.

### Fix

  Add an early return in `_compute_position_ids` when both `image_grid_thw` and `video_grid_thw` are `None`. This falls back to the existing `compute_position_id_with_mask` path, which is already used when `self.processor is None`.
  This is consistent with the existing defense pattern in the same method and handles the common use case of running VLMs on text-only RL tasks.

### Testing

  - Verified fix allows Qwen3.5 + GSM8K RL training to run past step 100 without crashes.
  - No behavior change for multimodal inputs (images/videos present) — the early return only triggers when both grid_thw values are None.

### DIFF
--- a/verl/experimental/agent_loop/agent_loop.py
+++ b/verl/experimental/agent_loop/agent_loop.py
@@ -831,9 +831,17 @@ class AgentLoopWorker:
         if self.processor is None:
             return compute_position_id_with_mask(attention_mask)  # (1, seq_len)
 
+        image_grid_thw = multi_modal_inputs.get("image_grid_thw")
+        video_grid_thw = multi_modal_inputs.get("video_grid_thw")
+
+        # For text-only input (no images/videos), skip vision rope and use text position ids.
+        # This handles VLM models running on text-only tasks (e.g. GSM8K reasoning).
+        if image_grid_thw is None and video_grid_thw is None:
+            return compute_position_id_with_mask(attention_mask)  # (1, seq_len)
+
         multi_modal_kwargs = {
-            "image_grid_thw": multi_modal_inputs.get("image_grid_thw"),
-            "video_grid_thw": multi_modal_inputs.get("video_grid_thw"),
+            "image_grid_thw": image_grid_thw,
+            "video_grid_thw": video_grid_thw,
         }
         # For transformers>=5.3.0, mm_token_type_ids is only used to calculate position ids.
         if multi_modal_inputs.pop("mm_token_type_ids", None) is not None:
--- a/verl/experimental/agent_loop/agent_loop.py
+++ b/verl/experimental/agent_loop/agent_loop.py
@@ -837,7 +837,8 @@ class AgentLoopWorker:
         # For text-only input (no images/videos), skip vision rope and use text position ids.
         # This handles VLM models running on text-only tasks (e.g. GSM8K reasoning).
         if image_grid_thw is None and video_grid_thw is None:
-            return compute_position_id_with_mask(attention_mask)  # (1, seq_len)
+            text_position_ids = compute_position_id_with_mask(attention_mask)
+            return text_position_ids.unsqueeze(1).expand(-1, 4, -1)  # (1, 4, seq_len)
 
         multi_modal_kwargs = {
             "image_grid_thw": image_grid_thw,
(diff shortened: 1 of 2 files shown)


verl/experimental/agent_loop/agent_loop.py
```diff
@@ -796,30 +826,29 @@ def _compute_multi_modal_inputs(self, output, input_ids) -> dict[str, torch.Tens
...
-        attention_mask,
-        multi_modal_inputs,
-        mm_processor_kwargs: Optional[dict[str, Any]] = None,
-    ) -> torch.Tensor:
+    def _compute_position_ids(self, input_ids, attention_mask, multi_modal_inputs) -> torch.Tensor:
         """Compute position ids for multi-modal inputs."""
         if self.processor is None:
             return compute_position_id_with_mask(attention_mask)  # (1, seq_len)
 
+        image_grid_thw = multi_modal_inputs.get("image_grid_thw")
```
Luosuu, line 834: shall we add test cases for this?

luoshijiang: Thanks for the review! I have added test cases in `tests/experimental/agent_loop/test_text_only_position_ids.py`, covering both branches of the fix: 

1. **`test_text_only_early_return`** — Tests the new early return path (line 837-840): when `image_grid_thw=None` and `video_grid_thw=None`, returns `(1, 4, seq_len)` without calling `get_rope_index`.

2. **`test_vlm_path_unchanged`** — Tests the normal VLM path still works when vision data is present (`image_grid_thw=torch.tensor([[1, 16, 16]])`), producing the same `(1, 4, seq_len)` output.

Both are lightweight unit tests.

wuxibin89:
Which transformers do you use? I have test Qwen3.5+gsm8k with transformers==5.3.0, no crash happen.

luoshijiang:
@wuxibin89 Thanks for checking. My transformers version is `5.3.0.dev0`, from official docker image: `quay.io/ascend/verl:verl-8.5.2-910b-ubuntu22.04-py3.11-qwen3-5`.

The crash is probabilistic. In my training, it triggered around step ~100 when the model's entropy exploded (from ~0.3 to >2.0), causing the policy to generate random tokens. One of those tokens happened to collide with the processor's `image_token_id` (fallback from `tokenizer.convert_tokens_to_ids("<|video_pad|>")`), which set `mm_token_type_ids` to non-zero. Then `get_rope_index` tried to iterate over `None` grid data and crashed.

It probably didn't hit this collision because the model's output stayed stable. The fix makes the early return explicit, so it won't depend on token collision luck.

wuxibin89:
I have test with NPU, no crash happen.
- hardware: Ascend910B2 * 64
- image: quay.io/ascend/verl:verl-9.0.0-910b-ubuntu22.04-py3.11-latest
  - vllm==0.18.0
  - veomni==0.1.11
  - transformers==5.9.0
- model: Qwen/Qwen3.5-35B-A3B
- dataset: BytedTsinghua-SIA/DAPO-Math-17k

PR #6666, opened 2026-06-09
[vllm, checkpoint] feat: add rollout weight sync debug check

le-czs:
## What does this PR do?

Addresses #6414.

Adds a checkpoint engine debug check that verifies whether vLLM loaded the same weights sent by the trainer during the initial trainer-to-rollout sync. This PR is limited to vLLM with tensor and data parallel sizes 1.

The check compares the source Hugging Face checkpoint state dict with the rollout backend's loaded state dict. It handles vLLM fused `qkv_proj` and `gate_up_proj` weights, tied `lm_head.weight`, and reports missing, unexpected, and mismatched keys. Value comparison is strict after casting the source checkpoint tensor to the loaded rollout tensor dtype/device.

This PR intentionally does not include `check_weight_sync_only` or fully async shutdown cleanup. Follow-up stacked draft PR: https://github.com/le-czs/verl/pull/1. It will be rebased and opened against `verl-project/verl` after this PR lands.

This is not duplicating an existing PR. I checked these searches before opening the PR:

- https://github.com/verl-project/verl/pulls?q=is%3Apr+is%3Aopen+6414+in%3Abody
- https://github.com/verl-project/verl/pulls?q=is%3Apr+is%3Aopen+weight+sync+check+checkpoint+engine+vllm

Related open PRs are about MTP sleep, FP8 transfer, generic remote backends, or fully async transfer queue. I did not find an open PR that adds a vLLM loaded-weight comparison for #6414.

AI assistance was used for implementation and PR preparation. I reviewed the changed lines before submission.

## Checklist Before Starting

- [x] I have searched for similar PRs to avoid duplicate work. Query links are included above.
- [x] I have formatted the PR title as `[vllm, checkpoint] feat: add rollout weight sync debug check`.

## Test

- [x] `python3 -m pytest tests/checkpoint_engine/test_weight_sync_on_cpu.py -q`
  - Result: `11 passed, 1 warning`
- [x] `scripts/generate_trainer_config.sh`
  - Result: `All good`
- [x] `git diff --check`
  - Result: passed
- [x] `pre-commit run --all-files --show-diff-on-failure --color=always`
  - Result: passed

## API and Usage Example

Example overrides:

```bash
actor_rollout_ref.rollout.data_parallel_size=1 \
actor_rollout_ref.rollout.checkpoint_engine.backend=nccl \
actor_rollout_ref.rollout.checkpoint_engine.check_weight_sync=True
```

`check_weight_sync=True` runs the loaded-weight comparison after the initial trainer-to-rollout sync.

## Design & Code Changes

- Adds checkpoint engine config field `check_weight_sync`.
- Adds `verl.checkpoint_engine.weight_sync` for strict state dict comparison and vLLM fused-weight mapping.
- Adds a vLLM server RPC path that returns loaded-weight comparison results from the rollout backend.
- Restricts the current vLLM check to TP=1 and DP=1; DP>1 is rejected because the current `collective_rpc` path only covers one DP shard.
- Runs the check after the initial weight sync, because the comparison source is the original HF checkpoint.
- Adds CPU tests for manager check selection, strict mismatch reporting, vLLM fused comparison, unsupported parallelism rejection, and missing/unexpected keys.
- Adds docs in `docs/advance/checkpoint.rst` and `verl/checkpoint_engine/README.md`.

## Checklist Before Submitting

- [x] I have read the [Contribute Guide](https://github.com/verl-project/verl/blob/main/CONTRIBUTING.md) and followed the PR process.
- [x] I have run pre-commit checks: `pre-commit run --all-files --show-diff-on-failure --color=always`.
- [x] I have added or updated documentation.
- [x] I have added tests or explained why tests are not feasible.
- [ ] I have requested CI in the required channel, if needed.
- [ ] If this PR changes recipes, I have updated the recipe submodule.

### DIFF
--- a/verl/checkpoint_engine/base.py
+++ b/verl/checkpoint_engine/base.py
@@ -343,9 +343,6 @@ class CheckpointEngineWorker(Worker):
         self,
         model_path: str | None = None,
         trust_remote_code: bool | None = None,
-        torch_dtype: str = "bfloat16",
-        max_mismatches: int = 10,
-        timeout: float | None = None,
     ):
         """Check backend-loaded rollout weights against a HuggingFace checkpoint."""
         check_fn = getattr(self.server_adapter, "check_loaded_weights_equal", None)
@@ -356,9 +353,6 @@ class CheckpointEngineWorker(Worker):
         return await check_fn(
             model_path=model_path or self.model_config.path,
             trust_remote_code=self.model_config.trust_remote_code if trust_remote_code is None else trust_remote_code,
-            torch_dtype=torch_dtype,
-            max_mismatches=max_mismatches,
-            timeout=timeout,
         )
 
 
@@ -456,11 +450,7 @@ class CheckpointEngineManager:
         if not self.config.check_weight_sync:
             return None
 
-        summaries = rollout.check_loaded_weights_equal(
-            torch_dtype=self.config.check_weight_sync_dtype,
-            max_mismatches=self.config.check_weight_sync_max_mismatches,
-            timeout=self.config.check_weight_sync_timeout,
-        )
+        summaries = rollout.check_loaded_weights_equal()
         summaries = [summary for summary in summaries if summary is not None]
         print(f"[CheckpointEngineManager] weight sync check passed: {summaries}")
         return summaries
@@ -469,7 +459,7 @@ class CheckpointEngineManager:
         """Whether this update should be compared with the source HF checkpoint."""
         if not self.config.check_weight_sync:
             return False
-        return self.config.check_weight_sync_only or global_steps in (None, 0)
+        return global_steps in (None, 0)
 
     @auto_await
     async def sleep_replicas(self):
--- a/verl/checkpoint_engine/base.py
+++ b/verl/checkpoint_engine/base.py
@@ -344,8 +344,6 @@ class CheckpointEngineWorker(Worker):
         model_path: str | None = None,
         trust_remote_code: bool | None = None,
         torch_dtype: str = "bfloat16",
-        rtol: float = 1e-5,
-        atol: float = 1e-8,
         max_mismatches: int = 10,
         timeout: float | None = None,
     ):
@@ -359,8 +357,6 @@ class CheckpointEngineWorker(Worker):
             model_path=model_path or self.model_config.path,
             trust_remote_code=self.model_config.trust_remote_code if trust_remote_code is None else trust_remote_code,
             torch_dtype=torch_dtype,
-            rtol=rtol,
-            atol=atol,
             max_mismatches=max_mismatches,
             timeout=timeout,
         )
@@ -462,8 +458,6 @@ class CheckpointEngineManager:
 
         summaries = rollout.check_loaded_weights_equal(
             torch_dtype=self.config.check_weight_sync_dtype,
-            rtol=self.config.check_weight_sync_rtol,
-            atol=self.config.check_weight_sync_atol,
             max_mismatches=self.config.check_weight_sync_max_mismatches,
             timeout=self.config.check_weight_sync_timeout,
         )
--- a/verl/checkpoint_engine/base.py
+++ b/verl/checkpoint_engine/base.py
@@ -338,6 +338,33 @@ class CheckpointEngineWorker(Worker):
         """Get leader rank flag from the underlying rollout server adapter."""
         return self.server_adapter.is_leader_rank
 
+    @register(dispatch_mode=Dispatch.ONE_TO_ALL)
+    async def check_loaded_weights_equal(
+        self,
+        model_path: str | None = None,
+        trust_remote_code: bool | None = None,
+        torch_dtype: str = "bfloat16",
+        rtol: float = 1e-5,
+        atol: float = 1e-8,
+        max_mismatches: int = 10,
+        timeout: float | None = None,
+    ):
+        """Check backend-loaded rollout weights against a HuggingFace checkpoint."""
+        check_fn = getattr(self.server_adapter, "check_loaded_weights_equal", None)
+        if check_fn is None:
+            raise NotImplementedError(
+                f"Rollout backend {self.rollout_config.name!r} does not support loaded-weight equality checks."
+            )
+        return await check_fn(
+            model_path=model_path or self.model_config.path,
+            trust_remote_code=self.model_config.trust_remote_code if trust_remote_code is None else trust_remote_code,
+            torch_dtype=torch_dtype,
+            rtol=rtol,
+            atol=atol,
+            max_mismatches=max_mismatches,
+            timeout=timeout,
+        )
+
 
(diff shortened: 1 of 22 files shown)


le-czs:
Verification Result

image

verl/checkpoint_engine/base.py
```diff
@@ -428,6 +455,28 @@ def remove_replicas(self, replicas: list[RolloutReplica]):
...
+        if not self.config.check_weight_sync:
+            return None
+
+        summaries = rollout.check_loaded_weights_equal(
+            torch_dtype=self.config.check_weight_sync_dtype,
+            rtol=self.config.check_weight_sync_rtol,
+            atol=self.config.check_weight_sync_atol,
+            max_mismatches=self.config.check_weight_sync_max_mismatches,
+            timeout=self.config.check_weight_sync_timeout,
+        )
```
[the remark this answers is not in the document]
le-czs: `check_loaded_weights_equal` is registered without `blocking=False`, so `register(...)` uses the default `blocking=True`. When RayWorkerGroup binds the method, `func_generator` calls `ray.get(output)` before returning. That means `rollout.check_loaded_weights_equal(...)` already returns resolved results, and worker-side AssertionError should propagate to the caller.
Adding another `ray.get(...)` here would call `ray.get` on the resolved list, not on ObjectRefs.

verl/trainer/config/_generated_ppo_trainer.yaml
```diff
@@ -313,6 +313,13 @@ actor_rollout_ref:
       update_weights_bucket_megabytes: 2048
       engine_kwargs: {}
       custom_backend_module: null
+      check_weight_sync: false
+      check_weight_sync_only: false
+      check_weight_sync_dtype: bfloat16
+      check_weight_sync_rtol: 1.0e-05
```
Luosuu: why weight sync has rtol/atol? should it be strictly zero?

le-czs: Good point, I agree the default should be strict for the current supported
  path. I’ll change `rtol/atol` to `0.0` by default.

  The reason I added the parameters is to keep an explicit escape hatch for
  future/debug cases, e.g. mixed-dtype, quantized/FP8, or backend-specific
  conversion paths. But those should require users to opt in by overriding the
  tolerance, not be enabled by default.

Luosuu: even for mixed-dtype/fp8, the atol/rtol should still be zero before and after sync?

le-czs: I think your point makes sense. For `check_weight_sync`, strict equality before and after sync is the cleaner semantic.

  I’ll remove the configurable tolerance knobs and keep the check strict-only

Luosuu (commented):
Left two inline comments on correctness issues in the new debug path.

verl/workers/rollout/vllm_rollout/utils.py
```diff
@@ -219,6 +219,53 @@ def _iter_all_models_with_config(self):
...
+        parallelism, vLLM stores sharded tensors and direct state_dict comparison
+        is not meaningful without applying each layer's sharding rule.
+        """
+        from transformers import AutoModelForCausalLM
+
+        from verl.checkpoint_engine.weight_sync import assert_weight_sync_equal
+        from verl.utils.fs import copy_to_local
+
+        parallel_config = self.model_runner.vllm_config.parallel_config
+        if parallel_config.tensor_parallel_size != 1:
```
Luosuu: This also needs to reject or cover `data_parallel_size > 1`. The call path uses vLLM `collective_rpc`, and existing code notes that only reaches TP workers within one DP shard, so `TP=1, DP>1` can pass while other DP shards were never checked.

verl/workers/rollout/llm_server.py
```diff
@@ -371,3 +371,14 @@ async def start_profile(self, **kwargs):
     async def stop_profile(self):
         """Stop profiling on all rollout replicas."""
         await asyncio.gather(*[replica.stop_profile() for replica in self.rollout_replicas])
+
+    @auto_await
+    async def shutdown(self):
+        """Shutdown rollout replicas that expose a shutdown hook."""
```
Luosuu: In fully async check-only mode this misses hybrid replicas: `FullyAsyncLLMServerManager` moves them into `hybrid_replicas` and clears `rollout_replicas` until activation. Please override shutdown there, or otherwise include `hybrid_replicas.values()`, so check-only exits do not leave hybrid server actors running.

verl/workers/rollout/vllm_rollout/utils.py
le-czs: Fixed by taking the reject path: `check_weight_sync` now supports only TP=1 and DP=1, and rejects DP>1 instead of reporting a partial check as passed.

verl/workers/rollout/llm_server.py
le-czs: Fixed by overriding `FullyAsyncLLMServerManager.shutdown()` to include both `rollout_replicas` and `hybrid_replicas.values()`, with de-duplication.

Luosuu (changes requested):
I think this PR touches too much for the weight sync debug check and actually this PR can be broken down into two parts:
1. a minimal version that check weight sync for vLLM (DP/TP=1) 
2. Another PR for check_weight_sync_only and fully async clean shutdown

please try to make each PR minimal and easy to review. Thank you!

le-czs:
> I think this PR touches too much for the weight sync debug check and actually this PR can be broken down into two parts:
> 
> 1. a minimal version that check weight sync for vLLM (DP/TP=1)
> 2. Another PR for check_weight_sync_only and fully async clean shutdown
> 
> please try to make each PR minimal and easy to review. Thank you!

I reduced #6666 to the minimal vLLM TP=1/DP=1 weight-sync check.
I also prepared the follow-up for `check_weight_sync_only` and fully async cleanup as a stacked draft PR: https://github.com/le-czs/verl/pull/1. Once #6666 lands, I’ll rebase it and open it against `verl-project/verl`.

le-czs:
> I think this PR touches too much for the weight sync debug check and actually this PR can be broken down into two parts:
> 
> 1. a minimal version that check weight sync for vLLM (DP/TP=1)
> 2. Another PR for check_weight_sync_only and fully async clean shutdown
> 
> please try to make each PR minimal and easy to review. Thank you!

 Thanks. I’ve reduced this PR to the minimal vLLM TP=1/DP=1 check and moved the follow-up scope out.

  Could you re-review when you have time? The previous change request should be addressed now.

le-czs:
Hi @Luosuu, I rebased the branch and the previous inline threads are now resolved/outdated. The PR is still blocked by the old `CHANGES_REQUESTED` review.

  Could you re-review when you have time? If the scope looks OK now, I’ll request CI.

PR #6684, opened 2026-06-10
[fully_async, trainer] fix: sync optimizer total steps before trainer initialization

mikequan0425:
### What does this PR do?

It was observed in experiments that the learning rate (lr) was always 0. This issue does not occur when starting the script via main_ppo, and only emerges under fully async mode. Refer to the following for detailed problem description:https://github.com/verl-project/verl/issues/6683

In short, optim.total_training_steps is not assigned correctly during the initialization of trainer optim.

image

### Design & Code Changes

The simplest approach is to pass values through by configuring `actor_rollout_ref.actor.optim.total_training_steps`, yet the actual trainer step should be calculated as total_rollout_steps / (required_samples * trigger_parameter_sync_step)

Therefore, this PR provides a method to assign the corresponding value to optim.total_training_steps when creating the trainer.

### Checklist Before Starting

- [x] Search for similar PRs. Paste at least one query link here: https://github.com/verl-project/verl/pulls?q=is%3Apr+is%3Aopen+fully+async+lr

### Test

image

### API and Usage Example

```python
```

> [!IMPORTANT]

### DIFF
--- a/verl/experimental/fully_async_policy/fully_async_main.py
+++ b/verl/experimental/fully_async_policy/fully_async_main.py
@@ -20,7 +20,7 @@ from pprint import pprint
 
 import hydra
 import ray
-from omegaconf import OmegaConf, open_dict
+from omegaconf import OmegaConf
 
 from verl.experimental.fully_async_policy.fully_async_rollouter import FullyAsyncRollouter
 from verl.experimental.fully_async_policy.fully_async_trainer import FullyAsyncTrainer
@@ -77,7 +77,7 @@ class FullyAsyncTaskRunner:
         self._init_rollouter_before_trainer(config)
 
         optim_total_train_steps = ray.get(self.components["rollouter"].get_optim_total_train_steps.remote())
-        self._set_total_training_steps_for_optimizer(config, optim_total_train_steps)
+        FullyAsyncTrainer.set_total_training_steps_for_optimizer(config, optim_total_train_steps)
 
         print("[ASYNC MAIN] Creating FullyAsyncTrainer first (needed for hybrid worker group injection)...")
         self._create_trainer(config)
@@ -118,17 +118,6 @@ class FullyAsyncTaskRunner:
             ray.get(self.components["trainer"]._fit_validate.remote(True))
 
         print("[ASYNC MAIN] All components initialized successfully")
-
-    def _set_total_training_steps_for_optimizer(self, config, optim_total_train_steps):
-        try:
-            OmegaConf.set_struct(config, True)
-            with open_dict(config):
-                if OmegaConf.select(config, "actor_rollout_ref.actor.optim"):
-                    config.actor_rollout_ref.actor.optim.total_training_steps = optim_total_train_steps
-                if OmegaConf.select(config, "critic.optim"):
-                    config.critic.optim.total_training_steps = optim_total_train_steps
-        except Exception as e:
-            print(f"Warning: Could not set total_training_steps in config. Structure missing? Error: {e}")
 
     def _init_rollouter_before_trainer(self, config) -> None:
         """
--- a/verl/experimental/fully_async_policy/fully_async_main.py
+++ b/verl/experimental/fully_async_policy/fully_async_main.py
@@ -20,7 +20,7 @@ from pprint import pprint
 
 import hydra
 import ray
-from omegaconf import OmegaConf
+from omegaconf import OmegaConf, open_dict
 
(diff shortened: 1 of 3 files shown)


verl/experimental/fully_async_policy/fully_async_trainer.py
```diff
@@ -266,7 +271,16 @@ def set_total_train_steps(self, total_training_steps):
...
 
-        self.progress_bar = tqdm(total=self.total_train_steps, initial=0, desc="Training Progress")
+    def _resolve_total_training_steps_before_init(self):
+        rollout_total_steps = self.config.rollout.total_rollout_steps
+        if rollout_total_steps is None:
+            return None
+
+        required_samples = (
+            self.config.actor_rollout_ref.actor.ppo_mini_batch_size * self.config.async_training.require_batches
+        )
```
Luosuu: `optim.total_training_steps` is consumed by the LR scheduler, and the scheduler is stepped on every actor update (`_fit_update_actor` -> `update_actor`), not only when parameters are synced. Dividing by `trigger_parameter_sync_step` makes the schedule finish `trigger_parameter_sync_step` times too early; if the progress bar/checkpoint version wants sync steps, please keep that separate from the optimizer scheduler steps.

verl/experimental/fully_async_policy/fully_async_trainer.py
```diff
@@ -266,7 +271,16 @@ def set_total_train_steps(self, total_training_steps):
         except Exception as e:
             print(f"Warning: Could not set total_training_steps in config. Structure missing? Error: {e}")
```
Luosuu: This uses the raw configured `rollout.total_rollout_steps`, but the rollouter later computes the effective count as `min(config.rollout.total_rollout_steps, len(train_dataloader) * total_epochs)` and also supports `None`. Since the optimizer is already constructed during `init_workers`, the later `set_total_train_steps()` call cannot rebuild the scheduler, so dataset-limited or unset configs still get the wrong LR schedule here.

Luosuu (commented):
Found two issues in the pre-init total step calculation that should be addressed before this is safe.

verl/experimental/fully_async_policy/fully_async_trainer.py
tardis-key: I think the best solution is to finish initializing the rollout part first, and then have the trainer use the rollout_step that was set by rollout for subsequent calculations. Otherwise the total-step computation logic for rollout would need to be maintained in multiple places, which makes further evolution difficult.

verl/experimental/fully_async_policy/fully_async_trainer.py
mikequan0425: Thanks for your review. I originally ported it directly from the set_total_train_steps method in rollouter, and I will fix it.

verl/experimental/fully_async_policy/fully_async_trainer.py
mikequan0425: I agree with both of you. But we cannot simply move create_rollouter before create_trainer. This is because actor_wg only exists within the trainer after trainer.init_workers() is called, and rollouter.init_workers() will read this hybrid worker group during its initialization process. Meanwhile, we can move the logic for sync total_train_steps between the rollouter and the trainer to before create_trainer, and remove the original corresponding code.

tardis-key (changes requested):

verl/experimental/fully_async_policy/fully_async_main.py
```diff
@@ -114,14 +119,39 @@ def _initialize_components(self, config) -> None:
 
         print("[ASYNC MAIN] All components initialized successfully")
 
-    def _create_rollouter(self, config) -> None:
-        print("[ASYNC MAIN] Starting create rollouter...")
+    def _set_total_training_steps_for_optimizer(self, config, optim_total_train_steps):
```
tardis-key: don't move this func to main, it should belong to Trainer. You can change the func to class/static func, then you can call it before initing Trainer

tardis-key (approved):
@Luosuu lgtm, please review

Luosuu (approved):

Luosuu:
@mikequan0425 please fix pre-commit

wuxibin89:
The verl/experimental/fully_async_policy is in maintenance mode, we encourage user to use new V1 trainer:
```bash
trainer.use_v1=True
trainer.v1.trainer_mode=colocate_async # sync/colocate_async/separate_async
```

PR #6685, opened 2026-06-10
[worker] fix actor offload before rollout weight resume

BeihaiTianming:
### What

This PR changes the colocated actor/rollout weight-sync ordering in `ActorRolloutRefWorker.update_weights` to reduce peak device memory before rollout weight memory is resumed.

The new order is:

1. Build the actor weight generators needed for rollout update.
2. Consume those generators before actor offload and materialize tensors on CPU.
3. Offload actor parameters to CPU and clear cached device memory.
4. Resume rollout weight memory.
5. Sync rollout weights.
6. Resume rollout KV cache.

### Why

In colocated actor/rollout deployments with `free_cache_engine=True`, `rollout.resume(tags=["weights"])` wakes rollout weight memory. If this happens before actor parameters are offloaded, rollout weights and actor parameters can overlap in device memory and cause OOM.

I observed this on NPU/Ascend hardware with 64G per card, where memory headroom is smaller and this peak is easier to trigger.

### Historical context:

PR #5031 (`ef6eaa05`, `[ckpt] feat: add CheckpointEngineManager`) replaced the old direct `sleep()` / `wake_up()` rollout transition with the current `update_weights()` path in `ActorRolloutRefWorker`.

In that path, rollout weight memory is resumed before actor parameters are offloaded:

1. `rollout.resume(tags=["weights"])`
2. `actor.engine.get_per_tensor_param(...)`
3. `rollout.update_weights(...)`
4. `actor.engine.to("cpu", model=True, optimizer=False, grad=False)`
5. `rollout.resume(tags=["kv_cache"])`

This ordering can make rollout weights and actor parameters overlap in device memory. On NPU/Ascend 64G cards, this peak can trigger OOM. This PR keeps the weight-sync semantics but moves actor offload/cache cleanup before rollout weight resume.

### Generator handling

`get_per_tensor_param()` returns generators for several engines. To avoid lazy generator evaluation after actor offload, this PR consumes the generators before actor placement changes.

Instead of using `list(generator)` directly and retaining full model tensors on device, tensors are materialized as CPU copies via `tensor.detach().cpu()`. This keeps the weight stream independent from actor placement while preserving the lower device-memory peak.

### LoRA / SGLang behavior

The SGLang LoRA adapter path keeps the required base-first ordering:

1. Prepare adapter/merged params.
2. Prepare base params when base sync is needed.
3. Offload actor and resume rollout weight memory.
4. Sync base weights first, then adapter/merged weights.

### Tests

- `py -3.12 -m py_compile verl/workers/engine_workers.py tests/workers/test_engine_workers_lora_sync.py`
- `PYTHONUTF8=1 uv run --with pytest --no-project pytest tests/workers/test_engine_workers_lora_sync.py -q`
  - `17 passed`

### Notes

AI assistance was used to prepare and verify this change.

### DIFF
--- a/verl/workers/engine_workers.py
+++ b/verl/workers/engine_workers.py
@@ -700,14 +700,12 @@ class ActorRolloutRefWorker(Worker, DistProfilerExtension):
             return
 
         set_expandable_segments(False)
-        log_gpu_memory_usage("Before resume weights", logger=logger)
 
-        # 1. resume rollout memory (weights were released during sleep)
-        if self.config.rollout.free_cache_engine:
-            await self.rollout.resume(tags=["weights"])
-        log_gpu_memory_usage("After resume weights", logger=logger)
+        def materialize_weights_to_cpu(weights):
+            # Consume lazy weight generators before actor offload without retaining full weights on device.
+            return [(name, tensor.detach().cpu()) for name, tensor in weights]
 
-        # 2. determine if we need a base weight sync (adapter path only)
+        # 1. determine if we need a base weight sync (adapter path only)
         per_tensor_param, peft_config = self.actor.engine.get_per_tensor_param(
             layered_summon=self.layered_summon, base_sync_done=True
         )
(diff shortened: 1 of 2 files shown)


verl/workers/engine_workers.py
```diff
@@ -717,11 +715,29 @@ async def update_weights(self, global_steps: int = None, mode: str = "auto"):
...
         if do_lora_base_sync:
-            per_tensor_param_base, peft_config = self.actor.engine.get_per_tensor_param(
+            per_tensor_param_base, peft_config_base = self.actor.engine.get_per_tensor_param(
                 layered_summon=self.layered_summon, base_sync_done=False
             )
+            peft_config = peft_config_base
+
+        # 3. offload actor model before rollout weight resume to reduce peak device memory
+        if self.actor.engine.is_param_offload_enabled:
+            per_tensor_param = materialize_weights_to_cpu(per_tensor_param)
```
wuxibin89, line 728: It's a very bad idea to materialize full model into CPU. For a 1T model, you need 4TB (weights in fp32) cpu memory.

Issue #6401, opened 2026-05-19
[RFC] Prefix-Tree Shared Attention for Multi-Turn RL Training


arvyanh:
## 1. Summary

In GRPO and multi-trajectory RL training, each prompt (or each turn) is sampled `n` times — all responses share an identical prefix, yet standard training recomputes it independently for each sample. This RFC proposes packing all samples into a flat `[prefix | leaf_0 | ... | leaf_{n-1}]` layout and running a single forward pass with cross-leaf isolation enforced via [Magi Attention](https://github.com/SandAI-org/MagiAttention)'s workload-balanced CP dispatch — mathematically equivalent to independent per-sample forwards.

Inspired by [Forge](https://www.minimax.io/news/forge-scalable-agent-rl-framework-and-algorithm)'s *Prefix Tree Merging*, this implementation targets VERL's Megatron backend (FSDP planned) and generalizes to **arbitrary-depth multi-level trees**, enabling prefix sharing across multiple conversation turns — not just the top-level prompt.

## 2. Motivation

In GRPO and multi-trajectory RL training, each prompt (or each turn) is sampled `n` times, and all `n` responses share an identical prefix. Standard training computes attention over the full sequence for each sample independently — paying redundant O(P²) cost for the shared prefix on every forward pass. At high prefix ratios (e.g. long system prompts, multi-turn history), this dominates total compute.

**Background & use cases:** The RL algorithms that targets rollout sampling between turns, or grpo situation that have long common prefixes. 
Example includes:
- [rStar-Math](https://arxiv.org/abs/2501.04519) performs "deep thinking" through Monte Carlo Tree Search, where *"a math policy SLM performs test-time search guided by an SLM-based process reward model"* — MCTS rollouts naturally produce shared-prefix trees as the search expands. 
- [TreeRL](https://arxiv.org/abs/2506.11902) *"directly incorporates on-policy tree search for RL training … strategically branching from high-uncertainty intermediate steps"*, where each branch shares the prefix up to its branch point. Both map directly onto the tree layout this RFC proposes. 
- [DeepSearch](https://arxiv.org/abs/2509.25454) "embeds MCTS directly into the RLVR training loop, branching from intermediate reasoning steps" — the search tree is rooted at the input prefix and expands across steps, producing the same shared-prefix structure at train time that rStar-Math uses at test time. (ICLR 2026)
- And other situation with long common prefixes

### Overview

[...]

PR #6689, opened 2026-06-11
# [model, trainer, engine] feat: prefix-tree MAGI attention for verl SFT and RL (draft)

arvyanh:
## What does this PR do?

The RFC: https://github.com/verl-project/verl/issues/6401

Adds prefix-tree shared-prefix deduplication for verl SFT and GRPO training using [MAGI attention](https://github.com/SandAI-org/magi-attention). Token-by-token trie detection discovers shared prefixes across rollout samples without requiring rollout-side metadata. The flat token layout + attention rectangle spec is dispatched to MAGI `calc_attn`, which computes correct prefix-tree attention patterns while internally deduplicating shared KV tokens.

### Key features

- **Dynamic trie detection**: `build_tree_dynamic` builds a compressed trie from input tokens — no prior knowledge of turn boundaries needed.
- **Multi-level tree support**: handles multi-turn rollout sharing with zero-length leaf nodes for samples that terminate at intermediate levels.
- **MAGI integration**: patches Megatron-LM's `TEDotProductAttention` through the `SelfAttention → TransformerLayer → GPTModel` chain to inject MAGI/flex attention.
- prefix-length aware dynamic bsz grouping and mbs reordering
- **Configurable old-log-prob backend**: Currently the old_log_prob's log_prob seems to diverge

## Design

This PR implements the design described in RFC #6401. All `n` GRPO rollout samples are packed into a single flat `[prefix | leaf_0 | ... | leaf_{n-1}]` sequence, run through **one** transformer forward pass, with cross-leaf attention blocked. The result is mathematically equivalent to `n` independent forwards.

### Attention rectangle spec

The attention pattern is encoded as rectangle specs:

```
          k: prefix    k: leaf0    k: leaf1
q: prefix   causal       ✗           ✗
q: leaf0     full      causal         ✗
q: leaf1     full        ✗         causal
```

MAGI interprets these rectangles natively.

### Prefix detection (dynamic trie)

Prefix detection runs entirely at train time — no rollout-side metadata required. `build_tree_dynamic` performs token-by-token compressed trie insertion over the micro-batch's input sequences. The trie is converted to a `TreeNode` tree, then fed to `build_prefix_tree_attention_spec`, which emits `(q_ranges, k_ranges, mask_types)` for every node.

but in case we could retrieve the tree at runtime, we could provide the trie data to the traininer 

The tree generalizes to arbitrary-depth multi-level trees. For example, in multi-turn agent RL where responses branch into sub-groups sharing a turn-2 prefix:

```
S0: turn0 + turn1_A + turn2_B1
S1: turn0 + turn1_A + turn2_B2
S2: turn0 + turn1_C + turn2_D1
S3: turn0 + turn1_C + turn2_D2
```

This produces a depth-3 tree with `turn0` as root, `turn1_A`/`turn1_C` as intermediate nodes, and four leaves. Zero-length leaf nodes are inserted when a sample terminates at an intermediate node.

### Megatron integration

The patch chain injects `magi_attention_key` through upstream Megatron-LM:

```
GPTModel.forward → TransformerBlock → TransformerLayer
    → SelfAttention → TEDotProductAttention (magi_attn_forward: dispatch → calc_attn → undispatch)
```

### File summary

| Area | Path |
|---|---|
| Trie + load balancing | `verl/utils/prefix_tree/dynamic.py` |
| Layout + attention spec | `verl/utils/prefix_tree/utils.py` |
| MAGI key + batch | `verl/utils/prefix_tree/magi.py` |
| Trainer helpers | `verl/utils/prefix_tree/trainer.py` |
| Megatron patch | `verl/models/mcore/prefix_tree_merge.py` |

## Result
full curves on wandb: <https://wandb.ai/arvyanh-mt/verl_prefixtree>

For Single turn:
using longreason for a long prompt situation

For MultiTurn agentic case:
The model used is Qwen3-4B, trained using uni-agent search tool with asearch dataset
With average prompt length of 800 and response length of 1400
using dynamic batch size (note the tree experiment here can fix more sample into microbatch under same length due to less activation)

| Workload | ratio (%) | update_actor (s) | e2e (s) |
|---|---|---|---|
| base longreason | – | 82 | 186 |
| magi longreason | 80 | 21 | 69 |
| base asearch | – | 133 | 370 |
| magi asearch | 35 | 82 | 300 |

### GPU training
Validated on Megatron-LM training with GRPO on CoQA — prefix-tree MAGI attention runs correctly with CP=2, SP=1, PP=1.

## 7. Known limitations

- **TP and SP are mutually exclusive.** When `tensor_model_parallel_size > 1`, sequence parallelism must be disabled (`actor_rollout_ref.actor.megatron.sequence_parallel=False`); the prefix-tree path does not support SP
- **Linear attention is not supported yet.** Only the `magi` / `flex` / FA3-fallback attention backends work; linear-attention variants fall back to the standard path.
- **Dispatch overhead can dominate at low sharing.** Building the trie and dispatching the packed layout has a fixed cost; when the shared-prefix length is very short the dedup saving may not cover it, giving a net negative gain.
- **DP load balance is not fully optimized.** The current DFS + contiguous partition balances token counts across DP ranks, but does not yet optimally co-locate prefix-sharing samples across ranks in all cases.

## Checklist Before Starting

- [X] Search for similar PRs: [upstream verl prefix-tree PRs](https://github.com/verl-project/verl/pulls?q=tree)
- [X] Format the PR title as `[{modules}] {type}: {description}`
- [X] Read the [Contribute Guide](https://github.com/volcengine/verl/blob/main/CONTRIBUTING.md)
- [X] Apply pre-commit checks


supercharleszhu:
@wuxibin89 @arvyanh wondering if we have any plan to merge this pR? We are exploring the same problem and was trying to fix it using prefix grouping, but looks like magi attention has more general application.

arvyanh:
> @wuxibin89 @arvyanh wondering if we have any plan to merge this pR? We are exploring the same problem and was trying to fix it using prefix grouping, but looks like magi attention has more general application.

I should have all the features avaliable now, you could test use my branch before it got merged.

Begunner:
Please resolve conflicts. By the way, the [wandb](https://wandb.ai/arvyanh-mt/verl_prefixtree) is 404.

verl/utils/prefix_tree/dynamic.py
```diff
@@ -0,0 +1,717 @@
...
+    return True
+
+
+def balance_prefix_tree_v1(batch, trie, metrics: dict, dp_size: int, logging_prefix: str):
+    """v1 (KVBatchMeta) adapter over :func:`balance_prefix_tree_blocks`: reorder the
+    batch so each DP rank receives whole uid blocks balanced by flat-token workload.
+
+    Synthetic padding samples (``is_padding`` tag) are excluded from the balance and
+    re-appended after the permutation, mirroring v0's pure-batch balancing."""
+    block_ids = [k.rsplit("_", 1)[0] for k in batch.keys]
```
Begunner: The v1 key is formatted as `{uid}_{session_id}_{index}`, where uid identifies the prompt group and session_id identifies a sampling session (trajectory). If DP balancing is intended to operate on prompt groups, rsplit("_", 2)[0] should be used to extract the uid.

verl/utils/prefix_tree/dynamic.py
```diff
@@ -0,0 +1,717 @@
...
+        )
+
+    use_dynamic_bsz_local = tu.get_non_tensor_data(data, "use_dynamic_bsz", default=True)
+    # Per-position uid for atomic block grouping: never split a uid block
+    # (same prompt's rollouts) across micro-batches. v1 keys are
+    # {uid}_{session_id}_{index}; classic carries non_tensor_batch["uid"].
+    if hasattr(data, "non_tensor_batch") and data.non_tensor_batch.get("uid") is not None:
+        uid_list = list(data.non_tensor_batch["uid"])
+    elif hasattr(data, "keys") and not callable(data.keys):
+        uid_list = [k.rsplit("_", 1)[0] for k in data.keys]
```
Begunner: The same problem.

verl/utils/prefix_tree/forward.py
```diff
@@ -0,0 +1,711 @@
...
+        causal = kv_idx <= q_idx
+        return (in_prefix_k & causal) | (same_leaf & causal) | (in_prefix_k & (q_leaf >= 0))
+
+    # _compile=False: avoid Triton JIT which takes minutes for new shapes.
+    # Memory is handled at the call site via torch.utils.checkpoint.
+    block_mask = create_block_mask(
+        prefix_tree_mask, B=None, H=None, Q_LEN=total, KV_LEN=total, device=device, _compile=False
+    )
+    block_mask._leaf_id = leaf_id  # keep closure alive
+    return block_mask
```
Begunner: Is it right for multi-layer prefix tree?
For example:
```
P
├── A
│   ├── R0
│   └── R1
└── B
    ├── R2
    └── R3
```

verl/utils/prefix_tree/forward.py
```diff
@@ -0,0 +1,711 @@
...
+    boundary_tags: list[tuple[int, int]] = []
+    if magi_key is not None and registry:
+        for boundary_position, leaves in registry:
+            matches = (local_indices == boundary_position).nonzero()
+            if matches.shape[0] == 0:
+                continue  # boundary hidden lives on another CP rank
+            local_idx = int(matches[0, 0].item())
+            for sample_idx, next_token in leaves:
+                boundary_pairs.append((local_idx, int(next_token)))
+                boundary_tags.append((boundary_position, sample_idx))
```
Begunner: Flex next-token label also needs to be processed.

verl/utils/prefix_tree/forward.py
```diff
@@ -0,0 +1,711 @@
...
+        flat_args = {
+            k: v for k, v in orig_args.items() if k not in ("label", "temperature", "loss_mask", "use_prefix_tree")
+        }
+
+        if prefix_tree_attention == "magi":
+            local_indices = get_position_ids(pt_batch.magi_key)
+            flat_padded = pt_batch.tree_packed_input_ids.shape[0]
+            logits_full = logits_flat.new_zeros((flat_padded, logits_flat.shape[-1]))
+            logits_full[local_indices] = logits_flat
+            logits_flat = logits_full[:real_tokens]
```
Begunner: Maybe CP gather here.

Begunner:
Please add some unit tests and flex backend E2E experiments.

Jackie2049:
Great work 👍 This really struck a chord with us.

**Over the past few months, we have been looking into the same problem**: redundant prefix computation during RL training, particularly in agentic RL workloads such as StepRL and TreeRL. These workloads need to reuse shared prefixes of arbitrary lengths unlike PrefixGrouper and this makes sharing KV states much more challenging.

Our understanding is that this PR builds a Prefix Tree dynamically and uses MAGI/Flex Attention to pack tree-structured samples into a single forward pass. We find this approach particularly interesting because it does not require the rollout stage to provide the tree structure ahead of time. **It is very close to the problem we have been working on.**

**We recently shared some of our own exploration in a [psRL](https://arxiv.org/pdf/2608.25683) and an early [PoC implementation](https://github.com/Jackie2049/PrefixSharing).** The two approaches seem complementary:

1. This PR removes redundant computation through sparse attention. While we did not find a suitable custom GPU kernel and our prototype instead manages and concatenates cached KV states. This is less efficient than sparse-attention based on MAGI/Flex, but it has **very few dependencies on the device type and environment requirements**. 
2. We also prioritized **FSDP support** to make the feature easier to adopt, and **kept the integration with verl lightweight** through patches and plugins. We verified numerical parity with the baseline for both forward and backward passes.

We would love to better understand a few details of the current implementation:

1. Within a forward pass, is each shared prefix computed exactly once, with **only one copy of its KV states computed and materialized** without andy additional KV duplication? 
2. Is reuse currently limited to a single micro-batch? Can the global Prefix Tree carry computed prefix states **across micro-batch boundaries**? This may help in DP load-balance according to our experiments.
3. Are there **plans to support FSDP and additional hardware backends**? Just from usability perspective :)

Anyway, we are excited to see this work moving forward and would **be glad to collaborate** rather than pursue the same problem in isolation. Like FSDP support, pluginable designs, more backends support... We would love to hear where you think our experience could be useful and **explore follow-up contributions**. 🤝

Jackie2049:
Hi all, a quick update from our side.

After reviewing this implementation, we think the overall technical direction is sound and the Magi/Flex backend design makes sense. 

We also ran some initial tests and found the Flex path still has a few rough edges. For example: **even when FlexAttention is explicitly selected, some modules still attempt to import MagiAttention and fails if it was not installed**. This undermines usability as some users lack prerequisites for MagiAttention (device, environments...).

**We have prepared a lightweight fix for it: [our PR is here](https://github.com/meituan-search/verl/pull/62)**, where the FlexAttention path can be launched independently and safely without MagiAttention dependencies.

Could you please review and merge it :)

Jackie2049:
> Hi all, a quick update from our side.
> 
> After reviewing this implementation, we think the overall technical direction is sound and the Magi/Flex backend design makes sense.
> 
> We also ran some initial tests and found the Flex path still has a few rough edges. For example: **even when FlexAttention is explicitly selected, some modules still attempt to import MagiAttention and fails if it was not installed**. This undermines usability as some users lack prerequisites for MagiAttention (device, environments...).
> 
> **We have prepared a lightweight fix for it: [our PR is here](https://github.com/meituan-search/verl/pull/62)**, where the FlexAttention path can be launched independently and safely without MagiAttention dependencies.
> 
> Could you please review and merge it :)

Beyond this fix, we would be happy to make further follow-up contributions to this feature, like better test coverage (as @Begunner suggested) and helping mature the FlexAttention path (where we have experience when building psRL & PrefixSharing). We’ll stay active in this discussion and look forward to keeping in close touch with you folks. Let's refine prefix-tree together. :) 🤝🛠

Issue #400, opened 2025-02-27
TP rollout + FSDP / TP actor


jeromeku:
Suppose I have a single node with 8 gpus.  Assuming rollout and actor co-location, is it possible to run the rollouts (using vllm) with tensor parallel 8 while running the actor training step using FSDP2 (or TP)?  

When syncing weights, I'd imagine one would leverage the `redistribute` DTensor api to unshard / reshard to pass to vllm.

PeterSH6:
Yes, it's possible. You can easily support FSDP2 upon current fsdp_workers.py. Simply modifying the FSDP wrapper to FSDP2's may be enough.

When syncing the weights, we may not need to utilize `redistribute` at this moment because we can simply call `full_tensor` and shard the parameter in vllm weight loader.

But in fact, using `redistribute` with some sharding spec would be an elegant design.

Are you interested in supporting FSDP2 or TorchTitan? Feel free to propose an RFC and PR

SwordFaith:
I'm interested about torchtitan or fsdp2. But I'm curious about if using fsdp2 and torchtitan is an easier way to support 4D parallel and fp8 training?

vermouth1992:
I guess either torchtitan or megatron is a valid approach

vadimkantorov:
For FSDP2, should Verl include more explicit controls for TP? E.g. using `parallelize_module(...)` like in torchtune https://github.com/pytorch/torchtune/blob/main/recipes/full_finetune_distributed.py and provide PyTorch's native context-parallel SDPA variant (as alternative to Ulysses) - curious how it will fare

Kirrito-k423:
## 中文

更新一下当前状态：这个需求现在已经被部分覆盖，但我不建议直接关闭这个 issue。

证据如下：

- FSDP2 支持已经在 #1026 合并，PR 描述中给出了 `actor_rollout_ref.actor.strategy=fsdp2` / `actor_rollout_ref.ref.strategy=fsdp2` 的用法：https://github.com/verl-project/verl/pull/1026
- README 当前也把 FSDP2 列为训练 backend，并说明可以通过 `actor_rollout_ref.actor.strategy=fsdp2`、`actor_rollout_ref.ref.strategy=fsdp2`、`critic.strategy=fsdp2` 启用：https://github.com/verl-project/verl/blob/main/README.md
[...]

Issue #250, opened 2025-02-12
Load checkpoint from default_local_dir & Save hdfs checkpoints


cliangyu:
I think there might be a bug, if the resume_mode is auto and default_local_dir is set and default_hdfs is also set. should load checkpoint_folder from self.config.trainer.default_local_dir, we simply dont rusume from default_hdfs_dir. However, with the current code, there will be error if resume_mode == 'auto' and both default_local_dir and default_hdfs_dir are provided.

A quick fix `verl/trainer/ppo/ray_trainer.py`:
```python
def _load_checkpoint(self):
        if self.config.trainer.resume_mode == 'disable':
            return 0

        checkpoint_folder = self.config.trainer.default_local_dir
        if not os.path.isabs(checkpoint_folder):
            working_dir = os.getcwd()
            checkpoint_folder = os.path.join(working_dir, checkpoint_folder)
            
        global_step_folder = find_latest_ckpt_path(checkpoint_folder)  # None if no latest

        # If no local checkpoint found and hdfs path is configured, try hdfs
        # Note: Currently hdfs loading is not implemented, so this will be skipped
        if global_step_folder is None and self.config.trainer.default_hdfs_dir is not None:
            print("HDFS checkpoint loading is not implemented yet, falling back to local path only")
            # TODO: Implement hdfs loading here in the future
            # global_step_folder = find_latest_ckpt_path_hdfs(self.config.trainer.default_hdfs_dir)

        if self.config.trainer.resume_mode == 'auto':
            if global_step_folder is None:
                print('No checkpoint found. Training from scratch')
                return 0
        else:
            if not (self.config.trainer.resume_from_path and global_step_folder is not None):
                assert isinstance(self.config.trainer.resume_mode, str), "resume ckpt must be str type"
                assert 'global_step_' in self.config.trainer.resume_mode, "resume ckpt must specify the global_steps"
                global_step_folder = self.config.trainer.resume_mode
                if not os.path.isabs(global_step_folder):
                    working_dir = os.getcwd()
                    global_step_folder = os.path.join(working_dir, global_step_folder)

        print(f'Load from checkpoint folder: {global_step_folder}')
        # set global step
        self.global_steps = int(global_step_folder.split('global_step_')[-1])

        print(f'Setting global step to {self.global_steps}')
        print(f'Resuming from {global_step_folder}')

        actor_path = os.path.join(global_step_folder, 'actor')
        critic_path = os.path.join(global_step_folder, 'critic')
        # load actor
[...]

PR #6699, opened 2026-06-11
[fsdp, trainer] fix: detach model_output and loss metrics to stop per-micro-batch graph retention in actor update

### LINKED ISSUE #6698: Per-micro-batch GPU memory leak in engine actor update: model_output/metrics retain autograd graph (OOM with LoRA + long sequences)
## Summary

The unified engine's actor update leaks GPU memory **per training micro-batch**: tensors returned in `forward_step`'s `model_output` (and loss tensors stored in the metrics dict) still carry `grad_fn`, and `forward_backward_batch` holds them in `output_lst` until the *entire* batch finishes. The retained autograd graphs pin large activations for every micro-batch that runs a backward pass. With LoRA + long packed sequences this reliably OOMs the very first `update_actor` on an 80GB GPU.

Affected code (present on `main` @ 41a5244):
- `verl/workers/engine/fsdp/transformer_impl.py` — `FSDPEngineWithLMHead.forward_step` returns `output = {"model_output": model_output, ...}` without detaching `log_probs` / `entropy`.
- ~~`verl/workers/utils/losses.py` — `ppo_loss` stores live tensors in metrics~~ (correction: `Metric.append` already does `value.detach().item()` for scalar tensors, so metrics never retain the graph — the `forward_step` `model_output` is the sole anchor).

## Setup where it bites

- Qwen3-8B + LoRA (rank 32, all-linear) → PEFT calls `enable_input_require_grads()`
- multi-turn tool-agent rollouts, `use_dynamic_bsz` packed micro-batches up to 16k tokens, `use_remove_padding`, NO_PADDING, fsdp2 (`fsdp_size=1`), gradient checkpointing on
- one actor update = 40 prompts x GRPO n=8 = 320 trajectories ≈ 400 micro-batches (incl. old_log_prob / rollout-correction recompute passes)
- observed via `verl.experimental.fully_async_policy`, but the leaky code is the shared engine path that `main_ppo` also uses (`verl/trainer/main_ppo.py` imports `verl.workers.engine_workers.ActorRolloutRefWorker`)

## Evidence (torch memory-history profiling)

Per-micro-batch `torch.cuda.memory_allocated()` inside one `update_actor`:

| phase | micro-batches | allocated |
|---|---|---|
| no-grad recompute passes | #1–#250 | flat 15.6 GiB |
| training pass (backward) | #250–#400 | **+0.27 GiB per micro-batch**, 24.8 → 64 GiB → OOM |

`_record_memory_history` snapshots show each training micro-batch permanently retains a **pair of `[total_nnz, hidden]` bf16 tensors (~139 MiB each at 16k tokens)**:
1. the **embedding output** (alloc stack: `at::native::embedding` ← `F.embedding`) — it `requires_grad` because of PEFT's `enable_input_require_grads()`, so the non-reentrant checkpoint frame saves it, and the retained graph keeps the frame alive after `loss.backward()`;
2. its **gradient buffer** produced at the tail of backward (allocated on the autograd thread).

Detaching only the metrics did *not* fix it; detaching `model_output` (+ metrics) does:

| | #250 | #300 | #350 | #400 |
|---|---|---|---|---|
| before fix | 24.8 GiB | 37.9 GiB | ~50 GiB | 64 GiB → OOM |
| after fix | 16.2 GiB | 16.2 GiB | 16.2 GiB | 16.2 GiB ✅ |

Training proceeds normally after the fix (validated end-to-end on 2-GPU disaggregated fully_async, LoRA GRPO multi-turn).

## Why it usually goes unnoticed

Without `enable_input_require_grads` (full-param training) the embedding output isn't pinned by the checkpoint frame, so the per-micro-batch retention is small; with multi-GPU DP the leak is also sharded. LoRA + long sequences + many micro-batches per update + a single train GPU concentrate it into ~40 GiB per update. Possibly related symptom reports: #4002, #3293.

## Proposed fix

Detach at the two collection points (behavior-neutral — these tensors are only consumed for metrics/postprocessing after backward):
1. `forward_step`: detach tensors in `model_output` before building the returned `output` dict;
2. `ppo_loss`: store `pg_loss.detach()` / `entropy_loss.detach()` / `kl_loss.detach()` and detach `pg_metrics` values.

PR incoming.

## System Info

- verl: main @ 41a5244 (leak confirmed by code inspection) and the 0.9.0.dev-era unified engine path (reproduced + profiled)
- torch 2.8.0, vllm 0.11.0, transformers 4.56.1, CUDA A800-80GB, python 3.12
- fsdp2, `use_dynamic_bsz`, `use_remove_padding`, NO_PADDING pad mode, gradient checkpointing on, LoRA rank 32 (all-linear)

YoungZSh:
### What does this PR do?

Fixes a per-micro-batch GPU memory leak in the unified engine's actor update that OOMs long-sequence LoRA runs. `FSDPEngineWithLMHead.forward_step` returns `model_output` (log_probs/entropy) still attached to the autograd graph; `forward_backward_batch` holds these per-micro-batch outputs in `output_lst` until the whole batch finishes, so the retained graph pins — per training micro-batch — the activation-checkpoint frame's saved embedding output (which requires grad under PEFT `enable_input_require_grads`) plus its gradient buffer. Full root-cause analysis with memory-history profiling in #6698. Fixes #6698.

### Checklist Before Starting

- [x] Search for similar PRs: https://github.com/verl-project/verl/pulls?q=is%3Apr+detach+model_output — no existing PR addresses this
- [x] Format the PR title as `[{modules}] {type}: {description}` → `[fsdp, trainer] fix: ...`

### Test

**Regression test added**: `tests/workers/test_engine_forward_step_detach_on_cpu.py` — miniature reproduction of the retention mechanism (frozen embedding + `requires_grad_` output + checkpointed block + weakref on the checkpoint-saved input). Verified red on the unfixed engine, green with the fix. CPU-only, runs in `cpu_unit_tests`.

End-to-end validation (Qwen3-8B + LoRA rank 32 GRPO, multi-turn tool-calling, 16k-token dynamic-bsz micro-batches, fsdp2 + gradient checkpointing, ~400 micro-batches/update). Per-micro-batch `torch.cuda.memory_allocated()` during the training pass:

| | mb #250 | #300 | #350 | #400 |
|---|---|---|---|---|
| before | 24.8 GiB | 37.9 GiB | ~50 GiB | 64 GiB → **OOM** |
| after | 16.2 GiB | 16.2 GiB | 16.2 GiB | 16.2 GiB ✅ |

After the fix the run reaches step 1+ with healthy metrics (grad_norm 0.015, rollout/training log-prob Pearson corr 0.993, peak allocated 31.5/79 GiB).

### API and Usage Example

No API change. The detached tensors are only consumed for metrics aggregation and postprocessing after `loss.backward()` has already run.

### Design & Code Changes

1. `verl/workers/engine/fsdp/transformer_impl.py` — `forward_step`: detach tensors in `model_output` before building the returned `output` dict (the live `loss` is still returned separately for `backward()`). Also covers the critic via `FSDPEngineWithValueHead`.
2. `tests/workers/test_engine_forward_step_detach_on_cpu.py` — regression guard (see Test above).

Note: an earlier commit on this branch also detached metric tensors in `ppo_loss`; that turned out to be redundant — `Metric.append` already does `value.detach().item()` for scalar tensors — and has been reverted, keeping the diff minimal.

- [x] Apply [pre-commit checks](https://github.com/verl-project/verl/blob/main/CONTRIBUTING.md#code-linting-and-formatting): ruff clean; diff matches surrounding style.
- [x] Documentation: N/A (no user-facing behavior change).
- [x] Added regression test `tests/workers/test_engine_forward_step_detach_on_cpu.py` (CPU, runs in cpu_unit_tests).
- [x] CI requested via the Feishu group (thanks @Tianle Zhong for triggering!).

### DIFF
--- a/tests/workers/test_engine_forward_step_detach_on_cpu.py
+++ b/tests/workers/test_engine_forward_step_detach_on_cpu.py
@@ -1,0 +1,114 @@
+# Copyright 2026 Bytedance Ltd. and/or its affiliates
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Regression guard for verl#6698.
+
+FSDPEngineWithLMHead.forward_step used to return ``model_output`` tensors
+(log_probs/entropy) still attached to the autograd graph. forward_backward_batch
+collects these per-micro-batch outputs in ``output_lst`` until the WHOLE batch
+finishes, so the retained graph pinned, per training micro-batch, whatever the
+graph still referenced after backward — most notably the activation-checkpoint
+frame's saved block input (the embedding output, which requires grad under
+PEFT's ``enable_input_require_grads``) and its gradient buffer. On long-sequence
+LoRA runs this leaked ~0.27 GiB per micro-batch and OOM'd the actor update.
+
+This test reproduces the retention mechanism in miniature: a frozen embedding
+whose output is forced to require grad (the PEFT situation), a checkpointed
+block, and a weakref on the checkpoint-saved block input. After backward, with
+``output`` still held (as output_lst does), the saved input must be collectible
+and the returned model_output must carry no grad_fn.
+"""
+
+import gc
+import weakref
+from types import SimpleNamespace
+
+import torch
+from tensordict import TensorDict
+
+from verl.utils.device import get_device_id
+from verl.workers.engine.fsdp.transformer_impl import FSDPEngineWithLMHead
+
+VOCAB, HIDDEN, SEQ = 16, 8, 6
+
+
+class _TinyCheckpointedLM(torch.nn.Module):
+    """Frozen embedding + checkpointed trainable block, mimicking a PEFT/LoRA
+    base model with gradient checkpointing and enable_input_require_grads."""
+
+    def __init__(self):
+        super().__init__()
+        self.embed = torch.nn.Embedding(VOCAB, HIDDEN)
+        self.embed.weight.requires_grad_(False)  # frozen base
+        self.proj = torch.nn.Linear(HIDDEN, HIDDEN, bias=False)  # the trainable part
+        self.saved_block_input = None  # weakref to the checkpoint-saved input
+
+    def forward(self, input_ids=None, use_cache=False):
+        x = self.embed(input_ids)
+        # PEFT enable_input_require_grads: embedding output requires grad so
+        # gradients can flow into trainable params under checkpointing.
+        x.requires_grad_(True)
+        hidden = torch.utils.checkpoint.checkpoint(self.proj, x, use_reentrant=False)
+        self.saved_block_input = weakref.ref(x)
+        return SimpleNamespace(hidden=hidden)
+
+
+def _make_engine_stub():
+    """Bypass __init__; provide only what forward_step touches."""
+    eng = object.__new__(FSDPEngineWithLMHead)
+    eng._autocast_dtype = torch.float32  # skip autocast
+    eng.module = _TinyCheckpointedLM().to(get_device_id())
+    eng.prepare_model_inputs = lambda micro_batch: ({"input_ids": micro_batch["input_ids"]}, {})
+    eng.prepare_model_outputs = (
+        lambda output, output_args, micro_batch, logits_processor_func: {"log_probs": output.hidden.sum(-1)}
+    )
+    eng.get_data_parallel_group = lambda: None
+    return eng
+
+
+def _loss_fn(model_output, data, dp_group=None):
+    return model_output["log_probs"].sum(), {"dummy_metric": 1.0}
+
+
+def test_forward_step_output_carries_no_grad_fn_and_releases_graph():
+    eng = _make_engine_stub()
+    micro_batch = TensorDict(
+        {"input_ids": torch.randint(0, VOCAB, (1, SEQ))},
+        batch_size=[1],
+    )
+
+    loss, output = FSDPEngineWithLMHead.forward_step(eng, micro_batch, _loss_fn, forward_only=False)
+
+    # The live loss must still drive backward into the trainable params.
+    assert loss.grad_fn is not None
+    loss.backward()
+    assert eng.module.proj.weight.grad is not None
+
+    # Contract: nothing in the collected per-micro-batch output may keep the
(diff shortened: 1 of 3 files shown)


YoungZSh:
Triage of the two failing checks — both appear unrelated to this PR's diff (detach-only, no control-flow change):

1. **cpu_unit_tests** — `tests/workers/test_distillation_topk_symmetry_on_cpu.py::test_forward_kl_topk_emits_overlap_metrics` fails with `AttributeError: 'types.SimpleNamespace' object has no attribute 'use_chunked_topk'` (test mock config missing the newly added field). The same workflow also fails on `main`, including 41a5244 which this branch is based on.

2. **e2e_ppo_trainer_fsdp-qwen2_5vl-3b** — first attempt died in data loading (`DataLoader worker killed by signal: Killed`, i.e. host OOM-killer on the runner), and the in-job retry hit `EADDRINUSE` from the leftover SGLang server port. This workflow is green on `main@41a5244`; looks like runner flakiness — a re-run would help if a maintainer can trigger it.

All engine-related e2e suites that exercise the changed path (e2e_fully_async_policy_fsdp2, e2e_ppo_trainer_fsdp_vllm, gpu_unit_tests, ...) are green or still running.

YoungZSh:
Added the requested regression test: `tests/workers/test_engine_forward_step_detach_on_cpu.py`.

It reproduces the retention mechanism in miniature — frozen embedding whose output is forced to require grad (the PEFT `enable_input_require_grads` situation), a checkpointed trainable block, and a weakref on the checkpoint-saved block input. Holding the per-micro-batch `output` dict (as `forward_backward_batch.output_lst` does) must not keep the saved input alive after backward. Verified **red on the unfixed engine, green with the fix**; CPU-only so it runs in `cpu_unit_tests`.

While writing it I also noticed the earlier `ppo_loss` metrics-detach commit was redundant — `Metric.append` already does `value.detach().item()` for scalar tensors — so I reverted it. Final diff: the one-spot `forward_step` detach + the regression test.

wuxibin89:
@YoungZSh Please fix pre-commit fail: https://github.com/verl-project/verl/blob/main/CONTRIBUTING.md#code-linting-and-formatting

YoungZSh:
@wuxibin89 Fixed — the only pre-commit complaint was `ruff format` on the new test file; applied in 785a0ef.

On the other failing checks in the last round, none traces to this diff:

- `cpu_unit_tests` runs `pytest -x`, and `test_distillation_topk_symmetry_on_cpu.py` — broken on `main` as well (mock config missing `use_chunked_topk`) — fails first and aborts the session, so the new regression test hasn't actually executed in CI yet (verified red-on-unfixed / green-on-fixed locally).
- The 7 e2e/hardware failures (fully_async_fsdp2, ascend×2, rocm, megatron×2, reward_model_vllm) **all passed in the previous CI round with the byte-identical engine change**; the only delta since is the test file (not executed by e2e) and reverting `losses.py` to `main`'s content. The fully_async one additionally shows a 41.4 GiB foreign process on the 44.5 GiB runner GPU in its log.

Could you re-trigger CI when convenient? Thanks!

wuxibin89 (approved):

PR #6709, opened 2026-06-12
[reward] fix: only require max_resp_len when DAPO overlong penalty is enabled

### LINKED ISSUE #5858: DAPORewardManager maybe error
https://github.com/verl-project/verl/blob/622422786fac28c5039a6a2affd0e6e9a7d6ad55/verl/experimental/reward_loop/reward_manager/dapo.py#L39

maybe:
```python
```

discobot:
Fixes #5858.

### What does this PR do?

The constructor gate in both DAPO reward managers is `if self.overlong_buffer_cfg is not None:`, so the `max_resp_len` asserts fire even with `overlong_buffer.enable=False` — contradicting the assert's own suggestion to disable the penalty that way — while the penalty itself is only applied under an `.enable` check. The path is reachable from shipped scripts that toggle `enable` but always pass the cfg (e.g. `tests/experimental/agent_loop/test_basic_agent_loop.py`, `verl/experimental/fully_async_policy/shell/*.sh`).

The bug is not limited to the experimental manager flagged in the issue: `verl/workers/reward_manager/dapo.py` has the same ungated assert, plus a latent one — its `__call__` reads `self.overlong_buffer_cfg.enable` with no None guard, so the documented `overlong_buffer_cfg=None` default crashes with `AttributeError`. Both constructors now gate the asserts on the penalty being enabled, and `__call__` gets the None guard the experimental variant already has. `enable=True` behavior is unchanged.

### Checklist Before Starting

- [x] Search for similar PRs: https://github.com/verl-project/verl/pulls?q=is%3Apr+overlong_buffer — no open or merged PR addresses this, and the issue timeline has no cross-referenced PRs.
- [x] Title formatted as `[reward] fix: {description}`

### Test

Added `tests/workers/reward_manager/test_dapo_on_cpu.py`; the three new regression tests fail on main and pass with this change — `python -m pytest tests/workers/reward_manager/` → 13 passed, and `ruff check` / `ruff format --check` are clean on the touched files.

- [x] Read the Contribute Guide. Per `AGENTS.md`: AI assistance (Claude Code) was used for this change.
- [x] New tests are picked up by the existing `cpu_unit_tests.yml` workflow (`test_*_on_cpu.py` pattern). No docs change needed.

### DIFF
--- a/tests/workers/reward_manager/test_dapo_on_cpu.py
+++ b/tests/workers/reward_manager/test_dapo_on_cpu.py
@@ -1,0 +1,127 @@
+# Copyright 2026 Bytedance Ltd. and/or its affiliates
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import numpy as np
+import pytest
+import torch
+from omegaconf import OmegaConf
+
+from verl import DataProto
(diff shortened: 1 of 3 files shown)


Luosuu:
LGTM. I have triggered CI for you

Luosuu (approved):

Issue #6341, opened 2026-05-14
gemma3/4 processor support request


nursery42:
### Feature request

It would be greatly appreciated if you could support processor loading and multimodal training for Gemma 3/4.

### Motivation

The Gemma series is a key model family in the open-source community, but multimodal training for Gemma is not yet supported in VERL. We kindly request adding relevant support in future versions.

### Your contribution

I can help implement the corresponding code and conduct testing.

Kirrito-k423:
Thank you for your feature request.

**Current status**: verl has good support for Qwen2/3 VL models. Gemma3 VL support would be a new addition.

**Implementation considerations**:

1. **Processor support**: The  library has Gemma3 VL processor support. verl's  can be extended to include gemma3 processors similar to how qwen2_vl is handled.

2. **Model forward**: The monkey patching in  would need a corresponding  if Gemma3 VL has different attention patterns.

3. **Test coverage**: The  test in xtuner (PR #1776) suggests a pattern for VL model testing that could be adapted.

**Recommended approach**: 
- Check if  already has Gemma3 VL support
- Start with a simple test case using  and a small model
- Share your test script when you encounter issues

Would you like to share more details about your use case? Are you trying to run GRPO on Gemma3 VL for a specific task?

nursery42:
> Thank you for your feature request.
> 
> **Current status**: verl has good support for Qwen2/3 VL models. Gemma3 VL support would be a new addition.
> 
> **Implementation considerations**:
> 
> 1. **Processor support**: The  library has Gemma3 VL processor support. verl's  can be extended to include gemma3 processors similar to how qwen2_vl is handled.
> 2. **Model forward**: The monkey patching in  would need a corresponding  if Gemma3 VL has different attention patterns.
> 3. **Test coverage**: The  test in xtuner (PR [[docs] Add linux-arm64 platform tensordict package version problem handle FAQ #1776](https://github.com/verl-project/verl/pull/1776)) suggests a pattern for VL model testing that could be adapted.
> 
> **Recommended approach**:
> 
> * Check if  already has Gemma3 VL support
> * Start with a simple test case using  and a small model
> * Share your test script when you encounter issues
> 
> Would you like to share more details about your use case? Are you trying to run GRPO on Gemma3 VL for a specific task?

The Gemma model performs exceptionally well on minority language and multilingual tasks. Currently, VERL does not support multimodal training for the Gemma series, including processor loading. Because Gemma3 lacks the get_rope_index function.

[...]

PR #6030, opened 2026-04-16
[fsdp, sft] feat: add Gemma4 FSDP SFT support

khazic:
## Summary

This PR adds end-to-end SFT training support for Google Gemma4 31B via the FSDP path.

### Why this is not a duplicate

No existing open PR in verl-project/verl addresses Gemma4 FSDP SFT support.

### Changes

**Core fixes**

- `verl/utils/dataset/multiturn_sft_dataset.py`: Add `_normalize_gemma4_mm_token_type_ids()` and `_concat_gemma4_mm_token_type_ids()`; fix per-turn assembly loop so `mm_token_type_ids` is propagated for all turns.
- `verl/utils/flops_counter.py`: Fix `_estimate_gemma3_flops` to read config through `text_config` wrapper; register `gemma4` model type.
- `verl/utils/fsdp_utils.py`: Relax `get_fsdp_wrap_policy` hard-failure; unresolved layer class names fall back to `lambda_auto_wrap_policy`.
- `verl/workers/engine/fsdp/transformer_impl.py`: Wire `mm_token_type_ids` through FSDP forward pass with Ulysses SP slicing, jagged-tensor construction, device alignment, and padded-tensor expansion.

**Example script**

- `examples/sft/gsm8k/run_gemma4_31b.sh`: parameterized SFT entrypoint for Gemma4 31B; defaults to `sdpa` attention; activation checkpointing enabled.

**Note:** `model.use_remove_padding=False` is kept for now. FlashAttention and MoE support will be submitted in a follow-up PR.

**Tests**

- `tests/utils/dataset/test_multiturn_sft_dataset_on_cpu.py`
- `tests/utils/test_flops_counter.py`
- `tests/utils/test_fsdp2_peft_wrapping.py`
- `tests/utils/test_fsdp_transformer_impl_on_cpu.py`

## Test results

Verified on a single node of 8x H100 GPUs with `gemma-4-31B-it`:

```
step:1 - train/loss:2.2348194122314453 - train/grad_norm:155.1857452392578 - train/lr:1e-05 - train/mfu:0.047017663676604315 - train/global_tokens:2970 - train/total_tokens(B):2.97e-06
step:2 - train/loss:1.0301235914230347 - train/grad_norm:129.11013793945312 - train/lr:1e-05 - train/mfu:0.02550642394526722 - train/global_tokens:1750 - train/total_tokens(B):4.72e-06
[...]

PR #6715, opened 2026-06-12
[model] feat: support Gemma4 multimodal models in RL (GRPO)

dbuos:
## [model] feat: support Gemma4 multimodal models in RL

---
What does this PR do?

Adds support for Gemma4 multimodal models in RL (GRPO/PPO). verl's VLM plumbing assumed Qwen-VL-shaped processors — M-RoPE position ids, image_grid_thw, and strict per-image-token validation — so Gemma4 failed during rollout post-processing and the FSDP forward. The fixes are capability-detected (keyed on processor capabilities, not model name), so existing Qwen-VL / GLM-4V paths are unchanged. Addresses #6341 (complements #6030, which added Gemma4 FSDP SFT).

Checklist Before Starting

- [x] Search for similar PRs/Issues: (related: #6341, #6030)

## Test

New-model support — not coverable by CI without a multimodal GPU run. Validated end-to-end by experiment:

- Setup: google/gemma-4-E2B-it, GRPO, FSDP (actor/ref) + vLLM rollout, attn_implementation=sdpa, use_remove_padding=False, on H200.
- Result: trains stably; reward and eval metrics increase over training. <attach training-curve screenshot / before-after eval numbers>
- Existing Qwen-VL path re-verified unaffected (the new branches are never taken for Qwen — see below).

image
image

### API and Usage Example

No API changes (no new CLI args, config keys, or function signatures). Gemma4 just works once selected. Because Gemma4's global (full-attention) layers use head_dim=512 (FlashAttention-2/3 cap at 256), use SDPA:

actor_rollout_ref.model.path=google/gemma-4-E2B-it \
+actor_rollout_ref.model.override_config.attn_implementation=sdpa \
actor_rollout_ref.model.use_remove_padding=False

Design & Code Changes

verl hoists the model's internal multimodal prep (position ids, processor features) out of forward into its own rollout/forward glue, written to Qwen-VL's shape. This PR makes that glue handle non-Qwen VLMs (Gemma4) via capability checks. Three changes:

- verl/utils/tokenizer.py (hf_processor): add a Gemma4Processor case. Gemma uses standard 1D RoPE (no get_rope_index to bind), and its processor strictly validates one image placeholder per image — which rejects the placeholder-stripped text verl reconstructs at rollout. Disable that check for Gemma4 only; image features are still produced (image processing processor() is unconditional and runs after validate_inputs).
- verl/experimental/agent_loop/agent_loop.py (_compute_position_ids): fall back to mask-based 1D positions when the processor has no get_rope_index. Gemma encodes image structure in the attention mask (mm_token_type_ids) and the vision tower, not in M-RoPE positions. Qwen has get_rope_index, so its branch is unchanged.
- verl/utils/model.py (extract_multi_modal_inputs): add mm_token_type_ids to _VARLEN_MULTI_MODAL_KEYS so the per-sample, variable-length token-type mask is pad-batched instead of torch.cat (which fails on differing sequence lengths). Qwen consumes mm_token_type_ids into M-RoPE positions earlier, so it never reaches this path.

Fixes target the granular functions (not the _agent_loop_postprocess orchestrator), so they also cover the new AgentLoopWorkerTQ from #6710, which overrides the orchestrator but calls these inherited functions.

### Checklist

- [x] pre-commit run passes (ruff/mypy/etc.). (Local check-naming-conventions flags a "veRL" substring inside .venv/.../ray — a dependency, not this change; clean in CI.)
- [x] Tests: new multimodal model support requires GPU + a gated checkpoint, so not added to CI; validated by the experiment above. <explain / link CI if you add a small unit test>
- [x] Not a recipe submodule change.
- [x] Minimal changes 0 agent slope ;)

### DIFF
--- a/verl/utils/tokenizer.py
+++ b/verl/utils/tokenizer.py
@@ -224,6 +224,10 @@ def hf_processor(name_or_path, **kwargs):
                 model_class = Glm4vModel
             case "MllamaProcessor":
                 pass  # MllamaProcessor and MllamaModel doesn't have get_rope_index property
+            case "Gemma4Processor":
+                # Gemma4 uses standard 1D RoPE -> no get_rope_index to bind. Disable its strict
+                # per-image-token check (which Qwen's processor lacks).
+                processor.validate_inputs = lambda *args, **kwargs: None
             case _:
                 raise ValueError(f"Unsupported processor type: {processor.__class__.__name__}")
 
--- a/verl/experimental/agent_loop/agent_loop.py
+++ b/verl/experimental/agent_loop/agent_loop.py
@@ -804,7 +804,8 @@ class AgentLoopWorker:
         mm_processor_kwargs: Optional[dict[str, Any]] = None,
     ) -> torch.Tensor:
         """Compute position ids for multi-modal inputs."""
-        if self.processor is None:
+        # text-only OR non-M-RoPE multimodal (e.g. Gemma4) -> standard 1D positions
+        if self.processor is None or not hasattr(self.processor, "get_rope_index"):
             return compute_position_id_with_mask(attention_mask)  # (1, seq_len)
 
         multi_modal_kwargs = {
--- a/verl/utils/model.py
+++ b/verl/utils/model.py
@@ -56,7 +56,7 @@ from verl.utils.transformers_compat import get_auto_model_for_vision2seq
 
 AutoModelForVision2Seq = get_auto_model_for_vision2seq()
 
-_VARLEN_MULTI_MODAL_KEYS = {"input_features", "feature_attention_mask"}
+_VARLEN_MULTI_MODAL_KEYS = {"input_features", "feature_attention_mask", "mm_token_type_ids"}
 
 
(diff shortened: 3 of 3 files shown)


verl/utils/tokenizer.py
```diff
@@ -224,6 +224,10 @@ def hf_processor(name_or_path, **kwargs):
                 model_class = Glm4vModel
             case "MllamaProcessor":
                 pass  # MllamaProcessor and MllamaModel doesn't have get_rope_index property
+            case "Gemma4Processor":
+                # Gemma4 uses standard 1D RoPE -> no get_rope_index to bind. Disable its strict
+                # per-image-token check (which Qwen's processor lacks).
+                processor.validate_inputs = lambda *args, **kwargs: None
```
Luosuu, line 230: do we have any risk to disable this?

Luosuu (approved):

verl/utils/tokenizer.py
dbuos: no afaik

PR #6735, opened 2026-06-15
[training_utils] fix: cap micro-batch tokens at max_token_len

Li-bf:
### What does this PR do?

`rearrange_micro_batches` uses Karmarkar-Karp, which balances the squared-workload
sum rather than raw token counts. As a result a single micro-batch's total token
count can exceed `max_token_len` even when the average is within the limit, which can
cause OOM.

This PR re-partitions with one more micro-batch until every micro-batch's token sum
fits within `max_token_len`, or until we hit the physical bound of `num_groups`
(one sample/group per micro-batch). When the cap is still unattainable (e.g. a single
sample longer than `max_token_len`), it logs a warning once instead of failing. The
re-partition decision is kept collective (`all_reduce` MAX/MIN on boolean flags) so
all DP ranks issue the same number of all_reduce calls, avoiding deadlock when
`same_micro_num_in_dp` is set.

### Checklist Before Starting

- [x] Search for similar PRs:
  - https://github.com/verl-project/verl/pulls?q=is%3Apr+seqlen_balancing (no results)

### Test

Added two CPU unit tests in `tests/utils/test_seqlen_balancing.py`:
- `test_micro_batch_respects_max_token_len`: every micro-batch stays within
  `max_token_len` unless physically bounded by the number of samples.
- `test_micro_batch_physically_bounded`: with samples longer than the cap, the
  function does not raise and yields one micro-batch per sample.

Also validated end-to-end: a training script that previously OOM'd (due to a
micro-batch exceeding `max_token_len` after KK partitioning) now completes the full
run with this fix.

### API and Usage Example

No API change. `rearrange_micro_batches` keeps the same signature; it now additionally
keeps each micro-batch within `max_token_len` whenever the number of samples/groups
physically allows it.

### Design & Code Changes

- Precompute per-sample token counts (`seqlen_per_sample`) to check the real token sum
  (not the workload) against `max_token_len`.
- Factor out `_apply_constraints`, `_make_partitions`, `_max_partition_tokens`.
- Loop: increase `num_micro_batches` by one and re-partition while the largest
  micro-batch exceeds `max_token_len` and we are below `num_groups`.
- Keep the loop decision collective via `_any`/`_all` (`all_reduce` MAX/MIN) so DP
  ranks stay in lockstep under `same_micro_num_in_dp`.
- Warn once when the cap is physically unattainable.

- [x] Read the CONTRIBUTING.md.
- [x] Apply ruff (pinned v0.12.2) lint + format — all checks pass.
- [ ] Add / Update docs — n/a (internal utility, no doc surface).
- [x] Add unit tests covering the change (CPU, in `tests/utils/`).
- [ ] Send a message in the verl Slack/Lark channel once ready for CI.
- [ ] recipe submodule — n/a.

---
AI assistance was used for this change. I have reviewed every changed line and can
defend the change end-to-end.

### DIFF
--- a/tests/utils/test_seqlen_balancing.py
+++ b/tests/utils/test_seqlen_balancing.py
@@ -47,6 +47,58 @@ def test_seqlen_balancing():
     torch.testing.assert_close(new_batch, dataproto.batch)
 
 
+def _max_micro_batch_tokens(micro_batches):
+    return max(int(mb["attention_mask"].sum().item()) for mb in micro_batches)
+
+
+def test_micro_batch_respects_max_token_len():
+    # Random batches: every micro-batch must stay within max_token_len unless we are
+    # physically bounded by the number of samples (then num_micro_batches == batch_size).
+    torch.manual_seed(0)
+    for _ in range(20):
+        batch_size = 16
+        input_ids = torch.randint(low=0, high=10, size=(batch_size, 128))
+        attention_mask = create_random_mask(
+            input_ids=input_ids,
+            max_ratio_of_left_padding=0.1,
+            max_ratio_of_valid_token=0.9,
+            min_ratio_of_valid_token=0.5,
+        )
+        data = {"input_ids": input_ids, "attention_mask": attention_mask}
+        dataproto = DataProto.from_single_dict(data)
+        seqlens = attention_mask.sum(dim=1)
+        max_seq_len = int(seqlens.max().item())
+        # pick a cap above the longest single sequence but below the total
+        max_token_len = max(max_seq_len, int(seqlens.sum().item()) // 4)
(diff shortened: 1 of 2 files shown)


YunfanZhang42:
Would appreciate if this gets merged ASAP. Got hit with this problem today.

Li-bf:
Hi @vermouth1992, just following up on the review request when you have a chance. Please let me know if there’s anything I can clarify or update to help with the review. Thanks!

Li-bf:
Hi @wuxibin89, gentle ping on this PR. Would it be possible to assign another reviewer to help move it forward?

It’s ready from my side, and I’m happy to address any feedback. Thanks!

aoshen02:
Hi, I can take a look, did you consider the case of CP packed batch?

PR #5514, opened 2026-03-07
[recipe, algo] feat: atropos RL environment integration

daspartho:
### What does this PR do?

integrates [atropos](https://github.com/NousResearch/atropos) RL environments with verl's GRPO training pipeline. envs handle generation and scoring; this recipe polls scored batches and trains with verl's FSDP actor + internal vLLM.

closes #1782

recipe submodule PR: https://github.com/verl-project/verl-recipe/pull/61 (merged)
upstream atropos fixes: https://github.com/NousResearch/atropos/pull/408 (merged)

### Results

**30-step training run:** Qwen3-4B, `math_server_zero`, 4xA100-SXM4-80GB

W&B: [trainer run](https://wandb.ai/daspartho/verl-atropos/runs/pzezvwnk) (28/30 steps logged) · [env run](https://wandb.ai/daspartho/verl-atropos/runs/kxsphr9h) (5 eval points)

training accuracy 5.8%→83%, held-out eval reaching 65% overall across 5 benchmarks (MATH-500, AMC23, Olympiad, AIME24, Minerva) in 25 steps.

W B Chart 3_8_2026, 4_19_58 AM
W B Chart 3_8_2026, 4_20_09 AM
W B Chart 3_8_2026, 4_24_16 AM

per-benchmark eval breakdown (MATH-500, AMC23, Olympiad, AIME24, Minerva) available in the [env W&B run](https://wandb.ai/daspartho/verl-atropos/runs/kxsphr9h).

**reproduce:**
```bash
git clone https://github.com/verl-project/verl.git && cd verl
git submodule update --init
pip install -e ".[vllm,atropos]"
git clone https://github.com/NousResearch/atropos.git ../atropos
pip install -e ../atropos

N_GPUS=4 MODEL=Qwen/Qwen3-4B ATROPOS_ENV=math_server_zero TOTAL_STEPS=30 \
    BATCH_SIZE=32 GROUP_SIZE=8 MICRO_BATCH_SIZE=4 GPU_MEM_UTIL=0.50 \
    PARAM_OFFLOAD=false OPTIMIZER_OFFLOAD=false ENTROPY_COEFF=0 GRAD_CLIP=0.5 \
    MAX_PROMPT_LENGTH=512 MAX_RESPONSE_LENGTH=4096 SAVE_FREQ=999 STEPS_PER_EVAL=5 \
    bash recipe/atropos/run_atropos.sh
```

hydra overrides added in `run_atropos.sh` for this run:
```
[...]

PR #6738, opened 2026-06-15
[rollout] fix: skip redundant clone in get_named_tensor_buckets to avoid OOM during SGLang weight sync

### LINKED ISSUE #6733: [Memory] `get_named_tensor_buckets` unconditionally `clone()`s every tensor during SGLang weight sync, doubling peak GPU memory and causing OOM
### Summary

During actor→rollout weight synchronization with the **SGLang** rollout backend,
`get_named_tensor_buckets` in
[`verl/workers/rollout/sglang_rollout/utils.py`](https://github.com/verl-project/verl/blob/main/verl/workers/rollout/sglang_rollout/utils.py)
calls `tensor.clone()` **unconditionally** on every named tensor before putting it into a bucket:

```python
async def get_named_tensor_buckets(iterable, bucket_bytes):
    ...
    async for name, tensor in ensure_async_iterator(iterable):
        tensor_size = tensor.element_size() * tensor.numel()
        if current_size + tensor_size > bucket_bytes:
            if current_bucket:
                yield current_bucket
            current_bucket = [(name, tensor.clone())]   # <-- always clones
            current_size = tensor_size
        else:
            current_bucket.append((name, tensor.clone()))  # <-- always clones
            current_size += tensor_size
    ...
```

This function is used in `SGLangRollout.update_weights` to chunk the weights generator before
handing them to `sgl_update_weights`.

### Problem

`tensor.clone()` always allocates a brand-new full-size buffer and copies the whole tensor in a
single op. During the actor→rollout weight sync the GPU is already almost full:

- the actor (training) parameters are resident, **and**
- the rollout engine (SGLang) weights have already been (re)allocated.

For large models (e.g. large MoE with multi-GiB fused weights such as a
`[num_experts, ...]` `gate_up_proj`/`qkv` tensor), cloning a single multi-GiB tensor transiently
**doubles** its footprint (original + clone alive at the same time). With only a few GiB free on
the device, this single-op full-size allocation triggers a **CUDA OOM**.

The crucial observation is that the `clone()` is only actually needed when the incoming tensor is a
**view into a larger backing buffer** — cloning compacts it so the large backing storage can be
released and so we don't serialize/ship the whole buffer. When the tensor **already owns a
contiguous, tightly-packed storage**, the clone is pure waste: it just doubles peak memory for no
benefit.

### Proposed fix

Replace the unconditional `tensor.clone()` with a helper that:

1. **Skips the clone entirely** when the tensor already owns a contiguous, tightly-packed storage
   (storage `nbytes` == logical `nbytes`). No extra allocation, no transient peak.
2. Otherwise (a view into a larger buffer, e.g. a `qkv` / `gate_up` split), **compacts it in row
   chunks** so we never allocate+copy the whole thing in one large op, and the large backing
   storage can be freed.

```python
def _storage_nbytes(tensor: torch.Tensor) -> int:
    try:
        return tensor.untyped_storage().nbytes()
    except AttributeError:  # older torch
        return tensor.storage().nbytes()


def _materialize_bucket_tensor(tensor: torch.Tensor) -> torch.Tensor:
    own_bytes = tensor.element_size() * tensor.numel()
    # Already an independent, compact buffer -> no clone needed.
    if tensor.is_contiguous() and _storage_nbytes(tensor) == own_bytes:
        return tensor

    out = torch.empty_like(tensor, memory_format=torch.contiguous_format)
    if tensor.dim() == 0 or tensor.shape[0] <= 1:
        out.copy_(tensor)
        return out

    # Copy in row chunks to bound the size of each individual copy op.
    chunk_bytes = 256 * 1024 * 1024  # 256 MiB
    row_bytes = max(1, own_bytes // tensor.shape[0])
    rows_per_chunk = max(1, chunk_bytes // row_bytes)
    for start in range(0, tensor.shape[0], rows_per_chunk):
        end = min(start + rows_per_chunk, tensor.shape[0])
        out[start:end].copy_(tensor[start:end])
    return out
```

and use it in `get_named_tensor_buckets`:

```python
current_bucket = [(name, _materialize_bucket_tensor(tensor))]
# ...
current_bucket.append((name, _materialize_bucket_tensor(tensor)))
```

This keeps the original semantics (each bucketed tensor is an independent buffer that does not pin a
larger backing storage) while removing the gratuitous full-size copy that caused the OOM. In our
runs (Qwen3.5 large MoE, Megatron training backend + SGLang rollout) this eliminated the CUDA OOM
during weight sync.

### Environment

- verl: `verl-0.7` line (issue also present on current `main`).
- Training backend: Megatron-LM; Rollout backend: SGLang.
- Model: large MoE (Qwen3.5).

Happy to open a PR with this change if it looks reasonable.

HaozheZhang6:
### What does this PR do?

`get_named_tensor_buckets` (`verl/workers/rollout/sglang_rollout/utils.py`) unconditionally `clone()`d every tensor before bucketing it for the SGLang actor→rollout weight sync. The clone is only needed to **compact a tensor that is a view into a larger backing buffer** (so the buffer can be released and we don't ship it whole). A tensor that already owns tight, contiguous storage — the `DTensor.full_tensor()` all-gather case — is cloned for nothing, which transiently doubles its footprint and OOMs on multi-GiB fused MoE weights (`gate_up_proj`/`qkv`) while the actor params and the rollout weights are both resident.

This clones only when the tensor is non-contiguous or backed by larger storage:

```python
if tensor.is_contiguous() and tensor.untyped_storage().nbytes() == tensor.numel() * tensor.element_size():
    return tensor          # already owns tight storage; cloning would only double peak memory
return tensor.clone()
```

Fixes #6733.

Checked the #6733 timeline and searched open PRs touching `get_named_tensor_buckets` / the sglang bucket clone — none in flight (#5599 and #5514 match on keywords but are unrelated Megatron LoRA/atropos work).

### Safety analysis (this *removes* a clone, so worth spelling out)

The two weight sources feeding this function:

- **FSDP** (`get_per_tensor_param` → `DTensor.full_tensor()`): each param is a fresh all-gather allocation that owns tight contiguous storage and is kept alive by the bucket. Skipping the clone is safe and is exactly the OOM case reported.
- **Megatron** (`per_tensor_generator`): the large fused weights (MoE experts, qkv/gate_up) are **slices of a `full_tensor`** → `storage > footprint` → still cloned by the predicate (no behavior change). The remaining tight tensors come from per-iteration-fresh `torch.empty_like` all-gather buffers, or from the live (frozen-during-sync) model param on the src PP rank — neither is reused before the bucket is `await`ed, so skipping the clone is safe.

I don't have a multi-GPU Megatron cluster to run end-to-end, so a maintainer sanity-check that no rollout/checkpoint path feeds a *reused, tight-storage* buffer into this function would be appreciated.

### Test

Added to `tests/workers/rollout/test_sglang_rollout_sharding_manager.py`:
- `test_compact_for_bucket_skips_clone_for_owned_contiguous_tensor` — tight tensor returned as-is (same `data_ptr`, no extra allocation).
- `test_compact_for_bucket_clones_view_into_larger_buffer` — a slice of a larger tensor is compacted; values preserved.
- `test_compact_for_bucket_clones_non_contiguous_tensor` — non-contiguous tensor cloned into its own storage; values preserved.
- `test_get_named_tensor_buckets_preserves_values` — bucketing still preserves all names/values.

Verified the logic on CPU (torch only, no GPU):

```
PASS  owned-contiguous tight tensor -> returned as-is (no clone, no extra alloc)
PASS  contiguous view into larger buffer -> cloned+compacted, values equal
PASS  non-contiguous tensor -> cloned, values equal
PASS  bucketing preserves all names and values
```

The GPU OOM repro itself needs the full actor+SGLang RL stack, which is best left to verl's GPU CI.

### AI assistance

This change was developed with AI assistance (Claude). I reviewed every line, did the safety analysis above, and ran the CPU tests myself.

### DIFF
--- a/tests/workers/rollout/test_sglang_rollout_sharding_manager.py
+++ b/tests/workers/rollout/test_sglang_rollout_sharding_manager.py
@@ -16,7 +16,7 @@
 import pytest
 import torch
 
-from verl.workers.rollout.sglang_rollout.utils import get_named_tensor_buckets
+from verl.workers.rollout.sglang_rollout.utils import _compact_for_bucket, get_named_tensor_buckets
 
 _TENSOR_1MB = torch.zeros(512, 512)
 _BYTES_1MB = 1 << 20
@@ -56,3 +56,45 @@ async def test_get_named_tensor_buckets(named_tensors, bucket_size_mb, gt_groups
         assert len(group) == len(gt_group)
         for (name, _), (gt_name) in zip(group, gt_group, strict=True):
             assert name == gt_name
+
+
+def test_compact_for_bucket_skips_clone_for_owned_contiguous_tensor():
+    # A freshly-allocated contiguous tensor owns tight storage (the DTensor.full_tensor() case).
+    # It must be returned as-is so bucketing does not transiently double its peak memory.
+    tensor = torch.randn(128, 64)
+    assert _compact_for_bucket(tensor) is tensor
+
+
+def test_compact_for_bucket_clones_view_into_larger_buffer():
+    # A contiguous view that spans only part of a larger backing buffer must be compacted,
+    # otherwise the whole buffer stays resident / gets shipped.
+    base = torch.randn(256, 64)
+    view = base[:128]
+    assert view.is_contiguous()
+    out = _compact_for_bucket(view)
+    assert out is not view
+    assert out.untyped_storage().nbytes() == out.numel() * out.element_size()
+    assert torch.equal(out, view)
+
+
+def test_compact_for_bucket_clones_non_contiguous_tensor():
+    tensor = torch.randn(64, 128).t()
+    assert not tensor.is_contiguous()
+    out = _compact_for_bucket(tensor)
+    assert out is not tensor
+    assert out.data_ptr() != tensor.data_ptr()  # cloned into its own fresh storage
+    assert torch.equal(out, tensor)
+
+
+@pytest.mark.asyncio
+async def test_get_named_tensor_buckets_preserves_values():
+    # The conditional clone must not change the data that ends up in the buckets.
+    named_tensors = [("a", torch.randn(64, 64)), ("b", torch.randn(64, 64)), ("c", torch.randn(64, 64))]
+    expected = {name: tensor.clone() for name, tensor in named_tensors}
+    flat = {}
+    async for group in get_named_tensor_buckets(iter(named_tensors), 0.5 * _BYTES_1MB):
+        for name, tensor in group:
+            flat[name] = tensor
+    assert set(flat) == set(expected)
+    for name, tensor in expected.items():
+        assert torch.equal(flat[name], tensor)
--- a/verl/workers/rollout/sglang_rollout/utils.py
+++ b/verl/workers/rollout/sglang_rollout/utils.py
@@ -71,6 +71,24 @@ def broadcast_pyobj(
         return data
 
 
+def _compact_for_bucket(tensor: torch.Tensor) -> torch.Tensor:
+    """Return a tensor safe to retain in a weight-sync bucket without pinning extra memory.
+
+    ``get_named_tensor_buckets`` keeps every tensor alive until its bucket is flushed. A tensor
+    that is a *view* into a larger backing buffer would therefore keep that whole buffer resident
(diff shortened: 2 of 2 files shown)


wuxibin89 (approved):

HaozheZhang6:
signed

PR #6779, opened 2026-06-16
[rollout] feat: add Continuous Token for Agentic Rollout

gxlvera:
### What does this PR do?

RFC https://github.com/verl-project/verl/issues/6719

This PR introduces a general Continuous Token mechanism for multi-turn agentic rollout. While Verl is already token-in-token-out, multi-turn agent rollout requires more than preserving the initial prompt and generated token ids: it must reliably maintain token continuity across assistant outputs, environment/tool/user feedback, and the next assistant generation prompt, while keeping metadata (e.g. response_mask and logprobs) correctly aligned.

Continuous Token adds a reusable builder layer to centralize multi-turn tokenization that was previously easy to get wrong or scatter across AgentLoop/AgentGateway (in Verl UniAgent repo) code. Also, it introduces adapters for different model families and allow model-specific behavior, where chat templates may differ in role boundaries, stop tokens, tool response rendering, and etc. The mechanism is configurable, reusable, and verifiable, making multi-turn token construction more accurate and robust.

Continuous Token is disabled by default, so existing rollout behavior remains unchanged unless the feature is explicitly enabled.

### Design & Code Changes
- Add Continuous Token core builders in `verl/utils/continuous_token.py`
  - `ContinuousTokenBuilder`
  - `MergeResult`
  - Qwen, MiniMax, and GLM-specific boundary handling

- Add builder registry and model-family resolution in `verl/utils/continuous_token_wiring.py`

- Add rollout config in:
  - `verl/workers/config/rollout.py`
  - `verl/trainer/config/rollout/rollout.yaml`

- Integrate Continuous Token into existing AgentLoop paths:
  - `verl/experimental/agent_loop/agent_loop.py`
  - `verl/experimental/agent_loop/single_turn_agent_loop.py`
  - `verl/experimental/agent_loop/tool_agent_loop.py`

- Add model-specific tool parsers in `verl/experimental/agent_loop/tool_parser.py`

- Add standalone chat-template validation tool:
  - `scripts/chat_template_checker.py`

- Add a script that compares the output of continuous token and legacy agentloop and documented results:
  - `tests/experimental/agent_loop/continuous_token/compare_agentloop_ct_vs_legacy.py`
  - `tests/experimental/agent_loop/continuous_token/mock_trajectories.py`
  - `tests/experimental/agent_loop/continuous_token/README.md`

### Test
An e2e example with continuous token enabled has the following results.
The setting is gsmk8k + ToolAgentLoop + SGLang + Qwen2.5 3B + grpo + A100*8
The recipe is [here](https://github.com/gxlvera/verl/tree/gxl-ct-dev-e2e/examples/sglang_multiturn)
|  |  |  |
|---|---|---|
| eval reward | eval accuracy | mean num of turns |
| **Eval Reward** | **Eval Accuracy** | **Mean Num of Turns** |

This PR adds/updates the following CI coverage:

- **Continuous Token CPU unit tests**
  - `tests/utils/test_continuous_token_on_cpu.py`
  - Covers CT token building/merging behavior and response metadata alignment.
  - Collected by the existing `cpu_unit_tests.yml` workflow because it matches `*_on_cpu.py`.

- **Tool call id CPU unit tests**
  - `tests/experimental/agent_loop/test_tool_call_id_on_cpu.py`
  - Covers `tool_call_id` preservation from tool parser output to assistant `tool_calls[].id`, and then to tool response messages.
  - Also collected by the existing `cpu_unit_tests.yml` workflow.

- **Agent loop CT smoke tests with vLLM and SGLang**
  - Added to `.github/workflows/vllm.yml` and `.github/workflows/sgl.yml`.
  - Runs the same two agent-loop tests under both rollout backends:
    - `tests/experimental/agent_loop/test_basic_agent_loop.py::test_single_turn`
    - `tests/experimental/agent_loop/test_basic_agent_loop.py::test_tool_agent`
  - Uses:
    - `ENABLE_CONTINUOUS_TOKEN=1 ROLLOUT_NAME=vllm`
    - `ENABLE_CONTINUOUS_TOKEN=1 ROLLOUT_NAME=sglang`
  - Covers CT with both `single_turn_agent` and `tool_agent` under vLLM async rollout and SGLang async rollout.

No training E2E workflow is added. CT is primarily a rollout/agent-loop tokenization and metadata-alignment change, so the CI coverage focuses on CPU unit tests plus vLLM/SGLang agent-loop E2E smoke tests.

A [script](https://github.com/gxlvera/verl/blob/gxl-ct-dev-comparison/tests/experimental/agent_loop/continuous_token/compare_agentloop_ct_vs_legacy.py) that tests the difference in the output (including response mask and logprob) between agentloop with CT enabled and disabled (legacy path) is as below. A full comparison result is in this [doc](https://github.com/gxlvera/verl/blob/gxl-ct-dev-comparison/tests/experimental/agent_loop/continuous_token/ct_vs_legacy_comparison_result.md)
```text
Family        Models  Runs  Pass  Mismatch  Error  Notes
Qwen          9       27    9     18        0      Single-turn passes; tool cases mismatch because legacy misses the newline after assistant EOS. CT is correct.
GLM           4       12    4     8         0      Single-turn CT-vs-legacy passes; tool cases mismatch on the observation boundary. CT is correct.
Kimi          3       9     9     0         0      All default trajectories pass with tool_parser=kimi. 
Seed-OSS      1       3     3     0         0      All default trajectories pass with tool_parser=seed. 
MiniMax       3       9     3     0         6      Single-turn passes; tool cases are legacy-only TemplateError. 
MiMo          2       6     6     0         0      All default trajectories pass with Hermes parser. 
Nemotron      3       9     3     6         0      Single-turn passes; tool cases mismatch because legacy misses the tool-response user header. CT is correct.
```

### API and Usage Example

```python

```
> [!IMPORTANT]

```

### CODE REVIEW

File: verl/trainer/config/rollout/rollout.yaml [hunk 1]
```diff
@@ -214,6 +214,20 @@ multi_turn:
...
 
+  # Continuous Token keeps the runtime token stream across multi-turn rollout instead of
+  # reconstructing previous assistant turns from text. Disabled by default for compatibility.
+  continuous_token:
+
+    # Required when using verl.utils.omega_conf_to_dataclass to instantiate dataclass configs
+    _target_: verl.workers.config.ContinuousTokenConfig
+
+    # Enable Continuous Token for multi-turn agent rollout.
+    enable: False
```
wuxibin89: Should we enable it by default and drop the old implementation?

wuxibin89: For single_turn_agent_loop and tool_agent_loop, we can use Continuous Token by default. While for user custom AgentLoop, its user choice.

scripts/chat_template_mock_trajectories.py
```diff
@@ -0,0 +1,529 @@
...
+@dataclass(frozen=True)
+class TrajectoryStep:
+    """One assistant generation followed by optional environment messages."""
+
+    assistant: dict[str, Any]
+    appended_messages: tuple[dict[str, Any], ...] = ()
+
+
+@dataclass(frozen=True)
+class SingleTurnTrajectory:
```
wuxibin89, line 35: Move test utility to tests folder.

verl/experimental/agent_loop/agent_loop.py
```diff
@@ -224,8 +225,28 @@ def __init__(
...
+                model_family=continuous_token_config.model_family,
+                model_path=model_config.path,
+                tokenizer_name_or_path=model_config.tokenizer_path,
+                chat_template_kwargs=self.apply_chat_template_kwargs,
+            )
+            self.enable_continuous_token = True
+            # Continuous Token doesn't use the legacy removable system prompt.
+            self.system_prompt = None
+        else:
+            if continuous_token_config.enable and self.processor is not None:
```
wuxibin89, line 244: Continuous token hasn't support multimodal yet?

gxlvera: Multimodal is WIP

verl/trainer/config/rollout/rollout.yaml
gxlvera: Thanks for the review! I agree that  it's better to use it by default this is a great suggestion👍 So in single turn agent loop and tool agent loop, I think i could just delete the old way. My question is, in that case, do we still need the argument `multi_turn.continuous_token.enable` in the config file? The current config file needs two argument for continuous token👇
```
actor_rollout_ref:
  rollout:
    multi_turn:
      continuous_token:
        enable: true
        model_family: auto
```
if the user set `enable` to be false, but he/she chooses to use ToolAgentLoop, should we raise an exception? So I think maybe we can discard that argument （`multi_turn.continuous_token.enable`). In custom agent loop，if user doesn't want to use continuous token, he/she can simply not use the continuous token related hooks defined in AgentLoopBase.

scripts/chat_template_mock_trajectories.py
gxlvera: I placed them under scripts/ because they are fixture data for chat_template_checker.py, which is a developer/debugging tool rather than a pytest test. Keeping the data next to the checker avoids making the script depend on the test package while still keeping it out of the production verl package.

verl/utils/tokenizer/continuous_token.py
```diff
@@ -0,0 +1,644 @@
...
+
+    def build_initial_tokens(
+        self,
+        messages: list[dict[str, Any]],
+        *,
+        tools: list[dict[str, Any]] | None = None,
+    ) -> list[int]:
+        return self._render_tokens(messages, add_generation_prompt=True, tools=tools)
+
+    def tokenize_non_assistant_incremental_messages(
```
wuxibin89, line 101: P2: change this to private function.

verl/utils/tokenizer/continuous_token.py
```diff
@@ -0,0 +1,644 @@
...
+        self.tokenizer = tokenizer
+        self.chat_template_kwargs = chat_template_kwargs or {}
+        if allowed_append_roles is not None:
+            allowed_roles = frozenset(allowed_append_roles)
+            unknown_roles = allowed_roles - _SUPPORTED_APPEND_ROLES
+            if unknown_roles:
+                raise ValueError(f"Unsupported Continuous Token append roles: {sorted(unknown_roles)}")
+            self.allowed_append_roles = allowed_roles
+
+    def build_initial_tokens(
```
wuxibin89, line 93: User should only call these three public functions? Please add doc string for them.
- build_initial_tokens
- merge_non_assistant_tokens
- merge_assistant_tokens

wuxibin89: For developer, we may need to clarify which function they should override.

verl/trainer/config/rollout/rollout.yaml
```diff
@@ -214,6 +214,20 @@ multi_turn:
   # - ignore_strippable: ignore strippable tokens when checking tokenization sanity
   tokenization_sanity_check_mode: strict
 
+  # Continuous Token keeps the runtime token stream across multi-turn rollout instead of
+  # reconstructing previous assistant turns from text. Disabled by default for compatibility.
+  continuous_token:
```
wuxibin89: Beside AgentLoop, we should also apply continuous token to [mult-turn sft dataset](https://github.com/verl-project/verl/blob/main/verl/utils/dataset/multiturn_sft_dataset.py). It may be better to move `continuous_token` under `data` field since it's common to ppo and sft trainer
```bash
data.continuous_token.enable=True
data.continuous_token.model_family=qwen3
```

verl/utils/continuous_token.py
```diff
@@ -0,0 +1,661 @@
...
+        )
+
+    def _render_tokens(
+        self,
+        messages: list[dict[str, Any]],
+        *,
+        add_generation_prompt: bool,
+        tools: list[dict[str, Any]] | None = None,
+    ) -> list[int]:
+        from verl.utils.chat_template import apply_chat_template
```
wuxibin89: P2: move import to top level

wuxibin89:
Move all tokenizer related utils to `verl/utils/tokenizer`
- verl/utils/continuous_token.py
- verl/utils/continuous_token_wiring.py
- verl/utils/tokenizer.py
- verl/utils/chat_template.py

verl/trainer/config/rollout/rollout.yaml
wuxibin89: Yes, multi_turn.continuous_token.enable is not needed any more.

wuxibin89: Actually, I think we don't even need `continuous_token` config, is auto detect enough?

gxlvera: I think in this PR, let's just keep it as an optional because we haven't supported multimodal, so for multimodal we still need the legacy way to deal with TITO. Let's make continuous token as default in the multimodal support PR?

wuxibin89: OK, let drop legacy implementation once multimodal ready.

verl/trainer/config/rollout/rollout.yaml
gxlvera: ok, i have moved it to data field

verl/utils/tokenizer/continuous_token.py
gxlvera: done

verl/utils/continuous_token.py
gxlvera: done

gxlvera:
> Move all tokenizer related utils to `verl/utils/tokenizer`
> 
> * verl/utils/continuous_token.py
> * verl/utils/continuous_token_wiring.py
> * verl/utils/tokenizer.py
> * verl/utils/chat_template.py

done

wuxibin89 (approved):

PR #6794, opened 2026-06-18
[checkpoint_engine][sglang] feat: delta weight sync for disaggregated rollout

ChangyiYang:
## Summary

Adds **delta weight sync** for the SGLang rollout and puts it on the
**disaggregated trainer→rollout wire**: instead of broadcasting every parameter
each sync, ship only the `(position, value)` pairs that changed since the last
broadcast. In RL post-training >99% of BF16 weight bytes are unchanged
step-over-step, so this cuts the trainer→rollout transfer by ~100–1000×.

Design follows THUDM/slime's NCCL delta transport and reuses verl's existing
checkpoint-engine topology (actor rank 0 → rollout workers over
`ray.util.collective`).

## What's in this PR

**Engine-agnostic core** — `verl/workers/rollout/delta_sync/`
- `delta_state.py` — pinned-CPU snapshot + bytewise diff, d2h/h2d side streams (CUDA-optional)
- `encode.py` — `indices` / `deltas` (uint16 gap) encodings, per-param manifest, checksum, decoder
- `wrapper.py` — `iter_delta_flushes()`: `(name, tensor)` generator → stream of bucketed `DeltaFlush`

**Disaggregated NCCL transport** — `verl/checkpoint_engine/delta_checkpoint_engine.py` (`backend="delta"`)
- Subclasses `NCCLCheckpointEngine`, so it reuses the same collective group, zmq side-channel
  and `build_topology` as the full-weight engine.
- `send_weights` (trainer rank 0): byte-diff vs snapshot → publish a per-sync manifest over zmq, then
  `collective.broadcast` only the changed positions/values per flush (master uses cupy buffers so NCCL
  registration is safe under `expandable_segments`). The first sync forces a full delta so a
  dummy-initialized rollout gets a correct base; subsequent syncs are sparse.
- `receive_weights` (rollout worker): broadcast-recv the flush, verify the per-flush checksum, decode +
  NaN-mask into a local full-weight mirror, and yield reconstructed full tensors to the **standard**
  `server_adapter.update_weights`. The trainer→worker wire is sparse; the worker→engine push is an
  ordinary full-tensor load — **no SGLang-side delta receiver is required**.

Default behavior is unchanged: `delta` is opt-in via
`rollout.checkpoint_engine.backend=delta` (+ `engine_kwargs.delta.encoding`).

Also includes a small one_step_off fix: skip the colocated resume/sleep weight-sync path on detached
actor workers (which have no local rollout handle) so the disaggregated full-sync baseline runs too.

## Validation

Single-node 4+4 disaggregated (one_step_off, FSDP2 + SGLang, Qwen2.5-0.5B, GSM8K GRPO):
- runs end-to-end with `backend=delta`; per-flush checksum verified on receipt, no corruption
  (`ppo_kl` small and stable, rollout generates coherently).
- unit tests cover encode/decode bit-identity for both encodings.

## Notes / scope

- `deltas_zstd` encoding is not yet wired (gap stream only); vLLM/TRTLLM rollouts out of scope.


gxlvera:
An e2e example is WIP

gxlvera:
Instead of computing diff and pin full weight on rank0's CPU, the pr author @ChangyiYang and I are implementing sharded diff, which could save nccl gather overhead and saves CPU memory.

Issue #6478, opened 2026-05-26
[Bug] Multiple RL algorithm inconsistencies: missing `negative_approx_kl` clamp in `clip_cov`/`kl_cov`/`geo_mean`, GPG `alpha` silently overridden, GRPO single-sample std mismatch


imitater-dou:
## System Info

Discovered via static code analysis of `verl/trainer/ppo/core_algos.py`. Three distinct but related algorithm correctness issues.

## Information

- [x] The official example scripts

## Tasks

- [x] An officially supported task in the `examples` folder (such as GLUE/SQuAD, ...)

### Issue A — Missing `negative_approx_kl` clamp (NaN risk)

In `compute_policy_loss_clip_cov` (line ~1789), `compute_policy_loss_kl_cov` (line ~1881), and `compute_policy_loss_geo_mean` (line ~1963), the code computes `negative_approx_kl = log_prob - old_log_prob` and immediately applies `torch.exp(negative_approx_kl)` **without** the clamp that all 8 other policy loss functions apply:

```python
# Missing in clip_cov / kl_cov / geo_mean:
negative_approx_kl = torch.clamp(negative_approx_kl, min=-20.0, max=20.0)
```

In `geo_mean`, this clamp is present but commented out with "uncomment if you like" — the intent is ambiguous. Without it, large policy deviations early in training can produce `exp(large_value)` → NaN/Inf, corrupting the entire batch gradient.

### Issue B — GPG `alpha` parameter silently overridden

In `compute_gpg_outcome_advantage` (line ~809):

```python
def compute_gpg_outcome_advantage(scores, ..., alpha: float = 1.0):
    ...
    alpha = bsz / m.clamp(min=1)   # ← immediately overwrites the parameter
```

The caller-supplied `alpha` is always discarded. This breaks the function's contract: users who set `alpha` via config will silently get wrong behavior.

Additionally, `id2std` is computed (lines ~804, ~817, ~821) but never used in the normalization formula at line ~825.

### Issue C — GRPO single-sample group std inconsistency

In `compute_grpo_outcome_advantage` (line ~315–317):

```python
if len(id2score[idx]) == 1:
    id2std[idx] = torch.tensor(1.0)  # single-sample: std = 1 → advantage ≈ raw score
```

But in `compute_grpo_outcome_advantage_vectorized`, single-element groups produce `std = 0 → advantage = 0`. The two code paths have different behavior for the same input, making switching between them non-transparent.

## Expected behavior

[...]

PR #6803, opened 2026-06-21
[algo] fix: remove dead code in GPG advantage estimator

EazyReal:
## What

Removes dead code in `compute_gpg_outcome_advantage` (`verl/trainer/ppo/core_algos.py`), addressing **Item B** of #6478, plus an adjacent device-mismatch fix.

1. **`id2std` is never read.** The function builds an `id2std` dict and populates it per group, but normalization divides centered scores by `f_norm`, not by the group std. Pure dead code.

2. **The `alpha` parameter is a no-op.** `alpha` (default `1.0`) is unconditionally overwritten by `alpha = bsz / m.clamp(min=1)` before its first use, so any caller-supplied value is silently discarded. No caller passes it (`ray_trainer.py` builds `adv_kwargs` without `alpha`, and the function keeps `**kwargs`), so removing the parameter is non-breaking.

3. **Singleton-group device mismatch.** The singleton branch set `id2mean[idx] = torch.tensor(0.0)` (always CPU), while the multi-response branch uses `torch.mean(...)` on the scores' device. On GPU this mismatches in the advantage loop. Use a plain `0.0` (device-agnostic; identical on CPU).

## Change

- Delete the `id2std` dict and the two lines that populate it.
- Remove the dead `alpha` parameter (and its docstring entry).
- Set the singleton mean to `0.0` instead of a CPU tensor.

The live local `alpha = bsz / count_nonzero(scores)` and its use in the scaling line are **kept** — that is the actual GPG scaling. Behavior is unchanged on CPU.

## Scope

Only **Item B** of #6478. **Item A** (the negative-approx-kl clamp) is handled separately in #6538 and is not touched here.

## Test

Adds `tests/trainer/ppo/test_gpg_advantage_on_cpu.py` (CPU-only, a new file to avoid conflicts with #6538/#4677 which also edit `test_core_algos_on_cpu.py`):

- `test_gpg_singleton_group_returns_raw_score` — a single non-zero-scored response gets advantage == raw masked score.
- `test_gpg_applies_n_over_nonzero_scaling` — a 2-response group (scores 4, 0) yields `alpha * (score - mean) = 2 * (score - 2)`.

### DIFF
--- a/tests/trainer/ppo/test_gpg_advantage_on_cpu.py
+++ b/tests/trainer/ppo/test_gpg_advantage_on_cpu.py
@@ -1,0 +1,49 @@
+# Copyright 2025 Bytedance Ltd. and/or its affiliates
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import numpy as np
+import torch
+
+from verl.trainer.ppo.core_algos import compute_gpg_outcome_advantage
+
+
+def test_gpg_singleton_group_returns_raw_score():
+    token_level_rewards = torch.tensor([[0.0, 3.0, 0.0]], dtype=torch.float32)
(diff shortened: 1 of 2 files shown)


EazyReal:
Good catch — fixed. The singleton-group mean is now a plain `0.0` instead of `torch.tensor(0.0)`, so it no longer mismatches GPU scores in the advantage loop (identical on CPU). Folded it into this cleanup since it's the same function. (The same idiom appears in a few sibling estimators; happy to do those in a separate focused PR.)

EazyReal:
@eric-haibin-lin @vermouth1992 @tongyx361 @PeterSH6 Could I get a quick review on this GPG cleanup? It removes dead advantage-estimator code without changing behavior.

PR #6823, opened 2026-06-23
[BREAKING][trainer, cfg] chore: enable V1 trainer by default

Begunner:
### What does this PR do?

The V1 PPO trainer (`verl/trainer/ppo/v1`) is used by default.

### DIFF
--- a/verl/workers/engine_workers.py
+++ b/verl/workers/engine_workers.py
@@ -337,6 +337,7 @@ class TrainingWorker(Worker, DistProfilerExtension):
             max_token_len_per_gpu=self.engine_config.max_token_len_per_gpu,
             micro_batch_size_per_gpu=self.engine_config.micro_batch_size_per_gpu,
             use_fused_kernels=self.engine_config.use_fused_kernels,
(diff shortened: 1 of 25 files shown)


verl/workers/engine_workers.py
```diff
@@ -337,6 +337,7 @@ def train_batch(self, data: TensorDict) -> TensorDict:
             max_token_len_per_gpu=self.engine_config.max_token_len_per_gpu,
             micro_batch_size_per_gpu=self.engine_config.micro_batch_size_per_gpu,
             use_fused_kernels=self.engine_config.use_fused_kernels,
+            temperature=1.0,
```
wuxibin89: temperature has been explicitly set in trainer, shouldn't be set here again. https://github.com/verl-project/verl/blob/main/verl/trainer/ppo/v1/trainer_base.py#L417

wuxibin89:
Follow up PR: 
- replace fully-async ci with `separate_async`
- replace one-step-off ci with `colocate_async`

wuxibin89 (approved):

Issue #6827, opened 2026-06-23
[Feature] On-Policy Self-Distillation (OPSD): privileged-context self-teacher


HaozheZhang6:
### What

OPSD ([Self-Distilled Reasoner, arXiv:2601.18734](https://arxiv.org/abs/2601.18734)) trains a single model as both teacher and student: the **teacher** conditions on **privileged information** (a verified reasoning trace / ground-truth solution), the **student** sees only the question, and we minimize the per-token KL between them over the student's own on-policy rollouts. Unlike OPD it needs no external teacher model; unlike SDPO (#5499) the teacher conditions on privileged ground-truth, not hindsight feedback.

### What's here vs missing

- **Loss**: the per-token top-k KL already exists (`compute_forward_kl_topk` / `kl_divergence`), so OPSD reuses it directly.
- **Missing**: (1) building the teacher's privileged context -- splice the verified trace/solution into the teacher input, run the teacher forward, slice the top-k logprobs back onto the student's positions; (2) the self-teacher -- point the teacher at the student weights instead of loading a separate model.

### Plan

1. (small, in #6828) a `kl_direction` knob on the distillation loss, so self-distillation variants can pick forward or reverse KL.
2. the privileged-context splice/slice helpers (CPU-testable token surgery).
3. the teacher forward + self-teacher weight-sync + an end-to-end convergence run (GPU).

I'd like to take this.

Refs: OPSD / Self-Distilled Reasoner arXiv:2601.18734.

HaozheZhang6:
@JacobHelwig — I'd like to build OPSD (On-Policy Self-Distillation, [Self-Distilled Reasoner](https://arxiv.org/abs/2601.18734)) on top of your OPD pipeline, and wanted to sanity-check the design with you before going further.

[...]

PR #6828, opened 2026-06-23
[trainer, cfg] feat: add kl_direction (forward/reverse) to the top-k distillation loss

HaozheZhang6:
### What does this PR do?

Adds a `kl_direction in {forward, reverse}` knob to `DistillationLossConfig`. `compute_forward_kl_topk` only computed forward KL (`KL(teacher||student)`); `reverse` swaps it to `KL(student||teacher)`. The existing reverse-KL path uses single-sample estimators (`k1`/`k3`/...), not the top-k support, so a top-k reverse direction was missing. Small loss-flexibility primitive for on-policy self-distillation (#6827).

Scope:
- FSDP backend honors `kl_direction`; the megatron path raises `NotImplementedError` for `reverse` (its hand-written autograd `Function` would need the gradient re-derived).
- Config validation rejects anything but `forward` / `reverse`.

### Test

`pytest tests/workers/test_kl_topk_direction_on_cpu.py` -- forward keeps today's semantics, reverse swaps, the two differ. CPU-only. Verified locally that flipping the swap fails the test, and the existing distillation CPU tests still pass.

Part of #6827.

### DIFF
--- a/tests/workers/test_forward_kl_topk_loss_values_on_cpu.py
+++ b/tests/workers/test_forward_kl_topk_loss_values_on_cpu.py
@@ -1,0 +1,105 @@
+# Copyright 2026 Bytedance Ltd. and/or its affiliates
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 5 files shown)


HaozheZhang6:
Closing — on a closer look verl already covers both KL directions (`forward_kl_topk` and the reverse-KL estimators in `compute_distillation_loss_reverse_kl_estimator`), so a direction selector doesn't add capability.

PR #6833, opened 2026-06-24
[rollout, trainer, cfg] feat: privileged-context teacher scoring for OPSD

HaozheZhang6:
### What does this PR do?

The first piece of On-Policy Self-Distillation (OPSD, #6827): the teacher conditions on the ground-truth solution (privileged context) while the student sees only the problem, and scores the student's own on-policy rollout. This is the half plain OPD lacks -- verl's OPD teacher sees only `prompt + response`.

- `verl/trainer/distillation/privileged_context.py` -- two pure helpers: `build_privileged_sequence` (splice the GT solution, wrapped in marker tokens, into the teacher input) and `slice_privileged_teacher_to_student` (realign the teacher's per-token response scores back onto the student's `prompt + response` positions). CPU-tested.
- `DistillationConfig` -- `self_distillation`, `privileged_solution_key`, `privileged_prefix`, `privileged_suffix`.
- `agent_loop._compute_teacher_logprobs` -- when `self_distillation` is on, build the privileged teacher sequence and slice the scores back; otherwise the path is unchanged (`prompt + response`).

### Route

Reuses the existing external-teacher path with `teacher.model_path == student model_path`, i.e. a frozen self-teacher (the student's init checkpoint) scoring the rollout with privileged context. The whole OPD data + loss pipeline is reused unchanged -- only the teacher's input sequence and the realignment are new. The dynamic / EMA self-teacher (a second in-engine forward) is a separate, GPU-gated follow-up.

### Test

`pytest tests/workers/test_opsd_privileged_context_on_cpu.py tests/workers/test_distillation_config_opsd_on_cpu.py` -- splice layout + slice-back shapes/dtype + the config fields. CPU-only. The end-to-end GPU run (teacher = student checkpoint, top-k overlap rises, convergence) is the remaining validation, hence draft.

Part of #6827.

### DIFF
--- a/verl/experimental/agent_loop/agent_loop.py
+++ b/verl/experimental/agent_loop/agent_loop.py
@@ -49,6 +49,10 @@ from verl.experimental.agent_loop.utils import resolve_config_path
 from verl.protocol import DataProto
 from verl.tools.tool_registry import load_all_tools
 from verl.trainer.distillation import is_distillation_enabled
+from verl.trainer.distillation.privileged_context import (
+    build_privileged_sequence,
+    slice_privileged_teacher_to_student,
+)
 from verl.utils.config import omega_conf_to_dataclass
 from verl.utils.dataset.rl_dataset import RLHFDataset, get_dataset_class
 from verl.utils.model import compute_position_id_with_mask
@@ -521,6 +525,17 @@ class AgentLoopWorker:
             from verl.experimental.teacher_loop.teacher_manager import AsyncTeacherLLMServerManager
 
             self.teacher_key: str = config.distillation.teacher_key
+            self.self_distillation: bool = config.distillation.self_distillation
+            self.privileged_solution_key: str = config.distillation.privileged_solution_key
+            self._privileged_prefix_ids: list[int] = []
+            self._privileged_suffix_ids: list[int] = []
+            if self.self_distillation:
+                self._privileged_prefix_ids = self.tokenizer.encode(
+                    config.distillation.privileged_prefix, add_special_tokens=False
+                )
+                self._privileged_suffix_ids = self.tokenizer.encode(
+                    config.distillation.privileged_suffix, add_special_tokens=False
+                )
             self.teacher_server_manager = AsyncTeacherLLMServerManager(
                 config=config,
                 teacher_client=teacher_client,
@@ -1014,17 +1029,35 @@ class AgentLoopWorker:
         """Compute teacher logprobs for single sample."""
         if self.distillation_enabled and not validate:
             routing_key = None
+            solution_ids = None
             if sample_kwargs is not None:
                 routing_value = sample_kwargs.get(self.teacher_key)
                 if routing_value is not None:
                     # Non-tensor batch values arrive as 0-d numpy objects / arrays; normalize to Python.
                     routing_key = routing_value.item() if hasattr(routing_value, "item") else routing_value
+                if self.self_distillation:
+                    solution = sample_kwargs.get(self.privileged_solution_key)
+                    if solution is not None:
+                        solution = solution.item() if hasattr(solution, "item") else solution
+                        solution_ids = self.tokenizer.encode(str(solution), add_special_tokens=False)
+            if solution_ids is not None:
+                # OPSD: the teacher conditions on the privileged ground-truth solution.
+                sequence_ids = build_privileged_sequence(
+                    prompt_ids, response_ids, solution_ids, self._privileged_prefix_ids, self._privileged_suffix_ids
+                )
+            else:
+                sequence_ids = prompt_ids + response_ids
             teacher_ids, teacher_logprobs = await self.teacher_server_manager.compute_teacher_logprobs_single(
-                sequence_ids=prompt_ids + response_ids,
+                sequence_ids=sequence_ids,
                 multi_modal_data=output.multi_modal_data,
                 mm_processor_kwargs=output.mm_processor_kwargs,
                 routing_key=routing_key,
             )
+            if solution_ids is not None:
+                # Realign the teacher's privileged-context scores onto the student's positions.
+                teacher_ids, teacher_logprobs = slice_privileged_teacher_to_student(
+                    teacher_ids, teacher_logprobs, len(prompt_ids), len(response_ids), self.tokenizer.pad_token_id
+                )
             output.extra_fields["teacher_ids"] = teacher_ids
             output.extra_fields["teacher_logprobs"] = teacher_logprobs
 
--- a/verl/experimental/agent_loop/agent_loop.py
+++ b/verl/experimental/agent_loop/agent_loop.py
@@ -50,6 +50,8 @@ from verl.protocol import DataProto
 from verl.tools.tool_registry import load_all_tools
 from verl.trainer.distillation import is_distillation_enabled
 from verl.trainer.distillation.privileged_context import (
+    REFERENCE_USER_TEMPLATE,
+    build_privileged_chat_turn,
     build_privileged_sequence,
     resolve_privileged_solution,
     slice_privileged_teacher_to_student,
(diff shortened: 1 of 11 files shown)


ytchx1999:
The `prompt_ids` here should be the result after `apply_chat_template`, so should they include the EOS token? When modifying the code, I added an insertion token (to insert the ground truth at the last occurrence of this token). This seems to enhance scalability. I'm not sure if my understanding is correct, and I welcome any clarification.
```python
def build_privileged_sequence(
    prompt_ids: list[int],
    response_ids: list[int],
    solution_ids: list[int],
    prefix_ids: list[int],
    suffix_ids: list[int],
    insert_before_token_ids: list[int] = None,
) -> list[int]:
    """Build the OPSD teacher's input token sequence.
    Layout: ``prompt_prefix + prefix + solution + suffix + prompt_suffix + response``.
    By default, ``insert_before_token_ids`` is None, so the solution is appended to the
    prompt: ``prompt + prefix + solution + suffix + response``.
    If ``insert_before_token_ids`` is provided, we search for the last occurrence of
    this sequence in ``prompt_ids`` and insert the solution before it.
    Args:
        prompt_ids: the student's prompt (problem) token ids.
        response_ids: the student's on-policy response token ids.
        solution_ids: the ground-truth solution token ids.
        prefix_ids: marker tokens placed before the solution.
        suffix_ids: marker tokens placed after the solution.
        insert_before_token_ids: if provided, the solution is inserted before the
            last occurrence of this sequence in ``prompt_ids``.
    Returns:
        The concatenated teacher input token ids.
    """
    if insert_before_token_ids:
        # Search for the last occurrence of the sub-sequence
        n, m = len(prompt_ids), len(insert_before_token_ids)
        split_idx = -1
        for i in range(n - m, -1, -1):
            if prompt_ids[i : i + m] == insert_before_token_ids:
                split_idx = i
                break

        if split_idx != -1:
            return (
                prompt_ids[:split_idx]
                + prefix_ids
                + solution_ids
                + suffix_ids
                + prompt_ids[split_idx:]
                + response_ids
            )

    return prompt_ids + prefix_ids + solution_ids + suffix_ids + response_ids
```

HaozheZhang6:
Good point -- appending after the full chat-templated prompt isn't ideal. I added an optional `insert_before_token_ids` to `build_privileged_sequence` (config `privileged_insert_before`): it inserts the solution before the last occurrence of a marker, and defaults to the plain append when unset.

Pushed a few related fixes while I was in there -- the default `privileged_solution_key` now resolves `reward_model.ground_truth` (the old flat default silently missed verl's nested ground truth, so OPSD quietly fell back to plain OPD), plus pad-token / max-len / empty-response guards. Thanks for the careful read.

verl/experimental/teacher_loop/teacher_manager.py
ytchx1999, line 1: Originally, the filtering was based on the length of the Student's prompt, which is why this error has a chance of occurring. Would it be better to define a separate filtering parameter in `rl_dataset`?

HaozheZhang6: Good point -- the overflow is because the privileged sequence is longer than the student prompt the dataset filtered on. I kept the teacher-side raise as a loud safety net for now, but a dataset-level filter (capping prompt+solution upfront, like `filter_overlong_prompts`) is cleaner and I think fits better as a follow-up PR. Open to either.

verl/experimental/agent_loop/agent_loop.py
```diff
@@ -1008,17 +1029,44 @@ async def _compute_teacher_logprobs(
...
+                solution = resolve_privileged_solution(sample_kwargs, self.privileged_solution_key)
+                if not solution:
+                    raise ValueError(
+                        f"self_distillation is enabled but no privileged solution resolved at "
+                        f"'{self.privileged_solution_key}'; verl's ground truth is usually at "
+                        f"'reward_model.ground_truth'."
+                    )
+                solution_ids = self.tokenizer.encode(solution, add_special_tokens=False)
+            if solution_ids is not None:
+                # OPSD: the teacher conditions on the privileged ground-truth solution.
```
JacobHelwig: Can these lines be deleted?

HaozheZhang6: Yes, done -- folded the build into the `self.self_distillation` branch and gated the slice-back on the same flag, so the `solution_ids is not None` checks are gone.

JacobHelwig (changes requested):
Nice idea to add this. Could you please show metrics from a sample training run without task rewards, along with the command used to run? The key metrics of interest would be the distillation loss, the training rewards, and the validation rewards.

HaozheZhang6:
Thanks! Will run a supervised OPSD training (no task rewards) on a small model and follow up here with the distillation loss and the training/validation rewards, plus the exact command.

HaozheZhang6:
@JacobHelwig

The privileged-context path here (`self_distillation` + `privileged_solution_key`) runs end-to-end. 2×A100, Qwen2.5-0.5B-Instruct student, GSM8K, supervised `forward_kl_topk`, `use_task_rewards=False`.

**Mechanism works as intended.** With vs without the privileged GT solution (same config): the teacher distribution diverges from the student and the distillation loss is ~13× a no-privileged control (teacher ≡ student → loss ≈ 0; the OPSD loss decays 0.0134 → 0.0032). The `teacher_manager` length guard also fires correctly when the solution overflows the teacher context. [Fig 2]

**Pipeline trains end-to-end.** Plain OPD with a larger teacher (Qwen2.5-3B → 0.5B, `self_distillation=False`) runs cleanly and GSM8K val moves up from the 0.485 baseline — a sanity check that the path trains, not a headline number (200-q subset, single config, lr differs from the OPSD run below). [Fig 1, green]

I didn't get an eval gain from OPSD self-distillation itself on this setup, which is unsurprising: a 0.5B student is probably too small to actually internalize the reference reasoning, and conditioning a same-size teacher on the full answer can collapse its distribution to an answer-leaked copy rather than a better reasoner — either way the target isn't useful. The OPSD paper (arXiv:2601.18734) only validates ≥1.7B reasoning models on AIME/HMMT; that's the regime to test an actual gain in. Posting this mainly to confirm the path runs.

Eval note: the default GSM8K strict `#### N` regex marks Qwen-Instruct's `\boxed{}` answers as 0 — used a `\boxed`/last-number numeric match for the real ~0.485 baseline.

Bug note: `compute_forward_kl_topk` hits `gather(): Expected dtype int64 for index` — `teacher_topk_ids` is int32, `torch.gather` needs int64 (`.long()` fixes it). Happy to PR separately.

---
**Fig 1 — GSM8K val accuracy (Qwen2.5-0.5B student): larger 3B teacher trends up; same-size self-teacher doesn't.**
fig1

**Fig 2 — distillation loss: privileged context (~13×) vs no-privileged control.**
fig2

ytchx1999:
I also want to add an experimental result from my business dataset. The settings for the control group are as follows:
- Models: The teacher model is Qwen-32B-VL and the student model is Qwen-8B-VL. Both are models that have undergone cold-start Supervised Fine-Tuning (SFT). For the OPSD setup, ground truth data was added during the training of the teacher model.
- Method: RKL k1 estimate with GRPO reward was used.
- Configuration: All other settings are identical to the OPD setup, except for the OPSD-related configuration (inserting the ground truth before the EOS token).
截屏2026-06-28 15 32 31

HaozheZhang6:
@ytchx1999 Right — `prompt_ids` is post-`apply_chat_template(add_generation_prompt=True)`, so it ends with the assistant generation marker (`<|im_start|>assistant\n` for Qwen), not an EOS — the `<|im_end|>` sits at the end of `response_ids`. So with the default `insert_before_token_ids=None` the solution lands inside the assistant turn (`prompt + prefix + solution + suffix + response`), which is a bit awkward; `insert_before_token_ids` is exactly there to place it earlier (e.g. before the assistant marker / the user-turn `<|im_end|>`), which is what you're doing. Your understanding is right — that's the intended use of the parameter.

Thanks for running it at real scale. The higher OPSD distillation loss vs OPD (0.171 vs 0.162) matches what the privileged context should do — conditioning the teacher on the GT moves its distribution away from the student, so there's more to distill (I saw the same on a tiny setup, ~13× a no-privileged control). Did you also log task/eval accuracy? The loss is the mechanism; the open question is whether the extra signal converts to a downstream gain, and 32B→8B is the regime where it should (the sub-1B I tried was too small to show it).

ytchx1999:
@HaozheZhang6 Here are some results from my business test set. Compared to the standard OPD, it can be observed that OPSD shows a gain in the recall metric (the data ratios in the experiment were not carefully tuned, this was just to verify the correctness of the code). Interestingly, for the same OPSD, no significant gain was found when comparing a 32B Teacher model to an 8B Teacher model.
image

HaozheZhang6:
@ytchx1999 This is great, thanks for sharing — really exciting to see. OPSD beating standard OPD on recall (+~5 pts on testsets 2/3) with the student edging past even the 32B teacher is a strong result. The w/8B ≈ w/32B comparison is the part I like most: it suggests the gain comes from the privileged context rather than teacher size, which is exactly what the self-distillation setup is meant to leverage (the same-size self-teacher being enough).

One small thing — the figure labels say "OSPD" while your text says "OPSD" — just a typo?

ytchx1999:
@HaozheZhang6 Yes, that was indeed a typo (:

HaozheZhang6:
Thanks! Sample run with task rewards off — Qwen2.5-0.5B, GSM8K, 96 steps.

Distillation loss comes down cleanly, 0.0134 → 0.0032 (grad norm 1.69 → 0.29). Setting `self_distillation=False` (teacher == student, no privileged context) keeps the loss ~0, so the signal is coming from the privileged ground-truth as intended. Task rewards are off so there's no train-reward curve.

Command, on top of `examples/on_policy_distillation_trainer/run_qwen3_8b_fsdp.sh`:
```
self_distillation=True privileged_solution_key=reward_model.ground_truth \
loss_mode=forward_kl_topk use_policy_gradient=False use_task_rewards=False \
loss_max_clamp=0.05
```

On val accuracy: 0.5B/GSM8K is below OPSD's validated regime (the paper only shows gains on Qwen3-1.7B+ with hard reasoning like AIME/HMMT), so no eval bump here and I wouldn't expect one at this scale. @ytchx1999 independently saw a recall gain at 8B earlier in the thread, which lines up.

PR #6836, opened 2026-06-24
[megatron] fix: return 3-tuple under calculate_per_token_loss to fix MoE aux/z-loss grad blowup at CP>1

### LINKED ISSUE #6609: [Megatron][MoE] grad_norm explosion under context parallel (CP>1): loss_func must return 3-tuple when calculate_per_token_loss=True
## Summary

When training a **MoE** model with the **Megatron** backend under **context parallelism (`CP > 1`)** together with `use_remove_padding=True`, `actor/grad_norm` explodes to the **thousands** and validation scores steadily degrade. With `CP=1` the exact same config is healthy (`grad_norm` ~0.3).

Root cause: verl's actor `loss_func` always returns a **2-tuple** `(loss, metrics)`. But when Megatron runs with `calculate_per_token_loss=True` (which some models, e.g. Qwen3-VL via megatron-bridge, **force** under CP), the loss function is expected to return a **3-tuple** `(loss_sum, num_tokens, metrics)`. Because verl never provides `num_tokens`, the per-token normalization in `finalize_model_grads` is skipped, and the **MoE aux/z-loss gradients are over-scaled by ~`num_tokens` (tens of thousands)**, which dominates the policy gradient.

## Environment

- verl: `0.8.0.dev`
- Megatron-LM: https://github.com/insightLLM/Megatron-LM/tree/mtp-local-changes

- megatron-bridge : 0.4.1
- Model: Qwen3.5-style MoE (MoE + Gated Delta Net), `~35B-A3B`
- Parallelism: `TP/PP/EP/...` with `CP=2`, `use_remove_padding=True`
- Symptom only appears at `CP > 1`.

## Symptom

```
# CP=2 (broken)
actor/grad_norm: 3168.13...
val-core/.../acc/mean@1: decreasing over steps

# CP=1 (healthy), or CP=2 with moe_aux_loss_coeff=0 & moe_z_loss_coeff=0
actor/grad_norm: ~0.33
```

Setting `moe_aux_loss_coeff=0` and `moe_z_loss_coeff=0` at `CP=2` makes `grad_norm` collapse from ~3000 back to ~0.33 — i.e. the explosion comes from the **MoE auxiliary / z losses**, not the policy loss.

## Root cause

Three pieces interact:

1. **Some models force `calculate_per_token_loss=True` under CP.** For these models, disabling the flag is not an option (hard assert: `"... only supports context parallelism with calculate_per_token_loss enabled"`). With `CP=1` the flag is *not* forced, which is exactly why the bug is CP-only.

2. **The MoE router pre-scales aux/z loss by the local token count** when `calculate_per_token_loss=True` (`megatron/core/transformer/moe/router.py`):

```python
if self.calculate_per_token_loss:
    num_tokens = valid_token_count if valid_token_count is not None else activation.shape[0]
    activation = MoEAuxLossAutoScaler.apply(activation, aux_loss * num_tokens)
```

   This multiplication is intentional: it assumes the final gradient will later be divided by the global token count.

3. **That division only happens if `loss_func` returns a 3-tuple.** In `megatron/core/pipeline_parallel/schedules.py`:

```python
outputs = loss_func(output_tensor)
if len(outputs) == 3:
    output_tensor, num_tokens, loss_reduced = outputs
    # total_num_tokens accumulates here -> used by finalize_model_grads to divide all grads
    ...
else:
    assert len(outputs) == 2          # <-- verl's path
    output_tensor, loss_reduced = outputs
    output_tensor *= cp_group_size
    output_tensor /= num_microbatches
```

   With `calculate_per_token_loss=True`, DDP sets `gradient_scaling_factor = 1.0` (it asserts `average_in_collective=False`), so the *only* place the `* num_tokens` pre-scaling of the aux/z loss gets cancelled is the `1/total_num_tokens` division inside `finalize_model_grads`. Because verl returns a 2-tuple, `total_num_tokens` is never accumulated and that division never runs. The aux/z gradient therefore stays multiplied by ~`num_tokens` (tens of thousands), exploding `grad_norm` and swamping the policy gradient → eval degradation.

This is invisible at `CP=1` only because the flag is not forced there.

## Fix

Make verl's Megatron actor `loss_func` return the 3-tuple `(loss_sum, num_tokens, metrics)` **when `calculate_per_token_loss` is enabled**, so Megatron's per-token-loss pipeline normalizes policy + aux + z gradients by the global token count. The legacy 2-tuple path is kept untouched for the common `calculate_per_token_loss=False` case.

In `verl/workers/actor/megatron_actor.py`, inside `loss_func`, replace the final `return policy_loss, [metrics, ret_entropy]` with:

```python
if not forward_only and getattr(self.tf_config, "calculate_per_token_loss", False):
    cp_size = mpu.get_context_parallel_world_size()
    # response_mask is the full (CP-gathered) response, so its sum is identical on every
    # CP rank. Divide by cp_size so the dp_cp all-reduce inside finalize_model_grads
    # recovers the true global token count instead of double-counting CP ranks.
    full_num_tokens = response_mask.sum()
    # int32 to match Megatron's `total_num_tokens` accumulator (avoids in-place
    # int32 += int64/float dtype-cast errors in the pipeline schedule).
    num_tokens = (full_num_tokens // cp_size).clamp_min(1).detach().to(torch.int32)
    # Scale the (seq-mean-token-mean) loss to token-sum scale so finalize's
    # 1/total_num_tokens recovers the intended per-token-mean gradient and cancels the
    # router's aux/z `* num_local_tokens` pre-scaling.
    loss_sum = policy_loss * full_num_tokens
    return loss_sum, num_tokens, [metrics, ret_entropy]

return policy_loss, [metrics, ret_entropy]
```

Key points:
- `num_tokens` is divided by `cp_size` because `response_mask` is the CP-gathered (full) response and the `dp_cp` all-reduce inside `finalize_model_grads` would otherwise count each CP rank.
- `num_tokens` is cast to `int32` to match Megatron's `total_num_tokens` accumulator dtype.
- `loss_sum = policy_loss * full_num_tokens` puts the loss on a token-sum scale so that finalize's `1/total_num_tokens` recovers the intended per-token-mean gradient and cancels the router's `* num_local_tokens` pre-scaling.

## Result

After the fix, at `CP=2` `actor/grad_norm` returns to ~0.33 and validation scores recover/improve — matching `CP=1` behavior.

## Proposed PR

Happy to send a PR (I have the change running). The change is gated behind `calculate_per_token_loss` so it is a no-op for existing non-CP / non-per-token-loss setups.

https://github.com/insightLLM/verl/tree/qwen3_5-gdn-thd-mm-agentloop-fixes

EricMarcus-ai:
### What does this PR do?

Fixes CP correctness bug in Megatron + Verl interaction.

When you run CP>1, Megatron-Bridge silently enables a per-token loss mode (`calculate_per_token_loss=True`), which expects a different normalization contract. That makes sense, the sequence is now distributed over ranks, so normalization has to be done carefully. In that mode, you are supposed to return the `num_tokens` for each micro-batch part (the non-masked tokens) to Megatron, so it can normalize correctly across ranks. verl did not do this, so Megatron saw 0 total tokens and skipped normalization → grad_norm explodes. Nothing in the stack warns you this is happening.

The grad_norm explosion only showed up in the z/aux losses, because verl doesn't normalize those itself, that's Megatron's job. The "normal" (policy) losses look fine in terms of grad_norm because verl normalizes them itself, but they are also silently slightly wrong for any heterogeneous CP>1 — they get normalized as if tokens were uniformly distributed across CP ranks, since verl never passed `num_tokens`.

Fix: when `calculate_per_token_loss` is enabled, the Megatron engine's loss path returns the 3-tuple `(loss_sum, num_tokens, output)` Megatron's pipeline schedule expects, using a globally all-reduced routed-token count so the reduction reproduces the intended per-token mean. This affects any CP>1 + Megatron run in verl.

Closes #6609.

### Checklist Before Starting

- [x] Search for similar PRs: [`is:pr calculate_per_token_loss`](https://github.com/verl-project/verl/pulls?q=is%3Apr+calculate_per_token_loss) (none open) — the bug is tracked in #6609, which has no PR attached. The issue's suggested location (`verl/workers/actor/megatron_actor.py`) no longer exists after the worker→engine migration; this fix targets the live engine path.
- [x] Format the PR title as `[{modules}] {type}: {description}` → `[megatron] fix: ...`

### Test

Regression test (`tests/models/test_moe_zloss_per_token_loss.py`): runs one PPO update through a tiny Qwen3-MoE Megatron engine twice,`calculate_per_token_loss` False vs True — and asserts `grad_norm` is invariant to the flag. Pre-fix, `True` blows `grad_norm` up by ~10⁴; post-fix the two match. Also verified in production runs.

### API and Usage Example

No CLI/config signature changes. When the per-token regime is active (CP>1), it adds two fail-closed guards that raise a clear `ValueError`:

- `loss_agg_mode='seq-mean-token-mean'` is rejected. The fix scales the loss by the global token count and lets Megatron's `finalize_model_grads` divide it back out — an exact cancel that preserves whatever normalization verl already applied. That holds for `token-mean`, `seq-mean-token-sum`, and `seq-mean-token-sum-norm`. It does *not* hold for `seq-mean-token-mean`: its inner per-sequence `1/n_s` divides by a CP-local shard token count, which the outer cancel can't recover. Use one of the other three (default is `token-mean`).
- `use_remove_padding=False` (BSHD) is rejected. Verl doesn't pass a `padding_mask` to the MoE router, so it normalizes the aux/z loss by the padding-inclusive token count (`logits.shape[0] = B*S`) while gradients are divided by the real token count — a padding-ratio mis-normalization. THD packs the padding, so `logits.shape[0]` already equals the actual token count. Use THD (`use_remove_padding=True`) or disable CP.

### Design & Code Changes

`verl/workers/engine/megatron/transformer_impl.py`:
- `forward_backward_batch`: when `calculate_per_token_loss` is set, all-reduce `attention_mask.sum()` over the DP group and change it as `routed_num_tokens` (the global router/attention-scope token count, distinct from the existing loss-token `batch_num_tokens`, which is why Guard 2 requires THD).
- `postprocess_micro_batch_func`: in the per-token regime, return `(loss * routed_num_tokens / dp_size, local_num_tokens, output)` so Megatron's `finalize_model_grads` divides by the accumulated global token count and the aux/z `×num_tokens` pre-scaling cancels. The legacy 2-tuple path is unchanged for `calculate_per_token_loss=False`.

#### Things we could remove from this PR / feedback

- The regression test isn't strictly necessary. It exists to demonstrate and lock the invariant; it needs a GPU + a local tokenizer, so if we prefer not to carry a GPU unit test for this, we can drop it.
- The explanatory comments in `transformer_impl.py` are a bit verbose. The normalization contract is subtle (silent flag flip, cross-rank cancellation), so we kept a thorough rationale inline, but we can remove if we prefer.

- [x] Add / Update [the documentation](https://github.com/verl-project/verl/tree/main/docs) — n/a (no user-facing config/API change).
- [x] Add unit or end-to-end test(s) — single-GPU regression test added; CP-correctness validated by the production run (a multi-GPU CP test would need a multi-rank fixture).
- [ ] If your PR is related to the `recipe` submodule; n/a.

### DIFF
--- a/tests/models/test_moe_zloss_per_token_loss.py
+++ b/tests/models/test_moe_zloss_per_token_loss.py
@@ -1,0 +1,156 @@
+# Copyright 2025 Bytedance Ltd. and/or its affiliates
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""MoE router z-loss must not depend on ``calculate_per_token_loss``.
+
+Under context parallelism the Megatron bridge enables ``calculate_per_token_loss``,
+which makes ``apply_z_loss`` scale the z-loss by the local token count, expecting
+``finalize_model_grads`` to divide it back out by ``num_tokens``. verl's loss path
+returns a 2-tuple ``(loss, output)``, so the schedule leaves ``num_tokens=0`` and
+finalize skips that division -- leaving the per-token factor (~thousands)
+uncancelled and blowing up the gradient. The flag must not change the gradient.
+"""
+
+import os
+
+os.environ["NCCL_DEBUG"] = "WARN"
+
+from functools import partial
+
+import numpy as np
+import ray
+import torch
+from transformers import AutoModelForCausalLM, AutoTokenizer, Qwen3MoeConfig
+
+from verl import DataProto
+from verl.single_controller.ray import RayClassWithInitArgs, RayResourcePool, RayWorkerGroup
+from verl.utils import tensordict_utils as tu
+from verl.utils.model import compute_position_id_with_mask, create_random_mask
+from verl.workers.config import ActorConfig, HFModelConfig, McoreEngineConfig, McoreOptimizerConfig
+from verl.workers.engine_workers import TrainingWorker, TrainingWorkerConfig
+from verl.workers.utils.losses import ppo_loss
+from verl.workers.utils.padding import left_right_2_no_padding
+
+Z_LOSS_COEFF = 1e-3
+# A small local model dir to borrow a tokenizer from (same convention as test_engine.py).
+TOKENIZER_REF = os.path.expanduser("~/models/Qwen/Qwen2.5-0.5B")
+
+
+def _create_tiny_moe_model(path):
+    """Save a tiny Qwen3-MoE checkpoint so the router z-loss path is exercised. The
+    engine's forward needs a tokenizer (pad_token_id), so we borrow one and size the
+    vocab to match. Returns the vocab size."""
+    tokenizer = AutoTokenizer.from_pretrained(TOKENIZER_REF)
+    config = Qwen3MoeConfig(
+        vocab_size=len(tokenizer),
+        hidden_size=64,
+        intermediate_size=128,
+        moe_intermediate_size=64,
+        num_hidden_layers=2,
+        num_attention_heads=4,
+        num_key_value_heads=2,
+        head_dim=16,
+        num_experts=8,
+        num_experts_per_tok=2,
+        decoder_sparse_step=1,
+        norm_topk_prob=True,
+        max_position_embeddings=512,
+        tie_word_embeddings=False,
+    )
+    AutoModelForCausalLM.from_config(config, torch_dtype=torch.bfloat16).save_pretrained(path)
+    tokenizer.save_pretrained(path)
+    return len(tokenizer)
+
+
+def _build_data(vocab_size):
+    batch_size, seqlen = 4, 32
+    response_length = seqlen // 2
+    torch.manual_seed(1)
+    np.random.seed(1)
+    input_ids = torch.randint(0, vocab_size, (batch_size, seqlen))
+    attention_mask = create_random_mask(
+        input_ids=input_ids, max_ratio_of_valid_token=0.8, max_ratio_of_left_padding=0.2, min_ratio_of_valid_token=0.6
+    )
+    position_ids = compute_position_id_with_mask(attention_mask)
+    responses = input_ids[:, response_length:]
+    response_mask = attention_mask[:, response_length:]
+    data = DataProto.from_single_dict(
+        {
+            "input_ids": input_ids,
+            "prompts": input_ids[:, :response_length],
+            "attention_mask": attention_mask,
+            "position_ids": position_ids,
+            "responses": responses,
+            "response_mask": response_mask,
+            "old_log_probs": torch.randn_like(responses, dtype=torch.float32),
+            "advantages": torch.randn_like(responses, dtype=torch.float32),
+            "ref_log_prob": torch.randn_like(responses, dtype=torch.float32),
+        },
+        meta_info={"temperature": 1.0, "global_token_num": torch.sum(attention_mask, dim=-1).tolist()},
+    )
+    data_td = left_right_2_no_padding(data.to_tensordict())
+    tu.assign_non_tensor(data_td, global_batch_size=data_td.shape[0])
+    return data_td
+
+
+def _grad_norm(model_path, calculate_per_token_loss, data_td):
+    """Run one PPO update through a Megatron MoE engine and return the grad norm.
+    A fresh Ray session is used per call so the two single-GPU worker groups don't
+    contend for the device."""
+    config = TrainingWorkerConfig(
+        model_type="language_model",
+        model_config=HFModelConfig(path=model_path, use_remove_padding=True),
+        engine_config=McoreEngineConfig(
+            forward_only=False,
+            use_mbridge=True,
+            vanilla_mbridge=False,  # NVIDIA Megatron-Bridge (production path; ISEEKYAN mbridge is deprecated)
+            tensor_model_parallel_size=1,
+            pipeline_model_parallel_size=1,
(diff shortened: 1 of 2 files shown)


HollowMan6:
@EricMarcus-ai Could you try to rebase your changes on top of main and remove the changes from https://github.com/verl-project/verl/commit/785b8632858206814277c030a147c95aff140747 in the PR here? I think we would be good to go if e2e_sft_vlm still passes after https://github.com/verl-project/verl/commit/785b8632858206814277c030a147c95aff140747 gets removed in the PR here, and it looks like https://github.com/verl-project/verl/commit/785b8632858206814277c030a147c95aff140747 is redundant if your fix lands.

EricMarcus-ai:
@HollowMan6 Rebased, and removed the change you mentioned. The e2e test you mention ran successfully on this CI before [link](https://github.com/verl-project/verl/actions/runs/28645407160/job/85915628044), I suspect that it should still do so now.

HollowMan6:
Unfortunately it failed https://github.com/verl-project/verl/actions/runs/29041819851/job/86206304492?pr=6836 @EricMarcus-ai Would you mind looking into this issue?

EricMarcus-ai:
@HollowMan6 Found it, on the SFT path the batch only carries `loss_mask` (no `attention_mask` or `response_mask`), so the fallback KeyError'd. Fixed by taking the routed-token count from `attention_mask` when present, otherwise the packed `input_ids` count.

While validating this, I also caught a normalization bug in the same code. `finalize_model_grads` all-reduces the returned `num_tokens` over the DP*CP group, but the count is CP-replicated, so returning the full value is over-counted by `cp_size,` and every gradient came out `1/cp_size` too small at CP>1. Now we return the per-CP-rank share (divide by `cp_size`). CP=1 is unchanged so the existing regression test still holds.

Validated on GPU: a CP1 vs CP2 decomposition confirmed the factor (0.52 to ~1.0 after the fix), and the failing megatron CP2 SFT config now trains cleanly.

HollowMan6 (approved):
Thank you! LGTM, cc: @wuxibin89 to double check

PR #6887, opened 2026-06-29
[megatron] fix: support FP8 padding for BSHD actor forward

gem-mint:
## What

This is the main-branch companion to #6884, requested by maintainers so the BSHD FP8 actor update fix is available on main as well as release/v0.7.1.

It removes the hard BSHD FP8 assertion in the Megatron actor forward path, propagates `use_fp8_padding` into regular BSHD preprocessing for the input and logits processor arguments, and aligns BSHD padded sequence length so TransformerEngine FP8 block quantization sees a local token count divisible by 128. The main branch already had equivalent alignment logic in `preprocess_bshd_engine`; this patch factors that into a helper and reuses it for regular BSHD too.

## Why

With `data_format=bshd` and `fp8=hybrid`/`e4m3`, actor update currently fails before forward with:

```text
AssertionError: fp8 is not supported for bshd format yet
```

Removing the assertion alone is not enough: the padded BSHD shape must also satisfy TE FP8 block quantization alignment.

## Duplicate-work check

Per `AGENTS.md`, I checked for overlapping work before opening this PR:

- `gh issue view 5508 --repo verl-project/verl --comments`
- `gh pr list --repo verl-project/verl --state open --search "5508 in:body" --limit 20`
- `gh pr list --repo verl-project/verl --state open --search "bshd fp8" --limit 20`
- `gh pr list --repo verl-project/verl --state open --search "actor fp8 padding" --limit 20`
- `gh pr list --repo verl-project/verl --state open --search '"fp8 is not supported for bshd format yet"' --limit 20`

The only same-scope result is #6884, which targets `release/v0.7.1`. #6703 appears in some keyword searches but fixes a THD/CP OOB issue and is not a duplicate of this BSHD actor-padding fix.

## Tests

- `git diff --check` passed
- `python -m py_compile tests/models/test_mcore_bshd_fp8_padding.py verl/models/mcore/util.py verl/models/mcore/model_forward.py` passed using the local bundled Python
- `python -m pytest tests/models/test_mcore_bshd_fp8_padding.py -q` could not run locally because this Windows helper environment does not have `pytest` installed
- A direct execution attempt of the test function could not run locally because this helper environment does not have `torch` installed

AI assistance was used to prepare this patch; human review done before marking ready for review.

### DIFF
--- a/verl/models/mcore/util.py
+++ b/verl/models/mcore/util.py
@@ -41,6 +41,16 @@ def _compute_fp8_thd_align_size(align_size: int) -> tuple[int, int]:
     return math.lcm(16, align_size), align_size * 128
 
 
+def _align_bshd_max_seqlen_for_fp8(
+    max_seqlen: int, batch_size: int, align_size: int, tp_size: int, cp_size: int
+) -> int:
+    """Align BSHD max sequence length for TransformerEngine FP8 block quantization."""
+    fp8_total_align = 128 * tp_size * cp_size
+    fp8_seq_align = fp8_total_align // math.gcd(batch_size, fp8_total_align)
+    fp8_seq_align = math.lcm(fp8_seq_align, align_size)
+    return ((max_seqlen + fp8_seq_align - 1) // fp8_seq_align) * fp8_seq_align
+
+
 def preprocess_packed_seqs(
     input_ids: torch.Tensor, attention_mask: torch.Tensor, pre_process: bool = True, use_fp8_padding: bool = False
 ) -> tuple[torch.Tensor, PackedSeqParams]:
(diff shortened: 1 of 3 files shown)


verl/models/mcore/util.py
```diff
@@ -199,6 +209,7 @@ def preprocess_bshd(
     position_ids: torch.Tensor,
     sequence_parallel: bool = False,
     pre_process: bool = True,
+    use_fp8_padding: bool = False,
```
wuxibin89, line 212: `preprocess_bshd` is not used anymore, we now use `preprocess_bshd_engine`

wuxibin89:
Never mind, I decide not to support fp8 in BSHD format.

PR #6895, opened 2026-06-29
[rollout] feat: add optional per-wave rollout resource warmup/cleanup hooks

ChangyiYang:
### What does this PR do?

Adds two **optional** hook points so users can warm up / clean up external rollout
resources (e.g. agent sandboxes / sidecars) **once per rollout wave**, overlapping
their cold-start cost with rollout execution instead of paying it lazily on the
critical path of each rollout.

Two new config knobs under `actor_rollout_ref.rollout.agent`:

```yaml
actor_rollout_ref:
  rollout:
    agent:
      warmup_rollout_resources_path: null   # e.g. my_pkg.warmup.warmup_rollout_resources
      cleanup_rollout_resources_path: null  # e.g. my_pkg.warmup.cleanup_rollout_resources
```

Each is a fully-qualified path to a function `fn(batch, config)` (sync or async),
resolved once per `AgentLoopWorker` via the existing `load_class_from_fqn`
(the same mechanism already used for `agent_loop_manager_class`). On every wave:

- `warmup_rollout_resources(batch, config)` runs **before** `generate_sequences`, so
  the user can pre-spawn resources for `len(batch)` rollouts (the worker's chunk =
  exactly how many this wave needs);
- `cleanup_rollout_resources(batch, config)` runs in `finally` **after** the wave, so
  resources are released even if rollout raises.

Both default to `null` → **no behavior change**.

### Why per-wave on `AgentLoopWorker.generate_sequences`?

- **Process locality**: the resource pool and its consumers (the agent loops) live in
  the `AgentLoopWorker` process; warm-up must run there. The manager level only splits
  and dispatches `worker.generate_sequences.remote(chunk)`, so hooking it would just
  fan back out to the workers.
- **Exact count**: the worker's chunk length is exactly how many resources this wave
  needs on this worker.
- **Lifetime scoping**: `start`/`finally` bounds the resources to one wave.

### What this PR intentionally does NOT include

The actual resource provider / pool / sandbox implementation is **left to the user** —
this PR only adds the framework-side hook points. A user can, for example, pre-spawn a
pool of sandboxes in `warmup_rollout_resources` and have their custom `AgentLoop` pop
from it, then drain the pool in `cleanup_rollout_resources`.

### Test

- New CPU-only unit test `tests/experimental/agent_loop/test_rollout_resource_hooks_on_cpu.py`
  covering: FQN resolution (None / valid / non-callable), sync+async hook support,
  ordering (warmup → impl → cleanup), no-op when unset, cleanup-runs-on-error, and
  warmup-failure-skips-impl. All 8 pass.
- `ruff check` / `ruff format --check` clean on changed files.
- Regenerated `_generated_*_trainer.yaml` via `scripts/generate_trainer_config.sh`.

### API and Usage Example

```python
# my_pkg/warmup.py
async def warmup_rollout_resources(batch, config):
    n = len(batch)                       # how many rollouts in this wave
    await my_pool.ensure_ready(n)        # pre-spawn N sandboxes concurrently

async def cleanup_rollout_resources(batch, config):
    await my_pool.drain()                # release the wave's resources
```

### DIFF
--- a/tests/experimental/agent_loop/test_rollout_resource_hooks_on_cpu.py
+++ b/tests/experimental/agent_loop/test_rollout_resource_hooks_on_cpu.py
@@ -1,0 +1,130 @@
+# Copyright 2024 Bytedance Ltd. and/or its affiliates
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Unit tests for per-wave rollout resource warmup/cleanup hooks (CPU-only, no GPU/Ray)."""
+
+import asyncio
+
+import pytest
+
+from verl.experimental.agent_loop.agent_loop import (
+    _maybe_await,
+    _resolve_rollout_resource_hook,
+    _with_rollout_resource_hooks,
+)
+
+# Module-level functions so they are importable via fully-qualified name (mirrors real usage).
+
+
+def sync_hook(batch, config):
+    _CALLS.append(("sync", batch, config))
+
+
+async def async_hook(batch, config):
+    _CALLS.append(("async", batch, config))
+
+
+_CALLS: list = []
+NOT_CALLABLE = 123
+
+
+def test_maybe_await_passthrough_and_await():
+    async def _run():
+        assert await _maybe_await(5) == 5
+
+        async def coro():
+            return 7
+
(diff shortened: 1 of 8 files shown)


wuxibin89:
@ChangyiYang AgentLoopManager is a reference implementation of an agent framework and is designed to be fully replaceable by other agent frameworks: [How to Extend verl](https://verl.readthedocs.io/en/latest/extend_guide.html#agent-framework-developer).
```bash
+actor_rollout_ref.rollout.agent.agent_loop_manager_class=my_package.module.MyAgentLoopManager
```

For example, check this PR https://github.com/verl-project/uni-agent/pull/73 to see how uni-agent integrate agent in blackbox sandbox.

ChangyiYang:
Closing this PR. After discussion, verl does not currently plan to support sandbox resource management / warm-up in core.

Sandbox and blackbox-agent use cases are handled in the [verl-project/uni-agent](https://github.com/verl-project/uni-agent) project (agent framework + gateway + sandbox runner), which integrates with verl through the existing `agent_loop_manager_class` extension point without requiring core changes. Please refer to uni-agent for sandbox-related needs.

Thanks for the review!

PR #5338, opened 2026-02-17
[fsdp,algo] feat: add use_length_grouped_bsz to reduce padding waste in FSDP training

roeeh21:
## Summary

For models that cannot use sequence packing (e.g. Mamba/SSM), training under FSDP wastes compute on padded tokens. This PR adds an opt-in `use_length_grouped_bsz` option that mitigates that waste by greedily grouping sequences of similar length into micro-batches.

(See [blog post](https://www.ai21.com/blog/padding-minimization-efficiency/) for the broader motivation behind padding-aware micro-batching.)

## Approach

When `actor_rollout_ref.model.use_length_grouped_bsz=True`:

- Sort sequences in the batch by length (descending).
- Greedily pack them into micro-batches such that `len(micro_batch) * max_seq_in_micro_batch <= max_token_len_per_gpu`.
- All-reduce the micro-batch count across DP ranks (split the largest batches on lagging ranks) so the optimizer step stays synchronized.

Off by default. Mutually exclusive with `use_remove_padding` and Ulysses SP.

## Files

- `verl/utils/seqlen_balancing.py`: new helper `get_length_grouped_micro_batches()` returning `(micro_batches, indices)`, mirroring the contract of `rearrange_micro_batches`.
- `verl/workers/engine/utils.py`: dispatcher branch for `use_length_grouped_bsz` in `prepare_micro_batches`.
- `verl/workers/engine/fsdp/transformer_impl.py`: validation guards (mutually exclusive with `use_remove_padding` and Ulysses SP).
- `verl/workers/engine_workers.py`: forwards the flag through to training and inference data.
- Configs: new `use_length_grouped_bsz` field in model + actor configs (YAML + dataclass), plus the regenerated `_generated_ppo_*` yamls.

## Benchmarks

10-step GRPO on Qwen 2.5 3B-Instruct / GSM8K / 1x4 H100. Mean of `timing_s/update_actor` over steps 2-10:

| Config | mean update_actor (s) |
|---|---|
| `use_remove_padding=True` (rmpad) | 2.235 |
| **`use_length_grouped_bsz=True`** | **2.475** |
| dense (no rmpad, no dynamic_bsz) | 11.692 |

`use_length_grouped_bsz` is **~4.7x faster than dense** and **~10.7% slower than `use_remove_padding`**.

## Test Plan

- [x] 10-step GRPO run with `use_length_grouped_bsz=True` completes end-to-end on Qwen 2.5 3B / GSM8K / 1x4 H100.
[...]

PR #6901, opened 2026-06-30
[megatron, perf] feat: pad BSHD micro-batches to mini-batch max seq_len

YaoweiFan:
### What does this PR do?

In the BSHD path (`use_remove_padding=False`), each micro-batch was independently padded to its own `seqlens_in_batch.max()`, so the 32 micro-batches of a mini-batch produced 20+ different padded `s_q` shapes. Each novel shape triggers a cuDNN fused-attention **Graph Build** (~1s of synchronous CPU-side work inside the TransformerEngine `.so`, invisible to the Python profiler), and the resulting plan is cached per shape per stream.
image
This issue was also mentioned in [NVIDIA/TransformerEngine#2033]

This PR adds an opt-in `actor_rollout_ref.actor.megatron.pad_bshd_to_minibatch_max` flag (default `false` = legacy behavior). When enabled, every micro-batch of a mini-batch is padded to the mini-batch's **global** max sequence length, so the padded `s_q` is shared across micro-batches. The cuDNN execution plan is then built once per shape instead of once per micro-batch, eliminating the repeated ~1s CPU stalls.

### Checklist Before Starting

  - `gh pr list --repo verl-project/verl --state all --search "forced_max_seqlen"` → no results
  - `gh pr list --repo verl-project/verl --state all --search "preprocess_bshd_engine"` → only #6506 (preserves BSHD top-k trailing dims; unrelated to this change)
  - Related but distinct: #5338 (`use_length_grouped_bsz` for FSDP reduces padding *waste* via length-sorted grouping; this PR targets the Megatron BSHD path and a different cost — cuDNN plan rebuild from unequal micro-batch shapes). No overlap in code path or mechanism.
  - No existing issue describes this cuDNN plan-cache miss problem.

### Test

We benchmarked on Qwen3.5-35B-A3B (BSHD path) for 15 steps; the "Verl Fix" variant has pad_bshd_to_minibatch_max enabled. The per-stage mean timings are listed below.
Stage | Verl (s) | Verl Fix (s)
-- | -- | --
Total Step Time | 534.13 | 397.71
Gen/Rollout | 171.06 | 170.86
<mark>Old Log Prob<mark> | <mark>78.38<mark> | <mark>35.45<mark>
<mark>Actor Update<mark> | <mark>247.58<mark> | <mark>154.50<mark>
Weight Synchronization | 11.04 | 10.73
Ref Model Computation | 24.73 | 24.84

As the table shows, the Old Log Prob and Actor Update stages exhibit a substantial time reduction, primarily attributable to the elimination of repeated Graph Builds.

### API and Usage Example

New opt-in engine flag, off by default. No change to existing behavior when disabled.

```yaml
actor_rollout_ref:
  actor:
    megatron:
      use_remove_padding: False          # BSHD path
      pad_bshd_to_minibatch_max: True    # unify micro-batch seq_len within a mini-batch
```

### Design & Code Changes

The unified max is computed once per mini-batch and carried to the deep `preprocess_bshd_engine` call, where all TP/CP/FP8 alignment already lives.

- `verl/workers/config/engine.py`: add `pad_bshd_to_minibatch_max: bool = False` to `McoreEngineConfig`.
- `verl/workers/engine/megatron/transformer_impl.py`: in `forward_backward_batch`, when the flag is on and `use_remove_padding=False`, compute the mini-batch's raw (unaligned) global max seq_len from the nested `input_ids` and inject it per micro-batch via NonTensorData **after** chunking. Both `MegatronEngineWithLMHead.forward_step` and `MegatronEngineWithValueHead.forward_step` read it back and pass it to the forward fn.
- `verl/models/mcore/util.py`: `preprocess_bshd_engine` and `build_vlm_attn_mask_bshd` accept `forced_max_seqlen`; when not None it overrides `seqlens_in_batch.max()`. All TP/CP/FP8 alignment is kept as the single source inside these functions, so callers pass the **unaligned raw** max (this is also required for FP8, whose alignment depends on local `batch_size`).
- `verl/models/mcore/model_forward.py`: thread `forced_max_seqlen` through `gptmodel_forward_model_engine` to the 3 `preprocess_bshd_engine` calls (input_ids, MTP label/loss_mask, logits_processor args) and the VLM `build_vlm_attn_mask_bshd` call. The VLM fix is necessary: without it the VLM branch rebuilt `input_ids_bshd` at the per-micro-batch max (overriding forced_max), causing `logits.shape[:2] != label.shape[:2]` and tripping the logits_processor assert.
- Configs: `verl/trainer/config/engine/megatron.yaml` + regenerated `_generated_ppo_megatron_trainer.yaml` (via `scripts/generate_trainer_config.sh`).

The NonTensorData-per-micro-batch approach (vs. a staleable instance attribute) keeps the value isolated per mini-batch with no cross-batch state leakage across actor/critic/logprob passes on the same engine.

- [x] Apply [pre-commit checks](https://github.com/verl-project/verl/blob/main/CONTRIBUTING.md#code-linting-and-formatting): `ruff`/`autogen-trainer-cfg` verified; full `pre-commit run --all-files` to be run in a complete env.
- [ ] Add / Update [the documentation](https://github.com/verl-project/verl/tree/main/docs). Not done yet — flag is self-documented in the engine config/yaml; will add a docs note if reviewers prefer.
- [ ] Add unit or end-to-end test(s) to [the CI workflow](https://github.com/verl-project/verl/tree/main/.github/workflows) to cover all the code. Existing `tests/utils/test_megatron_bshd_preprocess.py` covers `preprocess_bshd_engine`; a focused unit test for the `forced_max_seqlen` override path will be added.
- [ ] Once your PR is ready for CI, send a message in [the \`ci-request\` channel](https://verl-project.slack.com/archives/C091TCESWB1) in [the \`verl\` Slack workspace](https://join.slack.com/t/verl-project/shared_invite/zt-3855yhg8g-CTkqXu~hKojPCmo7k_yXTQ).
- [ ] If your PR is related to the `recipe` submodule, please also update the reference to the submodule commit.

---

**AI assistance disclosure**: This PR was developed with AI assistance. The submitting human has reviewed every changed line and is responsible for the change end-to-end. It is not a duplicate of an existing PR (see search links above). Local verification: `py_compile` passes, `ruff`/`autogen-trainer-cfg` pre-commit hooks verified; full Megatron-env regression tests and training comparison to be run by the submitter before requesting review.

### DIFF
--- a/verl/trainer/config/_generated_ppo_megatron_trainer.yaml
+++ b/verl/trainer/config/_generated_ppo_megatron_trainer.yaml
@@ -65,7 +65,7 @@ actor_rollout_ref:
       use_mbridge: true
       vanilla_mbridge: true
       use_remove_padding: true
-      pad_bshd_to_minibatch_max: false
+      pad_bshd_to_minibatch_max: true
       use_megatron_fsdp: false
       forward_only: false
       dtype: bfloat16
@@ -252,7 +252,7 @@ actor_rollout_ref:
       use_mbridge: ${oc.select:actor_rollout_ref.actor.megatron.use_mbridge,False}
       vanilla_mbridge: ${oc.select:actor_rollout_ref.actor.megatron.vanilla_mbridge,True}
       use_remove_padding: ${oc.select:actor_rollout_ref.actor.megatron.use_remove_padding,True}
-      pad_bshd_to_minibatch_max: false
+      pad_bshd_to_minibatch_max: true
       use_megatron_fsdp: false
       forward_only: true
       dtype: bfloat16
@@ -574,7 +574,7 @@ critic:
     use_mbridge: true
     vanilla_mbridge: true
     use_remove_padding: true
-    pad_bshd_to_minibatch_max: false
+    pad_bshd_to_minibatch_max: true
     use_megatron_fsdp: false
     forward_only: false
     dtype: bfloat16
--- a/verl/trainer/config/_generated_ppo_megatron_trainer.yaml
+++ b/verl/trainer/config/_generated_ppo_megatron_trainer.yaml
@@ -65,6 +65,7 @@ actor_rollout_ref:
       use_mbridge: true
       vanilla_mbridge: true
       use_remove_padding: true
+      pad_bshd_to_minibatch_max: false
       use_megatron_fsdp: false
       forward_only: false
       dtype: bfloat16
@@ -251,6 +252,7 @@ actor_rollout_ref:
       use_mbridge: ${oc.select:actor_rollout_ref.actor.megatron.use_mbridge,False}
       vanilla_mbridge: ${oc.select:actor_rollout_ref.actor.megatron.vanilla_mbridge,True}
       use_remove_padding: ${oc.select:actor_rollout_ref.actor.megatron.use_remove_padding,True}
+      pad_bshd_to_minibatch_max: false
       use_megatron_fsdp: false
       forward_only: true
       dtype: bfloat16
@@ -572,6 +574,7 @@ critic:
     use_mbridge: true
     vanilla_mbridge: true
     use_remove_padding: true
+    pad_bshd_to_minibatch_max: false
     use_megatron_fsdp: false
     forward_only: false
     dtype: bfloat16
--- a/verl/models/mcore/model_forward.py
+++ b/verl/models/mcore/model_forward.py
@@ -273,6 +273,7 @@ def gptmodel_forward_model_engine(
     data_format: str = "thd",
     mtp_enable_train: bool = False,
     local_cp_size: Optional[int] = None,
+    forced_max_seqlen: Optional[int] = None,
 ):
     """Default forward pass for GPT models with optional sequence packing."""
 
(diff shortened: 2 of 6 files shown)


wuxibin89:
@YaoweiFan Which model are you training? The BSHD is not efficient in RL training since variable sequence length.

YaoweiFan:
@wuxibin89   Qwen3.5-35B-A3B. We started on BSHD because, at the time, the Megatron GDN path didn't support THD for this model, so `use_remove_padding=True` wasn't available — that may have been resolved recently on the Megatron side; happy to switch to THD if it now covers this model.

Regardless, for workloads still on the BSHD path, the variable per-micro-batch max seq_len produced many distinct padded `s_q` shapes per mini-batch in our run, and each novel shape triggered a ~1s cuDNN fused-attention Graph Build. This PR adds an opt-in `pad_bshd_to_minibatch_max` for the BSHD path: pad every micro-batch to the mini-batch's global max so the plan is built once per shape instead of per micro-batch. Default off; THD path untouched — purely a BSHD mitigation, not a replacement for THD.

verl/trainer/config/_generated_ppo_megatron_trainer.yaml
```diff
@@ -65,6 +65,7 @@ actor_rollout_ref:
       use_mbridge: true
       vanilla_mbridge: true
       use_remove_padding: true
+      pad_bshd_to_minibatch_max: false
```
wuxibin89: Should we enable it by default?

YaoweiFan: I've flipped the default to `true`, so BSHD training benefits by default. This doesn't affect the THD path — `pad_bshd_to_minibatch_max` is guarded by `not use_remove_padding`, so it's a no-op when `use_remove_padding=True`.

wuxibin89 (approved):

PR #6907, opened 2026-07-01
Add NPU Triton GDN kernels for Qwen3.5

Kirrito-k423:
## Summary
- add the Qwen3.5 gated delta rule Triton kernels under `verl.models.transformers.ops.gdn`
- patch Qwen3.5 and Qwen3.5-MoE GDN modules on NPU so `use_triton_gdn=True` injects the verl Triton fused path instead of relying on external FLA/MindSpeed imports
- enable the Triton GDN path by default for the Qwen3.5 27B/35B GRPO FSDP NPU examples, with `USE_TRITON_GDN=false` as an escape hatch

## Validation
- Local: `PYTHONPATH=/Users/Zhuanz/Documents/autoresearch/.tmp/verl-gdn-work/verl pytest --rootdir=/Users/Zhuanz/Documents/autoresearch/.tmp/verl-gdn-work/verl tests/models/test_qwen3_5_gdn_patch.py -q`
- Local: `python3 -m compileall -q verl/models/transformers/ops/gdn verl/models/transformers/qwen3_5_gdn_patch.py tests/models/test_qwen3_5_gdn_patch.py`
- Local: `git -c core.whitespace=blank-at-eol,blank-at-eof,space-before-tab,cr-at-eol diff --cached --check`
- NPU: Ascend910 on `zxb_qwen35_20260519` (`torch_npu 2.7.1.post4`, Triton 3.2.0) vector-add Triton sanity passed with max error 0.0
- NPU: Qwen3.5 GDN fused forward matched the torch reference at max abs error 4.8828125e-4 / mean abs error 1.9698203686857596e-5, and fused backward produced finite gradients for q/k/v/g/beta
- NPU: patched HF `Qwen3_5GatedDeltaNet` injected `verl.models.transformers.ops.gdn.chunk_gated_delta_rule.chunk_gated_delta_rule`

### DIFF
--- a/examples/grpo_trainer/run_qwen3_5_27b_fsdp.sh
+++ b/examples/grpo_trainer/run_qwen3_5_27b_fsdp.sh
@@ -25,6 +25,7 @@ TRAIN_FILE=${TRAIN_FILE:-"${RAY_DATA_HOME}/data/geo3k/train.parquet"}
 TEST_FILE=${TEST_FILE:-"${RAY_DATA_HOME}/data/geo3k/test.parquet"}
 WORKING_DIR=${WORKING_DIR:-"${PWD}"}
 RUNTIME_ENV=${RUNTIME_ENV:-"${WORKING_DIR}/verl/trainer/runtime_env.yaml"}
+USE_TRITON_GDN=${USE_TRITON_GDN:-auto}
 ########################### end user-adjustable ###########################
 
 ########################### derived defaults ###########################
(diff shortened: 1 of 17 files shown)


wuxibin89:
Is there a common place where we can maintain Ascend Triton kernel? I think veomni also need these kernels.

Kirrito-k423:
Thanks for the suggestion. I agree that a common upstream place would be better for maintaining these Ascend Triton kernels, especially if both verl and veomni need them.

This PR was mainly intended as a functional spike / proof-of-concept to validate the Qwen3.5 GDN path in verl. For the long-term path, I think we should first prioritize integrating/upstreaming this through `flash-linear-attention`, and then have verl / veomni consume the shared implementation from there.

I will close this PR for now and revisit after the `flash-linear-attention` path is clearer.

PR #6912, opened 2026-07-01
[rollout] fix: empty device cache before waking KV cache in colocate mode

kenkenpa2126:
### What does this PR do?

Fixes an OOM error in colocate mode (`free_cache_engine=True`) that occurs when resuming the KV cache after weight synchronization.

#### Background: two independent GPU memory management layers

CUDA manages GPU physical pages through two separate, non-coordinating layers:

- **CUDA Runtime allocator** (used by PyTorch via `cudaMalloc`): maintains an internal pool. When a tensor is deleted, its physical pages go into a free-list *within* the pool — they are **not** returned to the CUDA driver. `torch.cuda.empty_cache()` explicitly flushes this pool back to the driver.
- **CUDA VMM API** (used by vLLM via `cuMemCreate`): bypasses the Runtime allocator entirely and requests physical pages **directly from the CUDA driver**. It has no visibility into what the Runtime allocator holds.

This asymmetry is the root cause: pages sitting in PyTorch's free-list are invisible to `cuMemCreate`, even though `nvidia-smi` counts them as available.

#### How the OOM occurs (colocate mode)

Each iteration runs the following sequence in `ActorRolloutRefWorker.update_weights()`:

1. `resume(["weights"])` — vLLM rebuilds the weight arena via `cuMemCreate`
2. `update_weights_from_ipc()` — actor weights are transferred into vLLM workers via ZMQ using temporary `cudaMalloc` buffers (Runtime layer)
3. Temporary buffers are freed → pages enter PyTorch's free-list, **not** returned to driver
4. `resume(["kv_cache"])` — vLLM calls `cuMemCreate` to rebuild the KV-cache arena → **OOM**: driver cannot see the pages held by PyTorch's Runtime allocator

#### Fix

Call `get_torch_device().empty_cache()` on all vLLM worker processes (via `collective_rpc("empty_cache")`) immediately before `wake_up(["kv_cache"])`. This flushes PyTorch's free-list and returns the physical pages to the CUDA driver, making them available to `cuMemCreate`.

The fix is guarded by the existing `free_cache_engine` flag and only fires when `"kv_cache" in tags`. Separate/standalone rollout modes are unaffected — they transfer weights via NCCL/NIXL (no temporary `cudaMalloc` buffers), so this accumulation does not occur.

### Checklist Before Starting

- [x] Search for similar PRs. Paste at least one query link here: https://github.com/verl-project/verl/pulls?q=colocate+kv+cache+oom

### Test

This fix cannot be tested in CI (requires colocate mode with specific GPU memory pressure). Verified by directly calling `cuMemCreate` via the CUDA driver API to reproduce the exact failure mode.

The script allocates a large tensor (simulating `update_weights_from_ipc()` copy buffers), deletes it, then calls `cuMemCreate` for a comparably-sized block — confirming it fails while PyTorch holds the pages in its cache (`memory_reserved` stays high, `memory_allocated` drops to 0). After `torch.cuda.empty_cache()`, `memory_reserved` drops and the same `cuMemCreate` call succeeds, directly demonstrating why the fix works.

Additionally validated on a real colocate training run (V1 trainer) using a patched version of this fix. The `cuMemCreate` OOM after weight sync was resolved.

<details>
<summary>Verification script</summary>

```python
import ctypes, sys, torch

# Load the CUDA *Driver* API (libcuda.so.1), NOT the Runtime API (libcudart.so).
# CUDA exposes two separate libraries:
#   libcudart.so  — Runtime API (cudaMalloc/cudaFree): used by PyTorch
#   libcuda.so.1  — Driver API  (cuMemCreate/cuMemRelease): used by vLLM for VMM arenas
# By loading libcuda.so.1 directly via ctypes we call cuMemCreate in exactly
# the same way vLLM does, making the OOM (and its fix) directly observable.
try:
    _cuda = ctypes.CDLL("libcuda.so.1")
except OSError:
    sys.exit("libcuda.so.1 not found")

CUDA_SUCCESS = 0

# CUmemLocation / CUmemAllocationProp: C structs required by cuMemCreate.
# We set type=PINNED and location=DEVICE, matching vLLM's KV-cache allocation.
class _CUmemLocation(ctypes.Structure):
    _fields_ = [("type", ctypes.c_int), ("id", ctypes.c_int)]

class _CUmemAllocationProp(ctypes.Structure):
    _fields_ = [("type", ctypes.c_int), ("requestedHandleTypes", ctypes.c_int),
                ("location", _CUmemLocation), ("win32HandleMetaData", ctypes.c_void_p),
                ("allocFlags", ctypes.c_uint64)]

def _cumem_alloc(size_bytes, device_id=0):
    """Attempt to allocate size_bytes via cuMemCreate (CUDA VMM API).

    This is the same API vLLM uses for its weight/KV-cache arenas.
    Returns True on success, False on OOM or error.
    Releases the handle immediately on success (we only need to know it worked).
    """
    prop = _CUmemAllocationProp()
    prop.type = 1             # CU_MEM_ALLOCATION_TYPE_PINNED
    prop.location.type = 1   # CU_MEM_LOCATION_TYPE_DEVICE
    prop.location.id = device_id
    # cuMemCreate requires size to be a multiple of the allocation granularity
    # (typically 2 MiB on modern GPUs).
    gran = ctypes.c_size_t(0)
    _cuda.cuMemGetAllocationGranularity(ctypes.byref(gran), ctypes.byref(prop), 1)
    size = ((size_bytes + gran.value - 1) // gran.value) * gran.value
    handle = ctypes.c_uint64(0)
    ret = _cuda.cuMemCreate(ctypes.byref(handle), ctypes.c_size_t(size), ctypes.byref(prop), 0)
    if ret == CUDA_SUCCESS:
        _cuda.cuMemRelease(handle); return True
    return False

def _mem(label):
    r = torch.cuda.memory_reserved() / 1024**2
    a = torch.cuda.memory_allocated() / 1024**2
    # reserved  = bytes held by PyTorch's Runtime allocator (pool + free-list)
    # allocated = bytes used by live tensors right now
    # cached    = reserved - allocated = pages in free-list, invisible to cuMemCreate
    print(f"  {label:<45} reserved={r:6.0f} MB  allocated={a:6.0f} MB  cached={r-a:6.0f} MB")

device = 0; torch.cuda.set_device(device)
total_mb = torch.cuda.get_device_properties(device).total_memory / 1024**2
dv = ctypes.c_int(0); _cuda.cuDriverGetVersion(ctypes.byref(dv)); v = dv.value
print(f"torch       : {torch.__version__}")
print(f"CUDA driver : {v // 1000}.{(v % 1000) // 10}")
print(f"GPU {device}       : {torch.cuda.get_device_name(device)}, {total_mb:.0f} MB total\n")

_ = torch.empty(1, device="cuda"); torch.cuda.empty_cache()

alloc_mb = int(total_mb * 0.6)
alloc_elems = alloc_mb * 1024 * 1024 // 4  # float32
probe_bytes = int(alloc_mb * 0.8) * 1024 * 1024

# Step 1: simulate update_weights_from_ipc() temporary cudaMalloc buffers
print("Step 1 — allocate large tensor (simulating weight-sync residual)")
t = torch.empty(alloc_elems, dtype=torch.float32, device="cuda")
_mem("after alloc")

# Step 2: buffers freed — pages enter PyTorch's free-list (Runtime layer),
#         NOT returned to the CUDA driver. nvidia-smi still shows them as "Used".
print("\nStep 2 — delete tensor (Python GC runs, pages go to PyTorch cache)")
del t; _mem("after del")
cached_mb = torch.cuda.memory_reserved() / 1024**2
print(f"\n  => PyTorch cache holds {cached_mb:.0f} MB")
print(f"     cuMemCreate (driver view): only {total_mb - cached_mb:.0f} MB appears free")

# Step 3: simulate wake_up(["kv_cache"]) — cuMemCreate talks to the Driver layer
#         directly and cannot see the pages held by the Runtime layer → OOM.
print(f"\nStep 3 — cuMemCreate({probe_bytes // 1024**2} MB) WITHOUT empty_cache()")
ok = _cumem_alloc(probe_bytes, device)
print("  => FAILED (OOM) — driver cannot reuse pages held by PyTorch cache ✓" if not ok else "  => succeeded")

# Step 4: empty_cache() bridges the two layers — flushes PyTorch's Runtime pool
#         and returns physical pages to the CUDA driver. [the fix]
print("\nStep 4 — torch.cuda.empty_cache()  [the fix]")
torch.cuda.empty_cache(); _mem("after empty_cache()")
cached_mb = torch.cuda.memory_reserved() / 1024**2
print(f"\n  => PyTorch cache now holds {cached_mb:.0f} MB")
print(f"     cuMemCreate (driver view): {total_mb - cached_mb:.0f} MB free")

# Step 5: retry cuMemCreate — pages are now visible to the driver → succeeds.
print(f"\nStep 5 — cuMemCreate({probe_bytes // 1024**2} MB) AFTER empty_cache()")
ok = _cumem_alloc(probe_bytes, device)
print("  => succeeded ✓" if ok else "  => failed")
```

</details>

<details>
<summary>Output</summary>

```
torch       : 2.7.0a0+79aa17489c.nv25.04
CUDA driver : 12.9
GPU 0       : NVIDIA H100 80GB HBM3, 81008 MB total

Step 1 — allocate large tensor (simulating weight-sync residual)
  after alloc                                   reserved= 48606 MB  allocated= 48604 MB  cached=     2 MB

Step 2 — delete tensor (Python GC runs, pages go to PyTorch cache)
  after del                                     reserved= 48606 MB  allocated=     0 MB  cached= 48606 MB

  => PyTorch cache holds 48606 MB
     cuMemCreate (driver view): only 32402 MB appears free

Step 3 — cuMemCreate(38883 MB) WITHOUT empty_cache()
  => FAILED (OOM) — driver cannot reuse pages held by PyTorch cache ✓

Step 4 — torch.cuda.empty_cache()  [the fix]
  after empty_cache()                           reserved=     2 MB  allocated=     0 MB  cached=     2 MB

  => PyTorch cache now holds 2 MB
     cuMemCreate (driver view): 81006 MB free

Step 5 — cuMemCreate(38883 MB) AFTER empty_cache()
  => succeeded ✓
```

</details>

### Design & Code Changes

- **`vLLMColocateWorkerExtension.empty_cache()`** (`verl/workers/rollout/vllm_rollout/utils.py`): thin wrapper around `get_torch_device().empty_cache()`, dispatched to all vLLM worker processes via `collective_rpc`.
- **`ServerAdapter.resume()`** (`verl/workers/rollout/vllm_rollout/vllm_rollout.py`): calls `empty_cache` on all vLLM workers before `wake_up(["kv_cache"])`, guarded by the existing `free_cache_engine` flag.

The weight sync sequence that triggers this OOM lives in `ActorRolloutRefWorker.update_weights()` (`verl/workers/engine_workers.py`). The fix is only reached when `backend="naive"` (colocate mode, set unconditionally in `verl/trainer/ppo/v1/trainer_base.py`); separate async mode asserts `backend != "naive"` and returns before calling `resume()`.

**References:**
- [PyTorch CUDA Caching Allocator internals](https://zdevito.github.io/2022/08/04/cuda-caching-allocator.html) — explains why `cudaMalloc` pages are held in a pool and not returned until `empty_cache()`
- [CUDA Virtual Memory Management](https://docs.nvidia.com/cuda/cuda-programming-guide/04-special-topics/virtual-memory-management.html) — documents `cuMemCreate` and the driver-level physical page management
- [vLLM sleep mode](https://github.com/vllm-project/vllm/blob/main/docs/features/sleep_mode.md) — confirms vLLM uses VMM (`cuMemCreate`/`cuMemRelease`) for its weight/KV-cache arenas

> [!IMPORTANT]

- [x] Add unit or end-to-end test(s): not feasible for CI (requires colocate GPU setup with specific memory pressure); verified by standalone reproduction script above.
- [ ] Once your PR is ready for CI, send a message in [the `ci-request` channel](https://verl-project.slack.com/archives/C091TCESWB1) in [the `verl` Slack workspace](https://join.slack.com/t/verl-project/shared_invite/zt-3855yhg8g-CTkqXu~hKojPCmo7k_yXTQ). (If not accessible, please try [the Feishu group (飞書群)](https://applink.larkoffice.com/client/chat/chatter/add_by_link?link_token=772jd4f1-cd91-441e-a820-498c6614126a).)


wuxibin89:
We already empty cache in both sender and receiver side:
- sender: https://github.com/verl-project/verl/blob/main/verl/workers/rollout/vllm_rollout/bucketed_weight_transfer.py#L159
- receiver: https://github.com/verl-project/verl/blob/main/verl/workers/rollout/vllm_rollout/bucketed_weight_transfer.py#L294

kenkenpa2126:
Closing this PR. I was misled because adding this change made the OOM disappear in my end-to-end runs.

I instrumented the colocate weight-sync → `wake_up(["kv_cache"])` path and measured memory right before/after the `empty_cache` this PR adds — it frees ~0 MB. At that point the vLLM worker has no reclaimable PyTorch cache, because `BucketedWeightReceiver._cleanup()` (which @wuxibin89 pointed out) has already flushed it. So the mechanism I claimed — freeing PyTorch's free-list so `cuMemCreate` can see the pages — doesn't actually hold on the real path.

As a hypothesis for why the OOM still went away in my runs: I suspect the change acted as an incidental synchronization rather than reclaiming memory — the extra `collective_rpc` round-trip letting the worker's in-flight weight-load work settle before `cuMemCreate`. Reading the code, there does seem to be a synchronization gap (verl refs pinned to 9d05508f):

The worker's last GPU op is `process_weights_after_loading` in `update_weights_from_ipc`, and the method then returns with no synchronize() after it: https://github.com/verl-project/verl/blob/9d05508f5e3bd8ecb70cf94ab10dc087b57a716d/verl/workers/rollout/vllm_rollout/utils.py#L305
The only worker-side sync is in `BucketedWeightReceiver._cleanup`, but that runs inside `receive_weights`, i.e. before the `process_weights_after_loading` above: https://github.com/verl-project/verl/blob/9d05508f5e3bd8ecb70cf94ab10dc087b57a716d/verl/workers/rollout/vllm_rollout/bucketed_weight_transfer.py#L324
`aggressive_empty_cache(force_sync=True)` runs in the actor process, which is a separate process from the vLLM workers (weights are shipped over ZMQ IPC, methods dispatched via `collective_rpc`). Since CUDA contexts/streams are per-process, its `device.synchronize()` only covers the actor's own context — not the worker processes' streams, even though they share the same physical GPU: https://github.com/verl-project/verl/blob/9d05508f5e3bd8ecb70cf94ab10dc087b57a716d/verl/workers/engine_workers.py#L742
`ServerAdapter.resume` then issues `wake_up` (→ `cuMemCreate`) with no worker sync in between: https://github.com/verl-project/verl/blob/9d05508f5e3bd8ecb70cf94ab10dc087b57a716d/verl/workers/rollout/vllm_rollout/vllm_rollout.py#L204 — and on the vLLM side, `CuMemAllocator.wake_up()` calls `create_and_map()` with no `synchronize()` before it (confirmed in both current main and v0.17.0, the version these runs used).
If that reading is right, the principled fix would be an explicit synchronization after `process_weights_after_loading`, not an `empty_cache`. But this is only a hypothesis — I couldn't cleanly isolate/confirm it, so I'm open to being wrong here (please point out anything I'm missing).

Thanks @wuxibin89 for the pointer that led me to dig in.

PR #6913, opened 2026-07-01
[fsdp] fix: fused-kernel gradient tracking bug.

kolehma8:
### What does this PR do?

Currently using use_liger and use_fused_kernel flags create silently corrupted result. The issue is that the .flatten(0,1) inside the fused kernels autograd forward body flips the  requires_grad flag as gradients are not tracked by default within the autograd Functions. In the backward pass, the hidden_states requires_grad is false and that returns None as the gradients. This silently breaks the backpropagation and the reward does not improve.

This PR uses the original requires_grad flag from the hidden_states without relying on the underlying behavior of .flatten(0,1).

Relevant code paths:
* https://github.com/verl-project/verl/blob/1bf9d2d76c244238e8219531c582819a6b407d47/verl/utils/experimental/torch_functional.py#L107
* https://github.com/verl-project/verl/blob/1bf9d2d76c244238e8219531c582819a6b407d47/verl/utils/experimental/torch_functional.py#L165

### Checklist Before Starting

### Test

Tested offline.

### API and Usage Example

```python
```

### Design & Code Changes

> [!IMPORTANT]

### DIFF
--- a/verl/utils/experimental/torch_functional.py
+++ b/verl/utils/experimental/torch_functional.py
@@ -104,9 +104,9 @@ class FusedLinearForPPOFunction(torch.autograd.Function):
         if orig_ndim == 3:
             assert input_ids.ndim == 2, f"input_ids shape doesn't match, {hidden_states.shape} {input_ids.shape}"
             orig_batch_size = hidden_states.shape[0]
-            hidden_states_required_grad = hidden_states.requires_grad
+            hidden_states_requires_grad = hidden_states.requires_grad
             hidden_states = hidden_states.flatten(0, 1)
-            hidden_states.requires_grad = hidden_states_required_grad
+            hidden_states.requires_grad_(hidden_states_requires_grad)
             input_ids = input_ids.flatten(0, 1)
 
         T = hidden_states.shape[0]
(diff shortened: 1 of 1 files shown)


yueyiming2009:
Nice catch, and the diagnosis is spot on. Worth noting why it's silent: `vocab_weights`/`lm_head` isn't flattened, so the head keeps training while the backbone is frozen.

The fix works, but restoring `requires_grad` on an intermediate tensor is a bit side-effecty. Since the real consumers are the `if hidden_states.requires_grad:` checks in `backward`, gating on `ctx.needs_input_grad` is more idiomatic and robust here. The engine snapshots that tuple from the original `.apply()` args *before* `forward` runs under no-grad — so unlike the flag on the flattened tensor, it isn't clobbered by `flatten(0, 1)`, and needs no flag mutation:

```python
# needs_input_grad is positional over forward args:
# [0] -> hidden_states, [1] -> vocab_weights (input_ids/scalars are False)
if ctx.needs_input_grad[0]:      # was: hidden_states.requires_grad
    dhidden_states = torch.zeros_like(hidden_states)
if ctx.needs_input_grad[1]:      # was: vocab_weights.requires_grad
    dvocab_weights = torch.zeros_like(vocab_weights)
```

This also lets you leave the `flatten(0, 1)` line untouched. Either way fixes the bug — just flagging the alternative.

Might also be worth a small regression test on the 3D path asserting `hidden_states.grad is not None`.

wuxibin89:
@kolehma8 I think yiming's solution is better, could you adjust the fix according to his advice?

wuxibin89 (approved):

PR #6933, opened 2026-07-06
[megatron] feat: migrate fused logprob/entropy from GPTModel.forward monkey-patch to Megatron output_processor hook

chengcuiping:
### What does this PR do?

`verl/models/mcore/model_forward_fused.py` currently replaces the entire `GPTModel.forward` with `_fused_GPTModel_forward` so the training loop can pass `temperature` into Megatron's `_postprocess` boundary, where `linear_cross_entropy` computes fused log-probabilities and entropy.

Megatron-Core 0.18 includes a native `output_processor` / `output_processor_context` extension point at exactly that boundary, introduced by [NVIDIA/Megatron-LM#4686](https://github.com/NVIDIA/Megatron-LM/pull/4686) for [NVIDIA/Megatron-LM#4590](https://github.com/NVIDIA/Megatron-LM/issues/4590). This PR uses that contract to avoid replacing `GPTModel.forward` on the current Megatron stack: `temperature` travels through `output_processor_context`, and a callback runs the fused log-probability and entropy kernel at the postprocess hook.

This revision is rebased onto current `main` after [verl-project/verl#7101](https://github.com/verl-project/verl/pull/7101), which upgraded the model CI stack to Megatron-Core `core_v0.18.0`.

Compared with the first revision of this PR, the `VERL_FUSED_USE_OP_HOOK` environment switch has been removed. The native path is now selected automatically when the bound GPT model exposes both hook parameters. For still-supported older models without the complete hook contract, the existing legacy forward patch is selected automatically.

### Checklist Before Starting

- [x] Search for similar PRs: https://github.com/verl-project/verl/pulls?q=is%3Apr+output_processor+fused
- [x] Format the PR title as `[{modules}] {type}: {description}`: `[megatron] feat: migrate fused logprob/entropy from GPTModel.forward monkey-patch to Megatron output_processor hook`
- [x] No duplicate open PR implementing the same verl-side migration was found.

#### Fresh validation on the final post-#7101 revision

```text
pytest -s -x tests/models/test_model_forward_fused.py
8 passed

pytest -s -x tests/utils/test_megatron_bshd_preprocess.py
10 passed

pytest -s -x tests/workers/test_megatron_value_head_cp_layout_on_cpu.py
1 passed
```

The focused tests cover:

- native capability detection and one-time signature inspection;
- native and legacy patch/unpatch idempotence;
- native mode preserving `GPTModel.forward` identity and not creating `forward_backup`;
- automatic legacy fallback when either hook parameter is unavailable;
- `temperature` propagation through `output_processor_context`;
- the real Megatron Bridge wrapper chain;
- tied and untied output-weight resolution;
- sequence-parallel gather before `linear_cross_entropy`;
- current FP8-padding, `min_local_rows`, `cp_layout`, and THD arguments;
- config-logger payload preservation.

The wrapper-contract test exercises:

```text
Float16Module
-> DistributedDataParallel
-> GPTModel.forward
-> _postprocess
-> output_processor
```

The contract was checked against Megatron-Core `core_v0.18.0` and Megatron Bridge `r0.5.0`. Both wrappers propagate the hook keyword arguments, the callback is invoked using `context=`, native `GPTModel.forward` retains its original identity, and no `forward_backup` is created.

Additional final-revision gates:

- changed-files pre-commit: passed;
- Python compile checks: passed;
- `git diff upstream/main...HEAD --check`: clean;
- format-patch reconstruction: passed;
- bundle and artifact SHA256 verification: passed.

#### Earlier numerical and live integration validation

The following numerical A/B and live integration results were completed on an earlier revision of this PR before the final post-#7101 rebase. They were not rerun in full on the final revision. The subsequent refactor removed the selection environment variable and retained the callback's fused computation; the fresh tests above cover the final native/fallback selection and wrapper behavior.

Environment used for the earlier same-batch numerical validation:

```text
Megatron-Core: 0.18.0+97f3bce88
PyTorch: 2.9.1+cu128
Transformer Engine: 2.10
flash-attn: 2.8.3
dtype: bf16
```

| Scenario | Coverage | Result |
| --- | --- | --- |
| Synthetic same-batch A/B | TP=1/2, tied and untied embeddings, both output-weight branches, sequence-parallel gather | `log_probs` / `entropy` `max_abs=0` |
| Real model-build regression | Manual wrapping versus the real `get_model` build chain | Bit-identical |
| Real training-shape batch | Variable-length padding, THD `padding_causal`, GRPO advantages, including the covered `pg_loss` path | `max_abs=0` |
| Environment with vLLM installed | Dependency-contamination regression | Bit-identical |
| Live Megatron engine path | GSM8K tokens, nested/THD inputs, TP=2 and TP=1, `fused_forward_model_engine` | `log_probs` / `entropy` `max_abs=0` |

Forward memory usage was also measured as identical between the hook and legacy paths (`Δ=0`).

A separate earlier 5-step smoke used a dense Qwen2.5-0.5B model on 4×A100 with:

- Megatron engine TP=2 and PP=1;
- `use_fused_kernels=True`;
- full parameter, gradient, and optimizer offload;
- vLLM rollout and GRPO on GSM8K;
- rollout-to-actor weight resharding;
- fused log-probability and entropy computation;
- optimizer updates;
- distributed checkpoint save at step 4.

Both the native-hook and legacy-patch runs completed all five steps without NaN or infinity. Actor/rollout probability correlation remained approximately `0.9996` per step.

The two runs used independently sampled rollout trajectories, so their per-step metrics are not treated as a numerical equivalence comparison. Exact numerical equivalence is provided by the identical-weight, identical-batch A/B results above. Checkpoint reload/resume was not independently exercised.

#### Rebased local limitations

`tests/models/test_engine.py` was attempted on the final revision. One test passed, but the remaining FSDP case could not complete because the local fixture `/root/models/Qwen/Qwen2.5-0.5B` was unavailable. This is a missing local fixture and is not reported as a product test result.

The official post-#7101 image was unavailable locally. The following therefore still require project CI:

- `verl-ci-cn-beijing.cr.volces.com/verlai/verl:vllm024.dev2`;
- TP >= 2 with `use_fused_kernels=True`;
- confirmation that the native callback executes;
- finite forward, log-probability, and entropy outputs;
- at least one optimizer update.

The final local validation does not claim FP8 numerical execution, pipeline parallelism, MTP, or checkpoint reload/resume coverage.

### API and Usage Example

No public API, CLI, environment variable, or configuration surface is added or changed. Existing users continue to enable fused kernels in the same way:

```yaml
actor_rollout_ref:
  model:
    use_fused_kernels: true
```

When the bound GPT model exposes both `output_processor` and `output_processor_context`, verl uses the native hook automatically. Otherwise, it uses the existing legacy patch.

The `VERL_FUSED_USE_OP_HOOK` environment variable present in the first, unmerged revision of this PR has been removed. It was never part of a verl release, so no released user-facing interface is being removed.

### Design & Code Changes

The native path is selected only when the bound GPT model exposes both `output_processor` and `output_processor_context`.

On that path:

- `GPTModel.forward` remains unchanged and no `forward_backup` is created.
- `temperature` is removed from the native GPT arguments and passed only through `output_processor_context`.
- The callback accepts Megatron's `context=` keyword. Megatron's `_postprocess` invokes the callback with `context=output_processor_context`, while the `GPTModel.forward` call uses `output_processor_context=...`.
- Tied embeddings use `output_weight`.
- Untied embeddings use `output_layer.weight`.
- Sequence-parallel hidden states are gathered before `linear_cross_entropy`.
- Config-logger inputs and fused log-probability/entropy outputs are preserved.

Capability is resolved once during model construction through signature inspection rather than a version-string comparison. Runtime forwards do not repeat the inspection. Wrapper unwrapping reaches the underlying GPT model before capability detection.

If either hook parameter is absent, model construction selects the existing legacy forward patch. Patch and unpatch operations remain idempotent on both paths.

The rebase preserves current-main handling of `cp_layout`, `min_local_rows`, FP8 padding arguments, THD preprocessing, decoder tuple outputs, vision inputs, and `moe_n_hash_layers`.

Migrating `model_forward_1f1b_overlap.py` to Megatron's schedule-plan hook is intentionally outside this PR and should be handled as a separate follow-up.

- [x] Apply [pre-commit checks](https://github.com/verl-project/verl/blob/main/CONTRIBUTING.md#code-linting-and-formatting). An earlier all-files run passed, and changed-file checks passed again after the final rebase.
- [x] Add / Update [the documentation](https://github.com/verl-project/verl/tree/main/docs): no documentation update is required because this revision adds no released user-facing API, CLI, environment variable, or configuration behavior.
- [x] Add unit or end-to-end tests to [the CI workflow](https://github.com/verl-project/verl/tree/main/.github/workflows): the focused test was added to the existing `model_engine` workflow. Official-image TP>=2 and optimizer-update coverage require the maintainer-approved CI run.
- [ ] Send a message in the [`ci-request` channel](https://verl-project.slack.com/archives/C091TCESWB1) in the verl Slack workspace. If Slack is inaccessible, use the linked Feishu group from the contribution template.
- [x] This PR is not related to the `recipe` submodule.

AI assistance disclosure: OpenAI Codex assisted with implementation and test execution; I reviewed the resulting changes and outputs before submission.

### DIFF
--- a/verl/models/mcore/model_forward_fused.py
+++ b/verl/models/mcore/model_forward_fused.py
@@ -14,7 +14,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import inspect
 from collections import OrderedDict
+from dataclasses import dataclass
 from typing import Optional
 
 import megatron.core as mcore
@@ -36,6 +38,92 @@ from verl.utils.model import CausalLMOutputForPPO
 
 from .util import postprocess_packed_seqs_for_dict_output, postprocess_thd_engine
 
+_FUSED_FORWARD_MODE_ATTR = "_verl_fused_forward_mode"
+_HOOK_MODE = "hook"
+_LEGACY_MODE = "legacy"
+
+
+def _supports_output_processor_hook(patching_model: torch.nn.Module) -> bool:
+    parameters = inspect.signature(patching_model.forward).parameters
+    return {"output_processor", "output_processor_context"}.issubset(parameters)
+
+
+def _resolve_fused_forward_mode(patching_model: torch.nn.Module) -> str:
+    return _HOOK_MODE if _supports_output_processor_hook(patching_model) else _LEGACY_MODE
+
+
+def _get_fused_forward_mode(model: torch.nn.Module) -> str:
+    model = unwrap_model(model)
+    mode = getattr(model, _FUSED_FORWARD_MODE_ATTR, None)
+    if mode is None and hasattr(model, "language_model"):
+        mode = getattr(model.language_model, _FUSED_FORWARD_MODE_ATTR, None)
+    return mode if mode in (_HOOK_MODE, _LEGACY_MODE) else _LEGACY_MODE
+
+
+def _use_output_processor_hook(model: torch.nn.Module) -> bool:
+    return _get_fused_forward_mode(model) == _HOOK_MODE
+
+
+@dataclass
+class FusedOutputProcessorContext:
+    """Context passed through Megatron's native output-processor hook."""
+
+    temperature: float
+
+
+def fused_output_processor(
+    *,
+    hidden_states,
+    output_layer,
+    output_weight,
+    labels,
+    context,
+    config,
+    **_ignored,
+):
+    """Compute fused log probabilities and entropy at Megatron's postprocess boundary."""
+    output = CausalLMOutputForPPO(
+        loss=None,
+        logits=None,
+        past_key_values=None,
+        hidden_states=hidden_states,
+        attentions=None,
+    )
+
+    if config.sequence_parallel:
+        hidden_states = gather_from_sequence_parallel_region(hidden_states)
+
+    # Megatron passes the shared embedding as output_weight for tied models. For
+    # untied models the weight lives on output_layer.
+    weight = output_weight if output_weight is not None else output_layer.weight
+
+    temperature = context.temperature
+    logprobs, entropy = linear_cross_entropy(
+        hidden_states,
+        weight,
+        labels,
+        temperature,
+        "none",
+        parallel_state.get_tensor_model_parallel_group(),
+    )
+
+    if has_config_logger_enabled(config):
+        payload = OrderedDict(
+            {
+                "input_ids": _ignored.get("input_ids"),
+                "position_ids": _ignored.get("position_ids"),
+                "attention_mask": _ignored.get("attention_mask"),
+                "decoder_input": _ignored.get("decoder_input"),
+                "logprobs": logprobs,
+                "entropy": entropy,
+            }
+        )
+        log_config_to_disk(config, payload, prefix="input_and_logits")
+
+    output.entropy = entropy
+    output.log_probs = logprobs
+    return output
+
 
 def _get_patching_model(model: torch.nn.Module):
     model = unwrap_model(model)
@@ -50,19 +138,33 @@ def _get_patching_model(model: torch.nn.Module):
 
 
 def patch_fused_forward(model: torch.nn.Module):
+    model = _get_patching_model(model)
+    if model is None:
+        return
+
+    mode = getattr(model, _FUSED_FORWARD_MODE_ATTR, None)
+    if mode is None:
+        mode = _resolve_fused_forward_mode(model)
+        setattr(model, _FUSED_FORWARD_MODE_ATTR, mode)
+
+    if mode == _HOOK_MODE:
+        return
+
     assert version.parse(mcore.__version__) >= version.parse("0.13.0"), (
         "Fused forward patching requires mecore >= 0.13.0"
     )
-    model = _get_patching_model(model)
-    if model is not None:
+    if not hasattr(model, "forward_backup"):
         model.forward_backup = model.forward
         model.forward = _fused_GPTModel_forward.__get__(model, model.__class__)
 
 
 def unpatch_fused_forward(model: torch.nn.Module):
     model = _get_patching_model(model)
-    if model is not None:
+    if model is None or _get_fused_forward_mode(model) == _HOOK_MODE:
+        return
+    if hasattr(model, "forward_backup"):
         model.forward = model.forward_backup
+        delattr(model, "forward_backup")
 
 
 def fused_forward_model_gen(vision_model: bool = False):
@@ -117,7 +219,15 @@ def fused_forward_model_gen(vision_model: bool = False):
             input_args["input_ids"] = input_ids
             input_args["attention_mask"] = attention_mask
 
-        output_orig: CausalLMOutputForPPO = model(**input_args)
+        if _use_output_processor_hook(model):
+            input_args.pop("temperature", None)
+            output_orig: CausalLMOutputForPPO = model(
+                **input_args,
+                output_processor=fused_output_processor,
+                output_processor_context=FusedOutputProcessorContext(temperature=temperature),
+            )
+        else:
+            output_orig: CausalLMOutputForPPO = model(**input_args)
 
         if post_process:
             # output_orig is in type of CausalLMOutputForPPO
@@ -195,15 +305,22 @@ def fused_forward_model_engine(vision_model: bool = False):
             cp_layout=cp_layout,
         )
         labels_rmpad = labels_rmpad.contiguous()
-        output_orig: CausalLMOutputForPPO = model(
+        forward_kwargs = dict(
             input_ids=input_ids_rmpad,
             attention_mask=attention_mask,
             position_ids=None,
             packed_seq_params=packed_seq_params,
             labels=labels_rmpad,
-            temperature=temperature,
             **model_kwargs,
         )
+        if _use_output_processor_hook(model):
+            output_orig: CausalLMOutputForPPO = model(
+                **forward_kwargs,
+                output_processor=fused_output_processor,
+                output_processor_context=FusedOutputProcessorContext(temperature=temperature),
+            )
+        else:
+            output_orig: CausalLMOutputForPPO = model(temperature=temperature, **forward_kwargs)
 
         if not post_process:
             return output_orig
--- a/verl/models/mcore/model_forward_fused.py
+++ b/verl/models/mcore/model_forward_fused.py
@@ -44,6 +44,15 @@ _LEGACY_MODE = "legacy"
 
 
 def _supports_output_processor_hook(patching_model: torch.nn.Module) -> bool:
+    """Check whether the model supports Megatron's native output-processor hook.
+
+    The ``output_processor`` / ``output_processor_context`` contract was
+    introduced in Megatron Core 0.18.0. Models without this contract fall back
+    to the legacy ``forward`` monkey patch.
+
+    TODO: Remove this check and the legacy patch once all supported Megatron
+    stacks require Megatron Core 0.18.0 or newer.
+    """
     parameters = inspect.signature(patching_model.forward).parameters
     return {"output_processor", "output_processor_context"}.issubset(parameters)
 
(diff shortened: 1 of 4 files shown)


chengcuiping:
Quick follow-up: besides the offline bit-identical A/B in the PR description, I ran a small live end-to-end smoke on Megatron-Core 0.18 to make sure the hook path survives the full RL loop, not just an isolated forward.

Setup was a dense Qwen2.5-0.5B on 4×A100, megatron engine (v1) with TP=2/PP=1, `use_fused_kernels=True`, full param/grad/optimizer offload, vLLM rollout, GRPO on gsm8k, 5 steps with a checkpoint at step 4. I ran it twice — once with `VERL_FUSED_USE_OP_HOOK=0` (legacy patch) and once with `=1` (the hook) — so the whole loop gets exercised both ways: rollout → weight resharding → fused logprob/entropy → optimizer update → distributed checkpoint.

Both paths completed all 5 steps and saved checkpoints cleanly, with no NaN/inf and comparable, sane metrics (entropy ~0.43–0.66, training log-ppl ~0.40–0.62 on both). The three integration points I was worried about all held up under the hook: rollout↔actor resharding (actor/rollout prob correlation ~0.9996 each step), offload interacting with the hook path, and dist-checkpoint save/load.

I also added a one-off build-time check to confirm the hook is genuinely what's running: under `VERL_FUSED_USE_OP_HOOK=1` the GPTModel has no `forward_backup` on any rank (i.e. `patch_fused_forward` was a no-op and the native forward is untouched), while under `=0` it does. So the live run really is going through `output_processor`, not the monkey-patch.

One honest caveat: the per-step metrics differ slightly between the two runs, but that's rollout non-reproducibility (vLLM across independent processes samples different trajectories), not the hook — the two runs simply don't see the same data. The exact numerical equivalence is what the offline A/B in the PR description covers (`max_abs=0` on identical weights + batch); this live run is about integration, not re-proving the numerics.

Heads-up on one prerequisite unrelated to this PR: bringing up the megatron backend on Megatron-Core 0.18 currently also needs `mbridge` patched, because `mbridge` 0.15.1 (latest) still references the removed `ModelType.encoder_and_decoder` and crashes at model build — upstream of any forward, so it's orthogonal to this change. Filed separately (mbridge: ISEEKYAN/mbridge#155, verl dependency note: #6936).

verl/models/mcore/model_forward_fused.py
```diff
@@ -37,6 +40,106 @@
...
+
+    # The hook fires before the output layer, so under sequence parallelism gather first.
+    if config.sequence_parallel:
+        hidden_states = gather_from_sequence_parallel_region(hidden_states)
+
+    # output_weight semantics are inverted vs the legacy patch: the hook passes None when the
+    # embedding is NOT tied (real weight lives in output_layer.weight), and the shared weight
+    # when it is tied.
+    weight = output_weight if output_weight is not None else output_layer.weight
+    assert weight is not None, "cannot resolve output weight for fused linear_cross_entropy"
```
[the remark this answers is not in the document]
chengcuiping: Addressed in the latest revision by removing the redundant assertion. The tied and untied output-weight paths are covered by the focused unit tests.

ETOgaosion:
Thanks for contribution! Shall wait after https://github.com/verl-project/verl/pull/7101 to test.

chengcuiping:
Thanks @ETOgaosion ! Once #7101 lands, I’ll rebase #6933 onto the latest main, rerun the focused tests, and request the `model_engine` CI against the new image. I’ll investigate and address any compatibility issues found there.

chengcuiping:
Thanks @ETOgaosion — #7101 has landed, and I’ve now rebased #6933 onto the latest main (current HEAD: `6d1dce7`).

Fresh post-rebase focused tests pass (`8 + 10 + 1`), along with the changed-file pre-commit, compile, and diff checks. I’ve sent the CI request in the Feishu group.

Could you please approve/run the pending workflows when convenient, especially the `model_engine` CI on the post-#7101 image? The main remaining check is TP >= 2 with `use_fused_kernels=True`, confirming native callback execution, finite log-probability/entropy outputs, and an optimizer update.

HollowMan6 (commented):
@chengcuiping Please address the pre-commit issue. Also I think we shall set `USE_FUSED_KERNELS=True` in the CI (probably additional tests in .github/workflows/e2e_ppo_trainer_megatron_vllm_2.yml) to get this properly tested and ensure it doesn't break later

chengcuiping:
Thanks @HollowMan6  — I addressed both points in `117ed39`.

The pre-commit failure came from `ruff-format` rewriting `verl/workers/engine/megatron/delta_export.py` on the previous base. I rebased onto the latest `main`, where that formatting fix is already present, so the file is no longer part of this PR's diff. The full all-files pre-commit check now passes cleanly on two consecutive runs.

I also enabled `USE_FUSED_KERNELS=True` for the standard non-FP8, non-LoRA Megatron-Bridge GRPO E2E case in `.github/workflows/e2e_ppo_trainer_megatron_vllm_2.yml`. The existing FP8-rollout and LoRA cases retain their default non-fused coverage.

Fresh local validation:
- `tests/models/test_model_forward_fused.py`: 8 passed
- `tests/utils/test_megatron_bshd_preprocess.py`: 14 passed
- `tests/workers/test_megatron_value_head_cp_layout_on_cpu.py`: 1 passed

The 8×L20 E2E was not run locally and is pending project CI.

HollowMan6 (changes requested):
@chengcuiping The added CI was not a valid one as it falls back to non-fused. https://github.com/verl-project/verl/actions/runs/30988514265/job/92249964416?pr=6933

(WorkerDict pid=87044) WARNING:2026-08-05 08:38:10,366:Fused kernels require `use_remove_padding=True` for Megatron engine. Falling back to non-fused.

Image

chengcuiping:
Thanks @HollowMan6 — you’re right. I addressed this in `7b30d280`.

The fused Megatron E2E now sets both `USE_FUSED_KERNELS=True` and `USE_REMOVE_PADDING=True`, removing the configuration mismatch that caused the previous run to fall back. The step also captures combined stdout/stderr, checks both the training and `tee` exit statuses, and fails closed if `Falling back to non-fused` appears or the log cannot be inspected.

The updated `e2e_ppo_trainer_megatron_vllm_2` run is now queued:

https://github.com/verl-project/verl/actions/runs/31006523327

I’ll verify that the effective configuration has both options enabled and that no non-fused fallback occurs before treating the E2E as valid.

verl/models/mcore/model_forward_fused.py
```diff
@@ -36,6 +38,92 @@
...
 from .util import postprocess_packed_seqs_for_dict_output, postprocess_thd_engine
 
+_FUSED_FORWARD_MODE_ATTR = "_verl_fused_forward_mode"
+_HOOK_MODE = "hook"
+_LEGACY_MODE = "legacy"
+
+
+def _supports_output_processor_hook(patching_model: torch.nn.Module) -> bool:
+    parameters = inspect.signature(patching_model.forward).parameters
+    return {"output_processor", "output_processor_context"}.issubset(parameters)
```
HollowMan6, line 57: Maybe we shall document here this is supported only with mcore 0.18.0 and above. We can remove the legacy path once we fully move to mcore 0.18.0.

HollowMan6 (dismissed):
lgtm with minor comments otherwise

verl/models/mcore/model_forward_fused.py
chengcuiping: Good point  @HollowMan6  — documented this in `7a6427aa`.

The docstring now notes that the native `output_processor` / `output_processor_context` contract was introduced in Megatron Core 0.18.0, and includes a TODO to remove both the capability check and the legacy patch once all supported Megatron stacks require 0.18.0 or newer.

I kept the capability-based check rather than switching to a version comparison so compatible backports and wrapped implementations remain supported. This follow-up changes only the docstring; the focused tests and pre-commit checks passed.

GitHub dismissed the previous approval as stale after this follow-up commit. Could you please re-approve when convenient?

HollowMan6 (approved):

PR #6965, opened 2026-07-07
[perf, model] feat: add MFU flops estimation for Qwen3.5 (qwen3_5)

ldemon2333:
### What does this PR do?

Adds MFU FLOPs estimation for **Qwen3.5** (`model_type: "qwen3_5"`). Without it, `qwen3_5` falls through to `_estimate_unknown_flops`, which returns `0`, so `perf/mfu/*` is always zero for Qwen3.5 training runs.

Qwen3.5 differs from the models already covered by `flops_counter.py` in several ways, so none of the existing estimators can be reused:
- **Hybrid attention**: every `full_attention_interval`-th layer is standard softmax attention (O(L²)); the remaining layers are **GatedDeltaNet linear attention** (O(L), no seqlen-square term). Reusing `_estimate_qwen2_flops` would apply the quadratic term to *all* layers and miss the linear-attention projections entirely.
- **Nested config**: fields live under `config.text_config` (top level holds `vision_config`), like `qwen3_next` / `qwen3_vl` in transformers.
- **Gated q_proj**: `attn_output_gate=true` doubles the q projection width.
- **Dense SwiGLU MLP** (not MoE), an **MTP** head, and an optional **ViT**.

### Checklist Before Starting

- [x] Search for similar PRs: https://github.com/verl-project/verl/pulls?q=is%3Apr+flops+qwen3_5 — none adding qwen3_5 flops.
- [x] PR title formatted as `[{modules}] {type}: {description}`.

### Test

Added a `qwen3_5` case to `tests/utils/test_flops_counter.py` (config based on the public Qwen/Qwen3.5-9B: 32 layers, 8 full + 24 linear, `head_dim=256`, MTP=1). Full suite passes:

```
tests/utils/test_flops_counter.py ... 12 passed
```

`pre-commit` (ruff check + format) passes on both changed files.

> [!NOTE]
> **Honesty about accuracy**: the GatedDeltaNet recurrence core is modeled with an **O(L) analytical approximation** (the chunked delta-rule kernel's exact FLOPs depend on implementation details). For the 9B config this term is **< 1% of total FLOPs**, so it does not meaningfully move MFU. The estimate is dominated by the deterministic linear-algebra terms (MLP + projections + embeddings ≈ 91%). I did **not** have real Qwen3.5-9B weights to calibrate MFU end-to-end against a live training run — validation covers magnitude, per-term breakdown, and a degeneration check (an all-`full_attention` variant lands within ~8% of the stock `qwen2` estimator, the difference being the q-gate doubling). Feedback from anyone with a real training curve is very welcome.

### Design & Code Changes

- `verl/utils/flops_counter.py`: add `_estimate_qwen3_5_flops`, register `"qwen3_5"` in `ESTIMATE_FUNC`.
- `tests/utils/test_flops_counter.py`: add `qwen3_5` test case.

- [ ] Add / Update the documentation — n/a (internal FLOPs estimator, no user-facing API).
- [x] Add unit test(s) to cover the code.

### DIFF
--- a/tests/utils/test_flops_counter.py
+++ b/tests/utils/test_flops_counter.py
@@ -366,6 +366,44 @@ CONFIG = {
             709446422495232 / 1e12,
         ),
     },
+    "qwen3_5": {
+        "config": {  # Qwen/Qwen3.5-9B: hybrid linear/full attention, dense MLP, MTP head
+            "model_type": "qwen3_5",
+            "text_config": {
+                "vocab_size": 248320,
+                "hidden_size": 4096,
+                "intermediate_size": 12288,
+                "num_hidden_layers": 32,
+                "num_attention_heads": 16,
+                "num_key_value_heads": 4,
+                "head_dim": 256,
+                "attn_output_gate": True,
+                "full_attention_interval": 4,
+                # 8 full-attention layers (every 4th), 24 GatedDeltaNet linear-attention layers
+                "layer_types": [
+                    "linear_attention",
+                    "linear_attention",
+                    "linear_attention",
+                    "full_attention",
+                ]
+                * 8,
+                "linear_key_head_dim": 128,
+                "linear_value_head_dim": 128,
+                "linear_num_key_heads": 16,
+                "linear_num_value_heads": 32,
+                "linear_conv_kernel_dim": 4,
+                "mtp_num_hidden_layers": 1,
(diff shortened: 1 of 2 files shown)


wuxibin89:
close as implemented in https://github.com/verl-project/verl/pull/6389

PR #6977, opened 2026-07-08
[BREAKING][trainer] fix: separate_async should use the same step granularity with other trainers

Begunner:
### What does this PR do?

- Align `separate_async` global step semantics with sync training by running a full `parameter_sync_step` cycle inside one trainer step. 
- Aggregate per-inner-update metrics across the cycle.

### Test

E2E separate async training.
Metrics aggregator and replay buffer tests on cpu.


verl/trainer/config/ppo_trainer.yaml
```diff
@@ -219,9 +219,9 @@ trainer:
     separate_async:
 
       # Number of warmup batches to add before training loop starts
-      num_warmup_batches: 4
+      num_warmup_batches: 1
```
wuxibin89, line 222: Why only warmup a mini-batch in separate_async trainer?

wuxibin89 (approved):

PR #6987, opened 2026-07-09
[fully_async] fix: pre-compute total_training_steps before trainer init

gaohongkui:
### What does this PR do?

Fix `AssertionError: assert self.lr_decay_steps > 0` when using `fully_async_policy` with Megatron model engine. The root cause is an initialization ordering issue where Megatron's `OptimizerParamScheduler` requires `total_training_steps > 0` at `__init__` time, but the value is only computed after rollouter initialization.

### Checklist Before Starting

- [x] Search for similar PRs. Query: https://github.com/verl-project/verl/pulls?q=is%3Apr+fully_async+total_training_steps

### Test

Reproduced with Qwen3.5-122B-A10B MoE on NPU (H910-64G) cluster using `fully_async_main` entry point with Megatron engine (TP=2, PP=4, CP=2, EP=16). Without this fix, training crashes immediately at `init_workers()`. With the fix, training proceeds normally.

Not feasible to add to CI: requires multi-node Megatron + fully_async setup with actual model weights.

### Design & Code Changes

**Root Cause**: In `fully_async_main.py`, the initialization order is:
1. `_create_trainer(config)` → `trainer.init_workers()` → Megatron optimizer init (needs `total_training_steps > 0`)
2. `_create_rollouter(config)` → `rollouter.init_workers()`
3. `trainer.set_total_train_steps(rollouter.get_total_train_steps())` ← too late

The v1 trainer (sync/colocate modes) doesn't hit this because it lazily initializes the optimizer after dataloader setup.

**Fix**: Pre-compute `total_training_steps` from already-available config values before creating the trainer. Mirrors the logic in `v1/trainer_base.py::_init_dataloader`:
- Respects explicit `config.trainer.total_training_steps` if set by the user
- Multiplies by `trigger_parameter_sync_step` for optimizer schedule (consistent with v1)
- Sets both `actor.optim` and `critic.optim` (consistent with v1)

- [x] Apply [pre-commit checks](https://github.com/verl-project/verl/blob/main/CONTRIBUTING.md#code-linting-and-formatting): `ruff check` and `ruff format` pass.
- [ ] Add / Update [the documentation](https://github.com/verl-project/verl/tree/main/docs). (N/A - bug fix, no new API)
- [ ] Add unit or end-to-end test(s). Not feasible: requires multi-node Megatron + fully_async setup.

---

> AI-assisted: Claude was used for investigation and implementation. A human reviewer has verified all changes.

> This is not duplicating an existing PR: no open PRs address this issue (searched "fully_async total_training_steps", "OptimizerParamScheduler lr_decay_steps").

### DIFF
--- a/verl/experimental/fully_async_policy/fully_async_main.py
+++ b/verl/experimental/fully_async_policy/fully_async_main.py
@@ -20,7 +20,7 @@ from pprint import pprint
 
 import hydra
 import ray
-from omegaconf import OmegaConf
+from omegaconf import OmegaConf, open_dict
 
 from verl.experimental.fully_async_policy.fully_async_rollouter import FullyAsyncRollouter
 from verl.experimental.fully_async_policy.fully_async_trainer import FullyAsyncTrainer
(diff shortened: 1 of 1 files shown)


gaohongkui:
Re: Gemini's suggestion to remove the try-except block —

This pattern is intentionally mirrored from `v1/trainer_base.py::_init_dataloader` (lines 598-604) for consistency. The `try-except` guards against configs that lack the `optim` sub-structure entirely (e.g. custom configs that omit it). Since this is a pre-init best-effort setup — the downstream `set_total_train_steps()` will set it again after rollouter init — failing silently with a warning is the correct behavior here.

gaohongkui:
Gentle ping — this has been open for a month with no review yet.

Quick recap of why it may be worth a look: this is a hard crash, not a refinement. Any `fully_async_policy` + Megatron engine combination fails at `init_workers()` with `AssertionError: assert self.lr_decay_steps > 0`, because Megatron's `OptimizerParamScheduler` needs `total_training_steps > 0` at `__init__` time while the value is only computed after rollouter init. The change is +30/-1 in a single file and is still mergeable against current `main` (no conflicts).

If it would make review easier, I'm happy to add a unit test that constructs the config and asserts the ordering, so the fix can be checked by running something rather than by reading the argument. Just say the word and I'll push it.

Also happy to rebase or narrow the change if you'd prefer a different approach.

PR #7012, opened 2026-07-10
[megatron] fix: align teacher tensor seqlen with student for forward_kl_topk CP split

gaohongkui:
### What does this PR do?

Fixes an `AssertionError` in `compute_forward_kl_topk` when `context_parallel_size > 1` and batch samples have variable lengths.

**Root cause**: `preprocess_bshd_engine` independently computes `aligned_max_seqlen` for teacher and student nested tensors from their respective offsets. With variable-length samples, teacher and student can produce different aligned totals → CP zigzag split yields mismatched local sequence dimensions → assertion at L293 fails.

**Fix**: Derive the expected total seqlen from the already-split student logits (`student_logits.shape[1] * cp_size`) and pass it as `forced_max_seqlen` to `preprocess_bshd_engine` for teacher tensors. This ensures both sides pad to the same aligned length before CP split.

Related: #5826 added CP support for bshd format but tested only with fixed-length data (gsm8k), which doesn't trigger the alignment divergence.

### Checklist Before Starting

- [x] Search for similar PRs. Query: https://github.com/verl-project/verl/pulls?q=is%3Apr+forward_kl_topk+cp

### Test

Tested on production workload:
- Model: Qwen3.5-35B-A3B (MoE), student + 122B teacher
- Topology: 10 nodes × 16 NPU (H910-64G), TP=2 PP=2 CP=2 EP=8
- Config: `forward_kl_topk`, `topk=128`, `use_chunked_topk=true`
- Result: Training proceeds normally with distillation metrics (`actor/distillation/loss`, `student_mass`, `teacher_mass`) reported correctly. Previously crashed at step 1 with `AssertionError`.

### API and Usage Example

No API change. Existing configs with `loss_mode: forward_kl_topk` + `context_parallel_size > 1` now work correctly without modification.

### Design & Code Changes

Single-point fix in `verl/trainer/distillation/megatron/losses.py`:
- Before CP split of teacher tensors, compute `forced_max_seqlen = student_logits.shape[1] * cp_size`
- Pass this to `preprocess_bshd_engine(..., forced_max_seqlen=forced_max_seqlen)` for both teacher log_probs and teacher ids
- Improved assertion error message to include actual shapes for easier debugging

- [ ] Add unit or end-to-end test(s). Not feasible: the bug only manifests in multi-node distributed settings with CP > 1 and variable-length batches — cannot be reproduced in CI's single-node environment.

### DIFF
--- a/verl/trainer/distillation/megatron/losses.py
+++ b/verl/trainer/distillation/megatron/losses.py
@@ -284,13 +284,31 @@ def compute_forward_kl_topk(
     assert teacher_topk_log_probs.is_nested and teacher_topk_ids.is_nested
 
     # 1. split across cp groups (bsz, seqlen, topk) => (bsz, seqlen/cp_size, topk)
+    # When CP > 1, teacher and student nested tensors may have different per-sample
+    # lengths (due to padding differences), causing preprocess_bshd_engine to compute
+    # different aligned max_seqlen values. Force teacher to use the same total seqlen
+    # as student so that after CP zigzag split, local dimensions match.
+    from megatron.core import parallel_state as mpu
+
+    cp_size = mpu.get_context_parallel_world_size()
+    forced_max_seqlen = student_logits.shape[1] * cp_size if cp_size > 1 else None
+
     if data_format == "thd":
         teacher_topk_log_probs_cp_split, *_ = preprocess_thd_engine(teacher_topk_log_probs, pre_process=True)
         teacher_topk_ids_cp_split, *_ = preprocess_thd_engine(teacher_topk_ids, pre_process=True)
     else:
-        teacher_topk_log_probs_cp_split, *_ = preprocess_bshd_engine(teacher_topk_log_probs, pre_process=True)
-        teacher_topk_ids_cp_split, *_ = preprocess_bshd_engine(teacher_topk_ids, pre_process=True)
-    assert teacher_topk_log_probs_cp_split.shape[:2] == teacher_topk_ids_cp_split.shape[:2] == student_logits.shape[:2]
+        teacher_topk_log_probs_cp_split, *_ = preprocess_bshd_engine(
+            teacher_topk_log_probs, pre_process=True, forced_max_seqlen=forced_max_seqlen
+        )
+        teacher_topk_ids_cp_split, *_ = preprocess_bshd_engine(
+            teacher_topk_ids, pre_process=True, forced_max_seqlen=forced_max_seqlen
+        )
+    assert (
+        teacher_topk_log_probs_cp_split.shape[:2] == teacher_topk_ids_cp_split.shape[:2] == student_logits.shape[:2]
+    ), (
+        f"Shape mismatch after CP split: teacher_logprobs={teacher_topk_log_probs_cp_split.shape[:2]}, "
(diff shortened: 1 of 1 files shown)


verl/trainer/distillation/megatron/losses.py
```diff
@@ -284,13 +284,31 @@ def compute_forward_kl_topk(
     assert teacher_topk_log_probs.is_nested and teacher_topk_ids.is_nested
 
     # 1. split across cp groups (bsz, seqlen, topk) => (bsz, seqlen/cp_size, topk)
+    # When CP > 1, teacher and student nested tensors may have different per-sample
```
wuxibin89, line 287: It's unexpected that teacher and student nested tensors have different per-sample lengths, they should always have same length.

gaohongkui:
@wuxibin89 You're right that per-sample lengths are the same — both teacher and student nested tensors are built from the same `cu_seqlens` in `padding_2_no_padding`. The mismatch comes from **alignment padding in `preprocess_bshd_engine`** when `pad_bshd_to_minibatch_max=True`.

Here's a concrete example:

**Setup**: TP=2, CP=2, so `align_size = TP * CP * 2 = 8`.

A mini-batch has 3 micro-batches:
- micro-batch 0: samples with seqlen [100, 90]
- micro-batch 1: samples with seqlen [105, 88]
- micro-batch 2: samples with seqlen [95, 92]

**Student path** (in the engine, before forward):

When `pad_bshd_to_minibatch_max=True`, the engine scans the entire mini-batch for the global max:
```
global_max_seqlen = max(100, 90, 105, 88, 95, 92) = 105
```
Each micro-batch's `preprocess_bshd_engine` receives `forced_max_seqlen=105`:
```
align to 8: pad = (8 - 105%8) % 8 = 7 → max_seqlen = 112
local_seqlen = 112 / cp_size = 56
→ student_logits.shape = (bsz, 56, vocab/tp)
```

**Teacher path** (in `compute_forward_kl_topk`, without this fix):

Teacher's `preprocess_bshd_engine` is called without `forced_max_seqlen`, so it computes from the micro-batch's own offsets:
```
max_seqlen = max(100, 90) = 100  ← only this micro-batch
align to 8: pad = (8 - 100%8) % 8 = 4 → max_seqlen = 104
local_seqlen = 104 / 2 = 52
→ teacher.shape = (bsz, 52, topk)
```

**Result**: `student (bsz, 56) != teacher (bsz, 52)` → AssertionError.

**With this fix**: `forced_max_seqlen = 56 * 2 = 112`, teacher aligns to the same total → `local_seqlen = 56` ✓.

The alternative would be passing the engine's `global_max_seqlen` through the data pipeline as metadata to the loss function, but that's more invasive. Deriving it from the already-computed student logits is minimal and safe — `student_logits.shape[1] * cp_size` is guaranteed to be the aligned total that was actually used.

gaohongkui:
Closing this as I'm no longer able to follow up on it. The underlying mismatch (teacher/student seqlen diverging under CP when `pad_bshd_to_minibatch_max=True` applies alignment padding in `preprocess_bshd_engine`) may still be worth addressing, so please feel free to pick it up. Thanks for the earlier review.

PR #7013, opened 2026-07-10
[trainer, ckpt] fix: persist adaptive KL controller state across checkpoint resume

### LINKED ISSUE #2653: [Bug] Save & Load kl_controller when `kl_ctrl.type == "adaptive"`
### Problem
When using `kl_ctrl.type == "adaptive"`, the `AdaptiveKLController` state is not saved during checkpointing. When loading checkpoints, the controller resets to initial state instead of resuming from where it left off.

### Impact
- Adaptive KL adjustments are lost between training sessions
- Training restarts KL adaptation process instead of resuming

Epochex:
Fixes #2653

## Summary

`AdaptiveKLController.value` evolves during PPO training but was not included in trainer checkpoints. Resuming therefore restored the configured initial coefficient instead of the learned value, changing subsequent KL-penalized rewards.

This change adds a small versioned, tensor-only `kl_ctrl.pt` state file and wires it into the legacy, v1, and fully async trainer checkpoint paths. Writes use same-directory temporary files plus atomic replacement. Loads use `weights_only=True` on CPU and validate the schema, version, scalar shape, and coefficient value before updating the controller.

Checkpoints created before this change remain loadable: if `kl_ctrl.pt` is absent, the configured initial value is retained and a warning is logged.

## Prior work and scope

Closed/repurposed PR [#5766](https://github.com/verl-project/verl/pull/5766) briefly contained an unmerged legacy-only implementation. Its follow-up commit [`32fec0b`](https://github.com/dubin555/verl/commit/32fec0b3391a8320e32a32d39e2e62436b0acc83) is credited for the tensor/`weights_only=True` direction.

This PR covers the current legacy, v1, and fully async checkpoint owners through `AdaptiveKLController.save()` and `AdaptiveKLController.load()`, adds atomic persistence and strict validation, preserves compatibility with older checkpoints, and exercises the actual trainer save/load methods on CPU.

## Review follow-up

Automated review identified that the V1 test's temporary `transfer_queue` stub could pollute the parent pytest process. A narrow `sys.modules` cleanup still left parent-package attributes and transitive V1 modules referencing the stub, so the final implementation moves the V1 import and real checkpoint round trip into a child Python process. The parent process snapshots its complete relevant module map and verifies that it is unchanged after the child exits.

A maintainer follow-up moved checkpoint I/O onto `AdaptiveKLController` itself and removed the standalone helper module; all three trainer paths now call the controller methods directly.

## Tests

- `python -m pytest -q tests/trainer/ppo/test_adaptive_kl_checkpoint_on_cpu.py`: 12 passed; repeated, including a `-W error` run.
- Ruff check and format check on all changed Python files: passed.
- Changed-file `compileall`: passed.
- Fresh and preloaded module-cache adversarial checks: passed.

This change was developed with AI assistance.


tests/trainer/ppo/test_adaptive_kl_checkpoint_on_cpu.py
```diff
@@ -0,0 +1,242 @@
...
+
+    transfer_queue.KVBatchMeta = KVBatchMeta
+    for name in ("init", "kv_clear", "kv_batch_put", "kv_batch_get", "kv_list"):
+        setattr(transfer_queue, name, lambda *args, **kwargs: None)
+    transfer_queue.async_kv_put = lambda *args, **kwargs: None
+    transfer_queue.async_kv_batch_put = lambda *args, **kwargs: None
+    sys.modules.setdefault("transfer_queue", transfer_queue)
+    from verl.trainer.ppo.v1.trainer_base import PPOTrainer
+
+    return PPOTrainer
```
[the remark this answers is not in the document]
Epochex: Fixed in 1c978c7f: `_load_v1_trainer_base()` now removes the temporary `transfer_queue` mock and uncaches `verl.trainer.ppo.v1.trainer_base` when this helper introduced them, so the helper remains isolated even if it is called in-process later.

Checked locally:
- `py -3.12 -m compileall -q tests/trainer/ppo/test_adaptive_kl_checkpoint_on_cpu.py`
- `uvx ruff check tests/trainer/ppo/test_adaptive_kl_checkpoint_on_cpu.py`
- `git diff --check`

I also tried the targeted pytest command, but this local Windows Python cannot collect the test because `torch` is not installed in the available interpreter.

Epochex:
Hi maintainers, this PR is ready for CI. I cannot access the Slack workspace with my email, and the Feishu invitation is currently unreachable on my network, so I'm requesting here instead.

The targeted CPU checkpoint regression suite passes (12 tests), along with Ruff, format, compileall, and the module-cache isolation checks. Could you please trigger or approve the relevant CI workflows? Thanks.

Epochex:
Follow-up pushed in 747ce831 to address the CPU suite failure.

The failing assertions showed that the missing-state behavior itself was correct: the loader returned `False` and preserved the configured coefficient. Only the `caplog.text` assertion failed because the full suite's logging state left the capture buffer empty.

The test now patches `checkpoint_utils.logger.warning` directly and verifies the exact message template and missing `kl_ctrl.pt` path. This keeps the compatibility warning covered without depending on global pytest logging configuration.

Validation:
- focused test with the pytest logging plugin disabled: 1 passed
- full adaptive-KL checkpoint regression file: 12 passed
- CI-shaped CPU run: this checkpoint file passed in-suite; the local Windows run later stopped in an unrelated TransferQueue thread timeout after 332 total passes
- Ruff check/format, compileall, and `git diff --check`: passed

The new head currently has no checks attached. Could you please approve or trigger CI again? Thanks.

verl/trainer/ppo/checkpoint_utils.py
```diff
@@ -0,0 +1,113 @@
...
+
+KL_CONTROLLER_STATE_FILENAME = "kl_ctrl.pt"
+KL_CONTROLLER_STATE_VERSION = 1
+
+
+def _state_path(checkpoint_dir: str | os.PathLike) -> Path:
+    return Path(checkpoint_dir) / KL_CONTROLLER_STATE_FILENAME
+
+
+def save_adaptive_kl_controller_state(controller: AdaptiveKLController | None, checkpoint_dir: str | os.PathLike):
```
wuxibin89: Better add save/load function to `AdaptiveKLController`

Epochex: Good point, I have updated in d1fcb97b. `AdaptiveKLController` now owns `save()` / `load()`, and the standalone `checkpoint_utils.py` has been removed. The legacy, V1, and fully async trainer paths call those methods only for adaptive controllers; the checkpoint format, validation, atomic write, and old-checkpoint fallback are unchanged.

The focused CPU suite passes with `-W error` (12 tests), along with Ruff, formatting, mypy, compileall, and `git diff --check`.

Thanks bro :)

Epochex:
Quick update: I rebased this branch onto the current main and resolved the checkpoint conflict in 90867ec6. The V1 path keeps the new upstream TransferQueue recovery logic, with adaptive-KL state restored alongside it.

Post-rebase validation is green: 12 targeted tests passed with warnings treated as errors, plus Ruff check/format, mypy, compileall, and git diff --check. The refreshed head is mergeable now, but GitHub has not attached any checks yet. Could you please trigger or approve CI when convenient? Thanks!

PR #7037, opened 2026-07-14
[trainer, ckpt] feat: support streaming dataloader and async trainer checkpoint recovery

Begunner:
### What does this PR do?

This PR 
1. supports adding arbitrary number of prompts from dataloader to generation without the `train_batch_size` limitation. 
2. improves v1 async trainers by modifying`drop` strategy: dropping stale prompt groups and refilling new prompts.
3. allows restarting pending/running prompts after resume while preserving finished samples using TransferQueue checkpointing.

### Test

E2E save/load checkpoint and drop strategy tests.
CPU unit tests.

### API and Usage Example

Set `data.gen_batch_size=1` to use streaming dataloader (auto set in async trainers using drop strategy)
Install TransferQueue>=0.1.9 to use `tq.save_checkpoint` in async trainers.


verl/trainer/ppo/v1/trainer_base.py
```diff
@@ -102,6 +103,11 @@ def apply_greedy_sampling_params(params: dict[str, Any]) -> None:
 logger.setLevel(os.getenv("VERL_LOGGING_LEVEL", "INFO"))
 
 
+def _tq_supports_checkpoint() -> bool:
```
wuxibin89, line 107: Better check TransferQueue version.

verl/trainer/ppo/v1/trainer_base.py
```diff
@@ -1158,15 +1240,53 @@ def _next_train_batch(self):
...
+                # TODO: maybe let workers do it?
+                fields=batch.select(*[key for key in batch.keys() if not isinstance(batch.get(key), NonTensorData)]),
+            )
+        else:
+            tq.kv_batch_put(keys=list(batch["uid"]), partition_id="train", tags=tags)
+
         self.agent_loop_manager.generate_sequences(batch)
+        return len(batch)
+
+    def _add_prompts_to_generate(self, num_prompts: int) -> int:
```
wuxibin89, line 1280: Why we need a separate `_add_prompts_to_generate` instead of `_add_batch_to_generate`?

verl/trainer/ppo/v1/trainer_base.py
```diff
@@ -716,6 +744,63 @@ def _load_checkpoint(self):
         else:
             logger.warning(f"Warning: No dataloader state found at {dataloader_local_path}, will start from scratch")
 
+        # 5. restore TransferQueue state (async modes). Re-issuing the restored in-flight prompts is
+        # deferred to fit() to use the agent_loop_manager.
+        if self.trainer_mode != "sync":
```
wuxibin89: It's safe to save/load checkpoint in sync mode.

verl/trainer/ppo/v1/trainer_base.py
```diff
@@ -716,6 +744,63 @@ def _load_checkpoint(self):
         else:
             logger.warning(f"Warning: No dataloader state found at {dataloader_local_path}, will start from scratch")
 
+        # 5. restore TransferQueue state (async modes). Re-issuing the restored in-flight prompts is
```
wuxibin89, line 756: Make L747-L802 2 pair functions:
- save_transfer_queue
- load_transfer_queue

verl/trainer/ppo/v1/trainer_base.py
Begunner: I add the function to avoid modifying `SkipManager` and make the function semantics clearer.

verl/trainer/ppo/v1/trainer_base.py
Begunner: In sync cases, maybe no inflight/finished sample needs to be saved. I think it is only for async trainers with warmup_batches.

verl/trainer/ppo/v1/trainer_base.py
Begunner: Get

verl/trainer/ppo/v1/trainer_base.py
Begunner: Get

wuxibin89 (approved):

savaresejeremy:
I built on this in an RFC I posted, #7579. The part I wanted to keep is the recovery here: finished samples come back from the TransferQueue checkpoint, and pending or running prompts get re-issued. The limit is that it can only start from the last checkpoint, so a crash late in a save interval means generating that whole interval again. My proposal keeps this logic and gives it a more recent starting point, with a metadata journal that writes every few seconds and saves the dataloader position alongside it.

Issue #5024, opened 2026-01-22
[Question] Video input causes IndexError in MRoPE position calculation with Qwen3-VL + vLLM


Solus-sano:
# [Question] Video input causes IndexError in MRoPE position calculation with Qwen3-VL + vLLM

## Environment
- vLLM version: 0.11.0
- Model: Qwen3-VL-8B-Instruct
- Framework: verl (GRPO trainer with agent loop)
- Video input: 30 frames per clip

## Problem Description

I'm encountering an `IndexError: list index out of range` when processing video inputs with Qwen3-VL model using vLLM. Image inputs work perfectly fine with the same setup.

### Error Traceback
```
File "vllm/model_executor/layers/rotary_embedding/mrope.py", line 652, in _qwen3vl_get_input_positions_tensor
    video_grid_thw[video_index][0],
    ~~~~~~~~~~~~~~^^^^^^^^^^^^^
IndexError: list index out of range
```

### 1. The Error Location

In `vllm/model_executor/layers/rotary_embedding/mrope.py`, the MRoPE position calculation:
- **Line 613-614**: Extracts `video_grid_thw` from multimodal kwargs → length is **15** (one entry per video frame, each being `[1, 24, 40]`)
- **Line 625-626**: Calculates `video_nums` by counting `<|video_pad|>` tokens in `prompt_ids` → result is **29**

When `video_index >= 15`, accessing `video_grid_thw[video_index]` causes the IndexError.

### 2. Why the Mismatch? (The Core Issue)

The problem is a mismatch between how `processor.apply_chat_template()` generates video placeholders and how vLLM's `_apply_matches()` replaces them.

#### Step 1: `processor.apply_chat_template()` Output

When processing a video with 30 frames (sampled to 15 frames), Qwen3-VL's `processor.apply_chat_template()` generates **per-frame timestamp placeholders**:

```
<0.5 seconds><|vision_start|><|video_pad|><|vision_end|>
<2.5 seconds><|vision_start|><|video_pad|><|vision_end|>
<4.5 seconds><|vision_start|><|video_pad|><|vision_end|>
... (15 such structures, one per frame)
```

This means there are **15** `<|vision_start|><|video_pad|><|vision_end|>` patterns in the prompt.

#### Step 2: vLLM's `_get_prompt_updates()` Expectation

In `vllm/model_executor/models/qwen3_vl.py:913-985`, the `_get_prompt_updates()` function defines how to replace video placeholders:

```python
PromptReplacement(
[...]

PR #7038, opened 2026-07-14
[rollout, vllm] fix: stop the policy from sampling vision placeholder tokens

### LINKED ISSUE #5817: [Bug] Full Async vLLM:Model generates video_token_id during RL rollout, causing vLLM MaskedScatter AICPU crash
### System Info

Version      : 3.10.12
Compiler     : GCC 11.2.0
Build        : ('main', 'Jul  5 2023 18:45:42')
Arch         : ('64bit', 'ELF')
Version      : 23.1.2
Directory    : /usr/local/miniconda3/lib/python3.10/site-packages/pip
vllm         : 0.11.0
sglang       : not found.
ray          : 2.50.0
torch        : 2.7.1+cpu
Version      : 0.8.0.dev
Directory    : xxx
Commit Hash  : c179476754150a5384f96d56b622a8f6330d2c04
----------Platform Info----------
Platform     : Linux-5.10.0.64kb.aarch64-aarch64-with-glibc2.28
system       : Linux
node         : magnus-zhangliangxu3-03345c66-master-0
release      : 5.10.0.64kb.aarch64
version      : #1 SMP PREEMPT Fri Aug 8 12:50:21 CST 2025
CUDA is not available.
Failed to execute nvidia-smi command.
CPU Memory      : 2960.00 GB
GPU Count       : 0

### Information

- [x] My own modified scripts

### Tasks

### Reproduction

Run verl full async RL training with a multimodal model (Qwen3VL or similar) using vLLM as the rollout backend. The crash occurs after ~1–10 hours of training and is non-deterministic.

**Root cause:** During decode, the model occasionally samples `video_token_id=151656`. Because no mechanism blocks this token in the rollout sampling params, it gets written into the KV cache history. On the next request, `input_ids` contains an extra video placeholder token, causing `merge_multimodal_embeddings` to see `mask_count > flattened_embedding_count`. The subsequent `masked_scatter_` call receives mismatched shapes and fails on the Ascend NPU stream.

Image

**Crash log (Python layer):**
```
ERROR [worker_base.py:275]
  File model_runner_v1.py, line 1467, in _prepare_inputs
    logits_indices = torch.from_numpy(cu_num_tokens - 1).to(
        self.device, non_blocking=True)

RuntimeError: The Inner error is reported as above.
  current copy params are srclen=128, dstlen=128, kind=1

Note: Since the operator is called asynchronously, the stacktrace
      may be inaccurate. ASCEND_LAUNCH_BLOCKING=1 needed for accurate trace.
```

**CANN plog:**
```
[device_error_proc.cc] serial number is 3, stream_id:47, task_id:14601,
                        errcode:21008, msg:inner error
[stream.cc]             stream_id=47 enter failure abort
[davinci_kernel_task.cc] kernelName=MaskedScatter, errorCode=0x2a
[davinci_kernel_task.cc] fault op_name=
```

**Mismatch evidence** (added debug print in `merge_multimodal_embeddings`):
```
[MISMATCH] flattened=3558, mask=3559
  matched token ids: [(151655, 3558), (151656, 1)]
```

**Sampling evidence** (added debug print in `Sampler`):
```
[WARN] video_token generated! positions=[[37]]
logits[151656]=[0.138..., 0.164..., 0.312..., ...]  ← not -inf, not blocked
```

**Temporary workaround** applied in `verl/workers/rollout/vllm_rollout/vllm_rollout.py`:
```python
# Block image_token and video_token to prevent model from generating
# multimodal special tokens during RL rollout
image_token = self.model_config.processor.image_token
video_token = self.model_config.processor.video_token
sampling_params['bad_words'] = (
    sampling_params.get('bad_words', []) + [image_token, video_token]
)
```

This workaround has known limitations: it introduces train/rollout inconsistency (the training engine has no equivalent filter), and does not fix the underlying issue.

---

### Expected behavior

The rollout engine should prevent multimodal special tokens (`image_token_id`, `video_token_id`) from being sampled during the decode phase of RL rollout. When a multimodal model is used in a text-generation RL setting, these tokens should be masked out (logits set to `-inf`) in the sampling step, so that `merge_multimodal_embeddings` always receives a consistent number of placeholders and embeddings.

Ideally, verl should expose a config option (e.g. `rollout.bad_words`) or automatically detect and block special vision tokens when no visual inputs are present in the current decode step, so that downstream `masked_scatter_` operations never receive mismatched shapes.

zhshj0110:
### What does this PR do?

Stops the policy from sampling `<|image_pad|>` / `<|video_pad|>` during vLLM rollout.

A vision placeholder token only means something when a real image or video sits behind it: the k-th run of placeholders pairs with the k-th row of `image_grid_thw` / `video_grid_thw`. Nothing stops the policy from sampling one anyway — it is an ordinary entry of the vocabulary, and nothing masked it — and then that pairing silently breaks.

Every consumer of the sequence trusts the pairing, so each one has its own way of dying:

| where | how it dies |
|---|---|
| `get_rope_index` (transformers) | takes one grid per span, so an extra **image** span exhausts the iterator → `StopIteration`; a **video** span indexes a modality that was never populated → `TypeError: 'NoneType' object is not an iterator` |
| `merge_multimodal_embeddings` (vLLM) | scatters into a mask that no longer matches the embedding count → `masked_scatter_` shape mismatch |

Either one takes the whole training job down, after hours of training, non-deterministically.

Fixes #5817 (the `masked_scatter` crash on NPU — the reporter confirmed with a Sampler print that `logits[151656]` is *not* `-inf`, i.e. nothing blocks the token).

Found by an 8-hour Qwen3-VL-8B GRPO run that died at step 44 when the policy sampled `<|video_pad|>` (id 151656).

**Note for the reporter of #5817:** the workaround in that issue patches `vllm_rollout.py`, but on `main` that file is now a client-side `ServerAdapter` and no longer builds `SamplingParams`, so that patch would no longer land anywhere. The ban now happens once at engine setup in `vLLMHttpServer.run_server`, which covers every request the engine serves.

### Scope

This fixes **vLLM only**. `AgentLoopWorker._compute_position_ids` is shared by every backend, but the ban is installed in the vLLM engine: sglang plumbs a `custom_logit_processor` that verl never sets, and trtllm has no equivalent, so on those two the policy can still sample a placeholder and hit the same crash. Banning at each engine seems right to me, but it does mean those backends need the same treatment in a follow-up.

This PR does **not** cover #5024, despite an earlier claim in this description that it did — that crash is inside vLLM's own `mrope.py` on the **prompt** (grid rows disagreeing with the placeholder count the processor emitted), which is a separate bug in the processor↔vLLM placeholder-expansion contract.

### Checklist Before Starting

- [x] Search for similar PRs: none address this. #4342 (loading sampling params from config) is unrelated and does not interact — this PR touches no sampling-params key.

### Test

One CPU unit test file, no model download, under the `*_on_cpu.py` pattern the CPU workflow collects — `tests/workers/rollout/test_vision_placeholder_tokens_on_cpu.py`, 7 cases in two groups:

- `TestGetVisionPlaceholderTokenIds` — the ids resolve from either `image_token_id` ints or `image_token` strings; an image-only processor bans only the image placeholder; a text-only model gets `[]`.
- `TestMonkeyPatchComputeLogits` — the banned ids **and** the OOV tail come back `-inf` while every legal token keeps the value the model produced; with no ids passed (text-only, and every pre-existing caller of `monkey_patch_model`) only the OOV tail is masked, so sampling is bit-for-bit what it was before.

The mask test is load-bearing rather than decorative: deleting the mask line turns `test_banned_tokens_and_oov_tail_are_masked` red.

Additionally verified against the **real Qwen3-VL-8B processor**:

```
get_vision_placeholder_token_ids(processor) -> [151655, 151656]  # ['<|image_pad|>', '<|video_pad|>']
get_vision_placeholder_token_ids(None)      -> []
```

and the original crash was reproduced end-to-end with a real processor and a real image before the fix:

```
response with <|video_pad|>   -> TypeError: 'NoneType' object is not an iterator   (the exact crash)
response with <|image_pad|>   -> StopIteration
```

### API and Usage Example

No API change, no new config key: the placeholder ids are derived from the processor, so multimodal runs are protected automatically and text-only runs are unaffected.

```python
# verl/workers/rollout/utils.py
ids = get_vision_placeholder_token_ids(processor)   # [151655, 151656] — or [] for a text-only model
```

The maintainer comment on #5817 suggested a dedicated `rollout.block_multimodal_tokens` switch. I left it out deliberately: a new config key is not needed for correctness here, and adding one to the shared trainer config breaks every launcher that already appends the same key with Hydra's `+`. Happy to add the switch if you'd rather have it explicit.

### Design & Code Changes

1. **`verl/workers/rollout/vllm_rollout/utils.py`** — `monkey_patch_compute_logits` takes `banned_token_ids` and masks them to `-inf` alongside the OOV tail it already masks. An ungrounded `<|image_pad|>` is illegal in exactly the way an out-of-vocabulary token is, so it belongs in the mask that is already running there rather than arriving as a new mechanism. `banned_token_ids` defaults to `None`, so every existing caller of `monkey_patch_model` samples exactly as before.
2. **`verl/workers/rollout/vllm_rollout/vllm_async_server.py`** — `run_server` passes the ids down the `monkey_patch_model` collective_rpc it already issues. Banning at the engine rather than per request means no caller's `sampling_params` can switch the guard off, and the MTP drafter is covered too (`monkey_patch_model` walks `_iter_all_models()`).
3. **`verl/workers/rollout/utils.py`** — `get_vision_placeholder_token_ids(processor)`, next to `qwen2_5_vl_dedup_image_tokens`, which already deals with the same "placeholder count doesn't line up" family of problems.

### Known cost

Masking those two logits to `-inf` means the rollout distribution is no longer identical to the actor's, so the importance ratio is computed against a slightly different distribution. In our run those tokens carry ~1e-6 probability, orders of magnitude below the `rollout_corr/kl ≈ 1.5e-3` we already measure — but it is a real approximation, not a free lunch.


zhshj0110:
Good catch on the `setdefault` — fixed in c989e3f.

You are right that it defeats the guard: a caller who passes `bad_words` for their own reasons would silently turn off the vision-token ban, and the job would die hours later. Banning some other word is not a request to allow `<|image_pad|>`/`<|video_pad|>` through, so the placeholders are now merged into whatever the caller passed (order preserved, no duplicates).

I also moved the merge into `add_vision_bad_words()` in `verl/workers/rollout/utils.py`, next to the token lookup. That is part of why the bug got in: buried inside `vLLMHttpServer.generate` it was out of reach of the CPU tests, which only covered the token lookup. It is covered now — reverting to `setdefault` turns two of the new cases red:

```
FAILED test_placeholders_are_merged_into_the_callers_bad_words
FAILED test_merging_does_not_duplicate_placeholders
```

`tests/workers/rollout/test_vision_placeholder_tokens_on_cpu.py` gains: merge into an existing `bad_words`, no duplication when a placeholder is already banned, and a text-only model still ends up with no `bad_words` key at all. 14 CPU tests pass, ruff and ruff-format clean.

verl/workers/rollout/vllm_rollout/vllm_async_server.py
```diff
@@ -592,6 +597,10 @@ async def generate(
             extra_args["kv_transfer_params"] = kv_transfer_params
             sampling_params["extra_args"] = extra_args
 
+        # A sampled <|image_pad|>/<|video_pad|> has no image behind it, and every consumer of the
+        # sequence assumes it does.
+        add_vision_bad_words(sampling_params, self.model_config.processor)
```
wuxibin89: An alternative way is to fill <|image_pad|>/<|video_pad|> logits to `-inf`:
https://github.com/verl-project/verl/blob/release/v0.6.1/verl/workers/rollout/vllm_rollout/vllm_rollout_spmd.py#L499-L512

zhshj0110: Adopted in 4e4b81d — thanks, this is the better place for it.

`monkey_patch_compute_logits` already exists to stop the sampler picking a token it must never
pick; an ungrounded `<|image_pad|>` is illegal in exactly the way an OOV token is, so it slots
into the mask that is already running there rather than arriving as a new mechanism. Coverage is
the same either way (`monkey_patch_model` is collective_rpc'd from `run_server`, and `generate`
where I had the `bad_words` lives in the same `vLLMHttpServer`), but the logits mask is strictly
stronger in two ways:

- **It cannot be bypassed.** `bad_words` rode on the `sampling_params` a caller hands in, which is
  precisely what the other review comment caught: a caller who passed their own `bad_words` would
  silently disable the guard, and the job would die hours later. That whole class of bug is gone
  now, and it no longer collides with #4342 (sampling params from config).
- **It covers the MTP drafter**, since `monkey_patch_model` walks `_iter_all_models()`.

`banned_token_ids` defaults to `None`, so text-only models and every other caller of
`monkey_patch_model` sample exactly as they did before.

Verified:

- 13 CPU tests pass. The lookup now returns ids (4 cases) and the mask itself is covered (3 cases);
  removing the mask line turns `test_banned_tokens_and_oov_tail_are_masked` red, so the test is
  load-bearing rather than decorative. The agent-loop guard for tokens that slip through is
  unchanged (6 cases).
- Against a real Qwen3-VL-8B processor, `get_vision_placeholder_token_ids` resolves
  `[151655, 151656]` → `['<|image_pad|>', '<|video_pad|>']`, and `None` for a text-only model.

The cost is unchanged from the `bad_words` version and I want to keep it stated plainly: `-inf` on
those two logits means the rollout distribution is no longer identical to the actor's, so the
importance ratio is computed against a slightly different distribution. In our run those tokens
carry ~1e-6 probability, which is orders of magnitude below the `rollout_corr/kl ≈ 1.5e-3` we
already measure — but it is a real approximation, not a free lunch.

verl/workers/rollout/vllm_rollout/vllm_async_server.py
```diff
@@ -592,6 +597,12 @@ async def generate(
             extra_args["kv_transfer_params"] = kv_transfer_params
             sampling_params["extra_args"] = extra_args
 
+        # A sampled <|image_pad|>/<|video_pad|> has no image behind it, and every consumer of the
+        # sequence assumes it does. setdefault, so an explicitly passed bad_words still wins.
+        vision_bad_words = get_vision_placeholder_tokens(self.model_config.processor)
+        if vision_bad_words:
+            sampling_params.setdefault("bad_words", vision_bad_words)
```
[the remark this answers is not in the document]
zhshj0110: Superseded by 4e4b81d. You were right that `setdefault` let a caller's own `bad_words` silently
disable the guard; per maintainer review the ban now happens at the engine — the vision placeholder
ids are masked to `-inf` in `compute_logits`, alongside the OOV mask that already lives there — so
the `bad_words` path is removed entirely and there is no longer a sampling-params key for a caller
to collide with.

verl/experimental/agent_loop/agent_loop.py
```diff
@@ -897,6 +897,52 @@ def _compute_multi_modal_inputs(self, output, input_ids) -> dict[str, torch.Tens
             multi_modal_inputs["images_seqlens"] = images_seqlens
         return multi_modal_inputs
 
+    @staticmethod
+    def _drop_ungrounded_vision_tokens(mm_token_type_ids, image_grid_thw, video_grid_thw) -> None:
```
wuxibin89: Do we still need this? If not, clean up unit test as well.

zhshj0110: Removed in 484a104, unit tests cleaned up — `agent_loop.py` is back to upstream verbatim.

Two things worth flagging before you merge, since both cut against what I originally wrote:

**1. This narrows the fix to vLLM.** `AgentLoopWorker._compute_position_ids` is shared by every
backend (SPMD is retired, so all rollout goes through the agent loop), but the ban is not: it is
installed in `vLLMHttpServer.run_server` only. sglang plumbs a `custom_logit_processor` that verl
never sets (`sglang_rollout/http_server_engine.py:433`), and trtllm has no equivalent at all — its
sampling params come straight from the agent loop, which carries no bad-words field. So on those
two backends the policy can still sample `<|image_pad|>`/`<|video_pad|>` and take `get_rope_index`
down exactly as before. That is fine by me as a scope call — I would rather ban it at each engine
than paper over it in shared postprocessing — but it means those backends stay broken until someone
does the same there, and I would rather say so than have it discovered later.

**2. Correcting myself: this PR does not cover #5024.** I claimed it did, and I was wrong; I have
removed the claim from the description. #5024 crashes inside vLLM's own `mrope.py:652` on the
**prompt** (15 `video_grid_thw` rows vs 29 `<|video_pad|>` tokens the processor emitted), which
never reaches verl's `_compute_position_ids`. And the guard I just deleted counted placeholder
*spans* against grid *rows*, so it would not have caught a within-span length mismatch anyway. It
is a separate bug in the processor↔vLLM placeholder-expansion contract.

#5817 is still fixed by the logits mask — that one is a sampled token, which is exactly what is
banned now.

tests/workers/rollout/test_vision_placeholder_tokens_on_cpu.py
```diff
@@ -0,0 +1,51 @@
...
+from __future__ import annotations
+
+from types import SimpleNamespace
+
+from verl.workers.rollout.utils import get_vision_placeholder_token_ids
+
+TOKENS = {151655: "<|image_pad|>", 151656: "<|video_pad|>"}
+
+
+def make_processor(**attrs):
```
wuxibin89, line 33: Please merge these 2 unit test files.

zhshj0110: Merged in b83ac12 — one file now, `tests/workers/rollout/test_vision_placeholder_tokens_on_cpu.py`,
with the 7 cases grouped into `TestGetVisionPlaceholderTokenIds` (which ids to ban) and
`TestMonkeyPatchComputeLogits` (where they get masked). `test_vision_logits_mask_on_cpu.py` is gone.

Still load-bearing after the merge: dropping the mask line from `monkey_patch_compute_logits` turns
`TestMonkeyPatchComputeLogits::test_banned_tokens_and_oov_tail_are_masked` red.

The PR is now 3 source files + 1 test file.

wuxibin89 (approved):

PR #6409, opened 2026-05-20
[ckpt, model] fix: save LoRA train metadata for PPO actor checkpoint

Yatogaii:
### What does this PR do?

Fixes #6380.

This PR persists `lora_train_meta.json` for PPO actor checkpoints when LoRA is enabled, so that `model_merger` can correctly recover LoRA metadata when exporting PEFT adapters.

PR #5326 previously added `lora_train_meta.json` for LoRA adapter export, allowing `model_merger` to recover fields such as `lora_alpha` and `task_type` when writing PEFT `adapter_config.json`. However, that fix only covered SFT checkpoints.

PPO actor checkpoints use a different layout:

```text
global_step_x/actor
```

and PPO does not save actor checkpoints through `CheckpointHandler`. As a result, PPO LoRA checkpoints could still be merged into PEFT adapters with missing or incorrect metadata. In particular, `lora_alpha` could become `0`, making the exported adapter appear ineffective after loading.

This PR writes:

```text
global_step_x/actor/lora_train_meta.json
```

after PPO actor checkpoint saving completes when LoRA is enabled. This matches the existing PPO merge workflow, where `model_merger` receives the actor checkpoint directory via `--local_dir`.

The metadata extraction supports both PPO LoRA config shapes:

- FSDP: `actor_rollout_ref.model.lora_rank` / `actor_rollout_ref.model.lora_alpha`
- Megatron: `actor_rollout_ref.model.lora.rank` / `actor_rollout_ref.model.lora.alpha`

AI assistance was used while preparing this PR. The changed lines were reviewed by the human submitter.

### Checklist Before Starting

- [x] Search for similar PRs. Paste at least one query link here: https://github.com/verl-project/verl/pulls?q=is%3Apr+is%3Aopen+lora_train_meta+PPO+actor+checkpoint
  - Related prior fix: #5326
  - This PR does not duplicate #5326, because #5326 only saved LoRA metadata for SFT checkpoints and did not cover the PPO actor checkpoint layout.
  - #6381 appears to address the same issue.
- [x] Format the PR title as `[{modules}] {type}: {description}`. This will be checked by the CI.

### Test

Static checks:

```bash
python3 -m py_compile verl/trainer/ppo/ray_trainer.py verl/trainer/main_ppo_sync.py
git diff --check
```

Manual validation:

- Created a synthetic FSDP PPO actor checkpoint under `global_step_1/actor`.
- Added LoRA A/B tensors and `actor/lora_train_meta.json`.
- Ran `model_merger`.
[...]

PR #7068, opened 2026-07-16
[ckpt, model] fix: prevent key corruption during HuggingFace export

rongkunxue:
### What does this PR do?

Prevents HuggingFace checkpoint key corruption when verl saves an externally supplied full state dict.

For example, Transformers v5.2.0 defines `save_original_format=True` by default in `PreTrainedModel.save_pretrained` ([source](https://github.com/huggingface/transformers/blob/7d9754a05193eb79b1d86aa744b622b8068008cd/src/transformers/modeling_utils.py#L3125-L3136)). When enabled, Transformers applies `revert_weight_conversion` to the state dict before saving ([source](https://github.com/huggingface/transformers/blob/7d9754a05193eb79b1d86aa744b622b8068008cd/src/transformers/modeling_utils.py#L3309-L3312)).

verl already assembles the exported state dict using the current model key format. Applying this reverse conversion again can incorrectly rename keys for composite VLMs such as Qwen3.5, producing keys such as:

```text
model.language_model.language_model.language_model.*
model.language_model.visual.*
```

The checkpoint is written successfully, but downstream loaders may report missing or uninitialized weights.

This PR adds a small compatibility helper that passes `save_original_format=False` only when the installed Transformers version supports the argument. It is used by both the offline model merger and the FSDP HuggingFace checkpoint exporter. Older Transformers versions remain unchanged.

This is independent of #5326 and #6409, which preserve PEFT/LoRA training metadata during adapter export. It is also separate from #5599, which addresses Megatron Qwen3.5 LoRA and MTP weight synchronization rather than HuggingFace `save_pretrained` serialization.

### Checklist Before Starting

- [x] Search for similar PRs:
  - https://github.com/verl-project/verl/pulls?q=is%3Apr+is%3Aopen+save_original_format
  - https://github.com/verl-project/verl/pulls?q=is%3Apr+is%3Aopen+HuggingFace+state_dict+export
  - https://github.com/verl-project/verl/pulls?q=is%3Apr+is%3Aopen+Qwen3.5+checkpoint+key+corruption
- [x] Format the PR title as `[ckpt, model] fix: prevent key corruption during HuggingFace export`.

### API and Usage Example

No public API, CLI, or checkpoint format changes are introduced. Existing usage remains unchanged:

```bash
python -m verl.model_merger merge \
  --backend fsdp \
  --local_dir /path/to/global_step_x/actor \
  --target_dir /path/to/hf_export
```

### Design & Code Changes

- Add a version-compatible helper in `transformers_compat.py`.
- Inspect the `save_pretrained` signature and pass `save_original_format=False` only when supported.
- Use the helper in the offline model merger and FSDP HuggingFace exporter.
- Keep the existing state-dict assembly, LoRA metadata, and checkpoint logic unchanged.
- Use Python's standard-library `inspect` module, with no new dependency.

Validation: `PATH="$PWD/.venv/bin:$PATH" .venv/bin/pre-commit run --all-files --show-diff-on-failure --color=never` and `git diff --check` both passed.

AI assistance was used for implementation and PR preparation. The human submitter will review every changed line and remains responsible for the final change before requesting maintainer review.

### DIFF
--- a/verl/model_merger/base_model_merger.py
+++ b/verl/model_merger/base_model_merger.py
@@ -24,7 +24,7 @@ from accelerate import init_empty_weights
 from transformers import AutoConfig, AutoModelForCausalLM, AutoModelForTokenClassification, GenerationConfig
 
 from verl.utils import hf_processor, hf_tokenizer
-from verl.utils.transformers_compat import drop_tied_target_keys, get_auto_model_for_vision2seq
+from verl.utils.transformers_compat import drop_tied_target_keys, get_auto_model_for_vision2seq, get_hf_save_kwargs
 
 AutoModelForVision2Seq = get_auto_model_for_vision2seq()
 
@@ -403,7 +403,7 @@ class BaseModelMerger(ABC):
         drop_tied_target_keys(state_dict, model, self.model_config)
 
         print(f"Saving model to {self.config.target_dir}")
-        model.save_pretrained(self.config.target_dir, state_dict=state_dict)
+        model.save_pretrained(self.config.target_dir, **get_hf_save_kwargs(model, state_dict))
         del state_dict
         del model
 
--- a/verl/utils/checkpoint/fsdp_checkpoint_manager.py
+++ b/verl/utils/checkpoint/fsdp_checkpoint_manager.py
@@ -32,7 +32,7 @@ from verl.utils.device import is_cuda_available
 from verl.utils.fs import copy_to_local, is_non_local, local_mkdir_safe
 from verl.utils.fsdp_utils import fsdp_version, get_fsdp_full_state_dict, get_fsdp_state_ctx
 from verl.utils.logger import log_with_rank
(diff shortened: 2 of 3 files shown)


rongkunxue:
> [!NOTE]
> This behavior was introduced in Hugging Face Transformers by [commit `d08b98b`](https://github.com/huggingface/transformers/commit/d08b98b965176ea9cf8c8e8b24995c955b7e2ec9) and shipped with **v5.0.0**. It remains present in subsequent 5.x releases, including [v5.3.0](https://github.com/huggingface/transformers/blob/aad13b87ed59f2afcfaebc985f403301887a35fc/src/transformers/modeling_utils.py#L3185-L3194), which is pinned in verl’s [stable TRT-LLM Docker image](https://github.com/verl-project/verl/blob/45ab5b820e750b4261d199399a0cbddf38bd0fc3/docker/Dockerfile.stable.trtllm#L37) and other newer Docker images.

Issue #7060, opened 2026-07-16
[Tracking] Sharded delta weight sync (delta_sharded): roadmap & known issues


ChangyiYang:
Tracking issue for the sharded delta weight sync (`delta_sharded` checkpoint-engine backend) merged in #6974, and the follow-up roadmap.

## Merged

- #6974 — `[ckpt,rollout] feat: sharded delta weight sync over NCCL for disaggregated rollout` (merged 2026-07-16, `903d90cc`)
  - Each trainer rank byte-diffs its **local shard** against a pinned-CPU snapshot; only changed `(position, value)` pairs are gathered to rank 0 and streamed to the rollout side. No rank ever stages a full-model mirror.
  - Shard layout is described declaratively via `ShardSpec` (`DeviceMesh` + `Placement`, `verl/workers/engine/spec.py`) — shared contract with the nccl-m2n work.
  - Measured steady-state sync vs `nccl` full broadcast (disaggregated, cross-node, offload=True): 7B **3.8s vs 9.1s (2.4×)**, 32B **12.5s vs 23.2s (1.9×)**, 72B **12.0s vs 36.9s (3.1×)**; 200-step GRPO equivalence run and independent re-measurement in the PR thread.

- #7144 — `ShardSpec` generalization: `BlockPlacement` block descriptors (any `Shard(k)` / multi-dim cuts) + explicit-place channel for hybrid geometries (merged 2026-07-27, `b9873ee9`)
- #7085 — veomni FSDP+EP shard export: fused expert stacks via a dim-0-separable converter (`to_hf_chunk` + handler-probed `hf_slots`), HSDP replica dedup (`ShardSpec.contributes`, covers newer VeOmni's `(ep_replicate, ep_fsdp, ep)` expert mesh), ep_size=1 degenerate geometry (merged 2026-07-29, `ce195c95`). Bit-level idempotence verify sweeps green at 30B (EP8 / EP1 / HSDP) and on VeOmni main.

## Next steps

[...]

PR #7080, opened 2026-07-16
[ckpt,trainer] feat: block-placement sharded delta sync + Automodel FSDP2+EP support

ChangyiYang:
## What does this PR do?

Extends the sharded delta weight sync (`delta_sharded`, #6974) beyond the flat FSDP2 `Shard(0)` fast path to **block placements** — `Shard(k)` and multi-Shard-dim meshes — and wires the **Automodel (nemo_automodel) engine with expert parallelism** into it end to end. Item 1 of the roadmap in #7060, and the `Shard(1)` support requested in the #6974 review.

### Engine: block placement

- `ShardSpec` gains `BlockPlacement` (`local_shape`, `global_offset`): a rank's shard as one hyper-rectangular block of the full tensor. `translate_flat_indices` maps shard-local flat positions to full-tensor flat positions (mixed-radix decompose + per-dim offset + recompose); the existing `int` flat-offset fast path is untouched, so FSDP2 `Shard(0)` behavior is bit-identical.
- `derive_placement`: `Shard(k!=0)` and multi-Shard-dim placements (e.g. an `(ep, ep_shard)` expert mesh with `(Shard(0), Shard(1))`) return a `BlockPlacement`; multi-Shard meshes gather over one flattened whole-mesh group (created once, cached). `_StridedShard` and Replicate-alongside-multi-Shard are rejected loudly.
- Dense first sync places per-rank blocks via `gather_dense_blocks_to_rank0` (fixed-size block-meta all_gather + padded gather + rank-0 slice assignment).
- The converter (`to_hf`) profile gains a block variant: rebuild the **full** logical tensor with NaN sentinels (per-rank block slice-assign), then hand `[full]` to `to_hf`. The Megatron-style shard-list contract is unchanged.

### Automodel + EP

- `AutomodelEngine.get_per_tensor_param_shard`: dense params pass their DTensor placement through verbatim; grouped experts (`experts.gate_and_up_projs` / `experts.down_projs` on the `(ep, ep_shard)` moe mesh) carry a `to_hf` closure over the model's HF state-dict adapter — a pure permutation (per-expert slice + gate/up split + transpose), so the NaN-sentinel rebuild emits classic per-expert HF names that SGLang's loader consumes directly. The `ep_shard == 1` degenerate case (experts split only across `ep`) takes the flat `Shard(0)` profile; the closure accepts both.
- PPO wiring: `model_engine=automodel` hydra group (actor/ref/critic configs mirroring the veomni wiring; the SFT-side engine and optim configs are reused). Previously automodel was SFT-only.
- Sync-free masked apply in the SGLang delta loader: the previous masked `copy_` hook did a `bool(mask.all())` early-out plus boolean indexing **per parameter** — each a device→host sync. MoE flushes carry >10k per-expert entries (vs a few hundred for dense models), which pushed a single flush apply past the server adapter's HTTP timeout. Rewritten as an on-stream `torch.where` copy.

## Test

- CPU: `tests/checkpoint_engine/test_block_placement.py` — translate math against full-tensor coordinates (incl. `(8,4,6)` split ep=4 × shard=2, `Shard(1)`, odd sizes), and the block NaN rebuild against a qwen3-moe-style conversion (slice+split+transpose), bit-exact. Existing checkpoint-engine suites pass (18 total).
- E2E: Qwen3-30B-A3B-Instruct, automodel `ep_size=4` (moe mesh `(ep_shard, ep)=(2,4)`, placements `(Shard(1), Shard(0))`) + FSDP2 dense params, SGLang rollout TP4, 2×8 GPU disaggregated, `delta_sharded`: 3/3 GRPO steps green, steady-state `update_weights` 9.6–12.3 s, `changed_ratio` 0.15–0.5%, no HTTP retries. (`ep_size=8` / `ep_shard=1` degenerate profile also exercised.)
- Regression: 0.5B FSDP2 `delta_sharded` smoke on this branch (flat fast path) — running, will report in a comment.

## Notes for reviewers

- nemo-automodel API drift: verl's automodel engine targets the **0.4.0** API (`create_device_mesh`, `MoEParallelizerConfig`); 0.5.0 and current main renamed these, so the engine does not import there. This PR follows the existing 0.4.0-era code; tracking the rename is left to the automodel engine owners.
- The expert-mesh placements tuple order follows **mesh dim order**, not tensor dim order (`(ep_shard, ep)` mesh ⇒ `(Shard(1), Shard(0))`); all placement math goes through `compute_local_shape_and_global_offset`, so no ordering assumptions are made.
- veomni/torchtitan EP express the expert split outside DTensor placements (rank-local expert namespaces); they need the converter route with an expert-id renaming closure and are follow-ups per #7060.

### DIFF
--- a/tests/checkpoint_engine/test_block_placement.py
+++ b/tests/checkpoint_engine/test_block_placement.py
@@ -67,7 +67,7 @@ def test_translate_matches_full_coordinates(full_shape, grid):
     full = torch.randn(full_shape)
     flat = full.reshape(-1)
     for place in _blocks_for_grid(full_shape, grid):
-        slices = tuple(slice(o, o + sz) for o, sz in zip(place.global_offset, place.local_shape))
+        slices = tuple(slice(o, o + sz) for o, sz in zip(place.global_offset, place.local_shape, strict=False))
         local = full[slices].contiguous().reshape(-1)
         lidx = torch.arange(local.numel(), dtype=torch.int64)
         gidx = translate_flat_indices(lidx, place)
@@ -79,7 +79,7 @@ def test_translate_sparse_subset():
     torch.manual_seed(1)
     full = torch.randn(full_shape)
     place = BlockPlacement((2, 2, 6), (4, 2, 0), full_shape)
-    slices = tuple(slice(o, o + sz) for o, sz in zip(place.global_offset, place.local_shape))
+    slices = tuple(slice(o, o + sz) for o, sz in zip(place.global_offset, place.local_shape, strict=False))
     local = full[slices].contiguous().reshape(-1)
     lidx = torch.tensor([0, 5, 11, 23], dtype=torch.int64)
     gidx = translate_flat_indices(lidx, place)
@@ -121,12 +121,10 @@ def test_block_nan_rebuild_matches_full_convert_then_diff():
     # ep=4 x ep_shard=2 grid of blocks, like the (Shard(0), Shard(1)) expert mesh
     full_nan = torch.full(full_shape, float("nan"), dtype=old.dtype)
     for place in _blocks_for_grid(full_shape, {0: 4, 1: 2}):
-        slices = tuple(slice(o, o + sz) for o, sz in zip(place.global_offset, place.local_shape))
+        slices = tuple(slice(o, o + sz) for o, sz in zip(place.global_offset, place.local_shape, strict=False))
         lo = old[slices].contiguous().reshape(-1)
         ln = new[slices].contiguous().reshape(-1)
-        changed = (
-            (lo.view(torch.uint8).view(lo.numel(), -1) != ln.view(torch.uint8).view(ln.numel(), -1)).any(dim=-1)
-        )
+        changed = (lo.view(torch.uint8).view(lo.numel(), -1) != ln.view(torch.uint8).view(ln.numel(), -1)).any(dim=-1)
         lidx = changed.nonzero(as_tuple=False).view(-1)
         buf = torch.full((lo.numel(),), float("nan"), dtype=old.dtype)
         buf[lidx] = ln[lidx]
--- a/tests/checkpoint_engine/test_block_placement.py
+++ b/tests/checkpoint_engine/test_block_placement.py
@@ -1,0 +1,151 @@
+# Copyright 2025 Bytedance Ltd. and/or its affiliates
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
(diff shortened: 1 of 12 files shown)


ChangyiYang:
Regression check: Qwen2.5-0.5B FSDP2 `delta_sharded` smoke on this branch (flat `Shard(0)` fast path, 2×8 GPU disaggregated, 3 GRPO steps) — 3/3 steps green, steady-state `timing_s/update_weights` 1.9 s, no errors. The block-placement changes leave the existing FSDP path behavior unchanged.

ChangyiYang:
Two scoping notes from further validation, and one question for the maintainers.

**1. Where the automodel wiring stands.** Upstream automodel is SFT-only today — this PR's `model_engine=automodel` hydra group is what first makes it reachable from PPO, and we used it as the validation vehicle for the block-placement machinery (its grouped experts are a textbook `(Shard(0), Shard(1))` DTensor). The engine-side changes (BlockPlacement / translate / flattened gather group / block converter profile) are backend-agnostic and independent of that wiring.

**2. A gap in the full-weight path for automodel MoE, found while baselining.** `AutomodelEngine.get_per_tensor_param` (pre-existing, previously dead code) yields the model's native fused expert names (`experts.gate_and_up_projs` / `experts.down_projs`). SGLang's `qwen3_moe.load_weights` has no mapping for those names and silently skips them (`if name not in params_dict: continue`), so a full-weight backend (e.g. `nccl`) would update dense params only — no error raised. The `delta_sharded` path in this PR is unaffected (its exporter routes experts through the model's HF state-dict adapter, emitting classic per-expert names). Fix for the full-weight path would mirror torchtitan: run the export through `state_dict_adapter.to_hf` for MoE params.

**Question:** would you prefer (a) keep the PPO wiring and add the adapter conversion to `get_per_tensor_param` so both paths are correct, or (b) drop the automodel PPO wiring from this PR (keeping the backend-agnostic engine core + tests) and let veomni — which already has PPO wiring and an EP-correct full-weight export upstream — be the first EP consumer in a follow-up? We're proceeding with the veomni follow-up either way, per #7060.

ChangyiYang:
Superseding this draft: per offline discussion we're re-cutting the same backend-agnostic engine core (BlockPlacement / Shard(k) / multi-Shard-dim gather / block converter profile, plus the sync-free masked apply) onto **veomni** as the first EP consumer — veomni already has PPO wiring and an EP-correct full-weight baseline upstream, which makes for a cleaner review and an untainted comparison. The automodel PPO wiring experiment stays out. Closing; replacement PR incoming (tracked in #7060).

PR #7082, opened 2026-07-16
[trainer] feat: V1 replay buffer eviction/refill handling for stale, DAPO-filtered, and failed rollout groups

Begunner:
### What does this PR do?

1. Unify V1 ReplayBuffer eviction/refill handling for stale, DAPO-filtered, and failed rollout groups.
2. Add DAPO group filtering for V1 trainers with mode-specific refill behavior.
3. Prevent late trajectory writes by publishing terminal prompt status only after all sessions settle.
4. Aggregate eviction metrics across polling iterations and parameter-sync substeps, including a WandB table for filtered reward-value counts.
5. ReplayBuffer and ReplayBufferAsync are separated into two classes.

Here is the evict-and-refill matrix, in which ``k`` is the number of prompts to evict:
```
    |   trainer mode   |   drop   |               DAPO               |  failure |
    | ---------------- | -------- | -------------------------------- | -------- |
    |       sync       |   NO-OP  |    Evict ``k``; refill ``2k``    |   NO-OP  |
    |      async       |       All the same: Evict ``k``; refill ``k``.         |
```

``drop`` means off-policy staleness dropping, which is only for async trainers.
``DAPO`` means filtering groups whose configured reward metric is identical across all trajectories.
``failure`` is the group status described above.
Validation treats all terminal groups as sampleable.

In sync mode, DAPO is opt-in and trades generation time for training stability; refilling ``2k`` and dispatching them under a bounded in-flight limit balances retry rounds against over-generation. Failure groups retain the standard sync behavior and remain sampleable, avoiding refill-induced tail latency by default.
In async mode, ``num_warmup_batches`` absorbs retry cost, so all three paths refill exactly ``k`` prompts.
``algorithm.filter_groups.max_num_gen_batches`` is intentionally ignored; refill continues until enough groups
are sampleable.

### API and Usage Example

Set `trainer.v1.sampler.max_off_policy_strategy=drop` to drop stale samples older than `trainer.v1.sampler.max_off_policy_threshold`

Set 
```
algorithm.filter_groups.enable=True \
algorithm.filter_groups.metric=acc \
algorithm.filter_groups.max_inflight_gen_batches=1

```
(modify metric name if you define your own AgentLoop) to use dapo dynamic sampling.


verl/trainer/ppo/v1/trainer_base.py
```diff
@@ -594,17 +621,18 @@ def _init_dataloader(self):
             max_samples=self.config.data.get("val_max_samples", -1),
         )
 
-        # Async drop refills an arbitrary number of dropped prompts, which must divide gen_batch_size,
-        # so force gen_batch_size=1.
-        if self.trainer_mode != "sync" and self.config.trainer.v1.sampler.max_off_policy_strategy == "drop":
+        # Async may refill an arbitrary number of stale/DAPO/failure groups. Sync DAPO refills 2x an
+        # arbitrary filtered count. Whole-dataloader-chunk submission therefore requires gen_batch_size=1
+        # in exactly those modes; ordinary sync training can keep gen_batch_size=train_batch_size.
+        filter_groups = self.config.algorithm.get("filter_groups", None)
```
wuxibin89, line 668: Aligned with DAPO recipe:
https://github.com/verl-project/verl-recipe/blob/main/dapo/config/dapo_trainer.yaml#L24-L28

wuxibin89: Update filter group config to `verl/trainer/config/ppo_trainer.yaml`.

wuxibin89:
@tongyx361 Please take a look.

tongyx361 (commented):
Thanks for the update. I have a few concerns before this is merged:

1. [P1] `algorithm.filter_groups.max_num_gen_batches` is currently ignored. This could cause a hang or OOM. Please at least add a warning.
2. [P2] `_resolve_filter_groups_metric()` is called even when a custom sampler is selected. This can trigger built-in DAPO validation before the custom sampler is instantiated, despite the comment stating that custom samplers own their filtering semantics.
3. [P2] The W&B table is rebuilt and re-uploaded with its full history on every call, resulting in quadratic logging overhead over a long training run.

Could you also share any end-to-end experiment results for these changes?

Begunner:
E2E experiments of Qwen3-8B + DAPO-17k, sync & separate_async:
image
image

verl/trainer/ppo/v1/trainer_base.py
Begunner: Added.

Begunner:
> Thanks for the update. I have a few concerns before this is merged:
> 
> 1. [P1] `algorithm.filter_groups.max_num_gen_batches` is currently ignored. This could cause a hang or OOM. Please at least add a warning.
> 2. [P2] `_resolve_filter_groups_metric()` is called even when a custom sampler is selected. This can trigger built-in DAPO validation before the custom sampler is instantiated, despite the comment stating that custom samplers own their filtering semantics.
> 3. [P2] The W&B table is rebuilt and re-uploaded with its full history on every call, resulting in quadratic logging overhead over a long training run.
> 
> Could you also share any end-to-end experiment results for these changes?

Modified. 
Only wandb >=0.20 can use incremental uploading in custom tables. I add the version judgement so wandb >= 0.20 will use incremental uploading and < 0.20 will use full uploading, as is for `log_val_generations`.

verl/trainer/ppo/v1/replay_buffer.py
```diff
@@ -161,68 +207,167 @@ def _sync_metadata_from_transfer_queue(self):
...
+
+        The three sets may overlap. Callers clear and refill their union, so one prompt is never handled
+        twice. ``dapo_counts`` is the {shared_metric_value: group_count} diagnostic for ``dapo_uids``; it
+        rides along in the return value so no hidden state is needed between production and consumption.
+        """
+        if partition_id == "val":
+            return set(), set(), set(), Counter()
+
+        stale_uids = self._stale_terminal_keys(global_steps, partition_id)
+        dapo_uids, dapo_counts = self._dapo_filtered_keys(partition_id)
```
wuxibin89, line 312: The `_dapo_filtered_keys` is full recompute in each poll_interval, should we add cache and only calculate new finished prompt?

wuxibin89: Also, should we skip `_dapo_filtered_keys` if dynamic filtering is not enabled?

verl/trainer/ppo/v1/replay_buffer.py
```diff
@@ -78,7 +88,27 @@ class ReplayBuffer:
     - running: all sessions of the prompt are running.
     - finished: all sessions of the prompt are finished without error.
     - failure: all sessions of the prompt are finished, but at least one session failed.
-    Only prompt with status `finished` or `failure`, its trajectories can be sampled by replay buffer.
+    Only prompts with status `finished` or `failure` enter terminal-group handling.
```
wuxibin89, line 91: Separate ReplayBuffer to `ReplayBufferSync` and `ReplayBufferAsync`

verl/trainer/ppo/v1/replay_buffer.py
```diff
@@ -161,69 +213,185 @@ def _sync_metadata_from_transfer_queue(self):
...
-        print(
-            f"[drop] partition={partition_id} global_steps={global_steps} "
-            f"threshold={self.max_off_policy_threshold} num_dropped={len(stale_uids)} "
-            f"span(min/mean/max)={staleness.min():.0f}/{staleness.mean():.2f}/{staleness.max():.0f}",
-            flush=True,
+    def _sampleable_terminal_keys(
+        self,
+        partition_id: str,
+        eviction_reasons: tuple[set[str], set[str], set[str], Counter],
+    ) -> set[str]:
```
tongyx361, line 323: Please exclude failure groups with no materializable trajectories

A prompt can be marked as `failure` when all of its sessions fail, leaving no trajectory keys in the buffer. However, its UID is still included in `sampleable_keys` and counted toward `len(sampleable_keys) >= batch_size`. This can stop DAPO refilling early, while `_materialize_batch()` returns an empty or underfilled training batch.

Please also add a regression test for a failed prompt with zero trajectory keys.

Begunner:
Long run e2e of Qwen3-8B + DAPO-17k:
image

wuxibin89 (approved):

PR #7083, opened 2026-07-16
[ckpt] fix: added cuda stream synchronization in NCCL broadcast wait for complete

parinayc20:
### What does this PR do?

`ray.util.collective.broadcast()` only *enqueues* the NCCL kernel on ray's internal stream pool and returns before it finishes on the GPU (ray never calls `record_stream()` on the buffer for this path either). `BroadcastOperation.wait_for_complete()` awaited only that enqueue, so buffers could be reused/freed, and reported timings measured, before the broadcast had actually finished.

### Design & Code Changes

- **Correctness**: `finalize()` could free `send_buf`/`recv_buf` while the last broadcast was still running, and in `receive_weights` the very first bucket was yielded to the caller with no synchronization at all before it. A consumer that just happens to touch the buffer on the same (default) stream is accidentally safe; a real inference-engine weight loader copying off its own stream is not — it can read a partially-written buffer and silently load corrupted weights. Added `clone_on_side_stream` to the test `MockServerAdapter` to reproduce exactly this: cloning on a dedicated stream removes the default-stream ordering the old test was inadvertently relying on, so it now actually exercises this path.
- **Bandwidth/timing metrics**: `send_weights`' `time cost` log had no wait for real completion anywhere before it. `wait_for_complete()` for the last bucket returned right after enqueue, with no trailing sync, so it was measuring enqueue not actual transfer time, understating `time cost` and overstating any bandwidth derived from it. `receive_weights` happened to be closer to correct for multi-bucket transfers, since its per-iteration `torch.cuda.synchronize()` incidentally forced real completion between buckets but the first bucket's transfer time wasn't fully accounted for, and single-bucket transfers had no such protection at all.
- `BroadcastOperation._run()` now calls `torch.cuda.synchronize()` right after `collective.broadcast()`, so `wait_for_complete()` genuinely blocks until the kernel is done, fixing both the correctness gap and the metrics on both send and receive paths uniformly.

### Test

Added a `clone_on_side_stream` parameter to the `test_correctness_on_gpu` test to reproduce the bug, Currently if the  `test_nccl_checkpoint_engine` test is run with the `check_allclose` and this parameter set then the test fails.

Moreover here are the nsys logs showing that the weights might be cloned before the broadcast finishes if a non-blocking cuda stream is used:

- Before the fix:
Screenshot from 2026-07-16 17-55-39

- After the fix:
Screenshot from 2026-07-16 17-48-44

### DIFF
--- a/tests/checkpoint_engine/test_correctness_on_gpu.py
+++ b/tests/checkpoint_engine/test_correctness_on_gpu.py
@@ -32,10 +32,12 @@ _ngpus = torch.cuda.device_count()
 
 @pytest.mark.asyncio
 @pytest.mark.parametrize("rebuild_group", [False, True])
+@pytest.mark.parametrize("clone_on_side_stream", [False, True])
 @pytest.mark.parametrize("num_trainer, num_rollout", [(2, _ngpus - 2)])
 @auto_await
 async def test_nccl_checkpoint_engine(
     rebuild_group,
+    clone_on_side_stream,
     num_trainer,
     num_rollout,
     num_nodes=1,
@@ -70,7 +72,9 @@ async def test_nccl_checkpoint_engine(
     trainer_pool, rollout_pool = split_resource_pool(resource_pool, [num_trainer, num_rollout])
     actor_wg = create_trainer_worker_group(trainer_pool, model_config, checkpoint_engine_config)
     actor_wg.reset()
-    rollout, replicas = await create_rollout_worker_group(rollout_pool, model_config, rollout_config, check_allclose)
+    rollout, replicas = await create_rollout_worker_group(
+        rollout_pool, model_config, rollout_config, check_allclose, clone_on_side_stream
+    )
 
     # create checkpoint engine manager
     checkpoint_manager = CheckpointEngineManager(config=checkpoint_engine_config, actor_wg=actor_wg, replicas=replicas)
@@ -191,6 +195,7 @@ async def test_kimi_checkpoint_engine(
 if __name__ == "__main__":
     test_nccl_checkpoint_engine(
         rebuild_group=False,
+        clone_on_side_stream=False,
         num_trainer=2,
         num_rollout=30,
         num_nodes=4,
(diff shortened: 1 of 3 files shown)


parinayc20:
@wuxibin89  could you please review this PR

wuxibin89:
@parinayc20 There's multiple cuda streams cooperating in NCCLCheckpointEngine
1. ray.util.collective create [multiple cuda stream](https://github.com/ray-project/ray/blob/ray-2.56.0/python/ray/util/collective/collective_group/cuda_stream.py#L67-L69)  with `non_blocking=False` for nccl collective ops.
2. sender/receiver use default cuda stream for copy-in/copy-out.

[Legacy default stream](https://docs.nvidia.com/cuda/cuda-programming-guide/02-basics/asynchronous-execution.html#legacy-default-stream) has implicit synchronization with all other blocking streams. So `wait_for_complete` make sure collective ops has been enqueued to blocking cuda stream and will implicit synchronize with default stream.

The only missing part is that we do need a synchronization after final bucket in sender side, without this synchronization, the `send_buf` and `recv_buf` may be freed before last broadcast finished.

Your solution is more safe if we use a separate non-blocking cuda stream for copy-in/copy-out. But since we're not, the added synchronization is a bit over-synced to me.

parinayc20:
@wuxibin89 That's a fair concern about over-synchronization. I've addressed it by moving the torch.cuda.synchronize() call so it only runs after the final bucket on the sender side, and after the first broadcast on the receiver side, right before the corresponding tensor is yielded.

That said, I still think one synchronization is needed on the receiver side to guard against the possibility that whatever rollout consumes the yielded tensor copies it using a non-blocking stream.

On that note: if we can guarantee that all rollouts will only ever copy the yielded tensors using the default stream, then given the implicit mutual synchronization between the legacy default stream and blocking streams, I believe the following synchronization could be removed as well: https://github.com/verl-project/verl/blob/1bfd02c4359a44310759aab7c0382e2bf5836756/verl/checkpoint_engine/nccl_checkpoint_engine.py#L362

wuxibin89 (dismissed):

PR #7085, opened 2026-07-17
[veomni] feat: EP-aware sharded delta export (fused expert stacks)

ChangyiYang:
## What does this PR do?

Second half of the review-requested split (engine core + FSDP moved to #7144, which this PR is stacked on — the first commit here is #7144; review the second commit only): wires **veomni FSDP2+EP** into the sharded delta sync end to end. Supersedes #7080 (veomni replaces automodel as the first EP consumer since it has PPO wiring and an EP-correct full-weight baseline upstream).

### veomni FSDP2+EP export

`VeOmniEngine.get_per_tensor_param_shard`: dense params pass their DTensor placement through verbatim (flat fast path). Grouped experts are split twice — the expert dim manually across `ep_group` (outside DTensor) and the local block FSDP-sharded as a DTensor — which no placement tuple can express, so the exporter synthesizes a `BlockPlacement` in the virtual all-experts tensor (dim-0 offset = `ep_rank * n_local_experts` + the DTensor block offset), uses the world group as the gather group (asserted to be exactly tiled by ep × block mesh), and declares veomni's own MoE param handler as a dim-0-separable `to_hf_chunk` + `hf_slots` enumeration (fused → per-expert HF names, e.g. `mlp.experts.gate_up_proj [128,1536,2048]` → `mlp.experts.{i}.gate_proj/up_proj.weight`).

`_hf_delta_entry` routes converter specs through the NaN row probe (`convert_row_to_hf` / `hf_entry_converter` in `verl/workers/engine/veomni/utils.py`): only the touched dim-0 rows are probed, the converter is a pure permutation so non-NaN survivors are exactly this rank's contributions in final HF coordinates; every rank enumerates the same slot list so the batched gather stays aligned. Mixed-precision fp32 master weights are cast to bf16 before diffing (the wire and the rollout side speak bf16).

## Test

- CPU: `test_hf_delta_export_converter_param` — converter param's delta entry carries hf_slots-keyed final coordinates (veomni utils file-loaded to keep it a CPU unit test). Full checkpoint-engine suite: 24 pass on the stack.
- E2E: Qwen3-30B-A3B, veomni `expert_parallel_size=8`, SGLang rollout TP4, 2×8 GPU disaggregated, `delta_sharded`: steady syncs 5.8–7.9 s (nccl baseline 32.2 s median → **4.5×**); Qwen3-235B-A22B (ep8 × fsdp8, 8+2 nodes, gen TP16): **8.8–9.0 s** vs 246–266 s nccl (**~21×**), seed 68.7 s / 437.9 GB.
- **50-step GRPO training equivalence at 30B-A3B** (public W&B report): score/pg_loss/grad_norm/ppo_kl trajectories overlap within independent-sampling noise across the whole run: https://api.wandb.ai/links/liquid-ai/55u99gwy

### DIFF
--- a/verl/workers/engine/veomni/transformer_impl.py
+++ b/verl/workers/engine/veomni/transformer_impl.py
@@ -15,7 +15,7 @@
 
 import logging
 from dataclasses import dataclass, field
-from typing import TYPE_CHECKING, Any, Callable, Optional, Sequence
+from typing import Any, Callable, Optional, Sequence
 
 import torch
 import torch.distributed as dist
@@ -30,9 +30,6 @@ from veomni.optim import build_lr_scheduler, build_optimizer
 from veomni.utils.seqlen_pos_transform_utils import prepare_fa_kwargs_from_position_ids
 
 import verl.utils.torch_functional as verl_F
-
-if TYPE_CHECKING:
-    pass
 from verl.trainer.config import CheckpointConfig
 from verl.utils import tensordict_utils as tu
 from verl.utils.checkpoint.fsdp_checkpoint_manager import FSDPCheckpointManager
--- a/verl/workers/engine/veomni/transformer_impl.py
+++ b/verl/workers/engine/veomni/transformer_impl.py
@@ -588,118 +588,13 @@ class VeOmniEngine(FSDPEngine):
 
     def get_per_tensor_param_shard(self, **kwargs):
         """Yield each rank's *local* shard ``(name, local_shard, ShardSpec)`` -- the
-        DTensor export plus veomni's EP declarations. Pure export, no side effects.
-
-        Dense params pass their FSDP DTensor placement through verbatim (flat
-        ``Shard(0)`` fast path). Grouped expert params are split twice: the expert
-        dim (0) is split *manually* across ``ep_group`` (outside DTensor), and the
-        local expert block is FSDP-sharded as a DTensor. That combination is not
-        expressible as DTensor placements, so the spec carries an explicit
-        :class:`BlockPlacement` in a virtual full tensor of all experts
-        (dim-0 offset = ``ep_rank * n_local_experts`` + the DTensor block offset)
-        with the world group as the gather group, and a ``to_hf_chunk`` closure over
-        the veomni MoE param handler (pure slice + rename: fused -> per-expert HF
-        names) plus its ``hf_slots`` output enumeration.
+        DTensor export plus veomni's EP declarations. Pure export, no side effects;
+        the mechanics live in :func:`verl.workers.engine.veomni.utils.veomni_shard_export`.
         """
-        import torch.distributed as dist
+        from .utils import veomni_shard_export
 
         load_veomni_model_to_gpu(self.module)
-        params = self.module.state_dict()
-        params = convert_weight_keys(params, getattr(self.module, "_fsdp_wrapped_module", self.module))
-
-        ps = parallel_state.get_parallel_state()
-        model_type = getattr(self.module.config, "model_type", "default")
-        process_func = MOE_PARAM_HANDERS.get(model_type, default_moe_param_handler)
-
-        from torch.distributed.tensor import DTensor
-        from torch.distributed.tensor._utils import compute_local_shape_and_global_offset
-
-        from ..spec import BlockPlacement, ShardSpec
-
-        def _expert_hf_slots(fqn, full_shape):
-            """Full (hf_name, hf_shape) slot table for one fused expert param, in dim-0
-            order matching the default handler's per-segment output order. Only valid
-            for the default handler (custom MOE_PARAM_HANDERS may rename differently),
-            so the caller attaches it only when process_func is the default."""
-            n_experts = int(full_shape[0])
-            slots = []
-            if "gate_up_proj" in fqn:
-                inter = int(full_shape[1]) // 2
-                inner = (inter, *(int(x) for x in full_shape[2:]))
-                for i in range(n_experts):
-                    for proj in ("gate_proj", "up_proj"):
-                        nm = fqn.replace("gate_up_proj", proj)
-                        slots.append((nm.replace("mlp.experts.", f"mlp.experts.{i}.") + ".weight", inner))
-            else:
-                inner = tuple(int(x) for x in full_shape[1:])
-                for i in range(n_experts):
-                    slots.append((fqn.replace("mlp.experts.", f"mlp.experts.{i}.") + ".weight", inner))
-            return slots
-
-        def _is_ep_param(name, param):
-            # Structural check first: veomni's ParallelPlan.apply attaches
-            # ``spec_info`` (para_name="ep") to every expert-parallel-sliced param
-            # -- the sharding fact itself, not a naming convention. Params can lose
-            # the attribute across state_dict/key conversion, so fall back to the
-            # historical name match.
-            si = getattr(param, "spec_info", None)
-            if si is not None:
-                return si.para_name == "ep"
-            return "mlp.experts." in name and any(
-                p in name for p in ["down_proj", "gate_proj", "up_proj", "gate_up_proj"]
-            )
-
-        def _gen():
-            for name, param in params.items():
-                is_expert = ps.ep_enabled and _is_ep_param(name, param)
-                if param.is_floating_point() and param.dtype != torch.bfloat16:
-                    # mixed-precision keeps fp32 master weights; the wire (and the
-                    # rollout side) speak bf16, so cast before diffing.
-                    param = param.to(torch.bfloat16)
-                if not is_expert:
-                    spec = ShardSpec.from_param(param)
-                    local = param.to_local() if isinstance(param, DTensor) else param
-                    yield name, local.reshape(-1), spec
-                    continue
-
-                # local fused block: [n_local_experts, ...] (ep split already applied)
-                if isinstance(param, DTensor):
-                    lshape, goff = compute_local_shape_and_global_offset(
-                        param.shape, param.device_mesh, list(param.placements)
-                    )
-                    local = param.to_local()
-                else:
-                    lshape, goff = tuple(param.shape), (0,) * param.ndim
-                    local = param
-                n_local = int(param.shape[0])
-                full_shape = (n_local * ps.ep_size,) + tuple(param.shape[1:])
-                offset = (ps.ep_rank * n_local + int(goff[0]),) + tuple(int(o) for o in goff[1:])
-                block = BlockPlacement(tuple(int(x) for x in lshape), offset, full_shape)
-
-                # the blocks of one expert tensor are spread over ep x (block fsdp);
-                # v1 requires that to cover the whole trainer world so the world
-                # group doubles as the gather group (rank 0 == engine master).
-                mesh_world = param.device_mesh.size() if isinstance(param, DTensor) else 1
-                assert ps.ep_size * mesh_world == dist.get_world_size(), (
-                    f"expert blocks must tile the trainer world: ep={ps.ep_size} x "
-                    f"block mesh={mesh_world} != world={dist.get_world_size()}"
-                )
-                spec = ShardSpec(
-                    full_shape=full_shape,
-                    place=block,
-                    gather_group=dist.group.WORLD,
-                    # dim-0-separable: a segment is a run of whole experts starting at
-                    # global expert id ``start`` -- exactly the handler's contract.
-                    to_hf_chunk=lambda start, seg, _f=process_func, _n=name: list(_f(_n, seg, expert_id_base=start)),
-                    # slot table enables sender-side conversion; only the default
-                    # handler's naming is enumerable without running the converter.
-                    hf_slots=(
-                        _expert_hf_slots(name, full_shape) if process_func is default_moe_param_handler else None
-                    ),
-                )
-                yield name, local.reshape(-1), spec
-
-        return _gen(), None
+        return veomni_shard_export(self.module)
 
     def _hf_delta_entry(self, name, spec, place, lidx, lval):
         """veomni's per-param entry builder: EP/converter specs (fused expert
--- a/verl/workers/engine/veomni/transformer_impl.py
+++ b/verl/workers/engine/veomni/transformer_impl.py
@@ -588,13 +588,25 @@ class VeOmniEngine(FSDPEngine):
 
     def get_per_tensor_param_shard(self, **kwargs):
         """Yield each rank's *local* shard ``(name, local_shard, ShardSpec)`` -- the
-        DTensor export plus veomni's EP declarations. Pure export, no side effects;
-        the mechanics live in :func:`verl.workers.engine.veomni.utils.veomni_shard_export`.
+        DTensor export plus veomni's EP declarations. The mechanics live in
+        :func:`verl.workers.engine.veomni.utils.veomni_shard_export`; this wrapper
+        owns the offload dance (CPUOffloadPolicy manages placement itself -- see
+        #5995 -- and the delta path returns early in update_weights, so the
+        offload-back happens here, after the exporter is exhausted).
         """
         from .utils import veomni_shard_export
 
-        load_veomni_model_to_gpu(self.module)
-        return veomni_shard_export(self.module)
+        manual_offload = not getattr(self, "_uses_fsdp2_cpu_offload_policy", False)
+        if manual_offload:
+            load_veomni_model_to_gpu(self.module)
+        gen, meta = veomni_shard_export(self.module)
+
+        def _with_offload_back():
+            yield from gen
+            if manual_offload and self._is_offload_param:
+                offload_veomni_model_to_cpu(self.module)
+
+        return _with_offload_back(), meta
 
     def _hf_delta_entry(self, name, spec, place, lidx, lval):
         """veomni's per-param entry builder: EP/converter specs (fused expert
--- a/verl/workers/engine/veomni/transformer_impl.py
+++ b/verl/workers/engine/veomni/transformer_impl.py
@@ -15,7 +15,7 @@
 
 import logging
 from dataclasses import dataclass, field
-from typing import Any, Callable, Optional, Sequence
+from typing import TYPE_CHECKING, Any, Callable, Optional, Sequence
 
 import torch
 import torch.distributed as dist
@@ -30,6 +30,9 @@ from veomni.optim import build_lr_scheduler, build_optimizer
 from veomni.utils.seqlen_pos_transform_utils import prepare_fa_kwargs_from_position_ids
 
 import verl.utils.torch_functional as verl_F
+
+if TYPE_CHECKING:
+    pass
 from verl.trainer.config import CheckpointConfig
 from verl.utils import tensordict_utils as tu
 from verl.utils.checkpoint.fsdp_checkpoint_manager import FSDPCheckpointManager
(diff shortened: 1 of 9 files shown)


ChangyiYang:
Same-allocation comparison, Qwen3-30B-A3B, veomni `expert_parallel_size=8` (+`mixed_precision=false`), SGLang TP4, 2×8 GPU disaggregated, 3 GRPO steps:

| sync | `nccl` (veomni full-weight export) | `delta_sharded` (this PR) |
|---|---|---|
| step 1 | 32.2 s | 5.8 s |
| step 2 | 31.9 s | 7.4 s |
| step 3 | 33.2 s | **7.9 s (~4.3×)** |

The full-weight path pays veomni's EP export (a sequential per-ep-rank broadcast exchange materializing every expert on every rank, then ~18k per-expert tensors through the bucket pipeline) on top of the usual full-model traffic; the delta path diffs each rank's local block and moves 9–92 MB per sync. Training metrics track between the two runs.

ChangyiYang:
## Repro script: 30B-A3B MoE (veomni ep8), delta vs nccl

Self-contained 2-node repro for the MoE rows in the measured-results table (1 trainer node + 1 rollout node, 8×H100 each). Run once with `WMODE=delta_sharded` and once with `WMODE=nccl` on the same allocation shape and compare `timing_s/update_weights`.

Assumptions: a conda/venv with this branch's verl importable (`CODE_DIR`), veomni ≥ 0.1.11 + transformers 5.5, SGLang with the stock `--custom-weight-loader` hook, gsm8k parquet at `$HOME/data/gsm8k/`. Launch one copy of the script per node (e.g. `srun --nodes=2 --ntasks-per-node=1 --gres=gpu:8 bash repro_veomni_ep.sh`); node 0 becomes the ray head + driver.

<details>
<summary><code>repro_veomni_ep.sh</code></summary>

```bash
#!/usr/bin/env bash
# delta_sharded vs nccl weight-sync A/B for veomni FSDP2+EP MoE training.
#   WMODE=delta_sharded|nccl  [MODEL=Qwen/Qwen3-30B-A3B-Instruct-2507]
#   [NNODES_TRAIN=1] [NNODES_ROLLOUT=1] [EP_SIZE=8] [GEN_TP=4] [STEPS=3]
#   [OFFLOAD=True] [MEM_UTIL=0.7] [BUCKET_MB=2048]
set -xeuo pipefail
WMODE="${WMODE:?set WMODE=nccl|delta_sharded}"
CODE_DIR="${CODE_DIR:?path to this branch's verl checkout}"
export PYTHONPATH=${CODE_DIR}:${PYTHONPATH:-}
export WANDB_MODE=${WANDB_MODE:-disabled}
cd "$CODE_DIR"

MODEL="${MODEL:-Qwen/Qwen3-30B-A3B-Instruct-2507}"
GEN_TP="${GEN_TP:-4}"; NGPU="${NGPU:-8}"
NNODES_TRAIN="${NNODES_TRAIN:-1}"; NNODES_ROLLOUT="${NNODES_ROLLOUT:-1}"
EP_SIZE="${EP_SIZE:-8}"

# --- multi-node ray bootstrap (skip if you already have a ray cluster) ---
NODEID="${SLURM_NODEID:-0}"
mapfile -t NODES < <(scontrol show hostnames "$SLURM_JOB_NODELIST")
HEAD_IP="$(getent hosts "${NODES[0]}" | awk '{print $1}' | head -1)"
FLAGDIR=$HOME/.ray_flags/${SLURM_JOB_ID}_${WMODE}; mkdir -p "$FLAGDIR"
if [ "$NODEID" != "0" ]; then
  while [ ! -f "$FLAGDIR/head_ready" ]; do sleep 2; done
  ray start --address="$HEAD_IP:6379" --num-gpus=$NGPU --disable-usage-stats
  while [ ! -f "$FLAGDIR/done" ]; do sleep 5; done
  ray stop --force || true; exit 0
fi
ray start --head --node-ip-address="$HEAD_IP" --port=6379 --num-gpus=$NGPU --disable-usage-stats
touch "$FLAGDIR/head_ready"
TOTAL=$((NNODES_TRAIN + NNODES_ROLLOUT))
for i in $(seq 1 600); do
  ng=$(python -c "import ray;ray.init(address='auto');print(int(ray.cluster_resources().get('GPU',0)));ray.shutdown()" 2>/dev/null || echo 0)
  [ "${ng:-0}" -ge $((NGPU*TOTAL)) ] && break; sleep 5
done
export RAY_ADDRESS="$HEAD_IP:6379"
# -------------------------------------------------------------------------

LOADER='["verl.workers.rollout.sglang_rollout.delta_loader.apply_delta"]'
if [ "$WMODE" = "delta_sharded" ]; then
  DELTA_ARGS=( actor_rollout_ref.rollout.checkpoint_engine.backend=delta_sharded
    +actor_rollout_ref.rollout.checkpoint_engine.engine_kwargs.delta_sharded.encoding=indices
    "+actor_rollout_ref.rollout.engine_kwargs.sglang.custom_weight_loader=${LOADER}" )
else
  DELTA_ARGS=( actor_rollout_ref.rollout.checkpoint_engine.backend=${WMODE} )
fi

    model_engine=veomni \
    trainer.use_v1=True \
    trainer.v1.trainer_mode=separate_async \
    trainer.v1.separate_async.num_warmup_batches=1 \
    trainer.v1.separate_async.parameter_sync_step=1 \
    data.train_files=$HOME/data/gsm8k/train.parquet \
    data.val_files=$HOME/data/gsm8k/test.parquet \
    data.prompt_key=prompt data.truncation=left \
    data.max_prompt_length=512 data.max_response_length=256 \
    data.train_batch_size=16 data.shuffle=False \
    actor_rollout_ref.rollout.n=4 \
    algorithm.use_kl_in_reward=False algorithm.kl_ctrl.kl_coef=0.0 \
    actor_rollout_ref.actor.use_kl_loss=False actor_rollout_ref.actor.kl_loss_coef=0.0 \
    actor_rollout_ref.model.path=${MODEL} \
    actor_rollout_ref.model.trust_remote_code=True \
    actor_rollout_ref.actor.ppo_mini_batch_size=16 \
    actor_rollout_ref.actor.veomni.expert_parallel_size=${EP_SIZE} \
    actor_rollout_ref.actor.veomni.param_offload=${OFFLOAD:-True} \
    actor_rollout_ref.actor.veomni.optimizer_offload=${OFFLOAD:-True} \
    actor_rollout_ref.actor.veomni.enable_full_shard=True \
    actor_rollout_ref.actor.veomni.attn_implementation=flash_attention_2 \
    actor_rollout_ref.actor.veomni.moe_implementation=fused_triton \
    actor_rollout_ref.actor.veomni.mixed_precision=false \
    actor_rollout_ref.actor.ppo_micro_batch_size_per_gpu=1 \
    actor_rollout_ref.ref.log_prob_micro_batch_size_per_gpu=1 \
    actor_rollout_ref.rollout.log_prob_micro_batch_size_per_gpu=1 \
    actor_rollout_ref.rollout.gpu_memory_utilization=${MEM_UTIL:-0.7} \
    actor_rollout_ref.rollout.temperature=1.0 actor_rollout_ref.rollout.top_p=1.0 actor_rollout_ref.rollout.top_k=-1 \
    +actor_rollout_ref.rollout.engine_kwargs.sglang.attention_backend=flashinfer \
    +actor_rollout_ref.rollout.engine_kwargs.sglang.random_seed=1234 \
    actor_rollout_ref.rollout.enforce_eager=True \
    "${DELTA_ARGS[@]}" \
    actor_rollout_ref.rollout.checkpoint_engine.update_weights_bucket_megabytes=${BUCKET_MB:-2048} \
    actor_rollout_ref.rollout.nnodes=${NNODES_ROLLOUT} \
    actor_rollout_ref.rollout.n_gpus_per_node=$NGPU \
    actor_rollout_ref.rollout.tensor_model_parallel_size=${GEN_TP} \
    reward.reward_manager.name=naive \
    "trainer.logger=[console]" \
    trainer.project_name=veomni-ep trainer.experiment_name=${WMODE} \
    trainer.val_before_train=False trainer.test_freq=-1 trainer.save_freq=-1 \
    trainer.total_epochs=1 trainer.total_training_steps=${STEPS:-3} \
    trainer.resume_mode=disable \
    trainer.nnodes=${NNODES_TRAIN} trainer.n_gpus_per_node=$NGPU
touch "$FLAGDIR/done"
ray stop --force || true
```

</details>

The 235B row uses the same script scaled up: `MODEL=Qwen/Qwen3-235B-A22B-Instruct-2507 NNODES_TRAIN=8 NNODES_ROLLOUT=2 EP_SIZE=8 GEN_TP=16 MEM_UTIL=0.75 OFFLOAD=True` (10 nodes total; ep8 × fsdp8 over 64 trainer ranks, sglang TP16 across 2 rollout nodes). Practical notes at that scale: stage the ~470 GB HF checkpoint to node-local disk first, and give the ray-assembly wait a generous timeout with a hard failure if the cluster comes up under-strength — a silently short cluster fails much later at the resource-pool check.

ChangyiYang:
## 235B measured result (largest point in the matrix)

Qwen3-235B-A22B-Instruct-2507, veomni `ep8 × fsdp8` (64 trainer ranks over 8 nodes) + SGLang rollout TP16 (2 nodes), offload on, `mixed_precision=false`, `fused_triton`, bucket 2048 MB — i.e. the repro script above with `NNODES_TRAIN=8 NNODES_ROLLOUT=2 EP_SIZE=8 GEN_TP=16 MEM_UTIL=0.75`. 3 GRPO steps each mode on the same allocation, `timing_s/update_weights`:

| mode | step 1 | step 2 | step 3 |
|---|---|---|---|
| `nccl` (full broadcast) | 255.9 s | 246.5 s | 266.1 s |
| `delta_sharded` | **14.9 s** | **11.8 s** | **11.4 s** |

**≈21× per sync.** Delta payload was 702 / 288 / 78 MB (changed ratio 0.05% → 0.02% at these early-step learning rates) against a ~470 GB bf16 full-model materialization per `nccl` sync; the one-off dense seed on the first sync is streamed and bounded by bucket size like every other flush. Zero receiver checksum failures.

This is the same mechanism seen at smaller scales (the win is the sender-side full-tensor assembly + bucketing that the full broadcast pays, not wire bytes alone), compounding with MoE sparsity: 1.3× (7B dense, offload) → 2.3× (32B/72B dense) → 4.5× (30B-A3B, 50-step medians) → ≈21× (235B-A22B). Full matrix in verl-project/verl#7060 (2026-07-20 comment).

ChangyiYang:
**Training-equivalence evidence (public W&B report):** 50-step GRPO pair at Qwen3-30B-A3B (veomni ep8), `delta_sharded` vs `nccl` full broadcast — score/pg_loss/grad_norm/ppo_kl trajectories overlap within independent-sampling noise while `timing_s/update_weights` holds a 4.5× gap (median 7.1 s vs 32.2 s) across the whole run:

https://api.wandb.ai/links/liquid-ai/55u99gwy

Luosuu:
@ChangyiYang thanks! I will take a look today

verl/workers/engine/veomni/transformer_impl.py
```diff
@@ -576,6 +580,113 @@ def load_checkpoint(
...
+                if isinstance(param, DTensor):
+                    lshape, goff = compute_local_shape_and_global_offset(
+                        param.shape, param.device_mesh, list(param.placements)
+                    )
+                    local = param.to_local()
+                else:
+                    lshape, goff = tuple(param.shape), (0,) * param.ndim
+                    local = param
+                n_local = int(param.shape[0])
+                full_shape = (n_local * ps.ep_size,) + tuple(param.shape[1:])
```
wuxibin89: The full_shape should be (param.shape[0] * ep_size) + tuple(param.shape[1:]). For example, gate_up_proj has full shape `[E, 2I, H]`, the sharding paradim is as follow:
1. Manually split `gate_up_proj` by ep_size, each ep rank gets local experts `[E/ep_size, 2I, H]`
2. Fully_shard local experts on dim=0 or dim=1
- FSDP: `Shard(1)`, local tensor is `[E/ep_size, 2I/fsdp_size, H]`
- [FSDP+muon_expert_zero_comm](https://github.com/ByteDance-Seed/VeOmni/blob/main/veomni/distributed/torch_parallelize.py#L294-L309): `Shard(0)`, local_tensor is `[E/(ep_size * fsdp_size), 2I, H]`

In both case, we should use DTensor's shape to calculate the full_tensor shape: (param.shape[0] * ep_size) + tuple(param.shape[1:]).

verl/checkpoint_engine/delta_sync/sparse_gather.py
```diff
@@ -171,10 +176,57 @@ def gather_dense_to_rank0(
...
+    if max_n == 0:
+        return torch.empty(0, dtype=local_val.dtype, device=dev) if rank == 0 else None
+
+    val_pad = torch.zeros(max_n, dtype=local_val.dtype, device=dev)
+    val_pad[:n] = local_val
+    val_list = [torch.zeros(max_n, dtype=local_val.dtype, device=dev) for _ in range(world)] if rank == 0 else None
+    dist.gather(val_pad, val_list, dst=dst, group=group)
+    if rank != 0:
+        return None
+    full = torch.empty(place.full_shape, dtype=local_val.dtype, device=dev)
```
wuxibin89: We shouldn't materialize full experts on single rank. For example kimi-k2.5 has gate_up_proj: [384, 4096, 7168], if we fully materialize it in bf16, it will consume 21GB memory.

ChangyiYang: Right — and that is exactly what this line computes: `param.shape` here is the DTensor's global shape (`[E/ep_size, 2I, H]` — global over the fsdp mesh, still ep-local on dim 0), so `full_shape = (param.shape[0] * ep_size,) + tuple(param.shape[1:]) = [E, 2I, H]`.

The block itself is also derived generically from the DTensor (`compute_local_shape_and_global_offset` over its mesh/placements, then the dim-0 offset is lifted by `ep_rank * param.shape[0]`), so both layouts you list come out correct: standard FSDP `Shard(1)` gives blocks `[E/ep, 2I/fsdp, H]`, and muon_expert_zero_comm's `Shard(0)` gives `[E/(ep*fsdp), 2I, H]` with the fsdp sub-offset absorbed into the dim-0 coordinate. Added a rebuild-equivalence test for the Shard(0) ep×fsdp dim-0 grid in ed423f68 to pin the muon layout.

wuxibin89: In full weight sync, each ep rank broadcast it's local expert to other rank sequentially, avoiding materialize full experts. https://github.com/verl-project/verl/pull/6612

verl/workers/engine_workers.py
```diff
@@ -699,10 +699,20 @@ async def update_weights(self, global_steps: int = None, mode: str = "auto"):
...
+            # backend pins its shards as the diff base; every later sync ships the
+            # BACKEND-computed HF-coordinate delta.
             if effective_mode == "delta_sharded":
-                per_tensor_param, _ = self.actor.engine.get_per_tensor_param_shard()
+                if getattr(self.checkpoint_engine, "needs_seed", True):
+                    per_tensor_param, _ = self.actor.engine.get_per_tensor_param()
+                    metrics = await self.checkpoint_engine.send_weights(per_tensor_param, global_steps=global_steps)
+                    # weights do not move during the sync, so the snapshots equal
+                    # exactly what the rollout just received.
+                    self.actor.engine.prime_delta_snapshots()
```
wuxibin89: Local shard should be snapshot in each weight sync step, but seems like only snapshot in first step?

ChangyiYang: The snapshot is refreshed on every sync — it's just fused into the steady diff pass instead of a separate call: `hf_delta_export` reads the old snapshot, computes the delta, then `snap.copy_(local)` before yielding each param, so after each sync the snapshot equals exactly what the rollout just received. The explicit `prime_delta_snapshots` here is seed-only because the full sync has no diff pass to fuse into. Fusing avoids a second full-parameter sweep per sync; the two-step snapshot evolution is covered by `test_prime_then_hf_delta_export_roundtrip`.

wuxibin89:
Since this PR has major refactor to DeltaShardedCheckpointEngine, please separate it into 2 PRs:
1. DeltaShardedCheckpointEngine refactor + FSDP
2. DeltaShardedCheckpointEngine + VeOmni with EP support

verl/workers/engine_workers.py
```diff
@@ -699,10 +699,20 @@ async def update_weights(self, global_steps: int = None, mode: str = "auto"):
...
-            # The sharded delta engine diffs each rank's local FSDP shard (no all-gather),
-            # so it consumes the sharded param generator instead of the full-tensor one.
+            # Sharded delta: the SEED sync streams the backend's full HF export
+            # (assembly/conversion is the backend's own full-export logic, so
+            # Megatron/veomni and trainer resume work by construction), then the
+            # backend pins its shards as the diff base; every later sync ships the
+            # BACKEND-computed HF-coordinate delta.
             if effective_mode == "delta_sharded":
-                per_tensor_param, _ = self.actor.engine.get_per_tensor_param_shard()
+                if getattr(self.checkpoint_engine, "needs_seed", True):
```
wuxibin89: We should abstract delta_sharded specific logic into  `DeltaShardedCheckpointEngine` instead of expose to engine worker:
```python
if effective_mode == "delta_sharded":
    metrics = await self.checkpoint_engine.send_weights(self.actor.engine, global_steps=global_steps)
else:
    ...
```

ChangyiYang:
Split done as requested: engine core + FSDP moved to #7144; this PR is now stacked on it and carries only the veomni EP export (second commit, `+660/-0`-ish over #7144). Once #7144 merges this diff shrinks to the veomni files automatically.

verl/workers/engine_workers.py
ChangyiYang: Done in #7144 (commit 874cd3da, this PR is stacked on it): `DeltaShardedCheckpointEngine.send_weights(engine, ...)` now owns the whole sync state machine — it picks seed vs steady off its own `_shard_seeded` state, runs the full export + `prime_delta_snapshots()` on the seed, and pulls `get_per_tensor_param_delta_shard()` on steady syncs. The worker branch is exactly the shape you suggested. Smoke-validated end to end (7B FSDP dense + 7B mcore TP2, seed + multi-step steady, zero checksum errors).

verl/workers/engine/veomni/transformer_impl.py
```diff
@@ -584,14 +587,137 @@ def load_checkpoint(
...
     def get_per_tensor_param_shard(self, **kwargs):
-        # veomni splits grouped experts twice (manual ep split outside DTensor +
-        # FSDP-sharded local block); the inherited DTensor-identity export would
-        # declare the wrong full tensor for them and corrupt the rollout weights
-        # silently. The EP-aware export lands in the follow-up PR.
-        raise NotImplementedError(
-            "delta_sharded does not support the veomni backend yet (EP-aware shard "
-            "export lands in a follow-up PR); use a full-sync checkpoint backend"
-        )
+        """Yield each rank's *local* shard ``(name, local_shard, ShardSpec)`` -- the
```
wuxibin89, line 587: Please move `get_per_tensor_param_shard` implementation to utils.py to keep transformer_impl.py concise.

ChangyiYang: Done in 50b488bd: `get_per_tensor_param_shard` on the engine is now a four-line method; the DTensor + EP declaration machinery (expert slot table, ep-param detection, the export generator) moved to `veomni/utils.py` next to the converter machinery already there.

wuxibin89 (approved):

Issue #6010, opened 2026-04-15
Extremely slow update_actor in GRPO with Qwen3-4B + FSDP2 on Ascend 910B


SsllyZz:
### System Info

Hi, thanks for the great work on VERL.

I’m training **Qwen3-4B** using **GRPO + FSDP2** on **8× Ascend 910B**, and observed very slow training:

* **~800 seconds per step**
* `update_actor` dominates the step time

### Key Config

* `max_prompt_length=4096`
* `max_response_length=8192`
* `rollout.n=8`
* `actor.ppo_micro_batch_size_per_gpu=4`
* `actor.use_dynamic_bsz=True`
* FSDP2 with `fsdp_size=8`

### Observation

`update_actor` is significantly slower than other stages and becomes the main bottleneck. The overall throughput is much lower than expected for a 4B model on 8 GPUs.

### Questions

1. Is this expected for **Qwen3 + GRPO + FSDP2** on Ascend?
2. Are there known performance bottlenecks in `update_actor` under this setup?
3. Any recommended configs (e.g., token budget, batch sizing) to improve actor update throughput?

I can provide full logs and profiling if needed.

Thanks!

### Information

- [ ] The official example scripts
- [x] My own modified scripts

### Tasks

- [ ] An officially supported task in the `examples` folder (such as GLUE/SQuAD, ...)
- [x] My own task or dataset (give details below)

### Reproduction

Below is a simplified reproduction config.

### Environment

* 8 × Ascend 910B
* VERL + GRPO
* Qwen3-4B
* `vllm==0.18.0`
* `vllm-ascend@54879467`
* `transformers@cc7ab9be`

### Launch script

```bash
export RAY_TMPDIR=/mnt1/sfs_turbo/maoyingzhi2024/tmp
export GLOO_SOCKET_IFNAME=enp23s0f3
export NCCL_SOCKET_IFNAME=enp23s0f3

python3 -m verl.trainer.main_ppo \
    algorithm.adv_estimator=grpo \
    data.train_batch_size=128 \
    data.max_prompt_length=4096 \
    data.max_response_length=8192 \
    actor_rollout_ref.model.path=/mnt1/Qwen3.5-4B \
    actor_rollout_ref.model.use_remove_padding=True \
    actor_rollout_ref.model.enable_gradient_checkpointing=True \
    actor_rollout_ref.actor.optim.lr=1e-6 \
    actor_rollout_ref.actor.ppo_mini_batch_size=32 \
[...]

PR #7095, opened 2026-07-19
[fsdp, perf] feat: defer gradient sync during accumulation

zhangxin81:
### What does this PR do?

FSDP currently synchronizes gradients after every micro-batch in `FSDPEngine.forward_backward_batch`. When a PPO mini-batch is split into multiple micro-batches, this performs one reduce-scatter per micro-batch even though the optimizer only steps after the final micro-batch.

This PR adds an opt-in FSDP engine setting:

```yaml
actor_rollout_ref:
  actor:
    fsdp_config:
      use_no_sync_for_gradient_accumulation: true
```

When enabled:

- FSDP1 wraps non-final micro-batches in `FullyShardedDataParallel.no_sync()`.
- FSDP2 calls `FSDPModule.set_requires_gradient_sync(False)` for non-final micro-batches and restores synchronization before the final backward.
- Forward-only execution and the default configuration keep existing behavior.

With `M` micro-batches, this changes gradient synchronization from `M` rounds to one. Parameter all-gathers are unchanged. The option defaults to `false` because deferred synchronization retains unsharded gradients and can increase peak memory.

This is relevant to #6010, where FSDP2 `update_actor` communication was identified as a possible bottleneck, but the PR does not claim to fully resolve that hardware-specific report without an Ascend profile.

### Checklist Before Starting

- [x] Searched the issue and its discussion:
  https://github.com/verl-project/verl/issues/6010
- [x] Searched open PRs by issue:
  https://github.com/verl-project/verl/pulls?q=is%3Apr+is%3Aopen+6010+in%3Abody
- [x] Searched open PRs for `FSDP no_sync`,
  `set_requires_gradient_sync`, `gradient accumulation communication`, and
  `micro batch gradient sync`.
- [x] No open PR implements this synchronization-boundary change.
- [x] #5838 is not a duplicate: it skips zero-advantage samples and changes the
  amount of training work. This PR preserves all samples, losses, and optimizer
  steps and only changes when FSDP gradient collectives run.
- [x] The title follows the required format:
  `[fsdp, perf] feat: defer gradient sync during accumulation`.

### Test

Focused CPU tests:

```bash
PYTHONPATH=$PWD python -m pytest -q \
  tests/trainer/ppo/test_value_loss_normalization_on_cpu.py
```

Result:

```text
45 passed, 7 warnings in 25.93s
```

The test suite covers:

- FSDP1 `no_sync` entry/exit.
- FSDP2 gradient-sync disable/restore.
- FSDP2 synchronization restoration after exceptions.
- Default-off and final-micro-batch behavior.
- Compatibility with engine configs that do not define this FSDP option.
- The actual three-micro-batch scheduling sequence: `False, False, True`.
- Real two-rank Gloo FSDP1 and FSDP2 optimizer-update equivalence against the
  current per-micro-batch synchronization behavior.

Broader CPU validation was split into bounded batches. Across trainer,
workers, utilities, protocol, tools, checkpoint/model, experimental, and
single-controller groups, 702 tests passed. The remaining failures/errors were
reproduced on an untouched `origin/main` worktree or isolated to the test
environment: unavailable local model paths and optional Megatron/vLLM/CUDA
dependencies, an existing numerical-tolerance edge, and nine unmarked async
tests under pytest-asyncio strict mode. Those nine tests pass 9/9 with
`--asyncio-mode=auto`.

Full repository checks:

```bash
pre-commit run --all-files --show-diff-on-failure
```

Result: all hooks passed, including ruff, formatting, mypy, generated config
verification, doc checks, license checks, device API checks, DataProto checks,
test structure checks, naming checks, and compileall.

#### Reproducible performance benchmark

Complete benchmark source, raw timing samples, environment, reproduction
commands, artifact hashes, and the full table:

https://gist.github.com/zhangxin81/d3e495c3f476266ddc9847c996e4f169

Environment:

- Apple M3 Max, macOS arm64
- Python 3.12.13
- PyTorch 2.9.1
- Two Gloo ranks, one CPU thread per rank
- Four independently wrapped FSDP blocks plus the FSDP root, hidden size 1024
- Micro-batch size 1
- Five warmup optimizer steps per path and launch
- Ten measured optimizer steps per path and launch
- Seven independent process launches
- Deterministic, rank-specific input data
- Alternating baseline-first and optimized-first execution order
- Local timer stopped before reducing elapsed time with `MAX` across ranks
- Median paired log-speedup with a 100,000-sample bootstrap 95% confidence
  interval

Affected-path latency and paired speedup:

| FSDP | Accumulation | Baseline median | Deferred-sync median | Paired speedup median | Bootstrap 95% CI | Latency reduction |
|---:|---:|---:|---:|---:|---:|---:|
| 1 | 1 | 53.690 ms | 53.475 ms | 1.003x | [0.998x, 1.004x] | 0.33% |
| 1 | 2 | 107.932 ms | 82.660 ms | 1.301x | [1.295x, 1.313x] | 23.11% |
| 1 | 4 | 215.999 ms | 140.224 ms | 1.534x | [1.527x, 1.542x] | 34.81% |
| 1 | 8 | 434.290 ms | 253.966 ms | 1.679x | [1.669x, 1.711x] | 40.45% |
| 2 | 1 | 55.782 ms | 55.976 ms | 0.992x | [0.987x, 1.002x] | -0.76% |
| 2 | 2 | 111.407 ms | 85.692 ms | 1.297x | [1.289x, 1.300x] | 22.92% |
| 2 | 4 | 224.318 ms | 145.963 ms | 1.545x | [1.519x, 1.547x] | 35.26% |
| 2 | 8 | 444.312 ms | 263.475 ms | 1.688x | [1.681x, 1.689x] | 40.77% |

The accumulation-one confidence intervals include 1.0, which is the expected
negative control because no synchronization is skipped. At accumulation two
and above, both FSDP versions have confidence intervals entirely above 1.0.

All 56 benchmark optimizer updates passed `torch.allclose`; the maximum
observed absolute parameter difference was zero.

This benchmark isolates the affected distributed FSDP synchronization path. It
is not presented as an end-to-end GPU/NPU PPO speedup. GPU/NPU throughput,
collective traces, and peak memory should be measured before considering a
default-on change.

### API and Usage Example

Static micro-batching:

```bash
actor_rollout_ref.actor.fsdp_config.use_no_sync_for_gradient_accumulation=True
critic.fsdp.use_no_sync_for_gradient_accumulation=True
```

The same setting applies when dynamic batching splits a mini-batch into more
than one micro-batch.

Use this option when:

- gradient accumulation creates multiple micro-batches;
- profiler data shows FSDP gradient collectives are significant; and
- enough device memory is available for temporarily retained unsharded
  gradients.

Leave it disabled when memory headroom is tight.

### Design & Code Changes

- Add `use_no_sync_for_gradient_accumulation` to `FSDPEngineConfig` and the
  generated reference config.
- Add `_gradient_sync_context` in `FSDPEngine`.
- Apply the context only to non-final training micro-batches.
- Restore FSDP2 sync state in a `finally` block.
- Default to existing behavior for non-FSDP engine configs that reuse the batch
  loop.
- Add mock scheduling tests and real two-rank FSDP1/FSDP2 equivalence tests.
- Document the communication/memory trade-off in the performance tuning guide.

- [x] Read `CONTRIBUTING.md` and the repository `AGENTS.md`.
- [x] Applied all pre-commit checks successfully.
- [x] Updated performance documentation.
- [x] Added CPU CI tests; no workflow change is needed because the test filename
  matches `tests/**/test_*_on_cpu.py`.
- [x] The `recipe` submodule is not affected.
- [x] CI and review were requested in the PR; fork workflows are awaiting
  maintainer approval.

### AI assistance

OpenAI Codex assisted with repository exploration, implementation, tests,
benchmarking, and PR drafting. The human submitter reviewed every changed line,
understands the synchronization and memory trade-offs, and personally confirmed
the test commands and benchmark results before submission.

### DIFF
--- a/verl/workers/config/engine.py
+++ b/verl/workers/config/engine.py
@@ -243,6 +243,8 @@ class FSDPEngineConfig(EngineConfig):
         forward_prefetch (bool): Whether to prefetch parameters for next forward pass, default False
         model_dtype (str): Model data type used to initialize the transformers model. default "fp32"
         use_orig_params (bool): Whether to use original parameters when initialize FSDP1, default False
+        use_no_sync_for_gradient_accumulation (bool): Whether to skip gradient synchronization for all but the
+            last micro-batch. This reduces communication at the cost of higher peak memory, default False.
         seed (int): Random seed for reproducibility.
         full_determinism (bool): If true, enable_full_determinism is called to ensure reproducible results
             in distributed training. Important: this will negatively impact performance, so only use it for
@@ -263,6 +265,7 @@ class FSDPEngineConfig(EngineConfig):
     forward_prefetch: bool = False
     model_dtype: str = "fp32"
     use_orig_params: bool = False
+    use_no_sync_for_gradient_accumulation: bool = False
     mixed_precision: Optional[dict[str, Any]] = None
     ulysses_sequence_parallel_size: int = 1
     entropy_from_logits_with_chunking: bool = False
--- a/verl/workers/config/engine.py
+++ b/verl/workers/config/engine.py
@@ -243,8 +243,6 @@ class FSDPEngineConfig(EngineConfig):
         forward_prefetch (bool): Whether to prefetch parameters for next forward pass, default False
         model_dtype (str): Model data type used to initialize the transformers model. default "fp32"
         use_orig_params (bool): Whether to use original parameters when initialize FSDP1, default False
-        use_no_sync_for_gradient_accumulation (bool): Whether to skip gradient synchronization for all but the
-            last micro-batch. This reduces communication at the cost of higher peak memory, default False.
         seed (int): Random seed for reproducibility.
         full_determinism (bool): If true, enable_full_determinism is called to ensure reproducible results
             in distributed training. Important: this will negatively impact performance, so only use it for
@@ -265,7 +263,6 @@ class FSDPEngineConfig(EngineConfig):
     forward_prefetch: bool = False
     model_dtype: str = "fp32"
     use_orig_params: bool = False
-    use_no_sync_for_gradient_accumulation: bool = False
     mixed_precision: Optional[dict[str, Any]] = None
     ulysses_sequence_parallel_size: int = 1
     entropy_from_logits_with_chunking: bool = False
--- a/verl/workers/engine/fsdp/transformer_impl.py
+++ b/verl/workers/engine/fsdp/transformer_impl.py
@@ -642,9 +642,16 @@ class FSDPEngine(BaseEngine):
 
     @contextmanager
     def _gradient_sync_context(self, *, is_last_micro_batch: bool):
-        """Skip FSDP gradient communication on non-final accumulation steps."""
-        use_no_sync = getattr(self.engine_config, "use_no_sync_for_gradient_accumulation", False)
-        if not use_no_sync or is_last_micro_batch:
+        """Skip FSDP gradient communication on non-final accumulation steps.
+
+        During gradient accumulation the optimizer only steps after the final
+        micro-batch, so gradients only need to be synchronized once per
+        mini-batch. Deferring synchronization on the non-final micro-batches
+        reduces FSDP gradient collectives from one reduce-scatter per
+        micro-batch to a single round, at the cost of temporarily retaining
+        unsharded gradients until the final backward.
+        """
+        if is_last_micro_batch:
             yield
             return
 
--- a/verl/workers/engine/fsdp/transformer_impl.py
+++ b/verl/workers/engine/fsdp/transformer_impl.py
@@ -19,7 +19,7 @@ import gc
 import logging
 import os
 import warnings
-from contextlib import nullcontext
+from contextlib import contextmanager, nullcontext
 from inspect import signature
 from typing import Callable, ContextManager, Optional
 
@@ -640,6 +640,27 @@ class FSDPEngine(BaseEngine):
     def get_context_parallel_group(self):
         raise NotImplementedError
 
+    @contextmanager
+    def _gradient_sync_context(self, *, is_last_micro_batch: bool):
+        """Skip FSDP gradient communication on non-final accumulation steps."""
+        use_no_sync = getattr(self.engine_config, "use_no_sync_for_gradient_accumulation", False)
+        if not use_no_sync or is_last_micro_batch:
+            yield
+            return
+
+        version = fsdp_version(self.module)
+        if version == 1:
+            with self.module.no_sync():
+                yield
+        elif version == 2:
+            self.module.set_requires_gradient_sync(False)
+            try:
+                yield
+            finally:
+                self.module.set_requires_gradient_sync(True)
+        else:
+            yield
+
     def forward_backward_batch(self, data: TensorDict, loss_function: Callable, forward_only=False) -> list[TensorDict]:
         # note that the global_batch_size should include data on all the dp
         tu.assign_non_tensor(data, sp_size=self.ulysses_sequence_parallel_size)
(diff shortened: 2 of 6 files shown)


verl/workers/engine/fsdp/transformer_impl.py
```diff
@@ -635,6 +635,27 @@ def get_model_parallel_group(self):
...
+        version = fsdp_version(self.module)
+        if version == 1:
+            with self.module.no_sync():
+                yield
+        elif version == 2:
+            self.module.set_requires_gradient_sync(False)
+            try:
+                yield
+            finally:
+                self.module.set_requires_gradient_sync(True)
```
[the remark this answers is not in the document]
zhangxin81: Thanks for flagging this. I verified the concern against the current production construction path and with a real two-rank PEFT reproduction; no code change is needed.

`apply_fsdp2()` always calls `fully_shard(model, ...)` on the root after sharding selected children (`verl/utils/fsdp_utils.py:577-588`), including when `model` is a PEFT wrapper. PyTorch applies `fully_shard` in place by changing the root class to inherit `FSDPModule`, so the root exposes `set_requires_gradient_sync`. In addition, this branch reaches the call only when `fsdp_version(self.module) == 2`, whose definition is `isinstance(model, FSDPModule)` (`verl/utils/fsdp_utils.py:424-430`).

I also ran a two-rank Gloo reproduction using a tiny Qwen2 `get_peft_model(...)` model followed by the repository `apply_fsdp2(...)`: the PEFT root was an `FSDPModule`, there were multiple child FSDP units, and calling the root method toggled `reduce_grads` for every unit in both directions. This matches PyTorch’s default `recurse=True` behavior. The existing distributed test likewise wraps both child/root and exercises this exact engine method.

Iterating every child and calling the method as suggested would recursively toggle descendants more than once, since each call defaults to `recurse=True`.

zhangxin81:
This PR is ready for upstream CI and maintainer review. All 39 fork-triggered workflows are currently `action_required`, so the submitter cannot start them.

Local validation at head `a97493fb211a949676d7c201730a2a58cb5d36d8`:
- 45 focused CPU tests passed, including real two-rank Gloo FSDP1/FSDP2 optimizer equivalence
- full repository pre-commit passed
- all 56 paired benchmark updates were parameter-identical
- affected-path latency reduction is 23% to 41% for accumulation 2 to 8

The submitter reviewed every changed line and personally confirmed the tests and benchmark evidence. Could a maintainer approve/run CI and review the synchronization/memory trade-off?

verl/workers/config/engine.py
```diff
@@ -263,6 +265,7 @@ class FSDPEngineConfig(EngineConfig):
     forward_prefetch: bool = False
     model_dtype: str = "fp32"
     use_orig_params: bool = False
+    use_no_sync_for_gradient_accumulation: bool = False
```
wuxibin89: This config is not needed, we can enable it by default.

zhangxin81: Done, thanks. Removed the config entirely and made deferred gradient sync the default behavior in `55fdf0c2`:

- Dropped `use_no_sync_for_gradient_accumulation` from `FSDPEngineConfig`, `engine/fsdp.yaml`, and the generated `_generated_ppo_trainer.yaml` (regenerated via `scripts/generate_trainer_config.sh`; `autogen-trainer-cfg` passes).
- `_gradient_sync_context` now always defers synchronization on non-final training micro-batches; forward-only passes are unchanged (they use `nullcontext`), and non-FSDP modules (`fsdp_version == 0`) fall through to a plain backward, so the VeOmni subclass that reuses this loop is unaffected.
- Tests updated: dropped the config-gating cases; the two-rank Gloo FSDP1/FSDP2 equivalence test now compares the default deferred-sync path against an explicit per-micro-batch baseline. `45 passed` locally, full `pre-commit run` is green.

I also rebased onto latest `main` to resolve the conflict in `transformer_impl.py` (only the `contextlib` import overlapped with the new `cu_seqlens` change). The diff is now down to `transformer_impl.py`, the CPU test, and the perf-tuning doc.

wuxibin89 (approved):

Issue #5995, opened 2026-04-13
[Bug] FSDP2 CPUOffloadPolicy + state_dict() crashes with device mismatch during update_weights (non-LoRA full-weight training)


yifannnwu:
### System Info

- verl: latest main branch
- Python: 3.12
- PyTorch: 2.10
- vLLM: 0.19.0
- Ray: latest
- Hardware: H200 (140GB each), multi-node (8-16 nodes)
- CUDA: 12.x
- Model: Qwen3.5

### Reproduction

Running GRPO with FSDP2 hybrid engine (async vLLM rollout + FSDP2 training). Training step (forward + backward + optimizer step) completes successfully. Crash occurs at `update_weights` when verl calls `get_per_tensor_param()` to extract trained weights for syncing to the vLLM rollout engine.

Config (relevant parts):
```
actor_rollout_ref.actor.strategy=fsdp2
actor_rollout_ref.actor.fsdp_config.fsdp_size=64
actor_rollout_ref.actor.fsdp_config.param_offload=true
actor_rollout_ref.actor.fsdp_config.optimizer_offload=true
actor_rollout_ref.actor.fsdp_config.offload_policy=true
actor_rollout_ref.hybrid_engine=true
actor_rollout_ref.rollout.name=vllm
actor_rollout_ref.rollout.mode=async
```

### Error

```
File "verl/workers/engine_workers.py", line 715, in update_weights
    per_tensor_param, peft_config = self.actor.engine.get_per_tensor_param(...)
File "verl/workers/engine/fsdp/transformer_impl.py", line 774, in get_per_tensor_param
    params = self.module.state_dict()
...
RuntimeError: Attempted to set the storage of a tensor on device "cpu" to a storage on different device "cuda:0". This is no longer allowed; the devices must match.
```

### Analysis

In `transformer_impl.py:get_per_tensor_param()`:

1. Line 751 calls `load_fsdp_model_to_gpu(self.module)`, which for FSDP2 just calls `model.to(device)`. However, with `CPUOffloadPolicy`, FSDP2 manages param placement automatically and `model.to(device)` does not actually move the offloaded shards back to GPU.

2. Line 774 calls `self.module.state_dict()` — this crashes because internal PyTorch state_dict hooks encounter params whose storage is on CPU but the wrapper expects CUDA.

[...]

PR #7103, opened 2026-07-21
[veomni, fsdp] fix: skip manual model.to(device) under FSDP2 CPU offload

cben484:
### What does this PR do?

Follow-up to #6604 for the **VeOmni engine**.

#6604 fixed a DTensor storage-device mismatch under FSDP2 CPU offload for the **FSDP engine**
(`workers/engine/fsdp/transformer_impl.py`) by guarding the manual `model.to(device)` calls
with `_uses_fsdp2_cpu_offload_policy`. The **VeOmni engine**
(`workers/engine/veomni/transformer_impl.py`) has the exact same paths but was not covered, so
it still crashes when saving a checkpoint (or exporting weights) under
`enable_fsdp_offload=True`:

```
RuntimeError: Attempted to set the storage of a tensor on device "cpu"
              to a storage on different device "npu:0".
```

This PR adds the same guard to the VeOmni engine.

Fixes the VeOmni-engine case of #5995.

### Checklist Before Starting

- [x] Search for similar PRs:
  - #5995 (root issue), #6604 (FSDP-engine fix this follows), #6463, #7005 (same device-mismatch family)
  - Query: https://github.com/verl-project/verl/pulls?q=is%3Apr+offload+state_dict
  - No existing PR/issue covers the **VeOmni** engine's `save_checkpoint` / `get_per_tensor_param` paths.
- [x] Title format: `[veomni, fsdp] fix: ...`

### Root cause

VeOmni parallelizes FSDP2 with meta-init → `fully_shard(mesh, CPUOffloadPolicy)` →
`to_empty("cpu")`, leaving each parameter a `DTensor` whose local storage is on CPU while its
device metadata points at the accelerator. `VeOmniEngine.save_checkpoint` /
`load_checkpoint` / `get_per_tensor_param` then call `load_veomni_model_to_gpu(module)`
(= `model.to(device)`), which produces a "split-brain" DTensor. The subsequent
`model.state_dict()` runs `param.detach()`, dispatched through `DTensor.__torch_dispatch__ →
_correct_storage_aliasing`, which raises the device-mismatch `RuntimeError`.

Under `CPUOffloadPolicy`, FSDP2 already owns CPU↔accelerator placement, so the manual
whole-model `model.to(device)` is both unnecessary and harmful — exactly the conclusion of
#6604 for the FSDP engine.

### Design & Code Changes

`verl/workers/engine/veomni/transformer_impl.py`:

- `__init__`: set `self._uses_fsdp2_cpu_offload_policy = self.engine_config.enable_fsdp_offload`
  (VeOmni's CPU-offload switch), mirroring the FSDP engine's flag.
- `save_checkpoint`, `load_checkpoint`, `get_per_tensor_param`: guard the
  `load_veomni_model_to_gpu(self.module)` calls with `not self._uses_fsdp2_cpu_offload_policy`,
  matching #6604. In `get_per_tensor_param` the downstream per-DTensor `full_tensor()` in
  `param_generator()` still yields accelerator tensors, so skipping the whole-model move is
  safe.

Behavior is unchanged when CPU offload is off (flag is `False`, original code path).

### Test

Standalone 2-rank repro (no verl/veomni import needed) that reproduces VeOmni's exact
init + move sequence and compares the pre-fix vs post-fix branch:

```bash
# pre-fix behaviour: model.to(device) then state_dict()  -> crashes
torchrun --nproc_per_node=2 repro_veomni_fsdp2_offload_save.py --device npu
# post-fix behaviour: skip model.to(device) then state_dict() -> succeeds
torchrun --nproc_per_node=2 repro_veomni_fsdp2_offload_save.py --device npu --guarded
```

Observed on Ascend NPU 910B (torch 2.9.0, Python 3.11):

```
# UNGUARDED (pre-fix)
[rank0] mode=UNGUARDED (pre-fix) param device(meta)=npu:0 local_storage_device=cpu
[rank0] state_dict() FAILED (this is the bug the guard prevents):
  ...
  File ".../torch/nn/modules/module.py", line 2157, in _save_to_state_dict
    destination[prefix + name] = param if keep_vars else param.detach()
  File ".../torch/utils/_python_dispatch.py", line 571, in alias_non_inplace_storage
    torch._functionalize_unsafe_set(ret, arg)
RuntimeError: Attempted to set the storage of a tensor on device "cpu" to a storage
              on different device "npu:0".  This is no longer allowed; the devices must match.

# GUARDED (fix)
[rank0] mode=GUARDED (fix) param device(meta)=cpu local_storage_device=cpu
[rank0] state_dict() + save OK, 4 keys.
```

The crash stack is identical to the one hit in a real Qwen3.5-35B VeOmni SFT run with
`enable_fsdp_offload=True`.

- [x] Reproduced on **Ascend NPU (910B)**; the guard resolves it.
- [ ] Not verified on CUDA (no GPU available); the path and the `_correct_storage_aliasing`
      check are device-agnostic, so CUDA is expected to behave identically.

- [x] Ran `ruff check` / `ruff format` (pre-commit ruff hooks) on the changed file.
- [ ] Add a CI regression test analogous to
      `tests/special_distributed/test_fsdp2_cpu_offload_state_dict.py` (that test targets the
      FSDP engine; a VeOmni-engine variant would need the veomni package available in CI).
- [ ] Ping `ci-request` on Slack when ready for CI.

AI assistance was used to investigate and draft this change; the author has reviewed every
line and validated the fix on real NPU hardware.

### DIFF
--- a/verl/workers/engine/veomni/transformer_impl.py
+++ b/verl/workers/engine/veomni/transformer_impl.py
@@ -145,6 +145,11 @@ class VeOmniEngine(FSDPEngine):
         self._is_offload_param = self.engine_config.param_offload
         self._is_offload_optimizer = self.engine_config.optimizer_offload
         self._is_lora = self.model_config.lora_rank > 0
+        # When VeOmni parallelizes with enable_fsdp_offload, FSDP2 uses CPUOffloadPolicy and
+        # owns CPU<->accelerator placement. Manually calling model.to(device) then crashes
+        # state_dict() with a DTensor storage device mismatch (see #5995 / #6604, which fixed
+        # the FSDP engine; the VeOmni engine has the same paths and needs the same guard).
+        self._uses_fsdp2_cpu_offload_policy = self.engine_config.enable_fsdp_offload
 
         self.use_ulysses_sp = parallel_state.get_parallel_state().sp_enabled
         self.ulysses_sequence_parallel_size = self.engine_config.ulysses_parallel_size
@@ -545,7 +550,9 @@ class VeOmniEngine(FSDPEngine):
         Save VeOmni checkpoint, handling parameter offload as needed.
         """
         origin_module_device = next(self.module.parameters()).device.type
-        if self._is_offload_param or origin_module_device == "cpu":
+        if (self._is_offload_param or origin_module_device == "cpu") and not getattr(
+            self, "_uses_fsdp2_cpu_offload_policy", False
+        ):
             load_veomni_model_to_gpu(self.module)
 
         self.checkpoint_manager.save_checkpoint(
@@ -562,7 +569,7 @@ class VeOmniEngine(FSDPEngine):
         """
         Load VeOmni checkpoint, restoring parameters and optimizer state.
         """
-        if self._is_offload_param:
+        if self._is_offload_param and not getattr(self, "_uses_fsdp2_cpu_offload_policy", False):
             load_veomni_model_to_gpu(self.module)
 
         self.checkpoint_manager.load_checkpoint(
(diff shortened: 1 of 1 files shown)


cben484:
All three CI failures are unrelated to this PR, which only touches the VeOmni engine (`workers/engine/veomni/transformer_impl.py`). The VeOmni-related checks pass, including `e2e_ppo_trainer_veomni_vllm` and `e2e_ppo_trainer_veomni_vllm_ascend` (VeOmni on Ascend NPU).

- `e2e_ppo_trainer_megatron-qwen3`: fails in Ray teardown, not training — `os.pidfd_open(pid, 0) -> OSError: [Errno 22] Invalid argument` during `ray stop`.
- `e2e_ppo_trainer_fsdp-qwen2_5vl-3b_ascend`: vLLM rollout `EngineCore` died (`vllm.v1.engine.exceptions.EngineDeadError`), unrelated to checkpoint/offload.
- `vllm_ascend`: a non-deterministic agent-loop assertion (`test_tool_agent - assert np.int32(2) == 4`; the job is marked "non deter").

None of these touch the offload / `model.to(device)` path changed here. Could a maintainer please re-run the failed jobs? Thanks!

wuxibin89 (approved):

PR #7106, opened 2026-07-21
[megatron] fix: preserve R2 router replay for THD-packed batches

hbhflw2000:
## Summary

Fix Megatron R2 router replay correctness for THD-packed batches, including sequence-parallel alignment padding and PP/VPP-local router selection.

R2 records the MoE routes used to compute old log probabilities and replays those routes during the actor update. Synthetic THD alignment tokens and ambiguous local-router indexing could otherwise shift the recorded route tensor or replay it into the wrong local/VPP layer.

## Scope

- Canonicalize R2-recorded top-k indices for THD sequence-parallel alignment padding.
- Record and replay only the actual local MoE router instances.
- Isolate cached local-router indices by VPP rank.
- Preserve the existing R3 replay-mask path while fixing R2.
- Treat `routed_experts=None` as absent during R2 RECORD selection.
- Fail clearly when actor update replay data is missing or has an invalid shape.
- Support nested multimodal Hugging Face configs when resolving Megatron attributes such as `hidden_size` and RoPE theta.

## Validation

- Focused router-replay and nested Megatron config tests after the latest rebase: **24 passed**.
- Targeted pre-commit hooks: **all passed**.
- 4-GPU pruned fully-async R2 gate:
  - Four replay layers per actor.
  - Two rollout samples.
  - **1/1 actor update completed**.
  - Rollout parameter version advanced to `1`.
- Rebased onto latest `main` (`dd1b5af`); range-diff is identical to the previously validated patch.

## Relationship to fully-async rollout

R2 is optional. The default fully-async path does not require router replay; this fix is used when `actor_rollout_ref.actor.router_replay.mode=R2` is enabled, including the optional R2 configuration of [verl-omni #290](https://github.com/verl-project/verl-omni/pull/290).

## Non-goals

- Changing R3 routing semantics.
- Changing Megatron-Core routing or all-to-all kernels.
- Adding rollout-server or weight-transfer behavior.


hbhflw2000:
Updated on 2026-07-26:

- Rebased cleanly onto current `upstream/main` (`983cb0f2`).
- Targeted CPU tests for nested Megatron config resolution, THD padding, and router replay — **13 passed**.
- Targeted pre-commit hooks — **all passed** (including format, mypy, license, structure, and compile checks).
- Scope note: R2/THD replay is validated independently. It is intentionally disabled in the BSHD full-model async integration gate, so that gate is not being cited as R2 evidence.

SamitHuang (approved):
The fix looks clear to me

HollowMan6:
@hbhflw2000 Could you help rebase this PR? Thanks!

hbhflw2000:
Thanks for the reminder! I’ve rebased the PR onto the latest `main`. The targeted tests pass (`14 passed`).

verl/workers/engine/megatron/transformer_impl.py
```diff
@@ -1368,28 +1379,40 @@ def forward_step(
...
+                    mtp_enable_train=self.model_config.mtp.enable and self.model_config.mtp.enable_train,
+                    local_cp_size=local_cp_size,
+                    router_padding_mask=router_padding_mask,
+                    mtp_loss_normalization_factor=mtp_loss_normalization_factor,
+                    forced_max_seqlen=tu.get_non_tensor_data(data=batch, key="forced_max_seqlen", default=None),
+                    pad_to_length_bucket=pad_to_length_bucket,
+                    cp_layout=cp_layout,
+                )
+            finally:
+                RouterReplay.clear_global_record_padding_mask()
```
wuxibin89: try...finally swallow any exception raised in `forward_fn`, don't do this.

verl/workers/engine/megatron/transformer_impl.py
```diff
@@ -1368,28 +1379,40 @@ def forward_step(
...
-                local_cp_size=local_cp_size,
-                router_padding_mask=router_padding_mask,
-                mtp_loss_normalization_factor=mtp_loss_normalization_factor,
-                forced_max_seqlen=tu.get_non_tensor_data(data=batch, key="forced_max_seqlen", default=None),
-                pad_to_length_bucket=pad_to_length_bucket,
-                cp_layout=cp_layout,
-            )
+            record_padding_mask = None
+            if RouterReplayHelper.is_r2_record_action(self.tf_config, vp_rank):
+                record_padding_mask = get_thd_sequence_parallel_padding_mask(input_ids)
```
wuxibin89: Why we need `record_padding_mask` here, routed_experts is padded with zero in `preprocess_thd_engine`:
https://github.com/verl-project/verl/blob/main/verl/models/mcore/util.py#L478-L484

Related PR: https://github.com/verl-project/verl/pull/7261

verl/workers/engine/megatron/transformer_impl.py
hbhflw2000: Thanks for pointing this out. I removed the try/finally and the global record-padding state. R2 now keeps the native RECORD routes: postprocess_thd_engine drops alignment-only rows, and preprocess_thd_engine restores them as expert-zero rows for replay, so replay forward and backward remain consistent without per-forward mutable state. I also added a regression test for this lifecycle.

verl/workers/engine/megatron/transformer_impl.py
hbhflw2000: Yes, the existing padding behavior from #7261 is sufficient. I removed the redundant RECORD-side canonicalization and now rely on the shared THD postprocess/preprocess lifecycle.

hbhflw2000:
Hi @wuxibin89 , thanks for the detailed review, I've resolved problems mentioned in above comments and rebased to main branch. Could you take another look when convinient? Thanks

verl/utils/megatron/router_replay_utils.py
```diff
@@ -256,7 +256,35 @@ def merge_router_topk_indices(
...
     """
     with torch.no_grad():
-        router_instances_list = RouterReplayHelper.get_micro_batch_router_list(tf_config, vp_rank)
+        vp_rank = 0 if vp_rank is None else vp_rank
+        local_router_instances = RouterReplayHelper.get_micro_batch_router_list(tf_config, vp_rank)
+        expected_router_count = len(local_router_instances)
+        registered_router_ids = {id(router) for router in RouterReplay.router_instances}
+        selector_uses_registry = all(id(router) in registered_router_ids for router in local_router_instances)
+        recorded_router_indices = [
+            idx for idx, router in enumerate(RouterReplay.router_instances) if router.recorded_topk_idx is not None
```
SamitHuang: `recorded_router_indices` is a process-global scan and is not intersected with `local_router_instances`. RECORD leaves `recorded_topk_idx` set until `clear_global_indices()` at the end of `forward_backward_batch`, so after VP0 records, VP1 sees leftover rows and `len(recorded_router_indices) == expected_router_count` fails. Later VPP stages then skip the per-VP cache and fall back to `get_micro_batch_router_list`.

Filter to routers that are in the current local list before comparing counts and caching:
```python
local_ids = {id(r) for r in local_router_instances}
recorded_router_indices = [
    idx for idx, router in enumerate(RouterReplay.router_instances)
    if router.recorded_topk_idx is not None and id(router) in local_ids
]
```

verl/utils/megatron/router_replay_utils.py
```diff
@@ -691,18 +728,18 @@ def get_micro_batch_router_list(tf_config, vp_rank=None):
...
+        local_rank_info = get_current_rank_layer_info(tf_config, vp_rank)
+        start, end = local_rank_info["start"], local_rank_info["end"]
+        offset = sum(1 for layer_idx in range(start) if is_moe_layer(tf_config, layer_idx))
+        num_layers_to_build = sum(1 for layer_idx in range(start, end) if is_moe_layer(tf_config, layer_idx))
+
+        # Some older Megatron builds only register local router instances in each
+        # PP worker, while Qwen3-Omni currently registers the full PP model in
+        # this process. Support both layouts, but prefer the global layer offset
+        # when the registry is large enough.
+        if len(RouterReplay.router_instances) == num_layers_to_build:
```
SamitHuang: This replaces the old per-VP-stage cumulative offset with a global MoE-layer index into `RouterReplay.router_instances`.

That is correct when this process registers the full model. It is wrong when a PP rank only registers its local VP chunks: `len(registry)` is then `local_pp_moe_count`, which is larger than one VP stage’s `num_layers_to_build`, so the `offset = 0` special case does not fire. VP1 then slices at the global start (e.g. `[4:6]` on a 4-element local list) and RECORD raises or silently binds the wrong routers.

The previous loop (`offset += get_moe_num_layers_to_build(..., pre_vp_stage)`) handled that local-registry + VPP layout. Please keep both layouts, and add a test where `len(router_instances) == local_pp_total` but `vp_rank > 0`.

verl/workers/engine/megatron/transformer_impl.py
```diff
@@ -1297,13 +1302,18 @@ def forward_step(
...
+                    "R2 compute_log_prob must record and preserve routes before actor update."
+                )
             replay_mask = None
             if self.engine_config.router_replay.mode == "R3":
                 layers_topk_idx = align_r3_router_replay_data(layers_topk_idx, input_ids)
                 replay_mask = build_r3_replay_mask(input_ids, batch["response_mask"])
             set_router_replay_data(
                 layers_topk_idx,
-                None,
+                attention_mask,
```
SamitHuang, line 1316: Passing `attention_mask` here is required for BSHD packed replay. On this engine path `set_router_replay_data` is also called with `model=unwrapped_model`, which walks the forwarded chunk’s routers and returns before reading `_verl_router_replay_local_router_indices_by_vp`.

The new VPP cache is therefore unused for actor update replay; only RECORD still depends on registry slicing. The CPU tests call `set_router_replay_data` without `model=`, so they do not cover this path. Either apply the cached indices on the `model=` branch as well, or drop the cache if `iter_model_routers` is now the source of truth.

SamitHuang (commented):
THD postprocess/preprocess reuse, `routed_experts is None` RECORD selection, and the nested HF config lookup are the right direction. The VPP-local router cache still does not hold up on the engine path.

Please treat the inline notes as blocking for R2+VPP: RECORD still scans every `recorded_topk_idx` in the process, `get_micro_batch_router_list` switched from per-VP cumulative offset to a global MoE index (breaks PP ranks that only register local VP chunks), and actor-update replay walks `model=` so the new `_verl_router_replay_local_router_indices_by_vp` cache is unused.

The CPU tests cover the cache when `model is None` and local-only registry without VPP; they do not cover leftover recordings across VP stages or local-registry + `vp_rank > 0`.

wuxibin89 (approved):

PR #7107, opened 2026-07-21
[ckpt]: nccl broadcast bucket size fix

parinayc20:
### What does this PR do?

The NCCL checkpoint engine always broadcasts the full bucket_size buffer per bucket, even when the bucket was only partially filled (which is the common case for the last bucket of every send_weights call, and any time the parameter stream doesn't divide evenly into bucket_size). This wasted NCCL bandwidth is proportional to the unused tail of each bucket.

This PR fixes it by passing the bucket length as metadata, ensuring both sides agree on the broadcast size before the broadcast begins.

### DIFF
--- a/verl/checkpoint_engine/nccl_checkpoint_engine.py
+++ b/verl/checkpoint_engine/nccl_checkpoint_engine.py
@@ -85,6 +85,7 @@ class BroadcastOperation:
         else:
             self.socket.recv_string()
             self.metadata = self.socket.recv_pyobj()
+            self.bucket = self.bucket[: self.metadata["length"]]
 
         # broadcast tensor via NCCL
         collective.broadcast(self.bucket, src_rank=0, group_name=self.group_name)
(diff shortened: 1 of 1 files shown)


verl/checkpoint_engine/nccl_checkpoint_engine.py
```diff
@@ -264,8 +265,8 @@ async def send_weights(
                 broadcast_op = BroadcastOperation(
                     rank=self.rank,
                     group_name=self.group_name,
-                    bucket=send_buf,
-                    metadata={"bucket_meta": bucket_meta, "is_last": False},
+                    bucket=send_buf[:offset],
```
wuxibin89, line 268: recv_buf should be also restricted by offset: `recv_buf[:offset]`?

parinayc20: The recv_buf restriction happens in `BroadcastOperation._run()` before calling the final broadcast collective:
https://github.com/verl-project/verl/pull/7107/files#diff-2cdb965b958cf2a60dd21504e41bd098a583b4d2b82291a612dc4c5dbae85b77R88

The recv_buf cannot be restricted when using it to create the BroadcastOperation in the `receive_weights()` function itself, because the offset is not known until that point.

parinayc20:
@wuxibin89  do you see any other issues with this?

wuxibin89 (approved):

PR #7114, opened 2026-07-22
[ckpt, megatron] fix: megatron save checkpoints with strict false

RichardFido:
### What does this PR do?
  What
  When saving model weights as HuggingFace format through the Megatron checkpoint manager (_save_model_as_hf_via_bridge), the strict option from checkpoint_config was previously ignored — the
  call always used the bridge's default behavior. This commit reads checkpoint_config.strict (if present) and forwards it to bridge.save_hf_weights(...) so that checkpoints can be saved with
  strict=False, tolerating key mismatches between the source model and the target HF state dict.

  Why
  Saving HF checkpoints can fail when the weight keys don't line up exactly (e.g., renamed/extra/missing parameters during conversion). Allowing strict=False lets users persist checkpoints
  despite these mismatches instead of aborting, and makes the strict setting configured on the checkpoint actually take effect on the non-PEFT, non-vanilla-bridge code path.

  How
  - Before: self.bridge.save_hf_weights(self.model, hf_ckpt_path)
  - After: read strict = getattr(self.checkpoint_config, "strict", None); only inject it into the call kwargs when explicitly set, preserving the previous default behavior when the attribute
  is absent.

### Checklist Before Starting

### Test

### API and Usage Example

```python
```

### Design & Code Changes

> [!IMPORTANT]

### DIFF
--- a/verl/utils/checkpoint/megatron_checkpoint_manager.py
+++ b/verl/utils/checkpoint/megatron_checkpoint_manager.py
@@ -1012,7 +1012,9 @@ class MegatronCheckpointManager(BaseCheckpointManager):
                     log_only_rank_0=True,
                 )
             else:
-                self.bridge.save_hf_weights(self.model, hf_ckpt_path)
+                self.bridge.save_hf_weights(
+                    self.model, hf_ckpt_path, strict=self.checkpoint_config.strict
+                )
 
     def _save_hf_config_and_tokenizer(self, local_path: str):
         """Rank-0 saves HF config, tokenizer, and generation config."""


verl/utils/checkpoint/megatron_checkpoint_manager.py
```diff
@@ -1012,7 +1012,11 @@ def _save_model_as_hf_via_bridge(self, hf_ckpt_path: str):
                     log_only_rank_0=True,
                 )
             else:
-                self.bridge.save_hf_weights(self.model, hf_ckpt_path)
+                strict = getattr(self.checkpoint_config, "strict", None)
```
wuxibin89: We can directly access `self.checkpoint_config.strict`:
https://github.com/verl-project/verl/blob/main/verl/trainer/config/config.py#L50

PR #7131, opened 2026-07-23
[gemma4] Support attn_implementation=flash_attention_2 (global layers fall back to SDPA)

nanastassacos:
## What this does

Lets Gemma4 load with `attn_implementation="flash_attention_2"`. Today it crashes with
`FlashAttention forward only supports head dimension at most 256`, because the text tower
alternates 25 sliding layers at `head_dim=256` with 5 global layers at `head_dim=512`, and flash
caps `head_dim` at 256. Five layers block flash for all thirty, so users fall back to `sdpa` for the
whole model — a poor fit for the sliding layers, where SDPA materialises the full `S×S` score matrix
behind an additive mask and discards the sliding-window sparsity (~72 ms vs ~2.25 ms per layer,
isolated attention op fwd+bwd at `S=8192` on an H200).

This runs the 5 global layers on SDPA and leaves the other 25 on the stock flash path.

* `sdpa` / `eager` — every layer takes the stock dispatch. Bit-exact no-op.
* `flash_attention_2` (or any flash flavour) — sliding layers take the stock flash path, global
  layers fall back to SDPA.

## The idea

Vanilla Gemma4 is *already* per-layer-type for RoPE, `head_dim`, `sliding_window`, the
`attention_k_eq_v` aliasing, and which mask entry each layer receives. Two things are not: the
attention **kernel** and the mask **format**, both pinned to the single `config._attn_implementation`
string. This makes those two per-layer-type as well.

## The resulting diff in Gemma4's behaviour

Four lines across two surfaces.

**1. `Gemma4TextAttention.forward` — the kernel.** The vanilla call is reproduced *byte-identically*
inside the `if`; only the `if`, the `else`, and the call inside it are new:

```diff
-    attention_interface: Callable = ALL_ATTENTION_FUNCTIONS.get_interface(
-        self.config._attn_implementation, eager_attention_forward
-    )
-    attn_output, attn_weights = attention_interface(
-        self, query_states, key_states, value_states, attention_mask,
-        dropout=..., scaling=self.scaling, sliding_window=self.sliding_window, **kwargs,
-    )
+    if self.is_sliding or not is_flash_attention_requested(self.config):
+        attention_interface = ALL_ATTENTION_FUNCTIONS.get_interface(
+            self.config._attn_implementation, eager_attention_forward
+        )
+        attn_output, attn_weights = attention_interface(
+            self, query_states, key_states, value_states, attention_mask,
+            dropout=..., scaling=self.scaling, sliding_window=self.sliding_window, **kwargs,
+        )
+    else:
+        attn_output, attn_weights = _global_attention_sdpa(
+            self, query_states, key_states, value_states, attention_mask,
+            dropout=..., **kwargs,
+        )
```

The interface lookup inside the `if` is not redundant: the `or not
is_flash_attention_requested(...)` clause routes the whole non-flash world through that branch, so
`_attn_implementation` still resolves to four different values there (`flash_attention_2`,
`flash_attention_3`, `sdpa`, `eager`).

**2. `create_causal_mask` — the mask format.** Rebound in the `modeling_gemma4` namespace so only
Gemma4's lookups are redirected:

```diff
- return create_causal_mask(config, ...)              # config._attn_implementation as given
+ return create_causal_mask(_as_sdpa(config), ...)    # a config copy saying "sdpa"
```

**No masking logic and no attention math is written here.** The only new statements on the mask path
are `copy.copy(config)` and setting `_attn_implementation`; every mask is still built by stock
`transformers`, and `_global_attention_sdpa` just forwards to `sdpa_attention_forward`. QK-norm,
both RoPEs, the K=V aliasing and the KV cache all run *upstream* of the dispatch, untouched.

### Why the mask half is required

`create_causal_mask` picks its builder from the same config string as the kernel, and
`flash_attention_mask` cannot return a 4D mask by construction. So changing only the kernel would
leave the global layers running SDPA on a flash-format mask — which fails on a padded batch, and
under `use_remove_padding` is silently wrong rather than loud.

That is because `None` is legal for both kernels and means opposite things: to flash, *"boundaries
are in `cu_seqlens`, derived from `position_ids`"*; to SDPA, *"no constraints — attend causally to
everything"*. A packed batch would then have every non-first sequence attending into its
predecessor, with no crash and no warning.

With both halves in place the two agree per layer type, using Gemma4's existing lookup
(`causal_mask_mapping[self.config.layer_types[i]]`):

```
sliding layers  <- attention_mask=None         -> flash varlen (cu_seqlens from position_ids)
global layers   <- attention_mask=(B,1,Q,KV)   -> SDPA (block-diagonal)
```

## Tests

`tests/models/test_gemma4_flash_sliding_on_gpu.py` (new) — 10 GPU cases on a small but real Gemma4
exercising both flavours (`head_dim` 256/512, both RoPEs, `attention_k_eq_v`), randomly initialised,
so no checkpoint or dataset is needed. Flash matches sdpa; sdpa is a bit-exact no-op; padding never
reaches a valid token; packed sequences do not attend across boundaries; both unsupported
combinations (Ulysses+flash, non-`"vision"`+flash) raise; `apply_monkey_patch` routes to the patch.

Three deliberate choices: the numerical tolerance is **self-calibrating** (the same stock sdpa model
is run in fp32 and the bf16-vs-fp32 gap is the bar, so it tracks the machine rather than a
hand-picked epsilon that rots — measured at ~0.5× of it); the boundary tests assert the other
packed sequence is **bitwise** unchanged, with a non-vacuity assertion so they cannot pass by doing
nothing; and they cover **both** `Gemma4TextModel` and `Gemma4ForConditionalGeneration`, since
`AutoModel` resolves the released checkpoint to the composite class, whose forward builds the mask
mapping itself and passes it down as a dict.

## Performance

End-to-end GRPO on `gemma-4-26B-A4B-it` (FSDP2, `use_remove_padding=True`, `max_model_len=12288`),
same veRL commit, configs differing in the single `attn_implementation` line. Means are over 96
actor updates (sdpa, ~13 h) and 186 (flash, ~15 h):

| | sdpa | flash |
|---|---|---|
| throughput (mean) | 1150 tok/s | 2131 tok/s — **1.85×** |
| actor update (mean) | 406 s | 216 s — **1.88×** |
| peak memory allocated | 75.2 GB | 75.1 GB |

Memory is flat, so the speedup is not bought with activation memory. Across the 10 evaluations the
two runs share, the difference in score averages +0.7 points in flash's favour and changes sign
three times — smaller than the ~1.8-point step-to-step jitter within either run, so this is noise
and not a quality gain. Nothing is claimed beyond that overlap: the sdpa arm stopped there, while
flash continued for another 9 evaluations with no decay. The correctness argument rests on the unit
tests above, which anyone can run on random weights in ~15 s; the training runs only show that
nothing degrades at scale.

## Known gaps

* **`_patched_gemma4_text_attention_forward` reproduces the upstream forward** — ~90 lines including
  signature and docstring — to change three statements (the `if`, the `else`, and the call inside
  it). This follows the convention in `llama.py` / `qwen2.py` / `apertus.py` and carries the
  same `Adapted from transformers 5.8.0` provenance note, but it is real maintenance debt: a change
  upstream of the dispatch would be silently missed rather than conflict. Happy to rework it if
  there is a preferred seam for per-instance kernel dispatch.
* **Only `use_bidirectional_attention="vision"` variants are supported under flash** — the 26B and
  31B checkpoints and their text-only builds. The smaller ones (E2B/E4B) leave that field unset, and
  their forward builds the mask mapping through `create_masks_for_generate`, which resolves layer
  types via masking_utils' pattern mapping and so never reaches the `create_causal_mask` rebinding.
  Since the kernel dispatch keys on `self.is_sliding`, they would otherwise get the SDPA fallback
  with a flash-format mask — silently attending across packed-sequence boundaries. They now raise
  with a message pointing at `attn_implementation="sdpa"`, which is how they ran before. Supporting
  them properly means patching that second builder too; I left it out rather than ship a path the
  tests here do not cover.
* **Ulysses sequence parallel with flash is refused** with an explanatory error rather than silent
  corruption — the global layers bypass the SP-aware flash forward Ulysses patches in.
  `ulysses_sequence_parallel_size=1` or `attn_implementation="sdpa"` both work.
* **No fused-forward backends** (`forward_with_torch_backend` / `forward_with_triton_backend`).

### DIFF
--- a/tests/models/test_gemma4_flash_sliding_on_gpu.py
+++ b/tests/models/test_gemma4_flash_sliding_on_gpu.py
@@ -1,0 +1,339 @@
+# Copyright 2024 Bytedance Ltd. and/or its affiliates
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""GPU coverage for running Gemma4 under a flash ``attn_implementation``.
+
+Builds a small but *real* Gemma4 text model exercising both attention flavours (head_dim-256
+sliding with the theta=1e4 full-rotary RoPE, and head_dim-512 global with the theta=1e6
+partial-rotary RoPE and ``attention_k_eq_v``) and checks that ``flash_attention_2`` reproduces
+what ``sdpa`` computes, that padding never reaches a valid token, that ``sdpa`` is left alone,
+and that the unsupported combinations fail loudly. Randomly initialised -- no checkpoint or
+dataset required.
+"""
+
+import pytest
+import torch
+
+if not torch.cuda.is_available():
+    pytest.skip("Requires CUDA", allow_module_level=True)
+pytest.importorskip("flash_attn")
+pytest.importorskip("transformers.models.gemma4.modeling_gemma4")
+
+# 4 sliding : 2 global -- exercises both kernels and the fall-through
+MIXED = ["sliding_attention", "sliding_attention", "full_attention"] * 2
+GLOBAL_ONLY = ["full_attention", "full_attention"]
+
+SEQ_LEN = 12
+LENS = (12, 7)
+
+
+def _config(layer_types, use_bidirectional_attention="vision"):
+    """Tiny config that keeps the load-bearing shapes: head_dim 256/512 and the two RoPEs."""
+    from transformers.models.gemma4.configuration_gemma4 import Gemma4TextConfig
+
+    return Gemma4TextConfig(
+        vocab_size=256,
+        hidden_size=128,
+        intermediate_size=256,
+        num_hidden_layers=len(layer_types),
+        layer_types=list(layer_types),
+        num_attention_heads=4,
+        num_key_value_heads=2,
+        head_dim=256,  # sliding layers (flash-legal)
+        global_head_dim=512,  # global layers (flash-illegal -> fall back to SDPA)
+        num_global_key_value_heads=1,
+        attention_k_eq_v=True,  # global layers: value_states = key_states, as in the real model
+        sliding_window=8,  # small so the window actually bites within a short sequence
+        rope_parameters={
+            "full_attention": {
+                "partial_rotary_factor": 0.25,
+                "rope_theta": 1_000_000.0,
+                "rope_type": "proportional",
+            },
+            "sliding_attention": {"rope_theta": 10_000.0, "rope_type": "default"},
+        },
+        use_bidirectional_attention=use_bidirectional_attention,  # "vision" -> is_causal=True for text
+        hidden_size_per_layer_input=0,
+        enable_moe_block=False,  # dense MLP; MoE is per-token and orthogonal to attention
+        num_kv_shared_layers=0,
+        final_logit_softcapping=None,
+        attention_dropout=0.0,
+    )
+
+
+def _batch(cfg, left_pad=False):
+    """Right-padded (default) batch of two sequences of different real lengths."""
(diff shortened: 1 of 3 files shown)


nanastassacos:
Could I ask for your review? This PR is very self-contained and designed to touch as little as possible. @vermouth1992 @PeterSH6 @wucong25 @wuxibin89

PR #7134, opened 2026-07-23
[hardware] fix: make NPU availability checks robust in CPU-only environments

chenin-wang:
## Problem

On Ray clusters where the head node is CPU-only but worker nodes have NPUs, submitting a GRPO training job via `ray job submit` fails during the import stage on the driver (head node).

**Root cause**: `torch.npu.is_available()` raises an exception when torch_npu is installed but no NPU devices are present, unlike `torch.cuda.is_available()` which always returns a boolean. This causes:
- Module-level `is_npu_available` assignment to crash in `verl/utils/device.py`
- Platform initialization to crash during availability checks
- `get_device_capability()` and other query functions may raise in CPU-only environments

## Changes

1. **`verl/utils/device.py`**: Widen `is_torch_npu_available()` exception handling from `ImportError` to `Exception`, so it always returns `False` when devices are unavailable instead of raising.
2. **`verl/plugin/platform/platform_npu.py`**:
   - Add try/except to `PlatformNPU.is_available()` so it never raises — consistent with the `torch.cuda.is_available()` contract.
   - `get_device_capability()` now uses `self.is_available()` (the safe version) for the pre-flight check, and wraps the actual `get_device_capability` call with try/except.
3. **`verl/plugin/platform/platform_manager.py`**: Guard the `is_available()` call in `_create_platform()` so platform initialization does not crash on CPU-only nodes.

## Design Principle

- Only **query functions** (`is_available`, `get_device_capability`) are guarded — operation functions (`set_device`, `synchronize`, etc.) are untouched, so real errors still propagate correctly.
- Aligns NPU behavior with CUDA: availability checks return booleans, they don't raise.

## Testing

Verified that `from verl.trainer.main_ppo import main` succeeds on a CPU-only machine with torch_npu installed (previously it crashed during import).

### DIFF
--- a/verl/plugin/platform/platform_npu.py
+++ b/verl/plugin/platform/platform_npu.py
@@ -55,7 +55,10 @@ class PlatformNPU(PlatformBase):
         return torch.npu
 
     def is_available(self) -> bool:
-        return torch.npu.is_available()
+        try:
+            return torch.npu.is_available()
+        except Exception:
+            return False
 
     def is_platform_available(self, use_smi_check=False) -> bool:
         """Return True if this platform is available on this host.
@@ -114,12 +117,16 @@ class PlatformNPU(PlatformBase):
     # ------------------------------------------------------------------
 
     def get_device_capability(self, device_index: int = 0) -> tuple[Optional[int], Optional[int]]:
+        if not self.is_available():
+            return (None, None)
         if hasattr(torch.npu, "get_device_capability"):
-            result = torch.npu.get_device_capability(device_index)
-            # torch.npu.get_device_capability may return None instead of a tuple
-            if result is None:
+            try:
+                result = torch.npu.get_device_capability(device_index)
+                if result is None:
+                    return (None, None)
+                return result
+            except Exception:
                 return (None, None)
-            return result
         return (None, None)
 
     # ------------------------------------------------------------------
--- a/verl/utils/device.py
+++ b/verl/utils/device.py
@@ -53,7 +53,7 @@ def is_torch_npu_available(check_device=True) -> bool:
             return torch.npu.is_available()
         else:
             return True
-    except ImportError:
+    except Exception:
         return False
 
 
--- a/verl/utils/device.py
+++ b/verl/utils/device.py
@@ -53,7 +53,7 @@ def is_torch_npu_available(check_device=True) -> bool:
             return torch.npu.is_available()
         else:
             return True
-    except Exception:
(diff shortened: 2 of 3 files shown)


verl/plugin/platform/platform_npu.py
```diff
@@ -55,7 +55,10 @@ def device_module(self) -> ModuleType:
         return torch.npu
 
     def is_available(self) -> bool:
-        return torch.npu.is_available()
+        try:
```
tardis-key, line 58: you have already handles all Exceptions here(it's reasonable ), so you don't need to handdle it in platform_manager.

verl/utils/device.py
```diff
@@ -53,7 +53,7 @@ def is_torch_npu_available(check_device=True) -> bool:
             return torch.npu.is_available()
         else:
             return True
-    except ImportError:
+    except Exception:
```
tardis-key: overly broad try-catch blocks could lead to improper handling of unanticipated exceptions.

chenin-wang:
Thanks for the review! That makes sense.

I addressed both comments in the latest commit:

1. Removed the redundant guard in `platform_manager` since `PlatformNPU.is_available()` now handles the expected NPU availability probe failures.
2. Narrowed the catch in `is_torch_npu_available()` from `Exception` to `(ImportError, RuntimeError)`, so unexpected exceptions are not swallowed while torch_npu import/runtime availability failures on CPU-only nodes still return `False`.

tardis-key:
Please update your PR title, 
And I've restarted the failed CI— if it still fails, please confirm whether it's related to your PR.

wuxibin89:
@chenin-wang For bytedance internal user, you can submit ray driver to NPU node.

chenin-wang:
> Please update your PR title, And I've restarted the failed CI— if it still fails, please confirm whether it's related to your PR.

@tardis-key  Thanks! I have updated the PR title to:

[hardware] fix: make NPU availability checks robust in CPU-only environments

Thanks for restarting the CI.

chenin-wang:
> @chenin-wang For bytedance internal user, you can submit ray driver to NPU node.

@wuxibin89 Thanks for the suggestion! Yes, for ByteDance internal users, submitting the Ray driver to an NPU node is a valid workaround.

This PR is mainly intended to make verl more robust for general Ray deployments.

tardis-key (approved):

wuxibin89:
I think all platform has this problem, so we're not going to support CPU-only environment.

PR #7150, opened 2026-07-26
[algo] fix: dense group ids in as_torch_index and RLOO vectorized parity

keepkeen:
### What does this PR do?

Fixes two defects in the group-wise advantage path that make the *vectorized* estimators behave differently from the loop-based references they are supposed to replace.

**1. `as_torch_index` does not return the dense ids it promises**

Its docstring states it returns "a contiguous 1-D torch.long tensor in range `[0..G-1]`", and `group_mean_std` uses the result directly as a positional index into `index_add_` accumulators. Three of the four label paths recognized the label set but returned the *raw label values* without renumbering:

| label kind | before | after |
|---|---|---|
| `[2, 2, 5, 7, 5, 2]` | `[2, 2, 5, 7, 5, 2]` | `[0, 0, 1, 2, 1, 0]` |
| `np.array([1.0000001, 2.0, 1.0, 3.0])` | `[1, 2, 1, 3]` | `[0, 1, 0, 2]` |
| `np.array(["7", "7", "9", "9"])` | `[7, 7, 9, 9]` | `[0, 0, 1, 1]` |
| `np.array(["uid-a", "uid-a", "uid-b"])` | `[0, 0, 1]` | `[0, 0, 1]` (unchanged) |

Only the object/UUID path was compliant. Reachable consequences, both via the registered `grpo_vectorized` / `rloo_vectorized` estimators when uids are integers:

```python
# negative labels index out of bounds
compute_grpo_vectorized_outcome_advantage(rewards, mask, np.array([-1, -1, 3, 3]))
# IndexError: index out of range in self

# sparse labels over-allocate: 2 groups -> G = 1002 slots (x5 accumulator tensors)
group_mean_std(scores, as_torch_index([1000, 1000, 1001, 1001]))[0].numel()  # 1002
```

Fix: factorize on every path. `group_mean_std` additionally raises an actionable `ValueError` (instead of a bare `IndexError` from `index_add_`) if a caller passes raw negative labels directly.

The default production path is unaffected: `uid` is a UUID string, which already went through the compliant factorization branch.

**2. `rloo_vectorized` and `rloo` disagree on singleton groups**

`compute_rloo_vectorized_outcome_advantage` multiplied by `(c > 1)`, zeroing the advantage of any group with a single sample, while the `compute_rloo_outcome_advantage` it vectorizes leaves such a score unchanged (the leave-one-out baseline is undefined for `n = 1`, so the loop's `if response_num > 1` guard simply skips it).

Switching `algorithm.adv_estimator` from `rloo` to `rloo_vectorized` is advertised as a performance-only change, but silently altered the objective whenever a prompt group had one surviving rollout (`rollout.n=1`, or groups thinned by filtering/failed rollouts). This PR aligns the vectorized variant with the loop reference rather than the other way round, so no existing `rloo` run changes behavior.

**Why the existing tests missed both**

- `test_as_torch_index_basic_integers` asserts only `len(torch.unique(g)) == 3` while its comment claims "Values should be contiguous 0..G-1" — the assertion passes on non-dense output.
- `_make_group_index` is documented as "ensuring each group has at least 2 samples" (`assert num_groups * 2 <= batch_size`), so `test_rloo_and_vectorized_equivalence` structurally cannot reach the branch where the two RLOO implementations diverge.

Both are tightened here, plus new regression tests for dense ids across every label kind, sparse/negative uids end-to-end through both vectorized estimators, and RLOO singleton-group parity.

### Checklist Before Starting

- [x] Search for similar PRs. Query link: https://github.com/verl-project/verl/pulls?q=is%3Apr+as_torch_index — plus:
  - `gh pr list --repo verl-project/verl --state open --search "as_torch_index"` → only #4933 (unrelated feature)
  - `--search "groupwise"` → none
  - `--search "group_mean_std"` → #6452, #4677, #4933
  - `--search "rloo_vectorized"` / `"rloo vectorized"` → #6452
  - `gh issue list --repo verl-project/verl --state all --search "as_torch_index"` / `"groupwise"` → no open issue

  Not a duplicate: #6452 fixes GRPO baselines contaminated by fully-rejected rollout rows (a different defect in `compute_grpo_outcome_advantage`); #4677 *adds* new vectorized estimators and does not touch `as_torch_index`, `group_mean_std`, or the body of `compute_rloo_vectorized_outcome_advantage` — it also claims its new estimators are "numerically equivalent to their loop-based counterparts", which is the property this PR makes actually enforceable, so the two changes are complementary rather than conflicting.

### Test

CPU-only; no GPU required.

```bash
pytest tests/utils/test_groupwise.py tests/trainer/ppo/test_core_algos_on_cpu.py -q
# 48 passed

# reverting only the two source files (tests kept) reproduces both defects:
# 16 failed, 32 passed
```

Wider regression sweep on the CPU-runnable subset (macOS, torch 2.12, CPU only):

```bash
pytest tests/trainer tests/utils -q \
  --ignore=tests/utils/megatron \
  --ignore=tests/utils/ckpt/test_megatron_checkpoint_manager_on_cpu.py \
  --ignore=tests/utils/test_shared_memory.py \
  --ignore=tests/utils/test_special_megatron_kl_loss_tp.py \
  --ignore=tests/utils/test_special_mstx_profile.py \
  --ignore=tests/utils/test_vllm_weight_name_normalization_on_cpu.py \
  --ignore=tests/utils/veomni
# 79 failed, 568 passed, 37 skipped
```

The 79 failures are byte-for-byte identical to the same command on unmodified `main` in this environment (they need CUDA/NCCL, vLLM, SGLang, or wandb/mlflow backends that are unavailable on a CPU-only macOS box); the failure lists were diffed with `comm` and this PR introduces none and fixes none of them.

Lint:

```bash
ruff check   verl/utils/groupwise.py verl/trainer/ppo/core_algos.py tests/utils/test_groupwise.py tests/trainer/ppo/test_core_algos_on_cpu.py  # All checks passed!
ruff format --check <same files>  # 4 files already formatted
```

### API and Usage Example

No public API or config change. `as_torch_index` returns dense ids for the label kinds listed above; since group labels only carry identity (the values were never meaningful downstream — `group_mean_std` uses them purely as positions), no caller semantics change.

### Design & Code Changes

- `verl/utils/groupwise.py`: new `_densify()` helper applied to every integer-recognizing path in `as_torch_index`; `group_mean_std` computes bounds with a single `torch.aminmax` and rejects negative indices with a message pointing at `as_torch_index`.
- `verl/trainer/ppo/core_algos.py`: `compute_rloo_vectorized_outcome_advantage` uses `torch.where(c > 1, leave_one_out, scores)` to match the loop reference, and canonicalizes labels through `as_torch_index` (same helper `grpo_vectorized` already uses) instead of an inline `np.unique`, which also makes it accept torch-tensor uids.

- [x] Read the [Contribute Guide](https://github.com/verl-project/verl/blob/main/CONTRIBUTING.md) and [AGENTS.md](https://github.com/verl-project/verl/blob/main/AGENTS.md).
- [x] Apply [pre-commit checks](https://github.com/verl-project/verl/blob/main/CONTRIBUTING.md#code-linting-and-formatting).
- [x] Add / update the [documentation](https://github.com/verl-project/verl/tree/main/docs) — n/a, no user-facing config or API change; the corrected behavior is described in the module/function docstrings.
- [x] Add unit or end-to-end test(s) — CPU tests added to the existing `tests/utils/test_groupwise.py` and `tests/trainer/ppo/test_core_algos_on_cpu.py`, both already covered by the CPU unit-test workflow.
- [ ] Slack `ci-request` message — will send once the PR is ready for CI.
- [x] **AI assistance was used** for this change (Claude and OpenAI Codex). The assistance and verification commands are disclosed above.\n- [ ] I have personally reviewed every changed line and can explain and defend the change end-to-end.


keepkeen:
recheck

PR #7160, opened 2026-07-27
[ray, single_controller] fix: sort placement groups by numeric IP so rank order follows node order

weich97:
### What does this PR do?

`sort_placement_group_by_node_ip` orders placement groups — and therefore global ranks — by the **string** form of each node's IP, so on a mixed-width allocation the rank order stops following the node order. This PR sorts by the parsed address instead. On equal-width allocations the result is byte-identical to today's order, so it is a no-op there.

Related open work: #4792 (IP-based worker sorting). Upstream PyTorch is making the same default change in `torchelastic`: pytorch/pytorch#191285 (approved by a repository collaborator) and the issue it came from, pytorch/pytorch#191190.

### Checklist Before Starting

- [x] Search for similar PRs. Paste at least one query link here: https://github.com/verl-project/verl/pulls?q=is%3Apr+placement+group+ip+sort — closest is #4792.

### The problem

`verl/single_controller/ray/base.py`:

```python
return sorted(pgs, key=lambda pg: pg_ip[pg.id])
```

Because the last octet is not zero-padded, string order and numeric order diverge whenever the allocated nodes span an octet-width boundary. For a 16-node job on `10.0.0.2…9` plus `10.0.0.66…73`, the resulting rank order is

```
2, 3, 4, 5, 6, 66, 67, 68, 69, 7, 70, 71, 72, 73, 8, 9
```

Node `.7` ends up isolated between two blocks of the other group. Ranks that are adjacent in ring collectives land on nodes that are far apart in the physical fabric.

### Test

**Unit tests (CPU only, no cluster):** `tests/single_controller/test_ip_sort_key_on_cpu.py` covers mixed-width numeric ordering, byte-identical behaviour on equal-width allocations, non-IPv4 fallback ordering, and malformed input.

**Reproduction of the underlying behaviour, no cluster needed:**

```python
ips = ["10.0.0.7", "10.0.0.66", "10.0.0.8", "10.0.0.70"]
sorted(ips)                                            # ['10.0.0.66', '10.0.0.7', '10.0.0.70', '10.0.0.8']
sorted(ips, key=lambda s: tuple(int(o) for o in s.split(".")))
                                                       # ['10.0.0.7', '10.0.0.8', '10.0.0.66', '10.0.0.70']
```

**Measured impact on a cluster:** on a 16-node MoE RL post-training job (GRPO, Qwen3-Coder-30B-A3B, EP=128, TP=4, Ascend 910C/HCCL), comparing the current string order against a contiguous order on the same nodes, everything else identical: **pure rank-order effect +8.4% step time** (weight-refit bucketing off) / **+9.3%** (on); 2×2 factorial, 5 epochs per cell, steady-state steps 3–12. The cost is a roughly constant seconds-per-step tax (it did not grow when mean generated length went 193→677 tokens).

Scoping this honestly: one interconnect, one model, one framework version. What transfers by code inspection is the permutation itself and the fact that ring collectives derive adjacency from rank order. Zero-padded / equal-width naming is immune; AWS IP-based private DNS names are not zero-padded and are mandatory on IPv4-only subnets.

### API and Usage Example

No API change: no CLI argument, config field, or function signature is added or altered. The only observable difference is the order in which placement groups are returned, and only for mixed-width allocations.

```python
# before: rank order follows the string form of the address
sorted(["10.0.0.9", "10.0.0.10"])          # ['10.0.0.10', '10.0.0.9']
# after: rank order follows the address
sorted(["10.0.0.9", "10.0.0.10"], key=_ip_sort_key)   # ['10.0.0.9', '10.0.0.10']
```

### Design & Code Changes

One module-level helper, so it is importable and unit-testable, plus its use as the sort key. It sorts by the parsed 4-tuple and falls back to string comparison for anything that is not a dotted quad (IPv6, hostnames), so non-IPv4 configurations keep today's order:

```python
def _ip_sort_key(ip: str):
    parts = ip.split(".")
    if len(parts) == 4:
        try:
            return (0, tuple(int(p) for p in parts), "")
        except ValueError:
            pass
    return (1, (), ip)
```

**Backward compatibility.** Equal-width allocations: order is byte-identical to today's. Mixed-width allocations: the order changes, which is the fix — it changes which node gets which rank, so jobs resuming from checkpoints that assumed the old order should restart instead. Happy to gate this behind a flag if maintainers prefer a staged rollout.

- [ ] Apply [pre-commit checks](https://github.com/verl-project/verl/blob/main/CONTRIBUTING.md#code-linting-and-formatting). Left unchecked deliberately rather than assumed: the branch is unchanged since submission and I have not re-run the hooks against current `main`. Say the word and I will re-run and push any formatting fix.
- [ ] Add / Update [the documentation](https://github.com/verl-project/verl/tree/main/docs). Not applicable as far as I can tell: there is no user-facing API or config change, and no page in `docs/` describes placement-group ordering. Glad to add one if you would like the behaviour documented.
- [x] Add unit or end-to-end test(s): `tests/single_controller/test_ip_sort_key_on_cpu.py`, CPU-only, no cluster required.
- [ ] Send a message in the `ci-request` Slack channel. This is the step I missed when opening the PR, which is why CI has never run here; it is being sent now.

### DIFF
--- a/tests/single_controller/test_ip_sort_key_on_cpu.py
+++ b/tests/single_controller/test_ip_sort_key_on_cpu.py
@@ -1,0 +1,46 @@
+# Copyright 2024 Bytedance Ltd. and/or its affiliates
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from verl.single_controller.ray.base import _ip_sort_key
+
+
+def test_mixed_width_ips_sort_numerically():
+    ips = ["10.0.0.7", "10.0.0.66", "10.0.0.8", "10.0.0.70"]
+    assert sorted(ips, key=_ip_sort_key) == ["10.0.0.7", "10.0.0.8", "10.0.0.66", "10.0.0.70"]
+
+
+def test_equal_width_order_is_byte_identical_to_lexicographic():
+    ips = ["10.0.0.101", "10.0.0.100", "10.0.0.103", "10.0.0.102"]
+    assert sorted(ips, key=_ip_sort_key) == sorted(ips)
+
+
+def test_full_octet_boundaries():
+    ips = ["192.168.1.5", "9.0.0.1", "10.42.13.2", "10.42.9.66", "10.42.9.7"]
+    assert sorted(ips, key=_ip_sort_key) == [
+        "9.0.0.1",
+        "10.42.9.7",
+        "10.42.9.66",
+        "10.42.13.2",
+        "192.168.1.5",
+    ]
+
+
+def test_non_ipv4_sorts_after_ipv4_and_keeps_string_order():
+    mixed = ["nodeb", "10.0.0.9", "nodea", "10.0.0.10", "fe80::1"]
+    assert sorted(mixed, key=_ip_sort_key) == ["10.0.0.9", "10.0.0.10", "fe80::1", "nodea", "nodeb"]
+
+
+def test_malformed_dotted_quad_falls_back_to_string():
+    vals = ["10.0.0.x", "10.0.0.2"]
+    assert sorted(vals, key=_ip_sort_key) == ["10.0.0.2", "10.0.0.x"]
--- a/verl/single_controller/ray/base.py
+++ b/verl/single_controller/ray/base.py
@@ -84,7 +84,19 @@ def sort_placement_group_by_node_ip(pgs: list[PlacementGroup]) -> list[Placement
         # all bunles should be on the same node
         node_id = specs["bundles_to_node_id"][0]
         pg_ip[pg.id] = node_ip[node_id]
-    return sorted(pgs, key=lambda pg: pg_ip[pg.id])
+    def _ip_sort_key(ip: str):
+        # Sort dotted-quad IPs numerically; lexicographic order breaks whenever the
+        # allocation spans an octet-width boundary (e.g. .9 vs .66), which lands
+        # ring-adjacent ranks on physically distant nodes. Non-IPv4 keeps string order.
+        parts = ip.split(".")
+        if len(parts) == 4:
+            try:
+                return (0, tuple(int(p) for p in parts), "")
+            except ValueError:
+                pass
+        return (1, (), ip)
+
+    return sorted(pgs, key=lambda pg: _ip_sort_key(pg_ip[pg.id]))
 
 
 @ray.remote
--- a/verl/single_controller/ray/base.py
+++ b/verl/single_controller/ray/base.py
@@ -67,6 +67,23 @@ def func_generator(self, method_name, dispatch_fn, collect_fn, execute_fn, block
     return type(method_name, (Functor,), {})()
 
 
+def _ip_sort_key(ip: str):
+    """Sort key that orders dotted-quad IPs numerically.
+
+    Lexicographic string order breaks whenever an allocation spans an octet-width
+    boundary (e.g. ``.9`` vs ``.66``), which lands ring-adjacent ranks on
+    physically distant nodes. Non-IPv4 addresses keep plain string order and sort
+    after all IPv4 addresses, so mixed deployments stay deterministic.
+    """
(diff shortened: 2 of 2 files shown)


weich97:
Update: moved `_ip_sort_key` to module level and added CPU unit tests (`tests/single_controller/test_ip_sort_key_on_cpu.py`) covering mixed-width numeric ordering, byte-identical behavior on equal-width allocations, non-IPv4 fallback ordering, and malformed input.

Also for context: PyTorch is adopting the same change **by default** in torchelastic's rendezvous — a maintainer confirmed the current lexicographic behavior "is only optimal when you zero-pad the node names and won't regress by switching to natural ordering" (pytorch/pytorch#191190), with the implementation in pytorch/pytorch#191285. Since verl reaches the same rank order through this IP-string sort, the two fixes are complementary.

Happy to gate behind a flag or adjust per maintainer preference.

weich97:
Gentle ping @wuxibin89 @vermouth1992 — would appreciate a review when you get a chance. CI is green, CLA is signed, and the change is intentionally minimal: a module-level `_ip_sort_key` that is byte-identical on equal-width allocations, plus CPU-only unit tests.

One update for context: the companion PyTorch PR (making natural ordering the default in torchelastic's rendezvous, filed at the torchelastic maintainer's invitation) has since received reviewer approval and is awaiting final maintainer sign-off — landing this would keep the two defaults consistent.

weich97:
@zw0610 @wuxibin89 @hongpeng-guo — following up, and first correcting something on my side rather than yours.

I had missed two requirements in the PR template: the title was not in the `[{modules}] {type}: {description}` form the CI checks (now fixed), and the PR body did not follow the template, including the submission checklist. Both are on me, and both are now addressed.

The remaining step I could not complete is the CI request in the `ci-request` Slack channel, which I believe is why this has stayed unchecked rather than anyone having objected to it. I will send that.

For scope: one module-level `_ip_sort_key` in `verl/single_controller/ray/base.py`, ordering placement groups by numeric IP rather than by string. On equal-width allocations the order is byte-identical to today's, so it is a no-op there; it differs only when address widths differ (`10.0.0.9` vs `10.0.0.10`), which is the case that currently interleaves nodes across chassis. Tests are CPU-only (`tests/single_controller/test_ip_sort_key_on_cpu.py`). Related open work: #4792. PyTorch is making the same default change in `torchelastic` (pytorch/pytorch#191285, collaborator-approved).

weich97:
Correction to my previous comment: I cannot send the `ci-request` message. The verl Slack workspace only admits accounts on the `anyscale.com`, `bytedance.com` and `together.ai` domains, and I do not have one; the Feishu link in the PR template appears to be from an older revision.

So the request is here instead. This PR has no check runs at all — not even the always-triggered ones (`check-pr-title`, `pre-commit`, `doc`) — which looks like the first-time-contributor workflow-approval gate rather than anything about the change itself. Could one of you click **Approve and run** on the workflows?

For what it is worth, the title now matches the `[{modules}] {type}: {description}` form that `check-pr-title.yml` enforces, and the new test carries the `_on_cpu.py` suffix so it is picked up by `cpu_unit_tests.yml`.

weich97:
@zw0610 @wuxibin89 @hongpeng-guo following up on this one. The blocker is not the code:

- The patch replaces the string sort of node IPs in `base.py` with a numeric key, and adds a CPU-only unit test. Sorting `10.0.0.1` through `10.0.0.16` as text interleaves them, which doubles cross-chassis ring edges on a two-chassis split.
- No CI has ever run on this branch. As I noted on 2026-08-16, I cannot send the `ci-request` message myself: the verl Slack workspace only admits accounts on specific corporate domains, so that route is closed to outside contributors like me.

Could a maintainer trigger CI, or point me at the current preferred route for contributors who are not in that Slack? Happy to rebase or adjust the patch if anything else is needed.

weich97:
Scope note on one measurement claim in the PR description, so that it does not conflict with the technical report that accompanies this work.

The description above says the ordering cost is "a roughly constant seconds-per-step tax (it did not grow when mean generated length went 193 to 677 tokens)". That statement applies to the probe set it was written from, and should not be read as a general conclusion.

Later controlled probes, pre-registered before the data existed, point the other way. On two chassis pairs the absolute ordering cost was 1.91x and 1.73x as large under the longer-generation condition. The report treats this as an association and not as a token-count causal effect, because workload, response cap and generated length move together across those endpoints and the design does not separate them; one earlier cap-only probe failed its own denominator self-check and its ratio was struck.

So the accurate summary is: the ordering cost is not established as constant in generated length, and the direction of the association in the controlled probes is that longer generation carries a larger absolute cost.

None of this changes the patch. The defect it fixes is that node IP addresses are compared as text, so a mixed-width allocation is delivered in byte order rather than numeric order; that is a correctness-of-ordering issue independent of how the resulting cost scales with workload.

PR #7161, opened 2026-07-27
[fsdp] refactor: move unfuse_moe_params to FSDP backend

wuxibin89:
### What does this PR do?

Move unfuse_moe_params from vllm to FSDP backend


verl/workers/engine/fsdp/transformer_impl.py
```diff
@@ -969,16 +969,14 @@ def get_per_tensor_param(self, layered_summon=False, base_sync_done=False, **kwa
...
         else:
             device = get_device_id()  # used when fsdp2 set cpu_offload_policy
-            # TODO: cast fp32 to bf16 to reduce weight sync overhead, need more fine-grained control, e.g MoE gate
             per_tensor_param = (
                 (
                     name,
-                    param.to(device, non_blocking=True).full_tensor().to(torch.bfloat16, non_blocking=True)
-                    if isinstance(param, DTensor)
-                    else param,
+                    param.to(device, non_blocking=True).full_tensor() if isinstance(param, DTensor) else param,
```
uqyxx, line 975: Hi @wuxibin89 , in verl/trainer/config/engine/fsdp.yaml, the default model_dtype is fp32, if "to(torch.bfloat16)" is removed here, the model will be in fp32 by default. Maybe should change the default setting to bf16 or let users know that we need to  pass this param explicitly in the script now, otherwise we may unintentionally run a different configuration.

PR #7181, opened 2026-07-28
[megatron] feat: delta_sharded on Megatron-Bridge param mappings (TP+EP, hybrid-Mamba)

ChangyiYang:
### What does this PR do?

The mcore backend joins the sharded-delta contract (#7144). #7085 (veomni EP) is merged; this branch is rebased onto latest main, so the diff is the megatron/engine increment only. Everything rides Megatron-Bridge's own per-param machinery — no legacy converters, no hand-written index math:

- **Export index**: `AutoBridge.get_conversion_tasks` enumerates parameters; each becomes a `ShardSpec` (TP partition dims → `BlockPlacement` over the TP group; expert tensors → virtual `[ep, *etp]` blocks over the etp×ep group; replicated → the engine's rank-0 direct path). No hand-derived geometry: the engine only consumes `gather_group` + the replicated-rank0 rule.
- **Comm-stub probe**: each mapping's `megatron_to_hf` runs on a probe copy whose process groups are `_ProbeGroup`s (real `size()`/`rank()`, any collective raises) and whose gather helpers are stubbed to pure local semantics — NaN sentinels through the REAL converter code give final HF coordinates with zero collectives and zero converter forks. This keeps converters with real value math (Mamba's TP de-interleave, GQA fusions) bit-exact, which the earlier "degenerate world" probe could not.
- **Idempotence verify sweep** (`verify_every=K` engine kwarg): every K-th sync appends a full re-export in the same receive session; the rollout loader snapshots every `copy_` destination before its real load path, re-applies, and fails loud on any bitwise divergence — a production-grade end-to-end oracle that tolerates multi-stage loaders (e.g. sglang's composed A_log loader).
- **Perf**: shape-pooled NaN probe buffers (Opt-A) and an empty-delta short-circuit reusing the slot cache (Opt-B); snapshots default to pageable host memory on mcore (`delta_pin_snapshots = False` class attr — pinning whole shard sets surfaces as CUDA OOM at 30B/235B scale).

Scope: TP+EP+ETP, PP=1/VPP=1, no LoRA, Megatron-Bridge flavor only.

### Test

| check | result |
|---|---|
| D1 equivalence (tiny Qwen2, real bridge, 1 GPU) | probe entries == bitwise diff of the bridge's own full HF export, 22/22 HF tensors |
| **TP2 differential oracle** (`tests/special_distributed/test_mcore_probe_differential.py`) | real `megatron_to_hf` vs merged probe outputs, bitwise, on qwen2 / qwen3_moe / **nemotron_h (Mamba)** / falcon_h1 |
| **7B TP2** e2e | uw 3.8–4.9s steady; idempotence sweeps (v=2/4/6) zero mismatch |
| **30B-A3B TP1×EP8 + moe_grouped_gemm** | uw 14.5–17.6s steady (hand-written-converter baseline: 17.2–19.7s) |
| **235B-A22B TP4×EP16×PP1 + grouped_gemm** (16 trainer GPUs + TP16 rollout) | seed 107.6s/437.9GB; steady uw **23.6/23.7s**, payload ~20GB/step; same-shape NCCL full sync: **235.0–237.6s/step → delta is 8.2–9.7×**, seed < half one NCCL step |
| **hybrid-Mamba e2e** (NVIDIA-Nemotron-Nano-9B-v2, TP1 and TP2) | 6 steps + 3 full idempotence sweeps (3.39B elems each) zero mismatch; 50-step GRPO W&B pair mechanically green |
| Opt-A regression | comm-stub refactor cost +25–40% at 235B; pooled buffers recovered to 23.6/23.7s (baseline 24.3–29.5s) |
| Opt-B | empty-delta short-circuit validated 7B TP2 e2e + sweeps |
| CPU suites | tests/checkpoint_engine 32 passed |

### Additional Info

- Perf follow-ups (not this PR): closed-form index tables for the hottest mapping families (#7060), PP steady relay.

### DIFF
--- a/tests/checkpoint_engine/test_sglang_loader.py
+++ b/tests/checkpoint_engine/test_sglang_loader.py
@@ -178,3 +178,52 @@ def test_dense_flush_applies_full_tensors():
 
     for name, expected in named:
         assert torch.equal(model.params[name].view(torch.int16), expected.view(torch.int16)), name
+
+
+def _dense_verify_flush(named, is_last=True, verify=True):
+    params, pieces, val_off = [], [], 0
+    for name, t in named:
+        flat = t.contiguous().view(-1)
+        params.append(
+            {
+                "name": name,
+                "dtype": str(t.dtype).replace("torch.", ""),
+                "shape": list(t.shape),
+                "pos_start": 0,
+                "pos_end": 0,
+                "pos_width": 4,
+                "val_start": val_off,
+                "val_end": val_off + flat.numel(),
+            }
+        )
+        pieces.append(flat)
+        val_off += flat.numel()
+    values = torch.cat(pieces)
+    spec = {
+        "encoding": "dense",
+        "verify": verify,
+        "is_last": is_last,
+        "params": params,
+        "checksum": int(checksum(torch.empty(0, dtype=torch.uint8), values)),
+    }
+    spec_t = torch.frombuffer(bytearray(json.dumps(spec).encode()), dtype=torch.uint8)
+    return [("__delta_spec__", spec_t), ("__values__", values)]
+
+
+def test_verify_sweep_passes_on_identical_state():
+    """A verify flush against a bit-identical model reports zero mismatches."""
+    named = _make_named()
+    model = _FakeModel([(n, t.clone()) for n, t in named])
+    apply_delta(model, _dense_verify_flush(named))  # must not raise
+
+
+def test_verify_sweep_fails_loud_on_divergence():
+    """A single flipped element in the server state must fail the sweep."""
+    import pytest
+
+    named = _make_named()
+    diverged = [(n, t.clone()) for n, t in named]
+    diverged[1][1].view(-1)[3] += 1.0
+    model = _FakeModel(diverged)
+    with pytest.raises(RuntimeError, match="verification FAILED"):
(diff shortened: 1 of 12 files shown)


ISEEKYAN:
hi @ChangyiYang, do we have a protocol file defining the format of delta, I prefer a training backend agnostic protocol (like HF safetensor for full parameter) and difference backends have their impl

ChangyiYang:
> hi @ChangyiYang, do we have a protocol file defining the format of delta, I prefer a training backend agnostic protocol (like HF safetensor for full parameter) and difference backends have their impl

hi @ISEEKYAN Yes, that's exactly how it works today: each training backend implements get_per_tensor_param_delta_shard, whose output format is the protocol — final HF-coordinate sparse entries, defined in verl/workers/engine/spec.py (wire encoding: delta_checkpoint_engine.py header). Everything past that boundary — gather, bucketing, the receiver — is backend-agnostic, and each impl fully controls how it produces its delta (snapshot diff, converters, or even diffing against a stored checkpoint).

wuxibin89 (approved):

wuxibin89:
@ChangyiYang  Does pipeline parallelism supported?

ChangyiYang:
@wuxibin89 PP is left as unsupported and will raise immediately. @gxlvera will support it

PR #7182, opened 2026-07-28
[fsdp] fix: make ref model CPUOffload configurable instead of hardcoded

Shadowking912:
### What does this PR do?

Fixes an issue where FSDP hardcodes `CPUOffload(offload_params=True)` for all non-actor (reference/reward) models, ignoring the user-configured `param_offload` setting in `verl/workers/engine/fsdp/transformer_impl.py`.

This forces CPUΓåöGPU parameter transfers on every reference forward micro-batch, even when the model fits entirely in GPU memory, leading to significant performance overhead (40-70x slower ref forward than actor forward for a 4B model).

**Motivation:** When a reference/reward model is small enough to fit in GPU memory (e.g., 4B parameters on H100 with 140GB), CPU offloading becomes pure overhead. The hardcoded offload forces unnecessary parameter swaps between CPU and GPU on every forward pass, causing severe latency degradation. By making this configurable, users can disable offloading when their ref model comfortably fits in VRAM, recovering 2-3x speedups in training step time.

**Changes:**
- **FSDP v1 (line 409):** Respect `param_offload` config when `forward_only=True` for reference models.
- **FSDP2 (line 442):** Apply the same configurable behavior to `CPUOffloadPolicy` for reference models.

The existing default behavior for actors is preserved. Users can now disable parameter offloading for reference/reward models by setting:

```yaml
actor_rollout_ref:
  ref:
    fsdp_config:
      param_offload: false
```

This keeps the reference/reward model weights on GPU when sufficient GPU memory is available.

**Observed impact (Gemma-3-4B, batch size 32, 1├ù H100 140GB):**
- Ref forward (total): **63ΓÇô95 s ΓåÆ 1.1 s**
- Ref per micro-batch: **8.7ΓÇô25.4 s ΓåÆ 0.2 s**
- Step time: **157 s ΓåÆ 59 s**
- Overall training step: **2.7├ù speedup** with no increase in peak GPU memory usage.

### Checklist Before Starting

- [x] Search for similar PRs: Verified no existing PRs address CPU offload configurability for reference models
  - Modules: `fsdp`
  - Type: `fix`
  - Title: `[fsdp] fix: Make ref model CPUOffload configurable instead of hardcoded`

### Test

Validated on Gemma-3-4B by disabling parameter offloading:

```yaml
actor_rollout_ref:
  ref:
    fsdp_config:
      param_offload: false
```

Verified that:
- Reference model parameters remain on GPU
- Reference forward latency drops from 8.7-25.4s to 0.2s per micro-batch
- Training proceeds correctly with no increase in peak GPU memory usage (stays at 140GB)
- Gradient computation and optimization continue normally

### API and Usage Example

No API changes. The existing `param_offload` configuration field is now respected for reference/reward models.

Users can now configure parameter offloading for reference/reward models:

```yaml
actor_rollout_ref:
  ref:
    fsdp_config:
      param_offload: false  # Disable CPU offload for ref model (keeps weights on GPU)
```

For FSDP reference/reward models, `param_offload` defaults to `True` when it is not specified. This preserves the existing behavior of offloading parameters to CPU.

Setting `param_offload: false` keeps reference/reward model parameters resident on the GPU and avoids CPU-GPU parameter transfers. This can improve performance when the model fits comfortably in GPU memory, but it increases peak GPU memory usage and may cause out-of-memory errors if the model does not fit alongside the rest of the workload. Users should opt out only after verifying available GPU memory.

For actor models, the default remains `param_offload: false`.

### Design & Code Changes

**File modified:** `verl/workers/engine/fsdp/transformer_impl.py`

**Changes:**
1. **FSDP v1 strategy (lines 400-411):**
   - Before: `if self.engine_config.forward_only: cpu_offload = CPUOffload(offload_params=True)`
   - After: `if self.engine_config.forward_only and self.engine_config.param_offload: cpu_offload = CPUOffload(offload_params=True)`

2. **FSDP2 strategy (lines 428-441):**
   - Before: `if self.engine_config.offload_policy or self.engine_config.forward_only: offload_policy = CPUOffloadPolicy(...)`
   - After: `if self.engine_config.offload_policy or (self.engine_config.forward_only and self.engine_config.param_offload): offload_policy = CPUOffloadPolicy(...)`

**Rationale:**
- Replaces hardcoded CPU offloading for forward-only (reference/reward) models with configuration-driven behavior
- Respects the existing `param_offload` field (already present in `EngineConfig`) for both FSDP v1 and FSDP2
- Preserves existing default behavior for actors (no CPU offload due to gradient accumulation issues)
- Maintains backward compatibility by making ref model offload configurable without breaking changes

> [!IMPORTANT]

- [x] Read the [Contribute Guide](https://github.com/verl-project/verl/blob/main/CONTRIBUTING.md)
  - Followed contribution guidelines; single focused change to FSDP configuration behavior

  - Pre-commit checks to be run before final submission

- [x] Add / Update the [documentation](https://github.com/verl-project/verl/tree/main/docs)
  - Configuration documentation updated to clarify `param_offload` behavior for reference models in FSDP config
  - Updated comments in `verl/workers/engine/fsdp/transformer_impl.py` to reflect configurable offload behavior

- [x] Add unit or end-to-end test(s) to the CI workflow
  - Not feasible: This change modifies existing FSDP configuration handling for reference models. It was validated through end-to-end training on Gemma-3-4B with measured performance improvements (2.7├ù step speedup). Full unit test coverage would require instantiating FSDP engines with various configurations, which is environment-dependent. The fix respects existing configuration fields already validated by existing test suites. Users can validate locally by setting `param_offload: false` in their FSDP config and observing ref forward latency improvements.

- [ ] Once your PR is ready for CI, send a message in the `ci-request` channel in the `verl` Slack workspace
  - Message to be sent: "PR #XXX ready for CI: [fsdp] fix: Make ref model CPUOffload configurable instead of hardcoded"

- [x] Recipe submodule: N/A (this PR does not modify the recipe submodule)


ErenAta16 (commented):
Making this configurable is reasonable, and gating on the flag the user already sets is the right mechanism. The part that needs to be deliberate is the default, because this silently increases GPU memory for every existing config.

`param_offload` defaults to `False`:

```python
# whether to offload param
param_offload: bool = False
```

so after this change a reference model only offloads when someone has explicitly set `ref.param_offload=True`. Before, `forward_only` alone was enough and every ref model offloaded unconditionally, which is what the comment being replaced said: "We force reference policy to use CPUOffload to save memory."

The reference policy in RLHF is a full copy of the policy weights. Keeping it resident instead of offloading is not a marginal change to the memory profile, and the people affected are exactly the ones who never touched `param_offload` because they never had to. A config that fit yesterday can OOM after this lands, and the traceback will point at FSDP allocation rather than at a default that moved.

Three ways out, in the order I would prefer them:

Default `param_offload` to `True` for `forward_only` engines specifically, so the new knob turns offload *off* for people who want it off and nobody who did nothing sees a change. That keeps the fix, which is really "let me disable this", and inverts the blast radius onto the opt-in group.

Or keep the default as written and say so loudly: a release note, and ideally a one-time warning when `forward_only and not param_offload` the first time a ref model is built, naming the flag. A silent memory regression is the worst version of this.

Or scope the change to the reward model if that is where the need came from, leaving the reference policy on its old behaviour.

Worth stating in the description which case motivated it. If someone has a ref model small enough that offloading is pure overhead, that is a good reason for the knob, and it also tells the reviewer whether the default should follow.

The FSDP1 and FSDP2 branches are changed consistently, and the comment updates match the new conditions in both, so the mechanics look right. This is only about which side of the switch the untouched configs land on.

Shadowking912:
Thanks for the feedback on backward compatibility. I have implemented **Option 1** (your preferred approach): defaulting `param_offload=True` for ref models to preserve existing behavior, while letting users opt-in to performance by setting `param_offload=False`. This inverts the blast radius to the opt-in group—no silent memory regressions for existing configs, and users with models that fit in GPU can recover the speedup. The PR already details the implementation (FSDPEngineConfig defaults logic, config changes to Optional[bool], FSDP v1/v2 offload checks) and passes all code quality checks. I have also updated the motivation in the description.

ErenAta16:
Late coming back to this, sorry. Defaulting `param_offload=True` for ref models is the right shape: it keeps every existing config on the behaviour it already has and puts the memory risk on the people who deliberately opt out, which is the group that can reason about whether their model fits.

The part I would put in the description if it is not there already is that second half stated plainly, that opting out can raise peak memory rather than only saving time. A default that is safe is easy to review; a documented opt-out that names its cost is what stops the first out-of-memory report from being a surprise.

I should be straight about the limit of this: I read the change, I have not run your branch, and I have no multi-GPU box to check the offload path on, so treat this as a design read and not a verification. The reviewer who can run it is the one who should sign it off.

Shadowking912:
Thanks, agreed. I updated the PR description to state explicitly that setting
`param_offload=False` for reference/reward models keeps their parameters on the
GPU, which can increase peak GPU memory usage and may cause OOMs if the model
does not fit alongside the rest of the workload.

The default remains enabled for forward-only FSDP models to preserve existing
behavior. I also added focused CPU configuration coverage for the default and
explicit override cases; it passes with 12 tests passing.

I agree that this does not replace multi-GPU runtime verification. The FSDP1
and FSDP2 offload paths still need validation from a reviewer with suitable
multi-GPU hardware.

ErenAta16:
You did add it, and I checked the description rather than taking the summary:

> Setting `param_offload: false` keeps reference/reward model parameters resident
> on the GPU and avoids CPU-GPU parameter transfers. This can improve performance
> when the model fits.

That is the condition stated plainly, which is the important half. The one thing still implied rather than named is what happens when it does not hold. Your two measurement lines both say "no increase in peak GPU memory usage", which is true of the 4B model you ran and is exactly the reason someone skims past the caveat: the numbers say memory is free, and the sentence that says it is conditional sits somewhere else. A reader with a 70B reference model needs the other half of the
sentence, that opting out makes the reference parameters permanently resident and can push a configuration that fits today into an out-of-memory failure.

One clause is enough. It costs nothing now and saves the first "this used to work" report later, which is the one that gets a default reverted rather than a config documented.

Nothing else from me. The default staying at `True` is the part that matters and you got that right from the start.

Shadowking912:
Thanks for the suggestion. Will add this also. Also how to get someone to test/merge this PR? Since you said you don't have a multi gpu setup to verify this.

ErenAta16:
Fair question, and it has a more concrete answer than "wait", so I went and
measured rather than guessed.

**The routing already worked.** CODEOWNERS line 17 is
`/verl/workers/engine/fsdp @eric-haibin-lin @vermouth1992 @ZihengJiang`, this PR
touches `verl/workers/engine/fsdp/transformer_impl.py`, and GitHub has already
requested `eric-haibin-lin` and `vermouth1992` on it. So it is not lost and there
is nobody left to notify. It is sitting in front of the two people who own that
path.

**The repo is not slow either**, which is the part that makes the wait odd. Eleven
PRs merged in the past week from eleven different outside authors, and their
numbers are #7764 through #7799. Yours is #7182 from 28 July, so branches opened
six hundred numbers later are passing it. That pattern usually means a PR is being
skipped over rather than queued behind anything.

**The one concrete difference I can see is the title.** Every merged PR in that
list uses the `[area] type:` form:

```
[ci] fix: nightlyCI_grpo_qwen3_5_2b_fsdp2_vl
[hardware] test: include AMD in platform auto-detection
[trainer, cfg] feat: filter V1 DAPO groups
[ckpt] fix: drain Mooncake completion slots
```

Yours is `Fix: Make ref model CPUOffload configurable instead of hardcoded`, with
no bracket. Your own description already opens with `## [fsdp] fix: ...`, so you
know the convention; it just never made it into the title, which is what a
maintainer scanning a list actually sees. Renaming to
`[fsdp] fix: make ref model CPUOffload configurable instead of hardcoded` costs
nothing and puts it in the same shape as everything that merges.

I checked labels too, in case that was the differentiator. It is not: the merged
PRs above carry no labels either, so do not spend effort there.

**On testing, I should be straight about my own limit.** I am on the requested
reviewer list here, and what I gave you was a design read. I have no multi-GPU
box, so I cannot run the offload path and will not pretend a review from me
substitutes for someone who can. The people who can are the two already
requested. If a rebase plus the title change does not draw them in a week or so,
asking in the project's community channel and linking the PR tends to work better
than another comment on the thread, because the thread only reaches people already
watching it.

Shadowking912:
Hi @eric-haibin-lin and @vermouth1992, could you please take another look at PR #7182 when you have time?

I updated the title to follow the repository convention and clarified in the description that `param_offload=False` for reference/reward models can increase peak GPU memory and cause OOMs.

The focused CPU configuration test passes with 12 tests passing. The remaining verification needed is a runtime check of the FSDP1 and FSDP2 offload paths on suitable multi-GPU hardware.

Thanks!

PR #4930, opened 2026-01-14
fix: Bypass Mode Configuration Timing Issue

zhangsheng377:
### What does this PR do?

> Add **concise** overview of what this PR aims to achieve or accomplish. Reference related GitHub issues and PRs that help with the review.
```log
RayPPOTrainer : fit : self.config.actor_rollout_ref.actor.policy_loss : {'_target_': 'verl.workers.config.PolicyLossConfig', 'loss_mode': 'bypass_mode', 'clip_cov_ratio': 0.0002, 'clip_cov_lb': 1.0, 'clip_cov_ub': 5.0, 'kl_cov_ratio': 0.0002, 'ppo_kl_coef': 0.1, 'rollout_correction': {'rollout_is': 'token', 'rollout_is_threshold': 2.0, 'rollout_rs': 'geometric', 'rollout_rs_threshold': 1.01, 'rollout_rs_threshold_lower': 0.99, 'rollout_token_veto_threshold': 0.0001, 'bypass_mode': True, 'loss_type': 'ppo_clip', 'rollout_is_batch_normalize': False}}
compute_policy_loss_bypass_mode : config : FSDPActorConfig(_target_='', strategy='fsdp', ppo_mini_batch_size=128, ppo_micro_batch_size=None, ppo_micro_batch_size_per_gpu=1, ppo_infer_micro_batch_size_per_gpu=None, use_dynamic_bsz=False, ppo_max_token_len_per_gpu=16384, ppo_infer_max_token_len_per_gpu=16384, clip_ratio=0.2, clip_ratio_low=0.2, clip_ratio_high=0.28, freeze_vision_tower=False, policy_loss=PolicyLossConfig(_target_='', loss_mode='bypass_mode', clip_cov_ratio=0.0002, clip_cov_lb=1.0, clip_cov_ub=5.0, kl_cov_ratio=0.0002, ppo_kl_coef=0.1, rollout_correction=None), 
```
After enabling the bypass mode, the `fit` function enters the `apply_bypass_mode` function. The logs show that the config is successfully modified inside `apply_bypass_mode`. However, the current code overlooks one point: the `loss_fn` is already initialized with parts of the config during the `init` phase, so modifying the config later in `fit` has no actual effect.

My solution is to separate the modification of the config from the modification of the batch within `apply_bypass_mode`, and perform the config modification during the `init` stage instead.

Fixes #4912

### Checklist Before Starting

- [ ] Search for similar PRs. Paste at least one query link here: ...
[...]

PR #7196, opened 2026-07-30
[trainer] feat: add opt-in rollout-vs-actor probs diagnostic for bypass mode

yueyiming2009:
### What does this PR do?

Adds an opt-in diagnostic that restores the rollout-vs-actor consistency metrics in **bypass mode**.

In bypass mode (`algorithm.rollout_correction.bypass_mode=true`) the trainer sets `old_log_probs = rollout_log_probs` and never recomputes the actor policy π_θ, so `calculate_debug_metrics` is never invoked and `training/rollout_actor_probs_pearson_corr` / `training/rollout_probs_diff_{max,mean,std}` are simply absent from the metrics. These are among the most useful async-RL diagnostics — they directly measure the training/inference gap and catch precision, tokenizer, and temperature drift that KL magnitude alone can miss. This is most acute for `PPOTrainerSeparateAsync`, which force-sets `bypass_mode = True` (`verl/trainer/ppo/v1/trainer_separate_async.py:70`) — i.e. the configuration with the *largest* expected train/infer gap currently has zero visibility into it.

The new flag `algorithm.rollout_correction.debug_rollout_actor_probs` (default `false`) makes the bypass branch of `_compute_old_log_prob` run **one extra no-grad `compute_log_prob` forward** purely for observability and feed the recomputed π_θ to the existing `calculate_debug_metrics`. `old_log_probs` is still set from `rollout_log_probs`, so the training computation is untouched and results are identical whether or not the diagnostic is enabled. The only cost is the extra inference forward — roughly what the non-bypass path already pays every step — and it lands inside the existing `marked_timer("old_log_prob")`, so it shows up in `timing_s/old_log_prob` rather than hiding.

### Checklist Before Starting

  - [`is:pr is:open bypass_mode`](https://github.com/verl-project/verl/pulls?q=is%3Apr+is%3Aopen+bypass_mode) → only #4930 (bypass mode config *timing*), unrelated.
  - [`is:pr is:open pearson`](https://github.com/verl-project/verl/pulls?q=is%3Apr+is%3Aopen+pearson) → only #4252, discussed below.
  - [`is:issue rollout_probs_diff`](https://github.com/verl-project/verl/issues?q=rollout_probs_diff) → no open issue requesting this.

  **Not a duplicate of #4252** ("use unscaled logits for accurate pearson correlation metric"). That PR changes *how* the existing metric is computed for the non-bypass path, by plumbing an unscaled `log_probs_for_metrics` through the actors. This PR does not touch the metric computation at all — it makes the existing, unmodified `calculate_debug_metrics` reachable from a code path where it currently never runs. The two are orthogonal and would not conflict. Worth noting for reviewers: #4252's premise (vLLM returns unscaled log-probs) was true when its issue #4162 was filed, but #4755 later added `rollout.logprobs_mode` defaulting to `processed_logprobs`, which makes vLLM's log-probs temperature-scaled and consistent with the actor recompute. #4162 is now closed.

- [x] Format the PR title as `[{modules}] {type}: {description}` → `[trainer] feat: add opt-in rollout-vs-actor probs diagnostic for bypass mode`

### Test

**Unit tests** — `tests/utils/debug/test_metrics.py` (picked up by `gpu_unit_tests.yml`):
**Smoke tests** -- run the separate async workflow and confirmed pcc chart showing up in mlflow.

```bash
pytest -s tests/utils/debug/test_metrics.py
# 4 passed
```

Three cases added, covering the metric this PR surfaces: identical rollout/actor log-probs → correlation `1.0`; perfectly anti-correlated probabilities → `-1.0`; all-False mask → `valid=0` and `nan` statistics. The existing `test_calculate_debug_metrics` now also asserts that all five debug metric keys are emitted.

**Honest scoping of the test coverage:** these tests exercise `calculate_debug_metrics`, which this PR does not modify. The new code path itself — `PPOTrainer._maybe_debug_rollout_actor_probs` — is not unit-tested, because reaching it requires a live `actor_rollout_wg` worker group and a TransferQueue partition; there is no existing fixture in the repo for driving `_compute_old_log_prob` in isolation. It was instead validated by GPU smoke test (below). Happy to add a mock-based test if a reviewer prefers.

**GPU smoke test** — ran an end-to-end bypass-mode job with the flag enabled and confirmed `training/rollout_actor_probs_pearson_corr` is emitted and charts correctly, where it was previously absent.

**Lint / repo checks** — `ruff check`, `ruff format --check`, `scripts/generate_trainer_config.sh` (regenerates byte-identical), `check_docstrings`, `check_docs_time_info`, `check_license`, `validate_structure` all pass.

### API and Usage Example

New config field, defaults to `false`, so existing configs are unaffected:

```yaml
algorithm:
  rollout_correction:
    bypass_mode: true
    debug_rollout_actor_probs: true   # NEW: opt-in diagnostic
actor_rollout_ref:
  rollout:
    calculate_log_probs: true         # required (already required by bypass mode)
```

Or from the CLI:

```bash
    algorithm.rollout_correction.bypass_mode=true \
    algorithm.rollout_correction.debug_rollout_actor_probs=true \
    actor_rollout_ref.rollout.calculate_log_probs=true \
    ...
```

Emitted metrics (identical to what the decoupled path already logs):

```
training/rollout_actor_probs_pearson_corr
training/rollout_probs_diff_valid
training/rollout_probs_diff_{max,mean,std}
```

### Design & Code Changes

- **`verl/trainer/ppo/v1/trainer_base.py`** — new `PPOTrainer._maybe_debug_rollout_actor_probs()`, called at the top of the bypass branch of `_compute_old_log_prob()`, *before* `old_log_probs` is written. It snapshots and restores `batch.extra_info` around the call so the diagnostic's flags (`calculate_entropy=False`, `compute_loss=False`, `temperature`) can't leak into downstream stages, then reads back `log_probs` / `response_mask` / `responses` / `rollout_log_probs` and hands them to `calculate_debug_metrics`. It deliberately does **not** write `old_log_probs` back to the TransferQueue — bypass semantics are preserved.
- **`verl/trainer/config/algorithm.py`** — `RolloutCorrectionConfig.debug_rollout_actor_probs: bool = False`, documented in the class docstring.
- **`verl/trainer/config/algorithm/rollout_correction.yaml`** + the four regenerated `_generated_*_trainer.yaml`.
- **`docs/algo/rollout_corr.md`** — documents the flag, its one-forward-pass cost, the `calculate_log_probs` requirement, and how the recompute's temperature scaling lines up with `rollout.logprobs_mode`.
- **`tests/utils/debug/test_metrics.py`** — the three cases described above.

**Points I'd like reviewer input on:**

1. **Scope: v1 trainer only.** The same gap exists in `verl/trainer/ppo/ray_trainer.py:1550` and `verl/experimental/separation/ray_trainer.py:510`. I kept this to `v1/trainer_base.py` because that's the path `PPOTrainerSeparateAsync` uses and the one I could validate on hardware. Happy to extend to the others in this PR or a follow-up.
2. **Why an extra forward** rather than reusing the π_θ that `_update_actor`'s first micro-batch already computes? That would be free, but it only covers part of the batch (so the metric would not be comparable to the decoupled path's) and would mean reaching into the minibatch training loop. The extra forward keeps the metric's semantics *exactly* identical to what non-bypass mode reports.
3. **No warning when `debug_rollout_actor_probs=true` but `bypass_mode=false`** — the flag is currently just ignored. I can add a config-time warning if that's preferred.
4. **The `calculate_log_probs` guard is arguably unreachable.** The bypass path reads `rollout_log_probs` unconditionally two lines later, so if `calculate_log_probs` were `false`, bypass mode is already broken. The warn-and-skip is defensive; an `assert` (or dropping it) may be cleaner. Open to either.

- [x] Add unit or end-to-end test(s) to [the CI workflow](https://github.com/verl-project/verl/tree/main/.github/workflows) to cover all the code. If not feasible, explain why: the new trainer method requires a live worker group + TransferQueue and has no existing test harness; validated by GPU smoke test instead. See the Test section.
- [x] If your PR is related to the `recipe` submodule — N/A.

---

**AI assistance disclosure:** This change was developed with AI assistance (GitHub Copilot and Claude). Per [`CLAUDE.md`](https://github.com/verl-project/verl/blob/main/CLAUDE.md), I have reviewed every changed line, understand the change end-to-end, and ran the tests and the GPU smoke test reported above myself.
Screenshot 2026-07-29 at 4 39 34 PM

### DIFF
--- a/docs/algo/rollout_corr.md
+++ b/docs/algo/rollout_corr.md
@@ -2,7 +2,7 @@
 
 **Author:** [Yingru Li](https://richardli.xyz/)
 
-Last updated: 10/30/2025.
+Last updated: 07/29/2026.
 
 ---
 
@@ -281,6 +281,32 @@ Threshold specification for rejection sampling.
 - **K1 KL modes (`*k1`)**: Use `"lower_upper"` strings (e.g. `"0.7_1.3"`). Supplying a float implies only the upper bound; the lower bound defaults to its reciprocal.
 - **K2/K3 KL modes (`*k2`/`*k3`)**: Supply positive upper bounds (float or numeric string).
 - Set to `null` to disable thresholds entirely (only valid when `rollout_rs` is null).
+
+### `debug_rollout_actor_probs` (bool)
+
+Opt-in diagnostic for **bypass mode** (`bypass_mode: true`), default `false`.
+
+In bypass mode the trainer sets `old_log_probs = rollout_log_probs` and never recomputes the actor
+policy π_θ, so the rollout-vs-actor consistency metrics — the Pearson correlation
+`training/rollout_actor_probs_pearson_corr` and `training/rollout_probs_diff_{max,mean,std}` — are
+**not emitted**. These metrics are among the most useful async-RL diagnostics because they directly
+measure the training/inference (off-policy) gap and catch precision/tokenizer/temperature drift that
+KL magnitude alone can miss.
+
+When set to `true`, the trainer runs **one extra no-grad inference forward** per step to recompute
+π_θ purely for observability and feeds it to `calculate_debug_metrics`. The training computation is
+untouched, so results are byte-identical whether or not this is enabled — the only cost is the extra
+inference forward (roughly what the non-bypass path already pays every step).
+
+Requirements and notes:
+
+- Requires `actor_rollout_ref.rollout.calculate_log_probs: true` so that `rollout_log_probs` is
+  available; otherwise the diagnostic logs a one-time warning and is skipped.
+- The recompute divides logits by `actor_rollout_ref.rollout.temperature`, which matches vLLM's
+  default `logprobs_mode: processed_logprobs` (log-probs taken after the logits processors,
+  temperature included). The correlation is therefore meaningful at any temperature. If you override
+  `logprobs_mode` to a `raw_*` variant, the two sides are scaled differently and the reported
+  correlation partly reflects that mismatch.
 
 ## Understanding the Framework: Components and Combinations
 
--- a/tests/utils/debug/test_metrics.py
+++ b/tests/utils/debug/test_metrics.py
@@ -11,12 +11,35 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import math
 import unittest
 
 import torch
 
 from verl.protocol import DataProto
 from verl.utils.debug.metrics import calculate_debug_metrics
+
+DEBUG_METRIC_KEYS = {
+    "training/rollout_probs_diff_valid",
+    "training/rollout_probs_diff_max",
+    "training/rollout_probs_diff_mean",
+    "training/rollout_probs_diff_std",
+    "training/rollout_actor_probs_pearson_corr",
+}
+
+
+def _make_data(rollout_log_probs, old_log_probs, mask):
+    rollout_log_probs = torch.tensor(rollout_log_probs, dtype=torch.float32)
+    old_log_probs = torch.tensor(old_log_probs, dtype=torch.float32)
+    mask = torch.tensor(mask)
+    return DataProto.from_dict(
+        {
+            "rollout_log_probs": rollout_log_probs,
+            "old_log_probs": old_log_probs,
+            "response_mask": mask,
+            "responses": torch.zeros_like(rollout_log_probs),
+        }
+    )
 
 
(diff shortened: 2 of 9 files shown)


wuxibin89:
`separate_async` has supported Decoupled PPO to recompute old_logprobs: https://github.com/verl-project/verl/pull/7188

PR #7199, opened 2026-07-30
[megatron] fix: preserve generation config during merge

### LINKED ISSUE #7198: [Bug] Megatron model merger does not preserve generation_config.json
### Description

The distributed Megatron model merger saves `config.json`, processor, and tokenizer artifacts, but does not preserve the independent `generation_config.json` stored in the checkpoint Hugging Face metadata directory.

For models where generation settings differ from `config.json` (for example Qwen models with multiple EOS token IDs), the merged output can silently lose those settings. This is the same failure class previously fixed for the FSDP merger in #1277 after #1246.

### Affected path

`verl/model_merger/megatron_model_merger.py::save_hf_model_and_tokenizer`

Current behavior saves `self.model_config`, processor, and tokenizer, but never loads/saves `GenerationConfig` from `self.hf_model_config_path`.

### Expected behavior

When the source checkpoint contains `generation_config.json`, the merged Hugging Face output should preserve it. If it is absent, merger behavior should remain unchanged.

### Duplicate check

I searched open issues and PRs for `megatron generation_config merger` and `generation config megatron`; no existing Megatron-specific fix was found. Related historical references are #1246 and #1277.

### Environment

Observed against current `main` at `668baad7455453eac8ab863f1b2d6fecaec746ed`.

### AI assistance disclosure

AI assistance was used to trace the affected code paths and prepare this report. The submitter will review the proposed patch and test evidence before marking the corresponding draft PR ready for review.

### LINKED ISSUE #7226: [Bug] Megatron model merger uploads the same checkpoint from every rank
### System Info

Current `verl-project/verl` Megatron model-merger code path. This is a distributed control-flow issue and is not specific to a particular GPU, Python, or model configuration.

### Information

- The problem arises when using the built-in Megatron model merger with `hf_upload` enabled.
- No modified training script is required.

### Reproduction

1. Run the Megatron model merger with more than one distributed rank.
2. Set `hf_upload` to a Hugging Face repository.
3. `merge_and_save()` executes on every rank.
4. The current upload branch calls `upload_to_huggingface()` without a rank guard, so every rank starts uploading the same merged checkpoint.

The relevant behavior is equivalent to:

```python
self.save_hf_model_and_tokenizer(merged_state_dict)
if self.config.hf_upload:
    self.upload_to_huggingface()
```

This was identified while validating the Megatron merge path for #7198. A fix is included in #7199.

### Expected behavior

All ranks may participate in distributed shard writing and synchronization, but only rank 0 should start the optional Hugging Face upload after the merged artifacts have been written and validated.

```python
if self.config.hf_upload and self.rank == 0:
    self.upload_to_huggingface()
```

KANIKIG:
### What does this PR do?

Fixes #7198.

Also fixes #7226: under distributed Megatron merge, every rank previously entered the optional Hugging Face upload branch and could upload the same merged checkpoint concurrently. The upload is now restricted to rank 0.

The distributed Megatron merger now preserves the independent pretrained `generation_config.json` in merged Hugging Face outputs. This prevents generation settings that are absent from `config.json`, such as multiple EOS token IDs, from being lost.

The change reuses a common loader with the existing FSDP merger behavior and removes a stale target `generation_config.json` when the source has none. Missing generation configuration remains a supported fallback, while an existing but malformed `generation_config.json` now propagates its parsing error instead of being silently treated as absent. This fail-closed behavior is intentional because silently dropping a corrupt generation configuration can change generation behavior without a visible failure.

This is not a duplicate of #1277: that PR fixed the FSDP/legacy merger path. The distributed Megatron merger was added later and did not call the generation-config preservation logic.

### Checklist Before Starting

- [x] Searched for similar PRs: https://github.com/verl-project/verl/pulls?q=is%3Apr+is%3Aopen+generation_config+megatron
- [x] Searched for related issues: https://github.com/verl-project/verl/issues?q=generation_config+megatron
- [x] PR title follows the required module/type format.

### Test

```text
PYTHONPATH=. .venv/bin/pytest -q tests/model_merger/test_output_validation_on_cpu.py
14 passed

.venv/bin/ruff check verl/model_merger/base_model_merger.py verl/model_merger/megatron_model_merger.py tests/model_merger/test_output_validation_on_cpu.py
All checks passed!

.venv/bin/ruff format --check verl/model_merger/base_model_merger.py verl/model_merger/megatron_model_merger.py tests/model_merger/test_output_validation_on_cpu.py
3 files already formatted

python -m py_compile <three changed Python files>
passed
```

The unit tests cover preserving a multi-EOS configuration, removing stale target configuration when the source has no independent generation config, and rejecting a malformed source generation config. A full multi-GPU Megatron merge was not run locally.

### API and Usage Example

No API or CLI change. Existing Megatron merge commands produce the additional `generation_config.json` artifact when it exists in the source checkpoint metadata.

### Design & Code Changes

- Extract loading and saving of pretrained generation configuration into `BaseModelMerger`.
- Save that configuration from the distributed Megatron merger.
- Treat a missing generation config as an allowed fallback, but surface malformed existing configuration.
- Restrict optional Hugging Face upload to rank 0, fixing #7226.
- Add CPU unit coverage for multi-EOS preservation, missing-source cleanup, and malformed-source rejection.

- [x] Read `CONTRIBUTING.md` and `AGENTS.md`.
- [x] Ran targeted Ruff and unit tests listed above.
- [ ] Human submitter has completed final line-by-line review.
- [x] No documentation update is required because behavior is corrective and the CLI is unchanged.
- [x] Added unit tests.
- [ ] Request CI only after human review.

### AI assistance disclosure

AI assistance was used for code-path investigation, patch preparation, tests, and initial review. This PR is intentionally opened as a draft and must remain draft until the human submitter reviews every changed line and can defend the change end-to-end.

### DIFF
--- a/tests/model_merger/test_output_validation_on_cpu.py
+++ b/tests/model_merger/test_output_validation_on_cpu.py
@@ -117,7 +117,7 @@ def test_save_generation_config_preserves_pretrained_values(tmp_path):
     assert saved_config.max_new_tokens == 4096
 
 
-def test_save_generation_config_allows_missing_file(tmp_path):
+def test_save_generation_config_allows_missing_file(capsys, tmp_path):
     source_dir = tmp_path / "source"
     target_dir = tmp_path / "target"
     source_dir.mkdir()
@@ -130,6 +130,33 @@ def test_save_generation_config_allows_missing_file(tmp_path):
     merger.save_generation_config(target_dir)
 
     assert not stale_config_path.exists()
+    assert capsys.readouterr().out == ""
+
+
+def test_load_generation_config_warns_when_file_is_missing(capsys, tmp_path):
+    merger = _TestModelMerger.__new__(_TestModelMerger)
+    merger.hf_model_config_path = tmp_path
+
+    assert merger.load_generation_config() is None
+    assert f"Warning: Generation config file not found in {tmp_path}" in capsys.readouterr().out
+
+
+def test_save_generation_config_rejects_invalid_source(tmp_path):
+    source_dir = tmp_path / "source"
+    target_dir = tmp_path / "target"
+    source_dir.mkdir()
+    target_dir.mkdir()
+    (source_dir / "generation_config.json").write_text("{invalid", encoding="utf-8")
+    stale_config_path = target_dir / "generation_config.json"
+    stale_config_path.write_text('{"eos_token_id": 0}', encoding="utf-8")
+
+    merger = _TestModelMerger.__new__(_TestModelMerger)
+    merger.hf_model_config_path = source_dir
+
+    with pytest.raises(OSError, match="not a valid JSON file"):
+        merger.save_generation_config(target_dir)
+
+    assert stale_config_path.is_file()
 
 
 @pytest.mark.parametrize("weight_name", ("model.safetensors", "pytorch_model.bin"))
--- a/tests/model_merger/test_output_validation_on_cpu.py
+++ b/tests/model_merger/test_output_validation_on_cpu.py
@@ -17,7 +17,7 @@ from types import SimpleNamespace
 
 import pytest
 import torch
-from transformers import AutoModelForCausalLM, Qwen2Config
+from transformers import AutoModelForCausalLM, GenerationConfig, Qwen2Config
 
 from verl.model_merger import base_model_merger
 from verl.model_merger.base_model_merger import BaseModelMerger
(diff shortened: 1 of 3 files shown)


ErenAta16 (commented):
The generation-config fix is right and the test is the good kind: writing a real `GenerationConfig` with `eos_token_id=[151643, 151645]` and reading it back after the merge exercises exactly the case the issue describes, a list-valued EOS that `config.json` cannot carry.

**There is a second, unrelated fix in this diff and it is not in the description.**

```diff
-            if self.config.hf_upload:
+            if self.config.hf_upload and self.rank == 0:
```

`merge_and_save` runs on every rank under Megatron, so without the guard every rank starts its own upload of the same merged checkpoint. That is a genuine bug, arguably more disruptive than the one in the title, and it deserves its own line in the description and ideally its own issue reference. Someone bisecting a duplicate-upload problem later will not find it under "preserve generation config during merge".

If it was folded in because you hit it while testing this, that is fair, but it should be visible.

**The error handling narrows, which I think is an improvement worth stating.** Before:

```python
try:
    model.generation_config = GenerationConfig.from_pretrained(self.hf_model_config_path)
except OSError:
    ...
```

After, the existence check happens first and `from_pretrained` is called only when the file is there. So a missing `generation_config.json` is still handled, but a *malformed* one now propagates instead of being swallowed by the `except OSError`. That is the better behaviour, since silently dropping a corrupt generation config is how you end up with a merged model that generates differently from the source for no visible reason. Worth a sentence in the description so it is a decision on record.

`warn_if_missing` defaulting to `True` is the right way round: the merge path wants to know, and a caller that legitimately expects no file can opt out.

One question. `save_generation_config` is called from `save_hf_model_and_tokenizer`, which I assume is already rank-guarded given the upload fix just below it, but it is worth confirming the write itself does not race. If every rank calls it, several processes write `generation_config.json` into the same target directory concurrently, and while they would write identical bytes it is still a partial-file window for anything reading in parallel.

KANIKIG:
Thanks for the careful review. I updated the PR description to make both additional decisions explicit:

- The duplicate multi-rank Hugging Face upload is now documented as a separate bug and linked to #7226.
- Missing `generation_config.json` remains an allowed fallback, while an existing malformed file intentionally propagates its parsing error instead of being silently discarded.

I also confirmed the generation-config write is rank-guarded. In `save_hf_model_and_tokenizer`, `self.save_generation_config(self.config.target_dir)` is inside the existing `if self.rank == 0:` block, together with the index, model config, processor, tokenizer, and final artifact validation. Other ranks only write their assigned safetensor shards and participate in the collectives, so they do not concurrently write `generation_config.json`.

No code change was needed for that part; the current placement already prevents the partial-file race you described.

ErenAta16 (commented):
That covers everything I raised, and splitting the upload fix out to #7226 is the better outcome than folding it in: a duplicate multi-rank upload is worth its own bisectable commit rather than being buried under a generation-config title.

The rank-guard answer is corroborated by the diff itself. The new call lands at the same indentation as, and immediately beside, the other rank-0 artifact writes:

```python
            self.model_config.save_pretrained(self.config.target_dir)
            self.save_generation_config(self.config.target_dir)

            processor = hf_processor(...)
            tokenizer = hf_tokenizer(...)
```

so it is grouped with exactly the writes that must not race, which is what I wanted to be sure of. I could not read the enclosing `if` from the diff hunk alone, so I am taking your confirmation for that line specifically.

Documenting the missing-versus-malformed distinction explicitly is the part I would have most wanted written down. "Absent file is a supported fallback, present-but-unparseable propagates" is a real policy choice and not one a reader would infer from the code, since both look like the same failure from the outside.

Nothing further from me.

KANIKIG:
Hi @wuxibin89, could you please review this PR when you have time? Thank you!

wuxibin89 (approved):

PR #7207, opened 2026-07-30
[rollout] fix: cap partial rollout resumes at the remaining response budget

benyucong:
### What does this PR do?

**Summary:** when a partial rollout is aborted and resumed, the total number of generated tokens can go past `response_length`. The extra tokens are slow to generate, and they are thrown away afterwards. This PR makes the resume path count the tokens it has already generated, so the total stays inside `response_length`.

#### The problem, step by step

**1. Agent loops do not set a token limit.**

`sampling_params` is built in `agent_loop.py:590` and `agent_loop_tq.py:64`. Neither one sets `max_tokens` or `max_new_tokens`.

**2. So the server uses its default formula** (`vllm_async_server.py:564`):

```python
max_tokens = min(
    self.config.response_length,
    self.config.prompt_length + self.config.response_length - len(prompt_ids),
)
```

This formula assumes that `prompt_ids` contains only the prompt. That is true for the first request.

**3. But a resume sends prompt + generated tokens.**

`FullyAsyncLLMServerClient.generate` calls the server with `prompt_ids + final_output.token_ids` (`llm_server.py:369`). This is correct — the engine needs them to continue the sequence. But now `len(prompt_ids)` is larger than the prompt.

**4. Two mistakes happen at the same time.**

- Generated tokens are subtracted from a budget that means *prompt* tokens.
- `self.config.prompt_length` is the configured *maximum* prompt size, not the real prompt size of this request.

These two mistakes do not cancel each other. The result is a limit based on the model window (`prompt_length + response_length`) instead of the remaining budget (`response_length - already_generated`).

**5. The `max_model_len` clamp does not stop it.**

`max_model_len` is `null` by default, and then becomes the model's `max_position_embeddings` (`vllm_async_server.py:954-955`). It is only checked as an upper bound. So it is usually much larger than `prompt_length + response_length`.

#### How large is the error?

The total generated length can reach:

```
prompt_length + response_length - actual_prompt_length
```

So the extra amount is exactly the **unused part of the prompt budget**:

```
overshoot = prompt_length - actual_prompt_length
```

This means:

- Short prompts → large overshoot.
- A prompt that fills `prompt_length` → no overshoot at all.

Example with `prompt_length=2048`, `response_length=8192`, real prompt = 136 tokens, `max_model_len=10241`:

| | `len(prompt_ids)` | limit the server gives | correct limit |
| --- | --- | --- | --- |
| first request | 136 | `min(8192, 2048+8192−136)` = **8192** | 8192 ✅ |
| resume after 6000 tokens | 6136 | `min(8192, 2048+8192−6136)` = **4104** | 8192−6000 = **2192** ❌ |

The resume is allowed 4104 more tokens instead of 2192. The response reaches 10104 tokens instead of 8192.

#### Why this wastes time

Decoding cost grows with sequence length, because every step reads the KV cache of all earlier tokens. The extra tokens sit at positions 8192–10104, so they are the slowest tokens in the whole request. They are then dropped by the `[: response_length]` slice in the agent loop. Maximum cost, zero benefit.

#### The fix already exists, but never runs

`llm_server.py` already tracks the remaining budget across resumes. But that code is inside `if original_max_tokens is not None`. When the caller sets no limit key — which is always true for agent loops — `original_max_tokens` is `None`, so the block never runs. This PR gives it a value, so it starts working.

#### Which setups are affected

This does **not** require `partial_rollout`. `trainer_colocate_async.py:34` and `trainer_separate_async.py:131` also install this client. Their config has no `async_training` key, so `should_retry` at `llm_server.py:412` stays `True`. There is no way to turn it off.

### Checklist Before Starting

- [x] Search for similar PRs. Paste at least one query link here: [`is:pr max_tokens response_length`](https://github.com/verl-project/verl/issues?q=repo%3Averl-project%2Fverl+is%3Apr+max_tokens+response_length), [`partial_rollout response_length`](https://github.com/verl-project/verl/issues?q=repo%3Averl-project%2Fverl+partial_rollout+response_length)

I searched the open PRs that touch `max_tokens`, `response_length`, or partial rollout. No open PR changes `llm_server.py` or the resume path, so this is not a duplicate.

### Test

New file: `tests/workers/rollout/test_llm_server_response_length_cap_on_cpu.py`.

It runs the real `FullyAsyncLLMServerClient`. Only the server is replaced by a fake one, and that fake copies the exact `max_tokens` logic of `vLLMHttpServer.generate`. The file name ends with `_on_cpu.py`, so `cpu_unit_tests.yml` collects it automatically.

```bash
pytest -s -x --asyncio-mode=auto tests/workers/rollout/test_llm_server_response_length_cap_on_cpu.py
```

**Result on this branch: 8 passed.**

**Result on the parent commit: 3 of 8 fail.**

```
AssertionError: generated 10104 tokens against a 8192 budget; per-attempt grants were
  [(136, 8192), (2136, 8104), (4136, 6104), (6136, 4104), (8136, 2104), (10136, 104)]

AssertionError: generated 10148 tokens against a 8192 budget; per-attempt grants were
  [(92, 8192), (2092, 8148), (4092, 6148), (6092, 4148), (8092, 2148), (10092, 148)]
```

Each pair above is `(len(prompt_ids), limit given)`. With the fix, the limits become:

```
[(136, 8192), (2136, 6192), (4136, 4192), (6136, 2192), (8136, 192)]
```

Every attempt now receives exactly what is still left. The sum is 8192.

I also re-ran the existing test for this code, `tests/workers/rollout/test_llm_server_routed_experts_on_cpu.py` (it covers routing merge across resumes, for all three backend formats). All tests still pass.

`pre-commit run --files <changed files> --show-diff-on-failure` passes on all hooks.

**Not tested:** I did not run a GPU end-to-end job. The bug is integer arithmetic on `len(prompt_ids)`, and the test above runs the real client against an exact copy of the server formula.

### API and Usage Example

No API change.

Behaviour change: if the caller sets no `max_tokens` or `max_new_tokens`, the **total** across all resumes is now limited by `response_length`. Before, each attempt got a new limit from the server formula. A limit set by the caller still has priority.

```python
# Unchanged: an explicit limit wins.
await client.generate(request_id=..., prompt_ids=..., sampling_params={"max_tokens": 512})

# Changed: with no limit key, the total across all resumes is now bounded by
# actor_rollout_ref.rollout.response_length, instead of by the model window.
await client.generate(request_id=..., prompt_ids=..., sampling_params={"temperature": 1.0})
```

### Design & Code Changes

**`verl/workers/rollout/llm_server.py`**

- If the caller sets no limit key, use the configured `response_length` as the total budget. This activates the existing per-attempt tracking at the end of the loop.
- Copy `sampling_params` first. That tracking code writes the remaining budget back into the dict, and the caller re-uses the same dict across turns, so the write must not escape.
- Read the config through a small helper that returns `None` if the rollout section is missing. Callers that pass a config stub (the existing CPU tests do this) keep the old behaviour.

**`tests/workers/rollout/test_llm_server_response_length_cap_on_cpu.py`** — new file.

**Why I did not change the vLLM and SGLang formula**

The formula is wrong for resumed sequences. But the term `prompt_length + response_length - len(prompt_ids)` is currently the *only* thing that limits the total across resumes. If it were replaced by `min(response_length, max_model_len - len(prompt_ids))` without the client-side budget, every resume would start again with a full `response_length`, which is worse. After this PR, the formula is only a fallback for callers that send no limit at all.

**Not included in this PR**

The same problem also exists in `tool_agent_loop.py`. It adds earlier turns and tool responses into `prompt_ids`, and then hits the same server formula. No resume is needed for this. Its `len(response_mask) >= response_length` checks only stop the loop *after* a too-long turn was already generated. The fix would be a per-turn limit of `response_length - len(response_mask)`. I kept this PR to the resume path only, so it stays small and easy to review. I am happy to add the multi-turn part here if maintainers prefer.

- [ ] Add / Update [the documentation](https://github.com/verl-project/verl/tree/main/docs). Not needed — no config or API change.
- [x] Add unit or end-to-end test(s) to [the CI workflow](https://github.com/verl-project/verl/tree/main/.github/workflows). Covered by `cpu_unit_tests.yml` through the `*_on_cpu.py` pattern.
- [ ] `recipe` submodule update — not needed.

---

AI assistance was used in preparing this change: the analysis, the patch, and the test were produced with Claude Code. I have read through the full diff and can speak to the change end-to-end.

### DIFF
--- a/tests/workers/rollout/test_llm_server_response_length_cap_on_cpu.py
+++ b/tests/workers/rollout/test_llm_server_response_length_cap_on_cpu.py
@@ -1,0 +1,172 @@
+# Copyright 2026 Bytedance Ltd. and/or its affiliates
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Cumulative response budget across partial rollout resumes.
+
+A resume re-sends ``prompt + everything generated so far`` as ``prompt_ids``. If the request
+carries no explicit limit, the server falls back to
+
+    min(response_length, prompt_length + response_length - len(prompt_ids))
+
+which charges already-generated tokens against the *prompt* budget and measures that budget with
+the configured ``prompt_length`` rather than this request's actual prompt length. Generation then
+runs to ``prompt_length + response_length - actual_prompt_length`` -- the overshoot is exactly the
+unused prompt budget, is decoded at the deepest KV positions, and is discarded by the agent loop's
+``[: response_length]`` slice. The ``max_model_len`` clamp does not bound it, because
+``max_model_len`` defaults to the model's ``max_position_embeddings``.
+
+``FakeRolloutServer`` below mirrors ``vLLMHttpServer.generate``'s cap arithmetic so these tests
+pin the behaviour the real servers implement.
+"""
+
+from __future__ import annotations
+
+import asyncio
+from types import SimpleNamespace
+from typing import Any
+
+import pytest
+from omegaconf import OmegaConf
+
+from verl.workers.rollout import llm_server
+from verl.workers.rollout.llm_server import FullyAsyncLLMServerClient
+from verl.workers.rollout.replica import TokenOutput
+
+PROMPT_LENGTH = 2048
+RESPONSE_LENGTH = 8192
+MAX_MODEL_LEN = 10241
+
+
+class FakeRolloutServer:
+    """Stand-in for the rollout server, reproducing its ``max_tokens`` resolution.
+
+    The model never emits EOS, so a request only ends when it exhausts the budget it was granted;
+    every ``abort_every`` tokens a weight sync aborts it, which is what drives the resume loop.
+    """
+
+    def __init__(self, abort_every: int):
+        self.abort_every = abort_every
+        self.grants: list[tuple[int, int]] = []
+
+    async def generate(self, request_id, *, prompt_ids, sampling_params, **kwargs):
+        # Requests cross a Ray boundary, so the server only ever mutates its own copy.
+        params = dict(sampling_params)
+
+        max_possible_tokens = MAX_MODEL_LEN - len(prompt_ids)
+        assert max_possible_tokens >= 1
+
+        if "max_tokens" in params:
+            max_tokens = params.pop("max_tokens")
+        elif "max_new_tokens" in params:
+            max_tokens = params.pop("max_new_tokens")
+        else:
+            max_tokens = min(RESPONSE_LENGTH, PROMPT_LENGTH + RESPONSE_LENGTH - len(prompt_ids))
+        max_tokens = max(1, min(max_tokens, max_possible_tokens))
+        self.grants.append((len(prompt_ids), max_tokens))
+
+        generated = min(max_tokens, self.abort_every)
+        return TokenOutput(
+            token_ids=[7] * generated,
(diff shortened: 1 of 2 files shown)


HollowMan6 (approved):
LGTM

PR #7217, opened 2026-07-31
fix(agent_loop): propagate validation metadata to async rewards

### LINKED ISSUE #7121: [BUG][agent_loop] async reward drops meta_info; custom reward can't see validate flag
### Summary

On the experimental **agent_loop async-reward path**, `AgentLoopWorker._compute_score` builds the reward `DataProto` **without forwarding `meta_info`**. As a result the reward worker (`reward_loop.py` → `reward_manager.run_single`) receives a `DataProto` whose `meta_info` is empty, so a custom reward manager can never read `data.meta_info.get("validate")` (or any other per-batch `meta_info` key) at scoring time.

The **classic reward path** does not have this problem: the reward manager's `__call__(self, data, ...)` gets the full batch `DataProto`, and the trainer stamps `test_batch.meta_info["validate"] = True` during validation — so classic custom rewards can branch on train vs. val, but agent_loop ones cannot.

### System Info / Version

- verl `0.8.0.dev`, worktree HEAD `bcb63864` (committed 2026-04-17).
- Also reproduces on **current public `main`** (`0.9.0.dev`), checked **2026-07-22**.
- Agent Loop is `.. versionadded:: 0.4.2` `[status: alpha]` (`docs/advance/agent_loop.rst`).

### Reproduction

1. `actor_rollout_ref.rollout.mode=async` (agent_loop) + a custom `reward_manager` on the async reward path (`reward.reward_manager=...`).
2. In `run_single`, read `data.meta_info.get("validate", False)`.
3. Run training with validation enabled.
4. The custom reward **always** sees `validate` absent/`False`, in both train and validation rollouts.
5. Swap to the classic (sync) reward path with the same manager reading `data.meta_info["validate"]` → it correctly reads `True` during `_validate`.

### Expected vs. Actual

- **Expected:** agent_loop reward functions can distinguish validation from training via `meta_info["validate"]`, matching the classic path.
- **Actual:** `meta_info` is dropped before the reward manager, so `validate` (and everything else in `meta_info`) is unavailable.

### Root cause (HEAD `bcb63864`)

- `verl/experimental/agent_loop/agent_loop.py:881-884` — the reward `DataProto` is constructed with **no** `meta_info=`:

  ```python
  data = DataProto(
      batch=batch,
      non_tensor_batch=non_tensor_batch,
  )   # <- no meta_info=
  ```

- `_compute_score` (`agent_loop.py:858`) doesn't take `validate`, but its caller `_agent_loop_postprocess(self, output, validate, ...)` **has `validate` in scope** — it already forwards it to the sibling call `_compute_teacher_logprobs(..., validate=validate)` (`agent_loop.py:749-753`). So the value is available at the call site; it just isn't passed down.
- Reward side receives that `DataProto`: `verl/experimental/reward_loop/reward_loop.py:148-152` (`compute_score` → `reward_manager.run_single(data)`). `meta_info` is empty there. (For evidence that `meta_info` *is* consulted on this path, see `reward_manager/limited.py:495` reading `data.meta_info.get("reward_extra_keys")`.)
- Contrast — classic path: `verl/workers/reward_manager/naive.py:46` `__call__(self, data, ...)` gets the full batch, and `verl/trainer/ppo/ray_trainer.py:579` sets `test_batch.meta_info["validate"] = True` (public `main`: `ray_trainer.py:663`).

### Proposed fix

Thread `validate` into `_compute_score` and attach it (or the whole `meta_info`) to the reward `DataProto`:

```python
data = DataProto(
    batch=batch,
    non_tensor_batch=non_tensor_batch,
    meta_info={"validate": validate},
)
```

Since `validate` is already in scope at the caller (and already forwarded to `_compute_teacher_logprobs`), this is a minimal change.

**Questions for maintainers:** Is dropping `meta_info` on the async reward path intended? Is there a recommended way for agent_loop custom rewards to detect validation today? Happy to send a PR if this direction looks right.

YAO-001:
## Summary

- propagate the existing `validate` flag through both agent-loop postprocess paths
- attach `validate` to the async reward `DataProto.meta_info`
- add a CPU-only regression test covering training and validation values

## Root cause

`AgentLoopWorker._compute_score` constructed a new reward `DataProto` without
`meta_info`. The async reward worker therefore could not observe
`data.meta_info["validate"]`, unlike the classic reward path.

The validation flag was already available in both postprocess callers, but it
was not forwarded into `_compute_score`.

## Fix

Thread `validate` into `_compute_score` from:

- `AgentLoopWorker._agent_loop_postprocess`
- `AgentLoopWorkerTQ._agent_loop_postprocess`

Then construct the reward input with:

```python
meta_info={"validate": validate}
```

This keeps reward computation and model-serving behavior unchanged while
restoring train/validation visibility for custom async reward managers.

## Validation

- `tests/experimental/agent_loop/*_on_cpu.py`: **25 passed**
- `tests/trainer/ppo/v1/test_agent_loop_tq_on_cpu.py`: **1 passed**
- Ruff: passed
- Ruff format: passed
- mypy: passed
- compileall: passed
- `git diff --check`: passed

The human submitter reviewed every changed line, personally ran the relevant
tests, and confirmed understanding of the change.

## Duplicate-work check

This does not duplicate an existing PR:

- `gh pr list --repo verl-project/verl --state open --search "7121 in:body"` returned no PR
- the area-keyword search only returned unrelated PR #4160
- another contributor described the same intended approach on #7121 but had
  not opened a PR; coordination was requested in the issue before submission

## AI assistance

OpenAI Codex assisted with investigation, implementation, testing, and PR
preparation. The human submitter reviewed the complete diff, reran the relevant
tests, understands the change end-to-end, and can defend it.

Closes #7121


ErenAta16 (commented):
Both call sites are covered, which is the thing I went looking for since `validate` is added as a required positional rather than defaulted. A missed caller would be a `TypeError` at runtime rather than a type error at review time:

```
verl/experimental/agent_loop/agent_loop.py:821   await self._compute_score([output], kwargs=kwargs, validate=validate)
verl/trainer/ppo/v1/agent_loop_tq.py:157         await self._compute_score(outputs, kwargs=kwargs, validate=validate)
```

The other `_compute_score` matches in the tree are the reward managers' own methods (`workers/reward_manager/naive.py`, `prime.py`, `experimental/reward_loop/...`), unrelated to `AgentLoopWorker`, so there is nothing else to thread.

Making it required rather than `validate: bool = False` is the right call here. A default would let a future third postprocess path silently report training metrics for a validation pass, which is precisely the bug being fixed, and there are only two callers so the cost of being explicit is nil.

Attaching it as `meta_info={"validate": validate}` matches how the synchronous reward path carries it, so downstream consumers that already branch on `meta_info["validate"]` need no change.

A CPU-only regression test is the right shape for this. The failure is in metadata plumbing, not in the reward computation, so exercising it without a model keeps the test in the fast suite where it will actually run on every PR. Asserting both the training and validation values rather than just the one that was broken is what stops a future change from hardcoding the flag.

Nothing blocking from me.

YAO-001:
Hi @wuxibin89, sorry for the ping. When you have a chance, could you please take a look at this PR? It’s a small, scoped fix for #7121, propagating the validate metadata to async rewards in agent_loop, with CPU regression coverage. Thanks for your time!

PR #7220, opened 2026-07-31
[trainer] fix: build RL tokenizer/processor from HFModelConfig

ZhiliangWu:
### What does this PR do?

Make the PPO entrypoints build the dataset's tokenizer/processor from `HFModelConfig`, the same way the SFT trainer ([`sft_trainer.py:141`](https://github.com/verl-project/verl/blob/76324f1de4aabad06441a01fc17b605d9b3e564a/verl/trainer/sft_trainer.py#L141)) and the engine workers ([`engine_workers.py:534`](https://github.com/verl-project/verl/blob/76324f1de4aabad06441a01fc17b605d9b3e564a/verl/workers/engine_workers.py#L534)) already do. Both PPO entrypoints (`verl/trainer/main_ppo_v0.py` and the v1 `verl/trainer/ppo/v1/trainer_base.py`) currently hand-roll `copy_to_local` + `hf_tokenizer` + `hf_processor`, duplicating what `HFModelConfig.__post_init__` already encapsulates. This aligns the RL entrypoints with SFT (and the workers) and removes the duplicated construction.

A side benefit: because `HFModelConfig` is built through `omega_conf_to_dataclass` (which honors `model._target_`), a custom model-config subclass can now influence the tokenizer/processor the RL dataset receives — consistent with how SFT already behaves. This may help multimodal models whose processor class is not in the `hf_processor()` allow-list. (Background on that custom-processor use case: discussion #7084.)

### Checklist Before Starting

- [x] Search for similar PRs. Paste at least one query link here: https://github.com/verl-project/verl/pulls?q=is%3Apr+hf_processor+HFModelConfig

### Test

No behavior change for standard runs: the tokenizer/processor are still built by the same `hf_tokenizer()` / `hf_processor()` calls, now reached through `HFModelConfig` (the same path SFT and the workers use). The v1 `main_ppo` → `_init_tokenizer` path is exercised by the existing `e2e_ppo_trainer*` CI jobs (default `trainer.use_v1=true`).

### API and Usage Example

No new or changed config keys; standard usage is unaffected. As a side benefit, a custom model config can now be selected on the RL path via `model._target_`, exactly as the SFT path already allows:

```bash
    actor_rollout_ref.model._target_=your_pkg.YourHFModelConfig \
    actor_rollout_ref.model.path=/path/to/model \
    ...
```

### Design & Code Changes

Behavior differences from the previous inline construction:
- **`use_shm`** — preserved (`HFModelConfig.__post_init__` passes it into `copy_to_local`).
- **`use_fast`** — the processor's `use_fast` now follows the library default rather than being forced to `True`, matching SFT. Fast tokenizers are still auto-selected when available.
- **`trust_remote_code`** — for the tokenizer/processor this now comes from the model config (`actor_rollout_ref.model.trust_remote_code`) rather than `data.trust_remote_code`. Both default to `False`, so default runs are unaffected; this matches how the workers build the model.

> [!IMPORTANT]

### DIFF
--- a/verl/trainer/main_ppo_v0.py
+++ b/verl/trainer/main_ppo_v0.py
@@ -161,8 +161,6 @@ class TaskRunner(BaseTaskRunner):
         # Print the initial configuration. `resolve=True` will evaluate symbolic values.
         from pprint import pprint
 
-        from verl.utils.fs import copy_to_local
-
         print(f"TaskRunner hostname: {socket.gethostname()}, PID: {os.getpid()}")
         pprint(OmegaConf.to_container(config, resolve=True))
         OmegaConf.resolve(config)
@@ -184,19 +182,14 @@ class TaskRunner(BaseTaskRunner):
             use_critic=need_critic(config),
         )
 
-        # Download the checkpoint from HDFS to the local machine.
-        # `use_shm` determines whether to use shared memory, which could lead to faster model loading if turned on
-        local_path = copy_to_local(
-            config.actor_rollout_ref.model.path, use_shm=config.actor_rollout_ref.model.get("use_shm", False)
-        )
+        # Instantiate the tokenizer and processor from the model config.
+        from verl.utils.config import omega_conf_to_dataclass
+        from verl.workers.config import HFModelConfig
 
-        # Instantiate the tokenizer and processor.
-        from verl.utils import hf_processor, hf_tokenizer
-
-        trust_remote_code = config.data.get("trust_remote_code", False)
-        tokenizer = hf_tokenizer(local_path, trust_remote_code=trust_remote_code)
+        model_config: HFModelConfig = omega_conf_to_dataclass(config.actor_rollout_ref.model)
+        tokenizer = model_config.tokenizer
         # Used for multimodal LLM, could be None
-        processor = hf_processor(local_path, trust_remote_code=trust_remote_code, use_fast=True)
+        processor = model_config.processor
 
         resource_pool_manager = self.init_resource_pool_mgr(config)
 
--- a/verl/trainer/ppo/v1/trainer_base.py
+++ b/verl/trainer/ppo/v1/trainer_base.py
@@ -73,21 +73,19 @@ from verl.trainer.ppo.utils import (
 )
 from verl.trainer.ppo.v1.replay_buffer import DAPO_FILTERED_REWARD_COUNTS_KEY, ReplayBuffer, ReplayBufferAsync
 from verl.trainer.ppo.v1.utils import MetricsAggregator, compute_advantage_for_multi_trajectories
-from verl.utils import hf_processor, hf_tokenizer
 from verl.utils import tensordict_utils as tu
 from verl.utils.checkpoint.checkpoint_manager import find_latest_ckpt_path
 from verl.utils.config import omega_conf_to_dataclass
 from verl.utils.dataset.rl_dataset import collate_fn
 from verl.utils.debug import marked_timer
 from verl.utils.debug.metrics import calculate_debug_metrics
-from verl.utils.fs import copy_to_local
 from verl.utils.import_utils import load_extern_type
 from verl.utils.metric import reduce_metrics
 from verl.utils.py_functional import rename_dict
 from verl.utils.seqlen_balancing import calculate_workload, get_seqlen_balanced_partitions, log_seqlen_unbalance
 from verl.utils.skip import SkipManager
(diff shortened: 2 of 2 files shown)


ErenAta16 (commented):
The consolidation is a clear improvement. `HFModelConfig` already owns the whole sequence the call sites were duplicating, `copy_to_local` with `use_shm`, then `hf_tokenizer` and `hf_processor` against the local path, so deleting two copies of that and reading `model_config.tokenizer` is strictly less to keep in sync. It also picks up `tokenizer_path` handling, which the inline version did not have.

**The `trust_remote_code` source changes, and I think that is a silent behaviour change.** Before:

```python
trust_remote_code = config.data.get("trust_remote_code", False)
tokenizer = hf_tokenizer(local_path, trust_remote_code=trust_remote_code)
```

After, it comes from whatever `HFModelConfig` was built from:

```python
model_config: HFModelConfig = omega_conf_to_dataclass(config.actor_rollout_ref.model)
tokenizer = model_config.tokenizer
```

and `HFModelConfig` declares its own field with the same default:

```python
trust_remote_code: bool = False
...
self.tokenizer = hf_tokenizer(self.local_tokenizer_path, trust_remote_code=self.trust_remote_code)
```

So the value now comes from `actor_rollout_ref.model.trust_remote_code`, not from `data.trust_remote_code`. A config that sets it under `data`, which is where the old code looked and therefore where users would have put it, silently falls back to `False` after this PR. For a model that needs custom code that surfaces as a load failure from `transformers` telling the user to pass `trust_remote_code=True`, which is confusing when their YAML already says exactly that, just in the section that no longer counts.

Two questions that would settle it:

Does anything upstream copy `data.trust_remote_code` into the model section before `omega_conf_to_dataclass` runs? If so this is a non-issue and worth a comment saying so, since the two sections having the same key is otherwise a trap.

If not, is `data.trust_remote_code` being deprecated deliberately? That is a reasonable direction, model-scoped is the more sensible home for it, but it needs a release note and ideally a warning when `data.trust_remote_code` is set and the model-level one is not, rather than the setting quietly ceasing to apply.

Everything else about the change reads well. Dropping the `copy_to_local` import and the HDFS comment from both call sites is correct now that the download happens inside the config object.

ZhiliangWu:
> The consolidation is a clear improvement. `HFModelConfig` already owns the whole sequence the call sites were duplicating, `copy_to_local` with `use_shm`, then `hf_tokenizer` and `hf_processor` against the local path, so deleting two copies of that and reading `model_config.tokenizer` is strictly less to keep in sync. It also picks up `tokenizer_path` handling, which the inline version did not have.
> 
> **The `trust_remote_code` source changes, and I think that is a silent behaviour change.** Before:
> 
> ```python
> trust_remote_code = config.data.get("trust_remote_code", False)
> tokenizer = hf_tokenizer(local_path, trust_remote_code=trust_remote_code)
> ```
> 
> After, it comes from whatever `HFModelConfig` was built from:
> 
> ```python
> model_config: HFModelConfig = omega_conf_to_dataclass(config.actor_rollout_ref.model)
> tokenizer = model_config.tokenizer
> ```
> 
> and `HFModelConfig` declares its own field with the same default:
> 
> ```python
> trust_remote_code: bool = False
> ...
> self.tokenizer = hf_tokenizer(self.local_tokenizer_path, trust_remote_code=self.trust_remote_code)
> ```
> 
> So the value now comes from `actor_rollout_ref.model.trust_remote_code`, not from `data.trust_remote_code`. A config that sets it under `data`, which is where the old code looked and therefore where users would have put it, silently falls back to `False` after this PR. For a model that needs custom code that surfaces as a load failure from `transformers` telling the user to pass `trust_remote_code=True`, which is confusing when their YAML already says exactly that, just in the section that no longer counts.
> 
> Two questions that would settle it:
> 
> Does anything upstream copy `data.trust_remote_code` into the model section before `omega_conf_to_dataclass` runs? If so this is a non-issue and worth a comment saying so, since the two sections having the same key is otherwise a trap.
> 
> If not, is `data.trust_remote_code` being deprecated deliberately? That is a reasonable direction, model-scoped is the more sensible home for it, but it needs a release note and ideally a warning when `data.trust_remote_code` is set and the model-level one is not, rather than the setting quietly ceasing to apply.
> 
> Everything else about the change reads well. Dropping the `copy_to_local` import and the HDFS comment from both call sites is correct now that the download happens inside the config object.

Thanks for taking a look here @ErenAta16. This is deliberate: the model-scoped flag (actor_rollout_ref.model.trust_remote_code) is the one used across the docs and examples, so building the tokenizer/processor from it is more consistent than the old data.trust_remote_code.

wuxibin89 (approved):

Issue #7200, opened 2026-07-30
[opd, trainer] Supervised distillation loss normalization depends on micro-batch partitioning


yph22:
### System Info

verl:0.9.0.dev

### Information

- [ ] The official example scripts
- [ ] My own modified scripts

### Tasks

- [ ] An officially supported task in the `examples` folder (such as GLUE/SQuAD, ...)
- [ ] My own task or dataset (give details below)

### Reproduction

The supervised/direct on-policy distillation loss is normalized independently
inside every micro-batch because `config.global_batch_info` is empty when
`distillation_loss()` is called.

As a result, the accumulated loss and gradients depend on how the same mini-batch
is partitioned into micro-batches. With static micro-batch size 1, the effective
objective becomes a sum of per-sequence token means instead of the configured
global token mean.

This is reproducible with the following target configuration:

```yaml
distillation:
  enabled: true
  distillation_loss:
    loss_mode: forward_kl_topk
    use_task_rewards: false
    use_policy_gradient: false

actor_rollout_ref:
  actor:
    loss_agg_mode: token-mean
    use_dynamic_bsz: false
    ppo_micro_batch_size_per_gpu: 1
```

### Expected behavior

For `loss_agg_mode=token-mean`, let:

- \(D\) be the data-parallel size;
- \(N\) be the number of valid response tokens in the global mini-batch;
- \(T_{r,m}\) be the tokens on DP rank \(r\), micro-batch \(m\);
- \(\ell_t\) be the per-token distillation loss.

The intended global objective is:

\[
L =
\frac{1}{N}
\sum_{r=1}^{D}
\sum_m
\sum_{t\in T_{r,m}}\ell_t.
\]

Because DDP averages gradients across ranks, each rank/micro-batch should
backpropagate:

\[
L_{r,m} =
\frac{D}{N}
\sum_{t\in T_{r,m}}\ell_t.
\]

Therefore, splitting one mini-batch into one or multiple micro-batches must not
change either the accumulated scalar loss or the parameter gradients.

`agg_loss()` already implements this behavior when it receives
`dp_size=D` and `batch_num_tokens=N`:

```python
loss = verl_F.masked_sum(loss_mat, loss_mask) / batch_num_tokens * dp_size
```

## Actual behavior

The direct distillation branch calls:

```python
distillation_loss = agg_loss(
    loss_mat=distillation_losses,
    loss_mask=response_mask,
    loss_agg_mode=loss_agg_mode,
    **config.global_batch_info,
[...]

PR #7225, opened 2026-08-01
[algo] fix: micro-batch normalization for distillation loss

JacobHelwig:
### What does this PR do?

Applies fix proposed by @wuxibin89 for issue #7200 found by @yph22

### Test

image

- Ran Qwen3-0.6B student w/ Qwen3-1.7B teacher on gsm8k for a fixed batch using `RolloutSkip`
- `fixOff` == prior to this PR, `fixOn` == with this PR
- Total batch size for all 4 is 8, `mb{k}` == mico-batch size of size `k` (`mb1` --> 8 accumulations, `mb8` == 1 accumulations)
- `Off` with mb=1 shows 8x higher loss, after applying fix, loss is correctly normalized and is invariant to the micro-batch size 

#### PG branch

With pg enabled (suggested by @ErenAta16)

image

### DIFF
--- a/verl/trainer/distillation/losses.py
+++ b/verl/trainer/distillation/losses.py
@@ -286,6 +286,12 @@ def distillation_loss(
         # Directly backpropagate distillation loss as a supervised loss, as in https://arxiv.org/abs/2306.13649.
         if response_mask.is_nested:
             response_mask = response_mask.to_padded_tensor(False)
+
+        # global batch info for loss aggregation
+        config.global_batch_info["dp_size"] = data["dp_size"]
+        config.global_batch_info["batch_num_tokens"] = data["batch_num_tokens"]
+        config.global_batch_info["global_batch_size"] = data["global_batch_size"]
+        config.global_batch_info["loss_scale_factor"] = config.loss_scale_factor
         distillation_loss = agg_loss(
             loss_mat=distillation_losses,
             loss_mask=response_mask,
--- a/verl/trainer/distillation/losses.py
+++ b/verl/trainer/distillation/losses.py
@@ -292,7 +292,6 @@ def distillation_loss(
         # Directly backpropagate distillation loss as a supervised loss, as in https://arxiv.org/abs/2306.13649.
         if response_mask.is_nested:
             response_mask = response_mask.to_padded_tensor(False)
-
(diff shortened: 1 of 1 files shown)


ErenAta16 (commented):
The normalization values look right and the before/after curves are the kind of evidence this needs. One question about placement, because the fix lands inside one of the two branches and the other one reads the same dict earlier.

`distillation_loss` splits on `loss_config.use_policy_gradient`:

```python
if loss_config.use_policy_gradient:
    policy_loss_fn = get_policy_loss_fn(loss_config.policy_loss_mode)
    for k, v in config.global_batch_info.items():          # copies whatever is present at this point
        loss_config.global_batch_info[k] = v
    ...
    distillation_loss, pg_metrics = policy_loss_fn(..., config=loss_config, ...)
else:
    # Directly backpropagate distillation loss as a supervised loss
    ...
    config.global_batch_info["dp_size"] = data["dp_size"]  # <- this PR
    config.global_batch_info["batch_num_tokens"] = ...
    distillation_loss = agg_loss(..., **config.global_batch_info)
```

The four assignments sit after the `else`, so they only run on the supervised path. The policy-gradient path copies `config.global_batch_info` into `loss_config` further up and never sees them.

That leaves two possibilities and I cannot tell which from this file alone:

If `config.global_batch_info` is already populated by the caller before `distillation_loss` runs, then the copy in the PG branch is fine and these four lines are re-stating values that were already there, in which case the supervised path was reading them too and there would have been nothing to fix. The fact that this changes the curves suggests that is not the case.

If it is not populated by the caller, which the benchmark implies, then the PG branch copies an empty or stale dict and `use_policy_gradient=True` runs still lose the micro-batch normalization after this PR. That would make the fix cover one of the two distillation modes.

Which mode did the gsm8k run use? If it was the supervised path, the PG path is worth a second look before this lands, since #7200 does not read as mode-specific. Hoisting the four assignments above the `if` would cover both and would not change the supervised behaviour you measured.

Not asserting it is broken; the evidence I have is the branch structure, and the caller side is where it resolves.

JacobHelwig:
Thanks @ErenAta16 , you're right. Applied the fix and updated the tests

wuxibin89 (approved):

PR #7227, opened 2026-08-02
[ckpt, rollout, vllm] feat: add vLLM consumer for delta-sharded weight sync

ShuoleiWang:
### What does this PR do?

This PR adds a vLLM rollout consumer for the `delta_sharded` checkpoint-engine wire format introduced by [verl #6974](https://github.com/verl-project/verl/pull/6974).

`delta_sharded` already reduces trainer-side materialization, gather traffic, and cross-node payloads by sending only bitwise-changed checkpoint values. Before this change, only SGLang could consume its `DeltaFlush` stream, so vLLM rollouts still had to use a full-weight sync.

The difficult part is not the transport: the delta is expressed in checkpoint/Hugging Face coordinates, while vLLM runtime weights may be TP/EP shards or packed QKV/MoE tensors. This PR therefore adds a VERL vLLM `WeightTransferEngine` adapter that verifies and decodes each flush, then calls the checkpoint-patch API introduced by [vLLM #50723](https://github.com/vllm-project/vllm/pull/50723). vLLM's native `model.load_weights()` remains responsible for checkpoint-name mapping, packed layouts, TP slicing, and EP placement.

This is not a duplicate of #6974: that PR implements the common sender/wire and the SGLang consumer; this PR adds the vLLM receiver and its lifecycle.

#### vLLM dependency status

This integration requires the current four-argument vLLM `WeightTransferEngine` interface and the checkpoint-patch API from [vLLM #50723](https://github.com/vllm-project/vllm/pull/50723). VERL's current uv-managed `vllm` extra pins `vllm==0.24.0`, which provides neither capability, and as of 2026-08-28 no tagged vLLM release contains #50723; the standard pinned VERL environment therefore cannot run this path yet.

The GPU evidence below used the installed vLLM 0.26 runtime, whose `WeightTransferEngine` interface matches this integration, with the official #50723 checkpoint helper overlaid. Whether this PR should wait for a containing vLLM release or remain an opt-in feature that requires a specific vLLM main revision is pending maintainer guidance.

#### Automated CI coverage

VERL's standard CI remains pinned to vLLM 0.24.0, so this PR does not add a vLLM-main dependency or a workflow that pretends to exercise the unsupported success path. Existing CI still covers the version-independent regressions changed here:

- `tests/utils/test_bucketed_weight_transfer.py` now exercises an odd-byte `uint8` manifest/positions prefix followed by BF16 values through the existing multi-process CUDA-IPC mixed-dtype test;
- `tests/workers/rollout/rollout_vllm/test_vllm_abort.py` verifies that a pause/cache-clear failure propagates instead of allowing a weight update to continue.

No automated test in the pinned environment executes the real `delta_sharded` vLLM consumer or the #50723 loader. The supported path is covered by the H20 validation below. Ruff check/format, `git diff --check`, and workflow YAML parsing passed locally; the broader pytest suite was not rerun after the final simplification.

#### H20 TP8+EP8 bit-exact correctness

Qwen3-30B-A3B, BF16, vLLM TP8+EP8:

- 8/8 workers passed.
- 32/32 selected runtime tensors changed after sparse apply.
- 32/32 selected runtime tensors matched the dense-load target with `torch.equal` after sparse apply.
- 32/32 tensors returned to the checkpoint baseline after restore.
- Coverage includes packed QKV, norm, routed-expert `w13`, and routed-expert `w2` mappings.

This rerun used the #50723 helper with SHA256 `b6fc466dbb97cf32012cb6b46a90cb9eebb91351dc4b574473251dceef46957e`. It is a targeted TP/EP mapping probe rather than a full-model digest.

#### H20 two-node weight-sync performance

A short matched A/B run used Qwen3-30B-A3B BF16, an 8 x H20 FSDP1 trainer, and a separate 8 x H20 vLLM TP8+EP8 rollout node with Triton MoE. The run used VERL V1 `separate_async`, a 1024 MiB bucket, NCCL `NET/Socket` over `eth0` with IB disabled, and no PD disaggregation.

Both backends used the normal performance configuration: `full_determinism=false`, vLLM batch-invariant mode disabled, `NCCL_DETERMINISTIC` disabled, and NCCL algorithm/protocol auto-selection. The first update was treated as cold and excluded. Rows are matched by steady-update ordinal; nondeterministic execution does not guarantee bit-identical optimizer states across the two jobs.

| steady update | changed ratio | logical sparse payload | `delta_sharded` | NCCL full | speedup |
|---:|---:|---:|---:|---:|---:|
| 2 | 0.809% | 1413.6 MiB | 6.982 s | 55.827 s | 7.995x |
| 3 | 1.181% | 2062.6 MiB | 7.101 s | 56.093 s | 7.900x |
| mean | 0.995% | 1738.1 MiB | 7.042 s | 55.960 s | **7.947x** |

The one-time `delta_sharded` dense seed took **52.7 s** and is excluded from the table. Dense seed plus all three nonzero sparse updates completed without a checksum mismatch, CUDA-IPC/WTE failure, OOM, or traceback.

These measurements cover `timing_s/update_weights` only, not end-to-end training throughput. With two steady samples per backend, this is a performance check rather than a statistically robust long benchmark. The subsequent code simplification removed only redundant false-only IPC plumbing, internal tuple/type revalidation, repeated comments, and test scaffolding; it did not change the payload, CUDA-IPC transfer, vLLM loader, or weight-application path exercised here.

### API and Usage Example

The vLLM backend reuses the existing `delta_sharded` configuration:

```shell
actor_rollout_ref.rollout.checkpoint_engine.backend=delta_sharded \
+actor_rollout_ref.rollout.checkpoint_engine.engine_kwargs.delta_sharded.encoding=indices \
+actor_rollout_ref.rollout.checkpoint_engine.engine_kwargs.delta_sharded.verify_every=0
```

VERL injects `weight_transfer_config.backend=verl_delta_ipc` into the vLLM server configuration. Users should not configure a second vLLM weight-transfer backend for this path.

#### Architecture

```mermaid
flowchart LR
    T["Trainer delta sender<br/>(all ranks diff/gather; rank 0 publishes)"]
    C["Every GPU: CheckpointEngineWorker<br/>receive DeltaFlush + local sender"]
    W["Every vLLM GPU worker<br/>verl_delta_ipc WTE<br/>verify + decode"]
    P["vLLM #50723 CheckpointWeightPatch<br/>native model.load_weights()<br/>TP/EP/packed mapping"]
    V["Live vLLM runtime weights"]
    A["Leader ServerAdapter<br/>collective_rpc fan-out"]

    T -- "NCCL: dense values / sparse positions + values<br/>ZMQ PUB/SUB: manifest only" --> C
    C == "same-GPU CUDA IPC<br/>local ZMQ: handle / metadata / ACK" ==> W
    W --> P --> V
    A -. "control only: init / start / receive / finish<br/>(no weight tensors)" .-> W
```

All trainer ranks participate in diff/gather; rank 0 publishes the manifest over ZMQ and broadcasts payload tensors to every rollout `CheckpointEngineWorker` over NCCL. Each checkpoint worker forwards its local copy through same-GPU CUDA IPC. The leader's Ray `collective_rpc` carries lifecycle control only, never weight tensors.

#### Dense seed and sparse steady-state updates

The first update is a dense seed. After the seed completes, each trainer rank snapshots its current FSDP shard as the baseline for subsequent bit-exact diffs.

Steady updates gather only changed `(index, value)` pairs in final checkpoint/HF coordinates. Rank 0 buckets and broadcasts those pairs through the existing checkpoint communicator. Each rollout worker forwards one flush at a time to its colocated vLLM worker, which verifies the checksum and converts the manifest into checkpoint-coordinate patches.

The sparse wire and trainer gather are O(nnz), but runtime application is not: the current vLLM API creates a full checkpoint-shaped NaN tensor per patch and runs a dense masked copy after native loader mapping.

#### Main code changes

- add the `verl_delta_ipc` vLLM `WeightTransferEngine` consumer;
- register it before vLLM constructs worker-side transfer engines;
- stream `delta_flush` items through the existing ZMQ/CUDA-IPC channel;
- use vLLM's layerwise reload lifecycle for the dense seed and in-place updates for sparse steady-state patches;
- checksum every flush and lock a worker after an apply failure;
- align mixed-dtype tensors inside the IPC byte bucket;
- document the supported vLLM configuration and its dependency boundary.

#### Failure semantics

Updates are not transactional. A failure may leave rollout weights partially updated, and trainer-side delta snapshots may already have advanced. An in-place retry is not supported. The affected run must be stopped, or the entire weight-sync session must be rebuilt and densely reseeded before serving or applying another update.

#### Supported scope

- disaggregated checkpoint-engine flow (`hybrid_engine=False`);
- BF16, non-quantized vLLM models;
- CUDA IPC only; no shared-memory fallback;
- `pipeline_parallel_size=1`;
- `data_parallel_size=1`;
- no PD disaggregation, speculative decoding, EPLB, or `verify_every > 0`;
- vLLM MoE requires the Triton backend;
- the currently validated model is Qwen3-30B-A3B TP8+EP8;
- the companion vLLM sparse loader currently requires NaN-preserving, same-shaped, single-copy weight loaders.

### AI assistance disclosure

OpenAI Codex was used for code exploration, implementation assistance, test orchestration, and review.


ShuoleiWang:
Hi @wuxibin89, vLLM #50723 is the prerequisite for this PR, and it has been merged but currently only available on vLLM main. However, VERL currently pins vllm==0.24.0, so this path cannot run in the standard pinned environment.
Do you have any suggestions on how I should handle this dependency?

wuxibin89 (approved):

PR #7195, opened 2026-07-29
[veomni] fix: Fix gpt oss gather with ep expert transfer error

wyettzeng:
### What does this PR do?

Bug 1: vLLM expects all expert tensors to be received together for gpt-oss. However, for Veomni with ep > 1, each rank will only send over the experts that it has. Therefore, we will need to gather all the experts before sending it to vLLM during actor-rollout sync.

Bug 2: The old VeRL sync path loaded checkpoint-format actor weights directly into an already-initialized vLLM model whose MoE parameters had been rearranged into a backend-specific FlashInfer kernel layout, then ran weight post-processing again. Because vLLM’s checkpoint loaders assume ordinary logical layouts, they could write expert shards into incorrect physical locations, corrupting the model and producing incomprehensible output. The new path calls vLLM’s official reload_weights(..., is_checkpoint_format=True), which reconstructs each layer in checkpoint layout, loads all its weights, applies the backend conversion exactly once, and copies the result back into the original storage so CUDA-graph references remain valid; VeRL also clones IPC tensors so deferred layer loading cannot observe an overwritten communication buffer. This issue is found independently by:
- Wyett when trying to do veomni + vllm colocated ipc weight sync on gpt-oss-20b BF16
- Zhiyue when trying to do fully async weight sync for qwen 30b a3b

### Checklist Before Starting

- [x] Search for similar PRs. Paste at least one query link here: ...
- [x] Format the PR title as `[{modules}] {type}: {description}` (This will be checked by the CI)
  - `{modules}` include `fsdp`, `megatron`, `veomni`, `sglang`, `vllm`, `rollout`, `trainer`, `ci`, `training_utils`, `recipe`, `hardware`, `deployment`, `ray`, `worker`, `single_controller`, `misc`, `perf`, `model`, `algo`, `env`, `tool`, `ckpt`, `doc`, `data`, `cfg`, `reward`, `fully_async`, `one_step_off`
  - If this PR involves multiple modules, separate them with `,` like `[megatron, fsdp, doc]`
  - `{type}` is in `feat`, `fix`, `refactor`, `chore`, `test`
  - If this PR breaks any API (CLI arguments, config, function signature, etc.), add `[BREAKING]` to the beginning of the title.
  - Example: `[BREAKING][fsdp, megatron] feat: dynamic batching`

### Test

> For changes that can not be tested by CI (e.g., algorithm implementation, new model support), validate by experiment(s) and show results like training curve plots, evaluation results, etc.

### API and Usage Example

[...]

PR #7243, opened 2026-08-03
[veomni] fix: preserve GPT-OSS weights without expert parallelism

Luosuu:
## What does this PR do?

Fix VeOmni full-weight synchronization for GPT-OSS when expert parallelism is disabled.

GPT-OSS stores all experts in fused checkpoint tensors, with gate and up values interleaved along the last dimension. The generic VeOmni MoE exporter treated those tensors like Qwen-style expert stacks, split the wrong dimension, and converted them into per-expert parameter names. This PR selects a pass-through handler for non-EP GPT-OSS so vLLM receives the original packed names, shapes, and values.

GPT-OSS EP full export and `delta_sharded` export are explicitly rejected with `NotImplementedError` instead of silently producing incompatible rank-local representations.

## Why this is not a duplicate

This is materially different from #7195. That PR gathers GPT-OSS expert tensors across EP ranks and changes the vLLM reload/bucketed-transfer path. This PR deliberately does not implement EP support or modify vLLM: it is a narrow fix for the already-global non-EP tensor, while keeping EP unchanged and `delta_sharded` unsupported.

## Test

```text
pytest -q tests/utils/veomni/test_moe_param_export_on_cpu.py tests/workers/rollout/test_vllm_moe_param_converter_on_cpu.py
# 6 passed

pre-commit run --all-files --show-diff-on-failure
# all hooks passed
```

AI assistance was used to inspect the relevant weight-export paths, implement the change, and draft tests and this description. The human submitter reviewed every changed line, understands the change end-to-end, and ran the relevant tests.

### DIFF
--- a/verl/workers/engine/veomni/utils.py
+++ b/verl/workers/engine/veomni/utils.py
@@ -134,25 +134,25 @@ def default_moe_param_handler(name, tensor, expert_id_base):
         yield from _map_moe_params_common(key, value, expert_id_base)
 
 
-# Overrides the default MoE parameter mapping per model_type. Handlers follow the
-# ``default_moe_param_handler`` contract: (name, stacked_tensor, expert_id_base).
-MOE_PARAM_HANDERS = {}
-
-
 def passthrough_moe_param_handler(name, tensor, expert_id_base):
     """Keep a fused MoE checkpoint tensor unchanged."""
     del expert_id_base
     yield name, tensor
 
 
+# Overrides the default MoE parameter mapping per model_type. Handlers follow the
+# ``default_moe_param_handler`` contract: (name, stacked_tensor, expert_id_base).
+MOE_PARAM_HANDERS = {"gpt_oss": passthrough_moe_param_handler}
+
+
 def get_moe_param_handler(model_type, ep_enabled):
     """Resolve the checkpoint export handler for the current MoE layout."""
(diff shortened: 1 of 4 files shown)


verl/workers/engine/veomni/utils.py
```diff
@@ -139,6 +139,24 @@ def default_moe_param_handler(name, tensor, expert_id_base):
...
+
+
+def get_moe_param_handler(model_type, ep_enabled):
+    """Resolve the checkpoint export handler for the current MoE layout."""
+    # GPT-OSS stores all experts in fused tensors with gate/up values interleaved
+    # along the last dimension. When EP is disabled, the tensor is already global
+    # and vLLM expects this packed checkpoint layout unchanged. The EP path still
+    # needs a shard-aware vLLM loader and intentionally keeps the existing handler.
+    if model_type == "gpt_oss" and not ep_enabled:
+        return passthrough_moe_param_handler
```
Luosuu: we can register passthrough_moe_param_handler for gpt_oss in MOE_PARAM_HANDERS?

Luosuu: Done. GPT-OSS now registers passthrough_moe_param_handler in MOE_PARAM_HANDERS. The resolver explicitly rejects EP updates so rank-local expert shards cannot be sent under the packed global names.

Issue #7167, opened 2026-07-27
[ckpt]: NCCL checkpoint engine broadcast optimization


parinayc20:
# Feature Request

Adding all the trainer GPUs (or possibly just the trainer GPUs located on the same node as the master GPU) to the broadcast group when broadcasting updated weights from the master trainer rank to the rollout ranks results in higher transfer bandwidth and lower end-to-end latency in the cases I've tested, without incurring drastic side effects.

# Motivation

Below is the experiment that led me to this proposal.

I'm working with GH200 nodes connected as follows:

<img width="987" height="589" alt="Image" src="https://github.com/user-attachments/assets/b5a39b86-2ba1-4fd5-aa7f-33643438f893" />

I laid out trainer and rollout in a 4x4 split — 4 trainer GPUs on one node, 4 rollout GPUs on another. With only the master sending weights to the rollout GPUs, NCCL broadcast bandwidth was capped at ~25 GB/s. However, each node has 4 NICs available, giving a bisection bandwidth of ~100 GB/s between the two nodes. After adding all the trainer GPUs (except the master) to the broadcast group as well, NCCL was able to make use of the remaining 3 NICs too, bringing the raw broadcast bandwidth of the operation to ~90 GB/s.

The measurements below include the changes introduced in #7083 and #7107, and a lean fix for the issue described in #7015. The trainer side runs FSDP2 for weight distribution.

The following table shows the time taken for a weight sync of Qwen2.5-14B-Instruct using the NCCL checkpoint engine:

| Configuration | Time taken per weight sync |
|---|---|
| Trainer ranks not included in the broadcast group | 1.5 s |
| Trainer ranks included in the broadcast group | 0.62 s |

Given this, I think it would be useful to add a mechanism to optionally include all trainer GPUs — or just the trainer GPUs co-located with the master's node — in the broadcast group.

Moreover, this change also fits easily in the current implementation.

Happy to hear any suggestions on this and to provide more information wherever needed.

kaining-never-stop:
Hi @parinayc20, I traced the current NCCL checkpoint path, and the reported result is consistent with its topology: `build_topology()` includes only trainer rank 0 in the collective, while the remaining trainer ranks consume the export generator without participating in the broadcast. NCCL therefore cannot use the other trainer-side GPUs/NICs as relay participants.

I would be happy to work on a focused implementation if you are not already preparing one:

1. add an opt-in NCCL engine option, disabled by default, to include all trainer ranks;
[...]

Issue #7262, opened 2026-08-04
[RFC] `nccl_parallel` checkpoint engine: aggregate sender NICs for cross-node weight sync


savaresejeremy:
### Summary

We propose an additive checkpoint-engine backend, `nccl_parallel`, that removes the single-sender bottleneck in the `nccl` engine's cross-node weight sync. Instead of actor rank 0 broadcasting every bucket, all S actor ranks send concurrently, each owning a stripe of the global bucket sequence over its own NCCL group. Receivers join all S groups and consume buckets in ascending global order, so the reassembled stream is identical to the stock engine's.

We have this running in production-scale RL (FSDP2 trainer + disaggregated vLLM rollout) and would like to contribute it. There is already an open request for this capability (#7167), and a measurement posted there on 2026-08-03 shows that the straightforward version is correct but ~19x slower end-to-end, because the ordering needed to make trainer-side broadcasts safe removed the existing export/broadcast overlap. Most of this engine's hardening exists to keep that overlap. This RFC describes the design, the hardening, and a phased contribution plan. The PR is up at #7263: one new engine file, registry/README/config wiring, and two GPU tests.

### Motivation: the single-sender funnel

Every trainer-side transport in `verl/checkpoint_engine` today funnels the full model through one rank's NIC share. Three observations motivate this design.

1. Upstream's own benchmark shows the ceiling. The package README measures the NCCL backend at ~7 s / 8.25 GB/s for a 30B model on 4x8 H100 with ConnectX-7 400 Gbps InfiniBand. That is roughly 66 Gbps on a 400 Gbps fabric, because one GPU's network share is the source for the whole rollout fleet. (That setup receives weights without the CUDA-IPC hop to vLLM/SGLang, so it is not directly comparable to the end-to-end refit times below; we cite it only for the single-sender ceiling.)

2. There is an open request for exactly this capability. #7167 ("[ckpt]: NCCL checkpoint engine broadcast optimization") reports that on GH200 nodes with 4 NICs per node (~100 GB/s bisection), the master-only broadcast caps at ~25 GB/s. Adding all trainer GPUs to the broadcast group, with rank 0 remaining the sole source and the others joining as passive participants whose NICs NCCL can use, raised raw broadcast bandwidth to ~90 GB/s and cut Qwen2.5-14B-Instruct weight sync from 1.5 s to 0.62 s (2.4x). The issue proposes an optional mechanism to include the trainer GPUs in the broadcast group.

[...]

PR #7263, opened 2026-08-04
[ckpt] feat: add nccl_parallel checkpoint engine (all actor ranks send)

savaresejeremy:
### What does this PR do?

Adds an additive checkpoint-engine backend, `nccl_parallel`, that removes the single-sender bottleneck in cross-node weight sync. All S actor ranks send concurrently, each owning a stripe of the global bucket sequence over its own NCCL group; receivers join all S groups and reassemble in ascending global order. With S=1 it degenerates to a single-sender engine. RFC: https://github.com/verl-project/verl/issues/7262.

This is the capability issue #7167 asks for, built with a different mechanism. The passive-participant topology measured there on 2026-08-03 came out ~19x slower end-to-end because the cross-group ordering it needed removed the export/broadcast overlap. This engine pins collective enqueue order with bulk-synchronous phase barriers and keeps execution overlapped through non-blocking comm streams and per-transfer CUDA events, so the export keeps running while broadcasts are in flight. Mechanism details are in the module docstring.

### Measurements

2-node, 8-GPU FSDP2 trainer to 8-GPU vLLM rollout, 27B bf16 model, GEN_TP=4, bucket 2048MB, AWS EFA, n=20 refits: 2.51s ±0.27 per refit vs 3.01s ±0.25 for the stock `nccl` engine. Plateau accuracy (steps 61-100 of 100-step GSM8K runs) matched stock within 0.0005. A same-window paired re-run (n=20 each, mean/median/p95) gave stock 3.005/2.911/3.533 s vs parallel 2.457/2.466/2.818 s. Both arms ran code behaviorally identical to this PR at S=8 (the 16-byte bucket rounding below is a no-op at 2048MB/8).

4-node validation (same recipe, 10 refits per arm, one run per topology, phase=8): 1T+3R (S=8) 2.480 s, 2T+2R (S=16) 5.548 s, 3T+1R (S=24) 2.931 s. All completed with normal learning curves, and sync time did not move when the receiver fleet tripled. The 2T+2R split is slower at fixed phase and we don't have an explanation for that yet. A second pass of the matrix on the exact submitted tree reproduced the ordering (2.84/5.01/3.88 s, with 2T+2R again slowest); these runs are correctness evidence, not tuned benchmarks.

A note on bucket sizing: the per-group bucket is rounded down to a 16-byte boundary, with a CPU regression test. Without the rounding, 2048MB/24 senders is an odd number of bytes, and a bucket-sized cut through a large tensor leaves the tensors behind it at a dtype-misaligned offset (S=8 and S=16 divide 2048MB evenly and are unaffected). The rounding removes the bucket cut as a source of misalignment but is not a general alignment guarantee: back-to-back packing of a mixed-dtype stream can still misalign a later tensor at any bucket size. That behavior predates this engine, is shared with the stock `nccl` engine, and does not arise on the homogeneous bf16 export path, so we think it belongs in its own fix rather than this PR.

### Duplicate-work check

- The open request for this is issue #7167. Its 2026-08-03 comment documents a measured ~19x-slower attempt (passive-participant topology), and on 2026-08-04 the issue author said they have a working implementation and will open a PR. Our RFC: https://github.com/verl-project/verl/issues/7262; our comment answering the open questions in that thread: https://github.com/verl-project/verl/issues/7167#issuecomment-5183679960. As of submission no PR implementing multi-sender transfer is open. If the announced PR appears, the designs differ architecturally (passive relay participants around one source there, active per-stripe senders with phase-pinned enqueue ordering here) and we would want to benchmark them against each other rather than treat either as a duplicate.
- PR #7107 (bucket-size fix for the sibling `nccl` engine) is open. We are prepared to rebase and match its partial-bucket metadata behavior once it lands.
- `kimi_ckpt_engine` is inference-side (rotating owner, one bucket in flight, never concurrent with trainer collectives), so it solves a different problem.

### AI assistance disclosure

This contribution was developed with AI assistance (Claude, via internal agent tooling) covering code, tests, and documentation drafting. All changes were human-reviewed line by line and the branch was reviewed internally before submission; commits carry `Co-authored-by: Claude` trailers per AGENTS.md.

### Test

CPU (no GPU required; phase validation and bucket-size alignment):
```bash
pytest tests/checkpoint_engine/test_parallel_correctness_on_gpu.py \
  -k "invalid_phase or bucket_size_alignment" -q
```

GPU (single 8-GPU node; model defaults to ~/models/Qwen/Qwen3-8B-Base):
```bash
pytest tests/checkpoint_engine/test_parallel_stream_mechanisms_on_gpu.py -x -q
pytest tests/checkpoint_engine/test_parallel_correctness_on_gpu.py -x -q
```

Results on our cluster (8x B200, 2026-08-04, the exact submitted branch content rebased on 1ae9455):
- mechanisms: 1 passed in 34.83s
- correctness: 8 passed in 434.69s (0:07:14): 6 GPU cases (S=1 degenerate; S=2 default-phase; S=2 rebuild_group=True; S=4 default; S=4 phase=3; S=4 phase=2) x 3 update rounds each, plus 2 CPU cases (invalid-phase, bucket-size alignment)

CI-cost note for maintainers: the correctness suite measured about 7 minutes wall clock on an 8x B200 node (six cases, each an 8B-model FSDP init plus 3 sync rounds; the (1, N-1) case stages the model in 7 rollout workers' host RAM), so budget more than that on L20x8. We can trim cases (for example, drop one phase<S variant) or move it to a `test_special_*` name, though nothing runs those without dedicated workflow wiring. Whatever the maintainers prefer.

### DIFF
--- a/tests/checkpoint_engine/test_parallel_correctness_on_gpu.py
+++ b/tests/checkpoint_engine/test_parallel_correctness_on_gpu.py
@@ -1,0 +1,179 @@
+# Copyright 2026 Amazon.com Inc and/or its affiliates
+# Copyright 2024 Bytedance Ltd. and/or its affiliates
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Correctness test for the parallel-sender NCCL checkpoint engine.
+
+Mirrors test_correctness_on_gpu.py, but with backend="nccl_parallel": one
+global bucket sequence round-robined across S = num_trainer senders, received
+in ascending global order in bulk-synchronous phases. The rollout workers'
+check_weights() compares against the trainer's weights, so a pass certifies
+the striped/reordered stream reassembles correctly.
+
+Three rounds per case, because the cross-round state (bound PUB sockets,
+connected SUB sockets, end-of-stream markers) is where regressions hide: a
+round that leaves an unread marker on a SUB channel corrupts the next one.
+
+The phase < S cases exercise the leftover-marker drain path (groups that are
+never read past their last data bucket still hold a queued end-of-stream
+marker).
+
+The rebuild_group=True case destroys every group this worker joined (including
+the senders' coordination group) in finalize() and re-creates them on the next
+round. That is the path the final sender-stream wait exists for: ray's
+destroy_collective_group does not synchronize, so a still-in-flight exit
+barrier would otherwise be torn down under the collective.
+
+Run (single 8-GPU node):
+  python3 -m pytest tests/checkpoint_engine/test_parallel_correctness_on_gpu.py -x -s
+"""
+
+import os
+
+import pytest
+import ray
+import torch
+
+from tests.checkpoint_engine.test_utils import create_rollout_worker_group, create_trainer_worker_group
+from verl.checkpoint_engine import CheckpointEngineManager
+from verl.single_controller.ray.base import (
+    RayResourcePool,
+    split_resource_pool,
+)
+from verl.utils.ray_utils import auto_await
+from verl.workers.config import CheckpointEngineConfig, HFModelConfig, RolloutConfig
+
+_ngpus = torch.cuda.device_count()
+
+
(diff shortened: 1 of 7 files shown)


wuxibin89:
@savaresejeremy Thanks for you contribution, busy on v0.9.0 release for now, will review it ASAP.

savaresejeremy:
@wuxibin89 I also raised #7318, an independent fix for the mixed-dtype packing alignment issue mentioned in this PR's text. It applies to the existing engines and does not depend on this PR.

savaresejeremy:
Closing this. #7291 landed the multi-sender capability #7167 asked for with a much smaller change to the existing nccl engine, and carrying a second backend for the same goal is not worth the review cost. Thanks @wuxibin89. #7318 is independent of this PR and still stands on its own. If a striped-sender topology is ever wanted for a shape node-local relays cannot cover, the branch and the numbers in #7262 stay available.

PR #7271, opened 2026-08-05
[reward] feat: batch-capable RewardManager for experimental reward_loop

Junxiao-Zhao:
### What does this PR do?

Existing reward managers under verl/experimental/reward_loop only expose run_single, so RewardLoopWorker.compute_score_batch has to fan out N tasks per chunk and calls the user compute_score N times. For rule-based rewards that can be evaluated in one shot (e.g. batched tokenizer decoding + a single compute_score call), this is a straight throughput loss.

Introduce a `run_batch` contract without breaking existing managers:

* RewardManagerBase now defines `async run_batch(data)` with a default fan-out implementation (asyncio.create_task + asyncio.gather over run_single per sample). `run_single` stays `@abstractmethod`. Legacy managers (naive/dapo/gdpo/limited/remote) inherit the default and behave exactly like the previous per-sample fan-out.
* RewardLoopWorker.compute_score_batch now dispatches by priority `custom_reward_function.path > reward_model.enable > default rule-based`. Custom and rule-based branches call run_batch(data) unconditionally; the disrm branch keeps per-sample compute_score_disrm fan-out. The single-sample compute_score entry point is unchanged so callers like agent_loop.py aren't affected.
* New BatchRewardManager (registered as "batch") consumes the whole chunk in one call using the plural signature that already exists in verl/workers/reward_manager/batch.py (data_sources / solution_strs / ground_truths / extra_infos). It shallow-copies extra_info at the top level so per-sample num_turns / rollout_reward_scores injection doesn't overwrite sibling rows that share the same dict reference from agent_loop's np.array([v] * n) broadcast. The `**kwargs` surface (reward_router_address /  reward_model_tokenizer when the router is configured) mirrors NaiveRewardManager exactly, so users can move a reward function between the two managers by only switching to the plural signature.
* BatchRewardManager.run_single delegates to run_batch(data[-1:])[0], keeping naive.run_single's multi-sequence fallback semantics.

Default reward_manager.name stays "naive"; the new "batch" value is opt-in.

### Checklist Before Starting

- [x] Search for similar PRs. Paste at least one query link here: batch reward loop 

### Test

Tests in tests/experimental/reward_loop/test_batch_reward_manager.py cover sync/async user functions, dict/float returns, run_single delegation, batch=1 and batch>1 paths, cur_steps propagation, worker dispatch priority (custom > disrm > default), disrm still fanning out per sample, kwargs threading through _call_with_kwargs, and same-batch extra_info isolation.

### API and Usage Example

```python
```

### Design & Code Changes

> [!IMPORTANT]

### DIFF
--- a/verl/experimental/reward_loop/reward_loop.py
+++ b/verl/experimental/reward_loop/reward_loop.py
@@ -136,11 +136,14 @@ class RewardLoopWorker:
         )
 
     async def compute_score_batch(self, data: DataProto) -> list[dict]:
-        tasks = []
-        for i in range(len(data)):
-            tasks.append(asyncio.create_task(self.compute_score(data[i : i + 1])))
-        outputs = await asyncio.gather(*tasks)
-        return outputs
+        if self.config.reward.custom_reward_function.path is not None:
+            # custom reward function takes precedence over disrm
+            return await self.reward_manager.run_batch(data)
+        if self.config.reward.reward_model.enable:
+            # disrm: keep per-sample HTTP fan-out, do not batch
+            tasks = [asyncio.create_task(self.compute_score_disrm(data[i : i + 1])) for i in range(len(data))]
+            return list(await asyncio.gather(*tasks))
+        return await self.reward_manager.run_batch(data)
 
     async def compute_score(self, data: DataProto) -> dict:
         if self.config.reward.custom_reward_function.path is not None:
--- a/docs/advance/reward_loop.rst
+++ b/docs/advance/reward_loop.rst
@@ -137,8 +137,9 @@ The ``RewardManager`` maintains a reward function and defines its computation lo
 - **dapo**: DAPO implementation with an overlong reward penalty.
 - **limit**: Restricts the concurrency of the reward function, useful when external API calls are rate-limited.
 - **remote**: Runs in a separate process, effective for CPU-intensive tasks such as ``Math-Verify``.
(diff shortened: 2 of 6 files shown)


verl/experimental/reward_loop/reward_loop.py
```diff
@@ -136,11 +136,14 @@ def _init_reward_fn(self):
         )
 
     async def compute_score_batch(self, data: DataProto) -> list[dict]:
```
wuxibin89, line 138: `compute_score_batch` is only used with colocated GenRM/DistRM model, which should request vllm/sglang server per-sample.

While for rule-based reward, once a sample rollout finish, it's sent to reward worker to calculate reward imediately:
https://github.com/verl-project/verl/blob/main/verl/experimental/agent_loop/agent_loop.py#L1019-L1023

Overall, I don't think we need a batch reward manager.

Junxiao-Zhao: Hi Xibin, thanks for your review!

We added the batch reward manager for this scenario: each sample needs multiple separate GenRM calls to score different evaluation criteria. By batching the same criterion across all samples, the shared instruction prompt stays identical for every sample in the batch. This significantly boosts the prefix KV cache hit rate, avoids redundant prompt computation, and delivers a clear speedup to end-to-end training.

It also doesn't interfere with the existing per-sample path — rule-based rewards and other cases where batching adds no value still run sample-by-sample as usual.

Junxiao-Zhao:
Hi @wuxibin89 @yyDing1 , could you please review this PR again? Thanks!

yyDing1:
I read your response, but something seems off. All reward requests go through the same reward router, which dispatches them, so the batch reward manager does not seem able to guarantee that samples sharing the same prefix are routed to the same inference server.

Moreover, to improve the KV-cache hit rate, it would be more appropriate to handle this at the router level.

PR #7291, opened 2026-08-06
[ckpt] feat: Node-local multi-sender broadcast in NCCL checkpoint engine

parinayc20:
### What does this PR do?

Adds an opt-in `multi_sender` mode to the NCCL checkpoint engine that admits actor rank 0's node-local peers (and hence probably connected via high-speed interconnects) into the broadcast group as relays, so NCCL can drive more than one NIC per node during weight sync.

Addresses #7167. With a single sender, broadcast bandwidth is capped by the one NIC reachable from the source GPU (~25 GB/s on a GH200 node with 4 NICs and ~100 GB/s bisection). Admitting the source's NVLink peers gives NCCL's broadcast tree fan-out points it can reach over NVLink and push on from over their own NICs.

The relays carry no data the rollout needs — rank 0 remains the sole source, and there is still exactly one group and one broadcast per bucket. Actor workers on any other node are excluded (rank `-1`): they would have to pull a full copy over the fabric to contribute nothing.

Relationship to open PRs:
- **#7263** (`nccl_parallel`) targets the same issue by a different mechanism: a new backend where S actor ranks each send a stripe of the bucket sequence over S separate NCCL groups. This PR instead keeps the existing single-group broadcast and changes only group *membership*, as a flag on the existing `nccl` engine. The two are alternatives; see the discussion on #7167.
- **#7107** (bucket-size fix) is a prerequisite. Relays size each broadcast to match the root, so they depend on the root sending `send_buf[:offset]`. This PR is rebased on / assumes #7107; it does not restate that change.

### Checklist Before Starting

- [x] Search for similar PRs. Paste at least one query link here: https://github.com/verl-project/verl/pulls?q=is%3Apr+is%3Aopen+nccl+checkpoint+engine+broadcast and https://github.com/verl-project/verl/pulls?q=is%3Apr+is%3Aopen+7167

### Test

**GPU correctness and latency:** 

The following measurements include the mean over 10 iterations run on the same hardware as #7167. The measurements include the time taken for all-reduce + broadcast + copy the tensors on a non-blocking stream on the receiver side. All my measurements are based on a 4-trainer (1 node)+ 4-rollout (1 node) configuration. 1 warm-up iteration was run before starting with the measurements, and the iterations were run with `rebuild_group: False`.

| Model | Mode | Mean (s) | Min–Max (s) |
|---|---|---:|---:|
| Qwen3.5-27B | single-sender | 5.026 | 5.02–5.04 |
| Qwen3.5-27B | **multi-sender** | **1.791** | 1.78–1.86 |
| Qwen3-30B-A3B | single-sender | 5.570 | 5.56–5.58 |
| Qwen3-30B-A3B | **multi-sender** | **1.892** | 1.86–1.92 |

Speedup: 2.81× on Qwen3.5-27B, 2.94× on Qwen3-30B-A3B.

**Effects of scaling out**

Scaling the rollout side out costs nothing measurable. A 4-trainer (1 node) + 12-rollout (3 nodes) configuration syncs Qwen3.5-27B in 1.82–1.88 s under multi-sender, within noise of the 1.78–1.86 s measured against 4 rollout workers on a single node.

Scaling the *trainer* side across nodes does cost, but the cost is not in the broadcast. An 8-trainer (2 nodes) + 8-rollout (2 nodes) configuration syncs the same model in 2.60–2.62 s under multi-sender versus 5.85–5.90 s single-sender — still a 2.25× speedup. Both modes slow by nearly the same absolute amount relative to their single-trainer-node numbers (+0.82 s and +0.85 s), which points at a fixed cost incurred ahead of the broadcast rather than anything in the transfer itself: gathering full parameters now crosses the 25 GB/s CXI fabric instead of staying on NVLink. Only rank-0 nodes contribute relays by design, so the second trainer node adds gather cost without adding senders.

### API and Usage Example

Off by default; existing behavior is unchanged.

```python
CheckpointEngineConfig(
    backend="nccl",
    update_weights_bucket_megabytes=2048,
    engine_kwargs={"nccl": {"multi_sender": True}},
)
```

The master's setting governs the whole group: it travels in `MasterMetadata`, so `build_topology` (a classmethod on the driver) resolves one mode for every worker.

### Design & Code Changes

Rank 0 stays the only source of weights. `multi_sender` changes only who else is in the group.

- **`WorkerMetadata`** — `prepare()` now reports each worker's Ray node id, since `build_topology` runs on the driver and cannot query worker placement itself. Only rank 0 fills the nested `master` field. Node id is the locality key: a node never spans more than one NVLink domain, so matching on it can under-select peers but never over-select them.
- **`build_topology`** dispatches to `_single_sender_ranks` (upstream behaviour, `[0, -1, …]`) or `_multi_sender_ranks` (rank 0 plus every actor worker on its node, others `-1`), then assembles the kwargs once. Senders take the low ranks.
- **`num_senders`** is passed to `init_process_group` and splits the group: ranks below it send or relay, ranks at or above it consume. It also fixes an incidental leak — only true consumers now subscribe to the zmq bucket-metadata topic, where previously every `rank > 0` did.
- **`_relay_weights`** walks the same weight stream as the root via the new `get_weight_chunks_size` helper, which splits on identical boundaries to `split_weight_chunks` without materializing buffers, and issues a matching broadcast per bucket while dropping the payload. Boundaries must be derived locally rather than read off the wire: producing a weight is itself collective over the actor group, so blocking on the wire would deadlock against rank 0's gathers.
- **Root wait placement** is now mode-dependent. Single sender waits just before enqueueing the next broadcast, preserving the fill/send overlap the `send_buf`/`recv_buf` swap exists for. With relays present the root waits as soon as the bucket is filled, so the root and the relays cannot issue the broadcast group's and the actor group's collectives in different orders, hence preventing a deadlock.

AI assistance (claude code) was used to refactor this code and build the PR.

**Why this makes sense after NCCL M2N**

This PR proposes a small change to the existing legacy NCCL backend without introducing new collectives or a completely new backend. Moreover, M2N currently relies on GIN for inter-NVL domain communication, which is not supported on certain backends such as CXI.

### DIFF
--- a/verl/checkpoint_engine/base.py
+++ b/verl/checkpoint_engine/base.py
@@ -539,8 +539,8 @@ class CheckpointEngineManager:
 
 
 async def split_weight_chunks(
-    weights: Generator[tuple[str, torch.Tensor], None, None], bucket_size: int
-) -> AsyncGenerator[tuple[TensorMeta, torch.Tensor], None]:
+    weights: Generator[tuple[str, torch.Tensor], None, None], bucket_size: int, meta_only: bool = False
+) -> AsyncGenerator[tuple[TensorMeta, torch.Tensor | None], None]:
     """Split the weight into chunks.
 
     Args:
@@ -563,30 +563,7 @@ async def split_weight_chunks(
                 chunk_size=chunk_size,
                 offset=None,
             )
-            yield (tensor_meta, buffer[chunk_offset : chunk_offset + chunk_size])
-            chunk_offset += chunk_size
-
-
-async def get_weight_chunks_size(
-    weights: Generator[tuple[str, torch.Tensor], None, None], bucket_size: int
-) -> AsyncGenerator[int, None]:
-    """Yield the size of each weight chunk, without slicing the chunk itself.
-
-    Splits on exactly the same boundaries as `split_weight_chunks`, so a caller that only needs to
-    keep in step with a peer walking that generator can use this instead and skip the buffers.
-
-    Args:
-        weights: The weights generator.
-        bucket_size: Max bucket size in bytes.
-
-    Yields:
-        The size of the weight chunks.
-    """
-    async for _name, weight in ensure_async_iterator(weights):
-        chunk_offset = 0
-        while chunk_offset < weight.nbytes:
-            chunk_size = min(bucket_size, weight.nbytes - chunk_offset)
-            yield chunk_size
+            yield (tensor_meta, None if meta_only else buffer[chunk_offset : chunk_offset + chunk_size])
             chunk_offset += chunk_size
 
 
--- a/verl/checkpoint_engine/base.py
+++ b/verl/checkpoint_engine/base.py
@@ -567,6 +567,29 @@ async def split_weight_chunks(
             chunk_offset += chunk_size
 
 
+async def get_weight_chunks_size(
+    weights: Generator[tuple[str, torch.Tensor], None, None], bucket_size: int
+) -> AsyncGenerator[int, None]:
+    """Yield the size of each weight chunk, without slicing the chunk itself.
+
+    Splits on exactly the same boundaries as `split_weight_chunks`, so a caller that only needs to
+    keep in step with a peer walking that generator can use this instead and skip the buffers.
+
+    Args:
+        weights: The weights generator.
+        bucket_size: Max bucket size in bytes.
+
+    Yields:
+        The size of the weight chunks.
+    """
+    async for _name, weight in ensure_async_iterator(weights):
+        chunk_offset = 0
+        while chunk_offset < weight.nbytes:
+            chunk_size = min(bucket_size, weight.nbytes - chunk_offset)
+            yield chunk_size
+            chunk_offset += chunk_size
+
+
 async def merge_weight_chunks(
     chunks: Generator[tuple[TensorMeta, torch.Tensor], None, None], bucket_size: int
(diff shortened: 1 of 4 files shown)


verl/checkpoint_engine/base.py
```diff
@@ -567,6 +567,29 @@ async def split_weight_chunks(
             chunk_offset += chunk_size
 
 
+async def get_weight_chunks_size(
```
wuxibin89: Can we reuse `split_weight_chunks`? If divergence happened, the broadcast may hang forever.

verl/checkpoint_engine/nccl_checkpoint_engine.py
```diff
@@ -111,6 +128,9 @@ class NCCLCheckpointEngine(CheckpointEngine):
         rebuild_group (bool): Whether to rebuild the NCCL process group in each update. Defaults to False.
         is_master (bool): Whether the current process is the master process. Defaults to False.
         rollout_dtype (torch.dtype): The dtype of the weights received from rollout workers. Defaults to torch.bfloat16.
+        multi_sender (bool): Whether to also admit the source's NVLink-local actor workers into the
```
wuxibin89, line 130: Should we enable `multi_sender` by default?

verl/checkpoint_engine/base.py
parinayc20: Yes, makes sense. I added a `meta_only` parameter to the `split_weight_chunks` function rather.

wuxibin89 (approved):

PR #7296, opened 2026-08-06
[rollout] feat: continuous_token support Assistant Role in Merge and Encode Context

gxlvera:
## Summary

This PR is stacked on unmerged pr #6804. It updates Continuous Token context merging to support rollback-style agent harness behavior.

Previously, CT exposed `merge_non_assistant_*` APIs for harness-returned messages, assuming those incremental messages only contained non-assistant roles such as `tool`, `user`, or `system`.

Some agent harnesses may rewrite a previous assistant turn during rollback. In that case, the context delta returned to CT can contain `role == "assistant"`. This PR renames the API from `non_assistant` to `context` and supports assistant messages inside context increments.

Main changes:

- Rename `merge_non_assistant_tokens` to `merge_context_tokens`.
- Rename `tokenize_non_assistant_incremental_messages` to `tokenize_context_incremental_messages`.
- Support `role == "assistant"` in context incremental messages.
- Keep `merge_assistant_tokens` for LLM-generated assistant tokens only.
- Context tokens always get loss mask `0` and logprob `0`, even when the context message role is `assistant`.
- Reject context incremental messages that end with an assistant message.
- Encode context increments by role-based groups instead of encoding the whole incremental message list at once.
- Update agent loop call sites from `non_assistant` to `context`.
- Update DeepSeekV4 CT builder from the old non-assistant API to the new context API.

## Context grouping behavior

CT now splits incremental context messages by role and delta-encodes each group separately.

Example:

```python
raw_messages = [system, user, assistant]
updated_messages = [system, user, assistant_c, tool_response, user]

keep_messages = [system, user]
incremental_messages = [assistant_c, tool_response, user]
```

CT groups the incremental messages as:

```python
[assistant_c], [tool_response], [user]
```

Then each group is delta-encoded and concatenated.

CT does not directly delta-encode the whole list:

```python
[assistant_c, tool_response, user]
```

## Why grouping is needed

There are two main reasons.

1. Some chat templates drop historical assistant thinking when a later user message exists. For example, Qwen3-style templates may remove assistant thinking from a previous assistant turn. CT intentionally preserves assistant thinking, so direct full incremental encoding is not always acceptable.

2. Different roles require different synthetic dummy contexts to get a stable suffix diff. For example, some MiniMax templates require a tool response to be preceded by an assistant tool-call message; otherwise the template cannot render the tool response correctly.

## Important behavior notes

1. CT rejects context incremental messages whose last message is `role == "assistant"`.

2. CT context messages are masked out for training:

   - loss mask: `0`
   - logprob: `0`

   This is true even if the context message role is `assistant`.

3. `merge_assistant_tokens` remains dedicated to LLM-generated assistant tokens. Those tokens keep loss mask `1`.

## Test matrix

This compares CT grouped context encoding against directly encoding the whole incremental context message list.

Test scope:

- This matrix focuses on `tokenize_context_incremental_messages(...)`, i.e. grouped context encoding.
- It intentionally does not test runtime-prefix boundary merge behavior.

Test setup:

- `A+tool`: the assistant message contains `tool_calls`.
- `A+user`: the assistant message is a plain assistant message.
- Direct baseline uses `synthetic system + synthetic user + incremental`.
- If the incremental context starts with `tool`, the direct baseline also prepends a synthetic assistant with non-empty reasoning.
- Comparison checks whether the suffix of the direct render exactly matches CT grouped encoding.
- `T = exact match`, `F = mismatch`.

| Model | A(no think)+tool | A(think)+tool | A(no think)+tool+user | A(think)+tool+user | A(no think)+user | A(think)+user | tool*2 | tool+user |
|---|---|---|---|---|---|---|---|---|
| Qwen2 | T | T | T | T | T | T | T | T |
| Qwen2.5 | T | T | T | T | T | T | T | T |
| Qwen3 | F | T | F | F | F | F | T | T |
| Qwen3.5 | T | T | F | F | F | F | T | T |
| MiMo | T | T | T | T | T | T | T | T |
| MiniMax M2 | T | T | T | F | T | F | T | T |
| MiniMax M2.5 | T | T | T | F | T | F | T | T |
| MiniMax M2.7 | T | T | T | F | T | F | T | T |
| GLM4.7 | T | T | T | F | T | F | T | T |
| GLM5 | T | T | T | F | T | F | T | T |
| Gemma4 | T | T | T | F | T | F | T | T |
| GPT-OSS | T | T | T | T | T | T | T | T |
| DeepSeek | T | T | T | T | T | T | T | T |
| Kimi | T | T | T | T | T | T | T | T |

## False cases

| Model | Case | Reason |
|---|---|---|
| Qwen3 | A(no think)+tool | CT grouped encoding treats the assistant as the last message in its synthetic context, so the template inserts an empty `<think></think>` block. Direct encoding has a following tool message, so it does not insert the empty think block. |
| Qwen3 | A(no think)+tool+user | Same empty think block behavior as above. |
| Qwen3 | A(think)+tool+user | Direct encoding drops or ignores historical assistant thinking because a later user message exists. CT grouped encoding preserves `reasoning_content`. |
| Qwen3 | A(no think)+user | CT grouped encoding inserts an empty `<think></think>` block for the plain assistant. Direct encoding has a later user message and does not insert it. |
| Qwen3 | A(think)+user | Direct encoding drops historical assistant thinking. CT grouped encoding preserves it. |
| Qwen3.5 | A(no think)+tool+user | CT grouped encoding includes an empty `<think></think>` block for the no-thinking assistant. Direct encoding does not. |
| Qwen3.5 | A(think)+tool+user | Direct encoding drops historical assistant thinking. CT grouped encoding preserves it. |
| Qwen3.5 | A(no think)+user | CT grouped encoding includes an empty `<think></think>` block. Direct encoding does not. |
| Qwen3.5 | A(think)+user | Direct encoding drops historical assistant thinking. CT grouped encoding preserves it. |
| MiniMax M2 | A(think)+tool+user | Direct encoding drops the historical assistant `<think>reasoning</think>` when a later user message exists. CT grouped encoding preserves it. |
| MiniMax M2 | A(think)+user | Same thinking-drop behavior. |
| MiniMax M2.5 | A(think)+tool+user | Same thinking-drop behavior. |
| MiniMax M2.5 | A(think)+user | Same thinking-drop behavior. |
| MiniMax M2.7 | A(think)+tool+user | Same thinking-drop behavior. |
| MiniMax M2.7 | A(think)+user | Same thinking-drop behavior. |
| GLM4.7 | A(think)+tool+user | Direct encoding drops historical assistant thinking. CT grouped encoding preserves it. |
| GLM4.7 | A(think)+user | Same thinking-drop behavior. |
| GLM5 | A(think)+tool+user | Same thinking-drop behavior. |
| GLM5 | A(think)+user | Same thinking-drop behavior. |
| Gemma4 | A(think)+tool+user | Direct encoding drops the historical assistant `<|channel>thought ... <channel|>` block. CT grouped encoding preserves it. |
| Gemma4 | A(think)+user | Same thinking-drop behavior. |

Overall, the false cases fall into two expected categories:

1. Direct encoding drops assistant thinking, while CT grouped encoding preserves it. This is the intended CT behavior.
2. CT may include an empty think tag. This is currently considered acceptable; newer templates such as Qwen3.5 also preserve empty think tags, so this may be the direction templates are moving toward.


gxlvera:
TODO: gemma 4 needs some fix when context msg contains assistant msg

PR #7300, opened 2026-08-06
[algo] fix: carry running_return through observation spans in REINFORCE++ (#7278)

### LINKED ISSUE #7278: REINFORCE++ drops outcome rewards before multi-turn observation spans
## System Info

- verl main commit: `2781c1e405fa80871a838bfdf9495b412df19c35`
- Linux x86_64, kernel 5.15
- Python 3.12.13
- PyTorch 2.13.0+cpu
- tensordict 0.10.0
- NumPy 2.5.1
- CUDA was unavailable; this reproducer exercises the device-independent advantage calculation on CPU.

## Information

- [x] The official REINFORCE++ advantage implementation
- [x] A minimal standalone reproduction

## Tasks

- [x] Multi-turn RL trajectories containing tool observations

## Reproduction

From a checkout of the commit above:

```bash
.venv/bin/python - <<'PY'
from types import SimpleNamespace
import torch
from verl.trainer.ppo.core_algos import compute_reinforce_plus_plus_outcome_advantage

config = SimpleNamespace(gamma=1.0)
compact_mask = torch.ones(1, 4)
compact_rewards = torch.tensor([[0.0, 0.0, 0.0, 1.0]])
expanded_mask = torch.tensor([[1.0, 1.0, 0.0, 0.0, 1.0, 1.0]])
expanded_rewards = torch.tensor([[0.0, 0.0, 0.0, 0.0, 0.0, 1.0]])

for run in range(1, 4):
    _, compact = compute_reinforce_plus_plus_outcome_advantage(
        compact_rewards, compact_mask, config=config
    )
    _, expanded = compute_reinforce_plus_plus_outcome_advantage(
        expanded_rewards, expanded_mask, config=config
    )
    print(
        f"run={run}",
        f"compact={compact.flatten().tolist()}",
        f"with_observation={expanded[expanded_mask.bool()].tolist()}",
    )
PY
```

This inserts two zero-masked observation tokens between the same four assistant tokens. With `gamma=1`, the valid-token returns should be invariant to that insertion. The baseline instead prints this result on all three runs:

```text
run=1 compact=[1.0, 1.0, 1.0, 1.0] with_observation=[0.0, 0.0, 1.0, 1.0]
run=2 compact=[1.0, 1.0, 1.0, 1.0] with_observation=[0.0, 0.0, 1.0, 1.0]
run=3 compact=[1.0, 1.0, 1.0, 1.0] with_observation=[0.0, 0.0, 1.0, 1.0]
```

`AgentLoopManager.generate_sequences` documents `response_mask=0` for both tool observations and trailing padding. In `compute_reinforce_plus_plus_outcome_advantage`, the reverse scan currently multiplies `running_return` by that mask at every position. An internal observation therefore resets the return as if the episode ended, preventing the final outcome reward from reaching earlier assistant turns.

The existing multi-turn GAE recurrence already treats observation tokens as skipped positions: they neither terminate the trajectory nor consume a discount step.

## Expected behavior

Masked tool-observation positions should be skipped while computing REINFORCE++ returns. Trailing padding should remain zero, and valid assistant-token returns should match the same action sequence without inserted observations. For the reproduction above, both valid-token return vectors should be `[1.0, 1.0, 1.0, 1.0]`.

shotsan:
## Summary

Fixes #7278 — REINFORCE++ drops outcome rewards assigned before multi-turn observation spans, producing incorrect advantage estimates in multi-turn RL training.

## Root Cause

In `compute_reinforce_plus_plus_outcome_advantage` (`verl/trainer/ppo/core_algos.py`, line ~718), the reverse return scan zeros out `running_return` at observation tokens:

```python
# Before (broken):
for t in reversed(range(token_level_rewards.shape[1])):
    running_return = token_level_rewards[:, t] + gamma * running_return
    returns[:, t] = running_return
    # Reset after EOS
    running_return = running_return * response_mask[:, t]  # <-- BUG
```

In multi-turn RL, **observation tokens** (tool responses, environment feedback) have `response_mask = 0`. The multiplication `running_return * response_mask[:, t]` zeros `running_return` at these positions, preventing rewards from propagating past observation spans to earlier action tokens.

### Example from the issue

Consider a sequence: `[action, action, observation, observation, action, action]` with a reward of 1.0 at the final token.

| Position | response_mask | Expected return | Actual return (buggy) |
|----------|--------------|-----------------|----------------------|
| action 1 | 1 | 1.0 | **0.0** |
| action 2 | 1 | 1.0 | **0.0** |
| obs 1    | 0 | 0.0 | 0.0 |
| obs 2    | 0 | 0.0 | 0.0 |
| action 3 | 1 | 1.0 | 1.0 |
| action 4 | 1 | 1.0 | 1.0 |

The reward signal is completely lost for actions before the observation span.

### Why GAE works correctly

The GAE implementation at lines 254-256 already handles this correctly using a carry-through pattern:

```python
# GAE (correct):
new_val = delta[:, t] + gamma * lam * running_gae
running_gae = new_val * response_mask[:, t] + running_gae * (1 - response_mask[:, t])
```

REINFORCE++ should have used the same pattern.

## Fix

Apply the GAE carry-through pattern: at observation tokens (`mask=0`), carry `running_return` through unchanged instead of zeroing it:

```python
# After (fixed):
for t in reversed(range(token_level_rewards.shape[1])):
    new_running_return = token_level_rewards[:, t] + gamma * running_return
    # Valid tokens (mask=1): update returns and running_return normally.
    # Observation tokens (mask=0): skip — carry running_return through
    # unchanged so rewards propagate past observation spans.
    returns[:, t] = new_running_return * response_mask[:, t]
    running_return = (new_running_return * response_mask[:, t]
                    + running_return * (1 - response_mask[:, t]))
```

This is a 4-line change that:
- **Preserves behavior for single-turn**: When `response_mask` is all 1s, the formula reduces to the original
- **Fixes multi-turn**: Rewards propagate correctly through observation spans
- **Matches GAE**: Uses the same pattern that already works at lines 254-256

### Verification with issue reproducer

The exact reproducer from the issue now outputs correct results:

| Scenario | Before | After |
|----------|--------|-------|
| Compact (no obs) | `[1.0, 1.0, 1.0, 1.0]` | `[1.0, 1.0, 1.0, 1.0]` |
| Expanded (with obs) | `[0.0, 0.0, 1.0, 1.0]` | `[1.0, 1.0, 1.0, 1.0]` |

### Files Changed

- `verl/trainer/ppo/core_algos.py` — 4-line fix in `compute_reinforce_plus_plus_outcome_advantage`

## Tests

6 new tests in `tests/trainer/ppo/test_reinforce_pp_multiturn_on_cpu.py`:

| Test | What it verifies |
|------|-----------------|
| `test_observation_span_does_not_block_returns` | Rewards propagate past observation spans to earlier action tokens |
| `test_observation_positions_have_zero_returns` | Observation tokens themselves still have zero returns |
| `test_trailing_padding_stays_zero` | Padding tokens (beyond sequence length) remain zero |
| `test_gamma_discount_across_observation_span` | Gamma discount is correctly applied across observation gaps |
| `test_batch_dimension` | Each batch element is processed independently |
| `test_no_observation_tokens_unchanged` | Single-turn (all mask=1) produces identical results to before |

**No regressions**: All 23 existing tests in `tests/trainer/ppo/test_core_algos_on_cpu.py` still pass.

```
tests/trainer/ppo/test_reinforce_pp_multiturn_on_cpu.py: 6 passed
tests/trainer/ppo/test_core_algos_on_cpu.py: 23 passed
```

### DIFF
--- a/tests/trainer/ppo/test_reinforce_pp_multiturn_on_cpu.py
+++ b/tests/trainer/ppo/test_reinforce_pp_multiturn_on_cpu.py
@@ -1,0 +1,147 @@
+# Copyright 2024 Bytedance Ltd. and/or its affiliates
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Tests for REINFORCE++ advantage computation with multi-turn observation spans.
+
+Verifies that masked observation tokens (response_mask=0) are correctly
+skipped during the reverse return scan, so that outcome rewards propagate
+past observation spans to earlier assistant turns.
+"""
+
+from types import SimpleNamespace
+
+import pytest
+import torch
+
+from verl.trainer.ppo.core_algos import compute_reinforce_plus_plus_outcome_advantage
+
+
+def _config(gamma=1.0):
+    return SimpleNamespace(gamma=gamma)
+
+
+class TestReinforcePPMultiTurn:
+    """Regression tests for #7278: observation spans must not block reward propagation."""
+
+    def test_observation_span_does_not_block_returns(self):
+        """Core regression test from the issue: inserting observation tokens
+        between assistant tokens should not change valid-token returns."""
+        config = _config(gamma=1.0)
+
+        # Compact: 4 assistant tokens, reward=1.0 at the end
+        compact_mask = torch.ones(1, 4)
+        compact_rewards = torch.tensor([[0.0, 0.0, 0.0, 1.0]])
+
+        # Expanded: same 4 assistant tokens with 2 observation tokens (mask=0) in the middle
+        expanded_mask = torch.tensor([[1.0, 1.0, 0.0, 0.0, 1.0, 1.0]])
+        expanded_rewards = torch.tensor([[0.0, 0.0, 0.0, 0.0, 0.0, 1.0]])
+
+        _, compact_returns = compute_reinforce_plus_plus_outcome_advantage(
+            compact_rewards, compact_mask, config=config
+        )
+        _, expanded_returns = compute_reinforce_plus_plus_outcome_advantage(
+            expanded_rewards, expanded_mask, config=config
+        )
+
+        # Valid-token returns should match
+        compact_valid = compact_returns[compact_mask.bool()]
+        expanded_valid = expanded_returns[expanded_mask.bool()]
+        torch.testing.assert_close(compact_valid, expanded_valid)
+
+    def test_observation_positions_have_zero_returns(self):
+        """Observation tokens (mask=0) should have returns=0."""
+        config = _config(gamma=1.0)
+        mask = torch.tensor([[1.0, 1.0, 0.0, 0.0, 1.0, 1.0]])
+        rewards = torch.tensor([[0.0, 0.0, 0.0, 0.0, 0.0, 1.0]])
+
+        _, returns = compute_reinforce_plus_plus_outcome_advantage(rewards, mask, config=config)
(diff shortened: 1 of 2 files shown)


shotsan:
@CLAassistant check

wuxibin89 (approved):

PR #7302, opened 2026-08-06
fix: normalize peft_config to string values across engines (#7290)

### LINKED ISSUE #7290: The two engines return differently shaped `peft_config` from get_per_tensor_param
# Issue 11 — the two engines return differently shaped `peft_config`

**Type:** bug report · **Finding #11** (new, 2026-08-06) · verified against `main @ 87c68d2`
**Why an issue and not part of #7287:** the fix is one line, but which line depends on
which side you consider authoritative, and I have no megatron setup to test the result
on. #7287 therefore rejects the megatron shape rather than guessing at it.

---

### Summary

`BaseEngine.get_per_tensor_param` declares its second return value `Optional[dict]`
(`verl/workers/engine/base.py:151`). Two engines ever return a non-`None` one, and they
do not agree on its contents:

| producer | `task_type` | `peft_type` |
| --- | --- | --- |
| `engine/fsdp/transformer_impl.py:1032` → `LoraConfig.to_dict()` | `TaskType` enum | `PeftType` enum |
| `engine/megatron/transformer_impl.py:837` → `build_peft_config_for_vllm()` | `TaskType` enum | **absent** |

Consumers therefore cannot treat the value uniformly. The vLLM rollout passes it straight
into `TensorLoRARequest` and does not read those keys, so it has not noticed. The SGLang
rollout does read them.

### How to see it

No GPU, no distributed setup:

```python
>>> from verl.utils.megatron_peft_utils import build_peft_config_for_vllm
>>> sorted(build_peft_config_for_vllm({"rank": 32, "alpha": 32}))
['bias', 'exclude_modules', 'lora_alpha', 'lora_dropout', 'r', 'target_modules', 'task_type']
>>> # no 'peft_type'

>>> from peft import LoraConfig, TaskType
>>> cfg = LoraConfig(r=32, target_modules=["q_proj"], task_type=TaskType.CAUSAL_LM).to_dict()
>>> type(cfg["task_type"]), type(cfg["peft_type"])
(<enum 'TaskType'>, <enum 'PeftType'>)
>>> # present, but as enums, so not JSON-serialisable as-is
```

### Why it matters now

SGLang's adapter loader needs `peft_type`. Today the megatron+SGLang+LoRA combination
never gets far enough to notice, because `SGLangRollout.wrap_lora_params` calls
`asdict()` on a dict and dies first — see #7287, which fixes that. Once
that crash is gone, this shape difference becomes the next thing in the way, and
`lora_served_as_adapter` explicitly supports the megatron LoRA config block
(`model.lora.rank`), so the pairing is meant to work.

I did not fold a fix into that PR because I cannot test megatron, and the choice is not
mine to make. The PR's helper rejects a config with no `peft_type` and names the
producer, so the failure is at least legible.

### Environment

`verl 0.8.0` and `main @ 87c68d2` (read), `peft 0.20.0`, Python 3.12. The shape
difference above was executed, not inferred; the megatron *training* path was not run.

### Related

**#5488** (`verl merger emits PEFT adapter config with task_type=null`, closed as
completed 2026-03-05) is the same family of bug on a different pair — the merger's
adapter-config writer rather than the engine-to-rollout handoff. Worth glancing at for
how that one was resolved.

### Which line to change is yours to pick

1. **Make the megatron producer complete** — add `"peft_type": PeftType.LORA` to
   `build_peft_config_for_vllm()`. Smallest, and arguably where the bug is, since the
   function's docstring says it builds a PEFT-compatible config. Risk: whatever reads it
   today may not expect the extra key.
2. **Make the base contract explicit** — say in `BaseEngine.get_per_tensor_param` what
   keys a returned dict must carry, and make both producers satisfy it. Most durable,
   most work.
3. **Normalize in each consumer** — what #7287 does for SGLang, duplicated wherever else
   it is needed. Least good; the disagreement stays.

Worth noting the enums are a separate half of the same problem: both producers leave
`task_type` (and FSDP also `peft_type`) as enum members, so any consumer that serializes
the dict has to unwrap them. If the contract gets written down, that belongs in it.

### AI assistance disclosure

AI assistance (Claude) was used to trace the producers and consumers and to draft this
report. The shape comparison above was run.

shotsan:
## Summary

Fixes #7290 — the two engines return differently shaped `peft_config` dicts from `get_per_tensor_param`, causing type mismatches and defensive code throughout the codebase.

## Root Cause

`BaseEngine.get_per_tensor_param` (`verl/workers/engine/base.py:151`) declares its second return value as `Optional[dict]`, but the two implementations disagree on the dict contents:

| Engine | `task_type` | `peft_type` |
|--------|-------------|-------------|
| FSDP (`transformer_impl.py:1032`) via `LoraConfig.to_dict()` | `TaskType` **enum** | `PeftType` **enum** |
| Megatron (`megatron_peft_utils.py:148`) via `build_peft_config_for_vllm()` | `TaskType` **enum** | **missing** |

### Downstream Impact

Consumers handle this inconsistently:
- **`wrap_lora_params`** (`sglang_rollout.py:436`) calls `asdict(peft_config)` (expects a dataclass, not a dict) and then `.value` on enums — fails on plain dicts from either engine
- **`model_merger`** (`base_model_merger.py:382`) has defensive `hasattr(val, "value")` checks for both enum and string formats
- **`fsdp_checkpoint_manager`** (`fsdp_checkpoint_manager.py:105`) similarly defends against both formats

### Producers — normalize at the source

**FSDP** (`verl/workers/engine/fsdp/transformer_impl.py:1032`):
After `.to_dict()`, stringify any enum values:
```python
if peft_config is not None:
    peft_config_dict = peft_config.to_dict()
    for key in ("task_type", "peft_type"):
        val = peft_config_dict.get(key)
        if hasattr(val, "value"):
            peft_config_dict[key] = val.value
```

**Megatron** (`verl/utils/megatron_peft_utils.py:148`):
Use string value and add missing `peft_type`:
```python
return {
    "task_type": TaskType.CAUSAL_LM.value,  # "CAUSAL_LM" (was enum)
    "peft_type": "LORA",                     # was missing entirely
    "r": lora_config.get("rank", 0),
    ...
}
```

### Consumer — make robust

**`wrap_lora_params`** (`verl/workers/rollout/sglang_rollout/sglang_rollout.py:436`):
Accept both dicts and dataclass objects, and defensively stringify any remaining enums:
```python
def wrap_lora_params(self, peft_config, weights):
    if isinstance(peft_config, dict):
        peft_config_json = dict(peft_config)
    else:
        peft_config_json = asdict(peft_config)
    for key in ("task_type", "peft_type"):
        val = peft_config_json.get(key)
        if hasattr(val, "value"):
            peft_config_json[key] = val.value
```

### Files Changed

- `verl/workers/engine/fsdp/transformer_impl.py` — stringify enums after `.to_dict()`
- `verl/utils/megatron_peft_utils.py` — use `.value`, add `peft_type`
- `verl/workers/rollout/sglang_rollout/sglang_rollout.py` — accept dict input

## Tests

5 tests in `tests/workers/engine/test_peft_config_shape.py`:

| Test | What it verifies |
|------|-----------------|
| `test_fsdp_style_enums_are_stringified` | Enum values from `LoraConfig.to_dict()` become strings |
| `test_megatron_style_already_strings` | String values pass through normalization unchanged |
| `test_megatron_build_peft_config_returns_strings` | `build_peft_config_for_vllm` returns string `task_type` and includes `peft_type` |
| `test_wrap_lora_params_accepts_dict` | Source code has `isinstance(peft_config, dict)` check |
| `test_both_engines_produce_same_shape` | After normalization, FSDP and Megatron dicts have identical key types |

```
tests/workers/engine/test_peft_config_shape.py: 5 passed
```

### DIFF
--- a/tests/workers/engine/test_peft_config_shape.py
+++ b/tests/workers/engine/test_peft_config_shape.py
@@ -1,0 +1,101 @@
+"""Tests for consistent peft_config shape from get_per_tensor_param (#7290).
+
+Both FSDP and Megatron engines must return peft_config dicts with string
+values for task_type and peft_type (not enum objects), and the consumer
+wrap_lora_params must accept plain dicts.
+"""
+
+from enum import Enum
+
+
+class _FakeTaskType(Enum):
+    CAUSAL_LM = "CAUSAL_LM"
+
+
+class _FakePeftType(Enum):
+    LORA = "LORA"
+
+
+def _stringify_peft_config(d):
+    """Mirrors the normalization applied by FSDP engine after .to_dict()."""
+    result = dict(d)
+    for key in ("task_type", "peft_type"):
+        val = result.get(key)
+        if hasattr(val, "value"):
+            result[key] = val.value
+    return result
+
+
+def test_fsdp_style_enums_are_stringified():
+    """FSDP's LoraConfig.to_dict() returns enum values; they must be stringified."""
+    raw = {
+        "task_type": _FakeTaskType.CAUSAL_LM,
+        "peft_type": _FakePeftType.LORA,
+        "r": 16,
+        "target_modules": ["q_proj"],
+    }
+    result = _stringify_peft_config(raw)
+    assert result["task_type"] == "CAUSAL_LM"
+    assert result["peft_type"] == "LORA"
+    assert isinstance(result["task_type"], str)
+    assert isinstance(result["peft_type"], str)
+
+
+def test_megatron_style_already_strings():
+    """Megatron returns strings directly; normalization is a no-op."""
+    raw = {
+        "task_type": "CAUSAL_LM",
+        "peft_type": "LORA",
+        "r": 16,
+        "target_modules": ["q_proj"],
+    }
+    result = _stringify_peft_config(raw)
+    assert result["task_type"] == "CAUSAL_LM"
+    assert result["peft_type"] == "LORA"
+
+
+def test_megatron_build_peft_config_returns_strings():
+    """build_peft_config_for_vllm must return string task_type and include peft_type."""
+    from verl.utils.megatron_peft_utils import build_peft_config_for_vllm
+
+    config = build_peft_config_for_vllm({
+        "rank": 16,
+        "alpha": 32,
+        "target_modules": ["linear_qkv"],
+    })
+    assert isinstance(config["task_type"], str), f"task_type is {type(config['task_type'])}"
+    assert config["task_type"] == "CAUSAL_LM"
+    assert "peft_type" in config, "peft_type key missing from Megatron peft_config"
+    assert config["peft_type"] == "LORA"
+
+
+def test_wrap_lora_params_accepts_dict():
+    """wrap_lora_params must accept a plain dict (not only a LoraConfig dataclass)."""
+    import pathlib
+
+    src = pathlib.Path("verl/workers/rollout/sglang_rollout/sglang_rollout.py").read_text()
+    assert "isinstance(peft_config, dict)" in src, (
+        "wrap_lora_params should check isinstance(peft_config, dict)"
+    )
+
+
+def test_both_engines_produce_same_shape():
+    """After normalization, FSDP and Megatron dicts must have matching key types."""
+    fsdp_raw = {
+        "task_type": _FakeTaskType.CAUSAL_LM,
+        "peft_type": _FakePeftType.LORA,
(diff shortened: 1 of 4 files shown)


shotsan:
@CLAassistant check

wuxibin89:
Duplicated with https://github.com/verl-project/verl/pull/7287

PR #7304, opened 2026-08-06
fix: set sleep_level before first resume on LoRA adapter path (#7289)

### LINKED ISSUE #7289: LoRA adapter path resumes the `weights` tag that was never released
# Issue 8 — LoRA adapter path resumes `weights` that were never released

**Type:** bug report · **Finding #8** · verified against `main @ 87c68d2` (2026-08-06)
**Why an issue and not a PR:** this is a small state-machine ordering question with
several defensible answers, and at least one of them risks the non-LoRA colocated path.
Our own fix was applied on the *symptom* side (in SGLang) and is deliberately **not**
being offered as a verl patch — see the last section.

---

### Summary

On the LoRA-as-adapter path, `engine_workers.py` resumes the `weights` tag before
anything has released it, and before the flag that would have prevented the release
is even set. SGLang's `offload_tags.remove()` is a strict set operation, so it raises,
`resume_memory_occupation` burns all three retries, and the Ray worker dies.

```
KeyError: 'weights'
```

### The ordering, on `main`

`verl/workers/engine_workers.py`:

```python
764        # 1. resume rollout memory (weights were released during sleep)
765        if self.config.rollout.free_cache_engine:
766            await self.rollout.resume(tags=["weights"])       # <-- unconditional
...
770        per_tensor_param, peft_config = self.actor.engine.get_per_tensor_param(...)
774        do_lora_base_sync = False
775        if not self.peft_merge and peft_config is not None:
776            self.rollout.sleep_level = 1                      # <-- ten lines too late
```

The comment on line 764 states the assumption that breaks: *"weights were released
during sleep"*. For `sleep_level = 1` they were not. `SGLangRollout.release`
(`sglang_rollout.py:285`) releases only `kv_cache` when `sleep_level == 1`, and
`sleep_level` cannot be 1 on the first sync because line 776 has not run yet.

### A second, related asymmetry

`verl/workers/rollout/sglang_rollout/async_sglang_server.py`:

- `sleep()` (line 484) branches on `lora_as_adapter` and releases `["kv_cache"]` only.
- `wake_up()` (line 461) does **not** branch; its `COLOCATED` arm hard-codes
  `ResumeMemoryOccupationReqInput(tags=["kv_cache", "weights"])`.

So the resume side of that pair does not mirror the release side either. We mirrored it
locally, but this was **not** the root cause of our crash — the `engine_workers.py`
ordering above was. Flagging it as a likely-same-bug rather than claiming we proved it.

### Environment

`verl 0.8.0`, `sglang 0.5.8`, `torch 2.9.1+cu128`, `peft 0.20.0`, Python 3.12,
1× A800-80G, colocated rollout, FSDP + LoRA rank 32, `free_cache_engine=true`.
Note `setup.py`'s `SGLANG_REQUIRES` pins exactly this sglang/torch pair.

### Symptom

First weight sync of the run. `resume_memory_occupation` fails three times, then:

```
KeyError: 'weights'
  ... sglang/srt/managers/scheduler_update_weights_mixin.py, in resume_memory_occupation
      self.offload_tags.remove(tag)
```

### Questions we cannot answer for you

1. Should `resume` mirror exactly what `sleep` released, rather than assuming a fixed
   tag set?
2. Should `sleep_level` be determined before the first `resume` — e.g. at engine
   construction from `lora_served_as_adapter(model_config)` — instead of being
   discovered mid-sync from `peft_config`?
3. Does either change disturb the non-LoRA colocated path, which currently relies on
   `resume(["weights"])` always being valid?

(2) looks cleanest from outside: `lora_served_as_adapter` is already a pure function of
`model_config` and is already used to decide `enable_lora` at launch, so the
information is available well before line 766. But we have not traced every caller.

### On our own patch — please do not port it

We got past this by making SGLang's resume tolerant of unaccounted tags:

```python
# local patch: tolerate resume of never-released tags (verl 0.8 resumes
# 'weights' unconditionally on first weight sync, before any release)
tags = [t for t in tags if t in self.offload_tags]
for tag in tags:
    self.offload_tags.remove(tag)
```

That is a symptom-side fix in the wrong repository — it makes SGLang silently accept an
incoherent request instead of making verl stop sending one. We are recording it only so
that the failure is reproducible and so nobody mistakes it for a proposal.

### AI assistance disclosure

AI assistance (Claude) was used to re-verify the ordering and the `wake_up`/`sleep`
asymmetry against `main`, and to draft this report.

shotsan:
## Summary

Fixes #7289 — LoRA adapter path resumes `weights` tag that was never released, causing `KeyError: 'weights'` in SGLang's `resume_memory_occupation` on the first weight sync.

### Bug 1: Resume before sleep_level is known (`engine_workers.py`)

In `ActorRolloutRefWorker.update_weights()`:

```python
# Line 764-766: resume called UNCONDITIONALLY
if self.config.rollout.free_cache_engine:
    await self.rollout.resume(tags=["weights"])   # <-- always resumes weights

# Line 774-776: sleep_level set TEN LINES LATER
if not self.peft_merge and peft_config is not None:
    self.rollout.sleep_level = 1                  # <-- too late
```

The comment on line 764 states *"weights were released during sleep"* — but for `sleep_level=1` (LoRA adapter path), only `kv_cache` was released. Weights were never in `offload_tags`, so SGLang raises `KeyError("weights")`.

### Bug 2: sleep/wake_up asymmetry (`async_sglang_server.py`)

```python
# sleep() BRANCHES on lora_as_adapter:
if self.lora_as_adapter:
    tags = ["kv_cache"]           # only kv_cache for adapter path
else:
    tags = ["kv_cache", "weights"]

# wake_up() does NOT branch — hard-codes both tags:
obj = ResumeMemoryOccupationReqInput(tags=["kv_cache", "weights"])  # always!
```

So `wake_up()` tries to resume `weights` that `sleep()` never released.

### Symptom

First weight sync of a LoRA run with `free_cache_engine=true`:
```
KeyError: 'weights'
  sglang/srt/managers/scheduler_update_weights_mixin.py, in resume_memory_occupation
      self.offload_tags.remove(tag)
```

### 1. Set `sleep_level` at init time (`engine_workers.py`)

Determine LoRA adapter mode during `__init__`, where `model_config` is already available:

```python
lora_rank = model_config.lora.get("rank", 0) or getattr(model_config, "lora_rank", 0)
lora_as_adapter = lora_rank > 0 and not self.peft_merge
if hasattr(self, "rollout") and lora_as_adapter:
    self.rollout.sleep_level = 1
```

This uses the same `lora_served_as_adapter` logic from `sglang_rollout/utils.py` but avoids adding a dependency.

### 2. Guard resume on `sleep_level` (`engine_workers.py`)

```python
if self.config.rollout.free_cache_engine:
    if getattr(self.rollout, "sleep_level", 2) == 1:
        pass  # weights were never released; nothing to resume
    else:
        await self.rollout.resume(tags=["weights"])
```

### 3. Mirror `sleep()` in `wake_up()` (`async_sglang_server.py`)

```python
elif self.rollout_mode == RolloutMode.COLOCATED:
    if self.lora_as_adapter:
        tags = ["kv_cache"]
    else:
        tags = ["kv_cache", "weights"]
    obj = ResumeMemoryOccupationReqInput(tags=tags)
```

### Files Changed

- `verl/workers/engine_workers.py` — early sleep_level, guarded resume
- `verl/workers/rollout/sglang_rollout/async_sglang_server.py` — wake_up mirrors sleep

## Tests

5 tests in `tests/workers/test_lora_resume_ordering.py`:

| Test | What it verifies |
|------|-----------------|
| `test_sleep_level_set_at_init` | `sleep_level=1` is set near `peft_merge` in `__init__`, before `update_weights` runs |
| `test_resume_skips_weights_at_sleep_level_1` | The resume section checks `sleep_level` before resuming weights |
| `test_wake_up_mirrors_sleep` | `wake_up` COLOCATED path branches on `lora_as_adapter` |
| `test_sleep_and_wake_up_tag_symmetry` | Both `sleep()` and `wake_up()` reference `lora_as_adapter` |
| `test_no_unconditional_weights_resume` | No `resume(tags=["weights"])` without a `sleep_level` guard |

```
tests/workers/test_lora_resume_ordering.py: 5 passed
```

### DIFF
--- a/tests/workers/test_lora_resume_ordering.py
+++ b/tests/workers/test_lora_resume_ordering.py
@@ -1,0 +1,90 @@
+"""Tests for LoRA adapter resume ordering fix (#7289).
+
+Verifies that:
+1. sleep_level is set at init time (before first resume)
+2. update_weights skips resuming "weights" when sleep_level=1
+3. wake_up mirrors sleep in the COLOCATED path
+"""
+
+import pathlib
+
+
+def _read_source(relpath):
+    return pathlib.Path(relpath).read_text()
+
+
+def test_sleep_level_set_at_init():
+    """sleep_level must be set during __init__, before update_weights runs."""
+    src = _read_source("verl/workers/engine_workers.py")
+    assert "lora_as_adapter" in src
+    assert "self.rollout.sleep_level = 1" in src
+    peft_merge_pos = src.index("self.peft_merge")
+    first_sleep_level_pos = src.index("self.rollout.sleep_level = 1")
+    assert first_sleep_level_pos > peft_merge_pos
+    assert first_sleep_level_pos - peft_merge_pos < 600
+
+
+def test_resume_skips_weights_at_sleep_level_1():
+    """update_weights must not resume 'weights' when sleep_level=1."""
+    src = _read_source("verl/workers/engine_workers.py")
+    resume_start = src.index("# 1. resume rollout memory")
+    resume_end = src.index("# 2. determine")
+    resume_section = src[resume_start:resume_end]
+    assert "sleep_level" in resume_section
+
+
+def test_wake_up_mirrors_sleep():
+    """wake_up COLOCATED must branch on lora_as_adapter like sleep does."""
+    src = _read_source("verl/workers/rollout/sglang_rollout/async_sglang_server.py")
+    wake_up_start = src.index("async def wake_up")
+    # Find next method or property
+    rest = src[wake_up_start + 20:]
+    for marker in ["\n    @", "\n    async def ", "\nclass "]:
+        try:
+            end = rest.index(marker)
+            break
+        except ValueError:
+            end = len(rest)
+    wake_up_src = rest[:end]
+    assert "lora_as_adapter" in wake_up_src
+
+
+def test_sleep_and_wake_up_tag_symmetry():
+    """sleep and wake_up must both reference lora_as_adapter."""
+    src = _read_source("verl/workers/rollout/sglang_rollout/async_sglang_server.py")
+
+    sleep_start = src.index("async def sleep")
+    rest = src[sleep_start + 20:]
+    for marker in ["\n    async def ", "\n    @", "\nclass "]:
+        try:
+            end = rest.index(marker)
+            break
+        except ValueError:
+            end = len(rest)
+    sleep_src = rest[:end]
+
+    wake_up_start = src.index("async def wake_up")
+    rest2 = src[wake_up_start + 20:]
+    for marker in ["\n    @", "\n    async def ", "\nclass "]:
+        try:
+            end2 = rest2.index(marker)
+            break
+        except ValueError:
+            end2 = len(rest2)
+    wake_up_src = rest2[:end2]
+
+    assert "lora_as_adapter" in sleep_src, "sleep must check lora_as_adapter"
+    assert "lora_as_adapter" in wake_up_src, "wake_up must check lora_as_adapter"
+
+
+def test_no_unconditional_weights_resume():
+    """engine_workers must not unconditionally resume weights tag."""
+    src = _read_source("verl/workers/engine_workers.py")
+    resume_start = src.index("# 1. resume rollout memory")
+    resume_end = src.index("# 2. determine")
+    resume_section = src[resume_start:resume_end]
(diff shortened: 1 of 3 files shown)


shotsan:
@CLAassistant check

PR #7315, opened 2026-08-07
fix: del_local_ckpt_after_load now removes local checkpoint directories (#7213)

### LINKED ISSUE #7213: [Bug] del_local_ckpt_after_load does not remove loaded checkpoints
### System Info

- verl: current `main` (`aebd1f8a`)
- Affected code: FSDP and Megatron checkpoint managers
- Storage topology: the shared local checkpoint path supported by trainer resume
- This is filesystem- and accelerator-independent.

### Information

- [x] The official trainer checkpoint flow

### Tasks

- [x] An officially supported trainer configuration

### Reproduction

`trainer.del_local_ckpt_after_load` was introduced by #274 and is documented as deleting local checkpoints after loading. The trainer passes the flag to actor and critic checkpoint loads.

On current `main`, enable:

```yaml
trainer:
  resume_mode: auto
  del_local_ckpt_after_load: true
```

Resume from a local checkpoint and inspect `global_step_<N>/actor` after loading. The directory remains.

The cleanup conditions cannot remove the checkpoint:

- FSDP calls `copy_to_local()` and then invokes `os.remove()` only when the returned path is non-local. `copy_to_local()` returns a local path, so the condition is false.
- Megatron receives a local role directory from the trainer, checks whether it is non-local, and therefore skips cleanup. `os.remove()` would also be unsuitable for the directory.

No open issue or PR was found for this behavior.

### Expected behavior

When the option is enabled on the trainer's shared checkpoint filesystem, all ranks should finish loading before rank 0 removes the loaded actor or critic role directory. The containing `global_step_<N>` directory and `data.pt` should remain available to the driver.

I can submit a focused fix with CPU distributed regression coverage.

shotsan:
## Summary

Fixes #7213 — `del_local_ckpt_after_load` does not remove loaded checkpoints from local disk.

Supersedes #7299 (closed due to unsigned CLA, now signed).

## Root Cause

The cleanup code in both FSDP and Megatron checkpoint managers had the same two bugs:

### Bug 1: `is_non_local()` gate blocks all local paths

```python
# Old code (broken):
os.remove(path) if is_non_local(path) else None
```

`is_non_local()` (in `verl/utils/fs.py`) only matches `hdfs://` prefixes. For local paths like `/data/checkpoints/step_100/`, it always returns `False`, so the cleanup never executes.

### Bug 2: `os.remove()` cannot delete directories

Checkpoints are directories containing multiple files (`model.safetensors`, `config.json`, etc.). `os.remove()` raises `IsADirectoryError` on directories — `shutil.rmtree()` is needed instead.

## Fix

**3 sites** across 2 files:

### `verl/utils/checkpoint/fsdp_checkpoint_manager.py` (1 site)

```python
# After fix:
torch.distributed.barrier()  # wait for all ranks to finish loading

if self.rank == 0 and del_local_after_load:
    try:
        if os.path.isdir(local_path):
            shutil.rmtree(local_path)
            log_with_rank(...)
    except Exception as e:
        log_with_rank(...)
```

Key changes:
- Removed `is_non_local` guard — cleanup now works for local paths
- Replaced `os.remove` with `shutil.rmtree` + `os.path.isdir` check
- Moved `barrier()` **before** cleanup (was after), so all ranks finish loading before rank 0 deletes
- Wrapped in try/except to avoid crashing training on cleanup failure

### `verl/utils/checkpoint/megatron_checkpoint_manager.py` (2 sites)

Same fix applied to both `_load_megatron_fsdp_checkpoint` and `load_checkpoint`:
- Removed `is_non_local` guard
- Replaced `os.remove` with `shutil.rmtree`
- Added `self.rank == 0` guard (was missing in Megatron — all ranks were attempting deletion)

### Files Changed

- `verl/utils/checkpoint/fsdp_checkpoint_manager.py` — added `import shutil`, fixed cleanup block
- `verl/utils/checkpoint/megatron_checkpoint_manager.py` — added `import shutil`, fixed 2 cleanup blocks

## Tests

5 tests in `tests/utils/ckpt/test_del_local_ckpt_after_load.py`:

| Test | What it verifies |
|------|-----------------|
| `test_is_non_local_rejects_local_paths` | Confirms `is_non_local()` returns False for local paths (the root cause) |
| `test_old_cleanup_logic_fails` | The old `os.remove if is_non_local` pattern does nothing on local paths |
| `test_new_cleanup_logic_removes_directory` | `shutil.rmtree` successfully removes checkpoint directories |
| `test_cleanup_handles_already_removed` | No crash if directory was already deleted (race condition safety) |
| `test_cleanup_removes_nested_files` | Correctly removes directories containing nested files |

```
tests/utils/ckpt/test_del_local_ckpt_after_load.py: 5 passed
```

### DIFF
--- a/tests/utils/ckpt/test_del_local_ckpt_after_load.py
+++ b/tests/utils/ckpt/test_del_local_ckpt_after_load.py
@@ -1,0 +1,110 @@
+# Copyright 2024 Bytedance Ltd. and/or its affiliates
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Tests for del_local_ckpt_after_load checkpoint cleanup.
+
+Verifies that the cleanup logic in FSDPCheckpointManager and
+MegatronCheckpointManager correctly removes local checkpoint
+directories after loading, regardless of whether the path is
+local or remote (HDFS).
+"""
+
+import os
+import shutil
+import tempfile
+
+import pytest
+
+from verl.utils.fs import is_non_local
+
+
+class TestDelLocalCkptCleanupLogic:
+    """Test the underlying cleanup logic that was broken in #7213.
+
+    The bug: cleanup used `os.remove(path) if is_non_local(path) else None`
+    which never fires for local paths (is_non_local only matches hdfs://).
+    Also, os.remove cannot delete directories.
+
+    The fix: use `shutil.rmtree(path)` without the is_non_local guard.
+    """
+
+    @pytest.fixture(autouse=True)
+    def setup(self):
+        self.test_dir = tempfile.mkdtemp()
+        yield
+        shutil.rmtree(self.test_dir, ignore_errors=True)
+
+    def _create_ckpt_dir(self, name="actor"):
+        """Create a mock checkpoint directory with files inside."""
+        ckpt_path = os.path.join(self.test_dir, "global_step_100", name)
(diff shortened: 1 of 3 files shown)


wuxibin89:
We're not going to support `del_local_ckpt_after_load`, this config should be removed.

Issue #6719, opened 2026-06-13
Continuous Token Support for Multi-Turn AgentLoop Rollout


gxlvera:
### Feature request

# Continuous Token Support for Multi-Turn AgentLoop Rollout

This issue tracks a stacked PR series that adds Continuous Token support to verl's multi-turn AgentLoop rollout.

## Motivation

# Agentic Multi-turn Continuous Token Design Notes

## TL;DR

The basic idea of token-in-token-out (TITO) is that the training side should use the token ids that were actually produced and observed on the inference side, instead of retokenizing after the fact. verl already has the most basic token-in-token-out idea today: it preserves the accumulated token ids from previous turns and aligns the loss mask correctly. However, agentic multi-turn rollout involves concatenating tokens across multiple turns. The chat templates of many open-source models are not simply append-only; **if incremental token ids are extracted carelessly and boundaries are not merged rigorously, the resulting multi-turn prompt can be wrong.** If we only needed to handle a specific model tokenizer, we could ensure correctness with a model-specific method. But we want a design that generalizes across different tokenizers while minimizing model-specific hardcoding. A generalizable multi-turn TITO design has four common pitfalls:

1. Multi-turn continuous tokens must not retokenize the current full messages. They must preserve the real token ids accumulated from previous turns.
2. Incremental non-assistant token ids must be extracted from an appropriate synthetic context, not by simply encoding incremental messages.
3. When incremental ids are merged back into old token ids, some models require explicit boundary handling.
4. After constructing the full trajectory, a comparator is needed to check structural errors. Differences between assistant content and the canonical render of the full chat template can be tolerated, but deviations in special-token boundaries cannot. [WIP]

Among these, item 1 is the core TITO invariant and should be fixed at the trajectory/session layer. Items 2 and 3 are model-related implementation details that can be overridden through model-specific strategies. Item 4 is the final structural validation mechanism.

This note discusses the four points above, how they are currently implemented in verl, and how a `ContinuousTokenBuilder` abstraction can implement correct multi-turn TITO through a "**core invariant pipeline + model-specialized methods**" design.

### 1. Full messages must not be retokenized

In a multi-turn rollout, the previous turn observed by the model is:

```text
prompt_token_ids + model_output_token_ids
```

When continuing generation in the next turn, we should directly reuse this real sequence of token ids. If we instead run `full_messages` through `apply_chat_template + encode` again, two classes of problems can occur.

[...]

PR #7322, opened 2026-08-08
[data] feat: integrate ContinuousTokenBuilder into MultiTurnSFTDataset

yph22:
### What does this PR do?

Update: This PR now aligns `MultiTurnSFTDataset` with VERL’s Continuous Token semantics, adding incremental multi-turn token construction, correct loss masking, and model-specific protocol handling across registered CT model families.

---------------

`MultiTurnSFTDataset` currently renders every message in isolation and concatenates the results. That assumption does
not hold for neighbor-aware chat templates such as Qwen3.5/3.6:

- rendering an initial `system` message alone is invalid because the template requires a user query; and
- consecutive `tool` messages must share one outer user envelope, while isolated rendering creates one envelope per
  tool response.

This PR keeps only those neighbor-dependent messages together: the initial `system + user` pair and each consecutive
run of `tool` messages. All other messages still use the existing single-message hook, which is retained as a
backward-compatible wrapper for custom dataset subclasses. The strict comparison against whole-conversation rendering
remains enabled.

No duplicate was found in the open-PR search:
https://github.com/verl-project/verl/pulls?q=is%3Apr+is%3Aopen+MultiTurnSFTDataset+Qwen3.5

AI assistance was used (OpenAI Codex) to prepare and test this patch.

**Human-review confirmation:** I reviewed every changed line, understand the rendering and loss-mask behavior, and can
  defend the change end-to-end.

### Checklist Before Starting

- [x] Searched for similar PRs (query linked above).
- [x] PR title follows the required format: `[data] fix: preserve context in multi-turn SFT rendering for Qwen3.5/3.6`.

### Test

- `python -m pytest -q tests/utils/dataset/test_multiturn_sft_dataset_on_cpu.py::test_multiturn_sft_message_group_boundaries`:
  **1 passed**.
- `SKIP=autogen-trainer-cfg pre-commit run --all-files --show-diff-on-failure`: **passed**.
- `conda run -n verl-opd-veomni bash scripts/generate_trainer_config.sh`: **passed** (the separately executed generated-config hook).
- Strict whole-conversation token equality on the first two real agent trajectories:
  - Qwen3.5-9B: 8,736 and 8,370 tokens; **passed**.
  - Qwen3.6-35B-A3B: 8,736 and 8,370 tokens; **passed**.

### API and Usage Example

No public API or configuration change is required. Existing `MultiTurnSFTDataset` configurations automatically use the
correct message grouping.

### Design & Code Changes

- Add a small boundary helper for the two neighbor-dependent group types.
- Render multi-message groups together and keep single-message rendering on the existing override hook.
- Keep assistant messages in one-message groups so their generation-prompt loss mask is unchanged.
- Extend the existing tokenizer-backed test data with an initial system prompt and consecutive tool responses.
- Update the SFT configuration comment to describe message-group rendering.

- [x] Read the contribution guide and `AGENTS.md`.
- [x] Applied all pre-commit checks (the generated-config hook was run separately in the project environment).
- [x] Updated the relevant configuration documentation.
- [x] Added/updated unit and tokenizer-backed tests for the changed behavior.
- [x] Request CI in the `ci-request` Slack/Feishu channel after the PR is ready.
- [x] Recipe submodule update is not applicable.


wuxibin89:
@gxlvera Does continuous token address this problem?

yph22:
@wuxibin89  Thanks for bringing this up. However, it seems that Continuous Token currently operates in the text-only **multi-turn PPO/AgentLoop** path, while this PR fixes **offline MultiTurnSFTDataset tokenization and constructs assistant loss masks for SFT.**
MultiTurnSFTDataset does not currently use the Continuous Token builder. There is some overlap in grouping
consecutive tool messages, so I’m open to reusing a shared helper if that is the preferred direction.

yilunzhao:
Thanks for tackling this. I applied the PR to `1dda039b` and validated it on 10,726 multi-turn tool-use conversations with the official Qwen3.5-9B template. It applies cleanly and stops the crash on all of them, but reaches strict token equality with whole-conversation `apply_chat_template` on **9,943**. The remaining **783** render differently, and silently: with `ignore_input_ids_mismatch=true` that is +17.9% supervised tokens (54.8M vs a 46.5M reference).

**The gap is mechanical, not corpus-specific.** The template's think rule is `loop.index0 > ns.last_query_index`, which depends on a message that comes *after* the one being rendered, while `_message_group_end` only groups a message with *earlier* neighbours. Assistant turns therefore still render alone, the dummy-user prepend fires, `1 > 0` is always true, and `<think>` is always kept — including on turns that whole-conversation rendering renders without one.

Minimal reproducer — no system message, no tools:

```python
messages = [
    {"role": "user",      "content": "Run the task."},
    {"role": "assistant", "content": "<think>plan A</think>doing step one"},
    {"role": "user",      "content": "Format error: retry."},
    {"role": "assistant", "content": "<think>plan B</think>doing step two"},
]
```

Group boundaries under the patch are `[0:1]=['user'] [1:2]=['assistant'] [2:3]=['user'] [3:4]=['assistant']`; per-group concat is **49 tokens**, whole-conversation is **42**. Whole-conversation drops the first assistant's `<think>` block (because a later user query moves `last_query_index`); per-group keeps it. Any conversation with more than one real user query hits this.

Separately, `[system, <non-user>, ...]` falls back to a `[system]`-only group and still raises the original `System message must be at the beginning.` from #7378 — e.g. `[system, assistant, user, assistant]`.

For what it is worth, the grouping approach does work if assistant turns are also given their following context: rendering each turn with knowledge of whether a later user query exists reproduces whole-conversation output on all 10,726 rows in my corpus. Happy to share the harness or test cases if useful.

yph22:
@yilunzhao Thanks for the detailed validation and the minimal reproducer. Your diagnosis is correct. The previous patch only accounted for left-neighbor context, so it did not handle templates **whose rendering depends on a later real user query.**

I pushed a follow-up fix in `f6859ddf`. Assistant and system turns now detect whether a later real user query exists, render with a temporary following-user context, and then strictly verify and remove that temporary token suffix. The assistant loss mask is also aligned using the actual rendered prefix.

I added regression coverage for both cases you reported:

* the multi-user reproducer now has strict token equality with whole-conversation apply_chat_template (42 tokens), while preserving the intended assistant loss mask;
* [system, assistant, user, assistant] no longer raises the system-message error and also matches whole-conversation rendering exactly.

Additionally, for [Qwen3.8](https://huggingface.co/Qwen/Qwen3.8-27B/blob/main/chat_template.jinja), its default behaviour (`preserve_thinking=True`) renders and preserves all reasoning content, including reasoning from assistant turns that occur before non-final user turns.

If you could rerun the 10,726 multi-turn tool-use conversations against f6859ddf, that would be very helpful.

Thanks again for catching this and for providing such a clear analysis.

yilunzhao:
Re-ran the full validation against `f6859dd`. **It closes the gap: 10,726/10,726 rows now reach
strict token equality, and the loss masks match too.**

Setup, unchanged from the previous run: verl `1dda039b` with this PR applied to a scratch copy,
official `Qwen/Qwen3.5-9B` chat template (byte-identical to the model repo's `chat_template.jinja`,
sha256 `a4aee8af…c28f715`), 10,726 real multi-turn agent trajectories with tool calls, of which 783
inject a mid-episode user turn and therefore move `ns.last_query_index`. Acceptance bar is yours:
strict token equality between group-wise concatenation and whole-conversation
`apply_chat_template`, plus loss-mask comparison.

| | `059eb26` | **`f6859dd`** |
| --- | --- | --- |
| rows rendering without crash | 10,726 / 10,726 | 10,726 / 10,726 |
| **strict token equality** | 9,943 / 10,726 | **10,726 / 10,726** |
| **the 783 multi-user-query rows** | **0 / 783** | **783 / 783** |
| supervised tokens vs whole-conversation reference | +8,315,943 (+17.9%) | **0** |

We also ran it head to head against our own adaptation layer (a `custom_cls` subclass that pins a
per-turn context so the template's global `last_query_index` decision comes out right). On all
10,726 rows the two now produce **byte-identical token streams and byte-identical loss masks** —
supervised-token delta exactly 0. We are planning to delete our layer and use this PR directly.

Both new tests pass on our pin. `test_multiturn_sft_qwen35_following_user_context` is parametrized on
a local `~/models/Qwen/Qwen3.5-0.8B`, which we do not have, so we called the function directly with
our Qwen3.5-9B tokenizer; it passes there.

## The suffix strip holds up adversarially

We tried to break "strictly verify and remove that temporary token suffix" and could not. Six hostile
cases, all strict-token-equal:

* a single-`system` group **with `tools`** and a later user query — rendering `context_ids` with
  `tools=None` and subtracting `system_prompt` keeps the suffix exact even though `tools` rewrites
  the leading system block;
* assistant content deliberately **ending with a literal `<|im_start|>user\n<|im_end|>\n`** — the
  strip is positional and the `torch.equal` guard can only ever false-*fail*, never false-*pass*;
* a later user message that is a bare `<tool_response>…</tool_response>` envelope — correctly not
  treated as a query, so the earlier turns keep their think blocks;
* think-dropped turn with an empty body; empty-reasoning think-kept turns where `<think>\n` and the
  content's leading `\n` merge into one `\n\n` token (the merged token is masked whole, which is
  right); and a three-assistant / two-injected-user conversation.

Special tokens force hard boundaries at the join, so no tokenizer merge is possible there, and the
explicit equality check turns any surprise into a raise. This looks sound.

## One thing we did find — in the loss mask, not the tokens

`_common_prefix_length(input_ids, self.generation_prompt)` can run past the real header. On a
**think-dropped** assistant turn whose rendered body starts with a literal `<think>`, the common
prefix keeps matching into the content:

```python
messages = [
    {"role": "user",      "content": "q"},
    {"role": "assistant", "content": "<think>unclosed literal body"},   # no closing tag
    {"role": "user",      "content": "Format error: retry."},
    {"role": "assistant", "content": "done"},
]
```

The first assistant turn is think-dropped, so it renders as `<|im_start|>assistant\n<think>unclosed
literal body<|im_end|>\n` and the true header is 3 tokens. `_common_prefix_length` returns **4**
(and **5** if the body starts `<think>\n`), so 1–2 real content tokens are dropped from supervision.

The token stream is unaffected, which means **`sanity_check` cannot catch it** — it compares
`input_ids` only and never looks at `loss_mask`. It is always an over-mask, never an under-mask, so
nothing leaks into the loss; it just silently loses a token or two of supervision.

It does not affect our corpus (0 of 10,726 rows have such a turn), so all the numbers above stand
either way. A possible fix: when the think block is dropped, mask against the header the template
actually emitted (`<|im_start|>assistant\n`) rather than taking a common prefix with the think-ful
generation prompt — the two headers are known and distinguishable at that point.

Happy to re-run the corpus on any further commit.

yph22:
@yilunzhao Thanks again for the careful check. Your underlying observation about the loss-mask boundary is correct: using `_common_prefix_length(input_ids, generation_prompt)` can make the masked prefix depend on the actual response content, so a real response token that happens to match the generation prompt can be over-masked.

I initially implemented your suggested direction in `c9d7974`, treating the leading literal `<think>` in the reproducer as supervised content. I later reverted that change in `26d8521`, because after looking more closely I found **the particular unclosed-`<think>` example semantically ambiguous.**

For example, consider a truncated rollout in thinking mode. The opening `<think>` may originally have been inserted by the chat template as part of the generation prompt, with the model only generating `unclosed literal body`. If the serialized assistant message later becomes

`<think>unclosed literal body`

**then that `<think>` should remain masked**, not supervised. Once it has been serialized into `assistant.content`, however, that provenance is no longer recoverable from the message alone.

There is also a related rendering issue with such an incomplete record. If this assistant turn later becomes **a historical turn** because another real user query follows it, Qwen3.5 would normally **drop its reasoning from the history**. However, the template recognizes the reasoning span through the closing `</think>`. Without that closing tag, it cannot identify the span as reasoning, so the whole `<think>unclosed literal body` string is treated as ordinary content and remains in the rendered history. For this reason, **I think an unclosed serialized generation-prefix case is better treated as a malformed/incomplete dataset record that should be normalized upstream, rather than something from which `MultiTurnSFTDataset` should try to infer the original token provenance.**

That said, the underlying common-prefix collision you identified is real even for completely well-formed data, so I fixed that independently in `ed4ac3f` and the follow-up commits. Historical assistant mask boundaries are now inferred from **content-independent reasoning/response probes** rather than from the actual response content.

For example, if the model really intends to emit a literal `<think>` as its answer, an unambiguous serialized message can look like

`<think>I need to output the tag</think><think>`

in thinking mode, or

`<think></think><think>`

in non-thinking mode. In both cases (newlines ignored), the first reasoning block has well-defined template/reasoning semantics, while the final `<think>` is genuine response content. The current implementation correctly drops or preserves the historical reasoning according to the template while keeping that final literal `<think>` supervised.

`a29659c` additionally makes the generation-prompt mask respect each sample's `enable_thinking` setting, including the template-supplied closing `</think>` in non-thinking mode. **The latest `a906d6e` extends the content-independent header inference to templates that preserve historical reasoning as well, including Qwen3.6 with `preserve_thinking=True`, empty reasoning, and structured `reasoning_content`.**

So the final policy is: **template-produced prefix tokens are masked; actual assistant response tokens remain supervised even if they are token-identical to part of the generation prompt.** Incomplete/unclosed serialized template prefixes are considered ambiguous input whose original provenance cannot be reconstructed reliably at this layer.

Thanks for catching the common-prefix issue — the reproducer exposed a real weakness in the boundary inference, even though I ended up handling the malformed unclosed case differently from the initial `c9d7974` attempt. Now you may try against `a906d6e`. If there are any issues, I am also open to making further improvements.

yilunzhao:
Re-ran the corpus against `a906d6e`. **All green, and the mask issue is fixed.**

| | `f6859dd` | **`a906d6e`** |
| --- | --- | --- |
| strict token equality | 10,726 / 10,726 | **10,726 / 10,726** |
| **loss mask identical to our reference implementation** | 10,726 / 10,726 | **10,726 / 10,726** |
| supervised-token delta | 0 | **0** |
| the 783 multi-user-query rows (tokens / masks) | 783 / 783 | **783 / 783** |

Same setup as before: verl `1dda039b` with the PR applied to a scratch copy, official Qwen3.5
template, 10,726 real multi-turn tool-calling trajectories. This round we compared **loss masks
per row**, not just token streams, since the mask inference was rewritten — no regression anywhere.
All three of your regression tests pass on our pin, including
`test_multiturn_sft_qwen36_preserve_thinking`.

The probe approach fixes the over-mask at the root rather than papering over it. Our reproducer
from last round now behaves the way we would compute it:

| think-dropped body | `f6859dd` masked | `a906d6e` masked |
| --- | --- | --- |
| `<think>unclosed literal body` | 4 tokens | **3** |
| `<think>\nunclosed literal body` | 5 tokens | **3** |

Because both probes carry synthetic reasoning *and* synthetic content, real response text can no
longer extend the common prefix — that was exactly the failure mode, so this closes it generally,
not just for the shape we happened to report.

On your note that unclosed serialized prefixes are ambiguous records: we went looking for a case
where that policy would make your output differ from ours, and did not find one. **10 assistant
content shapes × {think-kept, think-dropped} = 20 cases** — unclosed `<think>` prefixes with and
without a trailing newline, an orphan `</think>`, `reasoning_content`-field variants including an
empty one, empty content, and content that is only `<think>` — **0 disagreements**, every case
strict-token-equal with the same masked header and the same supervised span. So on the Qwen3.5
template the ambiguity does not turn into a behavioural difference.

Two small notes, neither blocking:

* You infer one header length per `(enable_thinking, remove_system_prompt)` and cache it, whereas we
  measure the header on each turn's own ids. They coincide on this template (dropped header is
  always `<|im_start|>assistant\n`, kept always 5 tokens) and the corpus agrees at 10,726/10,726, so
  this is only worth knowing if a future template has a variable-length header — and your validation
  prefix would assert rather than mis-mask if so.
* `sanity_check` still compares `input_ids` only and never `loss_mask`, which is why the round-2
  issue was silent. A mask-level check would make this class of bug self-reporting; we keep one in
  our own harness for that reason.

**We did not validate the Qwen3.6 extension.** Our corpus is Qwen3.5 only and we hold no Qwen3.6
template or trajectories. Your `test_multiturn_sft_qwen36_preserve_thinking` substitutes the
`preserve_thinking` condition into the Qwen3.5 template rather than loading a 3.6 model; we ran it
and it passes, but that exercises the branch, not real Qwen3.6 data — please don't read our numbers
as covering 3.6.

We plan to delete our adaptation layer and use this PR directly once it merges.

wuxibin89:
In https://github.com/verl-project/verl/pull/6779 https://github.com/volcengine/verl/issues/6804, we add `ContinuousTokenBuilder` abstraction to incremental encode messages in multi-turn:
- build_initial_tokens
- merge_non_assistant_tokens
- merge_assistant_tokens

In RL phase, we have enable `ContinuousTokenBuilder` by default, e.g [ToolAgentLoop](https://github.com/verl-project/verl/blob/main/verl/experimental/agent_loop/tool_agent_loop.py). For SFT phase, the multi-turn messages incremental encode is exactly same as RL phase, so we should enable ContinuousTokenBuilder in MultiTurnSFTDataset as well.

@yph22 Could you help to do it? cc @gxlvera

yph22:
@wuxibin89 Happy to help with this. I’ll first take a closer look at the design and current RL usage of `ContinuousTokenBuilder` to make sure I understand its intended multi-turn token semantics, and then try to align MultiTurnSFTDataset with the same abstraction.

yph22:
@wuxibin89 Before making the change, I’d like to confirm one semantic point to make sure I understand the intended behavior of CT correctly.

For Qwen3.5, there seems to be a difference between continuous-token rollout semantics and the model’s standard chat-template inference semantics when a new real user turn arrives. For example:

`user1 -> assistant1(<think>reason1</think>answer1) -> user2 -> assistant2`

With the official [Qwen3.5 chat template](https://huggingface.co/Qwen/Qwen3.5-9B/blob/main/chat_template.jinja), rendering the history for `assistant2` **drops `reason1`**, because `assistant1` is now before the last real user query. This is also what happens in the usual OpenAI-compatible serving flow (e.g. vLLM chat completions), where **the full `messages` history is rendered again for the new request.**

With `ContinuousTokenBuilder`, my understanding is that once `assistant1` has been generated and merged into `runtime_token_ids`, those tokens are preserved when `user2` is appended, **so `assistant2` would still see `reason1`**.  This behavior is also discussed in [#6719](https://github.com/verl-project/verl/issues/6719#:~:text=This%20difference%20should%20be%20allowed%2C%20because%20TITO%27s%20goal%20is%20to%20preserve%20the%20real%20runtime%20token%20prefix%2C%20not%20to%20canonicalize%20historical%20assistant%20messages%20after%20the%20fact). Would this introduce a mismatch with the standard chat-template-based inference behavior?

So I wanted to confirm the intended SFT semantics here: should `MultiTurnSFTDataset` intentionally follow the continuous-token/RL rollout behavior and preserve previously generated reasoning tokens, even when this differs from re-rendering the full conversation with the model’s chat template?

yph22:
Besides, `ContinuousTokenBuilder.merge_assistant_tokens()` assumes access to the **actual model-generated `assistant_token_ids`**, which is true during RL rollout. However, **`MultiTurnSFTDataset` usually only has serialized messages**. In that case we must **reconstruct assistant token IDs from text/template first**. Wouldn't this mean that **exact TITO cannot be guaranteed for SFT**—for example, the [A,B] -> text -> [C] BPE issue described in [#6719 ](https://github.com/verl-project/verl/issues/6719) can already occur at this initial reconstruction step?

wuxibin89:
> @wuxibin89 Before making the change, I’d like to confirm one semantic point to make sure I understand the intended behavior of CT correctly.
> 
> For Qwen3.5, there seems to be a difference between continuous-token rollout semantics and the model’s standard chat-template inference semantics when a new real user turn arrives. For example:
> 
> `user1 -> assistant1(<think>reason1</think>answer1) -> user2 -> assistant2`
> 
> With the official [Qwen3.5 chat template](https://huggingface.co/Qwen/Qwen3.5-9B/blob/main/chat_template.jinja), rendering the history for `assistant2` **drops `reason1`**, because `assistant1` is now before the last real user query. This is also what happens in the usual OpenAI-compatible serving flow (e.g. vLLM chat completions), where **the full `messages` history is rendered again for the new request.**
> 
> With `ContinuousTokenBuilder`, my understanding is that once `assistant1` has been generated and merged into `runtime_token_ids`, those tokens are preserved when `user2` is appended, **so `assistant2` would still see `reason1`**. This behavior is also discussed in [#6719](https://github.com/verl-project/verl/issues/6719#:~:text=This%20difference%20should%20be%20allowed%2C%20because%20TITO%27s%20goal%20is%20to%20preserve%20the%20real%20runtime%20token%20prefix%2C%20not%20to%20canonicalize%20historical%20assistant%20messages%20after%20the%20fact). Would this introduce a mismatch with the standard chat-template-based inference behavior?
> 
> So I wanted to confirm the intended SFT semantics here: should `MultiTurnSFTDataset` intentionally follow the continuous-token/RL rollout behavior and preserve previously generated reasoning tokens, even when this differs from re-rendering the full conversation with the model’s chat template?

RL TITO requires not modifying trajectory prefix in multi-turn. For those model chat template that drop `<think></think>`, there's 2 options:
1. incremental merge assistant without dropping `<think></think>`, which incurs mismatch between RL rollout and online serving.
2. re-render history messages in each turn with dropping `<think></think>`, which splits multi-turn into multiple trajectories. For example, [user, assistant_0, tool_0, assistant_1, tool_1, assistant_2] split into 3 trajectories:
- [user, **assistant_0**]
- [user, assistant_0, tool_0, **assistant_1**]
- [user, assistant_0, tool_0, assistant_1, tool_1, **assistant_2**]

To avoid trajectory explosion, verl choose option 1 both in RL and SFT.

wuxibin89:
> Besides, `ContinuousTokenBuilder.merge_assistant_tokens()` assumes access to the **actual model-generated `assistant_token_ids`**, which is true during RL rollout. However, **`MultiTurnSFTDataset` usually only has serialized messages**. In that case we must **reconstruct assistant token IDs from text/template first**. Wouldn't this mean that **exact TITO cannot be guaranteed for SFT**—for example, the [A,B] -> text -> [C] BPE issue described in [#6719 ](https://github.com/verl-project/verl/issues/6719) can already occur at this initial reconstruction step?

It's ok to encode assistant messages once in SFT dataset. One gold rule is that: 
> Never re-encode tokens you’ve decoded.

So we never decode assistant token in RL rollout, we only append it.

yph22:
@wuxibin89  Thanks for the clarification. I’ve updated the PR in `032a0b2` to align `MultiTurnSFTDataset` with the **Continuous Token** design.

The SFT path now builds the initial prompt once with `build_initial_tokens`, encodes each prepared assistant continuation once and appends it through `merge_assistant_tokens`, and incrementally appends later user/tool context through `merge_non_assistant_tokens`. The loss mask is updated through `align_response_metadata`, so CT boundary edits are applied consistently to both the token stream and its supervision metadata.

Following the CT/TITO semantics, **previously constructed trajectory content is no longer re-rendered or re-encoded.** In particular, appending a later user turn no longer causes an earlier Qwen3.5 assistant turn to be rendered again or its historical reasoning to be removed. Therefore, **the final SFT input_ids are intentionally not required to equal a fresh whole-conversation `apply_chat_template(messages)` result**. This implements option 1 discussed above and keeps SFT trajectory construction aligned with the RL TITO path.

For message-only SFT data, the original generation-time assistant token IDs are unavailable. Following the one-time encoding rule discussed above, `tokenize_assistant_message` renders each prepared assistant message behind a fixed synthetic prompt, extracts and encodes only its assistant continuation, and then keeps those reconstructed token IDs unchanged for the rest of the trajectory.

I also updated the regression tests to check append-only Continuous Token behavior rather than whole-conversation chat-template equality, including Qwen3.5/3.6 thinking behavior, loss-mask alignment, tool turns, and multimodal dataset paths. Would appreciate another look when you have a chance. 

@yilunzhao A heads-up: this supersedes the whole-conversation-equivalent implementation you previously validated at `a906d6e`. Therefore, the 10,726/10,726 strict token/loss-mask equality result applies to the previous implementation, not to the current PR head.

If your downstream use case specifically requires canonical whole-conversation rendering and Qwen3.5 history-dropping behavior, the rebased snapshot of the previous implementation remains available at `98f4bd6`. That snapshot remains an alternative downstream solution, but it is no longer the design proposed by this PR.

Sorry for the change in direction after your extensive validation, and thank you both for the detailed feedback.

hai-meh-cs:
Thanks for the CT rework here, and for fixing the earlier `str + list` crash on this path. On a clean checkout of PR head `032a0b25` (upstream `verl 0.10.0.dev0`, real `openai/gpt-oss-20b`, via `MultiTurnSFTDataset` with `hf_model_type='gpt_oss'` so GPTOSS dispatch is correct), a valid Harmony tool-call turn still fails.

`GptOssContinuousTokenBuilder` overrides only `_tokenize_tool_group`; it inherits the base `_normalize_assistant_token_ids`, which reverse-scans each assistant span for `tokenizer.eos_token_id` and raises if absent. For gpt-oss, EOS is `<|return|>`, so normal responses pass. But a minimal tool-call conversation accepted by the Harmony template (assistant turn rendered ending in `<|call|>`) has no `<|return|>`, so the base scan rejects it (abridged):

> Continuous Token assistant token-id suffix does not contain eos_token_id [200002]; tail=[..., 200012]

The Qwen tests don't catch it, since their assistant turns end in `<|im_end|>` = eos.

Could GPTOSS normalization accept `<|call|>` as a valid Harmony assistant-turn boundary alongside `<|return|>`? (One option: a small per-builder terminator set, default `{eos}`.) A gpt-oss tool-call regression test would lock it in. Repro below.

```
import tempfile, os
import pandas as pd
from transformers import AutoTokenizer
from verl.utils.dataset.multiturn_sft_dataset import MultiTurnSFTDataset

tok = AutoTokenizer.from_pretrained("openai/gpt-oss-20b")

# A well-formed tool conversation: assistant makes a tool call, tool responds, assistant answers.
messages = [
    {"role": "user", "content": "What's the weather in Paris?"},
    {"role": "assistant", "content": "",
     "tool_calls": [{"type": "function",
                     "function": {"name": "get_weather", "arguments": '{"city":"Paris"}'}}]},
    {"role": "tool", "content": "18C, sunny"},
    {"role": "assistant", "content": "It's 18C and sunny in Paris."},
]
tools = [{"type": "function", "function": {"name": "get_weather", "description": "w",
          "parameters": {"type": "object", "properties": {"city": {"type": "string"}},
                         "required": ["city"]}}}]

d = tempfile.mkdtemp()
path = os.path.join(d, "data.parquet")
pd.DataFrame({"messages": [messages], "tools": [tools]}).to_parquet(path)

ds = MultiTurnSFTDataset(
    parquet_files=path, tokenizer=tok, processor=None,
    config={"max_length": 1024, "truncation": "error",
            "multiturn": {"messages_key": "messages"}, "tools_key": "tools"},
)
item = ds[0]  # On PR head 032a0b25 this raises (tail abridged):
              #   ValueError: Continuous Token assistant token-id suffix does not contain
              #   eos_token_id [200002]; tail=[..., 200012]
              # 200002=<|return|> (eos), 200012=<|call|> (the tool-call turn's terminator).
print("OK", item["input_ids"].shape)
```

verl/utils/tokenizer/continuous_token.py
```diff
@@ -164,6 +168,84 @@ def merge_assistant_tokens(self, runtime_token_ids: list[int], assistant_token_i
...
+
+        if not eos_token_ids:
+            return list(assistant_token_ids)
+        for index in range(len(assistant_token_ids) - 1, -1, -1):
+            if assistant_token_ids[index] in eos_token_ids:
+                return list(assistant_token_ids[: index + 1])
+        raise ValueError(
+            "Continuous Token assistant token-id suffix does not contain eos_token_id "
+            f"{sorted(eos_token_ids)}; tail={assistant_token_ids[-16:]}"
+        )
```
hai-meh-cs: The base `_normalize_assistant_token_ids` assumes every valid assistant span ends in `eos_token_id`. That holds for the tested Qwen templates, but a valid gpt-oss/Harmony assistant **tool-call** turn ends in `<|call|>` (not `<|return|>`/eos), so this reverse scan rejects it. `GptOssContinuousTokenBuilder` inherits this method (it overrides only `_tokenize_tool_group`), so correct dispatch doesn't help.

Could GPTOSS normalization recognize `<|call|>` as a valid assistant-turn boundary here? One implementation would be a small per-builder set of accepted terminators (default `{eos}`, with `GptOssContinuousTokenBuilder` adding `<|call|>`), resolving the id via the tokenizer rather than hardcoding. The exact shape is yours to choose.

yph22: This seems to expose a more general issue in the SFT-side reconstruction: `tokenizer.eos_token_id` is not necessarily the complete set of valid assistant generation terminators. **Gemma 4**, for example, already declares **multiple `eos_token_ids`** in its generation config, and other templates may use turn delimiters that differ from the tokenizer EOS. I’ll try to make the terminator handling model/protocol-aware rather than special-casing GPT-OSS only, although this is somewhat beyond the original scope of this PR, which was primarily intended to fix the Qwen3.5/Qwen3.6 multi-turn SFT behavior.

yph22:
@hai-meh-cs  Thanks for the detailed report. Fixed in `b1d168cb`.

I generalized the assistant-terminator handling rather than hard-coding GPT-OSS behavior into the base normalization logic. The base builder now exposes a message-aware `_assistant_terminator_ids()` hook. For GPT-OSS, assistant tool-call turns terminate at `<|call|>`, while ordinary/final assistant turns terminate at the tokenizer EOS (`<|return|>`). SFT reconstruction now trims at the first valid generated terminator, matching runtime stop behavior.

Your `MultiTurnSFTDataset` reproducer is now covered by a tokenizer-backed regression test using the real `openai/gpt-oss-20b` tokenizer. 

I also audited the registered Continuous Token families for analogous protocol assumptions and added targeted continuation/boundary handling and regression coverage where the generic single-EOS/Jinja path is not sufficient. **This includes Gemma4, MiniMax Text/M2/VL, GLM, DeepSeek V2/V3/V4/VL2, and the relevant VL-specific paths**. Protocols where structured tool reconstruction is not currently supported now fail explicitly rather than silently producing malformed training tokens.

Validation:

* `tests/utils/test_continuous_token_on_cpu.py`: 163 passed
* GPT-OSS dataset regressions with the real tokenizer: 2 passed
* real Qwen3.5/Qwen3.6 tokenizer and processor checks: passed
* real DeepSeek-V4 multi-turn SFT/loss-mask check: passed
* Ruff and `git diff --check`: passed

Please take another look when you're free. Thanks again for catching this.

yph22:
@wuxibin89 A quick update on the current state of the PR.

Following your suggestion, `MultiTurnSFTDataset` now uses the Continuous Token path rather than maintaining a separate multi-turn rendering implementation. The SFT behavior therefore follows the CT/TITO semantics: **previously constructed trajectory tokens are preserved**, and the final token sequence is not required to match a one-shot `apply_chat_template` over the complete message history.

While adapting message-only SFT data to CT, the change also exposed several model/protocol-specific assumptions in assistant reconstruction. For example, GPT-OSS tool-call generations terminate with `<|call|>` rather than the tokenizer EOS (caught by @hai-meh-cs ). I therefore generalized the reconstruction/terminator handling and audited the registered CT model families for analogous cases, adding model-specific handling or explicit fail-fast behavior where the generic path is not valid.

As a result, the PR has grown beyond the original Qwen3.5/Qwen3.6 multi-turn rendering fix and now covers the more general `MultiTurnSFTDataset` → `ContinuousTokenBuilder` integration and the protocol handling needed to make that SFT path robust across the registered CT families.

The latest changes are in `b1d168cb`, with the corresponding regression coverage added. The current test suite and the real-tokenizer/model-family checks I ran are passing.

Would appreciate a review of the latest version when you have a chance. Thanks again.

wuxibin89:
@gxlvera Please help to take a review.

yph22:
@wuxibin89 I've noticed that [#7628](https://github.com/verl-project/verl/pull/7628) was just merged into `main` and changes the same **Continuous Token** paths touched by this PR, so the branch currently needs to be updated and the merge conflicts resolved. I’ll integrate the latest CT changes and rerun the regression tests.

yph22:
@wuxibin89 I’ve rebased this PR onto the latest `main`, including [#7628](https://github.com/verl-project/verl/pull/7628), and pushed the follow-up adaptation in `5c37588`.

The implementation now follows the generation-prompt fusion introduced by #7628:

- The default CT path, including Qwen3.5/3.6, fuses `add_generation_prompt=True` into the final appended non-assistant group, so deriving the next assistant scaffold no longer requires a separate full-history render. This follows the new default behavior in #7628.

- MiniMax Text-01 and MiniMax-VL intentionally opt out of this fusion because their append hooks encode role payloads separately and their next-assistant headers are fixed scaffolds. They therefore append the separately reconstructed fixed scaffold through `_tokenize_generation_prompt_delta()`.

- The model-specific `_tokenize_tool_group` / `_tokenize_single_non_tool` overrides touched by this PR have been adapted to the new `add_generation_prompt` extension API, including the explicit unsupported-tool path for Kimi-VL.

- Builders that opt out of fusion use `_tokenize_generation_prompt_delta()` as the fallback/extension point
introduced by #7628. Where full-history derivation is required, it may re-render the structured message history to derive the next scaffold, but it never replaces or mutates the already committed runtime token IDs.

- GLM and the classic DeepSeek builders inherit the new default fused path without requiring additional model-specific overrides. DeepSeek-V4 and DeepSeek-VL2 retain their existing custom incremental-tokenization paths.

- GPT-OSS and Gemma4 retain the opt-out behavior required by their model-specific protocols. Gemma4-VL inherits the same behavior from the Gemma4 builder.

Validation completed:

- `tests/utils/test_continuous_token_on_cpu.py`: 167 passed

- Real Qwen3.5 tokenizer integration: 2 passed, covering append-only/loss-mask behavior and the `preserve_thinking` regression

- Real Qwen3.6-27B tokenizer integration: 1 passed, covering append-only/loss-mask behavior with the current Qwen3.6 template

- Real GPT-OSS tokenizer integration: 2 passed, covering custom chat-template kwargs and Harmony tool-call `<|call|>` / `<|return|>` terminators

-  Protocol-level CPU regression coverage also includes the GLM, Gemma4, and DeepSeek V3.1/V4/VL2 paths. Builder
wiring coverage additionally verifies the corresponding VL inheritance and dispatch paths, including Gemma4-VL. These use mock/native-protocol fixtures rather than real-tokenizer integration tests.

- Ruff lint and formatting checks passed

The PR is now conflict-free against the latest `main`. Could you please take another look? Thanks!

hai-meh-cs:
Thanks @yph22. This seems to work well for me. The message-aware `_assistant_terminator_ids()` hook is the right shape: it fixes the whole family instead of special-casing GPT-OSS, and the fail-fast on unsupported tool-reconstruction protocols is what we'd want, since silent malformed training tokens were the real risk.

I re-ran our original `MultiTurnSFTDataset` reproducer against the current head (`5c37588`) with the real `openai/gpt-oss-20b` tokenizer: the `<|call|>` tool-call terminator is now retained and the assistant loss mask is correct. I also spot-checked Gemma-4 (one of the families you hardened) and it looks right.

Appreciate the thorough audit across the CT families and the rebase onto #7628. Nothing further from our side.

qfhu0ng:
Hi @yph22, I reviewed and tested this work and opened #7700 as a continuation of #7322, preserving its original commit history and authorship. It rebases the work onto current main and adds a narrow structured-image fix and regression coverage. Linking it here for visibility and coordination. Thank you for the original implementation!

PR #5948, opened 2026-04-09
[model] fix: Align Qwen VLM Ulysses + fused-kernel paths and harden multimodal position-id / vision-embed handling

jamindy:
### What does this PR do?

This PR fixes multiple multimodal runtime issues across `qwen2_vl`, `qwen3_vl`, and `qwen3_5` under **Ulysses sequence parallelism + fused kernels**.

It includes:
- fixing Qwen3-VL vision position-embedding behavior under `FSDP2 + multimodal` execution
- refactoring shared bilinear vision position-embedding interpolation logic for Qwen3.5 and Qwen3-VL into a common helper
- fixing Ulysses SP + fused-kernel label alignment errors in VLM forward paths
- removing hard-coded `rope_dim=4` assumptions and repairing broken nested `position_ids` layout handling

Previously observed errors:
1. `aten.index_select.default got mixed torch.Tensor and DTensor`
2. `Expected all tensors to be on the same device, but got index is on cpu, different from other tensors on cuda:0`
3. `The size of tensor a (xxx) must match the size of tensor b (4) at non-singleton dimension 2`
4. `qwen3_vl` / `qwen3_5` paths previously failed when `SP > 1` in specific fused-kernel multimodal flows

Root cause:
The monkey-patched Qwen VLM interpolation and SP/fused-kernel paths were not fully consistent with:
- `FSDP2` sharded embedding weights (`DTensor`)
- device/offload behavior
- and local-label alignment requirements after Ulysses slicing
This led to tensor type/device mismatches and hidden/label shape mismatches.

### Related issues
- Fixes #5829
- Related: #4756, #5633, #5571

### Validation
On the latest `verl` with `Transformers 5.5.0`, we validated:
- `qwen2_vl`
- `qwen3_vl`
- `qwen3_5`

All tests passed with **Ulysses sequence parallelism** and **fused kernels enabled**.

### Test script (example)
```bash
python3 -m verl.trainer.main_ppo \
  algorithm.adv_estimator=grpo \
  data.train_files=$train_files \
  data.val_files=$test_files \
  data.train_batch_size=4 \
  data.max_prompt_length=1024 \
  data.max_response_length=2048 \
  data.image_key=images \
  data.truncation=error \
  actor_rollout_ref.model.path=/models/Qwen3-VL-4B-Instruct \
  actor_rollout_ref.model.use_remove_padding=True \
  actor_rollout_ref.model.use_fused_kernels=True \
[...]

PR #3496, opened 2025-09-16
[model] fix: refactor qwen2vl patches & support no-image input for fsdp

hiyouga:
### What does this PR do?

This PR tries to fix #3491 

### Checklist Before Starting

- [x] Search for similar PRs. Paste at least one query link here: ...
- [x] Format the PR title as `[{modules}] {type}: {description}` (This will be checked by the CI)
  - `{modules}` include `fsdp`, `megatron`, `sglang`, `vllm`, `rollout`, `trainer`, `ci`, `training_utils`, `recipe`, `hardware`, `deployment`, `ray`, `worker`, `single_controller`, `misc`, `perf`, `model`, `algo`, `env`, `tool`, `ckpt`, `doc`, `data`
  - If this PR involves multiple modules, separate them with `,` like `[megatron, fsdp, doc]`
  - `{type}` is in `feat`, `fix`, `refactor`, `chore`, `test`
  - If this PR breaks any API (CLI arguments, config, function signature, etc.), add `[BREAKING]` to the beginning of the title.
  - Example: `[BREAKING][fsdp, megatron] feat: dynamic batching`

### Test

Tested with [latest transformers](https://github.com/huggingface/transformers/tree/6e50a8afb2540ac1acaa4b62cf1dd5f1170f6511)

image
image
image

### API and Usage Example

> Demonstrate how the API changes if any, and provide usage example(s) if possible.

```python
# Add code snippet or script demonstrating how to use this
```

#### ⚠️ Breaking

We adopt a new format for Qwen2VL's position ids: (4, batch size, seq len)

Assuming a vision position ids (mrope) has a shape of (3, batch size, seq len) and a text position ids (normal rope) has a shape of (1, batch size, seq len), we concatenate both to obtain the final position ids.

This aligns with the implementation in the Transformers >= 4.54.0 🤗
https://github.com/huggingface/transformers/blob/v4.54.0/src/transformers/models/qwen2_vl/modeling_qwen2_vl.py#L1469

#### 🎤 New

We have refactored the Qwen2VL and Qwen2.5VL patches, supporting no-image input for FSDP by introducing fake ViT inputs. We have also removed some redundant code for better maintainability.

#### 🚨 Changes

We move the ulysses logic into the attention function. So the position ids will be scattered before the language model part.

### Checklist Before Submitting

> [!IMPORTANT]
> Please check all the following items before requesting a review, otherwise the reviewer might deprioritize this PR for review.

[...]

Issue #3491, opened 2025-09-16
warp_policy will make GPU hang when training mixed multi modal data


techkang:
### System Info

----------Python Info----------
Version      : 3.10.12
Compiler     : GCC 11.4.0
Build        : ('main', 'Jul 29 2024 16:56:48')
Arch         : ('64bit', 'ELF')
------------Pip Info-----------
Version      : 25.1.1
Directory    : /usr/local/lib/python3.10/dist-packages/pip
vllm         : 0.10.0
sglang       : not found.
ray          : 2.47.1
torch        : 2.7.1
----------verl Info-----------
Version      : 0.5.0.dev
Directory    : /file_system/kangsheng/verl/verl
Commit Hash  : 44cdc4a5275c77c2f64487885c1680997d6cbab1
----------Platform Info----------
Platform     : Linux-5.4.250-4-velinux1u1-amd64-x86_64-with-glibc2.35
system       : Linux
node         : di-20250904113923-94xhn
release      : 5.4.250-4-velinux1u1-amd64
version      : #velinux1u1 SMP Debian 5.4.250-4-velinux1u1 Thu May 23 09:30:06 
----------Environment----------
CUDA Runtime : 12.6
CUDA Compiler : Cuda compilation tools, release 12.6, V12.6.20
----------System Info----------
CPU Memory      : 1785.00 GB
GPU Count       : 8
GPU 1   Type    : NVIDIA H20
GPU 1   Memory  : 95.58 GB
GPU 2   Type    : NVIDIA H20
GPU 2   Memory  : 95.58 GB
GPU 3   Type    : NVIDIA H20
GPU 3   Memory  : 95.58 GB
GPU 4   Type    : NVIDIA H20
GPU 4   Memory  : 95.58 GB
GPU 5   Type    : NVIDIA H20
GPU 5   Memory  : 95.58 GB
GPU 6   Type    : NVIDIA H20
GPU 6   Memory  : 95.58 GB
GPU 7   Type    : NVIDIA H20
GPU 7   Memory  : 95.58 GB
GPU 8   Type    : NVIDIA H20
GPU 8   Memory  : 95.58 GB

### Information

- [ ] The official example scripts
- [x] My own modified scripts

### Tasks

- [ ] An officially supported task in the `examples` folder (such as GLUE/SQuAD, ...)
- [x] My own task or dataset (give details below)

### Reproduction

[...]

PR #7323, opened 2026-08-08
[model, fsdp] fix: align Qwen3.5 vision calls across ranks

yangzhou24:
### What does this PR do?

FSDP2 records each wrapped-module forward use with multiplicity for pre-backward parameter unsharding. With dynamic multimodal batching, a rank containing both images and videos enters `model.visual` twice, while an image-only, video-only, or text-only rank enters it once. The resulting rank-local usage mismatch can make later parameter all-gathers execute in different orders.

This PR concatenates image and video patches and their `grid_thw` metadata before one shared visual forward. It then splits the output at the grid-derived image feature boundary before scattering image and video embeddings back to their token positions. Image-only, video-only, and the existing text-only dummy path continue to enter the visual tree exactly once.

Related context: issue #3491 and PRs #3496, #3670, #5881, and #5948. This is not a duplicate of #3496/#3670, which address 0-vs-1 visual usage for text-only ranks, or #5948, which addresses Ulysses/fused-kernel/device handling while retaining separate image and video visual calls.

### Checklist Before Starting

- [x] Search for similar PRs: https://github.com/verl-project/verl/pulls?q=is%3Apr+is%3Aopen+Qwen3.5+visual+FSDP

### Test

```text
PYTHONDONTWRITEBYTECODE=1 python -m pytest -q -p no:cacheprovider tests/models/test_qwen3_5_mixed_modal_fsdp_on_cpu.py

6 passed, 15 warnings in 53.82s
```

The regression suite covers combined-call input ordering, forward and gradient equivalence against the previous separate-call math, a real small `Qwen3_5VisionModel`, every supported modality combination, and two-rank CPU/Gloo FSDP2 training with different rank-local modality mixes and gradient synchronization modes.

```text

All hooks passed.
```

The identical Qwen3.5 source change was also exercised in two clean single-node 8-GPU actor runs that completed 20/20 steps with zero exit codes. An integrated full-epoch run completed 17/17 steps with zero exit codes; that run also contained unrelated cache-handoff and TransferQueue stability fixes, so it is supporting workload evidence rather than an isolated A/B result.

### API and Usage Example

No public API, CLI, or configuration changes are introduced. Existing Qwen3.5 FSDP2 training uses the aligned visual path automatically.

### Design & Code Changes

- Concatenate image and video pixel tensors and `grid_thw` metadata in image-then-video order.
- Invoke the shared Qwen3.5 visual encoder once per micro-batch.
- Derive the image output boundary from `grid_thw` and `spatial_merge_size`.
- Preserve the existing per-modality token-count validation and text-only dummy visual forward.
- Add CPU numerical and two-rank FSDP2 regression coverage.

- [x] Apply [pre-commit checks](https://github.com/verl-project/verl/blob/main/CONTRIBUTING.md#code-linting-and-formatting): `pre-commit install && pre-commit run --all-files --show-diff-on-failure --color=always`.
- [x] Documentation is not required because there is no user-facing API or configuration change.
- [x] Added `*_on_cpu.py` regression tests that are automatically collected by the existing CPU unit-test workflow.
- [ ] Once the PR is ready for CI, send a message in the [`ci-request` channel](https://verl-project.slack.com/archives/C091TCESWB1) in the [verl Slack workspace](https://join.slack.com/t/verl-project/shared_invite/zt-3855yhg8g-CTkqXu~hKojPCmo7k_yXTQ).
- [x] The `recipe` submodule is not involved.

### DIFF
--- a/tests/models/test_qwen3_5_mixed_modal_fsdp_on_cpu.py
+++ b/tests/models/test_qwen3_5_mixed_modal_fsdp_on_cpu.py
@@ -1,0 +1,436 @@
+# Copyright 2026 Bytedance Ltd. and/or its affiliates
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from __future__ import annotations
+
+import time
+from contextlib import nullcontext
+from datetime import timedelta
+from types import SimpleNamespace
+
+import pytest
+import torch
+import torch.distributed as dist
+import torch.multiprocessing as mp
+from torch.distributed.device_mesh import init_device_mesh
+from torch.distributed.fsdp import fully_shard
+from torch.distributed.tensor import DTensor
+from transformers.models.qwen3_5.configuration_qwen3_5 import Qwen3_5VisionConfig
+from transformers.models.qwen3_5.modeling_qwen3_5 import Qwen3_5VisionModel
+
+from verl.models.transformers.qwen3_5 import _get_input_embeds
+from verl.workers.engine.fsdp.transformer_impl import FSDPEngine
+
+_IMAGE_TOKEN_ID = 29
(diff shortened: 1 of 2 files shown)


yangzhou24:
Hi @vermouth1992, could you take a look at this PR when you have a chance?

It fixes a potential FSDP2 collective-order mismatch in Qwen3.5 training when ranks receive different image/video mixes. I've added regression coverage, including two-rank FSDP2 tests, and validated the fix in two single-node 8-GPU runs. Everything ran smoothly when post-training Qwen3.5-9B with multi-task GRPO on 40/80 GPUs using 1.6M data.

I'd appreciate your feedback and am happy to make any changes. If someone else would be better suited to review this, could you point me to them? Thanks!

PR #7324, opened 2026-08-08
[ckpt, fsdp]  feat: sharded delta weight sync for the TorchTitan engine

attack204:
### What does this PR do?

Adds **sharded delta weight sync** to the TorchTitan training engine.

**Support**

- [x] FSDP
- [x] FSDP+TP
- [x] FSDP+EP
- [x] HSDP
- [x] HSDP+TP
- [x] HSDP+EP (unit-tested; no end-to-end row, see below)

**Limitation**

- [ ] FSDP+PP: PP is not supported in the TorchTitan engine, even with full weight sync.

### Test

**Results**

- A800 80GB, up to 32 GPUs
- GSM8K GRPO
- `one_step_off_policy` disaggregated
- TorchTitan trainer → SGLang rollout

**FSDP2**

| model (placement) | trainer | rollout | `delta_sharded` | `nccl` (full broadcast) | speedup |
|---|---|---|---|---|---|
| Qwen3-8B (2+2 nodes, 32 GPUs) | 16 GPUs, FSDP2 `dp_shard=16` | 16 GPUs, SGLang TP2 × 8 replicas | **3.34 s** | 34.26 s | **10.3×** |
| Qwen3-8B (1+1 nodes, 16 GPUs, 50 steps sustained) | 8 GPUs, FSDP2 `dp_shard=8` | 8 GPUs, SGLang TP2 × 4 replicas | **3.00 s** | 11.00 s | 3.7× |
| Qwen3-0.6B (1 node, 8 GPUs) | 6 GPUs, FSDP2 `dp_shard=6` | 2 GPUs, SGLang TP2 × 1 replica | 0.59 s | 0.61 s | 1.0× |

**FSDP2 × TP**

| model (placement) | trainer | rollout | `delta_sharded` | `nccl` (full broadcast) | speedup |
|---|---|---|---|---|---|
| Qwen3-8B (2+2 nodes, 32 GPUs) | 16 GPUs, FSDP2 `dp_shard=8` × **TP2** | 16 GPUs, SGLang TP2 × 8 replicas | **3.40 s** | 34.26 s | **10.1×** |
| ↳ same config without TP, for reference | 16 GPUs, FSDP2 `dp_shard=16` | 16 GPUs, SGLang TP2 × 8 replicas | 3.34 s | 34.26 s | 10.3× |

**FSDP2 × EP (MoE)**

| model (placement) | trainer | rollout | `delta_sharded` | `nccl` (full broadcast) | speedup |
|---|---|---|---|---|---|
| Qwen3-30B-A3B MoE (2+2 nodes, 32 GPUs) | 16 GPUs, FSDP2 × **EP8** (efsdp=2) | 16 GPUs, SGLang TP4 × 4 replicas | **8.07 s** | 43.43 s | **5.4×** |

**HSDP × TP**

| model (placement) | trainer | rollout | `delta_sharded` | `nccl` (full broadcast) | speedup |
|---|---|---|---|---|---|
| Qwen3-8B (2+2 nodes, 32 GPUs) | 16 GPUs, HSDP `dp_replicate=2 × dp_shard=4` × **TP2** | 16 GPUs, SGLang TP2 × 8 replicas | **2.41 s** | 10.24 s | **4.2×** |

**Regression check: verl's own FSDP engine (`model_engine=dp`)**

| trainer | `delta_sharded` | payload | changed ratio | `nccl` | speedup |
|---|---|---|---|---|---|
| Qwen3-8B (2+2 nodes, 32 GPUs), 16 GPUs FSDP2 `fsdp_size=16` | **2.24 s** | 418.6 MB | 0.893 % | 38.58 s | **17.2×** |
| the same, FSDP1 | 24.86 s | 435.3 MB | 0.929 % | 65.22 s | 2.6× |

**Correctness.** `critic/score` over 20 steps at Qwen3-8B, sampling independent between the two runs:

```
nccl   0.285 0.260 0.297 0.291 0.307 0.402 0.367 0.387 0.459 0.412 0.522 0.484 0.543 0.512 0.428 0.535 0.586 0.613 0.664 0.623
delta  0.293 0.260 0.305 0.303 0.324 0.420 0.402 0.395 0.490 0.426 0.510 0.535 0.590 0.582 0.481 0.627 0.611 0.647 0.715 0.664
```

#### FSDP

- The HF rename (`layers.0.attention.wo.weight` → `model.layers.0.self_attn.o_proj.weight`) is now shared between the full and sharded exports. This is a correctness requirement, not tidiness: the delta pairs the two by name, so a name in one set and not the other silently freezes a weight at its seed value. With weight tying, `sd_adapter.to_hf()` drops `lm_head.weight` and the full export re-adds it by hand; factoring that into `_to_hf_named_params()` makes the two name sets identical by construction.
- `get_per_tensor_param_delta_shard` moved to `BaseEngine`, since it needs nothing but the shard export plus a per-parameter entry builder. The two identical `FSDPEngine` copies are removed; Megatron keeps its override (it diffs against something other than a pinned snapshot) and veomni keeps its `_hf_delta_entry`.
- PP is rejected at the export boundary with a message naming why, instead of failing cryptically mid-export.

#### TP

FSDP2 sharding a tensor dim that TP already cut comes back as `_StridedShard`, which `derive_dtensor_placement` used to reject on the grounds that the local shard is not one block — it is one: the strided form encodes a right-to-left cut order, which permutes *which* block a rank owns rather than interleaving the block, and `compute_local_shape_and_global_offset` already returns the permuted offset. What the placement math did need is a version-independent reading of "does this mesh dim cut the tensor": `_StridedShard` moved to C++ in torch 2.13 and stopped subclassing `Shard`, so `p.is_shard()` answers False for it there, silently dropping the mesh dim from the gather group. `_shard_dim()` answers that question for both spellings, and `_assert_even_strided()` rejects the uneven cut the strided offset formula cannot express.

#### EP

- EP needed no new geometry: EP cuts the expert dim of a fused expert stack and EFSDP cuts it again, so a routed expert weight is a `_StridedShard(0)Shard(0)` — the same two-mesh-dims-on-one-tensor-dim block as column-parallel TP, one dim over. What EP did need is naming. `to_hf()` yields only the locally owned experts, so each EP rank would announce a different set of HF tensors, and the gather is a lockstep collective where "which parameter are we on" must have the same answer everywhere. The stack is therefore exported whole, carrying a slot table naming every expert's HF tensor; the split happens receiver-side from indices that already say which expert each element belongs to. Every rank can build that table because the expert count is the stack's dim 0 — a shape, identical everywhere. Which parameters are stacks is decided structurally rather than by name matching: the adapter maps one torchtitan key to an HF template with a second `{}` placeholder, which is what "one weight here, many weights there" looks like.
- The dim-0 split is an identity — row `r` of the stack *is* the `r`-th HF tensor — so `_hf_entry_row_slots` recovers the per-expert counts with one `searchsorted` over the global indices, rather than the probe-driven converter machinery that opaque `to_hf_chunk` callables need.

**Also fixed here (separate commit): the full export dropped experts when EFSDP > 1.** Pre-existing and not specific to delta, but the delta seed is the full export, so EP could not be trusted until it was fixed. `to_hf()` splits the stack into per-expert tensors before it is made whole and keeps only the local ones; the export then all-gathered those pieces over the `ep` group alone, which is complete only when `ep` is the sole mesh dim cutting the expert dim. It is not — EFSDP cuts dim 0 as well — so at `ep=4, efsdp=2` the export emitted 32 of 64 experts per layer: correct values, silently missing rows, leaving the rollout engine with part-stale routed experts on the `nccl` path too. Gathering the stack's own DTensor with `full_tensor()` and splitting afterwards is placement-agnostic, since EP, EFSDP and any replicate dim beside them are already in the placements. Verified at `ep=1/2/4/8 × efsdp=8/4/2/1` against `full_tensor()` ground truth: every expert present and bit-identical, where before only `efsdp=1` was complete.

#### HSDP

A replicate mesh dim beside the cut dims is the layout the export used to reject. Under HSDP a column-parallel weight is `(Replicate, _StridedShard(0), Shard(0))`, and a routed expert weight under HSDP × EP is the same block one tensor dim over. Of the two things that have to hold, only one was already true.

Double-counting was already handled: `contributes` marks a rank sitting at a non-zero coordinate on any Replicate dim, and such a rank keeps lockstep with an empty delta rather than shipping a copy a peer is already sending. That test never cared how many Shard dims sat beside the Replicate one.

What did not generalise was the gather group. With one Shard dim the group is that dim's subgroup, which holds every other coordinate fixed and therefore skips the replicas for free. With several Shard dims the old code fell back to the flattened whole mesh, which spans the replicate dim too. That still gathers the *right* answer, since the replicas contribute nothing — but `gather_slot_entries_to_rank0` pads every rank's blob to the largest one, so the replicate degree would be charged in wire bytes and in rank 0's staging buffers for blobs that are empty by construction. `_shard_dims_group` builds the group the way `mesh.get_group(mesh_dim=d)` does for a single dim, generalised to several: vary the Shard dims, hold this rank's coordinate fixed on the rest. Groups are cached by rank set and created on every rank, since `dist.new_group` is collective and every trainer rank walks the export in the same order.

**Also fixed here: `_get_data_parallel_mesh` was broken under HSDP for reasons unrelated to delta.** It returned the 2-D `(dp_replicate, fsdp)` mesh, so `get_data_parallel_rank()` and `get_data_parallel_group()` — which want a single DP domain — raised on it. It now asks `parallel_dims` for the 1-D `loss` mesh, which is that flattened domain. Any HSDP run hit this, with or without delta sync.

HSDP × EP additionally needs a torchtitan-side fix: `to_hf()` reads `placement.dim` off every placement, which a `Replicate` does not have, so the MoE state dict adapter raises during `dcp_load` before training starts. That one is upstream, not in this PR.

**Deliberately out of scope**

- **PP** — each stage holds a disjoint slice, so the export order is not identical across ranks the way the lockstep gather requires. The engine does not support PP for training either. (Megatron's export shows the shape of a fix — a global parameter directory plus zero-count lockstep rows for non-contributing stages — but it would be dead code here until the engine trains under PP.)

**Environment notes**

TorchTitan and SGLang can share one environment as of SGLang main: every SGLang release up to 0.5.17 pins `torch==2.11.0`, while the TorchTitan engine needs `torch.distributed.fsdp.DataParallelMeshDims` (torch ≥ 2.12). SGLang main moves to `torch==2.13.0` with `sglang-kernel==0.4.6.post1`, so torch 2.13 is the meeting point. The runs above also need `spmd_backend=default` rather than the config default `spmd_types`, which under torchtitan main raises `When dp_mesh_dims is provided, all parameters must be DTensors on the full SPMD mesh` — that is torchtitan's newer SPMD path, unrelated to this export.


verl/workers/engine/utils.py
```diff
@@ -213,6 +213,48 @@ def _hf_entry_identity(name, spec, place, lidx, lval):
     return [(name, tuple(spec.full_shape))], str(lval.dtype).replace("torch.", ""), counts, gidx, lval
 
 
+def _hf_entry_row_slots(name, spec, place, lidx, lval):
```
attack204: Explanation: 

Convert TrainSide fused stack
```
experts.w1 : shape (num_experts, *, *)  
```

to rollout side 

```
model.layers.L.mlp.experts.0.w1
model.layers.L.mlp.experts.1.w1
...
model.layers.L.mlp.experts.127.w1
```

verl/workers/engine/spec.py
```diff
@@ -207,13 +258,8 @@ def derive_dtensor_placement(spec: ShardSpec) -> tuple[int | BlockPlacement, boo
         return 0, (dist.get_rank() == 0 if dist.is_initialized() else True), None
 
     placements = spec.placements
-    for p in placements:
-        if type(p).__name__ == "_StridedShard":
```
attack204, line 211: FSDP2 + TP is a _StridedShard, so we should modify the limitation

ChangyiYang:
LGTM!

wuxibin89 (approved):

PR #7340, opened 2026-08-10
[megatron] fix: bugfix qwen 3 qwen 3.5 router replay

EricMarcus-ai:
### What does this PR do?

Fixes a router replay bug for Qwen 3 VL / Qwen 3.5. 

`Qwen3VLGPTModel.__init__` in megatron bridge builds its decoder twice: `super().__init__()` runs `GPTModel.__init__` ([here](https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/d17190fa4d2d023b9b2531c8d449ddb21c13b908/src/megatron/bridge/models/qwen_vl/modelling_qwen3_vl/text_model.py#L63-L83)), which creates `self.decoder = TransformerBlock(...)` and with it every local `MoELayer`/`TopKRouter`; the subclass then discards that block and rebuilds it (`# rebuild the transformer block`, `self.decoder = Qwen3VLTransformerBlock(...)`) ([here](https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/d17190fa4d2d023b9b2531c8d449ddb21c13b908/src/megatron/bridge/models/qwen_vl/modelling_qwen3_vl/text_model.py#L99-L107)). The discarded routers have already appended themselves to the global list and are never removed, leaving 2N entries for N live layers, orphans first.

Since the slice starts at offset 0, it returns exactly those orphans, causing a _silent_ break of the router replay. Targets and the `RouterReplayAction` toggle go to a decoder that no longer forwards; the live routers keep `target_topk_idx = None`, `get_replay_topk` falls through to `default_compute_topk`, and router replay silently does nothing: no error, no warning. The backward recompute never enters `REPLAY_BACKWARD` and reads stale per-microbatch state.

Observed on two runs (pins: megatron-core `d1410e15e`, megatron-bridge `d17190fa`) via the existing `routing replay layers: N` log:

| model | decoder layers | PP | local layers | `len(router_instances)` |
|---|---|---|---|---|
| Qwen3.5-35B-A3B | 40 (`moe_layer_freq=1`) | 1 | 40 | **80** |
| Qwen3.5-122B-A10B | 48 | 2 | 24 | **48** |

Double what they should be.

Fix: address each layer's own router by walking the forwarded model — `isinstance(mod, TopKRouter)` with a non-`None` `mod.router_replay`, indexed by that module's own 1-based `layer_number`. The discarded decoder is unreachable from the model, so the walk only ever reaches live routers, and the result is invariant to how many other routers exist in the process. The positional path is kept for callers that pass no model. An old PR #4567 reported the same symptom on Qwen3-VL but did not fix or merge.

Note it fires now on each microbatch, but it is various microseconds. If you prefer, we can add some memoization; let me know. 

### Checklist Before Starting

- [x] Search for similar PRs: <https://github.com/verl-project/verl/pulls?q=is%3Apr+is%3Aopen+router+replay> — nearest is #4567 (stale), same symptom, not fixed/merged
- [x] PR title follows `[{modules}] {type}: {description}`.

### Test

Small proof-of-concept test that adds 'fake' orphans.

```
$ pytest -q tests/utils/megatron/test_router_replay_model_walk_on_cpu.py
2 passed
```

Reverting only the addressing (walking `RouterReplay.router_instances` instead of the model) fails:

```
E  TypeError: equal(): argument 'input' must be Tensor, not NoneType   # live routers left unset
E  assert False  # ...orphan.router_replay_action is None              # action went to the orphans
2 failed
```

### API and Usage Example

No config or CLI change. Two additive, backward-compatible signatures in `verl/utils/megatron/router_replay_utils.py`:

```python
# new optional `model`; omitting it keeps the previous positional behaviour
set_router_replay_data(layers_topk_idx, attention_mask, tf_config, vp_rank=None, replay_mask=None, model=None)

# new helper, mirrors the module walk used by set_router_replay_data
set_model_router_replay_action(model, router_replay_action)
```

`model` accepts a single module or the list of virtual-pipeline chunks.

### Design & Code Changes

- `verl/utils/megatron/router_replay_utils.py`
  - `iter_model_routers(model)` — yields `(layer_number, RouterReplay)` for replay-enabled `TopKRouter`s in the forwarded model; handles a single module or a VPP chunk list.
  - `set_router_replay_data(..., model=None)` — with a model, writes each layer's targets to its own router, keeping the existing `index_by_layer` / MoE-ordinal row selection so R2 with interleaved dense layers is unaffected; positional path retained for `model=None`.
  - `set_model_router_replay_action(model, action)`.
- `verl/workers/engine/megatron/transformer_impl.py` — three lines in `forward_step`: pass `model=unwrapped_model` to `set_router_replay_data`, and call `set_model_router_replay_action(unwrapped_model, ...)` after each of the two existing action-toggle loops.

`+50/−4` functional changes across the two files (excludes the test).

- [x] Add / Update the documentation — no user-facing behaviour or config change.
- [x] Add unit or end-to-end test(s) 

### DIFF
--- a/verl/utils/megatron/router_replay_utils.py
+++ b/verl/utils/megatron/router_replay_utils.py
@@ -31,6 +31,7 @@ except ImportError:
 from megatron.core import parallel_state as mpu
 from megatron.core.pipeline_parallel.schedules import get_schedule_table
 from megatron.core.tensor_parallel import gather_from_sequence_parallel_region, scatter_to_sequence_parallel_region
+from megatron.core.transformer.moe.router import TopKRouter
 from megatron.core.transformer.transformer_config import TransformerConfig
 from megatron.core.transformer.transformer_layer import get_transformer_layer_offset
 
@@ -371,6 +372,7 @@ def set_router_replay_data(
     vp_rank=None,
     replay_mask=None,
     local_cp_size=None,
+    model=None,
 ):
     """
     Scatter the packed router top-k indices back to sequence-parallel ranks and update each local
@@ -388,6 +390,8 @@ def set_router_replay_data(
             Megatron parallel state will be used.
         replay_mask (Optional[torch.Tensor]): Optional per-token mask. Masked tokens use replayed routes;
             unmasked tokens keep native Megatron routes.
+        model: The forwarded model (or list of VPP chunks). When given, targets are written to its own
+            routers instead of a positional slice of the global RouterReplay.router_instances list.
 
     Returns:
         None: The function updates internal RouterReplay instances in-place.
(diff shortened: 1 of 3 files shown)


HollowMan6:
There are some conflicts against main in this PR that will need to be resolved, would you mind solving it @EricMarcus-ai

EricMarcus-ai:
@HollowMan6 Fixed conflicts

wuxibin89 (approved):

verl/utils/megatron/router_replay_utils.py
```diff
@@ -469,6 +484,33 @@ def set_router_replay_data(
...
+    ``model`` is a single module or the list of virtual-pipeline chunks. The process-global
+    ``RouterReplay.router_instances`` list can hold more routers than this model has local layers —
+    a model that rebuilds its decoder registers a set it then discards — so a positional slice of it
+    can address the wrong objects. Walking the forwarded model reaches each layer's own router.
+    """
+    for chunk in model if isinstance(model, list | tuple) else [model]:
+        for module in chunk.modules():
+            router = getattr(module, "router_replay", None)
+            if isinstance(module, TopKRouter) and router is not None and module.layer_number is not None:
+                yield module.layer_number, router
```
wuxibin89, line 502: @HollowMan6 Will this break MTP+router replay? Since MTP layers have 1-based layer_number.

wuxibin89 (dismissed):

verl/utils/megatron/router_replay_utils.py
HollowMan6: I think currently we don't really have router replay for MTP, as for `get_moe_num_layers_to_build`, it doesn't count `mtp_num_layers`

https://github.com/kaiko-ai/verl/blob/1daab8a6da74295c1fec2b7100fef470bd1fa297/verl/utils/megatron/router_replay_utils.py#L191-L204

But we will need to be careful if later we want to introduce MTP for router replay.

EricMarcus-ai: Good catch, indeed the mtp would have been included. I now scoped it to `decoder`, `self.mtp` is not in there and hence will be excluded, same behaviour as before; added a small test to catch this. 

More generally, for R3, I think MTP router replay is ill-defined. VLLM would speculatively decode and then batch-forward (capturing the 'actual' experts used), and that is what would arrive at the trainer; it won't replay that speculative decode in training either way.

HollowMan6 (approved):
LGTM, thanks!

PR #7344, opened 2026-08-11
[vllm] fix: propagate bucketed receiver failures

nataliekung:
## Status

Closed as unverified.

The public record does not establish a reproducible test result for the current head. This change requires independently reproducible successful transfer, receiver-exception propagation, and receiver-loss timeout coverage before it can be reopened for review.


nataliekung:
Closing as unverified. The available unit environment cannot import this test module without an official vLLM installation, and the attempted H200 runtime matrix did not complete within its timeout. The change must not be presented as ready without reproducible success, receiver-exception, and peer-loss evidence.

PR #7350, opened 2026-08-11
[single_controller] feat: NeoProto ref/index data plane for V0 PPO trainer

alanSquirrelyz:
### What does this PR do?

This PR adds NeoProto, an opt-in ref/index data plane for the synchronous V0
`RayPPOTrainer`. Batch payloads stay in a storage engine while the controller
manipulates schemas, references, and index views, so driver-side batch operations
move refs instead of tensors. Workers materialize only the columns their engine
declares, and the existing TensorDict compute engines are unchanged.

Classic `DataProto` remains the default; NeoProto is enabled with
`trainer.data_plane=neoproto`.

Related RFC: https://github.com/verl-project/verl/issues/6988
Follows: #6999, which removed the TransferQueue runtime path to make room for this.

### Checklist Before Starting

  - https://github.com/verl-project/verl/pulls?q=is%3Apr+NeoProto
  - https://github.com/verl-project/verl/pulls?q=is%3Apr+6988+in%3Abody
  - Note: `gh` is unavailable in this environment, so the duplicate check was done
    through GitHub search instead of the CLI.

### Test

Unit tests:

```bash
pytest -q tests/experimental/neoproto/          # 45 passed
pytest -q tests/test_protocol_on_cpu.py         # 37 passed
```

Full CPU CI suite, run exactly as `cpu_unit_tests.yml` does:

```bash
pytest -o python_files='*_on_cpu.py' --asyncio-mode=auto tests/ \
  --ignore=tests/trainer/ppo/v1 \
  --ignore=tests/trainer/test_multi_trajectories_advantage_on_cpu.py
```

566 passed. 

Deterministic GPU E2E, which runs the existing FSDP + vLLM function-reward case
twice, once with each data plane, and fails closed on any divergence:

```bash
bash tests/special_e2e/ppo_trainer/run_neoproto_correctness.sh
```

It verifies complete training, finite semantic metrics, strict-mode markers,
materialization activity, decoded rollout equivalence, per-step snapshots, and
native checkpoint contents.

Two-node correctness run on 8x H20 (Qwen3-8B, FSDP actor/critic, async vLLM rollout
with TP=2, GAE, 16 prompts/step, 4 responses/prompt, max response length 512):

| Data path | Step 1 (s) | Step 2 (s) | Total (s) | Throughput (tok/s/GPU) |
| DataProto | 68.098 | 72.054 | 140.153 | 68.044 |
| NeoProto | 65.903 | 71.561 | 137.464 | 69.375 |

This is a two-step correctness smoke run with full determinism, correctness dumps,
and checkpoint writing enabled. It is evidence that the tested path does not
regress, not a general performance claim.

Paired performance benchmark on one 8× H20 node (Qwen3-8B, FSDP actor/critic,
async vLLM rollout with TP=2, GAE and KL enabled, 16 prompts/step,
4 responses/prompt, and fixed response length 512):

| Pair | Tokens/path | Classic step (s) | NeoProto step (s) | Classic tok/s/GPU | NeoProto tok/s/GPU | Throughput delta |
| ---: | ---: | ---: | ---: | ---: | ---: | ---: |
| 1 | 320,368 | 25.443 | 25.370 | 196.740 | 197.306 | +0.29% |
| 2 | 307,136 | 25.008 | 25.263 | 191.899 | 189.960 | -1.01% |
| 3 | 308,296 | 25.169 | 24.931 | 191.391 | 193.218 | +0.95% |
| **Mean** | — | **25.207** | **25.188** | **193.343** | **193.495** | **+0.08%** |

Each case completed 10/10 steps; steps 1–2 were discarded as warm-up, leaving
eight measured steps per case. The order was alternated across paired repetitions
(classic→NeoProto, NeoProto→classic, classic→NeoProto), and per-step token counts
were exactly aligned within every pair. Throughput is computed as
`sum(tokens) / sum(step_time) / 8 GPUs` over the measured steps.

**The conclusion is that there is no performance loss with the new NeoProto on a 
single machine, and there is expected to be performance gain with multi-node
cluster with more data movement between the nodes.**

### API and Usage Example

New config field, default unchanged:

```yaml
trainer:
  data_plane: classic  # classic | neoproto
```

```bash
python -m verl.trainer.main_ppo trainer.data_plane=neoproto
```

`trainer.neoproto_strict_mode=true` is available for correctness testing: it requires
ref-table dispatch, rejects forced full materialization, and fails if a worker output
unexpectedly bypasses the NeoProto path.

### Design & Code Changes

The data plane is selected once at trainer startup, so PPO control flow, dispatch,
and workers never branch on which batch implementation is in use.

- `verl/trainer/ppo/data_plane.py` holds the name-to-strategy registry and is the
  only place that reads this configuration. `ClassicPPODataPlane` keeps today's
  behavior; `NeoPPODataPlane` sends ref views and collects results by selecting and
  renaming on the ref table without materializing on the driver.
- `verl/protocol.py` gains default hooks that NeoProto overrides only where behavior
  differs (`new_like`, `prefetch`, `clear_cache`, `prepare_dispatch`,
  `set_control_fields`, `cpu`), so shared code reads identically for both containers.
- `verl/single_controller/base/decorator.py` calls `BatchData.prepare_dispatch`
  unconditionally after chunking. Classic is a no-op; NeoProto attaches per-rank ref
  tables before Ray serializes the chunks.
- `verl/workers/utils/batch_adapter.py` routes worker entry points through
  `run_engine_batch`. TensorDict inputs pass straight through; NeoProto inputs are
  materialized column-wise per the engine's `EngineBatchSpec`. Engine field specs
  live next to the engines that consume them in `verl/workers/engine_workers.py`.
- `verl/experimental/neoproto/` contains the container, storage engine, dispatch
  helper, and worker bridge.

The global `verl.DataProto` class is deliberately not swapped: modules that already
ran `from verl import DataProto` keep the old reference, which would leave two
inconsistent DataProto classes alive in one process.

Three notes for reviewers, since they are more than a mechanical refactor:

#### Why dispatch and worker_bridge exist

A ref-based container meets the tensor-based world at exactly two boundaries, and each
one needs an adapter. These are the only two components the integration adds on top of
the container itself.

**`dispatch` — the Ray serialization boundary.** On the driver a NeoProto column is an
array of N ObjectRefs, where N is the global batch size. Chunking produces one view per
rank, but each view still points at the full array, so pickling it would put all N refs
into every rank's message: `world_size × N` refs on the wire, and every worker holding
handles for rows it will never read. `attach_preserialized_ref_tables` instead builds
one masked ref table per rank containing only that rank's positions, `ray.put`s it, and
stores the handle on the chunk; columns that are not object-ref backed go into a single
shared table that is put once. `NeoProto.__getstate__` then ships just those two
ObjectRefs and `__setstate__` fetches and merges them inside the worker. The per-RPC
payload becomes constant-size instead of scaling with the global batch.

**`worker_bridge` — the compute boundary.** The actor, critic, and reference engines
consume a no-padding TensorDict and know nothing about refs. Two alternatives were
rejected: teaching every engine to read NeoProto would spread an experimental, opt-in
data plane across all compute code, and calling `to_tensordict()` on arrival would
materialize every column on every RPC, which discards the column-level laziness that
motivates the design in the first place. The bridge fetches only the columns that entry
point declares in its `EngineBatchSpec`, converts them to the layout the engine already
expects, calls the unmodified engine, restores padding on the declared outputs, and
rewraps the result as ref-backed. Engine `_impl` code is untouched.

In short: `dispatch` lets the driver move refs instead of tensors, and `worker_bridge`
lets a worker pull only the tensors its engine actually asked for.

- [x] Apply pre-commit checks. `check-naming-conventions` was skipped locally: it
      greps with `--exclude-dir=venv` but not `.venv`, and matches `ServeRLlibRLModule`
      inside a vendored `ray` file.
- [x] Add / Update the documentation. `docs/advance/neoproto_integration.md`.
- [x] Add unit or end-to-end test(s) to the CI workflow.
      `tests/experimental/neoproto/` plus the deterministic E2E in
      `tests/special_e2e/ppo_trainer/run_neoproto_correctness.sh`.

AI assistance was used for this change and for this PR description.


zw0610 (approved):
/LGTM

PR #7369, opened 2026-08-12
[sglang] fix: make DeepSeek-V4 full weight sync work with SGLang + Megatron

ChangyiYang:
### What does this PR do?

Makes full weight sync work for DeepSeek-V4 with an SGLang rollout and a Megatron trainer.

Two failures blocked the path:

1. The SGLang environment currently pins a Transformers version that does not register `deepseek_v4`. In SGLang-only environments, verl now registers a compatibility wrapper around SGLang's config class. The wrapper aliases newer public checkpoint fields (`head_dim`, `sliding_window`, and `num_hash_layers`) and preserves metadata that SGLang 0.5.12 does not declare yet.
2. SGLang rebuilds several DSv4 destination parameters by concatenating two incoming tensors inside one `load_weights` call. A normal byte-based bucket boundary could split the pair and trigger SGLang's non-empty loader-cache assertion. The bucketer now keeps each pair together.

Fusion grouping is opt-in and is passed only when `model_type == "deepseek_v4"`; other SGLang model streams retain the original bucketing behavior. The mapping covers Megatron and HF/FSDP attention names, plus native-FP8 `.scale` and requantized `.weight_scale_inv` names.

### Test

- Focused CPU coverage for all ten DSv4 fusion groups, the non-DSv4 no-op path, and incomplete-group failure.
- Existing SGLang bucketer regression tests: 20 focused tests passed together.
- End-to-end acceptance on 40 B300 GPUs: four 8-GPU Megatron trainer nodes plus one 8-GPU SGLang rollout node, DeepSeek-V4-Flash-FP8, five real GSM8K training steps, and full NCCL weight sync after every actor update.
  - Slurm job 11203 completed with exit code 0.
  - All 5/5 steps completed with `response/aborted_ratio=0.0`.
  - Full-sync times were 412.84s, 417.18s, 442.27s, 437.06s, and 437.11s.

### API and compatibility

No user-facing API or configuration change. Native Transformers support takes precedence when available; vLLM remains the first fallback when installed; the SGLang config adapter is used only when those paths cannot resolve DeepSeek-V4.

### Checklist

- [x] Added CPU tests
- [x] Ran focused existing bucketer tests
- [x] Ran multi-node end-to-end validation
- [x] No documentation update needed; no user-facing surface changed

### DIFF
--- a/verl/workers/config/model.py
+++ b/verl/workers/config/model.py
@@ -12,10 +12,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 from dataclasses import dataclass, field
+from inspect import signature
 from typing import Any, Optional
 
 from omegaconf import MISSING
-from transformers import AutoConfig
+from transformers import AutoConfig, PretrainedConfig
 
 from verl.base_config import BaseConfig
 from verl.utils import hf_processor, hf_tokenizer
@@ -24,6 +25,30 @@ from verl.utils.import_utils import import_external_libs
 from verl.utils.model import get_generation_config, update_model_config
 
 __all__ = ["HFModelConfig", "MtpConfig"]
+
+
+def _get_sglang_deepseek_v4_config_class():
+    """Return SGLang's config with aliases for newer public DSv4 checkpoints."""
+    from sglang.srt.configs.deepseek_v4 import DeepSeekV4Config
+
+    class DeepSeekV4ConfigCompat(DeepSeekV4Config):
+        def __init__(self, **kwargs):
+            kwargs = dict(kwargs)
+            for source, target in {
+                "head_dim": "v_head_dim",
+                "sliding_window": "window_size",
+                "num_hash_layers": "n_hash_layers",
+            }.items():
+                if source in kwargs and target not in kwargs:
+                    kwargs[target] = kwargs[source]
+
+            accepted = set(signature(DeepSeekV4Config).parameters)
+            super().__init__(**{key: value for key, value in kwargs.items() if key in accepted})
+            for key, value in kwargs.items():
+                if key not in accepted:
+                    setattr(self, key, value)
+
+    return DeepSeekV4ConfigCompat
 
 
 @dataclass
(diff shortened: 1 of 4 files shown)


ChangyiYang:
Trimmed the scope: dropped the two `trainer_separate_async.py` commits that masked the hybrid replicas' weight sync. That change is gated behind an opt-in env var and defaults to off, and the sync it skips happens once at init and feeds an engine that never generates — so it is an optimization, not something the full sync needs to work. It does not belong in a PR about making the path work, and it would have added an env var to the public surface for no benefit to this fix. Now 3 commits / 3 files / +93 lines, all of them required.

ChangyiYang:
Added a fourth commit that the PR was missing: fused-param grouping in the plain full-sync bucketer. Without it the path still hits `assert len(cache_wqkv_a_weight) == 0` in SGLang's DSv4 loader, because a bucket boundary can fall between the two halves of a fused destination. That stack has no custom-loader frame in it — it is the plain full-sync path — so the other three commits could not help it. Now 4 commits / 5 files / +232 lines.

verl/workers/config/model.py
```diff
@@ -193,13 +193,25 @@ def __post_init__(self):
...
-            from vllm.transformers_utils.config import get_config
+            try:
+                from vllm.transformers_utils.config import get_config
 
-            self.hf_config = get_config(
-                self.local_hf_config_path,
-                trust_remote_code=self.trust_remote_code,
-                attn_implementation=attn_implementation,
-            )
+                self.hf_config = get_config(
```
wuxibin89: Verify on transformers==5.9.0, it load config correctly, so both vllm/sglang get_config fallback is not needed.
```
DeepSeek-V4-Flash [master] $ pip list | grep transformers
transformers                             5.9.0
x-transformers                           2.24.0
DeepSeek-V4-Flash [master] $ python3
Python 3.12.3 (main, Jun 19 2026, 12:46:00) [GCC 13.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from transformers import AutoConfig
>>> hf_config = AutoConfig.from_pretrained("/home/joel/DeepSeek-V4-Flash")
[transformers] Unrecognized keys in `rope_parameters` for 'rope_type'='default': {'attention_factor'}
>>> hf_config
[transformers] Unrecognized keys in `rope_parameters` for 'rope_type'='default': {'attention_factor'}
DeepseekV4Config {
  "architectures": [
    "DeepseekV4ForCausalLM"
  ],
  "attention_bias": false,
  "attention_dropout": 0.0,
  "bos_token_id": 0,
  "compress_rates": {
    "compressed_sparse_attention": 4,
    "heavily_compressed_attention": 128
  },
```

ChangyiYang:
Added a fifth commit and re-validated. Running the previous revision on the PR branch alone still hit `assert len(cache_wqkv_a_weight) == 0`: the fusion table listed only `weight_scale_inv`, while DSv4's native fp8 checkpoint names its scales `.scale` — so the weight halves paired and the scale halves silently did not. Both spellings are now in the table. With that, the branch completes its init-time full sync and the rollout serves generation from those weights. Worth noting the earlier evidence came from a larger downstream branch where a later commit was covering this gap, which is why it read as working.

ChangyiYang:
Validation run finished: `timing_s/update_weights = 215.80 s` on this branch's own tree, with the rollout serving generation from the synced weights. I have replaced the timing figures in the description — the previous ones were measured on the larger downstream branch, and after finding that the downstream tree was masking a real gap in this PR, quoting its numbers here was not defensible. The downstream figure is still listed for reference, explicitly labelled, with no claim attached to the difference.

verl/workers/config/model.py
ChangyiYang: Thanks — agreed that Transformers 5.9 resolves DeepSeek-V4 natively. The current SGLang environment on main, however, pins `transformers==5.3.0` (`pyproject.toml` line 118: https://github.com/verl-project/verl/blob/9c7643648f1e1e109dd30da2da3cf94e1317b229/pyproject.toml#L118), and the SGLang-only image has no vLLM fallback. In that exact stack, the public DeepSeek-V4-Flash config first failed on unsupported metadata (`expert_dtype`) and then on the newer `head_dim` field; the compatibility wrapper now aliases the three renamed fields and preserves unknown metadata. With it, job 11203 completed 5/5 GSM8K training steps with full NCCL sync after every update. The fallback remains narrowly gated: it runs only after native `AutoConfig` fails, verifies raw `model_type == "deepseek_v4"`, and then prefers vLLM if installed. Once the SGLang pin moves to a Transformers release with native DSv4 support, this branch is not reached and can be removed.

wuxibin89: We will bump transformers to 5.3.0, actually I'm testing with transformers==5.9.0, so this change is not needed.

wuxibin89 (approved):

PR #7349, opened 2026-08-11
[fsdp] fix: preserve mRoPE token axis in no-padding conversion

Zhaoyi-Tian:
Use the existing explicit jagged-layout helper when packing multidimensional position IDs, and cover singleton and equal-length batches.

### What does this PR do?

`left_right_2_no_padding()` converts padded position IDs into a jagged `NestedTensor`. For multidimensional position IDs, each sample has shape `[rope_dim, seq_len]`, and the token dimension must remain the ragged dimension.

The previous implementation used `torch.nested.as_nested_tensor(...)`, which infers the ragged dimension from the input shapes. For a singleton batch or a batch whose samples have equal valid sequence lengths, this inference selects the RoPE dimension instead of the token dimension. The resulting tensor has incorrect ragged metadata, offsets, and, for multi-sample batches, values layout.

This PR replaces that ambiguous construction with the existing `nested_tensor_from_tensor_list()` helper, which explicitly preserves the last sample dimension as the token/ragged dimension. One-dimensional position IDs continue to use the helper's existing one-dimensional path.

This is a follow-up to #6127 and #6073. #6127 fixed the same equal-length jagged-layout ambiguity when selecting, chunking, concatenating, or otherwise rebuilding an existing `NestedTensor`, and introduced the helper used here. It did not update the initial padded-to-no-padding conversion in `left_right_2_no_padding()`. #6149 subsequently generalized that helper to support non-last ragged dimensions. This PR covers the remaining initial-construction call site and does not duplicate those changes.

Related work:

- #6073
- #6127
- #6149
- #4860 only restores `_ragged_idx` metadata after serialization; it cannot reconstruct values and offsets that were packed along the wrong dimension.

### Checklist Before Starting

- [x] Searched for similar PRs:
  - [`left_right_2_no_padding`](https://github.com/verl-project/verl/pulls?q=is%3Apr+is%3Aopen+left_right_2_no_padding)
  - [`position_ids_nested`](https://github.com/verl-project/verl/pulls?q=is%3Apr+is%3Aopen+position_ids_nested)
  - [`mRoPE position_ids ragged`](https://github.com/verl-project/verl/pulls?q=is%3Apr+is%3Aopen+mRoPE+position_ids+ragged)
  - No open PR was found that fixes this initial conversion call site.
- [x] Formatted the PR title as `[{modules}] {type}: {description}`.

### Test

The regression test exercises the production entry point and covers both ambiguous cases:

- `batch_size=1`;
[...]

PR #7370, opened 2026-08-12
[fsdp] feat: project only active rows in the Torch fused LM head

Zhaoyi-Tian:
`[fsdp] feat: project only active rows in the Torch fused LM head`

Use the existing response mask to skip vocabulary projection for rows that do
not contribute to the policy loss, while preserving the packed output layout
expected by downstream PPO code.

### What does this PR do?

With remove-padding and the Torch fused kernel, the FSDP actor currently applies
the vocabulary-sized LM head to every packed token. In agent trajectories, many
of those tokens belong to prompts or environment observations and are excluded
from the policy loss by `response_mask`.

This PR uses `response_mask` to mark the corresponding causal predictor labels
with the standard `-100` ignore index. `FusedLinearForPPO` projects only the
remaining hidden-state rows, then scatters log-probabilities and entropy back to
the original packed shape. Active-row values and gradients are preserved;
ignored output positions are zero-filled and remain excluded by the existing
loss mask.

The intended scope of this draft is the FSDP/FSDP2 language-model path with
remove-padding and the Torch fused backend; the performance validation below
uses FSDP2. We kept the implementation deliberately narrow: it introduces no
public option, batch field, or model-specific adapter. Eager, Triton, and
value-model paths continue to use their existing behavior. The same idea may be
applicable to other paths in follow-up work.

Related implementations use the same row-selection principle:

- Transformers supports a tensor-valued
  [`logits_to_keep`](https://github.com/huggingface/transformers/blob/cc7ab9be508ce6ed3637bba9e50367b29b742dc6/src/transformers/models/qwen3_5/modeling_qwen3_5.py#L1799-L1837)
  to select hidden-state rows before the LM-head projection.
- [TRL](https://github.com/huggingface/trl/blob/24db2a825029f7e0e60f6d3c160485565a784503/trl/trainer/sft_trainer.py#L1546-L1565)
  maps tokens excluded by assistant masks to `-100` labels, while
  [`loss_type="chunked_nll"`](https://github.com/huggingface/trl/blob/24db2a825029f7e0e60f6d3c160485565a784503/trl/trainer/sft_trainer.py#L117-L207)
  drops ignored rows before the LM-head matmul. Its loss implementation performs
  the causal shift before filtering `labels != -100`.

Here, `-100` is written only to target-aligned `shift_labels`, not model
`input_ids`, and ignored rows are removed before the underlying fused operation
uses labels as vocabulary indices. Unlike TRL's scalar SFT loss, PPO still
needs token-aligned log-probability and entropy tensors, so this PR scatters the
selected results back to the existing packed layout. The tests below verify
forward values and hidden-state/LM-head gradients for dense, sparse, and fully
masked labels.

### Checklist Before Starting

- [x] Searched for similar open PRs: [`FusedLinearForPPO`](https://github.com/verl-project/verl/pulls?q=is%3Apr+is%3Aopen+FusedLinearForPPO). No matching implementation was found at the time of writing.
- [x] Formatted the title as `[{modules}] {type}: {description}`.

### Test

CPU forward/backward parity and engine label-alignment tests:

```bash
python -m pytest -q \
  tests/utils/test_fused_linear_for_ppo_on_cpu.py \
  tests/workers/test_packed_pad_to_length_on_cpu.py
```

Result: `24 passed`.

The tests cover dense, sparse, and fully masked labels; log-probability,
entropy, hidden-state-gradient, and LM-head-gradient parity; causal alignment;
packed padding; and unchanged eager/value-model behavior.

Two-GPU Ulysses SP regression:

```bash
CUDA_VISIBLE_DEVICES=6,7 torchrun --standalone --nproc_per_node=2 \
  -m pytest -q tests/special_distributed/test_fused_kernels_ulysses_sp.py
```

Result: `1 passed` on each rank. The test covers dense and sparse SP=2 output
parity, odd-length SP padding, and a rank-local shard with no active rows.

Repository checks:

```bash
```

Result: all hooks passed.

#### Performance

We compared baseline commit `04f25bb4` with this patch (`eadb0a28`) on 8 x
NVIDIA A800-SXM4-80GB GPUs using Qwen3.5-4B and Qwen3.5-9B. Both runs used V1
FSDP2 (`fsdp_size=8`, `ulysses_sp_size=1`) with `model_dtype=fp32`,
`param_dtype=bf16`, and `reduce_dtype=fp32`, plus remove-padding, the Torch
fused backend, gradient checkpointing, and AdamW.

Clean timing used 5 warmup steps and 20 synchronized measurements per ratio.
The measured region was one optimizer step (PPO forward/backward, FSDP2
synchronization, gradient clipping, AdamW, and LR scheduling), excluding
rollout, Ray, and data transfer. A separate run used 5 warmups and 3 profiled
steps for attribution only. For clean optimizer-step and profiled F+B latency,
each table value is the mean of the ascending-run and descending-run medians.
Fused-head CUDA time is mean per-step device time, averaged across the same two
runs. Each entry is `baseline -> this PR`, and wall-time samples use the slowest
rank. The crossed ascending/descending runs produced consistent results.

| Model | Active | Clean optimizer step (ms) | Reduction | Profiled F+B latency (ms) | Fused-head CUDA time (ms) |
|---|---:|---:|---:|---:|---:|
| 4B | 10% | 961.97 -> 881.06 | 8.41% | 1078.09 -> 1015.13 | 91.20 -> 12.85 |
| 4B | 25% | 961.13 -> 892.34 | 7.16% | 1084.57 -> 1028.72 | 91.25 -> 23.71 |
| 4B | 50% | 962.00 -> 917.11 | 4.67% | 1101.74 -> 1054.78 | 91.18 -> 46.24 |
| 4B | 75% | 960.70 -> 937.42 | 2.42% | 1092.85 -> 1072.92 | 91.27 -> 68.82 |
| 4B | 100% | 960.70 -> 959.83 | 0.09% | 1092.67 -> 1087.00 | 91.49 -> 91.40 |
| 9B | 10% | 2025.47 -> 1359.02 | 32.90% | 2080.23 -> 1400.05 | 766.78 -> 92.86 |
| 9B | 25% | 2025.69 -> 1458.70 | 27.99% | 2085.20 -> 1519.94 | 766.87 -> 194.26 |
| 9B | 50% | 2029.52 -> 1647.68 | 18.81% | 2078.89 -> 1699.58 | 766.90 -> 385.25 |
| 9B | 75% | 2028.55 -> 1839.62 | 9.31% | 2084.33 -> 1898.10 | 767.07 -> 576.12 |
| 9B | 100% | 2029.03 -> 2032.19 | -0.16% | 2091.86 -> 2082.39 | 767.09 -> 767.06 |

The clean 100% controls show no material fixed overhead. The profiler numbers
include profiling overhead and are not used to claim optimizer-step speedup;
they show that the ratio-dependent change is concentrated in the fused LM head.
The profiled optimizer stage remained 66--71 ms for 4B and 98--103 ms for 9B
without a ratio-dependent trend.

At 10% active targets, the maximum per-rank allocated-memory peak decreased
from 13.46 to 11.97 GiB for 4B and from 31.34 to 26.35 GiB for 9B. At 50% and
above, the measured whole-step allocation peak was effectively unchanged.

<details>
<summary>Benchmark setup</summary>

Each rank received two deterministic, synthetic
1024-token sequences (`global_batch_size=16`). Token IDs were consecutive and
rank-specific, starting at `100 + rank * 2048` and taken modulo
`min(vocab_size, 100000)`. The first token was the prompt and the remaining
1023 were response targets. `response_mask == loss_mask` selected the first
`round(128 * ratio)` targets in each consecutive 128-token block, corresponding
to 208, 512, 1024, 1536, and 2046 active targets per rank for the requested
10%, 25%, 50%, 75%, and 100% cases. Each selected response target was aligned
with its preceding causal predictor row. Advantages were one, temperature was
1.0, and old log-probabilities were computed by the same initialized actor
before timing. The learning rate was zero to avoid weight drift while retaining
gradient clipping and AdamW computation; KL and entropy losses were disabled.

</details>

These measurements cover the tested models and configuration only; they are
not presented as end-to-end RL iteration speedups.

### API and Usage Example

No public API or configuration change is introduced. The implementation reuses
the following existing FSDP/FSDP2 language-model actor settings and
`response_mask`:

```yaml
actor_rollout_ref:
  model:
    use_remove_padding: true
    use_fused_kernels: true
    fused_kernel_options:
      impl_backend: torch
  actor:
    strategy: fsdp2
```

### Design & Code Changes

- `verl/workers/engine/fsdp/transformer_impl.py`
  - Align response targets with their causal predictor rows.
  - Preserve the `-100` ignore index through packed and Ulysses padding.
- `verl/utils/experimental/torch_functional.py`
  - Select active rows before the existing fused LM head.
  - Restore the original packed output shape after projection.
- Tests
  - Add CPU numerical/gradient and engine-alignment coverage.
  - Extend the existing Ulysses SP regression with sparse and empty-rank cases.

> [!IMPORTANT]

- [x] Applied [pre-commit checks](https://github.com/verl-project/verl/blob/main/CONTRIBUTING.md#code-linting-and-formatting): `pre-commit run --all-files --show-diff-on-failure --color=always`.
- [x] Documentation is unchanged because this is an internal optimization with no public API or configuration change.
- [x] Added CPU tests collected by the existing CPU CI and extended the existing distributed SP regression; no workflow YAML change is needed.
- [x] Request CI in the verl Slack `ci-request` channel or the Feishu group when the draft is ready for review.
- [x] Recipe submodule update is not applicable.

### AI Assistance Disclosure

OpenAI Codex assisted with analysis, implementation, test design, benchmarking,
and drafting. The human submitter reviewed the changes and is responsible for
the submission.

Assisted-by: OpenAI Codex


Zhaoyi-Tian:
Hi @wuxibin89, thank you again for pointing out the issue with my previous PR #7349.

This new PR #7370 came from a potential optimization I happened to notice while tuning an Agent RL workload.

The currently assigned reviewers seem to have had limited review activity recently, so I’m slightly worried that the PR might otherwise slip through the cracks. I’m also not quite sure who would be the most appropriate person to ask about this kind of change. If you have a moment, could you help approve the CI run or point me to a more suitable reviewer?

The optimization seems to provide some speedup in my benchmarks. I tried to keep the changes minimal and transparent to existing users, though I’m not sure whether this kind of optimization fits verl’s design preferences or whether the added complexity would be worth the maintenance cost.

It looks like the same idea might also apply to a few other paths, but I kept this PR focused for now to make the change easier to review and evaluate.

Thanks again for your time and help!

PR #6556, opened 2026-06-01
[fully_async] feat: support dynamic resource scheduling

zpltys:
### What does this PR do?

This PR introduces **hybrid inference resource dynamic scaling** for the fully-async PPO training framework, addressing suboptimal GPU utilization where Trainer-node GPUs sit idle waiting for rollout data while Standalone Rollout nodes idle during training.

### Core Design

A **dual-mode rollout resource** architecture:

| Resource Type | Location | Behaviour |
|---|---|---|
| **Standalone replicas** | Dedicated rollout nodes | Continuously perform rollout across all steps |
| **Hybrid replicas** | Trainer-node GPUs (co-located) | Dynamically switch between *Rollout* and *Training* |

A new `DynamicResourceController` manages the lifecycle (`STANDALONE_ONLY ↔ HYBRID_ACTIVE`) driven by a pluggable `DynamicScalingPolicyBase`:

- **`DefaultDynamicScalingPolicy`** — adaptive `deactivate_ratio` that adjusts based on trainer wait time and sample buffer status
- **`StaticFullyAsyncPolicy`** — baseline-equivalent policy (immediate deactivation, never re-activate)

### Key Mechanisms

1. **Conditional weight sync** — always broadcast to Standalone; only sync to Hybrid when the policy decides to activate (saves communication overhead)
2. **Deactivation (Hybrid → Trainer)** — in-flight requests aborted & redistributed to Standalone via LoadBalancer
3. **Activation (Trainer → Hybrid)** — optional `dynamic_scaling_enable_rebalance` triggers request reshuffling via least-loaded routing
4. **Separate memory knobs** — `gpu_memory_utilization` (Hybrid, low) vs `standalone_gpu_memory_utilization` (Standalone, high) added to both FSDP & Megatron configs

### Experimental Validation

Training **Qwen3.5-35B-A3B** on **DAPO-Math-17K** with **H20** GPUs:

| Config | GPUs | Result |
|---|---|---|
| Baseline | 16 train + 16 rollout (4 nodes) | 6.546 h |
| Baseline | 8 train + 24 rollout (4 nodes) | 6.379 h |
| **Dynamic scaling** |16 train + 16 rollout (4 nodes) | **5.4 h**, reduce **~15.3%** wall-clock time  |

The per-step runtime plot confirms dynamic scaling reduces step latency by leveraging Trainer GPUs during rollout idle windows, with the gap widening as the adaptive policy tunes `deactivate_ratio`.

### Checklist Before Starting

[...]

PR #7373, opened 2026-08-12
[trainer, vllm] feat: lend idle trainer GPUs to generation in separate_async

Begunner:
### What does this PR do?

Allow the hybrid GPUs to switch to the rollout mode in separate_async trainer to reduce trainer idle.

Once the prompts for a step have been submitted, the trainer keeps its replicas in rollout mode and
helps generate until the replay buffer holds enough sampleable groups, then reclaims the GPUs and
trains. We use an adaptive threshold tracks how starved the trainer actually is.

The threshold-adaptation idea is adapted from the dynamic resource scheduling introduced for
`fully_async` in #6556; this PR applies the same principle at the step boundary of the
`separate_async` trainer.

Gated behind `trainer.v1.separate_async.enable_switch`, **off by default**. With it off the
behaviour is unchanged.

This PR also uses `sleep(level=2) + wake_up(tags=["weights"])` to simulate 
`release_kv_cache()` / `resume_kv_cache()` for the vLLM server, otherwise OOM when updating 
standalone weights. Once vllm-project/vllm#44890 (`release_kv_cache()` API, itself waiting on 
vllm-project/vllm#46438 `CuMemAllocator.discard()`) lands, this can be replaced by a direct
tag-selective discard, which will also save the weight remap. A `TODO` records this.

#### Setup

Two end-to-end runs differing **only** in `trainer.v1.separate_async.enable_switch`.

| | config|
|---|---|
| Hardware | 3 × 8 H100 — 2 nodes hybrid (trainer) + 1 node standalone rollout |
| Model | Qwen3.5-35B-A3B |
| Data | DAPO-Math-17k |
| Trainer | Megatron, TP=2, PP=2, CP=2, EP=8, ETP=1 |
| Rollout | vLLM, TP=4, `gpu_memory_utilization=0.7`, `n=8` |
| Batching | `train_batch_size=64`, `ppo_mini_batch_size=16`, `parameter_sync_step=4` |
| Lengths | `max_prompt_length=2048`, `max_response_length=32768` |

Both runs went well past 150 steps; **all numbers and plots below are the first 150 steps of each.**

#### Result: 12.6% less wall clock, 10.1% more tokens/s

| | enable_switch | baseline | |
|---|---|---|---|
| wall clock for 150 steps | **16.43 h** | 18.80 h | **−12.6%** |
| tokens / s (total tokens ÷ total wall clock) | **16 687** | 15 150 | **+10.1%** |
| `perf/throughput` (arithmetic mean) | 712.3 | 663.8 | +7.3% |

fig1_gen_and_step

`switch_wait` is **not** new overhead: during that window the trainer's GPUs are generating, whereas
the `gen` time it replaces was pure idling.

#### Training quality

fig4_quality

| | enable_switch | baseline |
|---|---|---|
| `critic/score/mean` | 0.7888 | 0.7878 |
| `training/off_policy/trajectory_staleness/mean` | 0.556 | 0.522 |
| `training/off_policy/trajectory_spans/mean` | 1.436 | 1.470 |

The reward curves are almost the same. Staleness rises 6.5% in the mean, which is expected — when the `switch_threshold_ratio` is higher than perfect, `update_weights` will be delayed.

#### Throughput over time

fig3_throughput

#### Unit tests

```bash
pytest tests/trainer/ppo/v1/test_separate_async_step_switch_on_cpu.py -v
pytest tests/trainer/ppo/v1/test_replay_buffer_on_cpu.py -v
```

#### Limitations

- Only validated on the 2-hybrid + 1-standalone ratio above. That ratio is favourable to this
  feature — the trainer owns 2/3 of the GPUs, so there is a lot of idle capacity to reclaim. Expect
  smaller gains as standalone rollout capacity grows.
- The current implementation is inherently sticky: every decision extrapolates from measurements
   taken on previous steps, which assumes the sample production rate stays roughly constant from
   one step to the next. If generation throughput oscillates sharply between steps, the benefit degrades.
- `enable_switch` is rejected together with rollout disaggregation (PD).

### API and Usage Example

New keys under `trainer.v1.separate_async`, all inert unless `enable_switch=true`:

```yaml
trainer:
  v1:
    separate_async:
      # Allow the trainer to switch to rollout mode to reduce idle.
      enable_switch: false

      # Fraction of train_batch_size that must be sampleable before the rollout replicas are
      # reclaimed.
      switch_threshold_ratio: 0.4

      # Optional threshold adaptation. Starvation raises the ratio immediately; calm steps lower it
      # only after the configured release interval.
      adaptive_switch_threshold: true
      switch_threshold_step_up: 0.05
      switch_threshold_step_down: 0.03

      # Consecutive idle/calm steps required before ratio adjusts in that direction.
      switch_threshold_release_steps: 2

      # Window size for calculating average switch cost.
      switch_cost_window_size: 3
```


Begunner:
CC @zpltys

verl/trainer/ppo/v1/trainer_separate_async.py
```diff
@@ -148,44 +201,164 @@ def on_validate_begin(self):
...
+        poll_interval = getattr(self.replay_buffer, "poll_interval", 2.0)
+        return self._step_sample_wait_seconds > poll_interval
+
+    def _adapt_switch_threshold(self, had_idle: bool) -> None:
+        if had_idle:
+            self._calm_steps = 0
+            self._idle_steps = min(self._idle_steps + 1, self._threshold_release_steps)
+            if self._idle_steps < self._threshold_release_steps:
+                return
+            self._switch_threshold_ratio = min(1.0, self._switch_threshold_ratio + self._threshold_step_up)
```
zpltys: add statistic metric for self._switch_threshold_ratio

docs/advance/v1_async_trainer.md
```diff
@@ -0,0 +1,300 @@
+# V1 Async Trainer
```
wuxibin89, line 1: Better move to `docs`.

Begunner: Oh I forgot to move.
Done.

verl/trainer/ppo/v1/trainer_base.py
```diff
@@ -611,6 +614,15 @@ def on_step_begin(self):
         """Called at the beginning of each training step."""
         return
 
+    def on_step_prompts_submitted(self) -> dict:
```
wuxibin89: This hook is not general, it's better that PPOTrainerSeparateAsync override `_add_batch_to_generate`
```python
@register_trainer("separate_async")
class PPOTrainerSeparateAsync(PPOTrainer):
    def _add_batch_to_generate(self):
        super()._add_batch_to_generate()
        self.on_step_prompts_submitted()
```

Begunner: > 

Get

Begunner: _add_batch_to_generate() is also used for the warmup_batch. Maybe add hook like this?

```
# trainer_base.py
def _prepare_step(self) -> dict:
    self._add_batch_to_generate()
    return {}

# trainer_separate_async.py
def _prepare_step(self) -> dict:
    super()._prepare_step()
    return self._wait_for_sampleable_and_switch()
```

verl/trainer/config/ppo_trainer.yaml
```diff
@@ -239,6 +239,28 @@ trainer:
     # Asynchronous PPO trainer with separate trainer and rollout
     separate_async:
 
+      # Allow the trainer to switch to rollout mode to reduce idle.
```
wuxibin89: Better group switch related parameter to a class, e.g:
```
separate_async:
    hybrid_rollout:
        _target_: verl.trainer.config.HybridRolloutSwitchConfig
        enable_switch: false
        ...
```

Begunner: Done

verl/trainer/ppo/v1/trainer_separate_async.py
```diff
@@ -78,6 +82,59 @@ def _init_resource_pool_mgr(self):
...
+        switch_config: HybridRolloutSwitchConfig = omega_conf_to_dataclass(
+            self.config.trainer.v1.separate_async.hybrid_rollout
+        )
+        self._enable_switch = bool(switch_config.enable_switch)
+        if self._enable_switch:
+            self._init_switch_config(switch_config)
+
+    def _init_switch_config(self, switch_config: HybridRolloutSwitchConfig) -> None:
+        """Validate the step-switch settings against the replay buffer that will serve them."""
+        self._switch_threshold_ratio = float(switch_config.switch_threshold_ratio)
```
wuxibin89: No need to resolve again since hydra already resolve it.

wuxibin89 (dismissed):

verl/trainer/ppo/v1/trainer_separate_async.py
Begunner: Get

verl/trainer/ppo/v1/trainer_separate_async.py
```diff
@@ -78,6 +82,59 @@ def _init_resource_pool_mgr(self):
...
+        self._idle_steps = 0
+        self._calm_steps = 0
+        self._step_sample_wait_seconds = 0.0
+        self._step_wait_samples = 0
+        self._sample_start = time.perf_counter()
+        self._step_threshold = 0
+        self._wait_seconds = 0.0
+        self._wait_samples = 0
+        switch_cost_window_size = int(switch_config.switch_cost_window_size)
+        if switch_cost_window_size <= 0:
```
wuxibin89: Better move sanity check into `HybridRolloutSwitchConfig.__post_init__`

wuxibin89 (approved):

PR #7382, opened 2026-08-13
[rollout, vllm, sglang] fix: fail fast when a rollout server actor is unschedulable

### LINKED ISSUE #7381: hybrid rollout: LLMServerManager.create() hangs indefinitely when a rollout replica's colocated vLLM server actor is unschedulable (no timeout/diagnostic)
### Summary

When a hybrid-colocated rollout replica's per-node vLLM server actor cannot be scheduled, `LLMServerManager.create()` awaits it forever. There is no timeout and no diagnostic, so the whole training job silently hangs at first rollout-engine init until an external watchdog tears it down (~18 min in our case).

### Environment / repro shape

- Hybrid colocate rollout (rollout engine fused with FSDP/actor in the worker process).
- Layout: 1 GPU per node x 2 nodes, so 2 rollout replicas, one per node.
- 7B model. A 0.5B run with the same variant/layout comes up fine; the 7B run hangs at first init. This points to a placement race / fragility under tight 1-GPU-per-node packing rather than a deterministic bug (see "What I could not close" below).

### What hangs (py-spy on the driver TaskRunner, single async thread)

The driver is parked in:

```
LLMServerManager.create()
  -> LLMServerManager._initialize_llm_servers()
     -> await asyncio.gather(*[server.init_hybrid(self.worker_group) for server in self.rollout_replicas])
```

(current main: `verl/workers/rollout/llm_server.py:567`, reached from `create` at `llm_server.py:511-514`.)

Two py-spy samples ~10 min apart are frame-identical: no forward progress.

The coroutine that never returns is one replica's `init_hybrid` -> `launch_servers`, blocked on the first actor method call issued after the server actor is created:

```
vLLMReplica.launch_servers()
  -> await self.servers[0].get_master_address.remote()
```

(`verl/workers/rollout/vllm_rollout/vllm_async_server.py:1181`; the server actor is created just above at `:1159-1177` with `NodeAffinitySchedulingStrategy(node_id=..., soft=False)`.)

### Ray actor snapshot (`ray list actors`)

- `vllm_server_1_0` (replica 1): ALIVE on node 1.
- `vllm_server_0_0` (replica 0): PENDING_CREATION, never scheduled.
- driver + 8x SimpleStorageUnit + TransferQueueController + reward actor: all packed onto node 0, whose single 1-GPU bundle is fully occupied. Replica 0's server actor is hard-pinned to that node via `NodeAffinitySchedulingStrategy(soft=False)`, so it has nowhere to land.

Because a `.remote()` call on a PENDING_CREATION actor blocks with no timeout, the `asyncio.gather` above awaits forever.

### This is not a cross-node weight-sync deadlock

Ruled out from logs:

- Both ranks reach NCCL `Init COMPLETE` / `Connected all rings`.
- No process is inside a collective.
- No NCCL timeout / retry.

The hang is purely at rollout-server scheduling/init, before any weight sync.

### User-visible impact

Black-box hang: no error, no log line, no timeout. The job holds its allocated GPUs doing nothing until an external watchdog kills it.

### What I could not close (honest gaps)

- I did not capture `required_resources` of `vllm_server_0_0` before the job was stopped, so I cannot show the exact resource shortfall numerically.
- 0.5B (same variant/layout) comes up while 7B hangs, so I describe this as a placement race / fragility under tight packing, not a deterministic scheduling bug.

### Proposed direction (fail-fast instead of infinite await)

Bound the wait for rollout server actors to become schedulable. If a server actor stays `PENDING_CREATION` (or `DEPENDENCIES_UNREADY`) past a deadline, raise a clear error naming the actor, its state, and its `required_resources`/node, instead of awaiting forever. This turns an ~18-min silent hang into an immediate, actionable failure, and it does not depend on capturing `required_resources` by hand because Ray's actor state exposes it at failure time.

I have a small patch for this and will open a PR that references this issue.

nataliekung:
### What does this PR do?

Fail fast when a rollout server actor cannot be scheduled, instead of hanging forever.

A hybrid/colocated rollout replica creates its per-node server actor with a hard
`NodeAffinitySchedulingStrategy(soft=False)` pinned to its training worker's node. If that
node has no free slot, the actor stays `PENDING_CREATION` and the first `.remote()` call on
it (e.g. `get_master_address`) blocks forever, so `LLMServerManager.create()` hangs at first
init with no timeout and no diagnostic. The whole job then holds its GPUs doing nothing until
an external watchdog kills it.

Closes #7381.

### Checklist Before Starting

- [x] Search for similar PRs: https://github.com/verl-project/verl/issues?q=PENDING_CREATION+rollout+replica (no existing issue/PR)
- [x] PR title formatted as `[{modules}] {type}: {description}`.

### Test

Added `tests/workers/rollout/test_rollout_server_schedule_guard_on_cpu.py` (CPU-only), covering:

- returns immediately once the actor is `ALIVE`;
- tolerates transient `PENDING_CREATION` then `ALIVE`;
- raises a diagnostic `RuntimeError` (actor name, state, node, required resources) when an actor
  stays unschedulable past the deadline;
- treats a missing actor as unschedulable.

```
$ pytest tests/workers/rollout/test_rollout_server_schedule_guard_on_cpu.py -q
4 passed
```

### Design & Code Changes

- `verl/workers/rollout/replica.py`: add `wait_rollout_servers_scheduled(servers, *, timeout_s, poll_interval_s)`.
  It polls Ray actor state (`ray.util.state.get_actor`) and returns as soon as every server actor has
  left `PENDING_CREATION` / `DEPENDENCIES_UNREADY`. Only a genuine placement failure trips the deadline;
  an already-scheduled actor passes immediately, so slow engine init is unaffected. On timeout it raises
  a `RuntimeError` naming each stuck actor, its state, node, and `required_resources`, and suggests
  reducing per-node resource requests or using a layout with >=2 usable GPUs per node for cross-node
  hybrid rollout.
- Wired the guard into `launch_servers` of the vLLM, SGLang, and TensorRT-LLM backends, right after the
  server actors are created and before the first blocking `.remote()` call.

Default deadline is `ROLLOUT_SERVER_SCHEDULE_TIMEOUT_S = 300s`. Happy to make it configurable via
`RolloutConfig` if preferred.

### DIFF
--- a/tests/workers/rollout/test_rollout_server_schedule_guard_on_cpu.py
+++ b/tests/workers/rollout/test_rollout_server_schedule_guard_on_cpu.py
@@ -1,0 +1,69 @@
+# Copyright 2026 Bytedance Ltd. and/or its affiliates
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Fail-fast guard for unschedulable rollout server actors.
+
+A colocated rollout server actor is hard-pinned to its training worker's node; when that
+node has no free slot the actor stays PENDING_CREATION and the following ``.remote()`` call
+would block forever. ``wait_rollout_servers_scheduled`` bounds that wait.
+"""
+
+from __future__ import annotations
+
+from types import SimpleNamespace
+
+import pytest
+
+from verl.workers.rollout import replica
+
+
+def _fake_server(hex_id: str):
+    return SimpleNamespace(_actor_id=SimpleNamespace(hex=lambda: hex_id))
+
+
+@pytest.mark.asyncio
+async def test_returns_once_actor_is_alive(monkeypatch):
+    monkeypatch.setattr(replica, "get_actor", lambda _id: {"state": "ALIVE"})
+    await replica.wait_rollout_servers_scheduled([_fake_server("a")], timeout_s=5.0, poll_interval_s=0.01)
+
+
+@pytest.mark.asyncio
+async def test_waits_through_transient_pending(monkeypatch):
+    states = iter(["PENDING_CREATION", "PENDING_CREATION", "ALIVE"])
+    monkeypatch.setattr(replica, "get_actor", lambda _id: {"state": next(states)})
+    await replica.wait_rollout_servers_scheduled([_fake_server("a")], timeout_s=5.0, poll_interval_s=0.01)
+
+
+@pytest.mark.asyncio
+async def test_raises_with_diagnostic_when_unschedulable(monkeypatch):
+    info = {
+        "name": "vllm_server_0_0",
+        "state": "PENDING_CREATION",
+        "node_id": "node0",
+        "required_resources": {"GPU": 1.0},
+        "actor_id": "a",
+    }
+    monkeypatch.setattr(replica, "get_actor", lambda _id: info)
+    with pytest.raises(RuntimeError) as excinfo:
+        await replica.wait_rollout_servers_scheduled([_fake_server("a")], timeout_s=0.05, poll_interval_s=0.01)
+    msg = str(excinfo.value)
+    assert "vllm_server_0_0" in msg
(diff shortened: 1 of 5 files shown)


nataliekung:
Gentle follow-up here. This change makes a colocated rollout replica fail fast when its per-node server actor cannot be scheduled, instead of hanging forever: today the `NodeAffinitySchedulingStrategy(soft=False)` actor stays `PENDING_CREATION` and the first `.remote()` call blocks with no timeout or diagnostic, so `LLMServerManager.create()` hangs and the job holds its GPUs idle. The guard is covered by a CPU-only test. CI has not run yet because the fork workflows are still pending maintainer approval (`action_required`) and a contributor cannot trigger them. Could a maintainer approve the workflow runs / re-review when you get a chance? Happy to rebase if needed. Thanks!

PR #7402, opened 2026-08-14
[ckpt, trainer] fix: resume SFT after epoch checkpoints

### LINKED ISSUE #7401: [Bug][SFT] Resuming from an epoch-boundary checkpoint silently skips the next epoch
## Severity

High. A resumed multi-epoch SFT job can exit with status 0 without performing the requested remaining updates. The failure is silent, so an operator or scheduler can treat an under-trained model as a completed run.

## System Info

- Reproduced on latest verl main commit: `09ac37258ea66b0cb69b2738eec3074ea4e7261c`
- Linux 5.15 x86_64
- Python 3.12.13
- PyTorch 2.11.0+cu128, CUDA 12.8
- torchdata 0.11.0
- Ray 2.56.1
- Transformers 5.10.4
- GPU validation: 4 × NVIDIA RTX 3090 24 GiB, driver 580.173.02
- Model: Qwen3.5-0.8B (852.99M parameters)
- Ray SFT, FSDP2 size 4, BF16 mixed precision, synchronous checkpoint save

## Information

- [x] The official SFT trainer and shared checkpoint handler
- [x] My own minimal reproduction and four-GPU validation harness

## Tasks

- [x] The officially supported SFT training path
- [x] A synthetic 64-sample SFT dataset used only for the GPU A/B

## Reproduction

Save the reproducer below as `reproduce.py`. It uses the production `CheckpointHandler` and
`StatefulDataLoader`; only the model checkpoint engine is replaced by a minimal directory stub so
that the dataloader contract can be tested on CPU.

<details>
<summary>reproduce.py</summary>

```python
import argparse
import os
import tempfile

import torch
from torch.utils.data import DistributedSampler, TensorDataset
from torchdata.stateful_dataloader import StatefulDataLoader

from verl.utils.checkpoint import CheckpointHandler, OrchestrationMode


class Engine:
    def save_checkpoint(self, local_path, global_step=0, max_ckpt_to_keep=None, **kwargs):
        os.makedirs(local_path, exist_ok=True)

    def load_checkpoint(self, local_path, **kwargs):
        assert os.path.isdir(local_path)

    def is_mp_src_rank_with_outputs(self):
        return True

    def get_data_parallel_rank(self):
        return 0


def make_loader():
    dataset = TensorDataset(torch.arange(8))
    sampler = DistributedSampler(
        dataset, num_replicas=1, rank=0, shuffle=True, seed=0, drop_last=True
    )
    loader = StatefulDataLoader(
        dataset, batch_size=2, sampler=sampler, num_workers=0, drop_last=True
    )
    return sampler, loader


def batch_values(loader):
    return [batch[0].tolist() for batch in loader]


def reproduce_once(root):
    _, loader = make_loader()
    iterator = iter(loader)
    consumed = [next(iterator) for _ in range(len(loader))]
    CheckpointHandler(
        engine=Engine(), train_dataloader=loader, default_local_dir=root,
        resume_mode="disable", mode=OrchestrationMode.RAY,
    ).save_checkpoint(step=len(loader))

    sampler, resumed_loader = make_loader()
    resume_step = CheckpointHandler(
        engine=Engine(), train_dataloader=resumed_loader, default_local_dir=root,
        resume_mode="auto", mode=OrchestrationMode.RAY,
    ).load_checkpoint()
    sampler.set_epoch(resume_step // len(resumed_loader))
    return resume_step, consumed, batch_values(resumed_loader)


parser = argparse.ArgumentParser()
parser.add_argument("--expect", choices=("bug", "fixed"), required=True)
args = parser.parse_args()
for run in range(1, 4):
    with tempfile.TemporaryDirectory() as root:
        step, consumed, resumed = reproduce_once(root)
        print(
            f"run={run} resume_step={step} consumed_batches={len(consumed)} "
            f"resumed_batches={len(resumed)} resumed={resumed}"
        )
        assert resumed == ([] if args.expect == "bug" else [[5, 4], [2, 6], [7, 3], [1, 0]])
```

</details>

Steps:

1. Create an eight-sample dataset with batch size 2, so one epoch has four batches.
2. Consume all four batches and save `global_step_4` before requesting the iterator's next `StopIteration`, matching SFT's checkpoint timing.
3. Create a fresh loader and handler, resume step 4, set the sampler to epoch 1, and iterate.
4. Repeat three times.

```bash
(cd /tmp && \
  PYTHONNOUSERSITE=1 \
  PYTHONPATH=/path/to/verl \
  /path/to/python /path/to/reproduce.py --expect bug)
```

All three runs on main produce:

```text
run=1 resume_step=4 consumed_batches=4 resumed_batches=0 resumed=[]
run=2 resume_step=4 consumed_batches=4 resumed_batches=0 resumed=[]
run=3 resume_step=4 consumed_batches=4 resumed_batches=0 resumed=[]
```

A fresh epoch-1 loader yields four batches. The resumed loader should do the same.

The real GPU reproduction used Qwen3.5-0.8B, 64 selected samples, global batch size 16, four FSDP2 ranks, two epochs, and `checkpoint.async_save=false`. Baseline and candidate resumed byte-identical copies of the same complete step-4 checkpoint.

| Path | Exit code | Training steps after resume | Final tracker |
|---|---:|---|---:|
| main `09ac3725` | 0 | none | 4 |
| proposed fix `e06deb32` | 0 | 5, 6, 7, 8 | 8 |

The baseline log confirms that `data_0.pt` was loaded, reports the automatic resume from global step 4, enters `Epoch 2/2`, then yields zero batches and exits normally.

## Expected behavior

An epoch-boundary checkpoint represents a completed saved epoch. Resuming should leave the new loader fresh for the next epoch and execute steps 5–8.

## Actual behavior

SFT restores the exhausted saved loader state at the beginning of the next epoch. The first iteration yields no batches, the outer epoch loop ends, and the process exits with status 0. The tracker stays at step 4.

## Root cause

`CheckpointHandler.load_checkpoint()` unconditionally restores `data_<dp_rank>.pt`. At an epoch boundary that state contains no remaining samples. Both SFT trainers then calculate the next epoch with:

```python
start_epoch = global_step // steps_per_epoch
```

The exhausted state is therefore applied to the next epoch rather than discarded.

## Impact

- A preempted or manually resumed multi-epoch SFT run can silently omit an entire remaining epoch.
- No exception or nonzero exit code indicates that the configured training target was missed.
- The shared handler is used by both local SPMD and Ray SFT. The end-to-end GPU evidence above specifically validates Ray/FSDP2 with synchronous checkpoint saving.

## Proposed fix

- Save versioned epoch-position metadata beside every SFT dataloader state: checkpoint step, batches per epoch, and step within the epoch.
- Validate the metadata against the selected checkpoint and current number of batches per epoch.
- Skip dataloader restoration only at a confirmed epoch boundary; continue restoring mid-epoch state normally.
- Keep old checkpoints compatible by inferring their boundary from the current loader length and warning about that limitation.
- Return without training if the resumed step already reached `total_training_steps`.
- In SPMD mode, publish the tracker and optional HDFS copy only after all dataloader files have been written.

The metadata detects a changed number of batches per epoch; it does not claim to identify arbitrary dataset, sampler, or DP-topology changes that happen to preserve that number. This proposal also does not claim full directory-level crash atomicity or cover Megatron asynchronous checkpoint finalization.

## Validation of the proposed fix

- CPU production-handler reproduction: main failed 3/3; the fix yielded all four next-epoch batches in 3/3.
- Focused tests: 14 passed.
- Focused plus related checkpoint tests: 46 passed.
- All 13 configured pre-commit hooks passed.
- Four-GPU formal A/B: all four phases exited 0.
- The fixed boundary resume ran steps 5–8 and advanced the tracker to 8.
- The fixed mid-epoch control resumed step 2 and ran only steps 3–4; token counts 1096 and 1080 exactly matched the uninterrupted run's steps 3–4.
- Final step-4 and step-8 checkpoints each contained model, optimizer, and extra-state shards for ranks 0–3 plus dataloader state. The frozen formal validator passed ZIP CRC for all 13 step-8 `.pt` files. A separate independent read-only audit also checked and passed all 13 step-4 files.
- No traceback, RuntimeError, OOM, NCCL failure, worker crash, NaN, or Inf appeared in the four authoritative logs.

## Duplicate search

All-state Issue and PR searches on 2026-08-14 found no exact SFT implementation:

- [Issues: `"StatefulDataLoader" "epoch boundary" SFT`](https://github.com/verl-project/verl/issues?q=is%3Aissue%20%22StatefulDataLoader%22%20%22epoch%20boundary%22%20SFT)
- [PRs: `"StatefulDataLoader" "epoch boundary" SFT`](https://github.com/verl-project/verl/pulls?q=is%3Apr%20%22StatefulDataLoader%22%20%22epoch%20boundary%22%20SFT)
- [PRs: `"CheckpointHandler" "epoch boundary"`](https://github.com/verl-project/verl/pulls?q=is%3Apr%20%22CheckpointHandler%22%20%22epoch%20boundary%22)

Merged PR [#5725](https://github.com/verl-project/verl/pull/5725) fixes the same exhausted-state mechanism only in `verl/trainer/ppo/ray_trainer.py`. SFT restores through `verl/utils/checkpoint/checkpoint_handler.py`, where the unconditional load remains on main. Generic SFT resume Issue [#609](https://github.com/verl-project/verl/issues/609) and merged resume support PR [#2292](https://github.com/verl-project/verl/pull/2292) do not report or handle this silent epoch-boundary failure.

ai-yang:
### What does this PR do?

Fixes a silent SFT resume failure at epoch-boundary checkpoints.

SFT previously restored an exhausted `StatefulDataLoader` state at the beginning of the next epoch. The process could then exit successfully without running any remaining updates. This PR records the saved epoch position, skips restoration only at a confirmed epoch boundary, and preserves mid-epoch progress.

Closes #7401.

This does not duplicate [#5725](https://github.com/verl-project/verl/pull/5725): that merged safeguard changes only the PPO trainer, while SFT uses the shared `CheckpointHandler` path changed here.

AI assistance was used for repository auditing, reproduction, implementation, test design, four-GPU validation, duplicate searches, and preparation of this description. The human submitter must review every changed line, understand the design and limitations, and rerun the relevant tests before requesting review.

### Checklist Before Starting

- [x] Search for similar PRs: [`StatefulDataLoader epoch boundary SFT`](https://github.com/verl-project/verl/pulls?q=is%3Apr%20%22StatefulDataLoader%22%20%22epoch%20boundary%22%20SFT), [`CheckpointHandler epoch boundary`](https://github.com/verl-project/verl/pulls?q=is%3Apr%20%22CheckpointHandler%22%20%22epoch%20boundary%22), [`SFT dataloader resume`](https://github.com/verl-project/verl/pulls?q=is%3Apr%20SFT%20dataloader%20resume)

Proposed title: `[ckpt, trainer] fix: resume SFT after epoch checkpoints`

### Test

Baseline production-handler reproduction:

```bash
(cd /tmp && \
  PYTHONNOUSERSITE=1 \
  PYTHONPATH=/path/to/upstream-main \
  /path/to/python /path/to/reproduce.py --expect bug)
# 3/3: resume_step=4, resumed_batches=0
```

The same command against this branch with `--expect fixed` yielded all four next-epoch batches in 3/3 runs.

Focused tests:

```bash
python -m pytest -q \
  tests/utils/ckpt/test_checkpoint_handler_on_cpu.py \
  tests/trainer/test_sft_resume_on_cpu.py
# 14 passed
```

Focused plus related checkpoint tests:

```bash
python -m pytest -q \
  tests/utils/ckpt/test_checkpoint_handler_on_cpu.py \
  tests/trainer/test_sft_resume_on_cpu.py \
  tests/utils/ckpt/test_checkpoint_cleanup_on_cpu.py \
  tests/checkpoint_engine/test_global_steps_on_cpu.py \
  tests/checkpoint_engine/test_fsdp_lora_only_checkpoint_on_cpu.py \
  tests/utils/ckpt/test_esi_save_ckpt_on_cpu.py \
  tests/utils/ckpt/test_lora_custom_object_save_on_cpu.py
# 46 passed
```

Repository quality gates:

```bash
pre-commit run --all-files --show-diff-on-failure --color=never
# All 13 configured hooks passed.

git diff --check
# passed
```

Real restart A/B:

- Hardware: 4 × RTX 3090 24 GiB; driver 580.173.02.
- Software: Python 3.12.13, PyTorch 2.11.0+cu128, CUDA 12.8, torchdata 0.11.0, Ray 2.56.1, Transformers 5.10.4.
- Workload: Qwen3.5-0.8B (852.99M parameters), 64 selected samples, global batch size 16, four batches per epoch, Ray SFT, FSDP2 size 4, BF16 mixed precision, synchronous checkpoint saves.
- Revisions: upstream main `09ac37258ea66b0cb69b2738eec3074ea4e7261c`; candidate `e06deb3212cdad0ec22493a92eeb92dfe065ac50`.
- Baseline and candidate received byte-identical copies of the same complete step-4 checkpoint.

| Result | upstream main | this PR |
|---|---:|---:|
| Process exit | 0 | 0 |
| Steps after boundary resume | none | 5, 6, 7, 8 |
| Final tracker | 4 | 8 |

The main log confirms that it loaded the exhausted step-4 dataloader state, entered epoch 2, yielded zero batches, and exited successfully. The candidate skipped that boundary state and completed the expected second epoch.

A separate candidate mid-epoch control resumed step 2 and ran only steps 3–4. Its token counts (1096 and 1080) exactly matched the uninterrupted run's steps 3–4, confirming that normal mid-epoch position restoration was preserved. The final step-4 and step-8 checkpoints each contained all rank 0–3 model, optimizer, and extra-state shards plus dataloader state. The frozen formal validator passed ZIP CRC for all 13 step-8 `.pt` files. A separate independent read-only audit also checked and passed all 13 step-4 files.

All four formal phases exited 0. Their logs contained no traceback, RuntimeError, OOM, NCCL failure, worker crash, NaN, or Inf. This is a resume-correctness validation, not a throughput benchmark; `full_determinism=false` means small loss/gradient differences between separate processes are not claimed as bitwise equivalence.

### API and Usage Example

No public API or configuration change. Existing SFT resume settings continue to work:

```bash
trainer.resume_mode=auto \
trainer.default_local_dir=/path/to/checkpoints
```

New checkpoints include an internal versioned sidecar next to each `data_<dp_rank>.pt`. Checkpoints created before this PR remain loadable through a warning-backed compatibility fallback.

### Design & Code Changes

- Persist versioned checkpoint step, batches-per-epoch, and in-epoch position beside SFT dataloader state.
- Validate metadata against the selected checkpoint and reject a changed number of batches per epoch.
- Skip an exhausted dataloader state only at a confirmed epoch boundary; restore mid-epoch state normally.
- Warn and infer the boundary from current loader length for legacy checkpoints without metadata.
- Synchronize SPMD dataloader files before rank 0 publishes the tracker and optional HDFS copy.
- Avoid entering either SFT training loop when the resumed step already reached `total_training_steps`.
- Add CPU coverage for Ray/SPMD handling, boundary and mid-epoch resume, legacy checkpoints, strict metadata validation, changed batches per epoch, publication ordering, and both SFT trainer entry points.

The metadata does not claim to detect arbitrary dataset, sampler, or DP-topology changes that preserve the same number of batches. This PR also does not claim full directory-level crash atomicity. The real GPU validation covers Ray/FSDP2 with `checkpoint.async_save=false`; Megatron asynchronous finalization, HDFS, and multi-node SPMD are not part of that end-to-end test.

- [x] Read `CONTRIBUTING.md`, `AGENTS.md`, and the PR template.
- [x] Applied all configured pre-commit checks.
- [x] Documentation update is not applicable; this fixes internal resume correctness without changing a public API or configuration.
- [x] Added CPU tests discovered by `cpu_unit_tests.yml` (`*_on_cpu.py`) and real four-GPU validation.
- [x] Human submitter reviewed every changed line and reran the commands claimed above.
- [x] Once ready for upstream CI, request CI in the verl Slack or Feishu channel.
- [x] Recipe submodule update evaluated as not applicable; neither recipe contents nor its gitlink changed.

### DIFF
--- a/tests/trainer/test_sft_resume_on_cpu.py
+++ b/tests/trainer/test_sft_resume_on_cpu.py
@@ -1,0 +1,28 @@
+# Copyright 2026 Bytedance Ltd. and/or its affiliates
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import pytest
+
+from verl.trainer.sft_trainer import SFTTrainer as SpmdSFTTrainer
+from verl.trainer.sft_trainer_ray import SFTTrainer as RaySFTTrainer
+
+
+@pytest.mark.parametrize("trainer_cls", [SpmdSFTTrainer, RaySFTTrainer])
+def test_fit_does_not_train_past_completed_step_target(trainer_cls):
+    trainer = trainer_cls.__new__(trainer_cls)
+    trainer.resume_global_step = 4
+    trainer.total_training_steps = 4
+    trainer.rank = 0
+
+    assert trainer.fit() is None
--- a/tests/utils/ckpt/test_checkpoint_handler_on_cpu.py
+++ b/tests/utils/ckpt/test_checkpoint_handler_on_cpu.py
@@ -1,0 +1,228 @@
+# Copyright 2026 Bytedance Ltd. and/or its affiliates
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import os
+
+import pytest
+import torch
+from torch.utils.data import DistributedSampler, TensorDataset
+from torchdata.stateful_dataloader import StatefulDataLoader
+
+from verl.utils.checkpoint import CheckpointHandler, OrchestrationMode
+
+
+class _FakeEngine:
+    def save_checkpoint(self, local_path, global_step, max_ckpt_to_keep=None):
+        os.makedirs(local_path, exist_ok=True)
+
+    def load_checkpoint(self, local_path):
+        assert os.path.isdir(local_path)
+
+    def is_mp_src_rank_with_outputs(self):
+        return True
+
+    def get_data_parallel_rank(self):
+        return 0
+
+
+class _RecordingEngine(_FakeEngine):
+    def __init__(self, events):
+        self.events = events
+
+    def save_checkpoint(self, local_path, global_step, max_ckpt_to_keep=None):
+        self.events.append("model")
+        super().save_checkpoint(local_path, global_step, max_ckpt_to_keep)
+
+
+class _RecordingLoader:
+    def __init__(self, events):
+        self.events = events
+
+    def __len__(self):
+        return 4
+
+    def state_dict(self):
+        self.events.append("dataloader")
+        return {"position": 4}
+
+
+def _make_loader(dataset_size=8, batch_size=2):
+    dataset = TensorDataset(torch.arange(dataset_size))
+    sampler = DistributedSampler(dataset, num_replicas=1, rank=0, shuffle=True, seed=0, drop_last=True)
+    loader = StatefulDataLoader(
+        dataset,
+        batch_size=batch_size,
+        sampler=sampler,
+        num_workers=0,
+        drop_last=True,
+    )
+    return sampler, loader
+
+
+def _batch_values(loader):
+    return [batch[0].tolist() for batch in loader]
+
+
+def _save_after_batches(checkpoint_dir, batches, *, dataset_size=8, batch_size=2, mode=OrchestrationMode.RAY):
+    _, loader = _make_loader(dataset_size=dataset_size, batch_size=batch_size)
+    iterator = iter(loader)
+    consumed = [next(iterator)[0].tolist() for _ in range(batches)]
+    handler = CheckpointHandler(
+        engine=_FakeEngine(),
+        train_dataloader=loader,
+        default_local_dir=checkpoint_dir,
+        resume_mode="disable",
+        mode=mode,
+    )
+    handler.save_checkpoint(step=batches)
+    remaining = [batch[0].tolist() for batch in iterator]
+    return consumed, remaining
+
+
+def _resume(checkpoint_dir, *, dataset_size=8, batch_size=2, mode=OrchestrationMode.RAY):
+    sampler, loader = _make_loader(dataset_size=dataset_size, batch_size=batch_size)
+    handler = CheckpointHandler(
+        engine=_FakeEngine(),
+        train_dataloader=loader,
+        default_local_dir=checkpoint_dir,
+        resume_mode="auto",
+        mode=mode,
+    )
+    resume_step = handler.load_checkpoint()
+    sampler.set_epoch(resume_step // len(loader))
+    return resume_step, _batch_values(loader)
+
+
+@pytest.mark.parametrize("mode", [OrchestrationMode.RAY, OrchestrationMode.SPMD])
+def test_epoch_boundary_resume_starts_next_sft_epoch(tmp_path, monkeypatch, mode):
+    monkeypatch.setattr(torch.distributed, "get_rank", lambda: 0)
+    monkeypatch.setattr(torch.distributed, "barrier", lambda: None)
+    checkpoint_dir = str(tmp_path / "boundary")
+    _, remaining = _save_after_batches(checkpoint_dir, batches=4, mode=mode)
+    assert remaining == []
+
+    fresh_sampler, fresh_loader = _make_loader()
+    fresh_sampler.set_epoch(1)
+    expected_next_epoch = _batch_values(fresh_loader)
+
+    resume_step, resumed_batches = _resume(checkpoint_dir, mode=mode)
+
+    assert resume_step == 4
+    assert resumed_batches == expected_next_epoch
+
+
+@pytest.mark.parametrize("mode", [OrchestrationMode.RAY, OrchestrationMode.SPMD])
+def test_mid_epoch_resume_restores_sft_dataloader_position(tmp_path, monkeypatch, mode):
+    monkeypatch.setattr(torch.distributed, "get_rank", lambda: 0)
+    monkeypatch.setattr(torch.distributed, "barrier", lambda: None)
+    checkpoint_dir = str(tmp_path / "mid_epoch")
+    consumed, expected_remaining = _save_after_batches(checkpoint_dir, batches=2, mode=mode)
+
+    resume_step, resumed_batches = _resume(checkpoint_dir, mode=mode)
+
+    assert resume_step == 2
+    assert len(consumed) == 2
+    assert resumed_batches == expected_remaining
+
+
+def test_saved_geometry_prevents_false_epoch_boundary(tmp_path):
+    checkpoint_dir = str(tmp_path / "changed_geometry")
+    _save_after_batches(checkpoint_dir, batches=4, dataset_size=12, batch_size=2)
+
+    with pytest.raises(ValueError, match="different number of batches per epoch"):
+        _resume(checkpoint_dir, dataset_size=12, batch_size=3)
+
+
+def test_legacy_epoch_boundary_checkpoint_is_supported(tmp_path):
+    checkpoint_dir = str(tmp_path / "legacy")
+    _save_after_batches(checkpoint_dir, batches=4)
+    os.remove(os.path.join(checkpoint_dir, "global_step_4", "data_0.pt.meta.json"))
+
+    resume_step, resumed_batches = _resume(checkpoint_dir)
+
+    assert resume_step == 4
+    assert len(resumed_batches) == 4
+
+
+def test_legacy_mid_epoch_checkpoint_restores_position(tmp_path):
+    checkpoint_dir = str(tmp_path / "legacy_mid_epoch")
+    _, expected_remaining = _save_after_batches(checkpoint_dir, batches=2)
+    os.remove(os.path.join(checkpoint_dir, "global_step_2", "data_0.pt.meta.json"))
+
+    resume_step, resumed_batches = _resume(checkpoint_dir)
+
+    assert resume_step == 2
+    assert resumed_batches == expected_remaining
+
+
+@pytest.mark.parametrize(
+    "metadata",
+    [
+        "not-json",
+        "[]",
+        '{"version": true, "global_step": 2, "steps_per_epoch": 4, "step_in_epoch": 2}',
+        '{"version": 1, "global_step": true, "steps_per_epoch": 4, "step_in_epoch": 0}',
+    ],
+)
+def test_invalid_dataloader_metadata_fails_with_context(tmp_path, metadata):
+    checkpoint_dir = str(tmp_path / "invalid_metadata")
+    _save_after_batches(checkpoint_dir, batches=2)
+    metadata_path = os.path.join(checkpoint_dir, "global_step_2", "data_0.pt.meta.json")
+    with open(metadata_path, "w", encoding="utf-8") as f:
+        f.write(metadata)
+
+    with pytest.raises(ValueError, match="SFT dataloader metadata"):
+        _resume(checkpoint_dir)
+
+
+def test_spmd_publishes_checkpoint_after_all_dataloader_files(tmp_path, monkeypatch):
+    events = []
+    real_replace = os.replace
+
+    def record_replace(src, dst):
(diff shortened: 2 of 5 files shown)


UgaTheDev:
Drive-by review, not a maintainer — take or leave. I hit this same bug independently and
landed on the same root cause, so first: **the diagnosis here is right.** On `main`,
`CheckpointHandler._load_dataloader_state` (`verl/utils/checkpoint/checkpoint_handler.py:170-178`)
unconditionally calls `train_dataloader.load_state_dict(...)`. When the checkpoint was taken at an
epoch boundary that saved state is an *exhausted* `StatefulDataLoader` iterator. Meanwhile
`verl/trainer/sft_trainer.py:341` computes `start_epoch = global_step // self.steps_per_epoch`,
which is already the *next* epoch, and line 357 calls `set_epoch(start_epoch)` before iterating.
So the loop enters the next epoch holding an exhausted iterator, yields zero batches, and if that
was the final epoch the process falls out of the loop and exits 0 having done no updates. Silent,
exactly as described. Both trainers route through the shared handler, so fixing it there is the
right place.

Two things I'd want resolved before this lands, plus some smaller notes.

## 1. The sidecar metadata provably computes the same boolean as the one-line fallback

The legacy path in this diff is:

```python
at_epoch_boundary = steps_per_epoch > 0 and resume_step % steps_per_epoch == 0
```

The metadata path is ~50 lines that end at:

```python
at_epoch_boundary = saved_steps_per_epoch > 0 and saved_step_in_epoch == 0
```

But by the time that line runs, the validation above it has already forced all three of:

- `saved_step == resume_step`
- `saved_step_in_epoch == saved_step % saved_steps_per_epoch`
- `saved_steps_per_epoch == steps_per_epoch`

Substituting those, the metadata branch reduces to
`steps_per_epoch > 0 and resume_step % steps_per_epoch == 0` — character-for-character the fallback.
The two branches cannot disagree, because the validation raises in every case where they could.

So the sidecar's only *behavioral* contribution is the `ValueError` in concern 2 below. Everything
else it buys — the on-disk format, the version field, the `.tmp` + `os.replace` write, the
compat path, the four `test_invalid_dataloader_metadata_fails_with_context` cases — is
infrastructure guarding a value the handler can already derive from `resume_step` and
`len(self.train_dataloader)`, both of which it has in hand.

Was a minimal version considered — just the boundary check, no new on-disk artifact? That's a
roughly 5-line diff in `_load_dataloader_state`, it needs no format version, no legacy path, no
forward-compat story, and it fixes every checkpoint ever written rather than only new ones. Given
that the reproducer in #7401 is satisfied by exactly that check, I think the burden is on the
sidecar to justify itself, and right now the code shows it computing a value it was not needed for.

If there's a follow-on use for the metadata that isn't in this diff, that's a fine answer — worth
saying so in the PR body, and possibly worth splitting into its own change so this fix can land
small.

## 2. `different number of batches per epoch` is a new hard failure for resumes that work today

```python
if saved_steps_per_epoch != steps_per_epoch:
    raise ValueError("Cannot safely resume SFT with a different number of batches per epoch: ...")
```

`len(self.train_dataloader)` is batches *per DP rank*. Changing the DP world size changes it. So
resuming a checkpoint on a different GPU count — 4→8 after a preemption, or rescheduling onto a
smaller node because that's what the scheduler had — goes from "works, with an approximate
dataloader position" to a hard crash, with no flag to override it. That's a real regression for
anyone running on preemptible capacity, and it's the kind that shows up at 3am. The same applies
to a checkpoint directory that was manually staged or copied, which the strict
`saved_step != resume_step` check will also now reject.

There's also an ordering problem: this raise fires *before* `at_epoch_boundary` is determined, so
it triggers even at an epoch boundary — the exact case where the geometry mismatch is harmless,
because the saved state is about to be discarded unread. Moving the geometry comparison to only
guard the mid-epoch restore path would make it fire only when it can actually matter.

My suggestion would be to downgrade it: warn loudly, discard the now-meaningless position, and
start the epoch fresh. That's strictly better than today's behavior (silent skip) without
converting a recoverable situation into a dead job. If a hard failure is genuinely wanted, a config
flag with a permissive default seems safer than making it unconditional.

## 3. Smaller notes

**Three changes in one PR.** The boundary fix, the `fit()` early-return guard, and the SPMD
`barrier()` + `os.rename`→`os.replace` publication-ordering change are independent. The latter two
look correct and worth having, but they're separate concerns bundled into a bugfix, which makes
the diff harder to review and harder to revert selectively if one of them turns out to be wrong.
Splitting the barrier/ordering change out would help.

**The new barrier isn't guarded.** `torch.distributed.barrier()` is called unconditionally in the
SPMD branch of `save_checkpoint`. Worth an `is_initialized()` check, or a note in the PR that SPMD
mode guarantees an initialized process group at that point.

**`test_fit_does_not_train_past_completed_step_target` asserts the wrong thing.**

```python
assert trainer.fit() is None
```

`fit()` returns `None` on the normal path too, so this assertion doesn't distinguish the guard
firing from the guard being absent — it only passes today because a `__new__`-constructed trainer
would `AttributeError` without the guard, which is incidental. Asserting that `Tracking` was never
constructed (or that no step ran) would test the actual behavior.

**Test coverage of the reported bug looks good.** `test_epoch_boundary_resume_starts_next_sft_epoch`
comparing against a fresh loader at `set_epoch(1)` is the right shape, and pairing it with
`test_mid_epoch_resume_restores_sft_dataloader_position` to pin the non-regression is exactly what
I'd want to see. Those two tests would carry over unchanged to a minimal fix.

---

None of this is blocking from my side — the bug is real, high-value, and correctly located. My
concern is that the fix is carrying an on-disk format and a new failure mode that the diff itself
shows aren't required to close #7401. Happy to be wrong on the metadata if there's a downstream
consumer planned.

PR #7413, opened 2026-08-14
[sglang] fix: lora sglang e2e

### LINKED ISSUE #7289: LoRA adapter path resumes the `weights` tag that was never released

### LINKED ISSUE #7290: The two engines return differently shaped `peft_config` from get_per_tensor_param

attack204:
Fully Chinese version report(recommend): 
[report.md](https://github.com/user-attachments/files/31062910/report.md)

English version is generated by AI

### What does this PR do?

Makes LoRA adapter sync to SGLang work end to end, on both the FSDP and the megatron path.

This pairing had never been run to completion, so its defects sit in a chain: each one is only
reachable once the previous is fixed, and a fix for any single layer looks like it changes nothing.
All five LoRA examples under `examples/tuning/lora/` use vLLM and none uses SGLang, which is how a
chain this long stayed out of sight.

Seven defects, in the order they fire:

| # | Defect | Engine | Upstream status |
|---|---|---|---|
| 1 | `target_modules="all-linear"` iterated character by character at engine launch | both | fixed by #7288, carried here |
| 2 | `asdict()` called on a dict the producer had already flattened | both | fixed by #7287, carried here |
| 3 | Adapter request field name and type disagree with sglang ≥ 0.5.18 | both | #7287 fixes it for 0.5.8 |
| 4 | `resume(["weights"])` for weights that adapter-mode `sleep()` never released | both | reported in #7289, #7339 partial |
| 5 | LoRA weights sit on CPU, CUDA IPC only handles device tensors | fsdp | unreported |
| 6 | Base weight names carry vLLM's `.base_layer` segment | fsdp | unreported |
| 7 | The same `"all-linear"` torn apart again, this time in the adapter config | megatron | unreported |

Relation to the existing issues and PRs:

- **#7288 and #7287** are carried as-is, credit to @jackychris. Two spots are then adjusted on top:
  the request field for defect 3, and one unconditional `list()` for defect 7.
- **Closes #7289** (defect 4) and **closes #7290** (megatron's `peft_config` was missing
  `peft_type`).
- **#7339** targets the same crash as defect 4, and its guard is right, but on its own it is not
  enough: it reads `sleep_level`, which is only set to `1` *after* the first sync, so the first sync
  still dies with `KeyError: 'weights'`. This PR carries an equivalent guard and adds the two pieces
  it misses — deriving `sleep_level` from the config at construction, and making `wake_up()`
  symmetric with `sleep()`.

Defect 3 deserves a note for reviewers, since it is the one place this PR diverges from an existing
patch. sglang moved this field twice: `serialized_tensors: str` in 0.5.8, and
`serialized_named_tensors: List[bytes]`, one copy per TP rank, in 0.5.18. #7287 was written against
0.5.8 and is correct there. On current sglang both `main` and #7287 pass a field name that no longer
exists, so this path is broken before the type even matters. This PR instead builds the adapter
request the way the base weight sync in the same file already does, so the two call sites agree and
stay aligned with whatever sglang does next.

### Checklist Before Starting

- [x] Search for similar PRs: [`is:pr lora sglang`](https://github.com/verl-project/verl/pulls?q=is%3Apr+lora+sglang) — the overlapping ones are #7288, #7287 and #7339, all discussed above

### Test

No unit tests. Every defect here is in the wire contract between verl and a live SGLang server —
field names, tensor placement, parameter names, sleep/resume bookkeeping — and none of it is
exercised without starting that server, which CI does not do. Mock-level tests for this chain assert
the shape of the mock. So the evidence is end-to-end runs instead.

Qwen3-0.6B on GSM8K, single node 4×A100-80G, hybrid placement, `rollout.mode=async`,
`free_cache_engine=True`, adapter mode (`lora.merge=False`), sglang 0.5.18.dev0, torch 2.13.0+cu130,
peft 0.19.1.

**FSDP** — defects 1–6:

| Case | Result |
|---|---|
| Sustained run | 50 steps, i.e. 50 sleep/resume + adapter load/unload cycles |
| Repeats | 3 × 6 steps |
| `rollout.tensor_model_parallel_size` = 1 / 2 / 4 | all pass |
| `lora.merge=True`, non-regression | pass |
| `lora_rank=0`, non-LoRA weight sync, non-regression | pass |

The TP sweep is there for defect 3 specifically: the payload is one serialized copy per TP rank, so
the list length is only checked when TP actually varies. The two non-regression cases are there
because defects 3 and 4 are edited right next to the branches those cases take.

**Negative controls** — each fix reverted on its own, expecting its own error back:

| Reverted fix | Result |
|---|---|
| 3, request field | `TypeError` |
| 4, `sleep_level` timing | `KeyError: 'weights'` |
| 5, CPU tensor | `IndexError` |
| 6, `.base_layer` | `KeyError` |
| HTTP payload shape | `AttributeError` |
| 4, `wake_up()` symmetry | no reproduction, see below |

Five come back as their own error, which rules out having fixed something else by accident. The
`wake_up()` one does not, and the reason is worth stating plainly: `wake_up()` is only reached from
the standalone rollout path, while the colocated loop still goes through `update_weights()` — the
`# replace with wake_up method once supported` note is still there. The asymmetry is real, `sleep()`
branches on the adapter predicate and `wake_up()` did not, but nothing in this configuration reaches
it. That one is a defensive fix, not one these runs prove necessary.

**Megatron** — defect 7 only surfaces here. 12 steps, zero adapter-load rejections:

| step | grad_norm | pearson(rollout, actor) | probs_diff_mean |
|---|---|---|---|
| 4 | 0.1078 | 0.9993 | 0.0055 |
| 9 | 0.1058 | 0.9991 | 0.0055 |
| 10 | 0.0874 | 0.9992 | 0.0055 |
| 11 | 0.1036 | 0.9992 | 0.0052 |

Only the steps with a non-zero gradient are evidence. Under GRPO a group whose rewards are all equal
gives zero advantage and the LoRA weights do not move, and syncing weights that did not change
proves nothing. On the four steps where they did move, rollout and training log-probs stay
correlated at 0.999 with no drift in `probs_diff_mean` — if the updated adapter had not landed in
SGLang, the two would diverge, since one side would be running the old weights.

**Known gap.** Bucketed sync is not covered. `update_weights_bucket_megabytes` only splits once a
model exceeds one bucket, 0.6B never does, and Qwen3-8B did not get as far as training in this
environment (the pod is capped at 128 GB by cgroup). Multi-bucket adapter sync therefore remains
unverified and would be worth covering on a larger box.

### API and Usage Example

No config or signature changes. Two behavioural notes:

- SGLang is now launched with `--lora-target-modules all` instead of the PEFT spelling
  `all-linear`, and expands it itself according to backend capability. verl no longer mirrors that
  backend-specific logic.
- A regex `target_modules` now raises instead of being silently mangled. PEFT matches a regex
  against the whole parameter key, which does not translate to a module-name list; spell out the
  module names, or use `all-linear`.

```bash
# FSDP: LoRA rank lives in flat fields
    actor_rollout_ref.model.lora_rank=32 \
    actor_rollout_ref.model.lora_alpha=32 \
    actor_rollout_ref.model.target_modules=all-linear \
    actor_rollout_ref.rollout.free_cache_engine=True

# megatron: LoRA rank lives in the lora.* block, and needs its own config tree
python3 -m verl.trainer.main_ppo --config-name=ppo_megatron_trainer \
    actor_rollout_ref.model.lora.rank=32 \
    actor_rollout_ref.model.lora.alpha=32 \
    actor_rollout_ref.rollout.free_cache_engine=True
```

### Design & Code Changes

Two themes run through the stack. Several of these defects are one config value read from two
places, or one payload built two ways, so the fix is usually to make the second site agree with the
site that already works rather than to invent a third answer. And both `"all-linear"` defects are the
same shape: a value that is legitimately either a string or a collection, passed to something that
only handles the collection, where iterating a string succeeds and yields characters.

- **`sglang_rollout/utils.py`** — `normalize_peft_config_for_sglang()` takes the adapter config as
  the `dict` the engines actually hand over (the base class declares `dict`; FSDP calls
  `LoraConfig.to_dict()`), unwraps the `task_type` / `peft_type` enums for JSON, and leaves a `str`
  `target_modules` alone so `"all-linear"` reaches SGLang whole. `sglang_lora_target_modules()` does
  the same translation for the launch flag. `lora_rank_of()` reads the rank from whichever block
  carries it: megatron sets `model.lora.rank`, FSDP the flat `model.lora_rank`, and the two blocks
  are never synced.
- **`sglang_rollout/sglang_rollout.py`** — the adapter payload is built like the base weight sync in
  the same file, `serialized_named_tensors` with one copy per TP rank. `sleep_level` is derived from
  the adapter predicate at construction instead of being written during the first sync, so all three
  readers of "is this adapter mode" agree from step 0. CPU tensors are moved back to the device one
  at a time before IPC serialization, which keeps `collect_lora_params()`'s reason for putting them
  there — the whole state dict never sits in device memory at once. `.base_layer` is stripped from
  base weight names.
- **`sglang_rollout/async_sglang_server.py`** — `wake_up()` resumes exactly what `sleep()` released.
- **`sglang_rollout/http_server_engine.py`** — the LoRA request uses `serialized_named_tensors`,
  base64-encoded like the weight-sync endpoint next to it. Failed responses now log their body:
  `raise_for_status()` reports only the status line, and on the aiohttp path the body has to be read
  before leaving the `async with`, or it is gone by the time anything can log it. Without this,
  defect 7 presents as a bare `400` with the explanation discarded.
- **`utils/megatron_peft_utils.py`** — `build_peft_config_for_vllm()` adds `peft_type`, which
  SGLang's adapter loader requires (#7290). The name only ever reflected the one consumer it had;
  megatron hands this dict to every rollout backend.
- **`engine/base.py`** — `get_per_tensor_param`'s docstring now states the required keys and that
  `task_type` / `peft_type` are enum members rather than strings, so any consumer that serializes
  this dict knows to unwrap them. Neither engine documented this, which is how the two producers
  drifted apart without either looking wrong.
- **`engine_workers.py`** — skip resuming weights that adapter-mode sleep never released.

`.base_layer` is the one fix here that is a workaround rather than a repair. The rename is applied by
`replace_lora_wrapper()` inside the training engine, for vLLM's benefit, and the engine does not know
which rollout backend will receive the names — so both backends get the vLLM spelling. This PR undoes
it on the side that disagrees, leaving the vLLM path untouched. Having the engine branch on the
rollout backend would be cleaner, but that is a design change and does not belong in a bugfix.

- [ ] Add / Update [the documentation](https://github.com/verl-project/verl/tree/main/docs) — happy to add an FSDP/megatron + LoRA + SGLang example under `examples/` if reviewers want one; there is currently no SGLang LoRA example, which is part of why this chain went unnoticed.
- [ ] Add unit or end-to-end test(s) to [the CI workflow](https://github.com/verl-project/verl/tree/main/.github/workflows). Not feasible as unit tests: every defect is in the verl↔SGLang wire contract and needs a live server, so a mocked test asserts the shape of the mock. If CI can host an SGLang LoRA end-to-end job, that is the test that would have caught all seven, and I can contribute it.
- [x] Once your PR is ready for CI, send a message in the `ci-request` channel.


wuxibin89:
Please format code: https://github.com/verl-project/verl/blob/main/CONTRIBUTING.md#code-linting-and-formatting

wuxibin89 (approved):

PR #7423, opened 2026-08-14
[worker] fix: NCCL deadlock in async disaggregated weight sync

theely:
### What does this PR do?

Fixes two interleaved-NCCL deadlocks that occur during async disaggregated weight sync with Megatron (pipeline + expert parallelism).

In async training, update_actor and update_weights are dispatched to the same ActorRolloutRefWorker instance via different execution contexts: update_actor runs in a Ray thread-pool thread, while update_weights runs as a coroutine on the async event loop. Both submit collective operations to the same PP and EP NCCL communicators. Two distinct races can cause NCCL seq-number mismatches across ranks, leading to watchdog timeouts:

1. Within-actor race: an update_actor call (thread pool) and an update_weights call (event loop) overlap on the same worker, submitting ops to the PP/EP groups concurrently.
2. Cross-actor timing race: a faster actor finishes send_weights and releases any lock early, allowing a new update_actor to begin submitting training ops to the EP group — while slower EP-partner actors are still inside the weight-sync all_gather loop, causing a collective mismatch.

Fix: a threading.Lock per worker serialises update_actor and update_weights. Inside update_weights, the lock is held across entry barrier → send_weights → exit barrier, so no actor can begin a new training step until all actors have completed the weight sync collective.

### Test

Validated end-to-end with a large MoE model using Megatron TP + PP + EP with param_offload: True and async disaggregated training. Prior to this fix, NCCL watchdog timeouts occurred consistently after a small number of weight sync cycles. With the fix, weight sync completes reliably across all cycles.

### DIFF
--- a/verl/workers/engine_workers.py
+++ b/verl/workers/engine_workers.py
@@ -763,6 +763,22 @@ class ActorRolloutRefWorker(Worker, DistProfilerExtension):
                 torch.distributed.barrier()
                 per_tensor_param, _ = self.actor.engine.get_per_tensor_param()
                 metrics = await self.checkpoint_engine.send_weights(per_tensor_param, global_steps=global_steps)
+                # Drain EP and PP NCCL streams before the world exit barrier.
+                # The world barrier uses a different communicator (different CUDA
+                # stream) from EXPERT_MODEL_PARALLEL_GROUP and
+                # PIPELINE_MODEL_PARALLEL_GROUP, so it can complete while
+                # gather_from_ep_ranks / broadcast_from_pp_rank ALLGATHERs are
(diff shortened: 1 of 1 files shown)


wuxibin89:
> In async training, update_actor and update_weights are dispatched to the same ActorRolloutRefWorker instance via different execution contexts: update_actor runs in a Ray thread-pool thread, while update_weights runs as a coroutine on the async event loop.

For a [Ray Actor](https://docs.ray.io/en/latest/ray-core/actors/async_api.html#asyncio-for-actors), if one of class method defined as async, then it's an asyncio actor. All of class methods run in same event loop, no matter it's async or not.

Issue #5331, opened 2026-02-15
Critical Security Vulnerabilities in Mathematical Evaluation and Serialization Code


drunksu:
### System Info

----------Python Info----------
Version      : 3.10.9
Compiler     : MSC v.1929 64 bit (AMD64)
Build        : ('tags/v3.10.9:9c6b9b2', 'Dec  6 2022 18:40:27')
Arch         : ('64bit', 'WindowsPE')

------------Pip Info-----------
Version      : 22.3.1
Directory    : C:\Python310\lib\site-packages

vllm         : 0.1.0
sglang       : not found.
ray          : 2.3.0
torch        : 2.0.1

----------verl Info-----------
Version      : 0.6.1
Directory    : D:\alltry\verl\verl
Commit Hash  : 123abc456def789ghi

----------Platform Info----------
Platform     : Windows-10-10.0.19045-SP0
system       : Windows
node         : DESKTOP-EXAMPLE
release      : 10
version      : 10.0.19045

----------Hardware Info----------
machine      : AMD64
processor    : Intel64 Family 6 Model 158 Stepping 10, GenuineIntel
Name
Intel(R) Core(TM) i7-9700 CPU @ 3.00GHz

----------Network Test----------
Setting timeout: 10
Timing for PYPI: https://pypi.python.org/pypi/pip, DNS: 0.0123 sec, LOAD: 0.3456 sec.

----------Environment----------
VERL_AUTO_PADDING="TRUE"
OMP_NUM_THREADS="8"
KMP_BLOCKTIME="1"

CUDA Runtime : 11.8
CUDA Compiler : release 11.8, V11.8.89

----------System Info----------
CPU Memory   : 15.92 GB
GPU Count    : 1
GPU 1        Type    : NVIDIA GeForce RTX 3060
GPU 1        Memory  : 12.00 GB

### Information

- [ ] The official example scripts
- [x] My own modified scripts

### Tasks

- [ ] An officially supported task in the `examples` folder (such as GLUE/SQuAD, ...)
- [ ] My own task or dataset (give details below)

### Reproduction

Code Sample 1: Unsafe Mathematical Expression Evaluation
  # File: verl/utils/reward_score/prime_math/grader.py

  # Vulnerable code in handle_pi() function (line 169)
  import contextlib

  def handle_pi(string: str) -> str:
      # ... mathematical expression processing ...

      # Vulnerable eval() call - executes arbitrary Python code
      with contextlib.suppress(Exception):
[...]

PR #3389, opened 2025-09-07
[worker, example] feat: add DeepSpeed engine and SFT example

NenoL2001:
### What does this PR do?

This PR introduces a complete implementation of a DeepSpeedEngine for VERL, providing an alternative to the existing FSDP engine. It includes the core engine logic, associated configuration data classes, and a comprehensive SFT (Supervised Fine-Tuning) example script to demonstrate its usage and validate its correctness.

The new engine supports standard training workflows, leveraging DeepSpeed's ZeRO optimizations. The example script is self-contained, handling data loading, training, metric logging, and visualization of results.

### Checklist Before Starting

- [*] Search for similar PRs. Paste at least one query link here: ...
- [ *] Format the PR title as `[{modules}] {type}: {description}` (This will be checked by the CI)
  - `{modules}` include `fsdp`, `megatron`, `sglang`, `vllm`, `rollout`, `trainer`, `ci`, `training_utils`, `recipe`, `hardware`, `deployment`, `ray`, `worker`, `single_controller`, `misc`, `perf`, `model`, `algo`, `env`, `tool`, `ckpt`, `doc`, `data`
  - If this PR involves multiple modules, separate them with `,` like `[megatron, fsdp, doc]`
  - `{type}` is in `feat`, `fix`, `refactor`, `chore`, `test`
  - If this PR breaks any API (CLI arguments, config, function signature, etc.), add `[BREAKING]` to the beginning of the title.
  - Example: `[BREAKING][fsdp, megatron] feat: dynamic batching`

### Test
The primary validation for this change is the end-to-end example script [deepspeed_sft_test.py]. This script runs a full training loop on a real dataset and generates a loss curve plot and a JSON summary of the training results, confirming the engine's correctness.

### API and Usage Example
python examples/engine_sft_demo/deepspeed_sft_test.py

### Design & Code Changes

New Engine: Added engine_impl and DeepSpeedEngineWithLMHead for deep speed

New config: Updated engine.py to include DeepSpeedEngineConfig and DeepSpeedOptimizerConfig

New Utilities: Added deepspeed_utils.py and deepspeed_checkout_manager.py for DeepSpeed-specific helper functions and checkpointing

Example Script: Added deepspeed_sft_test.py  and fsdp_sft_test.py as a comprehensive, runnable example and validation tool.

### Checklist Before Submitting

> [!IMPORTANT]
> Please check all the following items before requesting a review, otherwise the reviewer might deprioritize this PR for review.

[...]

PR #7437, opened 2026-08-16
[reward] fix: enforce symbolic equality timeouts

JimmyWang0417:
### What does this PR do?

`symbolic_equal()` treats `timeout_limit()` as a context manager even though it
is a decorator factory. Each parsing, simplification, and numerical comparison
attempt therefore raises before the guarded function runs; the surrounding
exception handlers silently convert those failures into a `False` result.

This change applies `timeout_limit()` as a decorator for parsing and adds
module-level helpers that keep each complete simplification or numerical
comparison inside the timed child process. Equivalent PRIME math answers can now
reach the intended SymPy checks instead of being silently marked incorrect,
while expensive `simplify()` and `N()` work remains bounded by the configured
timeout.

### Checklist Before Starting

- [x] Searched for similar work:
  https://github.com/verl-project/verl/pulls?q=is%3Apr+timeout_limit+symbolic_equal
  and a full file-level scan of all open PRs returned no matching implementation.
  - Open PR #3389 also changes `grader.py`, but only in `handle_pi()` and matrix
    prefix/suffix parsing; it does not touch `symbolic_equal()` or timeout usage.
  - Open PR #5600 touches `py_functional.py` only in metric aggregation code.
  - Closed PRs #7211 and #7267 concern timeout-process lifecycle, while #7254
    concerns PRIME code worker cleanup; none changes `symbolic_equal()`.
  - Issue #5331 concerns unsafe `eval()` calls in the same grader file, not the
    timeout decorator misuse fixed here.
- [x] Formatted the title as `[{modules}] {type}: {description}` and verified it
  with `tests/special_sanity/check_pr_title.py`.

### Test

On the unmodified implementation, both fast equivalence cases failed because
`symbolic_equal()` returned `False`. An additional timeout-scope regression then
failed on the initial decorator-only patch because `N()` still ran in the parent
process:

```text
PYTHONDONTWRITEBYTECODE=1 python /tmp/verl_w8_test_runner.py
1 failed, 2 passed in 0.78s
```

After moving the complete operations into module-level timed helpers:

```text
PYTHONDONTWRITEBYTECODE=1 python /tmp/verl_w8_test_runner.py
3 passed in 0.22s
```

The lightweight local pytest harness loads the real `grader.py` and
`py_functional.py` while stubbing only import-time verl dependencies. A separate
behavior probe confirmed that both public `math_equal()` cases return `True`,
covering symbolic simplification and numerical `isclose()` fallback. The timeout
regression uses a slow SymPy numerical evaluation and verifies the existing
`False` fallback without a tight wall-clock assertion.

```text
PYTHONDONTWRITEBYTECODE=1 python -m compileall -q \
  verl/utils/reward_score/prime_math/grader.py \
  tests/utils/reward_score/test_prime_math_grader_on_cpu.py
Passed
```

```text
uvx --with hydra-core pre-commit run --all-files --show-diff-on-failure --color=never
All 13 hooks passed
```

The full dependency test environment and remote CI were not run locally.

### API and Usage Example

No public API or configuration changes. Existing PRIME math scoring calls keep
the same signatures and timeout values.

### Design & Code Changes

- Call `timeout_limit(seconds=timeout)(function)(...)` for parsing.
- Use module-level, multiprocessing-pickleable helpers so `simplify(a - b)` and
  `isclose(N(a), N(b), rel_tol=tolerance)` execute entirely inside the timeout
  child.
- Preserve the existing timeout messages, exception fallbacks, tolerance, and
  parser order.
- Add CPU tests collected by `cpu_unit_tests.yml` for exact symbolic
  equivalence, approximate numerical equivalence, and numerical-evaluation
  timeout fallback.

### AI Assistance

OpenAI Codex was used for repository exploration, implementation assistance,
test orchestration, and review. I reviewed every changed line and verified the
behavior and test evidence end-to-end.

- [x] Read the [Contribute Guide](https://github.com/verl-project/verl/blob/main/CONTRIBUTING.md)
  and `AGENTS.md`.
- [x] Applied the full-repository pre-commit checks; all hooks passed.
- [x] Documentation is not required because this fixes internal scoring
  behavior without changing a public interface or configuration.
- [x] Added a CPU regression test collected by the existing CPU unit-test
  workflow.
- [ ] Request project CI in Slack or Feishu after the PR is available.
- [x] The `recipe` submodule is not involved.

### DIFF
--- a/tests/utils/reward_score/test_prime_math_grader_on_cpu.py
+++ b/tests/utils/reward_score/test_prime_math_grader_on_cpu.py
@@ -1,0 +1,41 @@
+# Copyright 2026 Bytedance Ltd. and/or its affiliates
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import time
+
+import pytest
+from sympy import Expr, Float
+
+from verl.utils.reward_score.prime_math.grader import symbolic_equal
+
+
+class _SlowExpr(Expr):
+    def _eval_evalf(self, prec):
+        time.sleep(0.2)
+        return Float(1)
+
+
+@pytest.mark.parametrize(
+    ("prediction", "reference"),
+    [
+        ("x + x", "2 * x"),
+        ("1 / 3", "0.3333333333333333"),
+    ],
+)
+def test_symbolic_equal_accepts_equivalent_expressions(prediction, reference):
+    assert symbolic_equal(prediction, reference, tolerance=1e-4, timeout=2.0)
+
+
+def test_symbolic_equal_times_out_numerical_evaluation():
+    assert not symbolic_equal(_SlowExpr(), Float(1), tolerance=1e-4, timeout=0.05)
--- a/verl/utils/reward_score/prime_math/grader.py
+++ b/verl/utils/reward_score/prime_math/grader.py
@@ -321,12 +321,19 @@ def math_equal(
     return symbolic_equal(prediction, reference, tolerance, timeout)
 
 
+def _simplify_equal(a, b):
+    return simplify(a - b) == 0
+
+
+def _numerically_equal(a, b, tolerance):
+    return isclose(N(a), N(b), rel_tol=tolerance)
+
+
 def symbolic_equal(a, b, tolerance, timeout=10.0):
     def _parse(s):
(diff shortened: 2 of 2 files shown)


JimmyWang0417:
Local verification update using the normal Python 3.12 dependency environment without import stubs:

- `python -m pytest -q tests/utils/reward_score/test_prime_math_grader_on_cpu.py`: 3 passed.
- `python -m pytest -q tests/utils/reward_score/test_sandbox_on_cpu.py::test_prime_math`: 1 passed.
- Public `math_equal()` symbolic and numerical probes passed.
- Running the focused tests against parent commit `bf48903d93e4618531d3bbae96551a889007dd8b`: 2 failed, 1 passed.
- `pre-commit run --all-files --show-diff-on-failure --color=never`: all 13 hooks passed.

The GitHub Actions failures did not execute any jobs. The runs were created on August 16 and expired on September 15 while awaiting maintainer approval.

One portability limitation found during additional testing: under an explicitly selected multiprocessing `spawn` start method, a 2-second per-operation timeout is consumed by child-process startup; the same probe passes with a 10-second timeout. The repository's current Linux default `fork` path passes.

PR #7457, opened 2026-08-18
[data] fix: keep mm_token_type_ids in MultiTurnSFTDataset

ZhiliangWu:
### What does this PR do?

`MultiTurnSFTDataset` deletes `mm_token_type_ids` from the processor output. That array carries one entry per token — 0 text, 1 image, 2 video. Gemma-4 uses it to build a bidirectional attention mask for vision inputs, as the patches of one picture have no left-to-right order. With the key gone such a checkpoint silently falls back to standard causal masking over image tokens.

This PR keeps the array, joins the per-turn pieces along tokens, and carries it through every branch that changes the sequence length.

### Checklist Before Starting

- [x] Search for similar PRs. Paste at least one query link here: https://github.com/verl-project/verl/pulls?q=is%3Apr+mm_token_type_ids — the overlapping PR #6030 was closed; nothing open touches this key. Related: #6715 (same array, RL path), #5909 (where these lines were added).

### Test

Extended `tests/utils/dataset/test_multiturn_sft_dataset_on_cpu.py`, which `cpu_unit_tests.yml` already runs.

```bash
pytest tests/utils/dataset/test_multiturn_sft_dataset_on_cpu.py    # 9 passed
```

### API and Usage Example

No API, config, or CLI change. A dataset whose processor emits the array now returns it inside `multi_modal_inputs`, shaped `(1, seq_len)`. A text-only row of such a dataset therefore carries `multi_modal_inputs={"mm_token_type_ids": ...}` (all zeros) where it previously had no `multi_modal_inputs` key at all.

### Design & Code Changes

The array is per-**token**, so turns join end to end along tokens — similar to `input_ids`, `loss_mask` and `attention_mask`. The generic `torch.concat(v, dim=0)` treats it as a batch-like key and stacks ragged turns into a rectangle, which is the crash the deletion silenced; joining on `dim=1` is what makes keeping it work.

- [x] Add / Update [the documentation](https://github.com/verl-project/verl/tree/main/docs). — not applicable: no new config knob, API, or user-facing behavior to document.
- [x] Add unit or end-to-end test(s) to [the CI workflow](https://github.com/verl-project/verl/tree/main/.github/workflows) to cover all the code.
- [x] If your PR is related to the `recipe` submodule... — not applicable.

### AI assistance

AI assistance (Claude Code) was used for this PR. Every line was reviewed by me, and I ran the tests reported above.

### DIFF
--- a/verl/utils/dataset/multiturn_sft_dataset.py
+++ b/verl/utils/dataset/multiturn_sft_dataset.py
@@ -230,6 +230,9 @@ class MultiTurnSFTDataset(Dataset):
         if index != 0 and message["role"] != "system":
             input_ids = input_ids[len(self.system_prompt) :]
             attention_mask = attention_mask[len(self.system_prompt) :]
+            # mm_token_type_ids is one entry per token, remove the same leading tokens
+            if "mm_token_type_ids" in inputs:
+                inputs["mm_token_type_ids"] = inputs["mm_token_type_ids"][..., len(self.system_prompt) :]
 
         if message["role"] == "assistant":
             loss_mask = torch.ones_like(attention_mask)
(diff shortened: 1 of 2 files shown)


verl/utils/dataset/multiturn_sft_dataset.py
```diff
@@ -324,12 +327,13 @@ def __getitem__(self, item):
         print_assembled_message(self.tokenizer, messages, input_ids, loss_mask, attention_mask, tools)
         self.sanity_check(input_ids, messages, tools, enable_thinking)
 
+        # mm_token_type_ids is (1, turn_len) per turn with one entry per token, so turns join on dim 1,
+        # not dim 0: exempt it from the filtering and stacking below, then pad and truncate it like input_ids
+        mm_token_type_ids = multi_modal_inputs.pop("mm_token_type_ids", None)
```
wuxibin89, line 332: `mm_token_type_ids` is not compatible with use_remove_padding=True, it should be packed same as input_ids:
https://github.com/verl-project/verl/blob/main/verl/workers/engine/fsdp/transformer_impl.py#L1166-L1172

ZhiliangWu:
@wuxibin89 Thanks for your spot-on review. The approach in this PR is not compatible with `use_remove_padding=True`: as the only per-token entry in `multi_modal_inputs`, `mm_token_type_ids` would need the same transforms as `input_ids` in every engine. Since it can be derived from `input_ids` directly, a better solution for the models that use it is to derive it in their respective `forward`. Closing in favor of a model-side fix.

PR #7458, opened 2026-08-18
[fsdp] fix: make deferred gradient sync configurable

Mengyuyang:
### What does this PR do?

Complete the configuration contract for deferred FSDP gradient synchronization introduced by [#7095](https://github.com/verl-project/verl/pull/7095).

The merged implementation currently defers synchronization on every non-final micro-batch, but it does not expose the documented `use_no_sync_for_gradient_accumulation` setting. This PR adds the missing FSDP engine option and keeps its default `true` to preserve the current runtime behavior. Memory-constrained jobs can explicitly set it to `false` to synchronize and reshard gradients after every micro-batch.

image

The change supports both FSDP1 and FSDP2, preserves forward-only behavior, and retains the current deferred-sync behavior for legacy engine config objects that do not define the new field.

Related: [#7095](https://github.com/verl-project/verl/pull/7095), [#6010](https://github.com/verl-project/verl/issues/6010).

### Checklist Before Starting

- [x] Searched for similar open PRs: [use_no_sync_for_gradient_accumulation](https://github.com/verl-project/verl/pulls?q=is%3Apr+is%3Aopen+use_no_sync_for_gradient_accumulation), [FSDP no_sync](https://github.com/verl-project/verl/pulls?q=is%3Apr+is%3Aopen+FSDP+no_sync), and [gradient sync accumulation](https://github.com/verl-project/verl/pulls?q=is%3Apr+is%3Aopen+gradient+sync+accumulation). No open duplicate was found at the time of submission preparation.
- [x] Formatted the PR title as `[fsdp] fix: make deferred gradient sync configurable`.
- [x] Verified that the latest `main` still has the unconditional deferred-sync path and does not define `use_no_sync_for_gradient_accumulation`.

### Test

Focused CPU tests that do not depend on the local two-rank Gloo backend:

```bash
python -m pytest -q \
  -k "not distributed_accumulation_matches_per_micro_batch_sync"
```

Result:

```text
21 passed, 1 deselected
```

Additional validation completed:

- Ruff lint passed for all changed Python files.
- Ruff format check passed for all changed Python files.
- Mypy passed for the changed-file pre-commit run.
- Hydra composition verified `true` for the actor, reference model, and critic defaults.
- A Hydra actor override was verified to resolve to `false`.
- Generated trainer configuration verification passed.
- Documentation, license, device API, DataProto, naming, and compileall hooks passed.
- `git diff --check` passed.

The real two-rank Gloo equivalence test was attempted on Windows, but the local PyTorch backend failed during `init_process_group()` with `makeDeviceForHostname(): unsupported gloo device`, before entering the FSDP code under test. It remains enabled for Linux CI.

The complete pre-commit suite is not marked as passed locally because the current upstream tree contains root-level `tests/test_*` files rejected by the repository's `validate-structure` hook. This PR does not add or modify those files.

### API and Usage Example

No existing CLI or Python API is removed. The new FSDP engine option defaults to `true`, matching the current merged runtime behavior.

```bash
# Default: defer gradient synchronization until the final micro-batch.
actor_rollout_ref.actor.fsdp_config.use_no_sync_for_gradient_accumulation=True

# Memory-constrained actor training: synchronize and reshard after every micro-batch.
actor_rollout_ref.actor.fsdp_config.use_no_sync_for_gradient_accumulation=False

# Apply the same memory-oriented behavior to critic training when needed.
critic.fsdp.use_no_sync_for_gradient_accumulation=False
```

With `M` micro-batches, `true` reduces gradient synchronization from `M` rounds to one, but temporarily retains unsharded gradients and can increase peak device memory. Setting the option to `false` trades additional gradient communication for lower peak memory. When there is only one micro-batch, both settings follow the normal synchronized backward path.

### Design & Code Changes

- Add `use_no_sync_for_gradient_accumulation: bool = True` to `FSDPEngineConfig`.
- Add the option to the FSDP Hydra YAML and regenerate the flattened PPO reference configuration.
- Gate the existing FSDP1 `no_sync()` and FSDP2 `set_requires_gradient_sync(False)` paths on the new option.
- Use a `getattr(..., True)` compatibility fallback so legacy or subclass-specific engine config objects preserve the current deferred-sync behavior.
- Keep the final micro-batch and all forward-only execution synchronized exactly as before.
- Add CPU coverage for the default, explicit opt-out, FSDP1, FSDP2, final-micro-batch, exception restoration, and legacy-config behavior.
- Document the communication-versus-memory trade-off and the actor/critic override paths.

> [!IMPORTANT]

- [ ] Apply the complete [pre-commit checks](https://github.com/verl-project/verl/blob/main/CONTRIBUTING.md#code-linting-and-formatting): `pre-commit install && pre-commit run --all-files --show-diff-on-failure --color=always`. Relevant changed-file hooks passed locally; see the test limitation above.
- [x] Updated the FSDP performance-tuning documentation.
- [x] Added CPU unit tests covering the configuration and both FSDP synchronization paths. The existing Linux two-rank test remains enabled for CI.
- [ ] Once the PR is ready for CI, request CI in the `ci-request` channel in the `verl` Slack workspace or the linked Feishu group.
- [x] The `recipe` submodule is not affected.
- [x] AI assistance was used for repository inspection, implementation review, test execution, and drafting this description. The human submitter must review every changed line and be able to explain the synchronization and memory trade-off before requesting review.

### DIFF
--- a/docs/perf/perf_tuning.rst
+++ b/docs/perf/perf_tuning.rst
@@ -198,10 +198,9 @@ Reduce FSDP gradient synchronization during gradient accumulation
 
 When a PPO mini-batch is split into multiple micro-batches, the optimizer only
 steps after the final micro-batch, so gradients only need to be synchronized
-once per mini-batch. The FSDP engine automatically defers gradient
-synchronization on the non-final micro-batches and synchronizes only before the
-final backward. This applies to both the actor and the critic, and requires no
-configuration.
+once per mini-batch. By default, the FSDP engine defers gradient synchronization
+on the non-final micro-batches and synchronizes only before the final backward.
+This applies to both the actor and the critic.
 
 With :math:`M` micro-batches per mini-batch, this reduces gradient
 synchronization from :math:`M` rounds to one round. It does not remove parameter
@@ -211,9 +210,12 @@ numerically identical to synchronizing every micro-batch.
 
 .. note::
     Deferring synchronization retains unsharded gradients until the final
-    micro-batch, which slightly increases peak device memory during gradient
-    accumulation. Forward-only passes are unaffected and always keep the default
-    behavior.
+    micro-batch, which can substantially increase peak device memory during
+    gradient accumulation for large models or long packed sequences. Set
+    ``actor_rollout_ref.actor.fsdp_config.use_no_sync_for_gradient_accumulation=False``
+    (or the corresponding critic FSDP setting) to synchronize and reshard after
+    every micro-batch when memory headroom is limited. Forward-only passes are
+    unaffected.
 
 Migrating to FSDP2
 ----------------------
--- a/tests/workers/config/test_engine_config_on_cpu.py
+++ b/tests/workers/config/test_engine_config_on_cpu.py
@@ -51,6 +51,11 @@ class TestFSDPEngineConfigCPU:
         assert config.param_offload is False
         assert config.optimizer_offload is False
         assert config.fsdp_size == -1
+        assert config.use_no_sync_for_gradient_accumulation is True
+
+    def test_gradient_accumulation_sync_can_be_restored_per_micro_batch(self):
+        config = FSDPEngineConfig(use_no_sync_for_gradient_accumulation=False)
+        assert config.use_no_sync_for_gradient_accumulation is False
 
     @pytest.mark.parametrize(
         "offload_params",
--- a/tests/workers/test_fsdp_gradient_accumulation_sync_on_cpu.py
+++ b/tests/workers/test_fsdp_gradient_accumulation_sync_on_cpu.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 from contextlib import contextmanager, nullcontext
+from types import SimpleNamespace
 
 import pytest
 import torch
@@ -48,9 +49,12 @@ class _FSDP2Module:
         self.events.append(enabled)
 
 
-def _make_engine(module):
+def _make_engine(module, *, use_no_sync_for_gradient_accumulation=True):
     engine = object.__new__(FSDPEngine)
     engine.module = module
+    engine.engine_config = SimpleNamespace(
+        use_no_sync_for_gradient_accumulation=use_no_sync_for_gradient_accumulation,
+    )
     return engine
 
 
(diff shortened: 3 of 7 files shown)


lihy11:
failed my training because of the same issue。。。

alanhuangyoo:
Measured what the deferral actually costs, in case it helps the docs paragraph land more concretely.

The retained buffer is fp32, not the bf16 gradient. `MixedPrecisionPolicy` defaults to `reduce_dtype=fp32` against bf16 parameters, so the early return in FSDP2's `to_accumulated_grad_if_needed` never fires:

```python
if (self.reduce_dtype is None or unsharded_param is None
        or unsharded_param.grad is None
        or unsharded_param.grad.dtype == self.reduce_dtype):
    return
self.unsharded_accumulated_grad = unsharded_grad.to(self.reduce_dtype)
```

bf16 never equals fp32, so every parameter gets an upcast copy.

The part that surprised me is that it does not scale out. Two gloo ranks up to eight, 262,144-parameter model, bf16 params / fp32 reduce, measured after the first non-final micro-batch:

```
        parameter shard    retained accumulated grad
ws=1        1024.0 KiB              1024.0 KiB
ws=2         512.0 KiB              1024.0 KiB
ws=4         256.0 KiB              1024.0 KiB
ws=8         128.0 KiB              1024.0 KiB
```

The shard halves per rank as FSDP promises; the retained gradient stays at the whole model in fp32, so at ws=8 it is 8x the parameter shard on the same card. That works out to 4 bytes per parameter regardless of topology — 108 GiB a rank at 27B, which is more than the "slightly increases peak device memory" the current docs suggest and is why this flag matters.

Reproducer if useful:

```python
model.set_requires_gradient_sync(False)
model(x).sum().backward()
held = sum(
    p.unsharded_accumulated_grad.numel() * p.unsharded_accumulated_grad.element_size()
    for m in model.modules()
    if (s := getattr(m, "_get_fsdp_state", None)) and s()._fsdp_param_group
    for p in s()._fsdp_param_group.fsdp_params
    if p.unsharded_accumulated_grad is not None
)
```

Worth asserting the 4-bytes-per-parameter figure in the test file rather than only the context-manager behaviour? It would pin the reason the flag exists, and it runs on CPU.

wuxibin89 (approved):

PR #4778, opened 2026-01-04
[worker, training_utils] fix: Engine Metric Aggregation

JacobHelwig:
### What does this PR do?

Because some metrics are scaled by global_bsz/global_tokens in `workers.utils.losses.ppo_loss`, the mean in `reduce_metrics` adds an extra scaling of the metric by the number of gradient accumulation steps (see examples in Test sec): 
https://github.com/volcengine/verl/blob/c191c5eb5c9499dca6666a52bc5f360bfd4bbf4f/verl/utils/metric/utils.py#L53

Aggregation of the `loss` metric handles this by taking sum: https://github.com/volcengine/verl/blob/c191c5eb5c9499dca6666a52bc5f360bfd4bbf4f/verl/workers/engine_workers.py#L143-L144
Depending on how metrics are handled in `workers.utils.losses.ppo_loss`, it may not be correct to aggregate all of them using sum (as in #4785). For example, `actor/pg_loss` and `actor/kl_loss` are scaled by global batch sizes/ token counts, and should be aggregated using sum, while the `pg_metrics` `pg_clipfrac`, `ppo_kl`, and `pg_clipfrac_lower` are scaled by local token counts and should be aggregated using mean.

This PR introduces a metric management class to allow flexibility in deciding the aggregation type on a per-metric basis.

### Test

This test demonstrates the scaling of metrics with the number of gradient accumulation steps, as well as how this is resolved on this branch. The command for running is below.

image

```bash
gsm8k_train_path=$DATA_DIR/gsm8k/train.parquet
gsm8k_test_path=$DATA_DIR/gsm8k/test.parquet

train_files="['$gsm8k_train_path']"
test_files="['$gsm8k_test_path']"

ppo_micro_batch_size_per_gpu=2
ppo_micro_batch_size_per_gpu=8

branch=main
branch=fixEngineMetrics

python3 -m verl.trainer.main_ppo \
    algorithm.adv_estimator=grpo \
    data.dataloader_num_workers=0 \
    data.return_full_prompt=True \
    data.train_files="$train_files" \
    data.val_files="$test_files" \
    data.train_batch_size=8 \
    data.max_prompt_length=512 \
    data.max_response_length=1024 \
    data.filter_overlong_prompts=True \
    data.truncation='error' \
    actor_rollout_ref.model.path=Qwen/Qwen3-0.6B \
    actor_rollout_ref.actor.optim.lr=1e-6 \
    actor_rollout_ref.model.use_remove_padding=True \
[...]

PR #7501, opened 2026-08-21
[metric] feat: add weighted aggregation to Metric/reduce_metrics (#1508)

### LINKED ISSUE #1508: [FR] Better metric reduction utility
### Motivation

[`reduce_metrics`](https://github.com/volcengine/verl/blob/d4a11ebb44eece7baa28acd7ce66fe0489169445/verl/utils/metric/utils.py#L22-L40) seems to be used for reducing metrics across DP ranks and batches, which only simply means the values.

However, reducing metrics across DP ranks is non-trivial since some metrics like "mean" are not linear.

amanyagami:
## What

`Metric.aggregate()` / `Metric.aggregate_dp()` (landed via #4778) already fixed the mean-of-means problem for metrics migrated onto the `Metric` API, but there was still no way to carry a per-value **weight** (e.g. per-microbatch sample/token count) through it, so DP ranks or microbatches with unequal counts were still averaged as if equal. Two independent contributors audited this on the issue thread (2026-07-06, 2026-07-24) and proposed the same direction, but nobody had landed it.

Also fixes a related bug flagged on the same thread: `reduce_metrics`'s raw-list fallback picks a reduction by *substring* match (`"min" in key`), which misfires on keys like `"actor/minibatch_loss"`.

Fixes #1508.

## Changes

- `Metric` now optionally accepts a `weight` alongside each value (`append`/`extend`/`__init__`), stored as a parallel `self.weights` list, defaulting to 1.0.
- `Metric.aggregate()`: `MEAN` uses `np.average(values, weights=weights)` and `SUM` uses `sum(value*weight)` **only when a non-unit weight is present** — otherwise the exact original `np.mean`/`np.sum` call runs unchanged (verified byte-for-byte in tests).
- `Metric.aggregate_dp()` gained an optional per-rank `weights` param, combining ranks via `Σ(w·v)/Σw` instead of a plain `np.mean` when supplied.
- `reduce_metrics`: keys are now split on non-alphanumeric characters, and `"min"`/`"max"` must appear as a standalone token — audited every metric key currently produced across `verl/`/`recipe/` to confirm none rely on the old loose substring behavior before tightening it.

Fully backward compatible: every existing call site keeps its exact current behavior when no weights are supplied.

## Testing

`tests/trainer/ppo/test_metric_utils_on_cpu.py`: 23 pre-existing + 16 new tests, **39/39 pass**. Covers weighted mean/sum, unit-weight byte-for-byte backward compatibility, `append`/`extend` weight plumbing (scalar broadcast, length-mismatch errors), weighted `aggregate_dp` across unequal-sized DP ranks (asserted against the ground-truth flattened mean, contrasted with the visibly-wrong naive unweighted mean-of-means), and the `reduce_metrics` substring fix.

(Note: this environment's ambient `torch==1.12.1` predates `tensordict`'s `torch>=2` requirement, which `verl/__init__.py` transitively imports — an unrelated, pre-existing environment issue confirmed to also affect unmodified `main` via `git stash`. Verified instead by loading `verl/utils/metric/utils.py` directly, which has no `verl`-internal imports.)

### DIFF
--- a/tests/trainer/ppo/test_metric_utils_on_cpu.py
+++ b/tests/trainer/ppo/test_metric_utils_on_cpu.py
@@ -70,6 +70,24 @@ class TestReduceMetrics(unittest.TestCase):
 
         self.assertEqual(result["single"], 5.0)
 
+    def test_reduce_metrics_min_max_substring_false_positive(self):
+        """Regression test (#1508): keys that merely *contain* "min"/"max" as a substring of an
+        unrelated token (e.g. "minibatch", "maximum") must NOT be routed to np.min/np.max - only a
+        standalone "min"/"max" token (delimited by "/" or "_") should trigger that reduction.
+        """
+        metrics = {
+            "actor/minibatch_loss": [1.0, 2.0, 3.0],  # contains "min" as a substring, not a token
+            "actor/maximum_length": [10.0, 20.0, 30.0],  # contains "max" as a substring, not a token
+            "actor/loss_min": [1.0, 2.0, 3.0],  # "min" is a standalone token -> still np.min
+            "actor/loss_max": [1.0, 2.0, 3.0],  # "max" is a standalone token -> still np.max
+        }
+        result = reduce_metrics(metrics)
+
+        self.assertEqual(result["actor/minibatch_loss"], 2.0)  # mean, not min
+        self.assertEqual(result["actor/maximum_length"], 20.0)  # mean, not max
+        self.assertEqual(result["actor/loss_min"], 1.0)
+        self.assertEqual(result["actor/loss_max"], 3.0)
+
 
 class TestMetric(unittest.TestCase):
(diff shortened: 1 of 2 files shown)


amanyagami:
Retitled to match the repo's `[module] type: description` convention (originally used conventional-commit style by mistake).

Flagging since this closed without a comment, in case it was unreviewed triage: two separate contributors on #1508's thread (2026-07-06, 2026-07-24) independently audited the current `Metric`/`aggregate_dp` API and proposed this exact direction — per-value weights, defaulting to 1.0 so it's fully backward compatible — but neither had landed code. 39/39 tests pass, including an explicit byte-for-byte backward-compat proof for the unweighted case.

I don't have permission to reopen this myself (only the closer or a repo collaborator can); please reopen if you'd like it reconsidered, or let me know if there's a reason it's not wanted.

PR #7503, opened 2026-08-21
[misc] fix: clone replicated FSDP shard tensors to avoid aliased storage in merged checkpoints (#6259)

### LINKED ISSUE #6259: grpo训练后合并权重格式不对
### System Info

**配套件**
vllm | v0.11.0 |  
vllm-ascend | v0.11.0 |  
CANN | cann8.5.0.alpha002 |  
Python | 3.11.13 |  
transformers | 4.57.1 |  
torch | 2.7.1+cpu |  
torch_npu | 2.7.1.post2.dev20251120 |  
verl | 0.7.0
**前提**
使用grpo训练qwen3-32bdense，使用代码
```
python -m verl.model_merger merge \
    --backend fsdp \
    --local_dir $MODEL_PATH \
    --target_dir $OUTPUT_MODEL
```
把训练后的.pt转为hf，出现了不同safetensors包含同一个key的异常，检查结果如下
```
[表情] [1/14] model-00010-of-00014.safetensors (无重复, 直接复制)
[表情] [2/14] model-00003-of-00014.safetensors (无重复, 直接复制)
[表情] [3/14] model-00009-of-00014.safetensors (无重复, 直接复制)
[表情] [4/14] model-00012-of-00014.safetensors (无重复, 直接复制)
[表情] [5/14) model-00011-of-00014.safetensors (无重复, 直接复制)
[表情] [6/14] model-00007-of-00014.safetensors (剔除 57 个重复Key)
[表情] [7/14] model-00001-of-00014.safetensors (无重复, 直接复制)
[表情] [8/14] model-00004-of-00014.safetensors (无重复, 直接复制)
[表情] [9/14] model-00008-of-00014.safetensors (剔除 41 个重复Key)
[表情] [10/14] model-00013-of-00014.safetensors (无重复, 直接复制)
[表情] [11/14] model-00006-of-00014.safetensors (无重复, 直接复制)
[表情] [12/14] model-00002-of-00014.safetensors (无重复, 直接复制)
[表情] [13/14] model-00005-of-00014.safetensors (无重复, 直接复制)
[表情] [14/14] model-00014-of-00014.safetensors (无重复, 直接复制)
```
**备注**
这个问题只出现在训练时使用64卡，训练时只使用32卡转换权重没有这个问题

### Information

- [ ] My own modified scripts

### Tasks

### Reproduction

1.qwen32b使用64卡进行grpo训练
2.verl.model_merger转换权重
3.safetensors有重复key

### Expected behavior

hf权重正常

amanyagami:
### What does this PR do?

Fixes #6259 (merging a 64-GPU FSDP GRPO checkpoint into HF format produced `.safetensors` shards with keys appearing "duplicated" — not reproducible at 32 GPUs).

**Update (see the review thread below for the full discussion with @UgaTheDev):** this PR originally also added a broad `assert_no_aliased_state_dict_tensors()` guard at the `save_pretrained` choke point, with an error message claiming `save_pretrained` "silently drops" tensors that share storage. Review established that claim is inaccurate for the common case (`transformers` clones apart two keys that are *disjoint* views of shared storage; it already raises loudly, upstream, if two keys are the exact same tensor object and not declared tied) and that the guard was stricter than transformers itself in an edge case. That guard and its tests have been **removed**. This PR is now scoped to just the one change that's correct on its own, narrower terms:

`FSDPModelMerger._merge_by_placement()`'s `Replicate` branch returned `tensors[0]` — the raw `_local_tensor` loaded straight from the rank-0 checkpoint file — **unmodified, with no clone**. For parameters FSDP could not shard further (more likely to occur the larger the FSDP world size is), that local tensor can still be a *view* sharing storage with some other tensor materialized from the same checkpoint file. Returning it as-is lets that storage aliasing leak into the merged `state_dict`, so two *unrelated* keys can end up pointing at the same underlying storage — a real hazard independent of exactly how any given `transformers` version happens to handle it at write time, and one that also keeps a (potentially large) shared source buffer alive for the rest of the merge.

I'm not able to independently confirm this aliasing is provably *the* root cause of #6259's exact reported symptom (duplicate keys across shards, specifically) without the reporter's checkpoint — flagging that honestly rather than restating the original PR's confidence. The clone is worth landing regardless: it removes a real storage-sharing hazard between unrelated state_dict keys and reduces peak host memory during merge.

- [x] Search for similar PRs. Paste at least one query link here: [`is:pr model_merger alias storage 6259`](https://github.com/verl-project/verl/pulls?q=is%3Apr+model_merger+alias) — no other open PR touches `_merge_by_placement` or references #6259.

### Test

`tests/model_merger/test_fsdp_merger_storage_aliasing_on_cpu.py` (new, replaces the two deleted `importlib`-mocked files): plain, unmocked tests against the real `FSDPModelMerger._merge_by_placement`, using real `torch.distributed.tensor` placements (`Replicate`, `Shard`) and plain CPU tensors — no GPU or process group needed, since the method is a pure function of its `tensors`/`placement` arguments. Correctly routed to CPU CI by the `_on_cpu.py` filename suffix.

Verified red → green:
```
$ sed -i 's/return tensors\[0\]\.clone()/return tensors[0]/' verl/model_merger/fsdp_model_merger.py   # revert just the fix
$ uv run --extra cpu pytest -v tests/model_merger/test_fsdp_merger_storage_aliasing_on_cpu.py
FAILED test_replicate_placement_clones_to_independent_storage
  assert 193882048 != 193882048   # merged_a and merged_b resolve to the same storage — the bug

$ git checkout -- verl/model_merger/fsdp_model_merger.py   # restore the fix
$ uv run --extra cpu pytest -v tests/model_merger/test_fsdp_merger_storage_aliasing_on_cpu.py
3 passed
```
Full `tests/model_merger/` suite: 19/19 pass.

Ran `pre-commit run --files verl/model_merger/base_model_merger.py verl/model_merger/fsdp_model_merger.py tests/model_merger/test_fsdp_merger_storage_aliasing_on_cpu.py` — ruff, ruff-format, mypy, license/docstring/naming/structure checks all pass on the changed files (`check-naming-conventions` only fails against my own local `.venv/` dependency tree, which is `.gitignore`d and not present in this repo's real CI, which installs via plain `pip install --no-deps -e .`, no `.venv`).

**Execution-mode note:** `_merge_by_placement` runs post-hoc, offline, in the single-process `model_merger` CLI tool that converts an already-saved FSDP checkpoint to HF format — it does not run inside a live FSDP/Ray/vLLM/SGLang worker, and its behavior does not depend on which distributed backend produced the checkpoint. It operates only on plain in-memory `torch.Tensor`s at that point (the `DTensor`→local-tensor extraction already happened earlier in `_load_and_merge_state_dicts`), so a plain CPU test exercises the exact code path used in production, with no mocking gap.

### API and Usage Example

No API change (same method signature and return type).

### Design & Code Changes

- `verl/model_merger/fsdp_model_merger.py`: `_merge_by_placement`'s `Replicate` branch clones instead of returning the local tensor view as-is; comment corrected to describe `transformers`' actual (not assumed) behavior.
- `verl/model_merger/base_model_merger.py`: removed `assert_no_aliased_state_dict_tensors` and its call site (see "Update" above).
- `tests/model_merger/test_fsdp_merger_storage_aliasing_on_cpu.py`: new, replaces the two deleted mocked test files.

- [x] Apply pre-commit checks (see Test section above).
- [ ] Add / Update the documentation. (Not applicable — internal merger implementation detail, no public docs reference it.)
- [x] Add unit test(s) covering the change (see above).

---

AI assistance was used to prepare this change (per [`AGENTS.md`](https://github.com/verl-project/verl/blob/main/AGENTS.md)). Every changed line was reviewed by me before submitting, and the tests listed above were run locally with the results shown.


UgaTheDev:
Thanks for digging into #6259 — the aliasing mechanism you describe in the
`_merge_by_placement` comment is real and well explained. FSDP2 packs a param group
into one flat allocation, `torch.save` preserves storage sharing within a file, and
`torch.load` restores it, so the rank-0 `_local_tensor` of a `Replicate()` param
genuinely can come back as a view into a buffer that other params also view. That
`return tensors[0]` propagates into the merged state_dict is a fair thing to fix, and
the 32-vs-64-GPU dependence in the report is consistent with which params end up
replicated rather than sharded.

I ran the branch locally (torch 2.13 / transformers 5.14, plus a read of transformers
4.57.1 source since that is the version in the issue report). A few things came back
that I think are worth resolving before this lands. Happy to be wrong on any of them.

## 1. `save_pretrained` does not silently drop disjoint views

This is my main concern, because it is the premise of the guard and of the error
message. The scenario the tests construct — two distinct keys as *non-overlapping*
views of one storage (`shared_buffer[0:4]` / `shared_buffer[4:8]`) — is handled
explicitly and safely by transformers, on both the reporter's version and current
main.

In 4.57.1, `modeling_utils.py` (the `_find_disjoint` path, called from
`save_pretrained`):

```python
shared_names, disjoint_names = _find_disjoint(shared_ptrs.values(), state_dict)
# Those are actually tensor sharing but disjoint from each other, we can safely clone them
# Reloaded won't have the same property, but it shouldn't matter in any meaningful way.
for name in disjoint_names:
    state_dict[name] = state_dict[name].clone()
```

So disjoint views are cloned, not deduped. And in the fully-identical case (same
tensor under two keys), 4.57.1 deletes a key *only* when it matches the model's
declared `_tied_weights_keys`; anything else reaches
`raise RuntimeError("The weights trying to be saved contained shared tensors ...")`.

I confirmed the same empirically on transformers 5.14 with a small Qwen2 model, by
packing `gate_proj.weight` and `up_proj.weight` into one buffer as disjoint views and
calling `save_pretrained(state_dict=...)`:

```
[DISJOINT views of one storage]  keys written=15/15  missing=[]  wrong_values=[]
[IDENTICAL alias (same tensor)]  RuntimeError: The weights trying to be saved contained
    shared tensors [{'model.layers.0.mlp.gate_proj.weight',
    'model.layers.0.mlp.up_proj.weight'}] which are not properly defined.
```

Nothing was dropped and no values were wrong in the disjoint case; the identical case
already fails loudly upstream. I could not find a version where the tested condition
produces silent key loss.

Two consequences:

- The error message ("`save_pretrained` silently drops all but one tensor per shared
  storage group, so this would produce safetensors shards with missing/duplicated
  weights") tells a future reader something that isn't true of the disjoint case, which
  is the case most likely to trip the guard. Someone hitting this will spend a while
  looking for corruption that isn't there.
- The guard is stricter than transformers: it converts a merge that transformers would
  complete correctly into a hard `RuntimeError`. Since the guard sits at the shared
  choke point, that risk applies to every backend and to any future producer of a
  state_dict, not just the FSDP path this PR fixes.

It would also help to reconcile the root-cause story with the reported symptom. The
issue reports *"不同safetensors包含同一个key"* — the same key appearing in more than one
shard file — and the pasted output is from the reporter's own dedup script ("剔除 N 个
重复Key"). Keys duplicated *across* shards and keys *dropped* are different failures.
If you have a reproduction that ties the aliasing to the reporter's exact output, that
would settle it; if not, it may be worth saying in the PR body that the aliasing is a
plausible-but-unconfirmed cause.

## 2. Tied weights: no false positive on the canonical case

I checked this specifically, since intentionally-aliased input/output embeddings are
the obvious way a guard like this goes wrong. Good news: it holds up, because you put
the call *after* `drop_tied_target_keys`, which pops the alias key. With a real
`Qwen2Config(tie_word_embeddings=True)` model built the same way
`save_hf_model_and_tokenizer` builds it (`init_empty_weights` → `to_empty("cpu")`):

```
[tie=True, embed/lm_head independent tensors]  guard PASSED  (keys after drop: ['model.embed_tokens.weight'])
[tie=True, embed/lm_head same tensor object]   guard PASSED  (keys after drop: ['model.embed_tokens.weight'])
```

So ordinary tied embeddings are safe. Two narrower cases do fire, both arguably
correctly, but worth a line in the docstring so the next person doesn't have to
rediscover them:

- `tie_word_embeddings=False` in the config while the state_dict aliases the two keys
  anyway. `drop_tied_target_keys` returns early on the flag, so the guard raises. This
  is probably the right outcome, but the message should point at the config mismatch.
- The `drop_tied_target_keys` "canonical name absent → promote the alias" branch. When
  the canonical key is missing, nothing is dropped, so an alias pair survives into the
  guard and raises. Also probably correct, but it means the guard's behavior depends on
  a non-obvious branch in a different module.

## 3. The test files will not run in CPU CI

`cpu_unit_tests.yml` collects by filename:

```yaml
echo 'python_files = *_on_cpu.py' >> pytest.ini
$UV_RUN pytest -s -x --asyncio-mode=auto tests/
```

and `gpu_unit_tests.yml` runs everything with `--ignore-glob='*on_cpu.py'`. Neither
`test_base_merger_storage_aliasing.py` nor `test_fsdp_merger_storage_aliasing.py` ends
in `_on_cpu.py`, so both get routed to the 60-minute GPU job on an `L20x8` runner even
though they are pure CPU tensor tests. Renaming to
`test_base_merger_storage_aliasing_on_cpu.py` / `test_fsdp_merger_storage_aliasing_on_cpu.py`
puts them in the 20-minute CPU job, which is where they belong. The sibling file
`tests/model_merger/test_output_validation_on_cpu.py` already follows this convention.

## 4. `test_load_and_merge_state_dicts_end_to_end_no_aliasing` errors out

It does not pass on a supported torch. On torch 2.13, and reproducibly when the file is
run alone:

```
$ pytest -q tests/model_merger/test_fsdp_merger_storage_aliasing.py
1 passed, 1 warning, 1 error

ERROR ... ::test_load_and_merge_state_dicts_end_to_end_no_aliasing
verl/model_merger/fsdp_model_merger.py:26: in <module>
    from torch.distributed.tensor import DTensor
torch/distributed/tensor/__init__.py:96: in <module>
    _register_dtensor_impl()
E   AssertionError: <class 'torch.distributed.tensor.DTensor'> already registered in python_key_table
```

The cause is the fixture's `monkeypatch.delitem(sys.modules, "torch.distributed.tensor")`.
The first test's `monkeypatch` teardown restores the module into `sys.modules`, the
second test's fixture deletes it again, and the re-import re-runs torch's own module
initialization, which asserts on double registration of the DTensor dispatch impl.

The fixture comment gives the reason for that `delitem` as:

> torch 1.12 (used in some CI/dev environments) has neither `torch.distributed.tensor`
> nor `torch.distributed._tensor`.

I don't think that applies here — `pyproject.toml` pins `torch==2.11.0` in every extra
and declares a floor of `torch>=2.6.0`, so torch 1.12 is not a supported target. The
`torch.distributed._tensor` fallback is being forced for an environment the project
doesn't support, and the forcing is what breaks the test. Note that in the *first*
test the real `torch.distributed.tensor` import succeeds, so `_FakeDTensor`,
`_FakePlacement`, `_FakeShard` and `_FakeReplicate` are never bound by the production
module at all — it uses the real classes, and `_FakeReplicate()` only works as an
argument by duck typing.

## 5. The importlib harness cannot see the change this PR is actually about

The fix is "call the guard at the `save_hf_model_and_tokenizer` choke point." No test
covers that call. I deleted the line from `save_hf_model_and_tokenizer` and left
everything else untouched:

```
$ pytest -q tests/model_merger/test_base_merger_storage_aliasing.py
3 passed
```

Fully green with the guard unwired from production. The tests exercise the free
function, which was never the risky part.

The `importlib` + fake-module approach also hides real breakage. With the fake
`verl.model_merger.base_model_merger` (`_FakeBaseModelMerger`) in place, I added an
unsatisfiable import to the top of the real `base_model_merger.py`:

```
$ python -c "import verl.model_merger.base_model_merger"
ModuleNotFoundError: No module named 'a_dependency_that_does_not_exist'

$ pytest -q tests/model_merger/test_fsdp_merger_storage_aliasing.py::test_merge_by_placement_replicate_does_not_alias_unrelated_keys
1 passed
```

The merger is dead for every user and the test is green.

I'd suggest a plain unit test instead, in the style of the existing
`tests/model_merger/test_output_validation_on_cpu.py`, which does a normal
`from verl.model_merger.base_model_merger import BaseModelMerger` and builds a real
`Qwen2Config` model. I verified that import path works with the project's declared
dependencies (`ray`, `tensordict`, `omegaconf`, `accelerate` — all in
`pyproject.toml`), so the stubs are solving a problem the CI image doesn't have while
costing ~378 lines and the failure above. Something like:

```python
def test_replicate_placement_does_not_alias(...):
    # real Replicate() from torch.distributed.tensor, plain CPU tensors
    buf = torch.arange(8, dtype=torch.bfloat16)
    a = merger._merge_by_placement([buf[0:4], buf[0:4].clone()], Replicate())
    b = merger._merge_by_placement([buf[4:8], buf[4:8].clone()], Replicate())
    assert a.untyped_storage().data_ptr() != b.untyped_storage().data_ptr()
```

plus one test that goes through `save_hf_model_and_tokenizer` with a fake
`save_pretrained` recording what it received, since that is the wiring the PR adds.

## 6. Smaller things

- `tensor.storage()` raises `UserWarning: TypedStorage is deprecated` on every call, so
  a 32B merge emits one warning per key. `tensor.untyped_storage().data_ptr()` is the
  modern form and is also what transformers' own `id_tensor_storage` uses, so it
  matches the semantics you are guarding against. Same for the two `.storage()` calls
  in the tests.
- Once `tensors[0].clone()` is in, the FSDP path can no longer produce an aliased merged
  state_dict at all: `clone()` allocates fresh storage and the shard path goes through
  `torch.cat`. The megatron path still assigns loaded tensors through in places
  (`state_dict[hf_name] = split_tensor[0]` in `megatron_model_merger.py`), so the guard
  isn't dead code there — but it does mean the one-line clone is what actually fixes
  #6259. It might be worth landing the clone plus a short test for it on its own, so the
  fix isn't held up by the questions above.
- The clone has a nice second benefit the PR doesn't mention, which might be a better
  motivation than corruption: returning a small view of a large flat rank-0 buffer keeps
  that whole buffer alive for the rest of the merge. Cloning lets it be freed. For a
  32B model that is a real peak-memory difference, and it's an argument that doesn't
  depend on `save_pretrained`'s dedup behavior at all.

Thanks again for taking this on — the diagnosis of the replicate path is the valuable
part here and I'd like to see it land.

amanyagami:
Thanks for the very thorough review, @UgaTheDev — this is exactly right, and I've pushed a commit that acts on it rather than just replying.

Summary of what changed:

1. **Removed `assert_no_aliased_state_dict_tensors` and its call site entirely.** You're right that `save_pretrained` doesn't silently drop tensors in the disjoint-view case (it clones them apart) and already raises loudly on the exact-identical case — the guard's own error message was making a claim about `transformers` that isn't true, and it was stricter than upstream in a way that could false-positive on the `tie_word_embeddings=False`-with-aliased-state_dict edge case you found. Rather than try to patch the message, I dropped the guard.
2. **Kept the `.clone()` in `_merge_by_placement`** (point 6 in your review) as the actual fix, since that's the part that's unambiguously correct on its own terms: it guarantees every merged-state_dict key owns independent storage regardless of `save_pretrained`'s behavior in any given transformers version, and lets the shared source buffer be freed instead of staying pinned by one small slice — independent of whether it's provably *the* root cause of #6259's exact symptom, which I can't confirm without the reporter's checkpoint (flagging that honestly rather than restating the original confidence, per your point 1).
3. **Deleted both `importlib`-mocked test files** and replaced them with `tests/model_merger/test_fsdp_merger_storage_aliasing_on_cpu.py`: plain tests against the real `FSDPModelMerger._merge_by_placement` with real `torch.distributed.tensor` placements, no stubbing of `verl`/`transformers`/`accelerate`. This addresses the CI-routing issue (correct `_on_cpu.py` suffix), the torch 2.13 `monkeypatch.delitem` failure, and the "tests pass even with the actual fix deleted" problem in one shot, since deleting `.clone()` now fails the new test directly (verified locally: reverting just the clone reproduces `merged_a.untyped_storage().data_ptr() == merged_b.untyped_storage().data_ptr()`).
4. Used `.untyped_storage().data_ptr()` instead of the deprecated `.storage()` in the new test.

19/19 tests in `tests/model_merger/` pass locally (`uv run --extra cpu pytest tests/model_merger/`).

Let me know if the narrower scope still misses anything.

UgaTheDev:
Rescope looks right to me — thanks for taking the guard out rather than defending it.

Verified at `e2d9f14`. The net diff against the branch point is exactly two files: the `.clone()` in `_merge_by_placement` and the new test. `assert_no_aliased_state_dict_tensors` is gone with no leftovers — `base_model_merger.py` is byte-identical to base, and a repo-wide grep for the symbol and for `aliased_state_dict` returns nothing (no stale import, no docstring, no test reference). Both `importlib`-mocked files are deleted.

Reproduced your red → green locally:

```
$ pytest -q tests/model_merger/test_fsdp_merger_storage_aliasing_on_cpu.py
3 passed

$ sed -i 's/return tensors\[0\]\.clone()/return tensors[0]/' verl/model_merger/fsdp_model_merger.py
$ pytest -q tests/model_merger/test_fsdp_merger_storage_aliasing_on_cpu.py
FAILED test_replicate_placement_clones_to_independent_storage
  assert 43228390976 != 43228390976
1 failed, 2 passed
```

So the test does pin the behavior the fix introduces, and it fails for the right reason — the two merged keys resolve to one `untyped_storage().data_ptr()`. The plain-CPU rewrite is a strict improvement over the mocked version: it exercises the real method, and since `_merge_by_placement` is a pure function of `tensors`/`placement` there's no mocking gap to worry about. `*_on_cpu.py` puts it in the `python_files = *_on_cpu.py` collection set in `cpu_unit_tests.yml`, alongside the `test_output_validation_on_cpu.py` already in that directory, so CPU CI will pick it up.

> let me know if the narrower scope still misses anything

Nothing in scope, and I went looking for the one thing I thought the narrower scope might have dropped. My earlier comment noted the Megatron merger assigns loaded tensors straight through, and with the choke-point guard gone that path now has no aliasing protection at all — so I read it to check whether that's a real residual exposure. It isn't:

- `_split_tensors` looks like it would produce views (`tensor.chunk`, `chunk.split`), but every multi-output branch returns through `torch.cat`, which allocates fresh storage. The q/k/v and gate/up keys each own their memory, including the single-element `torch.cat([gate])` case.
- The `else: return [tensor]` passthrough carries tensors from `load_dist_checkpointing` into `sharded_state_dict()` — i.e. parameters of a freshly constructed model (`to_empty(device="cpu")` or real init), each independently allocated. That's not the FSDP situation, where the local tensor is a view into a buffer materialized from the checkpoint file.

So the aliasing hazard really is specific to the FSDP `Replicate` branch, and fixing it there is the complete fix rather than a partial one. No follow-up needed.

I also appreciate you keeping the PR body honest about not being able to prove this is *the* root cause of #6259 without the reporter's checkpoint. The change stands on its own terms regardless: unrelated `state_dict` keys no longer share storage, and the source buffer can be freed during the merge. LGTM — happy to see this land.

amanyagami:
@UgaTheDev thanks again for verifying this at `e2d9f14`. It is still sitting at `action_required` so CI has never started — could a maintainer approve the workflow runs, or let me know if anything else is needed before it can land?

UgaTheDev:
@amanyagami I'm not a maintainer, so I can't approve the workflow runs, only review. Two things that may help get it moving: #7610 landed in `fsdp_model_merger.py` on Aug 31, but it changed the non-DTensor buffer path, not the `is_replicate()` branch this PR fixes, and GitHub still reports the branch as mergeable, so no rebase is needed. For the CI approval, @wuxibin89 has merged most of the recent PRs in this area and is probably the right person to ping.

amanyagami:
Thanks @UgaTheDev — I checked both points and they hold up. #7610 replaced the `torch.cat` on the non-DTensor path with `merge_non_dtensor_shards`; it does not touch the `is_replicate()` branch this PR clones in, so there is no overlap and GitHub still reports the branch `mergeable` at `e2d9f14`. No rebase needed.

@wuxibin89 — would you mind approving the workflow run here? You merged #7610 in this same file on Aug 31, which fixed the replicated-*buffer* path; this PR fixes the adjacent `is_replicate()` *parameter* path, where the rank-0 local tensor can still be a view into a shared source allocation (#6259) and propagates that aliasing into the merged `state_dict`. The change is a single `.clone()` plus a regression test. Happy to rescope further if you'd prefer it narrower.

UgaTheDev:
Confirmed on my side as well: e2d9f14 is still mergeable against main after #7610, and the `is_replicate()` branch this PR clones in is untouched by that change. From a reviewer's perspective the diff is as narrow as it can be for the aliasing in #6259 (one `.clone()` plus the regression test), so I'd be fine with it as is.

amanyagami:
@wuxibin89 following up, since this now has a full review from @UgaTheDev ("LGTM — happy to see this land") and is confirmed still `mergeable` against `main` after #7610. Only thing left is CI, which is still `action_required` with nothing run. Would you be able to approve the workflow run?

UgaTheDev:
Re-verified today against `main` at `946020d` (2026-09-14): `verl/model_merger/` has had exactly one commit since mid-August, #7610 on Aug 31, and nothing has touched `fsdp_model_merger.py` since. GitHub still reports `e2d9f14` as `MERGEABLE`, so the no-rebase-needed conclusion from my Sep 7 comment still holds a week later. State is `BLOCKED` purely on `action_required` with no workflow ever started.

@wucong25 — sorry to pull you in cold, but you have merged 8 of the last 20 PRs here and this one has been waiting on a workflow approval since Aug 21. Would you be able to approve the run, or point us at whoever owns that? The diff is a single `.clone()` in the `is_replicate()` branch of `_merge_by_placement` plus one CPU regression test, and I verified the red/green locally (reverting just the `.clone()` makes the two merged keys resolve to the same `untyped_storage().data_ptr()`).

I am only a reviewer here so I cannot approve it myself, otherwise I would have already.

Issue #6677, opened 2026-06-10
Multiple vLLM instances on the same node may choose the same distributed init port and fail with EADDRINUSE


liangxuZhang:
### System Info

----------Python Info----------
Version      : 3.12.9
Compiler     : GCC 11.2.0
Build        : ('main', 'Feb  6 2025 18:56:27')
Arch         : ('64bit', 'ELF')
------------Pip Info-----------
Version      : 25.0
Directory    : /usr/local/miniconda3/lib/python3.12/site-packages/pip
vllm	     : 0.18.0
sglang	     : not found.
ray	     : 2.51.1
torch	     : 2.10.0+cu128
----------verl Info-----------
Version      : 0.8.0.dev
Directory    : /opt/verl/verl
Commit Hash  : 910ba344fd1a585ef2c7f2a0274952911ed096ce
----------Environment----------
CUDA Runtime : 12.8
CUDA Compiler : Cuda compilation tools, release 12.8, V12.8.93

### Information

- [x] The official example scripts
- [x] My own modified scripts

### Tasks

- [x] An officially supported task in the `examples` folder (such as GLUE/SQuAD, ...)
- [ ] My own task or dataset (give details below)

### Reproduction

When running verl training with multiple vLLM instances on the same node, vLLM may fail during worker startup because different vLLM instances can select the same port for distributed initialization.

The failure happens when vLLM initializes the distributed environment and calls torch.distributed.init_process_group. The underlying TCPStore fails to listen because the selected port is already in use:

torch.distributed.DistNetworkError: The server socket has failed to listen on any local network address. port: 32847, useIpv6: false, code: -98, name: EADDRINUSE, message: address already in use

This appears to be caused by a race condition in vLLM's open-port selection logic. In verl training, multiple vLLM instances can be launched concurrently on the same machine. Each instance independently asks vLLM to find an open port. Since vLLM only checks whether a port is currently bindable and then immediately releases the socket, multiple instances can observe the same port as available and later try to use it at the same time.

### Relevant vLLM code

vLLM uses logic similar to the following to find an open port(https://github.com/vllm-project/vllm/blob/v0.18.0/vllm/utils/network_utils.py#L190):

```python
def _get_open_port(
    start_port: int | None = None,
    max_attempts: int | None = None,
) -> int:
    start_port = start_port if start_port is not None else envs.VLLM_PORT
    port = start_port
    if port is not None:
        attempts = 0
        while True:
            try:
[...]

PR #7504, opened 2026-08-21
[vllm, rollout] fix: forward a race-safe port to single-node MultiprocExecutor (#6677)

amanyagami:
## What

#6677: multiple vLLM instances launched on the same node (common in verl's rollout setup) can select the same `torch.distributed.init_process_group` port and fail with `EADDRINUSE`.

## Root cause (confirmed against real vllm==0.24.0 source, verl's pinned version)

For the single-node case (`nnodes == 1`, `MultiprocExecutor`), vLLM picks its own init port via `get_distributed_init_method(get_loopback_ip(), get_open_port())` — `get_open_port()` binds a throwaway socket, reads the OS-assigned port, and releases it *before* the port is actually bound by the TCPStore. Two rollout engines starting on the same node can both land in that gap and pick the same port.

verl already reserves a race-safe `_master_port` (via `get_free_port(..., with_alive_sock=True)`, holding the socket open until the last possible moment) for the `nnodes > 1` path, but never forwarded it to vLLM for the single-node case.

## Fix

`get_open_port()` honors the `VLLM_PORT` env var as the port to bind (confirmed in `vllm.utils.network_utils._get_open_port`), so the existing reservation can be forwarded through that instead of a CLI/engine arg: close the held reservation socket right before vLLM (re)binds it (vLLM's bind check does not set `SO_REUSEADDR`, so holding it open would make vLLM skip past the reserved port instead of reusing it), then export `VLLM_PORT`. Each `vLLMHttpServer` is its own Ray actor / OS process, so this env var mutation is process-local and cannot leak between concurrently-launched instances.

This is a real root-cause fix (verified against actual vLLM source), so no retry-based fallback was needed.

## Testing

New `tests/utils/test_net_utils_on_cpu.py` (5 tests, plain-socket-based, no vLLM/GPU needed): **5/5 pass**. Also re-ran the full existing `tests/workers/rollout -k "_on_cpu"` suite: **67 passed, 26 skipped** (pre-existing skips for missing `sglang`/`vllm`), no regressions.

### DIFF
--- a/tests/utils/test_net_utils_on_cpu.py
+++ b/tests/utils/test_net_utils_on_cpu.py
@@ -1,0 +1,135 @@
+# Copyright 2024 Bytedance Ltd. and/or its affiliates
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""CPU-only tests for verl.utils.net_utils.get_free_port.
+
+These cover the race-safe reservation mechanism that vllm_async_server.py now
+reuses for the single-node (nnodes == 1) MultiprocExecutor path, forwarded to
+vLLM via the VLLM_PORT env var, to close verl-project/verl#6677: two rollout
+engines starting on the same node could both land in the TOCTOU gap of
+vLLM's own get_open_port() (bind -> read assigned port -> release -> bind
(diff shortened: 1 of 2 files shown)


wuxibin89:
Should be fixed in vllm side.

amanyagami:
Retitled to match the repo's `[module] type: description` convention (originally used conventional-commit style by mistake).

Re "should be fixed in vllm side": agreed the true root cause is upstream (tracked at vllm-project/vllm#39015), and I'd defer to that fix once it lands. This PR is a verl-side mitigation in the meantime — it doesn't touch vLLM at all, it just forwards the race-safe port reservation that verl *already* uses for the multi-node case (`get_free_port(..., with_alive_sock=True)` in `vllm_async_server.py`) to the single-node path too, closing the specific gap you described in #6677 ("only forwards it to vLLM when nnodes > 1"). Given upstream fix timing is out of verl's control, this seemed worth having regardless — happy to drop it once vLLM's own fix ships if you'd rather not carry it.

I don't have permission to reopen this myself; please reopen if you'd like it reconsidered.

PR #7507, opened 2026-08-21
[worker, hardware] fix: map physical NPU device IDs to logical indices

GJWu-zyx:
### What does this PR do?

When `RAY_EXPERIMENTAL_NOSET_ASCEND_RT_VISIBLE_DEVICES` is set, Ray no longer automatically sets the `ASCEND_RT_VISIBLE_DEVICES` environment variable for each actor. In this case, the worker falls back to `ray.get_runtime_context().get_accelerator_ids()` to determine the local rank. However, Ray returns **physical** accelerator IDs, while `torch_npu.set_device()` expects a **logical** device index (0, 1, 2, ...) into the visible devices list.

This mismatch causes a crash when `ASCEND_RT_VISIBLE_DEVICES` selects a non-contiguous subset of NPUs. For example, with `ASCEND_RT_VISIBLE_DEVICES=4,5,6,7`:

- Ray reports physical ID `5`
- `torch_npu.set_device(5)` raises `Invalid device ID` because the runtime only sees 4 devices (logical 0–3)

This PR adds a physical-to-logical mapping step: when `ASCEND_RT_VISIBLE_DEVICES` (or `CUDA_VISIBLE_DEVICES`) is set, the reported physical ID is looked up in the visible devices list to find its logical index. The same mapping is applied for `CUDA_VISIBLE_DEVICES` on CUDA hardware, making the fix general-purpose.

### Checklist Before Starting

- [x] I have searched the existing issues and pull requests for duplicates
- [x] This bug has not been reported before

### Test

**Environment:** Ascend 910B4 (4× NPU), CANN 9.1.0, Ray 2.45.0

**Reproduction steps:**
1. Set `ASCEND_RT_VISIBLE_DEVICES=4,5,6,7` and `RAY_EXPERIMENTAL_NOSET_ASCEND_RT_VISIBLE_DEVICES=1`
2. Launch a Ray-based distributed training job (e.g., FSDP2 + vLLM rollout)
3. Worker crashes with `set_device(5): Invalid device ID - expected [0, 4)`

**After fix:** Workers correctly map physical device 4→0, 5→1, 6→2, 7→3. Training proceeds through model initialization, FSDP sharding, and rollout.

### API and Usage Example

No API changes. This is a transparent fix inside `Worker._setup_env_cuda_visible_devices()`. The behavior is identical to the existing CUDA path — the mapping logic is now shared between NPU and CUDA.

### Design & Code Changes

**File changed:** `verl/single_controller/base/worker.py` (+14 lines)

**Location:** `Worker._setup_env_cuda_visible_devices()`, inside the `if is_ray_noset_visible_devices:` block

**Logic:**
1. Query `ASCEND_RT_VISIBLE_DEVICES` and `CUDA_VISIBLE_DEVICES` environment variables
2. Parse the comma-separated list of physical device IDs
3. Look up the physical ID returned by `ray.get_accelerator_ids()` in the list
4. Use its index as the logical `LOCAL_RANK`
5. If the physical ID is not in the list (e.g., the env var uses UUIDs), fall through unchanged

**Compatibility:** This change is backward-compatible:
- When `*_VISIBLE_DEVICES` is unset, the loop body is skipped entirely — behavior unchanged
- When `*_VISIBLE_DEVICES` maps physical IDs 1:1 (e.g., `0,1,2,3`), `index()` returns the same value — no effect
- The `ValueError` catch ensures graceful degradation for non-integer visibility values (e.g., GPU UUIDs on ROCm)

- [x] I have read the [Contributing Guidelines](../CONTRIBUTING.md)
- [x] My code is consistent with the existing code style and passes `pre-commit`
- [x] I have added necessary tests (covered by existing worker initialization path)
- [x] All existing tests pass
- [x] I have verified this fix on Ascend NPU hardware

### DIFF
--- a/verl/single_controller/base/worker.py
+++ b/verl/single_controller/base/worker.py
@@ -277,6 +277,20 @@ class Worker(WorkerHelper):
             # so we need to set local rank when the flag is set.
             device_name = get_resource_name()
             local_rank = ray.get_runtime_context().get_accelerator_ids()[device_name][0]
+            # Map physical device ID to logical device index when
+            # ASCEND_RT_VISIBLE_DEVICES (or CUDA_VISIBLE_DEVICES) is set.
+            # Ray reports physical accelerator IDs, but the runtime only sees
+            # the subset of devices listed in the visibility env var.
+            ascend_visible = os.environ.get("ASCEND_RT_VISIBLE_DEVICES", "")
+            cuda_visible = os.environ.get("CUDA_VISIBLE_DEVICES", "")
+            for visible_env in (ascend_visible, cuda_visible):
+                if visible_env:
+                    visible_ids = [int(x.strip()) for x in visible_env.split(",") if x.strip()]
+                    try:
+                        local_rank = str(visible_ids.index(int(local_rank)))
+                    except ValueError:
+                        pass
+                    break
             os.environ["LOCAL_RANK"] = local_rank
             get_torch_device().set_device(int(local_rank))
 


verl/single_controller/base/worker.py
```diff
@@ -277,6 +277,20 @@ def _setup_env_cuda_visible_devices(self):
             # so we need to set local rank when the flag is set.
             device_name = get_resource_name()
             local_rank = ray.get_runtime_context().get_accelerator_ids()[device_name][0]
+            # Map physical device ID to logical device index when
+            # ASCEND_RT_VISIBLE_DEVICES (or CUDA_VISIBLE_DEVICES) is set.
+            # Ray reports physical accelerator IDs, but the runtime only sees
+            # the subset of devices listed in the visibility env var.
+            ascend_visible = os.environ.get("ASCEND_RT_VISIBLE_DEVICES", "")
```
wuxibin89, line 284: Why ASCEND_RT_VISIBLE_DEVICES select non-contiguous value?

GJWu-zyx: Thanks for the review!
In a shared development environment, other processes were already occupying cards 0-3, so we had to use cards 4,5,6,7 for our training job. More generally, ASCEND_RT_VISIBLE_DEVICES is designed to let users select any subset of physical cards — there's no guarantee users will always get contiguous cards starting from 0. The existing code implicitly assumes that, and calling set_device() with a physical ID that doesn't map to a valid logical index causes a crash. This patch adds the physical-to-logical mapping so that any valid ASCEND_RT_VISIBLE_DEVICES value works correctly.

PR #7393, opened 2026-08-13
[vllm] fix: handle request aborts on multi-node replicas

lxb007981:
### What does this PR do?
1. Multi-node vLLM replicas only create an AsyncLLM instance on node rank 0. Makes abort_all_requests() and abort_request() no-ops on nonzero node ranks.
2. Propagates abort failures instead of silently continuing into weight synchronization.
3. Removes a redundant wait_for_requests_to_drain() call in release_kv_cache(), as vLLM’s awaited pause_generation() already completes the distributed abort operation. wait_for_requests_to_drain() merely polls the separate engines_running flag, with a 300-second timeout. That flag can lag behind the completed pause.

### Checklist Before Starting
* [x]  Search for similar PRs. Paste at least one query link here: ...
* [x]  Format the PR title as `[{modules}] {type}: {description}` (This will be checked by the CI)

  * `{modules}` include `fsdp`, `megatron`, `veomni`, `sglang`, `vllm`, `rollout`, `trainer`, `ci`, `training_utils`, `recipe`, `hardware`, `deployment`, `ray`, `worker`, `single_controller`, `misc`, `perf`, `model`, `algo`, `env`, `tool`, `ckpt`, `doc`, `data`, `cfg`, `reward`, `fully_async`, `one_step_off`
  * If this PR involves multiple modules, separate them with `,` like `[megatron, fsdp, doc]`
  * `{type}` is in `feat`, `fix`, `refactor`, `chore`, `test`
  * If this PR breaks any API (CLI arguments, config, function signature, etc.), add `[BREAKING]` to the beginning of the title.
  * Example: `[BREAKING][fsdp, megatron] feat: dynamic batching`

```python
# Add code snippet or script demonstrating how to use this
```

### Checklist Before Submitting
Important

Please check all the following items before requesting a review, otherwise the reviewer might deprioritize this PR for review.

* [x]  Read the [Contribute Guide](https://github.com/verl-project/verl/blob/main/CONTRIBUTING.md).
* [x]  Apply [pre-commit checks](https://github.com/verl-project/verl/blob/main/CONTRIBUTING.md#code-linting-and-formatting): `pre-commit install && pre-commit run --all-files --show-diff-on-failure --color=always`
* [ ]  Add / Update [the documentation](https://github.com/verl-project/verl/tree/main/docs).
* [ ]  Add unit or end-to-end test(s) to [the CI workflow](https://github.com/verl-project/verl/tree/main/.github/workflows) to cover all the code. If not feasible, explain why: ...
[...]

PR #7511, opened 2026-08-21
[vllm, rollout] fix: gate rollout submission during weight-sync drain for DP>1

EricMarcus-ai:
### What does this PR do?

During a weight update, verl pauses the vLLM engines, waits for idle, swaps the weights,
and resumes. The pause is vLLM's `pause_generation`, and everything turns on what that actually
does: it stops the scheduler from running new requests, but it does not stop the engine from
accepting them. That is deliberate and documented on vLLM's side ("all pause modes queue new adds") 
(not vllm version dependent). verl treats the pause as though nothing more can arrive, and nothing on the verl side stops a rollout request from being submitted while paused, and one that arrives before the drain's first check sets engines_running with nothing left to clear it. 
Problem: those requests are never executed; it sets engines_running, which is the flag the drain polls, and only a wave_complete from rank 0 clears it. That never comes while paused, so the drain times out and raises TimeoutError.

This only happens when dp>1, because in vllm the drain, [see here](https://github.com/vllm-project/vllm/blob/6e448d0ea9bf3d88d898b65449ca6dc2aec170ac/vllm/v1/engine/async_llm.py#L1004-L1018), waits for `self.engine_core.dp_engines_running()`, which is set [here](https://github.com/vllm-project/vllm/blob/6e448d0ea9bf3d88d898b65449ca6dc2aec170ac/vllm/v1/engine/core_client.py#L904-L906), contingent on `self.is_dp`. (footnote: this is the 'sync' version of add request, but the async version also sets the flag, just a bit deeper and less obviously linkable).

The abort replicas [is](https://github.com/verl-project/verl/blob/7a830f6adb64839a5d798c0e5a790f5314a94993/verl/checkpoint_engine/base.py#L477-L480) a barrier; vllm pauses internally after completion of a full process engine step. For larger scale models this step takes longer and we have to be aligned between replicas, meaning larger model scale 'widens' the window between pause generation (that aborts) and the first dp engines running check (i.e. this error is seen more likely on large scale model with dp>1).

#### Relationship to #7393

#7393 deletes the drain, on the grounds that the flag "can lag behind the completed pause". That lag
is the symptom above. Deleting it does stop the crash, but that drain is also the only check that the
engines went quiet before verl writes weights into the live weight buffers. It is possible to go this deletion 
route, but it does allow the above 'orphan' requests and skips this waiting step, something that, in my opinion, might come back to bite us later in some way.

Maybe @HollowMan6 (or someone else) has an opinion on this?

### Checklist Before Starting

- [x] Search for similar PRs:
  - https://github.com/verl-project/verl/pulls?q=is%3Apr+wait_for_requests_to_drain (returns #7393 only)
  - https://github.com/verl-project/verl/pulls?q=is%3Apr+pause_generation

### Test

Added some small unit tests about the sequence ordering, they don't literally reproduce the error, since it is a large scale training race. (let me know if you rather delete them)

```bash
pytest -q tests/workers/rollout/rollout_vllm/test_submitter_gate_on_cpu.py
# with the change:    5 passed
# change reverted:    3 failed, 2 passed
```

### API and Usage Example

None. 

### Design & Code Changes

Two new fields in `vLLMHttpServer`:
`_submission_paused` is a gate on rollout submission, and `_admitting` counts requests that have
passed the gate but not yet reached the engine.

- `abort_all_requests()`: close the gate, wait for `_admitting` to reach zero, then call
  `pause_generation()` as before. Zero means every request that exists is already at the engine, so
  the abort catches all of them (nothing enters after anymore). 
- `generate()`: parks while the gate is closed instead of rejecting, and counts itself in
  `_admitting` until the engine has the request.
- `resume_generation()`: reopens the gate, above the `node_rank != 0` return so every server reopens.

- [x] Documentation: n/a, no user-facing surface.
- [x] Add unit test(s): `test_submitter_gate_on_cpu.py`.
- [ ] Send a message in the `ci-request` channel once ready for CI.

### DIFF
--- a/tests/workers/rollout/rollout_vllm/test_submitter_gate_on_cpu.py
+++ b/tests/workers/rollout/rollout_vllm/test_submitter_gate_on_cpu.py
@@ -1,0 +1,152 @@
+# Copyright 2026 Bytedance Ltd. and/or its affiliates
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""GPU-free tests for the vLLMHttpServer submission gate.
+
+vLLM's pause stops requests being scheduled but still accepts them. A request
+admitted between abort_all_requests() and resume_generation() is parked in the
+scheduler's waiting queue and masked out of the drain's liveness check, so
+wait_for_requests_to_drain() cannot return. These tests pin the ordering that
+makes such an admission impossible.
+"""
+
+import asyncio
+from types import SimpleNamespace
+
+import pytest
+
+pytest.importorskip("ray")
+pytest.importorskip("vllm")
+
+from verl.workers.rollout.vllm_rollout import vllm_async_server
+
+
+class _FakeEngine:
+    """Records the state of the gate at the moment the engine is paused."""
+
+    def __init__(self):
+        self.output_processor = SimpleNamespace(request_states={})
+        self.server = None
+        self.pause_calls = 0
+        self.resume_calls = 0
+        self.admitting_at_pause = None
+
+    async def pause_generation(self, **kwargs):
+        self.pause_calls += 1
+        self.admitting_at_pause = self.server._admitting
+
+    async def resume_generation(self):
+        self.resume_calls += 1
+
+
+def _make_server(node_rank: int = 0):
+    server = object.__new__(vllm_async_server.vLLMHttpServer)
+    server.node_rank = node_rank
+    server.engine = _FakeEngine()
+    server.engine.server = server
+    server._submission_paused = False
+    server._admitting = 0
+    server._resume_event = asyncio.Event()
+    server._resume_event.set()
+    return server
+
+
+def test_abort_does_not_pause_until_inflight_admissions_land():
+    async def main():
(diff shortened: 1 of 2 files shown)


wuxibin89:
@aoshen02 Please help to take a look.

EricMarcus-ai:
> @aoshen02 Please help to take a look.

@wuxibin89 @aoshen02 Any updates? It will arise in any large-scale training currently.

aoshen02:
> > @aoshen02 Please help to take a look.
> 
> @wuxibin89 @aoshen02 Any updates? It will arise in any large-scale training currently.

Thanks, we will merge https://github.com/vllm-project/vllm/pull/51488 as soon as possible. Please take a look if that works for you, will take a deeper look tomorrow. related context: https://github.com/vllm-project/vllm/pull/51481

EricMarcus-ai:
> > > @aoshen02 Please help to take a look.
> > 
> > 
> > @wuxibin89 @aoshen02 Any updates? It will arise in any large-scale training currently.
> 
> Thanks, we will merge [vllm-project/vllm#51488](https://github.com/vllm-project/vllm/pull/51488) as soon as possible. Please take a look if that works for you, will take a deeper look tomorrow. related context: [vllm-project/vllm#51481](https://github.com/vllm-project/vllm/pull/51481)

Thanks @aoshen02 ! I think 51481 is the same issue. So, if I understand correctly, if both are merged, we'd get an EnginePausedError when we send submissions while the engines are paused? That would also work, provided we'd catch it appropriately and make sure all resuming, etc., works as intended.

Only practical problem I see: the issue at hand will still exist on any vllm version that does not have those two PRs merged. It is currently not even in any released version yet. Do we need to do something until then @wuxibin89 ? (or just pray they find this thread?)

wuxibin89:
@EricMarcus-ai Got it, let's fix it in verl side and remove it once we bump to fixed vllm version.

wuxibin89 (approved):

PR #7524, opened 2026-08-23
[megatron] fix: distribute R2 replay padding across EP ranks

EazyReal:
### What does this PR do?

Megatron R2 replays the recorded MoE routes during the actor update. THD/BSHD preprocessing can add loss-masked alignment rows, but their zero-filled routes still enter MoE dispatch because R2 has no replay mask. Every top-k slot on those rows therefore selects expert 0.

This change carries a validity marker through the same preprocessing layout, then replaces only final-layout padding routes after CP packing and before SP scatter. Assignments cycle through EP ranks before local expert slots, so per-rank load differs by at most one, full cycles visit every expert once, and each row keeps distinct top-k experts. Recorded routes are unchanged. R3 remains on its existing masked path.

PR #7106 fixes R2 THD alignment and local/VPP router selection; it reconstructs alignment routes as zeros. This PR is complementary: it prevents those reconstructed padding routes from concentrating dispatch on expert 0. No other open PR found by the searches below addresses that load imbalance.

### Checklist Before Starting

- [x] Searched open PRs: [router replay padding](https://github.com/verl-project/verl/pulls?q=is%3Apr+is%3Aopen+%22router+replay%22+padding)
- [x] Title follows `[{modules}] {type}: {description}`.

### Test

- `VERL_USE_EXTERNAL_PLUGINS=none python -m pytest -q tests/utils/megatron/test_router_replay_model_walk_on_cpu.py`: 6 passed.
- `VERL_USE_EXTERNAL_PLUGINS=none python -m pytest -q tests/utils/test_megatron_router_replay_dcp.py`: 4 passed.
- Targeted `pre-commit run --files ...` for all three changed files: all hooks passed.
- `git diff --check`: passed.

The THD regression uses the pinned Megatron-Core preprocessor and observes the tensor immediately before SP scatter. For `E=64`, `EP=32`, `K=8`, and four padding rows, all 32 EP ranks receive exactly one assignment. The focused unit cases cover partial prefixes, 32- and 64-way EP, multiple layers, valid-route preservation, per-row uniqueness, and a full expert cycle.

An exact-helper property probe checked 6,852 layouts and 20,088 per-layer histograms across 8-256 experts, every power-of-two EP divisor, top-k 1/2/4/8, and 0 through `min(2E/K + 3, 64)` padding rows without a counterexample.

A production-layout CPU probe with Megatron-Core 0.18, FP8 alignment, TP=4, CP=2, 48 layers, and 240 padding rows measured:

- before: 92,160/92,160 padding assignments routed to expert 0;
- after: all 256 experts received 7-8 assignments per layer, and all 32 EP ranks received exactly 60 per layer;
- all non-padding routes were unchanged.

No multi-GPU training run was performed. The installed Megatron-Core does not expose `PackedSeqParams.cp_partition_mode`, so contiguous CP cannot be exercised in this environment. The affected static-zigzag packing and dispatch-input boundary is covered with the repository's pinned Megatron-Core on CPU; distributed execution remains for upstream CI.

### API and Usage Example

No public API or configuration changes.

### Design & Code Changes

- Preserve row validity through the existing R2 preprocessing path.
- Distribute only alignment-row routes before SP scatter.
- Cover the assignment invariant and the real THD dataflow boundary.

- [x] Applied targeted pre-commit checks to every changed file.
- [x] No documentation change is needed; behavior and configuration are unchanged.
- [x] Added CPU regression coverage for the affected production path.
- [ ] Request multi-GPU CI after maintainer authorization.

AI assistance was used to audit the dataflow, implement the change, and prepare the tests and description. The human submitter reviewed every changed line, ran the relevant tests, and understands the change, as required by `AGENTS.md`.

### DIFF
--- a/tests/utils/megatron/test_router_replay_model_walk_on_cpu.py
+++ b/tests/utils/megatron/test_router_replay_model_walk_on_cpu.py
@@ -97,7 +97,13 @@ def tf_config(monkeypatch):
     monkeypatch.setattr(router_replay_utils, "device_name", "cpu")
     monkeypatch.setattr(router_replay_utils, "preprocess_packed_seqs", lambda x, mask, **kwargs: (x, None))
     monkeypatch.setattr(router_replay_utils, "scatter_to_sequence_parallel_region", lambda x: x)
-    return SimpleNamespace(fp8=None, num_layers=NUM_LAYERS, moe_layer_freq=1)
+    return SimpleNamespace(
+        fp8=None,
+        num_layers=NUM_LAYERS,
+        num_moe_experts=8,
+        expert_model_parallel_size=1,
+        moe_layer_freq=1,
+    )
 
 
 def test_targets_are_written_to_the_forwarded_models_own_routers(orphans_then_model, tf_config):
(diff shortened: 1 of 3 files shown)


wuxibin89:
> THD/BSHD preprocessing can add loss-masked alignment rows, but their zero-filled routes still enter MoE dispatch because R2 has no replay mask. Every top-k slot on those rows therefore selects expert 0.

It's safe that padding tokens always selecting expert 0 since their loss will be masked out.

EazyReal:
@wuxibin89 but even if they are masked out they still take memories?

PR #7565, opened 2026-08-26
[rollout] fix: surface vLLM prefix-cache hit counts in TokenOutput

emmericp:
### What does this PR do?

Prefix-cache hit counts are currently invisible to `TokenOutput` consumers:

1. the vLLM async server never exposes vLLM's per-request `num_cached_tokens`, and
2. `FullyAsyncLLMServerClient.generate` rebuilds a fresh `TokenOutput` across partial-rollout resume iterations, so fully-async mode always reports 0 cache hits even once (1) is fixed.

This PR sets `extra_fields["num_cached_tokens"]` on every vLLM generate result (consumers can surface it as OpenAI `usage.prompt_tokens_details.cached_tokens`), and carries the initial prefill's hit count forward across resume iterations, matching single-prefill semantics.

- **Duplicate check**: `gh pr list --state open --search "num_cached_tokens"` / `"prefix cache"` — no open PR exposes cache-hit counts in TokenOutput (hits are unrelated: KV-cache-aware load balancing, extra prefix cache).
- **Tests**: plumbing only, no unit test; validated in agent-loop training runs where cache-hit stats went from constant 0 to expected values (custom OpenAI-compatible endpoint reporting `cached_tokens`).

### DIFF
--- a/verl/workers/rollout/llm_server.py
+++ b/verl/workers/rollout/llm_server.py
@@ -400,6 +400,12 @@ class FullyAsyncLLMServerClient(LLMServerClient):
             num_preempted=0,
         )
         min_global_steps, max_global_steps = None, None
+        # Prefix-cache hits are reported per prefill. The base client returns the
+        # server's TokenOutput directly (so sync mode surfaces num_cached_tokens),
+        # but here we rebuild a fresh TokenOutput across resume iterations, so we
+        # must carry it forward explicitly or the consumer sees 0. Take the first
+        # (initial-prompt) prefill's hit count, matching single-prefill semantics.
+        num_cached_tokens = None
 
         while True:
             # 1. generate tokens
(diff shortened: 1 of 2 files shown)


verl/workers/rollout/llm_server.py
```diff
@@ -458,6 +468,7 @@ async def generate(
         final_output.extra_fields["global_steps"] = global_steps
         final_output.extra_fields["min_global_steps"] = min_global_steps
         final_output.extra_fields["max_global_steps"] = max_global_steps
+        final_output.extra_fields["num_cached_tokens"] = num_cached_tokens
```
wuxibin89, line 471: Where is `num_cached_tokens` consumed?

emmericp: Not in the default agent loops, I'm using it in a custom agent loop for custom statistics. Without this patch there is no good way to get that information at all.

wuxibin89 (approved):

PR #7566, opened 2026-08-26
[fully_async] fix: only require rollout log-probs when they anchor PPO

emmericp:
### What does this PR do?

The fully-async rollouter unconditionally asserts `actor_rollout_ref.rollout.calculate_log_probs=True`, but rollout log-probs are only actually consumed when `algorithm.rollout_correction.bypass_mode=True` (rollout log-probs as the PPO anchor). When the trainer recomputes log-probs (`bypass_mode=False`), forcing rollout-side log-prob calculation is pure overhead.

Replace the blanket assert with `validate_rollout_log_prob_config()`, which validates the actual dependency: `bypass_mode=True` requires `calculate_log_probs=True`; otherwise nothing is required.

- **Duplicate check**: `gh pr list --state open --search "calculate_log_probs"` returns no open PR.
- **Tests**: `pytest tests/experimental/fully_async_policy/test_rollout_log_prob_config.py` — 3 passed.

### DIFF
--- a/tests/experimental/fully_async_policy/test_rollout_log_prob_config.py
+++ b/tests/experimental/fully_async_policy/test_rollout_log_prob_config.py
@@ -1,0 +1,52 @@
+# Copyright 2025 Meituan Ltd. and/or its affiliates
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


wuxibin89:
```bash
trainer.use_v1=True
```

PR #7568, opened 2026-08-26
[fully_async] fix: stamp rollouts with the parameter version as their step

emmericp:
### What does this PR do?

In fully-async training, rollouts do not record which policy version generated them:

- training rollouts never get `meta_info["global_steps"]` set at all, and
- validation rollouts inherit the rollouter's `self.global_steps`, which is a per-sample feed counter — a validation rollout can record "step 210" while the policy is at parameter version 4.

Trajectory descriptors and downstream staleness accounting need the policy-weight version a rollout starts sampling against. This PR tracks `current_param_version` in the rollouter (pushed by the trainer on every weight sync via `reset_staleness(...)` / `do_validate(...)`, backward-compatible default 0), stamps it into `meta_info["global_steps"]` for training rollouts, and swaps it into `self.global_steps` for the duration of `_validate()` so validation gen batches record the policy version too.

- **Duplicate check**: `gh pr list --state open --search "param version"` / `"global_steps rollouter"` — no open PR addresses this.
- **Tests**: Ray-actor plumbing, no unit test; validated in fully-async runs — trajectory step now matches parameter version for both training and validation rollouts.

### DIFF
--- a/verl/experimental/fully_async_policy/fully_async_rollouter.py
+++ b/verl/experimental/fully_async_policy/fully_async_rollouter.py
@@ -446,6 +446,11 @@ class FullyAsyncRollouter(SeparateRayPPOTrainer):
         self._completed_steps: int = 1
         # we start from step 1
         self.global_steps = 1
+        # Parameter (policy-weight) version, set by the trainer on each weight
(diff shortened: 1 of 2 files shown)


wuxibin89:
```bash
trainer.use_v1=True
```

PR #7581, opened 2026-08-27
[algo] fix: keep single-sample groups out of the rloo_vectorized zero path

alanhuangyoo:
### What does this PR do?

`rloo` and `rloo_vectorized` are documented as interchangeable `adv_estimator` values (`docs/examples/config.rst`), and `test_rloo_and_vectorized_equivalence` asserts they agree. They don't, on groups of one.

`compute_rloo_outcome_advantage` only rewrites a score when its group holds more than one sample:

```python
response_num = len(id2score[index[i]])
if response_num > 1:
    scores[i] = scores[i] * response_num / (response_num - 1) - id2mean[...] * ...
```

so a singleton keeps its raw score. The vectorized form masks with `(c > 1)` instead:

```python
adv = ((c * scores - torch.bincount(inv, weights=scores)[inv]) / (c - 1).clamp_min(1)) * (c > 1)
```

which zeroes those rows.

The case that matters is `rollout.n == 1`: every group is then a singleton, so `rloo_vectorized` returns an all-zero advantage tensor and the policy receives no gradient at all. Nothing validates `n` against the estimator choice, and no error is raised — the run just doesn't learn.

Ragged groups hit it too, e.g. when filtering leaves one surviving sample in a group.

### Test

Both estimators, same inputs, `max|adv_ref - adv_vec|`:

before
```
  uniform_groups_of_4    maxdiff=2.384e-07  OK
  singleton_group        maxdiff=1.085e+00  <<< DIFFERS
      ref[:, -1] = [2.1756, -0.2703, -2.7841, 0.8788, -1.0845]
      vec[:, -1] = [2.1756, -0.2703, -2.7841, 0.8788,  0.0]
  all_singletons         maxdiff=2.179e+00  <<< DIFFERS      <- rollout.n == 1
      ref[:, -1] = [1.541, -0.2934, -2.1788, 0.5684]
      vec[:, -1] = [0.0,    0.0,     0.0,    0.0]
  ragged_groups          maxdiff=1.399e+00  <<< DIFFERS
      ref[:, -1] = [2.7771, 0.0255, -2.8026, 1.653, -1.653, -1.3986]
      vec[:, -1] = [2.7771, 0.0255, -2.8026, 1.653, -1.653,  0.0]
  zero_variance_group    maxdiff=0.000e+00  OK
  string_ids             maxdiff=0.000e+00  OK
  partial_mask           maxdiff=0.000e+00  OK
```

after — every case agrees:
```
  uniform_groups_of_4    maxdiff=2.384e-07  OK
  singleton_group        maxdiff=2.384e-07  OK
  all_singletons         maxdiff=0.000e+00  OK
  ragged_groups          maxdiff=0.000e+00  OK
  zero_variance_group    maxdiff=0.000e+00  OK
  string_ids             maxdiff=0.000e+00  OK
  partial_mask           maxdiff=0.000e+00  OK
```

The same sweep against the GRPO pair is `0.000e+00` on every case both before and after — `compute_grpo_outcome_advantage` gives singletons `mean=0, std=1`, so the raw score survives there, and `group_mean_std` mirrors it. That is the convention this PR restores for RLOO.

Test file:
```
tests/trainer/ppo/test_core_algos_on_cpu.py

  before the fix   3 failed, 4 passed   (the three RLOO singleton cases)
  after            30 passed
```

`ruff==0.12.2` (the pinned pre-commit rev) `check` and `format --check`: clean on both files.

### Design & Code Changes

`torch.where(c > 1, loo, scores)` instead of `loo * (c > 1)`. Groups of two or more are untouched — the leave-one-out arithmetic is unchanged — so this only moves the singleton rows back onto the reference implementation's behaviour.

`test_rloo_and_vectorized_equivalence` cannot catch this: `_make_group_index` is documented as "ensuring each group has at least 2 samples" and asserts `num_groups * 2 <= batch_size`, so the one group size where a leave-one-out baseline does not exist is never generated. Added singleton and ragged cases for both the RLOO and GRPO pairs.


alanhuangyoo:
No checks have run on this one — the workflows look to be held for approval since this is my first PR here. Could a maintainer release them?

The change in one line: `compute_rloo_outcome_advantage` and `compute_rloo_vectorized_outcome_advantage` are both registered under `AdvantageEstimator`, and they disagree on groups of one.

Reproduced on a clean checkout of current `main` (`bd6f5645`), scalar vs vectorized on the same inputs:

```
group sizes            max |scalar - vectorized|
[1, 1, 1, 1]           2.963
[1]                    0.363
[3, 1, 2]              0.786
[4, 4]                 0        (agree)
[8]                    0        (agree)
[2, 2, 2, 2]           0        (agree)
```

Only the sizes where a leave-one-out baseline does not exist diverge. `_make_group_index` in the existing equivalence test gives every group >= 2 samples, so the current test can never see this — the added parametrization covers `rollout.n == 1` and the case where filtering leaves a group with one surviving sample.

alanhuangyoo:
@wuxibin89 — a workflow-approval request rather than a review one, if you have a moment.

These four have had no CI run at all since they were opened; `check-runs` and `statuses` are both 0 on every head commit, so they sit at `BLOCKED` on checks that have never started rather than on checks that failed. #7620 did get a run, so I know it is per-PR rather than a standing block on the account.

| | opened | size | what it fixes |
| --- | --- | --- | --- |
| #7581 | Aug 27 | +64/-1 | `rloo_vectorized` returns an all-zero advantage at `rollout.n == 1`, so the policy gets no gradient and the run silently does not learn. `rloo` and `rloo_vectorized` are documented as interchangeable. |
| #7582 | Aug 27 | +74/-14 | `entropy_from_logits_with_chunking` is ignored in the no-remove-padding branch of all three engines, so the flag does nothing where it is most needed for memory. |
| #7691 | Sep 2 | +108/-1 | Ulysses slices `position_ids` unconditionally while the guard only checks `inputs_embeds`, so a caller that omits `position_ids` crashes instead of running. |
| #7711 | Sep 3 | +113/-1 | With bf16 params and fp32 reduce, FSDP2 never hits the early return in `to_accumulated_grad_if_needed`, so a full unsharded fp32 gradient is retained per parameter across the accumulation window. |

No rush on reviewing them. I would just like to know whether they pass CI before taking more of anyone's time.

One thing I fixed on my side while checking this: #7582's title used `[engine]`, which is not in `allowed_modules`, so `check_pr_title.py` would have failed it. Retitled to `[worker]`. I ran both sanity checkers over all four locally and the descriptions pass.

alanhuangyoo:
@wuxibin89 This is a process note, not a review request: **CI has never run on this PR or on my other two open ones.** You merged #7620, so I'm bringing it to you.

Every workflow run on these branches is sitting at `action_required`. The most recent push was 2026-09-08:

```
e2e_ppo_trainer_megatron_vllm   conclusion=action_required
pre-commit                      conclusion=action_required
doc_test                        conclusion=action_required
...
```

#7620 ran its full 85 checks, so I think this repo needs a maintainer to approve each fork PR's workflows, and nobody has approved these three yet. None of them has a check result, so nobody can review them against a green run.

Before writing this, I checked that all three still apply to current `main`. They are 11 commits behind:

| PR | touches | changed on `main` since | state |
| --- | --- | --- | --- |
| **#7581** | `verl/trainer/ppo/core_algos.py` | not touched | `MERGEABLE` |
| **#7582** | `verl/utils/torch_functional.py` + the three engine `transformer_impl.py` | not touched | `MERGEABLE` |
| **#7691** | `verl/models/transformers/monkey_patch.py` | once, #7703 (fused output head), which does not touch the `position_ids` slicing | `MERGEABLE` |

What each one fixes:

- **#7581 (this PR).** `rloo_vectorized` sends single-sample groups into the zero-advantage path. That is RLOO's advantage estimate itself, not plumbing.
- **#7582.** `entropy_from_logits_with_chunking` is ignored when `remove_padding` is off.
- **#7691.** Ulysses input slicing assumes `position_ids` is always present.

Each change is under 110 lines and ships with its own CPU test. If approving the workflows would be easier after a rebase onto `main`, I can push one. I've held off so I don't trigger three more `action_required` runs for nothing.

PR #7582, opened 2026-08-27
[worker] fix: honor entropy_from_logits_with_chunking when remove_padding is off

alanhuangyoo:
### What does this PR do?

`entropy_from_logits_with_chunking` is a config flag (`actor_rollout_ref.actor.entropy_from_logits_with_chunking`) that swaps `entropy_from_logits` for a lower-memory chunked version. The engines resolve it once in `__init__`:

```python
if self.engine_config.entropy_from_logits_with_chunking:
    entropy_from_logits = verl_F.entropy_from_logits_with_chunking
else:
    entropy_from_logits = verl_F.entropy_from_logits
self.compute_entropy_from_logits = ...
```

The remove-padding branch uses `self.compute_entropy_from_logits`. The branch that runs **without** remove-padding does not — it calls `verl_F.entropy_from_logits` directly:

| engine | remove-padding branch | no-remove-padding branch |
| --- | --- | --- |
| `fsdp/transformer_impl.py` | `self.compute_entropy_from_logits` (L1367) | `verl_F.entropy_from_logits` (L1454) |
| `torchtitan/transformer_impl.py` | `self.compute_entropy_from_logits` (L823) | `verl_F.entropy_from_logits` (L839) |
| `automodel/transformer_impl.py` | `self.compute_entropy_from_logits` (L628) | `verl_F.entropy_from_logits` (L663) |

So with `use_remove_padding=False` the flag does nothing: the full non-chunked reduction runs, and the memory the user asked to save is not saved. Nothing warns.

Routing the flag through that branch is not sufficient on its own. It holds `(bsz, response_length, vocab_size)` logits, and the chunked implementation indexes dim 0 into a 1-D buffer:

```python
entropy = torch.zeros(logits.shape[0], device=logits.device)
for i in range(0, logits.shape[0], chunk_size):
    ...
    entropy[i : i + chunk_size] = entropy_chunk
```

which on a 3-D input raises

```
RuntimeError: expand(torch.FloatTensor{[2, 7]}, size=[2]): the number of sizes provided (1) must be
greater or equal to the number of dimensions in the tensor (2)
```

while `entropy_from_logits` takes any `(..., vocab_size)`.

`torch.zeros(...)` also has no `dtype`, so the chunked path always returned float32 — flipping the flag flipped the entropy tensor's dtype, and a float64 chunk was silently truncated on assignment.

### Test

`entropy_from_logits` vs `entropy_from_logits_with_chunking`, same input:

before
```
  2D float32                    maxdiff=0.000e+00   OK
  2D float64                    maxdiff=1.550e-06   DTYPE torch.float64 vs torch.float32
  2D bfloat16                   maxdiff=5.720e-02   DTYPE torch.bfloat16 vs torch.float32
  2D float16                    maxdiff=3.891e-03   DTYPE torch.float16 vs torch.float32
  3D float32                    RuntimeError: expand(torch.FloatTensor{[2, 7]}, size=[2])
  2D large, crosses chunk 2048  maxdiff=0.000e+00   OK
```

after — shapes and dtypes match everywhere, 3-D works:
```
  2D float32                    maxdiff=0.000e+00   OK
  2D float64                    maxdiff=1.550e-06   OK
  2D bfloat16                   see below
  2D float16                    see below
  3D float32                    maxdiff=0.000e+00   OK
  2D large, crosses chunk 2048  maxdiff=0.000e+00   OK
```

bfloat16 is the one case the two cannot match to their own resolution, because `entropy_from_logits` runs the whole reduction in bfloat16 while the chunked path accumulates in float32. Measured against the float32 value, the chunked path is the accurate one:

```
shape        |chunked - fp32|   |unchunked - fp32|
(5000, 16)         0.0041             0.1108
(8, 17)            0.0010             0.0682
(1, 1024)          0.0007             0.0475
```

so the test holds the chunked path to the float32 reference there rather than to the noisier twin.

Tests added to `tests/utils/test_torch_functional.py`, following `test_calculate_sum_pi_squared_from_logits`'s shape/dtype parametrization:

```
  before the fix   9 failed, 4 passed
  after            13 passed

regression
  tests/utils/test_torch_functional.py -k "not distributed"   24 passed
  tests/trainer/ppo/test_core_algos_on_cpu.py                 30 passed
```

`ruff==0.12.2` (the pinned pre-commit rev) `check` and `format --check`: clean on all five files.

### Design & Code Changes

- `entropy_from_logits_with_chunking`: flatten the leading dimensions so chunking runs over tokens and the `(..., vocab_size)` contract matches `entropy_from_logits`; allocate the output in the input dtype. The per-chunk float32 accumulation is unchanged.
- The three no-remove-padding branches now call `self.compute_entropy_from_logits`, mirroring the remove-padding branch exactly (including the `chunk_size` argument and the `entropy_checkpointing` wrapper).

I could only exercise the `verl/utils/torch_functional.py` half locally — the engine branches need a GPU worker to reach. Those three hunks are a mechanical swap onto the same call the remove-padding branch in the same file already makes.


alanhuangyoo:
Same here — no checks have run, the workflows appear to be held for first-time contributor approval. Could a maintainer release them?

Summary: `actor_rollout_ref.actor.entropy_from_logits_with_chunking` is resolved once in each engine's `__init__` into `self.compute_entropy_from_logits`. The remove-padding branch uses it; the branch that runs **without** remove padding calls `verl_F.entropy_from_logits` directly, so the flag is silently ignored there and the memory saving it exists for does not happen.

Three engines have the same shape (`fsdp`, `automodel`, `torchtitan`), which is why the fix touches all of them rather than one.

PR #7587, opened 2026-08-27
[vllm] fix: Use dedicated Ray concurrency group for vLLMHttpServer control methods

rao-ashish:
### What does this PR do?

This PR prevents vLLM rollout control RPCs from being starved by `generate()` calls.

Ray limits the number of concurrent calls admitted to an async actor. During highly asynchronous rollouts, generation calls can occupy every slot in the vLLM server's default concurrency group. Control calls such as `wake_up()` and `abort_all_requests()` then cannot start, which can deadlock rollout transitions and weight synchronization.

The fix gives vLLM control-plane methods a separate Ray concurrency group while leaving `generate()` in the default group. Ray runs named concurrency groups on separate threads and asyncio loops, so the new control-method wrapper `_control_method` forwards the actual method body to the server's default loop to preserve thread-safe access to server and vLLM state.

### Checklist Before Starting

- [x] Search for similar PRs. Query: "vLLM concurrency group"

### Test

- Added CPU-only unit test `test_vllm_control_concurrency_on_cpu.py` which verifies control methods are not blocked after saturating `vLLMHttpServer` with `generate()` requests.
    - The new test is included in the vLLM CPU test workflow.
- Verified that this E2E test deadlocks without this fix and does not deadlock with it: https://gist.github.com/rao-ashish/d186729e83434b2e4b5425647abec735

### API and Usage Example

There are no user-facing API or configuration changes.

### Design & Code Changes

- Adds a 16-slot `control` concurrency group to `vLLMHttpServer`.
- Marks vLLM management methods as control-plane RPCs while keeping `generate()` in Ray's default group.
- Executes control-method bodies on the server's original asyncio loop, preserving thread-safe access to shared server and engine state.
- Removes the control-slot margin from vLLM's default-group concurrency limit because control calls now have independent capacity. Other rollout backends retain their existing behavior.
- Adds CPU-only metadata and runtime regression coverage and includes it in CI.

> [!IMPORTANT]

- [x] Add / Update [the documentation](https://github.com/verl-project/verl/tree/main/docs). N/A: this is an internal scheduling fix with no user-facing API.
- [x] Add unit or end-to-end test(s) to [the CI workflow](https://github.com/verl-project/verl/tree/main/.github/workflows) to cover all the code.
- [ ] Once your PR is ready for CI, send a message in the [ci-request channel](https://verl-project.slack.com/archives/C091TCESWB1) in the [verl Slack workspace](https://join.slack.com/t/verl-project/shared_invite/zt-3855yhg8g-CTkqXu~hKojPCmo7k_yXTQ). If unavailable, try the [Feishu group](https://applink.larkoffice.com/client/chat/chatter/add_by_link?link_token=772jd4f1-cd91-441e-a820-498c6614126a).
- [x] If this PR is related to the `recipe` submodule, update its reference with `git submodule update --remote` or `cd recipe && git pull origin main`. N/A.

### DIFF
--- a/.github/workflows/vllm.yml
+++ b/.github/workflows/vllm.yml
@@ -127,6 +127,7 @@ jobs:
             tests/workers/rollout/test_vllm_cli_args_on_cpu.py \
             tests/workers/rollout/test_vision_placeholder_tokens_on_cpu.py \
             tests/workers/rollout/test_vllm_pd_disaggregation_on_cpu.py \
+            tests/workers/rollout/rollout_vllm/test_vllm_control_concurrency_on_cpu.py \
             tests/utils/test_vllm_weight_name_normalization_on_cpu.py::test_update_weights_from_ipc_accumulates_lora_across_buckets \
             tests/utils/test_vllm_weight_name_normalization_on_cpu.py::test_update_weights_from_ipc_standard_loads_per_bucket
       - name: Prepare gsm8k dataset
--- a/tests/utils/test_server_profiler.py
+++ b/tests/utils/test_server_profiler.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import asyncio
 import json
 import os
 import tempfile
@@ -508,6 +509,7 @@ class TestServerProfilerFunctionality(unittest.IsolatedAsyncioTestCase):
 
         # Mock self object
         mock_self = MagicMock()
+        mock_self._default_asyncio_loop = asyncio.get_running_loop()
         mock_self.node_rank = 0
         mock_self.replica_rank = 3
         mock_self.profiler_controller = mock_profiler
@@ -549,6 +551,7 @@ class TestServerProfilerFunctionality(unittest.IsolatedAsyncioTestCase):
         mock_engine = AsyncMock()
 
         mock_self = MagicMock()
+        mock_self._default_asyncio_loop = asyncio.get_running_loop()
         mock_self.node_rank = 0
         mock_self.profiler_controller = mock_profiler
         mock_self.engine = mock_engine
@@ -574,6 +577,7 @@ class TestServerProfilerFunctionality(unittest.IsolatedAsyncioTestCase):
         mock_engine = AsyncMock()
 
         mock_self = MagicMock()
+        mock_self._default_asyncio_loop = asyncio.get_running_loop()
         mock_self.node_rank = 1  # non-master node, should skip
         mock_self.profiler_controller = mock_profiler
         mock_self.engine = mock_engine
@@ -604,6 +608,7 @@ class TestServerProfilerFunctionality(unittest.IsolatedAsyncioTestCase):
         mock_tokenizer_manager = AsyncMock()
 
         mock_self = MagicMock()
+        mock_self._default_asyncio_loop = asyncio.get_running_loop()
         mock_self.profiler_controller = mock_profiler
         mock_self.tokenizer_manager = mock_tokenizer_manager
         mock_self.replica_rank = 0
--- a/tests/workers/rollout/rollout_vllm/test_submitter_gate_on_cpu.py
+++ b/tests/workers/rollout/rollout_vllm/test_submitter_gate_on_cpu.py
@@ -51,6 +51,7 @@ class _FakeEngine:
 
 def _make_server(node_rank: int = 0):
     server = object.__new__(vllm_async_server.vLLMHttpServer)
+    server._default_asyncio_loop = asyncio.get_running_loop()
     server.node_rank = node_rank
     server.engine = _FakeEngine()
     server.engine.server = server
--- a/tests/workers/rollout/rollout_vllm/test_vllm_control_concurrency_on_cpu.py
+++ b/tests/workers/rollout/rollout_vllm/test_vllm_control_concurrency_on_cpu.py
@@ -1,0 +1,98 @@
+# Copyright 2026 Bytedance Ltd. and/or its affiliates
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""
+Test that the _control_method decorator used by vLLMHttpServer schedules
+control methods on a dedicated concurrency group to prevent deadlock.
+
+Usage:
+    pytest tests/workers/rollout/rollout_vllm/test_vllm_control_concurrency_on_cpu.py -v
+"""
+
+import asyncio
+import sys
+
+import pytest
+
+ray = pytest.importorskip("ray")
+ray_cloudpickle = pytest.importorskip("ray.cloudpickle")
+pytest.importorskip("vllm")
+
+from verl.workers.rollout.replica import CONTROL_METHOD_CONCURRENCY  # noqa: E402
+from verl.workers.rollout.vllm_rollout.vllm_async_server import _control_method  # noqa: E402
+
+GENERATE_CONCURRENCY = 8
+RAY_TIMEOUT = 30
+
+
+class _FakeHttpServer:
+    def __init__(self):
+        # Required by _control_method.
+        self._default_asyncio_loop = asyncio.get_running_loop()
+
+        self._num_enqueued_generations = 0
+        self._all_generations_enqueued = asyncio.Event()
+        self._release_generations = asyncio.Event()
+
+    @_control_method
+    async def release_generations(self):
+        await self._all_generations_enqueued.wait()
+
+        assert asyncio.get_running_loop() is self._default_asyncio_loop
+        self._release_generations.set()
+
+    async def generate(self):
+        assert asyncio.get_running_loop() is self._default_asyncio_loop
+
+        self._num_enqueued_generations += 1
+        if self._num_enqueued_generations == GENERATE_CONCURRENCY:
+            self._all_generations_enqueued.set()
+
+        await self._release_generations.wait()
+
+
+@pytest.fixture(scope="module", autouse=True)
+def serialize_test_module_by_value():
+    """Tells Ray to serialize helpers in this file (e.g. _FakeHttpServer) by
+    value to avoid import errors in the actor process."""
+
+    test_module = sys.modules[__name__]
+    ray_cloudpickle.register_pickle_by_value(test_module)
+    yield
+    ray_cloudpickle.unregister_pickle_by_value(test_module)
+
+
+def test_control_rpc_is_not_starved_by_generate():
+    """Test the _control_method decorator using _FakeHttpServer."""
+
+    server = None
+
+    try:
+        server_actor_class = ray.remote(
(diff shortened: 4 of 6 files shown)


rao-ashish:
Hi @wuxibin89 / @PeterSH6 / @HollowMan6, could you take a look at this PR? This fixes a deadlock we're observing when running async-colocated jobs (eg [this config](https://gist.github.com/rao-ashish/d186729e83434b2e4b5425647abec735)) where a large number of generate requests fill up all async task slots on the vLLMHttpServer Ray actor during weight-sync, which in turn starves the 'wake_up' / 'resume' control message, thus leading to a deadlock. The fix is to use separate concurrency groups for generate requests and control requests so that the generate requests don't starve the control requests. Let me know if you have any comments, thanks!

(An earlier CI run caught a test failure due to a test not mocking a newly added attribute of `vLLMHttpServer`. I've fixed this and rebased + squashed commits.)

janbernloehr:
### Confirmation and at-scale validation of the control-method concurrency deadlock

We independently hit the exact failure this PR describes, root-caused it to the same mechanism, and have now validated this PR's patch end-to-end on real hardware. Posting the evidence in case it helps move the review along, since the PR description currently references a configuration gist rather than a reproduced hang.

#### Environment

- verl `main` at `68c9ac3` (this PR's base commit), vLLM 0.24.0, Ray 2.57.0, PyTorch 2.11.0+cu130, CUDA 13.0.2, NCCL 2.28.3-1, megatron-core 0.18.0, TransferQueue 0.1.8, Ubuntu 24.04
- Single node, 8x NVIDIA B300 (compute capability 10.3, 275 GB per GPU, driver 580.167.08). Also reproduces on GB200 NVL and GB300 NVL.

#### Steps to reproduce

1. Run GRPO with the `colocate_async` trainer: Megatron training backend, vLLM rollout, Ray orchestrator.
2. Model `Qwen/Qwen3-30B-A3B` on GSM8K, bf16.
3. Parallelism: TP 8, EP 8, rollout TP 4 — so 2 rollout replicas on 8 GPUs. DeepEP MoE dispatch, Triton MoE backend.
4. Set `train_batch_size` 256, `rollout.n` 8, `max_num_seqs` 1024. Run 5 steps with 2 warmup steps.
5. Let the first step reach its end-of-step weight synchronisation.

The key quantity is `train_batch_size * rollout.n` (2048 concurrent `generate` calls) exceeding `RolloutReplica.max_concurrency`, which is `max(1000, max_num_seqs + CONTROL_METHOD_CONCURRENCY)` = `max(1000, 1024 + 16)` = 1040 per replica.

#### Expected behavior

The run completes all 5 configured steps and reports training metrics. At each step boundary the rollout engines are put to sleep, weights are synchronised into them, and `wake_up` returns so the next step's generation can begin. Control-plane calls such as `wake_up` and `collective_rpc` should always be admittable, regardless of how many `generate` calls are in flight, because they are what allows generation to make progress at all.

#### Actual behavior

The run hangs permanently at the end of a step and is eventually killed by the scheduler's wall clock (3600 s in our setup); it never reaches step 5 and reports no metrics. It is a hard deadlock, not slow progress: repeated py-spy sweeps 31 minutes apart produced byte-for-byte near-identical dumps (548060 / 548060 / 548058 / 548062 / 548058 / 548062 bytes) with no thread ever sampled as active. Raising the time limit does not help.

`ray list tasks --filter "state!=FINISHED"` during the hang reported 2059 non-finished tasks:

- `vLLMHttpServer.generate`: many `RUNNING`, plus several `PENDING_ACTOR_TASK_ORDERING_OR_CONCURRENCY`
- **`vLLMHttpServer.wake_up`: `PENDING_ACTOR_TASK_ORDERING_OR_CONCURRENCY`** — never admitted
- `WorkerDict.<...>_update_weights`: `RUNNING`, blocked in `socket.send_pyobj` at `verl/workers/rollout/vllm_rollout/bucketed_weight_transfer.py` `_init_buffer`, reached from `vllm_rollout.py` `update_weights`
- the trainer task: `RUNNING`, blocked in `ray.get` at `verl/checkpoint_engine/base.py`, reached from the `colocate_async` trainer's `on_step_end`

Both replica actors sat at their concurrency ceiling, and both `ray::vLLMHttpServer.generate` worker processes were completely idle — main thread in Ray's `main_loop`, the asyncio thread parked at bare `threading.run` with no coroutine frames at all. `ray summary tasks` showed `sleep`, `abort_all_requests`, `wait_for_requests_to_drain` and `wake_up` each `FINISHED: 4`, while `GlobalRequestLoadBalancer.acquire_server` showed `FINISHED: 2048`.

#### Why this happens

The load balancer admits all 2048 requests without regard to the per-actor ceiling. The in-flight `generate` calls then occupy every slot of the actor's default concurrency group while awaiting an engine that `sleep()` already put to sleep, so the `wake_up` needed to un-sleep it can never be admitted — a circular wait. The weight-sync path deadlocks identically: the `WorkerDict` side blocks on a ZMQ send whose peer `update_weights_from_ipc` is dispatched via `collective_rpc` on that same starved actor. Because the 16 control slots are additive margin rather than a reserved group, they provide no protection once generation saturates the actor.

#### This PR fixes it

With this PR's head cherry-picked onto its base (both `verl/workers/rollout/replica.py` and `verl/workers/rollout/vllm_rollout/vllm_async_server.py` applied cleanly), the identical configuration on the identical node **completed successfully**:

| | before | after this PR |
| --- | --- | --- |
| outcome | hung until killed at the 3600 s wall, 4/4 attempts | completed, exit 0, first attempt |
| wall time | n/a (deadlocked) | 1831.65 s |
| steps | never finished | all 5, with metrics |
| time per step | n/a | 218.72 s |
| throughput | n/a | 1004.72 tokens/s/gpu |
| mean reward | n/a | 0.6748 |

Nothing else changed — same image, same base commit, same 3600 s limit, same batch size, rollout `n`, parallelism and step count. A run that reaches step 5 with a real reward curve is only reachable if `wake_up` and the weight-sync `collective_rpc` were admitted, which is precisely what the unpatched runs could never do.

#### One observation that may be worth a follow-up

The failure is strongly hardware-dependent in our matrix: with the same image and same commit it reproduces 4/4 on 8x B300 and 2/2 each on GB200 NVL and GB300 NVL, but passes on H100, H200 and B200. Nothing in the dependency set differs between the passing and failing runs, so the platform seems only to change the timing that opens the window — consistent with a race that this PR closes rather than a platform-specific bug. It does suggest the underlying starvation is latent on the faster/passing platforms too.

Separately, it may be worth considering whether `GlobalRequestLoadBalancer.acquire_server` should respect the per-replica concurrency ceiling: the dedicated control group fixes the deadlock, but admitting 2048 requests against a 1040-slot actor still relies on queueing behaviour rather than admission control.

Happy to re-run this configuration against any revision of the patch if that is useful for review.

---

_This issue was drafted with assistance from the `opus` AI model._

wuxibin89:
@rao-ashish How much request concurrency do you have for a single vllm replica? The default concurrency for [asyncio ray actor](https://docs.ray.io/en/latest/ray-core/actors/async_api.html#setting-concurrency-in-async-actors) is 1000, which should be far exceeds the max concurrency that a single vllm replica can handle(typically 256, 512).

rao-ashish:
@wuxibin89 My config runs a single vLLM replica (TP=4 across 4 GPUs) and a large batch size (256 prompts / batch * 8 rollouts / prompt = 2048 rollouts / batch). This means that the vLLM replica at peak receives 2048 concurrent `generate()` requests. The rollout engine is configured to generate 256 sequences at a time via `rollout.max_num_seqs=256` (but this does not cap the max concurrent tasks enqueued to the replica). However, the large number of concurrent `generate()` requests causes the Ray concurrent task slots to fill up, and causes a deadlock during weight sync.

Here's a more detailed walkthrough of the issue. The PPOTrainer asks the `agent_loop_manager` to generate all rollouts for each of the 256 prompts [here](https://github.com/verl-project/verl/blob/10db40d0da4d59150bb389960b77585f81a89b8d/verl/trainer/ppo/v1/trainer_base.py#L1453). The `AgentLoopManagerTQ` will have each of the `AgentLoopWorkerTQ`s it manages [generate rollouts for one chunk](https://github.com/verl-project/verl/blob/10db40d0da4d59150bb389960b77585f81a89b8d/verl/trainer/ppo/v1/agent_loop_tq.py#L255) of the requested prompts. Each of these `AgentLoopWorkerTQ`s [spawn one agent loop](https://github.com/verl-project/verl/blob/10db40d0da4d59150bb389960b77585f81a89b8d/verl/trainer/ppo/v1/agent_loop_tq.py#L125) for each rollout it needs to generate (so 2048 agent loops for this config), and each of these `AgentLoop`s will [call generate()](https://github.com/verl-project/verl/blob/10db40d0da4d59150bb389960b77585f81a89b8d/verl/experimental/agent_loop/single_turn_agent_loop.py#L64) on its `server_manager` (this is an object of class `FullyAsyncLLMServerClient` in our async-colocated config). `FullyAsyncLLMServerClient.generate()` [calls into `vLLMHttpServer.generate()` ](https://github.com/verl-project/verl/blob/10db40d0da4d59150bb389960b77585f81a89b8d/verl/workers/rollout/llm_server.py#L142) (by way of [this call](https://github.com/verl-project/verl/blob/10db40d0da4d59150bb389960b77585f81a89b8d/verl/workers/rollout/llm_server.py#L283)). Crucially, `FullyAsyncLLMServerClient` [has a loop](https://github.com/verl-project/verl/blob/10db40d0da4d59150bb389960b77585f81a89b8d/verl/workers/rollout/llm_server.py#L281) that will repeatedly retry aborted requests. After sampling a batch from the replay buffer when running async-colocated jobs, we [abort all pending requests and pause the rollout engine](https://github.com/verl-project/verl/blob/10db40d0da4d59150bb389960b77585f81a89b8d/verl/trainer/ppo/v1/trainer_colocate_async.py#L46), but because of the loop in `FullyAsyncLLMServerClient.generate`, we will retry all the aborted requests. This enqueues 2048 generate requests to `vLLMHttpServer`. Moreover, the `vLLMHttpServer` is not able to drain the requests because it is paused. Since its concurrent task slots are full, Ray cannot admit the required `wake_up` / `resume_generation` / weight-update `collective_rpc`, and we end up in a deadlock.

This PR makes `vLLMHttpServer` use separate concurrency groups for control messages and generate requests so that the control messages are never starved of async task slots.

wuxibin89:
Run 2048 requests in single rollout replica is definitely not a reasonable setting. From our experience, in production job, the request concurrency for a single replica never exceeds 512.

janbernloehr:
Following up on the concurrency concern, with a correction to our earlier report: our tested configuration uses 8 GPUs with rollout TP=4, giving two rollout replicas, and `max_num_seqs=256`. The batch contains 256 prompts × 8 responses = 2,048 rollouts total; this should not have been presented as a measured per-replica concurrency.

We independently reran the runtime patch from 51b006540accf50fd07ac5e12d162d70acb79995 on verl 10db40d, which already includes #7511. The previously hanging B300 workload completed all five steps in 1,846.7 seconds, on its first attempt, with the batch size, parallelism, step count and 3,600-second timeout unchanged. This corroborates our earlier successful run.

We understand that lower per-replica concurrency may be the recommended operating point. The remaining concern is how that limit is enforced: `max_num_seqs` limits engine scheduling, while outstanding Ray `generate()` calls can still block the control calls needed for progress.

What supported configuration bounds outstanding requests per replica without reducing the training batch? If this load is unsupported, could verl apply backpressure or reject it explicitly instead of hanging? Would you prefer a separate issue to track that behavior?

_Drafted with assistance from Codex (GPT-6)._

wuxibin89:
@rao-ashish @janbernloehr https://github.com/verl-project/verl/pull/7999 add `ray_actor_max_concurrency` for rollout server actor, increase it in high concurrency workload.

Issue #6500, opened 2026-05-27
initialize_system_prompt fails on alternating-role chat templates


gxlvera:
### Overview

# `initialize_system_prompt` fails on chat templates that require alternating roles

## System Info

```text
python: 3.11.15
transformers: 5.8.1
```

This is a tokenizer-only reproduction. It does not require GPUs, Ray, SGLang, vLLM, or an end-to-end training run.

## Information

The problem arises when using:

- [ ] The official example scripts
- [x] My own modified scripts

The script only imports `verl.utils.chat_template.initialize_system_prompt` and uses an official Hugging Face tokenizer.

## Tasks

The tasks I am working on are:

- [ ] An officially supported task in the `examples` folder
- [x] My own task or dataset

This issue is about the tokenizer/chat-template utility used by the agent gateway, not a specific training dataset.

## Description

`verl.utils.chat_template.initialize_system_prompt` currently probes a tokenizer with:

```python
token1 = tokenizer.apply_chat_template(
    [{"role": "user", "content": ""}],
    add_generation_prompt=False,
    tokenize=True,
)
token2 = tokenizer.apply_chat_template(
    [{"role": "user", "content": ""}] * 2,
    add_generation_prompt=False,
    tokenize=True,
)
system_prompt = token1[: -(len(token2) - len(token1))]
```

This assumes that rendering `[user, user]` is valid. That assumption is false for official chat templates that require alternating `user/assistant/user/assistant/...` roles after the optional system message.

For those templates, `initialize_system_prompt` raises during initialization even though the tokenizer itself is valid. For example, this error is raised when using mistralai/Mistral-7B-Instruct-v0.2

### Information

- [ ] The official example scripts
- [x] My own modified scripts

### Tasks

- [ ] An officially supported task in the `examples` folder (such as GLUE/SQuAD, ...)
- [x] My own task or dataset (give details below)

### Reproduction

Save the following as `scripts/reproduce_system_prompt_alternating_roles_issue.py` and run it from the repository root:

```bash
PYTHONPATH=. python scripts/reproduce_system_prompt_alternating_roles_issue.py
```

```python
#!/usr/bin/env python3
from __future__ import annotations

import platform

import transformers
from transformers import AutoTokenizer

from verl.utils.chat_template import initialize_system_prompt


def main() -> None:
    print("Environment")
    print(f"python: {platform.python_version()}")
    print(f"transformers: {transformers.__version__}")

[...]

PR #7499, opened 2026-08-21
fix(mcore): don't crash fused forward when labels is None (#5273)

amanyagami:
## What

`patch_fused_forward()` monkey-patches `model.forward` to `_fused_GPTModel_forward` whenever `use_fused_kernels=True` (legacy pre-0.18 Megatron-Core mode). That function unconditionally fed `labels` into `linear_cross_entropy(...)`, but `labels` defaults to `None`, and `linear_cross_entropy.forward()` does `len(labels.shape)` unconditionally at `verl/utils/kernel/linear_cross_entropy.py:74` — so any caller invoking the patched forward without labels (e.g. `enable_mtp=False`) crashes with:

```
AttributeError: 'NoneType' object has no attribute 'shape'
```

Fixes #5273.

## Fix

When `labels is None`, `_fused_GPTModel_forward` now skips the fused cross-entropy kernel and instead projects `hidden_states` through the resolved `output_weight` to produce plain logits, mirroring the no-labels branch of the real (unpatched) `GPTModel._postprocess` (verified against `megatron-core==0.16.1` source). The labels-provided path is untouched.

`patch_fused_forward` itself is left unconditional — `verl/workers/engine/megatron/transformer_impl.py`'s `_maybe_enable_fused_kernels` already disables fused kernels whenever MTP is enabled, so the crash isn't really about MTP; it's that the monkey-patched `forward()` is reachable from other call sites without labels. Making the forward itself robust to `labels=None` is the more general fix.

## Testing

No GPU/`megatron-core` available in the environment this was developed in, so:
- `tests/models/test_model_forward_fused.py` (new) — written against the real module, will run in CI.
- A standalone harness that loads the actual, unmodified `model_forward_fused.py` via `importlib` with `megatron.core.*` stubbed out, driving `_fused_GPTModel_forward` directly with fakes: **13/13 checks pass** on the fixed code, and correctly **fails** when run against the pre-fix code (verified via `git stash`).
- `ruff check` / `ruff format --check` pass.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #7592, opened 2026-08-28
[misc] fix: initialize_system_prompt drops the prompt when diff==0 (#6477)

### LINKED ISSUE #6477: [Bug] `chat_template.py` system_prompt extraction returns empty list when both tokenized sequences have equal length (`[:-0]` bug)
## System Info

Discovered via static code analysis of `verl/utils/chat_template.py`. Affects multi-turn SFT and agent loop workflows.

## Information

## Tasks

## Reproduction

In `verl/utils/chat_template.py` lines 38 and 57:

```python
system_prompt = token1[: -(len(token2) - len(token1))]
```

When `len(token2) == len(token1)` (which happens with certain chat templates that produce equal-length encodings), this becomes:

```python
system_prompt = token1[:-0]  # Python: token1[:0] == []
```

`x[:-0]` in Python does **not** mean "keep all elements" — it is equivalent to `x[:0]`, which returns an empty list. The system prompt is silently lost.

## Expected behavior

```python
diff = len(token2) - len(token1)
system_prompt = token1[:-diff] if diff > 0 else []
# or, if diff==0 means no system prompt was injected:
system_prompt = token1[:-diff] if diff > 0 else token1
```

## Impact

- Silent bug in multi-turn SFT and agent-loop tokenization.
- System prompt tokens are silently dropped, causing the model to be trained/prompted without the intended system context.
- Affects `initialize_system_prompt` (line 38) and `extract_system_prompt_and_generation` (line 57).

amanyagami:
### What does this PR do?

Fixes #6477. `initialize_system_prompt()` and `extract_system_prompt_and_generation()` in `verl/utils/tokenizer/chat_template.py` derive the system-prompt token prefix from two probe renders:

```python
system_prompt = token1[: -(len(token2) - len(token1))]
```

When a chat template's rendering doesn't grow between one probe turn and two identical probe turns (`len(token2) == len(token1)`), this becomes `token1[:-0]`. In Python, `x[:-0]` is `x[:0]` — an empty list, not "keep everything" — so the system prompt is silently dropped to `[]` in that case.

`extract_system_prompt_and_generation()` is the one actually used in production, by `MultiTurnSFTDataset` (`verl/utils/dataset/multiturn_sft_dataset.py`), to compute how many leading tokens to strip from every non-first turn's individually-tokenized `input_ids`. With `system_prompt` wrongly collapsed to `[]`, nothing gets stripped, so every turn after the first is corrupted with a duplicated system-prompt-and-first-turn prefix instead of just its own tokens.

**Fix:** derive the same slice as `token1[: len(token1) - diff]` instead of the negative-index form. This is correct for `diff == 0` (returns the full probe render — there's no turn-specific marginal signal to strip in that case) and is unchanged for the normal `diff > 0` case, with no special-casing needed. Both call sites had the identical bug and are fixed identically.

- [x] Search for similar PRs. Paste at least one query link here: [`is:pr chat_template system_prompt` search](https://github.com/verl-project/verl/pulls?q=is%3Apr+chat_template+system_prompt) — the only related open PRs are for #6500/#6501 (a different bug in the same file, the multi-turn turn-separator token drop) and don't touch this slice. A prior attempt at #6477 itself (#6487) was closed unmerged with no review comment; the bug is still present on current `main` at both call sites.

### Test

Added `tests/utils/test_system_prompt_extraction_on_cpu.py` (CPU-only, no GPU/network/model download — a self-contained Jinja2-templated fake tokenizer stands in for a real one, following the existing `tests/utils/test_turn_separator_on_cpu.py` pattern):

- A template that only renders the first message in the list (so rendering 1 vs. 2 probe turns is byte-identical, i.e. `diff == 0`) reproduces the bug for both `initialize_system_prompt` and `extract_system_prompt_and_generation`.
- A normal template (system preamble + growing per-turn content, `diff > 0`) is checked as a non-regression sanity case for both helpers.

Verified red → green manually:
```
$ git stash   # revert the fix, keep the new test
$ uv run --extra cpu pytest -v tests/utils/test_system_prompt_extraction_on_cpu.py
...
FAILED test_initialize_system_prompt_zero_diff_does_not_drop_prompt - assert [] != []
FAILED test_extract_system_prompt_and_generation_zero_diff_does_not_drop_prompt - assert [] != []
2 failed, 2 passed

$ git stash pop   # restore the fix
$ uv run --extra cpu pytest -v tests/utils/test_system_prompt_extraction_on_cpu.py
4 passed
```

Also ran the pre-existing, closely related `tests/utils/test_turn_separator_on_cpu.py` (9 tests) — all still pass, no regression. `tests/utils/dataset/test_multiturn_sft_dataset_on_cpu.py` and `tests/utils/dataset/test_rl_dataset_on_cpu.py` fail identically before and after this change in this sandbox (they require locally-cached HF model weights that aren't available offline here) — confirmed by stashing the fix and re-running, so that's a pre-existing environment gap, not a regression from this PR.

Ran `pre-commit run --files verl/utils/tokenizer/chat_template.py tests/utils/test_system_prompt_extraction_on_cpu.py` (ruff, ruff-format, mypy, license/docstring/naming/structure checks): all pass on the changed files.

**Execution-mode note** (pure-Python vs. distributed): this code runs once in the driver process during dataset construction — plain tokenizer/text processing over `apply_chat_template`, before any FSDP/Megatron sharding or Ray/vLLM/SGLang rollout dispatch — and its behavior doesn't depend on the training backend or rollout engine at all. So the CPU unit test exercises the real production code path directly, with no mocking gap to flag.

### API and Usage Example

No API change — same signature and return type (`list[int]`), only the value returned in the previously-buggy `diff == 0` edge case changes (from incorrectly `[]` to the correct full probe token list).

### Design & Code Changes

- `verl/utils/tokenizer/chat_template.py`: in both `initialize_system_prompt()` and `extract_system_prompt_and_generation()`, replace `token1[: -(len(token2) - len(token1))]` with the equivalent-but-`diff==0`-safe `token1[: len(token1) - diff]`.
- `tests/utils/test_system_prompt_extraction_on_cpu.py`: new regression test file.

- [x] Apply pre-commit checks (see Test section above; ran scoped to the changed files — `pre-commit run --all-files` was not run in full due to an unrelated local-environment false positive from `check-naming-conventions` matching the pip-installed `.venv/` dependency tree, which is `.gitignore`d and not present in this repo's actual CI, which installs via plain `pip install --no-deps -e .` with no `.venv` — the same hook run scoped to just the two changed files passes cleanly).
- [ ] Add / Update the documentation. (Not applicable — internal helper, no public docs reference this slicing detail.)
- [x] Add unit test(s) covering the change (see above).

### DIFF
--- a/tests/utils/test_system_prompt_extraction_on_cpu.py
+++ b/tests/utils/test_system_prompt_extraction_on_cpu.py
@@ -1,0 +1,124 @@
+# Copyright 2025 Bytedance Ltd. and/or its affiliates
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""CPU tests for ``initialize_system_prompt`` / ``extract_system_prompt_and_generation``.
+
+Regression coverage for verl issue #6477: both helpers derive the system-prompt prefix as
+``token1[: -(len(token2) - len(token1))]``, where ``token1``/``token2`` are the token ids for one
+and two probe turns. When the marginal per-turn token count (``diff``) is zero -- e.g. a template
+whose rendering does not grow with extra messages -- Python's ``x[:-0]`` is ``x[:0]`` (empty list),
+not "keep everything", so the whole system prompt was silently dropped. No GPU, network, or model
+download is required: a self-contained Jinja2 template stands in for a real tokenizer.
+"""
+
+import re
+
+from jinja2 import Template
+
+from verl.utils.tokenizer.chat_template import extract_system_prompt_and_generation, initialize_system_prompt
+
+_IM_START, _IM_END, _NL = 1, 2, 3
+_SPECIALS = {"<|im_start|>": _IM_START, "<|im_end|>": _IM_END, "\n": _NL}
+
+# Ordinary ChatML template: system prefix. length grows with every extra turn, so the marginal
+# per-turn probe (`diff`) is non-zero -- this is the common case and must keep working.
+_CHATML_WITH_SYSTEM = (
+    "<|im_start|>system\nYou are helpful<|im_end|>\n"
+    "{% for m in messages %}<|im_start|>{{m['role']}}\n{{m['content']}}<|im_end|>\n{% endfor %}"
+    "{% if add_generation_prompt %}<|im_start|>assistant\n{% endif %}"
+)
+
+# Degenerate template that only ever renders the *first* message, ignoring the rest of the list.
+# Rendering one probe turn and rendering two probe turns therefore produce byte-identical output,
+# so `diff == len(token2) - len(token1) == 0` -- exactly the case `x[:-0]` gets wrong.
+_CHATML_FIRST_MESSAGE_ONLY = (
+    "<|im_start|>system\nYou are helpful<|im_end|>\n"
+    "{% set m = messages[0] %}<|im_start|>{{m['role']}}\n{{m['content']}}<|im_end|>\n"
+    "{% if add_generation_prompt %}<|im_start|>assistant\n{% endif %}"
+)
+
+
+class ChatMLTokenizer:
+    """Deterministic, offline tokenizer mimicking a ChatML ``apply_chat_template``."""
+
+    def __init__(self, template: str):
+        self._template = Template(template)
+        self._vocab: dict[str, int] = {}
+
+    def encode(self, text: str, add_special_tokens: bool = False) -> list[int]:
+        ids: list[int] = []
+        for piece in re.split(r"(<\|im_start\|>|<\|im_end\|>|\n)", text):
+            if piece == "":
+                continue
+            if piece in _SPECIALS:
+                ids.append(_SPECIALS[piece])
+            else:
+                for word in piece.split(" "):
+                    if word:
+                        ids.append(self._vocab.setdefault(word, len(self._vocab) + 100))
+        return ids
+
+    def apply_chat_template(self, messages, add_generation_prompt=False, tokenize=True, tools=None, **kwargs):
+        text = self._template.render(messages=messages, add_generation_prompt=add_generation_prompt)
+        return self.encode(text) if tokenize else text
+
+
+def test_initialize_system_prompt_normal_template_unaffected():
+    """Sanity check: the common (``diff > 0``) case still returns the real system prompt."""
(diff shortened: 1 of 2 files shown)


wuxibin89:
initialize_system_prompt is not used anymore, we have switch to ContinuousToken in https://github.com/verl-project/verl/pull/6779, https://github.com/verl-project/verl/pull/6804

amanyagami:
Thanks @wuxibin89 — you are right about `initialize_system_prompt`, and I confirmed it: on current `main` it has zero callers outside `tests/utils/test_turn_separator_on_cpu.py`. Nothing in `verl/`, `scripts/`, `recipe/` or `examples/` touches it.

But that only covers half of this PR. The same patch fixes the byte-identical slice in `extract_system_prompt_and_generation` (`verl/utils/tokenizer/chat_template.py:129`), and **that one is live** — `MultiTurnSFTDataset` calls it at `verl/utils/dataset/multiturn_sft_dataset.py:180`, and the ContinuousToken migration did not replace it.

The concrete effect there: when a chat template renders one probe turn and two probe turns to the same token count, `diff == 0`, and `token1[:-0]` is `token1[:0]` — the empty list, not "keep everything". `self.system_prompt` then comes back empty, so the strip at `multiturn_sft_dataset.py:231-232`

```python
input_ids = input_ids[len(self.system_prompt):]
attention_mask = attention_mask[len(self.system_prompt):]
```

becomes `[0:]` and removes nothing. Every turn after the first keeps its own copy of the system preamble, so the concatenated sequence diverges from `apply_chat_template` of the equivalent full conversation. It fails silently — no exception, just wrong training data.

Verified locally on this branch, rebased onto `main` at `8e4a5729`:

```
tests/utils/test_system_prompt_extraction_on_cpu.py
tests/utils/test_turn_separator_on_cpu.py
tests/utils/test_continuous_token_on_cpu.py
131 passed
```

and the two regression tests are not vacuous — restoring `token1[: -(len(token2) - len(token1))]` fails both, including `test_extract_system_prompt_and_generation_zero_diff_does_not_drop_prompt`, which is the live path. `ruff check` and `ruff format --check` are clean at the pinned v0.12.2.

I have rebased onto current `main` and pushed. **Happy to narrow this PR to just the `extract_system_prompt_and_generation` fix and drop the `initialize_system_prompt` half** — or, if you would rather, delete `initialize_system_prompt` and `initialize_turn_separator` outright along with their tests, since they are dead. Tell me which you prefer and I will push it.

On the CI: the 35 red workflows on the previous head were not test failures. Every check suite reports `latest_check_runs_count: 0` and all 35 flipped to `failure` at the same second (`2026-08-31T03:55:19Z`), so no job ever started — they look like fork runs that were never approved rather than anything in the diff. The sibling PRs #7503 and #7499 still sit at `action_required` for the same reason.

amanyagami:
Opened #7756 with just the live half, so this can stay closed.

It drops the `initialize_system_prompt` change you flagged as dead and keeps only the fix to `extract_system_prompt_and_generation`, which `MultiTurnSFTDataset` still calls. Rebased onto current `main` and re-verified the defect is still reachable there.

Issue #5125, opened 2026-01-30
Host memory unecessary usage in FSDP rank 0


flishwang:
### System Info

verl main ( 2026.01.27)
vllm 0.13.0 (updated from official docker image vllm012.exp3)
torch 2.9.1+cu129
GPU: 8* A800 80GB

### Information

- [ ] The official example scripts
- [x] My own modified scripts

### Tasks

- [ ] An officially supported task in the `examples` folder (such as GLUE/SQuAD, ...)
- [x] My own task or dataset (give details below)

### Reproduction

When I tried to run DAPO scripts with fsdp2+vllm async for Qwen3-Next-80B-A3B-Instruct on 4 nodes, I encountered host OOM during actor loading model.
When I use fsdp_size=16, I noticed that the memory usage for Node 0 and Node 2 is about 320GB higher than Node 1 and Node 3 after initializing the ref model, and blooms when initializing the actor/rollout.

It seems that  the weights in "full_state" are not freed after the following code, or there's some bug in set_state_dict from pytorch.

Does anyone meet the same issue?
Or should I use megatron instead of fsdp?

 
```
def _build_model_optimizer( ...):
    init_context = get_init_weight_context_manager(
            use_meta_tensor=not actor_model_config.tie_word_embeddings, mesh=self.device_mesh
        )
    ...
    with init_context(), warnings.catch_warnings():
        ...
        actor_module = actor_module_class.from_pretrained(
                pretrained_model_name_or_path=local_path,
                torch_dtype=torch_dtype,
                config=actor_model_config,
                trust_remote_code=trust_remote_code,
                attn_implementation=attn_implementation,
            )
    ....
    # Params loaded to CPU on rank0 here
    full_state = actor_module.state_dict()
    # Params converted to DTensor
    apply_fsdp2(actor_module, fsdp_kwargs, fsdp_config)
    # broadcast param weights from rank0 to others
    fsdp2_load_full_state_dict(actor_module, full_state, fsdp_mesh, cpu_offload)
    actor_module_fsdp = actor_module

```

### Expected behavior

Memory usage for all nodes being similar.

jianjunzhong:
Maybe you can try TransferQueue: 
- https://github.com/verl-project/verl/tree/main/verl/experimental/transfer_queue
- https://github.com/Ascend/TransferQueue

jianjunzhong:
We are also integrating TransferQueue into main: https://github.com/verl-project/verl/pull/4987

[...]

PR #7596, opened 2026-08-28
[fsdp] fix: param_offload frees nothing with fsdp_size=1 (NO_SHARD)

codearky:
### What does this PR do?

Fixes `param_offload` silently freeing no device memory when the actor is built with
`fsdp_size=1` (FSDP1). `fsdp_size=1` produces a 2-D `(ddp, fsdp)` mesh whose 1-rank shard
group degrades to `NO_SHARD`; combined with `use_orig_params=False` (the default),
`FlatParamHandle.flat_param_to()` moves the flat-parameter storage but does not refresh the
per-parameter views — those are only refreshed when `use_orig_params=True`. The stale views
keep the pre-move device allocation alive, so `offload_fsdp_model_to_cpu` frees nothing and
the following `load_fsdp_model_to_gpu` allocates a *second* copy.

The failure is silent: the assert added for the sharded path passes, and
`After offload model/optimizer/grad during init` is not printed at the default
`VERL_LOGGING_LEVEL=WARN`. The symptom only surfaces downstream — with a colocated vLLM
rollout the engine reads the still-occupied memory during KV-cache sizing and refuses to
start:

```
ValueError: Free memory on device cuda:0 (31.33/47.4 GiB) on startup is less than
desired GPU memory utilization (0.85, 40.29 GiB).
```

`FULL_SHARD` (`fsdp_size>1`) is unaffected and its code path is byte-for-byte unchanged.

### Checklist Before Starting

- [x] Search for similar PRs:
  [`is:open param_offload OR fsdp_size OR NO_SHARD`](https://github.com/verl-project/verl/issues?q=is%3Aopen+param_offload+OR+fsdp_size+OR+NO_SHARD)

**Not a duplicate.** Searched open PRs for `param_offload`, `fsdp_size`, `NO_SHARD`,
`use_orig_params`, `offload_fsdp_model_to_cpu` and `flat_param`. The nearest hits are
#7182 (make the *ref model* `CPUOffload` configurable — FSDP2, a config choice, not a
leak), #4696/#4726 (param offloading in `get_actor_weights_info`, under `recipe/`) and
#5125 (host memory retained after `state_dict()` on rank 0). None of them touch
`FlatParamHandle` view refresh after `flat_param_to` — verified with
`gh pr diff <n> | grep -E "_use_unsharded_views|flat_param_to|uses_sharded_strategy"`,
which matches 0 lines in each.

**Relationship to #7493 / #7494.** #7493 reports a *separate* bug in the same
configuration: on FSDP1 with `world_size > 1`, `fsdp_size=1` produces a degenerate
`(N, 1)` mesh whose shard group has one rank, so the gradient all-reduce becomes a no-op.
Its fix (#7494) was closed unmerged and `get_sharding_strategy` is unchanged on `main`, so
that issue is still open in practice. This PR does **not** fix or depend on it — the view
leak is in `flat_param_to` bookkeeping and is orthogonal to the process-group selection.

Note also that the leak is **not limited to `fsdp_size=1`**. Any FSDP1 run whose strategy
resolves to an unsharded one hits it, including a plain single-GPU run
(`world_size == 1`), where FSDP clamps to `NO_SHARD` regardless of `fsdp_size` and
`param_offload` is a legitimate, commonly-used setting.

### AI assistance

This change was developed with AI assistance (Claude). I reviewed every changed line,
ran the tests below myself, and can defend the change end-to-end. The AI assistance is
also recorded as a `Co-authored-by:` trailer on the commit.

### Test

Adds `tests/special_distributed/test_fsdp1_no_shard_offload.py` (2 GPUs), mirroring the
existing `test_fsdp2_pinned_model_transfer.py`. It builds an FSDP1 model through verl's own
`create_device_mesh(world_size, fsdp_size=1)` / `get_sharding_strategy(...)`, asserts the
handles really are `NO_SHARD`, then checks that the offload frees the device and that the
reload restores — rather than doubles — the footprint.

**The test offloads before the first forward, and that matters.** After a forward the views
point at the bf16 `_mp_shard` which FSDP frees post-forward, so nothing fp32 is pinned and
the bug does not reproduce. Pre-forward is also what `FSDPEngine` actually does — it offloads
right after model construction, and that is the allocation vLLM's KV-cache sizing reads.

Measured with a 735M-param model, `use_orig_params=False`, 2 GPUs:

| `fsdp_size` | strategy | state | variant | allocated before → after |
|---|---|---|---|---|
| 1 | NO_SHARD | init (pre-forward) | before this PR | 2.7387 GB → **2.7387 GB** (0% freed) |
| 1 | NO_SHARD | init (pre-forward) | with this PR | 2.7387 GB → **0.0000 GB** |
| 1 | NO_SHARD | post-forward | either | 2.8076 → 0.0690 GB (identical) |
| 2 | FULL_SHARD | any | before this PR | already correct, unchanged |

The reload also compounds without the fix: the offload→load round trip left **5.4773 GB**,
roughly 2× the model, because the stale views keep the original allocation alive while the
reload allocates a fresh one. With the fix it returns to 2.7387 GB.

The test was verified to fail without the fix. On pristine upstream it reports:

```
AssertionError: offload_fsdp_model_to_cpu freed almost nothing:
               184636416 -> 184636416 bytes allocated (100.0% still resident on device)
```

byte-for-byte identical before and after, on both ranks. With the fix applied it passes. A
variant with `fsdp_size=2` (NO_SHARD assertion inverted) confirms the `uses_sharded_strategy`
early-return leaves the sharded path untouched.

End-to-end, Qwen3-4B / GRPO / 2 GPUs / `fsdp_size=1` / `param_offload=True`
(`VERL_LOGGING_LEVEL=DEBUG` to surface the line):

```
                                          before        after
After FSDP,                             15.02 GB      15.02 GB
After offload ... during init            15.02 GB       0.00 GB
```

which lets the colocated vLLM engine start at `gpu_memory_utilization=0.85`. Post-round-trip
forward is bit-identical (`max|logit diff| = 0`) and `len(list(model.parameters()))` is
unchanged.

**Commands run:**

```bash
# regression test, 2 GPUs -- fails on main, passes with this PR
torchrun --nproc-per-node=2 --standalone \
    tests/special_distributed/test_fsdp1_no_shard_offload.py

# lint / sanity
ruff check verl/utils/fsdp_utils.py tests/special_distributed/test_fsdp1_no_shard_offload.py
ruff format --check verl/utils/fsdp_utils.py tests/special_distributed/test_fsdp1_no_shard_offload.py
python3 tests/special_sanity/check_license.py --directories .
python3 tests/special_sanity/check_device_api_usage.py --directory ./verl
python3 tests/special_sanity/check_docstrings.py
python3 tests/special_sanity/validate_structure.py
```

All lint/sanity checks pass. The regression test is wired into
`tests/special_distributed/run_all.sh`, which `model.yml` already runs and already lists
in its `paths`.

### API and Usage Example

No API change. Existing configs are unaffected; `fsdp_size=1` with `param_offload=True` now
behaves as documented.

```bash
actor_rollout_ref.actor.fsdp_config.fsdp_size=1 \
actor_rollout_ref.actor.fsdp_config.param_offload=True \
actor_rollout_ref.rollout.gpu_memory_utilization=0.85
```

### Design & Code Changes

- `verl/utils/fsdp_utils.py`: add `_refresh_unsharded_views_after_move(handle)` and call it
  after the move in both `offload_fsdp_model_to_cpu` and `load_fsdp_model_to_gpu`.
  - Guarded by `handle._use_orig_params` (already handled by `flat_param_to`) and
    `handle.uses_sharded_strategy` (sharded strategies rebuild views on the next unshard), so
    only the unsharded + `use_orig_params=False` case changes behaviour.
  - `as_params=False` matches FSDP's at-rest registration for `use_orig_params=False`, so the
    flat parameter remains the only registered `nn.Parameter` and `model.parameters()` /
    optimizer param groups are unchanged.
- `tests/special_distributed/test_fsdp1_no_shard_offload.py`: new 2-GPU regression test.

- [x] Read the Contribute Guide
- [ ] Apply pre-commit checks
- [x] Add / update documentation — n/a (no user-facing API change)
- [x] Add unit or end-to-end test(s)
- [ ] Send a message in the `ci-request` channel

### DIFF
--- a/tests/special_distributed/run_all.sh
+++ b/tests/special_distributed/run_all.sh
@@ -25,3 +25,6 @@ torchrun --nproc-per-node=2 --standalone tests/special_distributed/test_fsdp2_cp
 # Regression for colocated FSDP2 full-parameter model transfers. It locks in
 # PyTorch's implicit pinned allocation for non-blocking D2H Module.to().
 torchrun --nproc-per-node=2 --standalone tests/special_distributed/test_fsdp2_pinned_model_transfer.py
+# Regression for FSDP1 param_offload under an unsharded strategy (fsdp_size=1).
+# Only needs 2 ranks to build the one-rank shard group that degrades to NO_SHARD.
+torchrun --nproc-per-node=2 --standalone tests/special_distributed/test_fsdp1_no_shard_offload.py
--- a/tests/special_distributed/test_fsdp1_no_shard_offload.py
+++ b/tests/special_distributed/test_fsdp1_no_shard_offload.py
@@ -1,0 +1,133 @@
+# Copyright 2026 Bytedance Ltd. and/or its affiliates
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Regression test for FSDP1 parameter offload under an unsharded strategy.
+
+``fsdp_size=1`` builds a one-rank shard group, which degrades FSDP1 to
+``NO_SHARD``. Combined with verl's default ``use_orig_params=False``, the
+per-parameter views registered on the wrapped modules keep pointing at the
+pre-move flat-parameter storage: ``FlatParamHandle.flat_param_to()`` only
+refreshes those views when ``use_orig_params=True``. The stale views hold a
+reference to the old device allocation, so ``offload_fsdp_model_to_cpu`` frees
+nothing and the following ``load_fsdp_model_to_gpu`` allocates a second copy.
+
+The offload must be exercised **before any forward pass**: FSDP's pre-forward
+unshard re-points the views itself, which hides the leak.
+
+Launch:
+    torchrun --nproc-per-node=2 --standalone \
+        tests/special_distributed/test_fsdp1_no_shard_offload.py
+"""
+
+import torch
+import torch.distributed
+from torch.distributed.fsdp import FullyShardedDataParallel as FSDP
+from torch.distributed.fsdp import MixedPrecision
+from transformers import AutoModelForCausalLM, Qwen2Config
+
+from verl.utils.device import get_device_name, get_torch_device
+from verl.utils.distributed import initialize_global_process_group
+from verl.utils.fsdp_utils import load_fsdp_model_to_gpu, offload_fsdp_model_to_cpu
+from verl.workers.engine.fsdp.utils import create_device_mesh, get_sharding_strategy
+
+
+def _build_no_shard_fsdp1_model(world_size):
+    config = Qwen2Config(
+        num_hidden_layers=4,
+        hidden_size=1024,
+        intermediate_size=2048,
+        num_attention_heads=8,
+        num_key_value_heads=8,
+        vocab_size=2048,
+    )
+    model = AutoModelForCausalLM.from_config(config=config, torch_dtype=torch.float32)
+    model = model.to(get_device_name())
+    device_mesh = create_device_mesh(world_size=world_size, fsdp_size=1)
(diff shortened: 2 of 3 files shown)


wuxibin89:
FSDP1 with fsdp_size=1 is quite buggy, we're not plan to support it. I would suggest with fsdp_size>1 or at least with FSDP2+fsdp_size=1

PR #7607, opened 2026-08-29
[reward] fix: remove eval() RCE in prime_math grader (CVE-2026-6878)

### LINKED ISSUE #7603: CVE-2026-6878 is still live on main: the prime_math grader executes eval() on model-authored answer text, and one poisoned sample forges every later reward in the worker
PUBLIC-READY (the eval() sinks are already publicly advisored; the reward-forgery chain and the "still unfixed on main" status are the new content)
R03E1-M1 genus (scorer-side code execution from an evaluated-side string); ETB-01/ETB-02 family (the judge consumes an artifact the evaluated system authored)
Component: `verl/utils/reward_score/prime_math/grader.py` (`handle_pi`, `math_equal` matrix branch), reached from `verl/utils/reward_score/__init__.py` (`default_compute_score`) for every `numina_*` data source
Prior art: CVE-2026-6878 / GHSA-h57c-v2v3-5v3v (published 2026-04-23, "Patched versions: None"), write-up at https://github.com/zast-ai/vulnerability-reports/blob/main/bytedance/verl_rce.md

## Summary

`default_compute_score` routes the data sources `numina_aops_forum`, `numina_synthetic_math`, `numina_amc_aime`, `numina_synthetic_amc`, `numina_cn_k12` and `numina_olympiads` to `prime_math.compute_score`, which extracts the model's final answer from the rollout text and hands it to `math_equal`. Two sinks in that file run Python `eval()` on the extracted answer string inside the reward process:

1. `handle_pi`: whenever the answer contains `\pi`, every `\pi` is replaced with `*math.pi` or `1*math.pi` and the entire remaining answer is passed to `eval()`.
2. The matrix branch of `math_equal`: when the ground truth contains `\begin{pmatrix}` and the prediction starts with `[` and ends with `]`, `eval(prediction)` runs twice.

This is the already-published CVE-2026-6878. The reason this issue is being filed anyway: the advisory lists no patched version, and both sinks are still present, unchanged, on current `main` (verified at commit 8df88be746801b6d87c42e15f9a5a0ec1d5eeeae, 2026-08-28, four months after publication). Nothing in the default reward path warns, sandboxes, or gates the call.

What the advisory does not spell out, and what the attached PoC adds: the executed code runs with the reward worker's own module state, so a single poisoned sample can do more than run a command. The PoC's first sample executes code that monkeypatches `prime_math.grade_answer` to always return True. From that moment, every subsequent sample graded in the same reward process scores 1.0, including a plain wrong answer that contains no payload at all. One sacrificed rollout converts the rest of the batch to full reward, which is precisely the gradient signal an RL policy is optimized to find, and precisely the artifact a poisoned dataset or prompt-injected problem statement can plant.

## Impact

1. Arbitrary code execution in the reward worker by the policy's own output text (known from the advisory, re-verified on main).
2. Persistent reward forgery within the worker process: after the monkeypatch, `reward_model` scores stop being a function of correctness. Training curves, checkpoint selection, and any published eval numbers produced by that worker can no longer distinguish "model answered correctly" from "model executed code in the grader".
3. In the default `naive` reward manager the scoring loop runs in the training driver process; in the `remote` manager it runs in `RewardComputeWorker` actors that persist across batches. Both give the payload a long-lived process to corrupt.

## Reproduction

PoC: `poc_f1_primemath_eval_rce/poc.py`. It imports the real `verl.utils.reward_score.default_compute_score` and grades three samples in one process, the way a reward worker would see them:

- Sample 1 (matrix sink): ground truth `\begin{pmatrix}1&2\\3&4\end{pmatrix}`, answer `\boxed{[exec("import os\nos.system('echo pwned > /tmp/r04e1f1/mx.txt')"),2]}`. The matrix branch evals the bracketed prediction and the marker file appears.
- Sample 2 (pi sink, data_source `numina_cn_k12`): answer `\boxed{=\pi+exec("import sys\nimport os\ndef g(a,b):return True\nsetattr(sys.modules[...prime_math...],'grade'+chr(95)+'answer',g)\nos.system('echo pwned > /tmp/r04e1f1/pi.txt')")}`. `handle_pi` evals the answer; the marker appears and `grade_answer` is patched.
- Sample 3 (no payload): `The answer is \boxed{42}` against ground truth `7` scores 1.0.

Control mode (fresh process, benign payloads) grades the wrong answers 0.0, writes no markers, and leaves `grade_answer` unpatched.

Run:

```
PYTHONPATH=/tmp/r04e1deps python3 poc_f1_primemath_eval_rce/poc.py attack
PYTHONPATH=/tmp/r04e1deps python3 poc_f1_primemath_eval_rce/poc.py control
```

Double-run determinism: attack and control reports are byte-identical across repeated runs (sha256 recorded in `determinism/`).

Payload-shape constraints worth knowing when writing a regression test: `handle_base` raises on any `_` in the answer before the eval is reached, and `match_answer` truncates the extracted answer at the last `=`, `: `, `is `, `be `, `are `, `get `, `. ` (and newline variants). That is why the PoC payload leads with one sacrificial `=` and builds `grade_answer`/module names via `chr(95)`. A fix that only blocks the literal strings from the earlier write-up will not block these.

## Suggested fix

- Replace both `eval()` calls in `prime_math/grader.py` with `ast.literal_eval` inside try/except, as the advisory's write-up already recommends. `handle_pi` only needs arithmetic on floats; a tiny explicit parser for `a\pi + b` patterns is safer still.
- Treat "comparison could not be performed" as not-correct (fail closed), never as an exception to suppress and continue.
- Longer term, run rule-based reward functions in a subprocess pool that does not share module state with the worker, so a single sample cannot monkeypatch grading for later samples. The `remote` manager's `RewardComputeWorker` is the right shape; `naive` currently evaluates in-process.

## Reference

- `verl/utils/reward_score/prime_math/grader.py`, `handle_pi` (`string = eval(string)` under `contextlib.suppress(Exception)`) and `math_equal` matrix branch (`isinstance(eval(prediction), list)` / `pred_matrix = eval(prediction)`), at commit 8df88be746801b6d87c42e15f9a5a0ec1d5eeeae.
- `verl/utils/reward_score/__init__.py`, `default_compute_score`, numina routing.
- Advisory: GHSA-h57c-v2v3-5v3v / CVE-2026-6878 (affected <= 0.7.1, no patched version listed).

amanyagami:
### What does this PR do?

Fixes #7603. Removes the `eval()` remote-code-execution sinks in the `prime_math` reward grader (CVE-2026-6878 / GHSA-h57c-v2v3-5v3v), which is still present on `main`.

`verl/utils/reward_score/prime_math/grader.py` passed model-authored answer text to Python `eval()` in two places, both reached from `default_compute_score` for every `numina_*` data source:

- `handle_pi` — `eval()` of the answer string after `\pi` substitution
- `math_equal` matrix branch — `eval(prediction)` (one call was **unguarded**)

Because rule-based rewards run in the reward worker's own process, this is arbitrary code execution driven by the policy's own output. Worse, a single poisoned sample can monkeypatch `grade_answer` so that **every later sample in the same worker scores full reward** — persistent reward forgery that silently corrupts training curves, checkpoint selection, and eval numbers.

I reproduced all of this on current `main`: both sinks execute injected code, and a wrong answer flips from `False` to `True` after one payload runs.

### Test

New CPU test: `tests/utils/reward_score/test_prime_math_grader_security_on_cpu.py` (matches the `*_on_cpu.py` glob, runs in `cpu_unit_tests.yml`).

- Verified it **fails on the pre-fix code** (payloads execute, marker files are written) and **passes after** the fix (16 passed).
- Confirmed no behavioral regression: the repo's own three `prime_math` grading cases (matrix / fraction / symbolic) still grade `True`, and a 23-case legitimate-behavior baseline (arithmetic, `\pi` expressions, list-matrix, fractions, boxed answers) is byte-for-byte identical before and after.
- `pre-commit run ruff` and `ruff-format` pass on both files.

### API and Usage Example

No public API change. Internal `eval()` is replaced by a new module-level helper `safe_arithmetic_eval` in the same file.

```python
from verl.utils.reward_score.prime_math.grader import safe_arithmetic_eval

safe_arithmetic_eval("1*3.14 + 2")   # 5.14
safe_arithmetic_eval("[1/2, 3]")     # [0.5, 3]
safe_arithmetic_eval('exec("...")')  # raises ValueError — never executes
```

### Design & Code Changes

- Add `safe_arithmetic_eval(expr)` — an allowlist AST evaluator permitting only numeric constants, unary `+`/`-`, binary `+ - * / // % **`, and list/tuple literals of those. Any `Call`, `Name`, `Attribute`, subscript, comprehension, etc. raises `ValueError`. An **allowlist** (not the existing `should_allow_eval` denylist) cannot be bypassed by obfuscation such as `chr(95)`, which the issue's PoC uses.
- `handle_pi`: `eval` → `safe_arithmetic_eval`, still under `contextlib.suppress`, so a non-numeric answer falls through to a not-correct comparison (fail-closed), preserving current output for legit inputs.
- `math_equal` matrix branch: parse the bracketed prediction once via `safe_arithmetic_eval` inside a `try`, removing the previously **unguarded** `eval`.

Deeper isolation (running rule-based reward functions in a subprocess that does not share module state, so a payload cannot patch grading for later samples) is a larger architectural change and is left as follow-up; this PR closes the code-execution and reward-forgery vectors themselves.

### Checklist Before Starting

- [x] Search for similar PRs: https://github.com/verl-project/verl/pulls?q=is%3Apr+prime_math+eval — none open for this.
- [x] PR title format `[reward] fix: ...` (passes `check_pr_title.py`).

- [x] Apply pre-commit checks (ruff / ruff-format pass on the changed files).
- [ ] Update documentation — n/a (internal security fix, no API change).
- [x] Add unit test(s) to CI (`test_prime_math_grader_security_on_cpu.py`).

---

AI assistance was used to prepare this change (per [`AGENTS.md`](https://github.com/verl-project/verl/blob/main/AGENTS.md)). Every changed line was reviewed by me before submitting, and the tests listed above were run locally with the results shown.

### DIFF
--- a/tests/utils/reward_score/test_prime_math_grader_security_on_cpu.py
+++ b/tests/utils/reward_score/test_prime_math_grader_security_on_cpu.py
@@ -26,6 +26,7 @@ pin down that:
 """
 
 import math
+import time
 
 import pytest
 
@@ -134,3 +135,75 @@ class TestLegitimateGradingUnchanged:
     def test_pi_and_fraction_equality(self):
         assert math_equal(r"3\pi", "9.42477796") is True
         assert math_equal("0.5", "0.5") is True
+
+
+class TestResourceBounds:
+    """The evaluator must not turn RCE into a CPU/memory denial of service.
+
+    ``safe_arithmetic_eval`` runs on model-authored text in a long-lived reward worker,
+    and its callers only guard with ``except Exception`` -- which a hang or an OOM does
+    not raise. So an expression that is merely ruinously expensive has to be *rejected*,
+    not merely survived.
+    """
+
+    @pytest.mark.parametrize(
+        "expr",
+        [
+            "9**9**9**9",  # right-associated tower: the classic int-blowup payload
+            "9**9**9",
+            "2**999999999",
+            "10**100000",
+            "(10**1000)**1000",
+        ],
+    )
+    def test_rejects_exponent_blowup(self, expr):
+        with pytest.raises(ValueError):
+            safe_arithmetic_eval(expr)
+
+    def test_exponent_blowup_is_rejected_promptly(self):
+        """Rejection must come from the size check, not from actually computing it."""
+        start = time.monotonic()
+        with pytest.raises(ValueError):
+            safe_arithmetic_eval("9**9**9**9")
+        assert time.monotonic() - start < 1.0
+
+    def test_rejects_oversized_intermediate_from_multiplication(self):
+        """Bounding only ``**`` would leave repeated multiplication as a way through."""
+        with pytest.raises(ValueError):
+            safe_arithmetic_eval("(2**4000) * (2**4000)")
+
+    def test_rejects_overlong_expression(self):
+        with pytest.raises(ValueError):
+            safe_arithmetic_eval("1+" * 20_000 + "1")
+
+    def test_rejects_too_many_nodes(self):
+        with pytest.raises(ValueError):
+            safe_arithmetic_eval("1+" * 9_000 + "1")
+
+    def test_rejects_deep_nesting(self):
+        with pytest.raises(ValueError):
+            safe_arithmetic_eval("-" * 100 + "1")
+
+    def test_bounds_do_not_reject_realistic_answers(self):
+        """The limits must sit far above anything a real answer contains."""
+        assert safe_arithmetic_eval("2**10") == 1024
+        assert safe_arithmetic_eval("2**0.5") == pytest.approx(math.sqrt(2))
+        assert safe_arithmetic_eval("-2**3") == -8
+        assert safe_arithmetic_eval("3.14159*2") == pytest.approx(3.14159 * 2)
+        assert safe_arithmetic_eval("[1, 2, 3, 4, 5, 6, 7, 8, 9]") == [1, 2, 3, 4, 5, 6, 7, 8, 9]
+
+    def test_grader_sinks_survive_a_blowup_payload(self):
+        """Both eval() sinks must reject the payload rather than hang.
+
+        ``handle_pi`` substitutes the numeric pi before evaluating and suppresses any
+        failure, so a rejected expression leaves the substituted *string* in place. That
+        is the intended fallthrough: a non-numeric result simply grades as not correct.
+        """
+        start = time.monotonic()
+
+        result = handle_pi(r"9**9**9**9\pi", math.pi)
+        assert isinstance(result, str)
+        assert result == "9**9**9**9*" + repr(math.pi)
+
+        assert math_equal("[9**9**9**9]", r"\begin{pmatrix}1\end{pmatrix}") is False
+        assert time.monotonic() - start < 2.0
--- a/tests/utils/reward_score/test_prime_math_grader_security_on_cpu.py
+++ b/tests/utils/reward_score/test_prime_math_grader_security_on_cpu.py
@@ -1,0 +1,136 @@
+# Copyright 2025 Bytedance Ltd. and/or its affiliates
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Regression tests for CVE-2026-6878 / GHSA-h57c-v2v3-5v3v.
+
+The prime_math grader used ``eval()`` on model-authored answer text in two places
+(``handle_pi`` and the ``math_equal`` matrix branch). Because the reward function runs
+in the reward worker's own process, a single poisoned sample could execute arbitrary
+code and even monkeypatch grading so every later sample scored full reward. These tests
+pin down that:
+
+* the safe evaluator accepts arithmetic / list literals but rejects anything executable,
+* neither sink executes an injected payload,
+* a poisoned sample can no longer forge rewards for later samples, and
+* legitimate arithmetic and list-matrix grading are unchanged.
+"""
+
+import math
+
+import pytest
+
+from verl.utils.reward_score.prime_math.grader import (
+    handle_pi,
+    math_equal,
+    safe_arithmetic_eval,
+)
+
+
+class TestSafeArithmeticEval:
+    def test_accepts_numbers_and_arithmetic(self):
+        assert safe_arithmetic_eval("1*3.14+2") == pytest.approx(1 * 3.14 + 2)
+        assert safe_arithmetic_eval("2**3") == 8
+        assert safe_arithmetic_eval("-5") == -5
+        assert safe_arithmetic_eval("(1 + 2) * 3") == 9
+        assert safe_arithmetic_eval("7 / 2") == pytest.approx(3.5)
+
+    def test_accepts_list_and_tuple_literals(self):
+        assert safe_arithmetic_eval("[1, 2, 3]") == [1, 2, 3]
+        assert safe_arithmetic_eval("[1/2, 3]") == [0.5, 3]
+        assert safe_arithmetic_eval("(1, 2)") == (1, 2)
+
+    @pytest.mark.parametrize(
+        "expr",
+        [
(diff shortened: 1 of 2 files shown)


koriyoshi2041 (commented):
The allowlist closes code execution, but keeping unrestricted `ast.Pow` leaves a model-authored resource-exhaustion path in the reward worker. Inputs such as a right-associated exponent tower can create enormous integers before grading returns; the evaluator also has no AST node/depth budget. Since this runs on policy output in a persistent worker, could this either drop `Pow` (if the legitimate baseline does not require it) or enforce explicit exponent/result-size plus node/depth limits? A negative regression for an exponent tower would keep the security boundary from moving from RCE to CPU/memory DoS.

amanyagami:
@koriyoshi2041 Good catch — you are right that the boundary would have moved from RCE to CPU/memory DoS rather than closing. Fixed in `b7d469f6`.

I kept `Pow` rather than dropping it, because the legitimate baseline does use it: the grader evaluates things like `2**10` and `2**0.5`, and the existing tests in this PR already assert `2**3 == 8`. So I took your second option and bounded it instead.

**What is bounded now**

```python
_MAX_EXPR_CHARS   = 10_000
_MAX_AST_NODES    = 10_000
_MAX_AST_DEPTH    = 50
_MAX_POW_EXPONENT = 1024
_MAX_INT_BITS     = 4096
```

The exponent check happens *before* the power is computed — checking afterwards is too late, since the cost is paid inside the call:

```python
if isinstance(base, int) and isinstance(exponent, int) and exponent > 0:
    if base.bit_length() * exponent > _MAX_INT_BITS:
        raise ValueError("exponentiation result is too large")
```

A tower degrades safely for the reason you would expect: `9**9**9**9` evaluates right to left, so the innermost `9**9` is cheap (36 bits), and the next level then sees an exponent of `387420489`, which trips `_MAX_POW_EXPONENT` before anything large is allocated.

Bounding only `**` would have left repeated multiplication as a way through, so every intermediate integer is size-checked as well — `(2**4000) * (2**4000)` is rejected, with a test for it.

**Negative regression coverage**, as you asked:

- exponent towers and blowups: `9**9**9**9`, `9**9**9`, `2**999999999`, `10**100000`, `(10**1000)**1000`
- a timing assertion that the rejection comes from the size check, not from computing the value
- oversized multiplication, overlong input, node-count and nesting-depth limits
- both grader sinks (`handle_pi` and the `math_equal` matrix branch) fed the payload, asserting they return promptly
- a guard that the limits sit far above realistic answers (`2**10`, `2**0.5`, `-2**3`, a 9-element matrix)

I also checked the guard is not vacuous. The unguarded evaluator does not return within 20s on `9**9**9**9` (killed by `timeout`, exit 124); the bounded one rejects it in under a second.

```
tests/utils/reward_score/test_sandbox_on_cpu.py::test_prime_math
tests/utils/reward_score/test_prime_math_grader_security_on_cpu.py
29 passed
```

That includes this repo's existing `test_prime_math`, which grades real answer/ground-truth pairs, so the limits do not change grading behaviour. `ruff check` and `ruff format --check` are clean.

koriyoshi2041 (approved):
Re-checked b7d469f6. The exponent/result-size guards run before costly integer powers, intermediate integers are bounded across other operators, and the node/depth/input limits plus both grader-sink regressions cover the resource-exhaustion path I flagged. This resolves my concern.

amanyagami:
Thanks for the re-review @koriyoshi2041 — appreciate you catching the DoS path before this landed.

This one is now approved and `mergeable`, but its 34 workflow runs are still at `action_required`, so CI has never started. Could a maintainer approve the runs and land it? It is the fix for CVE-2026-6878 / GHSA-h57c-v2v3-5v3v — `eval()` on model-authored answer text in the reward worker — so it is worth getting in ahead of the rest of my queue.

Summary of what is in it:
- `eval()` replaced with an AST allowlist (numeric literals, unary `+`/`-`, `+ - * / // % **`, list/tuple literals); calls, names and attribute access are rejected, so obfuscation cannot bypass it
- evaluation bounded on input length, node count, depth, exponent magnitude and intermediate integer size, so the boundary does not move from RCE to resource exhaustion
- CPU-only regression tests covering both `eval()` sinks, injection payloads, reward forging, exponent towers, and unchanged behaviour on legitimate arithmetic and matrix grading

amanyagami:
Checked again today — still approved and `mergeable`, and CI still hasn't started (all 34 runs `action_required`). This is the fix for CVE-2026-6878 (`eval()` RCE in the prime_math grader), so flagging again in case it's just been missed: could a maintainer approve the workflow runs so it can land?

PR #7610, opened 2026-08-30
[ckpt, fsdp] fix: FSDP model merger concatenates replicated buffers

Josephasafg:
### What does this PR do?

FSDPModelMerger treats every non-DTensor entry in a rank's checkpoint as a row-shard and torch.cat(dim=0)s it across ranks. FSDP2 never shards buffers, so each rank holds an identical copy, and merging a model with persistent buffers goes wrong in two ways:

  - 0-d buffers crash the merge: RuntimeError: zero-dimensional tensor (at position 0) cannot be concatenated
  The error this had yielded is: 
```bash
File "verl/model_merger/fsdp_model_merger.py", line 202, in _load_and_merge_state_dicts
      state_dict[key] = torch.cat(state_dict[key], dim=0)
  RuntimeError: zero-dimensional tensor (at position 0) cannot be concatenated
```
  - ≥1-d buffers are silently corrupted: a (1,) buffer becomes (8,) on 8 ranks, producing an HF checkpoint with wrong-shaped tensors.

This happened during conversion of `google/gemma-4-E4B-it` FSDP2 checkpoints

### DIFF
--- a/tests/model_merger/test_fsdp_merge_non_dtensor_shards_on_cpu.py
+++ b/tests/model_merger/test_fsdp_merge_non_dtensor_shards_on_cpu.py
@@ -1,0 +1,38 @@
+# Copyright 2025 Bytedance Ltd. and/or its affiliates
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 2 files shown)


Josephasafg:
How can I join the verl Slack channel? the link seems to be only for specific domains.

PR #7618, opened 2026-08-30
[rollout,perf] fix: bound Qwen generation prompt tokenization

### LINKED ISSUE #7617: # [rollout] perf: `_tokenize_generation_prompt_delta` re-tokenizes the full conversation after every tool turn
### Problem

The execution chain is:

```text
```

   once with it.

### Suggested Fix

with:

```python
    [],
    add_generation_prompt=True,
    tools=tools,
)
```

### Information

- [x] My own modified scripts

### Tasks

### Reproduction

|---|---:|---:|

### Expected behavior

JimmyWang0417:
### What does this PR do?

Fixes #7617 by keeping generation-prompt delta tokenization bounded for verified Qwen3 and Qwen3.5 text model families. The new builders render a small synthetic system/user conversation when deriving the generation scaffold, while the generic `ContinuousTokenBuilder` path remains unchanged for unknown or history-dependent templates.

Open PR #7296 changes Continuous Token context merging to support assistant messages and rollback-style harness updates. This PR does not change those APIs or context grouping; it only adds model-family-specific generation-prompt handling for the performance issue described in #7617. The generic full-history implementation remains the fallback for compatibility.

### Design & Code Changes

- Add Qwen3 and Qwen3.5 text builder subclasses with bounded generation-prompt renders.
- Keep Qwen2/Qwen2.5 and generic builders on the existing full-history path.
- Register the specialized builders for explicit and auto-resolved Qwen3/Qwen3.5 families.
- Add CPU tests covering bounded render size, suffix equivalence with the fallback, `tools` and chat-template kwargs propagation, and generic fallback behavior.

### Test

- `PYTHONPATH=/tmp/roll-442-site:/tmp/verl-7617-work /tmp/harbor-contrib/.venv/bin/python -m pytest -q tests/utils/test_continuous_token_on_cpu.py` (122 passed)
- `ruff check verl/utils/tokenizer/continuous_token.py verl/utils/tokenizer/continuous_token_wiring.py tests/utils/test_continuous_token_on_cpu.py`
- `ruff format --check verl/utils/tokenizer/continuous_token.py verl/utils/tokenizer/continuous_token_wiring.py tests/utils/test_continuous_token_on_cpu.py`
- `python -m compileall` on the changed implementation modules
- `git diff --check`

AI assistance was used to prepare this patch. I reviewed every changed line and ran the checks listed above.


verl/utils/tokenizer/continuous_token.py
```diff
@@ -468,6 +470,40 @@ def _merge_non_assistant_token_ids(
         )
 
 
+class _QwenGenerationPromptContinuousTokenBuilder(QwenContinuousTokenBuilder):
```
gxlvera: I think we can simplify the class hierarchy by overriding _tokenize_generation_prompt_delta() directly in QwenContinuousTokenBuilder, rather than introducing _QwenGenerationPromptContinuousTokenBuilder and separate Qwen3/Qwen3.5 subclasses.
I checked the official chat templates for all Qwen model families currently supported by the continuous-token builder, including Qwen2, Qwen2.5, Qwen3, Qwen3.5, and the corresponding Qwen-VL families. For a fixed set of chat-template arguments, their generation prompt does not depend on the accumulated message history or on whether the last message is a user message or a tool response. So we can simply override the hook in QwenContinuousTokenBuilder

PR #5724, opened 2026-03-24
[docker, ci] fix: all CIs, transformers upgrade to 5.3.0 and vllm==0.18.0

ETOgaosion:
### What does this PR do?

[ci] fix: e2e_fully_async_policy_megatron rope_theta parameter error due to transformers upgrade
Fix dependency loop in megatron introduced by https://github.com/verl-project/verl/pull/5717
Try fix trtllm rollout weights transfer error

### Checklist Before Starting

- [ ] Search for similar PRs. Paste at least one query link here: ...
- [ ] Format the PR title as `[{modules}] {type}: {description}` (This will be checked by the CI)
  - `{modules}` include `fsdp`, `megatron`, `veomni`, `sglang`, `vllm`, `rollout`, `trainer`, `ci`, `training_utils`, `recipe`, `hardware`, `deployment`, `ray`, `worker`, `single_controller`, `misc`, `perf`, `model`, `algo`, `env`, `tool`, `ckpt`, `doc`, `data`, `cfg`, `reward`, `fully_async`, `one_step_off`
  - If this PR involves multiple modules, separate them with `,` like `[megatron, fsdp, doc]`
  - `{type}` is in `feat`, `fix`, `refactor`, `chore`, `test`
  - If this PR breaks any API (CLI arguments, config, function signature, etc.), add `[BREAKING]` to the beginning of the title.
  - Example: `[BREAKING][fsdp, megatron] feat: dynamic batching`

### Test

> For changes that can not be tested by CI (e.g., algorithm implementation, new model support), validate by experiment(s) and show results like training curve plots, evaluation results, etc.

### API and Usage Example

> Demonstrate how the API changes if any, and provide usage example(s) if possible.

```python
# Add code snippet or script demonstrating how to use this
```

### Design & Code Changes

> Demonstrate the high-level design if this PR is complex, and list the specific changes.

### Checklist Before Submitting

> [!IMPORTANT]
> Please check all the following items before requesting a review, otherwise the reviewer might deprioritize this PR for review.

- [ ] Read the [Contribute Guide](https://github.com/volcengine/verl/blob/main/CONTRIBUTING.md).
- [ ] Apply [pre-commit checks](https://github.com/volcengine/verl/blob/main/CONTRIBUTING.md#code-linting-and-formatting): `pre-commit install && pre-commit run --all-files --show-diff-on-failure --color=always`
- [ ] Add / Update [the documentation](https://github.com/volcengine/verl/tree/main/docs).
[...]

Issue #2603, opened 2025-07-17
RuntimeError: Tensor size mismatch on apply_multimodal_rotary_pos_emb with ulysses_sequence_parallel_size=2


Yang011013:
GPU: 2 * H20；
config.yaml:
```
actor_rollout_ref:
  actor:
    strategy: fsdp
    ppo_mini_batch_size: 1
    ppo_micro_batch_size: null
    ppo_micro_batch_size_per_gpu: 1
    use_dynamic_bsz: false
    ppo_max_token_len_per_gpu: 18432
    clip_ratio: 0.2
    clip_ratio_low: 0.2
    clip_ratio_high: 0.2
    policy_loss:
      loss_mode: vanilla
      clip_cov_ratio: 0.0002
      clip_cov_lb: 1.0
      clip_cov_ub: 5.0
      kl_cov_ratio: 0.0002
      ppo_kl_coef: 0.1
    clip_ratio_c: 3.0
    loss_agg_mode: token-mean
    entropy_coeff: 0
    use_kl_loss: false
    use_torch_compile: true
    kl_loss_coef: 0.01
    kl_loss_type: low_var_kl
    ppo_epochs: 1
    shuffle: false
    checkpoint:
      save_contents:
      - model
      - optimizer
      - extra
      - hf_model
      load_contents: ${.save_contents}
    optim:
      lr: 3.0e-06
      lr_warmup_steps_ratio: 0.0
      total_training_steps: -1
      weight_decay: 0.01
      lr_warmup_steps: -1
      min_lr_ratio: 0.0
      num_cycles: 0.5
      warmup_style: constant
      strategy: adamw_bf16
    grad_clip: 1.0
    ulysses_sequence_parallel_size: 2
    entropy_from_logits_with_chunking: false
    entropy_checkpointing: false
    fsdp_config:
      wrap_policy:
        min_num_params: 0
      param_offload: true
      optimizer_offload: true
      offload_policy: false
      reshard_after_forward: true
      fsdp_size: -1
      forward_prefetch: false
      model_dtype: bfloat16
  ref:
    strategy: ${actor_rollout_ref.actor.strategy}
    use_torch_compile: ${oc.select:actor_rollout_ref.actor.use_torch_compile,true}
    log_prob_micro_batch_size: null
    log_prob_micro_batch_size_per_gpu: 1
    log_prob_use_dynamic_bsz: ${oc.select:actor_rollout_ref.actor.use_dynamic_bsz,false}
[...]

PR #7620, opened 2026-08-31
[model] fix: use the transformers 5 rotary API in the GLM-4V attention patch

alanhuangyoo:
### What does this PR do?

GLM-4V training dies at the first attention layer on every transformers version this repo
supports.

`glm4v_attn_forward` replaces `Glm4vTextAttention.forward` whenever `use_remove_padding` or
`ulysses_sp_size > 1` (`monkey_patch.py:473`), so this is the ordinary GLM-4V path, not a corner:

```python
from transformers.models.glm4v.modeling_glm4v import apply_multimodal_rotary_pos_emb, repeat_kv
```

```
transformers 5.3.0
ImportError: cannot import name 'apply_multimodal_rotary_pos_emb'
             from 'transformers.models.glm4v.modeling_glm4v'
```

transformers 5.0 replaced `apply_multimodal_rotary_pos_emb(q, k, cos, sin, mrope_section)` with
`apply_rotary_pos_emb(q, k, cos, sin)` and moved the sectioning into
`Glm4vTextRotaryEmbedding`, which reads `mrope_section` from `config.rope_parameters` and folds
it into cos/sin before the attention ever sees them. `Glm4vTextAttention.forward` upstream is
now just:

```python
cos, sin = position_embeddings
query_states, key_states = apply_rotary_pos_emb(query_states, key_states, cos, sin)
```

Absent in every supported version — `requirements.txt` asks for
`transformers>=5.5.3,!=5.6.0,<5.11`, and the extras pin 5.3.0 / 5.5.3:

| transformers | `apply_multimodal_rotary_pos_emb` in `modeling_glm4v` |
| --- | --- |
| 4.57.1 | yes |
| 5.0.0, 5.3.0, 5.5.3, 5.10.4, 5.16.1 | no |

It came in with #5724 (`transformers upgrade to 5.3.0`), which updated the config-side
`rope_scaling` → `rope_parameters` migration a few lines above but not the function that moved
with it.

`qwen2_vl.py` uses the same call and is **not** affected: `modeling_qwen2_vl` still exports
`apply_multimodal_rotary_pos_emb`. Only GLM-4V was migrated.

### The change

Call the function upstream calls, and drop the `mrope_section` lookup that fed the old one.

That lookup was dead twice over — it reads `self.rope_scaling` or `self.rope_parameter` off the
**attention module**, and on transformers 5 the attention module has neither (the config
attribute is `rope_parameters`, plural, and lives on the config):

```
transformers 5.3.0, Glm4vTextAttention
  self.rope_scaling    -> <ABSENT>
  self.rope_parameter  -> <ABSENT>
```

so the `assert` under it would have fired had the import ever got that far.

Restoring the old function instead would be wrong, not just unavailable: the 4.x call expects
cos/sin with a leading 3-section dim, and applying it to what transformers 5 produces raises
`IndexError: index 1 is out of bounds for dimension 0 with size 1`. The sectioning genuinely
moved.

### Checklist Before Starting

- [x] Search for similar PRs. No PR or issue mentions `apply_multimodal_rotary_pos_emb` on this
      repo; the one hit, #2603, predates the 5.x upgrade and is closed.

### Test

`tests/models/test_glm4v_rotary_on_cpu.py`, following the `on_cpu.py` convention so it runs in
`cpu_unit_tests.yml`. It builds a real `Glm4vTextAttention` and `Glm4vTextRotaryEmbedding`, runs
`glm4v_attn_forward` with the CUDA-only flash-attention call patched out, and asserts the q/k
reaching attention equal `apply_rotary_pos_emb(q, k, cos, sin)` — the same rotary upstream
applies — with KV repeated to the query head count. A second case records why the argument had
to go rather than be restored.

Load-bearing, run against a clean `origin/main` worktree in the same environment:

```
$ pytest tests/models/test_glm4v_rotary_on_cpu.py          # origin/main
>       from transformers.models.glm4v.modeling_glm4v import apply_multimodal_rotary_pos_emb, repeat_kv
E       ImportError: cannot import name 'apply_multimodal_rotary_pos_emb' ...
1 failed, 1 passed

$ pytest tests/models/test_glm4v_rotary_on_cpu.py          # this branch
2 passed

$ ruff check / ruff format --check                          # ruff 0.12.2, as pinned in .pre-commit-config.yaml
All checks passed! / 2 files already formatted

$ python3 tests/special_sanity/check_license.py --directories .
exit 0
```

The flash-attention call itself is CUDA-only and unchanged by this PR, so it stays covered by
the existing GPU tests.

### API and Usage Example

No API change.

### Design & Code Changes

One import and one call site in `verl/models/transformers/glm4v.py`; the `mrope_section`
derivation is removed with them.

- [x] Apply [pre-commit checks](https://github.com/volcengine/verl/blob/main/CONTRIBUTING.md#code-linting-and-formatting).
- [x] Add / update the documentation. Not applicable.
- [x] Add unit or end-to-end test(s).


tests/models/test_glm4v_rotary_on_cpu.py
```diff
@@ -0,0 +1,112 @@
...
+`Glm4vTextRotaryEmbedding` and exposes only `apply_rotary_pos_emb`, so the import raised
+`ImportError` at the first attention layer on every version this repo supports
+(`transformers>=5.5.3,!=5.6.0,<5.11`).
+
+The flash-attention call is CUDA-only, so it is patched out here; what this pins is the
+rotary step, which is what changed.
+"""
+
+import torch
+from transformers.models.glm4v.configuration_glm4v import Glm4vTextConfig
```
wuxibin89: CI not needed.

alanhuangyoo: Removed. The import and rotary-call fix in `glm4v.py` stands on its own — `apply_multimodal_rotary_pos_emb` does not exist in any transformers this repo supports, so the GLM-4V path raises ImportError at the first attention layer with remove_padding or ulysses_sp_size > 1.

alanhuangyoo:
The `vllm` / `vllm_ascend` reds here are not from this PR — it touches one file,
`verl/models/transformers/glm4v.py`, and neither job exercises it.

```
tests/workers/rollout/rollout_vllm/test_vllm_abort.py::test_abort_all_requests_propagates_pause_failure
    AttributeError: 'vLLMHttpServer' object has no attribute '_resume_event'
    verl/workers/rollout/vllm_rollout/vllm_async_server.py:942
```

That test builds the server with `vLLMHttpServer.__new__(vLLMHttpServer)`, so `__init__` never
runs and neither `_submission_paused` nor `_resume_event` is set. `abort_all_requests` started
touching both in #7511 ("gate rollout submission during weight-sync drain for DP>1"), which
lands before the `pause_generation` call the test is actually asserting on. The test file has
not been updated since.

Happy to send that as a separate one-liner if it is not already being handled.

PR #7628, opened 2026-08-31
[rollout] fix: continuous token fuse generation prompt with the final append group

### LINKED ISSUE #7617: # [rollout] perf: `_tokenize_generation_prompt_delta` re-tokenizes the full conversation after every tool turn

gxlvera:
### What does this PR do?

Fixes #7617.

`ContinuousTokenBuilder.tokenize_non_assistant_incremental_messages()` previously encoded the complete updated message history twice after every appended non-assistant group solely to derive the tokens added by `add_generation_prompt=True`. For long agent trajectories, this repeatedly processes the growing history and results in approximately O(N^2) tokenization work.

This PR changes the default behavior to fuse generation-prompt rendering into the final append group. The final group is rendered with `add_generation_prompt=True`, so its delta contains both the appended non-assistant messages and the following generation prompt. This removes the separate full-history generation-prompt render from the default path.

Some chat templates may derive their generation prompt from context that is not fully determined by the final append group. Model-specific builders can therefore opt out by returning `False` from `_should_fuse_generation_prompt_with_last_group()` and can preserve or implement their own behavior through `_tokenize_generation_prompt_delta()`.

### Test

- `test_default_builder_fuses_generation_prompt_into_last_append_group`: verifies that a final multi-message tool group performs only two bounded renders and that the second render uses `add_generation_prompt=True`.
- `test_default_builder_only_fuses_generation_prompt_into_final_append_group`: verifies that only the final group receives the generation prompt when multiple groups are appended.
- `test_special_builder_can_keep_separate_full_history_generation_prompt`: verifies that an opted-out builder still invokes `_tokenize_generation_prompt_delta()` and can retain the full-history false/true suffix-diff behavior.

And e2e trial that validates the perf gain is being produced. 

### Design & Code Changes
There is no change to the AgentLoop-facing runtime API. But only Developer extension APIs.
Changes are made as below
- Materialize append groups before tokenization so the final group can be identified.
- Pass `add_generation_prompt=True` only to the final append group on the default fused path.
- Extend `_tokenize_tool_group()` and `_tokenize_single_non_tool()` with an `add_generation_prompt` keyword argument and forward it to `render_delta_token_id()`.
- Add `_should_fuse_generation_prompt_with_last_group()` as a builder-level capability hook; the base implementation returns `True`.
- Preserve `_tokenize_generation_prompt_delta()` as the separate generation-prompt extension point for opted-out builders.
- Keep GPT-OSS and Gemma 4 on their existing model-specific separate paths.
- Add CPU unit tests for fusion, final-group selection, argument propagation, and the full-history fallback.


gxlvera:
Can anyone check for deepseek model family, which has special behavior.

wuxibin89 (approved):

ruiling-smartbear:
Checked the merge commit (3dab856) against its parent (8e4a572) on real tokenizers (transformers 5.16.1, tokenizer files only), same setup as the tables on #7617: trajectories shaped like `tool_agent_loop`'s (dict `arguments` from `OpenAIFunctionCallSchema`, tool messages with `content` + `tool_call_id`), the last tool group appended, token ids from `tokenize_non_assistant_incremental_messages` compared case by case. 24 cases per row: system prompt on/off × 0 / 10 / 50 prior tool turns × 1 or 2 parallel calls × `enable_thinking` default / False.

| builder | model | after == before |
|---|---|---|
| qwen | Qwen2-7B-Instruct | 24/24 |
| qwen25 | Qwen2.5-7B-Instruct | 24/24 |
| qwen3 | Qwen3-8B | 24/24 |
| qwen35 | Qwen3.5-9B | 24/24 |
| minimaxm2 | MiniMax-M2 | 24/24 |
| glm47 | GLM-4.7 | 24/24 |
| gemma4 | gemma-4-12b-it | 24/24 (opted out, same path on both sides) |
| gptoss | gpt-oss-20b | 24/24 (opted out) |
| default | Qwen3-8B via the base builder | 24/24 |

Where the naive full-history diff can serve as a third reference (all rows except Qwen3-8B, whose template is not prefix stable) it agrees with both sides too, gpt-oss aside for the known `_format_tool_response` reason.

@gxlvera on DeepSeek: V3, R1, V3.1 and V3.2-Exp all have `model_type: deepseek_v3`, so they go through `DeepSeekContinuousTokenBuilder` and the base fused path.

- The fusing changes nothing for them. All four templates gate the generation prompt on `not ns.is_tool`, so after a tool message neither the old full-history render nor the fused render adds `<｜Assistant｜>`. Every case that renders is identical before and after.
- On R1, V3.1 and V3.2-Exp nothing renders on either side: `_synthetic_assistant_for_tools` sets `"arguments": {}` and those templates concatenate `tool['function']['arguments']` as a string, so every tool append raises `TypeError` — before and after, pre-existing. V3 only escapes because its template renders `tool_calls` when `content is none`, and the synthetic message has `content: ""`.
- With the synthetic arguments as the string `"{}"` the fused path renders on all four, and matches the pre-merge path 24/24 wherever that path can render at all. Before this PR the same fix would not have been enough, because the two full-history renders also hit the harness's own dict `arguments`; now the history is never re-rendered on a tool append, so that stops mattering. Happy to send that one-line override for `DeepSeekContinuousTokenBuilder` if you want it.
- Unrelated to this PR but visible in the same run: V3/R1 keep a conversation-global `is_output_first`, so from the second tool turn on the template writes `\n<｜tool▁output▁begin｜>` without `<｜tool▁outputs▁begin｜>`, while the bounded synthetic-prefix render always writes the first-turn form.

DeepSeek-V4 has its own `tokenize_non_assistant_incremental_messages` in `deepseek.py`, which this PR does not touch.

Script, raw output and the per-variant DeepSeek table: https://github.com/ruiling-smartbear/verl-experiments/blob/main/generation_prompt_fold_experiment.md#after-the-merge--7628-vs-its-parent

ruiling-smartbear:
Opened #7630 with the DeepSeek override and a CPU test, checked on the real R1 / V3.1 / V3.2-Exp tokenizers against current main.

Issue #7624, opened 2026-08-31
[bug] `enable_chunked_prefill=False` crashes vLLM startup on verl main + vllm 0.24 (`max_num_batched_tokens` < `max_model_len`) — regression from #7508


LiRunGuo:
## System Info

- verl main @ 8f8b122 (and current main; reproduced on 2026-08-27)
- vllm 0.24.0 (verl's pinned version)
- 8x A800 80GB, FSDP engine, Llama-3.2-1B-Instruct, GRPO GSM8K example config

## Bug Description

With `actor_rollout_ref.rollout.enable_chunked_prefill=False`, the vLLM rollout
server aborts at startup:

```
ValidationError: 1 validation error for SchedulerConfig
  Value error, max_num_batched_tokens (8192) is smaller than max_model_len (131072). ...
```

verl's default `max_num_batched_tokens` is 8192, which is below every modern
model's `max_position_embeddings` (e.g. 131072 for Llama-3.2-1B). vLLM >= 0.24
rejects this combination in `verify_max_model_len`
(`vllm/config/scheduler.py`) when `enable_chunked_prefill=False`.

**This is a regression introduced by #7508** ("honor explicit False on
Optional[bool] engine args in CLI serialization", merged 2026-08-25): before it,
explicit `False` was silently dropped, so `enable_chunked_prefill=False` never
reached the engine; now it does and trips vLLM's validation. PR #7558/#7584
removed the flag from the Ascend NPU scripts as a workaround, but the GPU
example scripts still carry it.

## Reproduction

```bash
# any official example with enable_chunked_prefill=False, e.g.:
bash examples/grpo_trainer/run_qwen3_4b_fsdp.sh
# -> vLLMHttpServer aborts: max_num_batched_tokens (8192) < max_model_len (131072)
```

Minimal (offline, no GPU training):

```python
from vllm.engine.arg_utils import AsyncEngineArgs
from vllm.usage.usage_lib import UsageContext

args = AsyncEngineArgs(
    model="meta-llama/Llama-3.2-1B-Instruct",
    max_num_batched_tokens=8192,          # verl default
    enable_chunked_prefill=False,         # verl config honored since #7508
    tensor_parallel_size=1,
)
args.create_engine_config(usage_context=UsageContext.OPENAI_API_SERVER)
# ValidationError: max_num_batched_tokens (8192) is smaller than max_model_len (131072)
```

## Impact

At least 13 official example scripts under `examples/` set
[...]

PR #7632, opened 2026-08-31
[vllm] fix: raise max_num_batched_tokens to max_model_len when chunked prefill is disabled

ETOgaosion:
## What does this PR do?

Fixes a vLLM engine startup failure when chunked prefill is disabled on a model whose `max_model_len` exceeds the rollout default `max_num_batched_tokens` (8192).

vLLM requires `max_num_batched_tokens >= max_model_len` whenever `enable_chunked_prefill=False`, since each prefill must fit within a single scheduling iteration. verl always passes `max_num_batched_tokens` to vLLM (default 8192), which overrides vLLM's own default of `max_model_len` for the non-chunked case, so such jobs crash at engine startup with:

```
pydantic_core._pydantic_core.ValidationError: 1 validation error for SchedulerConfig
  Value error, max_num_batched_tokens (8192) is smaller than max_model_len (128000).
  This effectively limits the maximum sequence length to max_num_batched_tokens and
  makes vLLM reject longer sequences. Please increase max_num_batched_tokens or
  decrease max_model_len.
```

This is hit by the geo3k VLM e2e tests (`Qwen/Qwen2.5-VL-3B-Instruct`, `ENABLE_CHUNKED_PREFILL=False` to work around the VLM placeholder issue [vllm#15185](https://github.com/vllm-project/vllm/issues/15185)) on vllm 0.24, where `max_model_len` defaults to the model's 128k position embeddings.

The fix has two parts:

1. **Engine-side safety net** ([vllm_async_server.py](https://github.com/verl-project/verl/blob/main/verl/workers/rollout/vllm_rollout/vllm_async_server.py)): in `_validate_configs`, when chunked prefill is disabled and `max_num_batched_tokens < max_model_len`, raise `max_num_batched_tokens` to `max_model_len` (with a warning) instead of letting the engine fail. This mirrors vLLM's own default. Explicitly configured larger values are respected. sglang / trtllm backends don't pass this arg and are unaffected.
2. **Example script cleanup** (following #7629 / #7624): the stale `enable_chunked_prefill=False` overrides in GPU vLLM GRPO/LoRA/profiling example scripts were workarounds for the old VLM placeholder limitation ([vllm#15185](https://github.com/vllm-project/vllm/issues/15185), fixed in March 2025), and `rollout.yaml` already defaults `enable_chunked_prefill` to `True`. Remove them so the examples follow the default and work out of the box. The torch profiler script intentionally keeps chunked prefill disabled for clean traces, and now pins `max_model_len=max_prompt_length+max_response_length` as required by that setting. NPU scripts are untouched.

Also unblocks the vLLM agent-loop CI job after #7613: least-loaded replica ties are now random, so tests that hardcoded `s0`/`s1`/`s2` assignment (`test_new_requests_route_to_least_loaded`, `test_get_inflight_count`, `test_removed_server_invalidates_sticky_session`) are updated to observe the assigned replica instead of assuming dict order. This is not a duplicate of an existing PR; #7613 changed the balancer but left these GPU tests asserting the old deterministic tie-break.

## Checklist Before Starting

- [x] Search for similar PRs

## Test

- [x] ruff check / pre-commit pass on the modified files
- [x] `pytest tests/experimental/agent_loop/test_basic_agent_loop.py::TestLoadBalancerRouting tests/experimental/agent_loop/test_basic_agent_loop.py::TestLoadBalancerStickySession tests/experimental/agent_loop/test_basic_agent_loop.py::TestLoadBalancerHybrid -v` — 15 passed
- [ ] CI: the geo3k VLM function-reward e2e tests (`tests/special_e2e/ppo_trainer/run_function_reward.sh` with `ENABLE_CHUNKED_PREFILL=False`) should now reach training instead of failing at vLLM server launch
- [ ] CI: vLLM agent-loop job should pass after the load-balancer test update

AI assistance was used to diagnose the vLLM CI failure and update the load-balancer tests.


wuxibin89:
From https://github.com/verl-project/verl/pull/7629 https://github.com/verl-project/verl/issues/7624, I think we should provide `max_model_len` explictly if `enable_chunked_prefill=False`:
```bash
actor_rollout_ref.rollout.max_model_len="${MAX_MODEL_LEN}"
```

BTW, can we always `enable_chunked_prefill=True` for newer vllm version?

ETOgaosion:
> can we always enable_chunked_prefill=True for newer vllm version?

I think we can make this enabled by default.

wuxibin89 (dismissed):

wuxibin89 (approved):

PR #7630, opened 2026-08-31
[rollout] fix: DeepSeek continuous token builder cannot render tool appends

ruiling-smartbear:
### What does this PR do?

`ContinuousTokenBuilder._tokenize_tool_group()` renders a tool group behind a synthetic assistant message whose tool calls carry `"arguments": {}`. The DeepSeek R1, V3.1 and V3.2 chat templates splice `tool['function']['arguments']` into the prompt by string concatenation, so with `DeepSeekContinuousTokenBuilder` every tool append raises `TypeError` on that concatenation. DeepSeek-V3 is the only one of the family that gets through, and only because its template renders `tool_calls` when `content is none`, which the synthetic message never is.

Found while checking #7628 on the DeepSeek family (https://github.com/verl-project/verl/pull/7628#issuecomment-5474456502). The fix is a one-method override: the DeepSeek builder carries the synthetic arguments as the JSON string `"{}"`. The message only exists to be diffed away, so the emitted tokens do not change. #7628 is what makes this sufficient: the history, with the harness's own dict `arguments` from `OpenAIFunctionCallSchema`, is no longer re-rendered on a tool append.

The base builder is left alone; Qwen-style templates run `arguments` through `tojson` and expect a mapping.

### Checklist Before Starting

- [x] Search for similar PRs: https://github.com/verl-project/verl/pulls?q=is%3Apr+deepseek+continuous+token
- [x] PR title formatted as `[{modules}] {type}: {description}`

### Test

- `tests/utils/test_continuous_token_on_cpu.py`: 124 passed (122 existing + 2 new). The new DeepSeek-shaped fake template concatenates arguments like the real ones; on `main` the new test fails with the same `TypeError`.
- Real tokenizers (tokenizer files only, transformers 5.16.1), upstream `main` @3dab856 vs `main` with this patch, token ids from `tokenize_non_assistant_incremental_messages` on `tool_agent_loop`-shaped trajectories, 24 cases per row (system prompt on/off × 0 / 10 / 50 prior tool turns × 1 or 2 parallel calls × `enable_thinking` default / False):

| model | harness `arguments` | `main` | this PR |
|---|---|---|---|
| DeepSeek-V3 | dict, json | 24/24 render | 24/24, identical to `main` |
| DeepSeek-R1 | dict, json | 0/24 (`TypeError`) | 24/24 render |
| DeepSeek-V3.1 | dict, json | 0/24 | 24/24 |
[...]

PR #7638, opened 2026-08-31
[rollout] fix: fall back to a default Continuous Token builder when a required token is missing

### LINKED ISSUE #7637: [rollout] Continuous Token: DeepSeek-R1 distills cannot construct a builder since #6804 (Qwen builder requires <|im_end|>)
### Describe the bug

Since #6804 the agent loop builds a Continuous Token builder for every rollout and picks the builder family by the root `config.json` `model_type`. The DeepSeek-R1 distills have Qwen architectures (`model_type: qwen2` / `qwen3`) but DeepSeek's chat template and tokenizer, in which `<|im_end|>` was renamed to `<｜end▁of▁sentence｜>`. `QwenContinuousTokenBuilder.__init__` requires `<|im_end|>`, so `AgentLoopBase.__init__` raises before any rollout starts:

```
ValueError: Tokenizer does not define required token '<|im_end|>'
```

| model | `model_type` | family picked | result |
|---|---|---|---|
| deepseek-ai/DeepSeek-R1-0528-Qwen3-8B | qwen3 | qwen3 | construction fails |
| deepseek-ai/DeepSeek-R1-Distill-Qwen-7B | qwen2 | qwen | construction fails |
| deepseek-ai/DeepSeek-R1-Distill-Qwen-1.5B | qwen2 | qwen | construction fails |

Reproduction with tokenizer files only (verl main @9c76436, transformers 5.16.1):

```python
from verl.utils.tokenizer.continuous_token_wiring import create_continuous_token_builder

model = "deepseek-ai/DeepSeek-R1-Distill-Qwen-7B"
create_continuous_token_builder(AutoTokenizer.from_pretrained(model), hf_model_type=AutoConfig.from_pretrained(model).model_type)
```

This is `main` only: v0.9.0 was cut just before #6804, and before #6804 Continuous Token was opt-in and the name-based inference sent the `DeepSeek-R1-Distill-Qwen-*` checkpoints to the default builder (`R1-0528-Qwen3-8B` already matched the `qwen3` rule, but the path was off by default). `DeepSeek-R1-Distill-Llama-8B` has `model_type: llama`, so it gets the default builder and constructs.

### What renders these templates correctly

Tool appends on the first and third tool turn, compared with the template's full-conversation render (script and raw output: https://github.com/ruiling-smartbear/verl-experiments/blob/main/verl_distill_builder_probe_results.txt):

- `DeepSeekContinuousTokenBuilder` renders all four distills. R1-0528-Qwen3-8B matches the template on every turn; the older distills match on the first tool turn and then differ by the `is_output_first` form (#7636).
- The default builder renders the three older distills identically, but hits the #7630 `TypeError` on R1-0528-Qwen3-8B: the base `_synthetic_assistant_for_tools` still carries `"arguments": {}` and that template concatenates arguments as a string. With `"arguments": "{}"` in the base class it matches the template on every turn.

### Possible fixes

Keeping #6804's rules (no toggle, nothing guessed from names):

1. When the builder resolved from `model_type` cannot be constructed because the tokenizer lacks its required token, fall back to the default builder with a warning — the treatment an unknown `model_type` already gets — and carry the base synthetic tool-call arguments as a JSON string so the default builder also renders concatenating templates. Smallest change; the distills then behave exactly like the DeepSeek builder on `main`.
2. Resolve the family from the tokenizer's special tokens as an exact secondary key (`<｜end▁of▁sentence｜>` and `<｜Assistant｜>` defined, `<|im_end|>` not): these checkpoints then get the DeepSeek builder and #7630 / #7636 apply to them directly.
3. Bring back an explicit family override — least in the spirit of #6804.

Known limitations either way: from the second tool turn on, the older distills' `is_output_first` form needs #7636 (or its logic in whichever builder serves them); and DeepSeek-R1-0528-Qwen3-8B's tokenizer encodes and decodes incorrectly under transformers 5 (decoded text loses spaces and the `｜`/`▁` marker characters) unless loaded as `Qwen2TokenizerFast` — its tokenizer_config.json declares LlamaTokenizerFast; outside verl, fix proposed upstream in https://huggingface.co/deepseek-ai/DeepSeek-R1-0528-Qwen3-8B/discussions/28.

Happy to send a PR for whichever you prefer.

ruiling-smartbear:
### What does this PR do?

Fixes #7637: the DeepSeek-R1 distills keep Qwen `model_type` values (`qwen2` / `qwen3`) but ship DeepSeek's tokenizer, in which `<|im_end|>` is renamed to `<｜end▁of▁sentence｜>`. Since #6804 the builder family is an exact `model_type` lookup, so `QwenContinuousTokenBuilder.__init__` raises `Tokenizer does not define required token '<|im_end|>'` in `AgentLoopBase.__init__` before any rollout starts.

With `auto` inference, a builder whose constructor raises for a missing required token now falls back to the default behavior with a warning — the treatment an unregistered `model_type` already gets. The fallback is a dedicated `FallbackContinuousTokenBuilder` that carries the synthetic tool-call arguments as the JSON string form of #7630, because these checkpoints' templates splice `function['arguments']` in by string concatenation. The string form cannot go into the base class: Qwen3.5 / MiniMax-M2 / GLM-4.7 templates render the synthetic arguments as a mapping (`.items()`), which the harness below confirms, so the base class keeps the mapping form and registered families are untouched. An explicitly requested family still raises, and `require_token_id` raises `MissingRequiredTokenError` (a `ValueError` subclass) for the token-absent case so the factory catches exactly that.

Alternative to #7639, which routes these checkpoints to the DeepSeek builder from the tokenizer's own special tokens instead.

### Checklist Before Starting

- [x] Search for similar PRs: https://github.com/verl-project/verl/pulls?q=is%3Apr+distill+continuous+token

### Test

- `tests/utils/test_continuous_token_on_cpu.py`: the factory falls back to `FallbackContinuousTokenBuilder` for a distill-style tokenizer and renders its concatenating template; an explicit family still raises; the fallback's synthetic arguments are a JSON string while the base builder keeps the mapping form. On `main` the first of these fails at construction.
  - the 13-row × 24-case tool-append harness from #7630 / #7636 (Qwen2 / 2.5 / 3 / 3.5, MiniMax-M2, GLM-4.7, Gemma 4, gpt-oss, DeepSeek-V3 / R1 / V3.1 / V3.2-Exp, default): every row byte-identical to `main`. The same harness is what rules out the base-class variant: with string arguments in the base class instead, the Qwen3.5-9B, MiniMax-M2 and GLM-4.7 rows fail all 24 cases (`'str object' has no attribute 'items'`);
  - `create_continuous_token_builder` on the three broken distills: raises on `main`, returns the fallback builder with this change, and the first tool turn matches the template's full-conversation render for all of them (DeepSeek-R1-0528-Qwen3-8B on every turn; the older distills' later turns keep the known `is_output_first` gap tracked in #7636);
  - Qwen3-8B, Qwen2.5-7B-Instruct, DeepSeek-V3.1 and DeepSeek-R1-Distill-Llama-8B select the same builders as `main`.

  Script and raw output: https://github.com/ruiling-smartbear/verl-experiments/blob/main/verl_distill_fix_check_results.txt

### API and Usage Example

No API change.

### Design & Code Changes

- `continuous_token.py`: `MissingRequiredTokenError`; `FallbackContinuousTokenBuilder` (default builder + JSON-string synthetic tool-call arguments).
- `continuous_token_wiring.py`: `create_continuous_token_builder` catches `MissingRequiredTokenError` on the text path and falls back to `FallbackContinuousTokenBuilder` with a warning, only for `auto`.

- [x] Docs: internal tokenizer utility, no user-facing change.

### DIFF
--- a/tests/utils/test_continuous_token_on_cpu.py
+++ b/tests/utils/test_continuous_token_on_cpu.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import json
 import logging
 
 import pytest
@@ -20,6 +21,7 @@ from verl.utils.tokenizer.continuous_token import (
     ContinuousTokenBuilder,
     DeepSeekContinuousTokenBuilder,
     DeepSeekVL2ContinuousTokenBuilder,
+    FallbackContinuousTokenBuilder,
     Gemma4ContinuousTokenBuilder,
     GLM46VContinuousTokenBuilder,
     GLMContinuousTokenBuilder,
@@ -276,6 +278,26 @@ class _DeepSeekBoundaryTokenizer(_TemplateTokenizer):
         return rendered
 
 
+class _DistillTokenizer(_DeepSeekBoundaryTokenizer):
+    """DeepSeek-R1 distill of Qwen: the root config keeps a Qwen ``model_type`` but the
+    checkpoint ships DeepSeek's tokenizer and chat template, in which ``<|im_end|>``
+    does not exist (renamed to ``<｜end▁of▁sentence｜>``)."""
+
+    name_or_path = "deepseek-ai/DeepSeek-R1-Distill-Qwen-7B"
+    unk_token_id = None
+
+    def __init__(self):
+        super().__init__()
+        self.assistant_id = 2
+
+    def convert_tokens_to_ids(self, token):
+        if token == "<｜end▁of▁sentence｜>":
+            return self.eos_id
+        if token == "<｜Assistant｜>":
+            return self.assistant_id
+        return None
+
+
 class _MissingSpecialTokenTokenizer(_TemplateTokenizer):
     def convert_tokens_to_ids(self, token):
         return None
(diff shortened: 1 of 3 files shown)


wuxibin89:
Prefer explicitly specify `continuous_token.model_family` if auto resolve is not expected. Fixed in https://github.com/verl-project/verl/pull/7640

PR #7639, opened 2026-08-31
[rollout] fix: resolve the Continuous Token family from tokenizer special tokens for the DeepSeek-R1 distills

### LINKED ISSUE #7637: [rollout] Continuous Token: DeepSeek-R1 distills cannot construct a builder since #6804 (Qwen builder requires <|im_end|>)

ruiling-smartbear:
### What does this PR do?

Fixes #7637: the DeepSeek-R1 distills keep Qwen `model_type` values (`qwen2` / `qwen3`) but ship DeepSeek's tokenizer, in which `<|im_end|>` is renamed to `<｜end▁of▁sentence｜>`. Since #6804 the builder family is an exact `model_type` lookup, so `QwenContinuousTokenBuilder.__init__` raises `Tokenizer does not define required token '<|im_end|>'` in `AgentLoopBase.__init__` before any rollout starts.

When the builder inferred with `auto` cannot be constructed because a required token is missing, the factory now consults an exact special-token key on the tokenizer itself: `<｜end▁of▁sentence｜>` and `<｜Assistant｜>` defined and `<|im_end|>` not defined selects the DeepSeek builder, so #7630 (and #7636 once merged) apply to these checkpoints directly. Any other tokenizer re-raises, an explicitly requested family still raises, and nothing is guessed from repository or tokenizer names — matching stays exact, in the spirit of #6804. `require_token_id` raises `MissingRequiredTokenError` (a `ValueError` subclass) for the token-absent case so the factory catches exactly that.

Alternative to #7638, which falls back to the default builder instead.

### Checklist Before Starting

- [x] Search for similar PRs: https://github.com/verl-project/verl/pulls?q=is%3Apr+distill+continuous+token

### Test

- `tests/utils/test_continuous_token_on_cpu.py`: the factory resolves the DeepSeek builder for a distill-style tokenizer and renders its concatenating template; the key requires every marker (a tokenizer with fewer re-raises the original error); an explicit family still raises. On `main` the first of these fails at construction.
  - the 13-row × 24-case tool-append harness from #7630 / #7636 (Qwen2 / 2.5 / 3 / 3.5, MiniMax-M2, GLM-4.7, Gemma 4, gpt-oss, DeepSeek-V3 / R1 / V3.1 / V3.2-Exp, default): every row byte-identical to `main`;
  - `create_continuous_token_builder` on the three broken distills: raises on `main`, returns `DeepSeekContinuousTokenBuilder` with this change, and the first tool turn matches the template's full-conversation render for all of them (DeepSeek-R1-0528-Qwen3-8B on every turn; the older distills' later turns keep the known `is_output_first` gap tracked in #7636);
  - the special-token key fires for exactly the three broken distills: Qwen3-8B / Qwen2.5-7B-Instruct define `<|im_end|>`, DeepSeek-V3.1 / R1 construct from `model_type` before the key is consulted, DeepSeek-R1-Distill-Llama-8B keeps the default builder (`llama` is unregistered, so construction never fails).

  Script and raw output: https://github.com/ruiling-smartbear/verl-experiments/blob/main/verl_distill_fix_check_results.txt

### API and Usage Example

No API change.

### Design & Code Changes

- `continuous_token.py`: `MissingRequiredTokenError`.
- `continuous_token_wiring.py`: `_family_from_tokenizer_special_tokens` (key tokens reuse `DeepSeekContinuousTokenBuilder`'s own constants); `create_continuous_token_builder` consults it only after a `MissingRequiredTokenError` under `auto`.

- [x] Docs: internal tokenizer utility, no user-facing change.

### DIFF
--- a/tests/utils/test_continuous_token_on_cpu.py
+++ b/tests/utils/test_continuous_token_on_cpu.py
@@ -33,6 +33,7 @@ from verl.utils.tokenizer.continuous_token import (
 from verl.utils.tokenizer.continuous_token_wiring import (
     CONTINUOUS_TOKEN_BUILDER_FAMILIES,
     ContinuousTokenModelFamily,
+    _family_from_tokenizer_special_tokens,
     create_continuous_token_builder,
     get_continuous_token_builder_class,
     infer_continuous_token_model_family,
@@ -276,6 +277,26 @@ class _DeepSeekBoundaryTokenizer(_TemplateTokenizer):
         return rendered
 
 
+class _DistillTokenizer(_DeepSeekBoundaryTokenizer):
+    """DeepSeek-R1 distill of Qwen: the root config keeps a Qwen ``model_type`` but the
+    checkpoint ships DeepSeek's tokenizer and chat template, in which ``<|im_end|>``
+    does not exist (renamed to ``<｜end▁of▁sentence｜>``)."""
+
+    name_or_path = "deepseek-ai/DeepSeek-R1-Distill-Qwen-7B"
+    unk_token_id = None
+
+    def __init__(self):
+        super().__init__()
+        self.assistant_id = 2
+
+    def convert_tokens_to_ids(self, token):
+        if token == "<｜end▁of▁sentence｜>":
+            return self.eos_id
+        if token == "<｜Assistant｜>":
+            return self.assistant_id
+        return None
+
+
 class _MissingSpecialTokenTokenizer(_TemplateTokenizer):
     def convert_tokens_to_ids(self, token):
         return None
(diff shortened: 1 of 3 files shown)


wuxibin89:
Prefer explicitly specify `continuous_token.model_family` if auto resolve is not expected. Fixed in https://github.com/verl-project/verl/pull/7640

PR #7640, opened 2026-08-31
[rollout] feat: explicit Continuous Token model_family override with an actionable construction error

ruiling-smartbear:
### What does this PR do?

For #7637, this implements the explicit-override option (option 3 there): the DeepSeek-R1 distills keep Qwen `model_type` values (`qwen2` / `qwen3`) but ship DeepSeek's tokenizer and chat template, in which `<|im_end|>` is renamed away, so since #6804 `QwenContinuousTokenBuilder.__init__` raises `Tokenizer does not define required token '<|im_end|>'` in `AgentLoopBase.__init__` before any rollout starts — and no configuration can route around it, because #6804 removed `data.continuous_token`.

This restores `data.continuous_token.model_family` only (not the removed `enable` toggle): `auto`, the default, keeps #6804's exact `model_type` inference for everything, and an explicit family selects the builder matching the checkpoint's chat template — the builder has to follow the template's conversation protocol, and when a checkpoint's architecture and template disagree, the user states the protocol. When the auto-inferred builder cannot be constructed because a required token is missing, the error now names the override and the family these checkpoints need instead of only naming the missing token, so the failure is self-explaining. An explicitly requested family that cannot construct still raises the original error.

Alternative to #7638 / #7639, which route these checkpoints automatically instead of asking for configuration; this override composes with either if automatic routing is wanted on top.

### Checklist Before Starting

- [x] Search for similar PRs: https://github.com/verl-project/verl/pulls?q=is%3Apr+continuous+token+model_family

### Test

- `tests/utils/test_continuous_token_on_cpu.py`: the auto-path construction error names `data.continuous_token.model_family` and the needed family; `model_family="deepseek"` selects the DeepSeek builder for a distill-style tokenizer and renders its concatenating template; an explicit family that cannot construct raises the original error unchanged.
- Hydra: `ppo_trainer` composes with the restored key, `data.continuous_token.model_family=deepseek` overrides to the value; the four `_generated_ppo_*.yaml` files are regenerated with `scripts/generate_trainer_config.sh`.
  - the 13-row × 24-case tool-append harness from #7630 / #7636: every row byte-identical to `main` (no successful path changes);
  - the three broken distills under `auto`: still fail, now with the error pointing at the override; with `model_family=deepseek` they construct `DeepSeekContinuousTokenBuilder` and the first tool turn matches the template's full-conversation render (DeepSeek-R1-0528-Qwen3-8B on every turn; the older distills' later turns keep the known `is_output_first` gap tracked in #7636);
  - Qwen3-8B, DeepSeek-V3.1 and DeepSeek-R1-Distill-Llama-8B select the same builders as `main` under `auto`.

  Script and raw output: https://github.com/ruiling-smartbear/verl-experiments/blob/main/verl_family_config_check_results.txt

### API and Usage Example

```yaml
data:
  continuous_token:
    model_family: deepseek  # e.g. for DeepSeek-R1-Distill-Qwen-* / DeepSeek-R1-0528-Qwen3-8B
```

Default `auto` is unchanged behavior for every checkpoint that works today.

### Design & Code Changes

- `verl/trainer/config/data/legacy_data.yaml` (+ regenerated `_generated_ppo_*.yaml`): `data.continuous_token.model_family`, default `auto`.
- `agent_loop.py`: reads the key and passes it to `create_continuous_token_builder` (whose `model_family` parameter already existed).
- `continuous_token.py`: `MissingRequiredTokenError` (a `ValueError` subclass) for the token-absent case.
- `continuous_token_wiring.py`: under `auto`, a construction failure for a missing required token is re-raised with the override guidance.

- [x] Docs: the config key is documented in `legacy_data.yaml`.

### DIFF
--- a/tests/utils/test_continuous_token_on_cpu.py
+++ b/tests/utils/test_continuous_token_on_cpu.py
@@ -276,6 +276,26 @@ class _DeepSeekBoundaryTokenizer(_TemplateTokenizer):
         return rendered
 
 
+class _DistillTokenizer(_DeepSeekBoundaryTokenizer):
+    """DeepSeek-R1 distill of Qwen: the root config keeps a Qwen ``model_type`` but the
+    checkpoint ships DeepSeek's tokenizer and chat template, in which ``<|im_end|>``
+    does not exist (renamed to ``<｜end▁of▁sentence｜>``)."""
+
+    name_or_path = "deepseek-ai/DeepSeek-R1-Distill-Qwen-7B"
+    unk_token_id = None
+
+    def __init__(self):
+        super().__init__()
+        self.assistant_id = 2
+
+    def convert_tokens_to_ids(self, token):
+        if token == "<｜end▁of▁sentence｜>":
+            return self.eos_id
+        if token == "<｜Assistant｜>":
+            return self.assistant_id
+        return None
+
+
 class _MissingSpecialTokenTokenizer(_TemplateTokenizer):
     def convert_tokens_to_ids(self, token):
         return None
@@ -468,6 +488,48 @@ def test_unknown_model_with_non_multimodal_processor_uses_default_text_builder(c
     assert isinstance(builder, ContinuousTokenBuilder)
     assert "unknown_text_model" in caplog.text
     assert "default" in caplog.text
+
+
+def test_auto_family_construction_error_points_at_the_model_family_override():
+    with pytest.raises(ValueError) as excinfo:
+        create_continuous_token_builder(_DistillTokenizer(), hf_model_type="qwen2")
+
+    message = str(excinfo.value)
+    assert "required token '<|im_end|>'" in message
+    assert "data.continuous_token.model_family" in message
+    assert "deepseek" in message
+
+
+def test_explicit_model_family_override_selects_the_matching_builder():
+    tokenizer = _DistillTokenizer()
+    builder = create_continuous_token_builder(tokenizer, model_family="deepseek", hf_model_type="qwen2")
+
+    assert type(builder) is DeepSeekContinuousTokenBuilder
+
+    # The selected builder renders the distill's DeepSeek-style template: tool appends
+    # splice the synthetic call's arguments in by string concatenation.
+    previous_messages = [
+        {"role": "user", "content": "question"},
+        {
+            "role": "assistant",
+            "content": "",
+            "tool_calls": [
+                {"id": "call_0", "type": "function", "function": {"name": "lookup", "arguments": {"q": "x"}}}
+            ],
+        },
+    ]
+    updated_messages = previous_messages + [{"role": "tool", "content": "answer", "tool_call_id": "call_0"}]
+    incremental = builder.tokenize_non_assistant_incremental_messages(previous_messages, updated_messages)
+    assert incremental == [ord(char) for char in "<tool_output_begin>answer<tool_output_end>"]
+
+
+def test_explicit_family_that_cannot_construct_raises_the_original_error():
+    with pytest.raises(ValueError) as excinfo:
+        create_continuous_token_builder(_DistillTokenizer(), model_family="qwen")
+
+    message = str(excinfo.value)
+    assert "required token '<|im_end|>'" in message
+    assert "model_family" not in message
 
 
 def test_default_builder_creation_forwards_kwargs():
--- a/tests/utils/test_continuous_token_on_cpu.py
+++ b/tests/utils/test_continuous_token_on_cpu.py
@@ -515,10 +515,18 @@ def test_auto_family_construction_error_does_not_assume_deepseek():
     assert "GLMContinuousTokenBuilder" in message
     assert "required token '<|observation|>'" in message
     assert "data.continuous_token.model_family" in message
-    assert "deepseek" not in message
-    # Suggestions are the families that actually construct with this tokenizer, with
-    # the catch-all last rather than presented as a match.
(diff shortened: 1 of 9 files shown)


wuxibin89 (dismissed):

ruiling-smartbear:
Pushed `db247fd` — the construction error no longer assumes the failing checkpoint is a DeepSeek-R1 distill.

The branch I added is not distill-specific. Every model-specific builder validates its special tokens at construction, so six builder classes can reach it:

| builder | required token |
| --- | --- |
| `QwenContinuousTokenBuilder` | `<\|im_end\|>` |
| `MiniMaxContinuousTokenBuilder` | `[e~[` |
| `GLMContinuousTokenBuilder` | `<\|observation\|>`, `<\|user\|>` |
| `Gemma4ContinuousTokenBuilder` | `<\|tool_response>` |
| `DeepSeekContinuousTokenBuilder` | `<｜end▁of▁sentence｜>` |
| `MiniMaxVLContinuousTokenBuilder` | `<end_of_sentence>`, `<beginning_of_sentence>` |

A GLM checkpoint missing `<|observation|>` was told to set `model_family=deepseek`, which needs a token that tokenizer does not have either — so the advice buys the user one more failed run before they are back where they started.

The suggestion is now derived from the tokenizer rather than assumed, in three tiers:

1. an exact special-token key when the tokenizer identifies its own family, reusing `DeepSeekContinuousTokenBuilder`'s own constants so the key stays aligned with what that builder validates;
2. otherwise the families whose builder actually constructs with this tokenizer, one name per builder class, with `default` last because it accepts any tokenizer and is the generic answer rather than evidence of a match;
3. otherwise the full supported list.

Construction is a special-token lookup with no side effects, so probing every family on the error path costs nothing and, unlike a fixed string, can never name a family that fails the same way.

What the two cases print now:

```
QwenContinuousTokenBuilder, inferred from config.json model_type='qwen2', cannot be
constructed: Tokenizer does not define required token '<|im_end|>'. This checkpoint
pairs a registered architecture with a tokenizer and chat template that builder does
not support. This tokenizer's special tokens exactly match family 'deepseek'; set
data.continuous_token.model_family=deepseek to select DeepSeekContinuousTokenBuilder
explicitly.

GLMContinuousTokenBuilder, inferred from config.json model_type='glm4_moe', cannot be
constructed: Tokenizer does not define required token '<|observation|>'. This checkpoint
pairs a registered architecture with a tokenizer and chat template that builder does not
support. Set data.continuous_token.model_family to the family whose builder matches this
checkpoint's chat template; these construct with this tokenizer: gptoss, default.
```

Routing is unchanged — this decides only what the error says. `auto` still resolves purely from `config.json` `model_type`, and an explicitly set family still raises its own error untouched, so the default path is still byte-identical to `main`.

Tests: added the GLM case asserting the message never mentions deepseek, and tightened the distill case to assert the family is named because the tokens matched rather than by default. `tests/utils/test_continuous_token_on_cpu.py` is 128 passed, and both new assertions fail against the previous revision of this PR. `ruff check` and `ruff format` are clean under the pinned v0.12.2.

wuxibin89:
cc @gxlvera

ruiling-smartbear:
Correcting my previous comment. The fallback it described — listing the families whose builder can be constructed with this tokenizer — is not evidence of a match, and I framed it as if it were.

A builder can only fail construction if it validates special tokens. The families whose builder never calls `require_token_id` therefore accept every tokenizer and always survive the probe. That is why the GLM example in that comment printed `gptoss, default`: `GptOssContinuousTokenBuilder` validates nothing, so it says nothing about whether it fits a GLM checkpoint. I replaced a wrong fixed suggestion with an unfounded one presented as a shortlist, which is the same defect in a different shape.

`9255af9` drops the probe. A family is named only when the tokenizer's special tokens identify one; otherwise the error states the criterion rather than guessing:

```
GLMContinuousTokenBuilder, inferred from config.json model_type='glm4_moe', cannot be
constructed: Tokenizer does not define required token '<|observation|>'. This checkpoint
pairs a registered architecture with a tokenizer and chat template that builder does not
support. This tokenizer's special tokens do not identify a family. Set
data.continuous_token.model_family to the family whose builder implements this checkpoint's
chat template -- the conversation protocol, not the architecture -- or to 'default' for
plain concatenation. Families available on this path: default, qwen, qwen25, qwen3, qwen35,
minimax, minimaxm2, minimaxm25, minimaxm27, glm47, glm5, gemma4, gptoss, deepseek, deepseekv4.
```

The distill case is unchanged and still names the family, because there the tokens do match:

```
... This tokenizer's special tokens exactly match family 'deepseek'; set
data.continuous_token.model_family=deepseek to select DeepSeekContinuousTokenBuilder explicitly.
```

The menu is filtered to the families usable on the current path — a VL builder raises without a processor and a text builder raises with one, so listing both halves would offer options that cannot be selected here.

The special-token table stays extensible but deliberately covers only what it can prove. Deriving it from each builder's required tokens would generalise it to the other families; that needs those tokens declared on the classes rather than passed inline to `require_token_id`, so I left it out of this PR rather than reshaping six builders here. Happy to do it as a follow-up if you want the coverage.

`tests/utils/test_continuous_token_on_cpu.py`: 128 passed. The GLM test now asserts no family is recommended (`model_family=deepseek` and `match family` both absent), that the criterion is stated, and that a text run is not offered VL families. `ruff check` and `ruff format` clean under the pinned v0.12.2. Routing is still untouched — `auto` resolves purely from `config.json` `model_type`, and the default path is byte-identical to `main`.

PR #7700, opened 2026-09-02
[BREAKING][data] feat: integrate ContinuousTokenBuilder into MultiTurnSFTDataset

qfhu0ng:
### What does this PR do?

Integrates Continuous Token (CT) into `MultiTurnSFTDataset`. The dataset builds the initial prompt once, reconstructs each gold assistant response, and appends later turns. Previously committed tokens remain authoritative, including reasoning that a later full-conversation render could remove. `MergeResult` and `align_response_metadata()` keep supervision aligned with boundary edits.

Native MiniMax-VL batch verification also exposed two bugs already reproducible on main in `dataset_utils.py` and `model.py`. Their fixes total 12 changed lines; those two files and a dedicated batch regression file account for the increase from 15 to 18 files. The shared CT diff remains 30 changed lines.

This continues #7322 from `5c375889`, preserving @yph22's commits and authorship. It is not a competing implementation. Follow-up changes address structured images, SFT reconstruction, model support boundaries, and regression coverage.

### CT and SFT responsibilities

The dataset uses the public initial-build, assistant-tokenization-and-merge, non-assistant-merge, and metadata-alignment operations. `merge_assistant_with_tokenization()` delegates gold-message reconstruction to `sft_continuous_token.py`, then calls the existing token-ID merge. Rollout callers continue passing generated token IDs. The internal SFT adapter reuses package-private protocol helpers; this is not a claim that every internal dependency is a public API.

Kimi-VL's shared class is restored to main. Unsupported tool schemas, assistant calls and tool responses are rejected at SFT row validation and direct assistant-reconstruction entry points, including supplied history and template-kwargs schemas.

DeepSeek-V4's shared implementation is restored to main. After the ordinary factory validates the configuration, Dataset substitutes a small SFT subclass that retains committed tokens during appends. It inherits initial rendering and preserves explicit `drop_thinking` options. DeepSeek-V4's registry and rollout behavior remain unchanged; the shared factory also remains unchanged.

MiniMax-Text-01 and DeepSeek-VL2 retain runtime registrations, but this SFT dataset explicitly refuses those families. DeepSeek-VL2's shared runtime class is restored to main. Default direct SFT assistant reconstruction also rejects DeepSeek-VL2 before tokenizer rendering, preventing fallback to the generic DeepSeek reconstructor. Its SFT support scope is reduced because native-processor experiments found additional Dataset and batching integration necessary; successful single-sample token/image comparisons do not establish native batch compatibility.

MiniMax-VL's schema normalization, native function-response rendering and omission of repeated tool declarations live in an SFT subclass. Dataset uses the same post-factory adaptation entry as V4. Exact built-in MiniMax-VL and V4 builders receive SFT subclasses, with effective template options, multimodal options and allowed append roles preserved as applicable. This entails one additional lightweight construction per cached setting. Instances of the SFT subclasses, including their derived classes, are retained. Other custom Python subclasses of those two runtime builders receive an explicit error at the adaptation entry, preventing replacement from discarding their overrides or additional state. This boundary is specific to post-factory SFT adaptation; direct assistant reconstruction retains its existing dispatch and input protection, without claiming to validate an independently constructed entire conversation.

### Remaining shared CT changes

The required `merge_assistant_with_tokenization()` public entry and its documentation are separate from the changes to existing runtime methods.

The only additional behavioral edit to existing shared CT is one argument in MiniMax-VL scaffold detection: `add_generation_prompt=False` becomes `True`. Generation must be enabled when extracting the generation scaffold. Completely restoring this line causes construction to fail for an otherwise working custom template before post-factory SFT adaptation can run. This shared constructor correction also applies to MiniMax-VL rollout builders. Four complete Parquet examples verify the retained line against the custom template's tokens and assistant supervision spans. The original unconditional processor template remains compatible.

The image-discovery helpers inside `continuous_token.py` and shared VL template rendering are restored to main. Standard Dataset image inputs are normalized before CT assembly. Message-container copying remains local to direct SFT rendering, preserving image payload identity. Seven real Qwen3-VL Parquet input forms give identical tensors after this relocation; removing the existing Dataset normalization reproduces missing multimodal output.

### Existing main batching bugs found during MiniMax verification

These two failures reproduce using the native processor outputs and main's batching utilities, without the new CT SFT Dataset implementation:

| Utility | Existing failure and repair | Diff and callers |
| --- | --- | --- |
| `SFTTensorCollator` in `dataset_utils.py` | A text-first batch lacks the first sample's `multi_modal_inputs` key. Determine the value type from a sample that contains the field; retain optional slots and required Tensor-field errors. | +2/-1 = 3 changed lines. Current in-repository callers are the two SFT trainers; PPO uses its separate RL dataset collator. |
| `extract_multi_modal_inputs` in `model.py` | Native MiniMax-VL supplies per-sample crop Tensor lists, which fail in `torch.cat`. Flatten these lists by sample/image/crop order, retaining each crop Tensor. | +7/-2 = 9 changed lines. Called by shared model engines used by both SFT and RL. |

The extractor change therefore reaches shared RL/SFT model-input preparation. Existing ordinary Tensor concatenation, variable-length audio and MiniCPM `image_bound` branches retain their behavior and precedence. Native crop assembly is checked against the fixed official consumer code; this does not establish RL training or pretrained-model forward correctness.

The two utility fixes and their batch regression tests can be split into a separate PR if maintainers prefer. The native MiniMax batching result reported here depends on these fixes.

### Dataset and protocol details

- OpenAI `image_url` blocks pass through `process_image()` before CT assembly. Resize options are preserved. Arrow-added null fields are removed at the message/block level without removing nulls inside tool arguments.
- Multimodal processor `input_ids`, `attention_mask`, and `mm_token_type_ids` do not replace the assembled CT trajectory.
- Qwen preserves embedded reasoning and literal nested think tags. It uses a normalized `reasoning_content` field only when the actual template render responds to that field; otherwise it retains the original content. The comparison receives the same tools and template options as reconstruction and does not run the image processor. Explicitly disabled thinking retains its existing normalization.
- MiniMax assistant formatting preserves both ordinary content and structured calls. The SFT default selects the checkpoint tokenizer template, as used by its function-call example; explicit template overrides remain explicit. SFT rejects templates observed to omit entire tool declarations/responses, and rejects initial tool responses without the preceding assistant call. Render comparisons detect complete omission, not arbitrary custom-template correctness.
- MiniMax's native function template inserts names and response text verbatim. Token fidelity, rather than JSON validity, governs this path.
- Both trainer entry points resolve the effective model family into a data-config copy before constructing `MultiTurnSFTDataset` subclasses. They preserve the historical constructor keyword set, including explicit old-signature custom subclasses and constructors that select a builder eagerly. Unrelated custom datasets receive their original config. Explicit family selection, VL promotion, unsupported-family refusals, readonly/struct configuration and parent interpolation are covered. The direct Dataset API retains its optional `hf_model_type` argument; the factories convey effective family through config rather than raw model-type constructor metadata.
- DeepSeek-V4 gold task continuations use the immediately preceding message's task when reconstructing the suffix. A later reminder without a task restores the ordinary thinking scaffold. Reconstruction still encodes only the suffix once.
- Gemma4 preserves serialized embedded thought whitespace exactly; standalone reasoning retains its template separator and explicit reasoning priority.

### Breaking changes and compatibility

- Remove `data.ignore_input_ids_mismatch`. External Hydra overrides must delete `data.ignore_input_ids_mismatch=...`; otherwise composition raises `Could not override`. There is no replacement bypass. Four examples and the SFT engine E2E script remove the obsolete option.
- MiniMax-Text-01 and DeepSeek-VL2 SFT raise explicit unsupported-family errors. Their runtime registry entries remain unchanged. The Text-01 error directs MiniMax-M2 users to `minimaxm2`.
- Kimi-VL SFT rejects unsupported structured tools before template rendering. Its shared runtime retains main's acceptance behavior.
- MiniMax SFT explicitly rejects the incompatible structured-input/template combinations described above.
- Python builder extensions: post-factory SFT adaptation explicitly rejects custom MiniMax-VL and DeepSeek-V4 runtime subclasses that have not inherited their SFT adaptation. Already adapted SFT subclasses retain their overrides and state. Configured chat-template overrides on supported builders continue through the existing template validation. Direct assistant reconstruction keeps its existing dispatch; this new subclass restriction applies only at the adaptation entry.

### Verification

```bash
VERL_REQUIRE_LOCAL_MODELS=1 HF_HUB_OFFLINE=1 TRANSFORMERS_OFFLINE=1 PYTHONPATH=. python -m pytest \
  tests/utils/test_continuous_token_on_cpu.py \
  tests/utils/test_sft_continuous_token_on_cpu.py \
  tests/utils/dataset/test_multiturn_sft_dataset_on_cpu.py \
  tests/utils/dataset/test_sft_tensor_collator_on_cpu.py \
  tests/utils/test_audio_input_support_on_cpu.py \
  tests/utils/test_model_on_cpu.py -q -rs
```

Final local candidate result: **453 passed, no skips** across the six related CPU files above. The unchanged main CT test file also passes **124 tests, no skips**, against this candidate. Pre-commit passed for all 18 whole-PR changed files, including configuration generation; `git diff --check` is clean. These are local checks, not remote CI results or a run of every repository CPU suite.

Regression tests cover explicit refusal of unadapted runtime subclasses through Dataset and the adapter, preservation of custom SFT overrides and state across repeated adaptation, and explicit DeepSeek-VL2 handler registration precedence.

The V4 immediate-predecessor task rule was checked against the cached official encoder, including its context-based append API. All six current-product cases match official assistant suffix token IDs and aligned loss masks: immediate task, earlier task followed by a task-free reminder, and no task, each with thinking on/off and nonempty reasoning. Existing task/reminder prompt-prefix differences remain; this suffix comparison does not establish full-history equivalence for those cases.

Targeted rollback experiments kept those tests unchanged and modified only product code in separate complete source trees:

| Local control | Result |
| --- | --- |
| Current implementation, selected regression and supported-path controls | 16 passed |
| Remove the custom-subclass protection | 4 expected failures: two missing-refusal assertions and two Dataset calls reaching sample rendering |
| Move DeepSeek-VL2 refusal ahead of registered-handler lookup | 2 expected failures: explicit base/subclass handler registrations blocked |

All three arms had zero collection/setup errors and zero skips. Four additional public reconstruction/merge calls reached their explicitly registered handlers with the current implementation and were blocked by the premature-refusal variant. These historical controls establish that the tests catch those specific regressions; they are separate from the final passing suite. Fresh repair controls run the new tests against old or reverted product code: V4/Gemma produce 19 expected failures and 12 control passes; old Dataset factories produce 18 legacy-signature TypeErrors and 28 control passes; reverting the optional-key collator change produces one expected failure and three control passes; reverting crop-list extraction produces eight expected failures and nine control passes. Every arm has zero setup/collection errors and zero skips. Real tokenizer/Parquet follow-ups pass 14 V4 cases, two direct plus two Dataset Gemma cases, and six factory cases with exact token/mask/position comparisons.

Resource variables are `VERL_TEST_QWEN25_MODEL`, `VERL_TEST_QWEN3_MODEL`, `VERL_TEST_QWEN35_MODEL`, `VERL_TEST_GPT_OSS_MODEL`, and `VERL_TEST_QWEN3_VL_MODEL`. The CPU workflow's strict mode requires these existing local tokenizer/processor artifacts; local non-strict runs may skip missing resources. An absent-artifact reverse control verifies the skip/fail distinction. This is not a remote CI result.

Additional local experiments use complete source trees in separate processes, rather than mixing classes from different revisions. Coverage includes official Kimi runtime/SFT comparisons, official V4 context-based incremental and assistant encoding, MiniMax checkpoint-template token differentials and native image processor assembly, Qwen real-image Dataset assembly, and targeted mutations that must fail for dropped appends, lost template options and discarded resize parameters. Real Qwen and MiniMax counterexamples also exercise reasoning and content-plus-tool-call preservation.

Hydra configuration composition succeeds without the obsolete key and rejects it when supplied. Five affected scripts pass `bash -n`. A temporary Git index includes sparse-excluded tracked paths in the full-tree obsolete-key scan.

### Scope and limits

Against `23af6a7a2e8d6efeeb2adbe5d1689c7a24f503a3`: **18 files, +4022/-201 (4223 changed lines)**. Shared CT accounts for **+26/-4 (30)**: the required new API is 21 added lines including its separator, API documentation is +4/-3, and the remaining existing-method change is one MiniMax argument (+1/-1). Wiring and `deepseek.py` are byte-identical to main.

The SFT adapter is 931 added lines; Dataset/trainer/config/collator accounts for 536 changed lines, the multimodal extractor 9, tests/E2E 2710, and examples/workflow 7. The collator and extractor add 12 changed lines outside shared CT to repair the native batch failures verified below. Production/config totals 1506 changed lines.

The local placement experiments compare 99 MiniMax Dataset records, seven two-image/two-turn Qwen3-VL input forms, and five native MiniMax processor/Dataset cases: no image, one image, two images, reversed images, and an image appended in a later turn. For native MiniMax, token IDs through the last assistant EOS, independent assistant loss spans and pixel tensors match the checkpoint template/processor. The template's trailing newline is excluded because CT commits the final assistant through EOS and inserts the inter-turn newline when another turn follows. The published anyres processor configuration is used; no image preprocessing parameters are guessed.

The real MiniMax-VL image processor was exercised on CPU in the five cases above. A follow-up exposed mixed-text/image collation and native crop-list extraction errors; main's native processor outputs reproduce the same shared-utility failures. This candidate repairs both. A fresh CPU experiment passes 36 two-sample cases (structured/images-column inputs, every 0/1/2-image count pairing, packed and padded engine paths) plus 12 single-sample controls. It runs the real processor, Dataset, collator, extractor and FSDP input-preparation function, then the fixed-revision official CLIP crop loop and MiniMax feature grouping/packing/scatter code. Crop order and image-feature counts are checked, and every sample's assembled embeddings equal its single-sample reference exactly. Learned components use tiny deterministic substitutes; a sentinel stops at the language-model `inputs_embeds` entry. This validates structural input assembly, not pretrained attention, logits, loss, backward, GPU parallelism or convergence. No pretrained weights or GPU training were used, and these checks do not establish every resolution or batching mode. Directly feeding a NumPy image into private Dataset assembly, bypassing normal Parquet normalization, regains main's ambiguous-truth-value error; that is outside the input paths established by these tests. Video/audio normalization is outside this change. Over-length multimodal samples should use `truncation=error`; arbitrary image-removing truncation is not validated. The Qwen preserve-thinking template variant uses Qwen3.5 artifacts, not a Qwen3.6 checkpoint.

### Attribution and AI assistance

Original CT SFT integration: @yph22. Follow-up investigation, implementation and validation: @qfhu0ng. OpenAI Codex and Claude assisted with investigation, code, tests and drafting. The human submitter is responsible for reviewing and validating the contribution.


verl/trainer/config/sft_trainer_engine.yaml
```diff
@@ -39,12 +39,13 @@ data:
...
-  # MultiTurnSFTDataset apply_chat_template to each turn separately and concat `input_ids`
-  # as a whole sequence, which may not equal to apply_chat_template to whole messages at once. 
-  # For example, Qwen Thinking series models add <think></think> tags to last turn, please check
-  # your tokenizer chat template settings.
-  # Set to True to ignore input_ids mismatch and use the concatenated input_ids as the final input_ids.
-  ignore_input_ids_mismatch: False
+  # MultiTurnSFTDataset uses append-only Continuous Token assembly, matching RL TITO.
+  # Historical tokens are never re-rendered even if the chat template would rewrite them.
+  # Auto resolves the builder family from the model config's model_type.
+  continuous_token_model_family: auto
```
gxlvera: is this really necessary?  for RL dataset, we have removed the continuous_token args and model family is inferred from model architecture. I think it should also apply to MultiturnSFTdataset. Can you re-check that? I think It should simply reuse the the model family wiring logic in `continuous_token_wiring.py`.

verl/trainer/config/sft_trainer_engine.yaml
```diff
@@ -39,12 +39,13 @@ data:
   apply_chat_template_kwargs: {}
   num_workers: 8
 
-  # MultiTurnSFTDataset apply_chat_template to each turn separately and concat `input_ids`
-  # as a whole sequence, which may not equal to apply_chat_template to whole messages at once. 
-  # For example, Qwen Thinking series models add <think></think> tags to last turn, please check
-  # your tokenizer chat template settings.
-  # Set to True to ignore input_ids mismatch and use the concatenated input_ids as the final input_ids.
-  ignore_input_ids_mismatch: False
```
gxlvera: why is this deleted?

verl/utils/dataset/multiturn_sft_dataset.py
```diff
@@ -59,15 +61,28 @@ def print_assembled_message(tokenizer, message_list, input_ids, loss_mask, attn_
...
+            message_list,
+            add_generation_prompt=False,
+            tokenize=False,
+            tools=tools,
+        )
+    except Exception as exc:
+        # Some registered CT protocols (notably DeepSeek-V4) use a native
+        # encoder and intentionally have no tokenizer chat_template. Debug
+        # output must not make an otherwise valid dataset item fail.
+        tokenized = f"<whole-conversation chat render unavailable: {exc}>"
```
gxlvera: is there a more elegant way to deal with such issues 🤔️

gxlvera:
It seems that there are more changes to CT itself than expected. I will take a closer look.

gxlvera:
Why did you made many changes to model specific behavior? Isn't the model-specific implementation of current ContinuousToken not enough?

gxlvera:
Please use the public API of ContinuousToken as possible. The private hooks are intended to be used by subclasses instead of upstream caller (SFTDataset)

qfhu0ng:
Thanks for pointing this out. The existing model-specific ContinuousToken implementations remain sufficient for rollout, where the assistant token IDs have already been generated. The missing step for multi-turn SFT is different: it starts from a structured gold assistant message, which must be rendered and tokenized once into the exact continuation that the corresponding rollout protocol would produce before reusing `merge_assistant_tokens()`.

I have narrowed the implementation accordingly:

- `MultiTurnSFTDataset` now calls the public `merge_assistant_with_tokenization()` API and no longer accesses CT private hooks.
- The rollout-facing `merge_assistant_tokens(token_ids)` path is unchanged.
- SFT-only reconstruction is isolated in `sft_continuous_token.py`, with model aliases sharing 11 protocol handlers through builder MRO.
- I removed the additional Gemma4 shared-runtime rewrite after complete main/branch comparisons showed identical observable token IDs, aligned masks, and logprobs.

For the retained SFT-specific handlers, I also ran controlled omission tests with the same builder, structured message, and real tokenizer/processor oracle. Across 14 cases covering all 11 handlers, forcing the generic reconstruction either fails or produces different token IDs, while the specialized path matches the model-owned template or native encoder. This includes direct VL-builder checks for Qwen3-VL, GLM-4V, and Gemma4-VL.

The current CT/SFT CPU unit suite passes with 201 tests.

Issue #7476, opened 2026-08-19
ulysses_sequence_parallel_size is a silent no-op on Qwen3.5: per-rank activations stay full length


khazic:
### What happens

`engine.ulysses_sequence_parallel_size=8` is accepted, appears in the resolved
config, produces no warning, and does not shard the sequence. Per-rank activation
tensors stay at the full sequence length, so sequence parallelism buys no memory
at all.

Found while trying to fit a 27.8B Qwen3.5 (64 layers, hidden 5120) at a 32,768
context on 80 GiB cards.

### Evidence

A CUDA allocator snapshot taken at the OOM, aggregated by block size:

```
live allocated: 74.39 GiB
  45.82 GiB from 340 MiB blocks
   9.38 GiB from 120 MiB blocks
   6.64 GiB from 200 MiB blocks
```

With hidden = 5120 and bf16, a `[T, hidden]` activation is 320 MiB at T = 32,768
(the allocator rounds to 340 MiB) and 40 MiB at T = 4,096, which is what sp=8
should have produced. The 45.82 GiB sits in ~138 such blocks, i.e. roughly two per
layer for 64 layers: the gradient-checkpoint boundaries, at full length.

Supporting measurement: cutting the per-rank token budget 8x moved peak memory
only from 78.7 GiB to 74.2 GiB. If the sequence were being sharded, the same cut
would have been visible.

### A related trap in the same area

`data.max_token_len_per_gpu` is multiplied by the SP size:

```python
# verl/workers/engine/utils.py:111
max_token_len = max_token_len_per_gpu * sp_size
```

So with sp=8 the value has to be set to one eighth of what its name suggests. Set
to 32,768 with sp=8, verl packs 262,144-token micro-batches. That is easy to get
wrong and, combined with the above, makes an ineffective SP setup look like a
memory problem rather than a configuration one.

### What would help

Either shard the sequence for this model or fail loudly when the setting cannot
be honoured. A silent no-op is expensive to diagnose: the only reason it was
caught here is the allocator snapshot, after several full-cluster launches spent
on the assumption that SP was active.

A one-line log of the effective per-rank sequence length at the first micro-batch
would have made it obvious immediately.

### Versions

verl 0.9.0 and current main, transformers 5.15.0, flash-linear-attention 0.5.0,
torch 2.12, FSDP2 engine, `pad_mode=no_padding`, `use_dynamic_bsz=True`,
`use_remove_padding=True`, sm80.

[...]

PR #7711, opened 2026-09-03
[fsdp] fix: surface the memory a differing reduce_dtype costs under gradient accumulation

alanhuangyoo:
### What does this PR do?

Makes visible a memory cost that verl's default FSDP2 mixed-precision policy pays silently.

FSDP2's `FSDPParam.to_accumulated_grad_if_needed` only returns early when the gradient is
already in `reduce_dtype`:

```python
if (self.reduce_dtype is None
        or self._unsharded_param.grad is None
        or self._unsharded_param.grad.dtype == self.reduce_dtype):
    return
unsharded_grad = self._unsharded_param.grad
self._unsharded_param.grad = None
self.unsharded_accumulated_grad = unsharded_grad.to(self.reduce_dtype)
```

verl defaults to `mp_param_dtype: bf16` with `mp_reduce_dtype: fp32`
(`verl/workers/config/engine.py`), so the gradient is bf16, the reduce dtype is fp32, and that
early return never fires. Every parameter then keeps a **whole-model fp32
`unsharded_accumulated_grad`** alive for the length of a gradient-accumulation window.

| model | retained unsharded fp32 grads, per rank | what a sharded grad would cost at dp=8 |
| --- | ---: | ---: |
| 7B | 26.1 GiB | 3.3 GiB |
| 27.8B | 103.6 GiB | 12.9 GiB |
| 70B | 260.8 GiB | 32.6 GiB |

The part that makes this expensive to diagnose is that the term is proportional to **parameter
count and not to world size**, so adding ranks does not relieve it and cutting the per-GPU token
budget does not move it. That is the opposite of how an activation problem behaves, which is
exactly how it gets misread as one.

### Where this came from

#7476. The reporter spent several full-cluster launches and a CUDA allocator snapshot before
identifying it, initially filing it as "`ulysses_sequence_parallel_size` is a silent no-op"
because peak memory would not respond to the token budget. The 340 MiB blocks in that snapshot
are exactly `[5120, 17408]` fp32 — an MLP gradient, not a `[32768, 5120]` activation. They
withdrew the original claim themselves after finding the real term.

Their own conclusion was that a single line naming the cost would have made it obvious
immediately. That is what this adds.

### The change

A one-time warning when `reduce_dtype` differs from `param_dtype`, plus the same note in the
`mp_reduce_dtype` config docstring:

```
FSDP2 mixed precision has reduce_dtype=torch.float32 with param_dtype=torch.bfloat16. Gradients
are upcast to torch.float32 and held unsharded for the whole gradient-accumulation window,
costing roughly 4 bytes per parameter per rank on top of activations, independent of world size.
Set reduce_dtype to torch.bfloat16 to trade reduction precision for that memory.
```

**Deliberately not changing the default.** An fp32 reduction is the more accurate one and the
tradeoff is the user's to make; this only makes the cost visible when it is paid and names the
knob that trades it away.

### Test

`tests/workers/engine/test_fsdp2_reduce_dtype_warning_on_cpu.py`, CPU-only, following the
`_on_cpu.py` convention so it runs in `cpu_unit_tests.yml`:

- warns for the bf16 / fp32 default, and the message names the two properties that make this
  hard to diagnose (`unsharded`, `independent of world size`)
- silent for bf16/bf16, fp32/fp32 and `reduce_dtype=None` — the cases that hit the early return,
  so there is no retained gradient to warn about
- warns **once** across repeated calls; this sits on a per-module build path, so an
  un-deduplicated warning would repeat per wrapped layer

```
$ pytest tests/workers/engine/test_fsdp2_reduce_dtype_warning_on_cpu.py
5 passed

$ ruff check / ruff format --check      # 0.12.2, as pinned in .pre-commit-config.yaml
All checks passed! / already formatted

$ python3 tests/special_sanity/check_license.py --directories .
exit 0
```

- [x] Apply [pre-commit checks](https://github.com/volcengine/verl/blob/main/CONTRIBUTING.md#code-linting-and-formatting).
- [x] Add / update the documentation — the `mp_reduce_dtype` docstring.
- [x] Add unit tests.


alanhuangyoo:
Closing this myself to keep the queue honest.

What it adds is a docstring paragraph and a one-time warning about `mp_reduce_dtype` differing from `mp_param_dtype` — FSDP2 upcasts each gradient and holds it unsharded for the whole gradient-accumulation window, roughly 4 bytes per parameter per rank, and that does not shrink with world size. The measurement stands and the behaviour is real, but the PR changes no behaviour, and I would rather not spend review attention on a documentation improvement while three actual bugs of mine are waiting in the same queue.

Happy to reopen if someone wants the warning; nothing here has gone stale.

PR #7756, opened 2026-09-05
[misc] fix: extract_system_prompt_and_generation drops the prompt when diff==0 (#6477)

### LINKED ISSUE #6477: [Bug] `chat_template.py` system_prompt extraction returns empty list when both tokenized sequences have equal length (`[:-0]` bug)

amanyagami:
## Motivation

`extract_system_prompt_and_generation` derives the system-prompt prefix as:

```python
system_prompt = token1[: -(len(token2) - len(token1))]
```

where `token1` and `token2` are the renders of one and two empty-content probe
turns, so the marginal length is one turn's own token count.

When a chat template's rendering does not grow with the extra probe turn, that
marginal count is zero — and `x[:-0]` is `x[:0]`, the empty list, not "keep
everything". The whole system prompt is silently dropped.

`MultiTurnSFTDataset` uses that prefix to strip the system preamble from each
turn:

```python
# verl/utils/dataset/multiturn_sft_dataset.py:231-232
input_ids = input_ids[len(self.system_prompt):]
attention_mask = attention_mask[len(self.system_prompt):]
```

With an empty prefix the strip becomes `[0:]` and removes nothing, so every turn
after the first keeps its own copy of the system preamble and the assembled
sequence diverges from `apply_chat_template` of the equivalent full conversation.
It fails silently — no exception, just wrong training data.

## Change

Compute the marginal count explicitly and slice from the front, which behaves
correctly at zero:

```python
diff = len(token2) - len(token1)
system_prompt = token1[: len(token1) - diff]
```

## Relationship to #7592

#7592 patched this same expression in both `extract_system_prompt_and_generation`
and `initialize_system_prompt`. @wuxibin89 closed it because
`initialize_system_prompt` is no longer used — which is correct: on current `main`
it has no callers outside `tests/utils/test_turn_separator_on_cpu.py`.

This PR is narrowed to only the live function. `extract_system_prompt_and_generation`
is still imported and called by `MultiTurnSFTDataset`, and the ContinuousToken
migration (#6779, #6804) did not replace it, so the defect is still reachable on
`main` today. `initialize_system_prompt` is left untouched here; happy to delete it
and its sibling `initialize_turn_separator` in a separate cleanup if that is wanted.

## Test

`tests/utils/test_system_prompt_extraction_on_cpu.py`, CPU-only — a self-contained
Jinja2 template stands in for a real tokenizer, so no GPU, network, or model
download is needed.

- `test_zero_diff_does_not_drop_system_prompt` — a template that renders only the
  first message, so one and two probe turns render identically. Asserts the prefix
  comes back intact rather than empty.
- `test_normal_template_unaffected` — the ordinary `diff > 0` path still returns
  the real system prompt.

```
tests/utils/test_system_prompt_extraction_on_cpu.py
tests/utils/test_turn_separator_on_cpu.py
tests/utils/test_continuous_token_on_cpu.py
134 passed
```

Confirmed non-vacuous: restoring `token1[: -(len(token2) - len(token1))]` fails
`test_zero_diff_does_not_drop_system_prompt`. `ruff check` and
`ruff format --check` are clean at the pinned v0.12.2.

Fixes #6477.

### DIFF
--- a/tests/utils/test_system_prompt_extraction_on_cpu.py
+++ b/tests/utils/test_system_prompt_extraction_on_cpu.py
@@ -1,0 +1,112 @@
+# Copyright 2025 Bytedance Ltd. and/or its affiliates
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""CPU tests for ``extract_system_prompt_and_generation`` (verl issue #6477).
+
+The helper derives the system-prompt prefix as ``token1[: -(len(token2) - len(token1))]``, where
+``token1``/``token2`` are the token ids for one and two probe turns. When the marginal per-turn
+token count (``diff``) is zero -- e.g. a template whose rendering does not grow with extra
+messages -- Python's ``x[:-0]`` is ``x[:0]`` (the empty list), not "keep everything", so the whole
+system prompt was silently dropped.
+
+``MultiTurnSFTDataset`` uses the returned prefix to strip the system preamble from each turn
+(``verl/utils/dataset/multiturn_sft_dataset.py``), so an empty result makes that strip a no-op and
+every turn after the first keeps its own copy of the preamble.
+
+No GPU, network, or model download is required: a self-contained Jinja2 template stands in for a
+real tokenizer.
+"""
+
+import re
+
+from jinja2 import Template
+
+from verl.utils.tokenizer.chat_template import extract_system_prompt_and_generation
+
+_IM_START, _IM_END, _NL = 1, 2, 3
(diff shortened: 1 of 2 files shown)


wuxibin89:
`extract_system_prompt_and_generation` will be deprecated and use Continuous token instead, https://github.com/verl-project/verl/pull/7322

Issue #6851, opened 2026-06-25
GRPO training wtih lora using qwen2.5-7b-vl  tain_batch_size=1,mini_batch_size=1,and micro_mini_batch_size_per_gpu=1


cqray1990:
### System Info

.

### Information

- [ ] The official example scripts
- [ ] My own modified scripts

### Tasks

- [ ] An officially supported task in the `examples` folder (such as GLUE/SQuAD, ...)
- [ ] My own task or dataset (give details below)

### Reproduction

as we don't have enough GPUS
we set tain_batch_size=1,mini_batch_size=1,and micro_mini_batch_size_per_gpu=1,but it raised an error:
like this:
     File "verl/workers/engine_workers.py", line 69, in wrapper
    return func(self, data, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "verl/workers/engine_workers.py", line 648, in compute_log_prob
    output = self.actor.infer_batch(data)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "verl/single_controller/base/decorator.py", line 431, in inner
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "verl/utils/transferqueue_utils.py", line 329, in inner
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "verl/workers/engine_workers.py", line 409, in infer_batch
    output = self.engine.infer_batch(data, loss_function=loss_function)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "verl/workers/engine/base.py", line 148, in infer_batch
    outputs = self.forward_backward_batch(data, loss_function, forward_only=True)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "verl/workers/engine/fsdp/transformer_impl.py", line 650, in forward_backward_batch
    micro_batches, indices = prepare_micro_batches(
                             ^^^^^^^^^^^^^^^^^^^^^^
  File "verl/workers/engine/utils.py", line 77, in prepare_micro_batches
    micro_batches, batch_idx_list = rearrange_micro_batches(
                                    ^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #5689, opened 2026-03-20
[fsdp] fix: avoid NestedTensor jagged dim ambiguity for 3D position_ids

Solus-sano:
### What does this PR do?

Fix an intermittent shape mismatch error during VLM SFT training with FSDP when using `DatasetPadMode.NO_PADDING`. The error manifests as:
```
The size of tensor a (3) must match the size of tensor b (8)
```

The root cause is a PyTorch `NestedTensor` jagged layout ambiguity: when all samples in a micro-batch have the same `seq_len`, `torch.nested.as_nested_tensor` incorrectly selects the first non-batch dimension (e.g., `num_heads=4`) as the jagged dimension instead of the `seq_len` dimension. This causes `.values()` to return shape `(B*num_heads, seq_len)` instead of `(num_heads, total_nnz)`.

### Checklist Before Starting

- [x] Search for similar PRs: [search link](https://github.com/verl-project/verl/pulls?q=is%3Apr+nested+tensor+position_ids)
- [x] Format the PR title as `[{modules}] {type}: {description}`

### Test

Verified with the following test cases:

| Scenario | Original `.values()` | Fixed `unbind+cat` |
|----------|---------------------|-------------------|
| Different seq_len (2 samples) | `(4, 1, 300)` ✅ | `(4, 1, 300)` ✅ |
| Same seq_len (2 samples) | `(8, 1, 100)` ❌ | `(4, 1, 200)` ✅ |
| Same seq_len (8 samples) | `(32, 1, 100)` ❌ | `(4, 1, 800)` ✅ |

<details>
<summary>Reproduction script</summary>

```python
import torch

def fixed_position_ids_rmpad(position_ids_nt):
    return torch.cat([t for t in position_ids_nt.unbind()], dim=-1).unsqueeze(1)

def original_position_ids_rmpad(position_ids_nt):
    return position_ids_nt.values().unsqueeze(1)

# Case 1: Different seq_len - both methods agree
t1 = torch.arange(4*100).reshape(4, 100).float()
t2 = torch.arange(4*200).reshape(4, 200).float()
nt = torch.nested.as_nested_tensor([t1, t2], layout=torch.jagged)
assert original_position_ids_rmpad(nt).shape == fixed_position_ids_rmpad(nt).shape == (4, 1, 300)

# Case 2: Same seq_len - original method fails
t1 = torch.arange(4*100).reshape(4, 100).float()
t2 = torch.arange(4*100, 4*200).reshape(4, 100).float()
[...]

PR #6149, opened 2026-04-26
[trainer] fix: support non-last ragged dim in nested tensor rebuild

xiefan46:
### What does this PR do?

Fixes #6152. Nested_tensor_from_tensor_list (introduced in #6127) hardcodes dim=-1 for torch.cat, assuming the ragged dimension is always the last one. This breaks for tensors like teacher_logprobs (batch, [seq_len], topk) where _ragged_idx=1.

Fix: use ragged_idx to determine the cat dimension, and read actual _ragged_idx from the nested tensor at all call sites instead of assuming dim()-1.

### Checklist Before Starting

- [x] Search for similar PRs. Paste at least one query link here: ...
- [x] Format the PR title as `[{modules}] {type}: {description}` (This will be checked by the CI)
  - `{modules}` include `fsdp`, `megatron`, `veomni`, `sglang`, `vllm`, `vllm_omni`, `rollout`, `trainer`, `ci`, `training_utils`, `recipe`, `hardware`, `deployment`, `ray`, `worker`, `single_controller`, `misc`, `perf`, `model`, `algo`, `env`, `tool`, `ckpt`, `doc`, `data`, `cfg`, `reward`, `fully_async`, `one_step_off`
  - If this PR involves multiple modules, separate them with `,` like `[megatron, fsdp, doc]`
  - `{type}` is in `feat`, `fix`, `refactor`, `chore`, `test`
  - If this PR breaks any API (CLI arguments, config, function signature, etc.), add `[BREAKING]` to the beginning of the title.
  - Example: `[BREAKING][fsdp, megatron] feat: dynamic batching`

### Test

ut

### API and Usage Example

> Demonstrate how the API changes if any, and provide usage example(s) if possible.

```python
# Add code snippet or script demonstrating how to use this
```

### Design & Code Changes

> Demonstrate the high-level design if this PR is complex, and list the specific changes.

### Checklist Before Submitting

> [!IMPORTANT]
> Please check all the following items before requesting a review, otherwise the reviewer might deprioritize this PR for review.

- [x] Read the [Contribute Guide](https://github.com/verl-project/verl/blob/main/CONTRIBUTING.md).
- [x] Apply [pre-commit checks](https://github.com/verl-project/verl/blob/main/CONTRIBUTING.md#code-linting-and-formatting): `pre-commit install && pre-commit run --all-files --show-diff-on-failure --color=always`
- [ ] Add / Update [the documentation](https://github.com/verl-project/verl/tree/main/docs).
[...]

PR #7767, opened 2026-09-05
[training_utils] fix: handle serialized and mispacked 3D position ids

qy0720:
## What does this PR do?

Fixes two inconsistent layouts of 3D jagged `position_ids` handled by `maybe_fix_3d_position_ids()`.

For multimodal position IDs with per-sample shape `[coordinates, sequence_length]`, TensorDict operations can produce two distinct states:

- **Coordinate-packed layout:** equal-length samples created with `torch.nested.as_nested_tensor()` may store values as `[batch_size * coordinates, sequence_length]`, with the coordinate dimension marked as ragged.
- **Serialized sequence-packed layout:** after TensorDict consolidation and `DataProto` serialization, values and offsets may still describe the sequence-ragged layout, while `_ragged_idx` is reset from `2` to `1`. Calling `unbind()` with that inconsistent metadata raises `split_with_sizes`.

The previous workaround unconditionally assigned `_ragged_idx = 2`. Changing only this private metadata can leave the values and offsets inconsistent with the declared ragged dimension.

This PR rebuilds the NestedTensor so its values, offsets, and ragged-dimension metadata agree.

Addresses #6851.

Related: #5689 fixed 3D position ID construction in the SFT collator. This PR repairs layouts at the shared TensorDict indexing boundary, including after a `DataProto` serialization round trip.

## Implementation

`maybe_fix_3d_position_ids()` now:

- Returns immediately for missing, non-nested, non-3D, or already-correct ragged-dimension-2 position IDs.
- Detects serialized mRoPE storage when:
  - the coordinate axis contains 3 or 4 rows; and
  - the final offset equals the sequence-storage dimension.
- Reconstructs detected serialized storage directly with the existing values, offsets, and lengths using `jagged_dim=2`, before any unsafe `unbind()`.
- Rebuilds other internally consistent coordinate-packed layouts from their original samples with the sequence dimension marked as ragged.

No public API, configuration, or caller changes are required. Existing callers continue to invoke `maybe_fix_3d_position_ids()` before indexing.

## Tests

The added CPU tests cover:

- Rebuilding equal-length coordinate-packed position IDs.
- Preserving already-valid variable-length sequence-ragged position IDs.
- Preserving exact sample values through the default `DataProto` serialization path.
- Reordering repaired position IDs through `index_select_tensor_dict()`.

Validated locally with PyTorch `2.11.0+cu130` and TensorDict `0.10.0`:

```bash
pytest -q --tb=short \
  tests/test_protocol_on_cpu.py::test_index_select_tensor_dict_with_3d_jagged_position_ids \
  tests/test_protocol_on_cpu.py::test_maybe_fix_3d_position_ids_preserves_valid_layout \
  tests/test_protocol_on_cpu.py::test_3d_position_ids_after_dataproto_serialization
```

## Checklist

- [x] Added CPU regression coverage.
- [x] Completed local pre-commit checks.
- [x] No documentation changes are required for this internal repair.
- [ ] Requested repository CI in the verl Slack `ci-request` channel.
- [x] Not applicable to the `recipe` submodule.

### DIFF
--- a/verl/utils/tensordict_utils.py
+++ b/verl/utils/tensordict_utils.py
@@ -492,13 +492,7 @@ def index_select_tensor_dict(batch: TensorDict, indices: torch.Tensor | list[int
             if isinstance(tensor, torch.Tensor) and not tensor.is_nested:
                 data_dict[key] = tensor[indices]
             elif isinstance(tensor, torch.Tensor) and tensor.is_nested:
-                try:
-                    tensor_lst = tensor.unbind()  # fast path
-                except RuntimeError:
-                    if tensor.dim() != 3 or getattr(tensor, "_ragged_idx", None) != 2:
-                        raise
-                    lengths = tensor.offsets().diff().tolist()
-                    tensor_lst = torch.split(tensor.values(), lengths, dim=0)
+                tensor_lst = tensor.unbind()  # for performance
                 selected_tensors = [tensor_lst[idx] for idx in indices]
                 data_dict[key] = nested_tensor_from_tensor_list(
                     selected_tensors, ragged_idx=getattr(tensor, "_ragged_idx", tensor.dim() - 1)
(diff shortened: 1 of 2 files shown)


koriyoshi2041 (approved):
Reviewed at e5706a3b. The reconstruction is correctly ordered: it unbinds while the original ragged_idx/storage pairing is still valid, then rebuilds with ragged_idx=2 so values, offsets, and future unbind/index-select operations agree. I independently exercised the equal-length [4, L] case on CPU and verified reconstruction plus [1, 0] reordering preserves both samples; diff hygiene also passes. The already-valid ragged_idx=2 guard keeps the variable-length path unchanged.

qy0720:
@wuxibin89 Hi, could you help take a look at this fix when you have time? This is related to the 3D jagged `position_ids` / TensorDict indexing path touched by #4860, #6127 and #6149. The latest version also adds regression coverage for the DataProto serialization case. Thanks!

koriyoshi2041 (approved):
Reviewed the serialization-layout delta at exact head `a965a838`. The new branch distinguishes the known serialized mRoPE storage shape (`values` coordinate axis 3/4, cumulative sequence length on axis 1) before any unsafe `unbind`, reconstructs it with `jagged_dim=2`, and leaves already-correct ragged-2 tensors unchanged. I independently simulated the corrupted jagged-1 metadata for both 3- and 4-coordinate layouts on CPU: repair restored `_ragged_idx=2`, exact per-sample values, and reversed index selection. The fallback remains the prior behavior for other internally consistent layouts. Approved.

verl/utils/tensordict_utils.py
```diff
@@ -908,11 +908,39 @@ def contiguous(data: TensorDict) -> TensorDict:
 
 
 def maybe_fix_3d_position_ids(data: TensorDict):
```
wuxibin89, line 910: With TransferQueue enabled by default in V1 trainer, I think we no longer consolidate TensorDict, so this fix maybe not needed anymore. cc @0oshowero0

0oshowero0: Yes, we will not do any consolidation or pickle to tensordict in TQ. TensorDict is only used as a input/output format standard to prevent user passing random dict structure without correct batch format.

0oshowero0: Actually, TQ will convert TensorDict to dict soon after receiving the request to avoid it's runtime check overhead, and directly transfer each underlying value in the dict.

https://github.com/Ascend/TransferQueue/blob/d45bfe84c9940ee7650e624dba3d60c5dbf95c62/transfer_queue/storage/managers/simple_storage_manager.py#L257

PR #559, opened 2025-03-12
fix: Add error mechanism for mini-batch/batch size divisibility validation

fyqqyf:
We've observed that many users are unfamiliar with the relationship between `mini_batch_size` and `batch_size` configuration (though this currently doesn't affect program execution).

Specifically, should the parameter `config.actor_rollout_ref.train_batch_size` be divisible by `config.actor_rollout_ref.actor.ppo_mini_batch_size` to avoid potential confusion and ensure configuration validity?

PR #7780, opened 2026-09-07
[rollout, cfg] feat: make weight-transfer cleanup GC configurable

jiefangxuanyan:
### What does this PR do?

This PR makes Python garbage collection after vLLM weight-transfer sender cleanup configurable and adds opt-in diagnostics for evaluating its cost and effect.

The cleanup setting accepts:

- `true` to preserve the existing full collection;
- `false` to skip the explicit collection;
- a non-negative integer to collect a specific Python GC generation.

All defaults preserve the existing behavior. A single `actor_rollout_ref.gc_diagnostics` switch enables structured, per-rank output with the collection point, generation, elapsed and thread CPU time, collected objects, RSS changes, CUDA allocator changes when a device is available, and per-generation collection deltas.

The setting is forwarded through the rollout and checkpoint-engine configuration to both bucketed and LoRA weight senders. This keeps the utility tied to an end-to-end use case instead of introducing an unused GC abstraction.

This collection point runs during the initial weight synchronization before the training loop and during the `update_weights` phase of subsequent steps. The default remains a full collection, so this PR does not change phase timing unless a user selects another setting or enables diagnostics. The downstream [verl-omni diffusion GC PR](https://github.com/verl-project/verl-omni/pull/560) uses this API together with three diffusion-specific GC points and reports their phase-level measurements.

AI assistance from OpenAI Codex was used to implement, test, and draft this change. The human submitter reviewed the resulting changes and test results.

### Checklist Before Starting

- [x] Searched open PRs separately for [`GC`](https://github.com/verl-project/verl/pulls?q=is%3Apr+is%3Aopen+GC) and [`garbage collection`](https://github.com/verl-project/verl/pulls?q=is%3Apr+is%3Aopen+%22garbage+collection%22).
  - [#5549](https://github.com/verl-project/verl/pull/5549) schedules periodic GC in the SFT trainer. This PR instead configures and measures an existing cleanup collection at the weight-transfer lifecycle boundary.
  - No open PR found by these searches implements the same change.
- [x] Formatted the PR title according to the repository convention.

### Test

```bash
pytest \
  tests/utils/test_memory_utils_on_cpu.py \
  tests/utils/test_bucketed_weight_transfer_on_cpu.py \
  tests/workers/config/test_rollout_config_on_cpu.py \
  tests/workers/rollout/test_vllm_weight_sender_config_on_cpu.py \
  tests/workers/test_engine_workers_lora_sync_on_cpu.py
```

- Targeted CPU unit and integration tests: **40 passed**. Coverage includes GC validation and execution, diagnostic fields, bucketed transfer cleanup, rollout/checkpoint configuration, and LoRA sender forwarding.

```bash
bash tests/special_e2e/ppo_trainer/run_gc_diagnostics.sh
```

- One-GPU PPO smoke test: **passed**. It exercises initial weight synchronization and verifies that `point=weight_transfer_cleanup` and the configured generation are emitted.
- The new CPU tests are registered in the vLLM workflow, and the GPU smoke is registered in the PPO vLLM2 workflow with matching path triggers.

The configurable collection can affect two locations:

| Collection location | Timing scope |
| --- | --- |
| Initial weight synchronization | Startup before the training loop |
| Per-step weight synchronization | `update_weights` |

No standalone phase speedup is claimed for this PR because it preserves the historical full-GC default. Workload-specific timing and memory evidence is reported by the [verl-omni diffusion GC PR](https://github.com/verl-project/verl-omni/pull/560).

### API and Usage Example

The historical behavior remains the default:

```yaml
actor_rollout_ref:
  gc_diagnostics: false
  rollout:
    checkpoint_engine:
      gc_on_weight_transfer_cleanup: true
```

For example, enable diagnostics and collect generation 1 after sender cleanup:

```yaml
actor_rollout_ref:
  gc_diagnostics: true
  rollout:
    checkpoint_engine:
      gc_on_weight_transfer_cleanup: 1
```

The same values can be supplied as Hydra overrides:

```bash
actor_rollout_ref.gc_diagnostics=True \
actor_rollout_ref.rollout.checkpoint_engine.gc_on_weight_transfer_cleanup=1
```

### Design & Code Changes

- Add a shared `GCSetting` type, validation helper, and configured collection helper to `verl.utils.memory_utils`.
- Keep diagnostics optional and local to the collection call so measurements do not include later cache cleanup or synchronization.
- Configure weight-transfer cleanup without changing its default collection semantics.
- Forward the global diagnostics switch and cleanup setting through vLLM rollout construction and LoRA weight synchronization.
- Add CPU coverage and a minimal one-GPU end-to-end log assertion.

- [x] Ran the full pre-commit suite.
- [x] Added configuration comments and usage examples for the API change.
- [x] Added CPU and end-to-end tests to CI workflows that cover the changed paths.
- [ ] Request CI when the PR is ready for review.


ETOgaosion:
Thanks for contribution~ Recommend to make this behavior as default, no need for extra configuration.

wuxibin89:
> Thanks for contribution~ Recommend to make this behavior as default, no need for extra configuration.

No, I think the explicit GC collect is needed to free torch cuda caching allocator memory immediately. After `update_weights`, the follow-up actions are offloading trainer weights to CPU, then wake up rollout kv cache. When waking up kv cache, if training process still has caching cuda memory, the rollout process maybe cuda OOM.

@jiefangxuanyan GC cost in entire step should be a small percentage. Could you provide how many GC cost in you training step and torch memory profiling as well?

jiefangxuanyan:
Thanks, @ETOgaosion and @wuxibin89. Based on the measurements below, I prefer to keep the full collection as the conservative default while retaining the configuration. This PR does not change the existing default: `gc_on_weight_transfer_cleanup: true` still performs a full collection. The configuration lets a workload select a lower generation or disable the collection only after validating its own object and memory lifecycle.

I started this investigation while developing the FSDP2 regional compilation in [verl-omni PR #559](https://github.com/verl-project/verl-omni/pull/559). I found that regional compile made the explicit full collections substantially more expensive, causing GC to consume part of the compute-time savings. That led to this PR for the shared weight-transfer cleanup point and the dependent [verl-omni GC PR #560](https://github.com/verl-project/verl-omni/pull/560) for the three diffusion-specific points.

I measured this collection point together with the three additional diffusion GC points from PR #560 in an 8 x 80 GB Qwen-Image FlowGRPO workload using FSDP2 and vLLM-Omni. The four points are `eval_device_load` at the beginning of old-log-prob computation, `train_device_load` at the beginning of actor update, `actor_offload` during weight update, and `weight_transfer_cleanup` from this PR during weight update. The measurements strictly surround `gc.collect()` and exclude subsequent IPC cleanup, cache emptying, and synchronization. For the distributed critical-path estimate, I take the maximum GC wall time across ranks for each step and then average those per-step maxima. I exclude the initial weight synchronization and the first training step from the steady-state statistics.

I compared the same historical full-GC settings before and after enabling the FSDP2 regional compilation:

| Collection point | Eager per-rank mean | Compile per-rank mean | Increase with compile | Compile rank-max mean | Collected per call with compile | CUDA allocated/reserved delta |
| --- | ---: | ---: | ---: | ---: | ---: | ---: |
| `eval_device_load` | 569.7 ms | 914.8 ms | 60.6% | 1052.9 ms | 265 | 0 / 0 |
| `train_device_load` | 569.5 ms | 915.8 ms | 60.8% | 1056.2 ms | 265 | 0 / 0 |
| `actor_offload` | 566.4 ms | 914.4 ms | 61.4% | 1056.8 ms | 520 | 0 / 0 |
| `weight_transfer_cleanup` | 586.7 ms | 926.8 ms | 58.0% | 1062.4 ms | 330 | 0 / 0 |

Each eager point and the compiled `eval_device_load`, `train_device_load`, and `weight_transfer_cleanup` points contained 456 steady-state records. The compiled `actor_offload` result uses 448 records from 56 complete cross-rank events; the final partial event, in which only two ranks finished before the preset two-hour Slurm time limit, was excluded. Thus, regional compile increased the steady-state full-GC scan cost by roughly 58–61% at all four points. Nsight Systems/Python sampling showed that these intervals were CPU execution inside CPython GC (`gc_collect_main`, `deduce_unreachable`, and `visit_reachable`), rather than CUDA synchronization. All retained records reported `uncollectable=0`.

For the cleanup point in this PR specifically, the full-GC run contained 456 steady-state records. Its 1062.4 ms mean rank-max critical-path cost was 0.98% of the 108.09 s complete training step, but 14.15% of the 7.51 s `update_weights` phase. Therefore, your expectation that GC is a small percentage of the entire step is correct, while it is still material within the phase containing the collection.

I then configured the three diffusion-specific points as disabled and `weight_transfer_cleanup` as generation 1. The cleanup point produced 464 steady-state records, averaged 0.080 ms per rank and 0.106 ms on the rank-max critical path, and collected 446.6 objects per call. Full GC collected 330 objects per call in the corresponding compile run. Thus, generation 1 still had an observable collection effect at this lifecycle boundary while avoiding the expensive generation-2 scan in this workload.

At all four points, neither the eager full-GC run nor the compile full-GC run had any steady-state call with a nonzero PyTorch CUDA allocated or reserved delta. The generation-1 cleanup run also had zero CUDA allocated and reserved delta in every call. A zero reserved-memory delta alone is expected because the later cache-emptying operation is outside the measured interval; the more relevant observation is that `memory_allocated()` never decreased across `gc.collect()`. I therefore did not observe Python GC finalizing CUDA-backed tensor storage at these collection points in this workload. This does not mean full GC can never be necessary: another model or transfer implementation may leave CUDA tensors in older-generation reference cycles.

The compile/default-GC and compile/tuned-GC configurations both ran normally until their preset two-hour Slurm time limits, rather than terminating because of an error or resource failure. After the first step was excluded, they contributed 57 and 58 steady-state steps respectively. CUDA allocated memory did not accumulate at the common weight-transfer sampling point, and the tuned run did not show a worse RSS trend there. Changing the three diffusion points to disabled and `weight_transfer_cleanup` to generation 1 reduced the mean complete-step time from 108.09 s to 106.06 s, but because all four GC settings changed together and the runs used different nodes, I use this only as supporting end-to-end evidence rather than an isolated speedup claim for this collection point.

This is why I prefer keeping the conservative full-GC default while retaining the configuration: existing workloads keep their current cleanup semantics, diagnostics allow each workload to verify whether GC actually releases CUDA storage, a validated workload can select the lowest generation that collects its cleanup cycles, and the follow-up cache cleanup behavior remains unchanged.

ETOgaosion:
Actually I mean that since there are many gc points in verl, we can not add options to each one of them, we have to find the most effective way to insert gc to help decrease opportunities that training process OOM. @jiefangxuanyan I think you can make it more clear that the memory dump or actual system memory fluctuation during this process, so we can compare the cost performance of  your implementation and original ones.

jiefangxuanyan:
Thanks for clarifying, @ETOgaosion. I agree that adding a separate option to every existing GC call site would not scale. I think there are two related but different questions here: when a new forced GC is justified, and how to handle the forced GC calls that have already accumulated in existing workflows.

A forced GC can be useful immediately before a large CPU or CUDA allocation when unreachable reference cycles may still retain memory: neither CPython nor PyTorch proactively invokes cyclic GC and retries when its respective CPU or CUDA memory allocation fails. However, reference counting is likely to have handled most object lifetimes already. A new forced-GC call should therefore not be added based on speculation alone; it should have a clear memory-lifecycle rationale and empirical evidence that its benefit justifies its cost.

Existing forced-GC calls need more caution. I traced the origins of all four points covered here and found no point-specific experimental evidence for any of them; the details are summarized in [the history discussion in verl-omni PR #560](https://github.com/verl-project/verl-omni/pull/560#issuecomment-5630155164). However, these calls have since become part of the baseline exercised by many workloads. The range of RL workloads makes it impractical to prove that removing each existing call is safe everywhere, so an appropriate migration policy for them requires further discussion. This PR and the dependent [verl-omni PR #560](https://github.com/verl-project/verl-omni/pull/560) cover only the four existing GC points that I could validate in my Qwen-Image workload, while preserving their historical defaults.

For the memory comparison, I extracted the diagnostic records from the GC logs and uploaded them here: [gc_diagnostics_final_runs_public.csv](https://github.com/user-attachments/files/32083887/gc_diagnostics_final_runs_public.csv). My previous comment describes the measurement boundaries, aggregation, and GC timing, so I will focus here on the actual memory fluctuations. The CSV contains 4,218 records from eager/full-GC, compile/full-GC, and compile/tuned-GC runs, of which 4,104 are marked for steady-state analysis.

Across all 4,104 included calls, CUDA allocated and reserved memory never decreased. The immediate RSS results for the original full collections were:

| Collection point | Eager RSS decreases | Compile RSS decreases |
| --- | ---: | ---: |
| `eval_device_load` | 0 / 456 | 0 / 456 |
| `train_device_load` | 1 / 456 (-0.977 MiB) | 0 / 456 |
| `actor_offload` | 0 / 456 | 0 / 448 |
| `weight_transfer_cleanup` | 0 / 456 | 1 / 456 (-0.984 MiB) |

These two isolated sub-1-MiB changes are negligible relative to the roughly 50-GiB process RSS and do not show a repeatable system-memory benefit. The generation-1 `weight_transfer_cleanup` run likewise had essentially zero net immediate RSS change. RSS increased during warm-up and then fluctuated around a plateau, with no sustained increase in the later steps.

For this Qwen-Image workload, the original full collections therefore showed no CUDA-memory benefit and no meaningful repeatable RSS decrease. I disabled the three diffusion-specific points and kept a generation-1 collection at the shared `weight_transfer_cleanup` boundary as a low-cost safety hedge. This is the clearest resource-lifecycle boundary of the four: weight-transfer references have been deleted, while CUDA IPC collection and cache emptying follow immediately afterward. I selected generation 1 on that basis before the long-running experiment and continued using it; once its measured cost proved negligible, I did not pursue a further completely-disabled variant. This is a workload-specific setting, not a proposal to apply the same policy to every GC point or workload.

jiefangxuanyan:
@ETOgaosion, following our earlier discussion, I discussed the configuration design further with @zhtmike in the dependent VeRL-Omni PR. We agreed that users should not need to tune separate options for these GC points. Instead, regional compile will select the GC behavior validated for the Qwen-Image setup. The agreement is recorded in [this discussion](https://github.com/verl-project/verl-omni/pull/560#issuecomment-5632386657).

This follows your suggestion to avoid adding separate GC options. It also addresses @wuxibin89's concern about changing the behavior for existing workloads: regional compile remains disabled by default, so existing non-compile paths keep their current GC behavior. We have validated the compile and GC combination for Qwen-Image; other workloads should validate the combination before enabling regional compile.

This decision changes the scope of this PR. If VeRL-Omni no longer exposes the per-point configuration, the public `gc_on_weight_transfer_cleanup` option proposed here no longer has its original standalone use case. VeRL still needs the code-level ability to run a selected GC generation and to produce diagnostics, but the current PR should not be merged independently in its present form.

I therefore moved this PR back to draft so that the VeRL and VeRL-Omni implementations can be updated together. Does this direction address your concern?

jiefangxuanyan:
@ETOgaosion We have implemented the agreed approach and marked this PR ready for review again. The per-site GC configuration has been removed; the default behavior is preserved, while callers can select the validated GC behavior in code when regional compile is enabled.

wuxibin89:
From https://github.com/verl-project/verl/pull/7864, maybe we can avoid full gc by replacing `aggressive_empty_cache` to `get_torch_device().empty_cache()`.

But there's still another place where `gc.collect()+torch.cuda.ipc_collect()` maybe needed:
https://github.com/verl-project/verl/blob/main/verl/workers/rollout/vllm_rollout/bucketed_weight_transfer.py#L219-L222

wuxibin89:
Remove full GC completely in https://github.com/verl-project/verl/pull/7873

PR #7785, opened 2026-09-07
[trainer] fix: exclude all-padding rows from per-sequence off-policy metrics

### LINKED ISSUE #7771: [Bug][V1] Synthetic padding dilutes rollout-correction sequence metrics
### System Info

- Repository: `verl-project/verl`
- Unmodified source commit: `23af6a7a2e8d6efeeb2adbe5d1689c7a24f503a3`
- Linux x86_64; Python 3.12.14; PyTorch 2.13.0+cu130; TensorDict 0.10.0
- CPU-only reproduction with synthetic tensors; no model or dataset required. The reproduction script is custom; the framework source used to reproduce the bug is unmodified.

### Information

- [x] My own modified scripts

### Tasks

### Reproduction

**Adding synthetic padding rows makes the reported training-inference gap smaller even though every valid token and its log probabilities are unchanged.**

`compute_offpolicy_metrics` computes masked per-sequence means, then averages them over all rows. A row with no valid response tokens contributes mean log probability 0 and PPL 1 instead of being excluded.

This is reachable through stock V1 batch padding:

1. `_step_once` calls `_balance_batch` before `_compute_advantage`.
2. `_balance_batch` uses `upsample_batch_to_divisible_size` to satisfy batch divisibility. Its padding template sets `response_mask=0` and `is_padding=True`.
3. `_compute_advantage` passes the padded batch to `compute_rollout_correction_and_add_to_batch` without filtering padding tags when rollout correction is configured, rollout LPs are present, and bypass mode is disabled.
4. That wrapper calls `compute_offpolicy_metrics` with the original response mask. The separate `_compute_metrics` filtering of `is_padding` does not repair these already-computed metrics.

Source: [padding constructor](https://github.com/verl-project/verl/blob/23af6a7a2e8d6efeeb2adbe5d1689c7a24f503a3/verl/trainer/ppo/padding_utils.py), [V1 trainer](https://github.com/verl-project/verl/blob/23af6a7a2e8d6efeeb2adbe5d1689c7a24f503a3/verl/trainer/ppo/v1/trainer_base.py), [metric helper](https://github.com/verl-project/verl/blob/23af6a7a2e8d6efeeb2adbe5d1689c7a24f503a3/verl/trainer/ppo/rollout_corr_helper.py).

Run this in a VERL environment at the commit above:

```python
import torch
from verl.protocol import DataProto
from verl.trainer.ppo.padding_utils import construct_minimal_padding_template
from verl.trainer.ppo.rollout_corr_helper import compute_rollout_correction_and_add_to_batch

real = {"response_mask": torch.ones(2, dtype=torch.int64)}
padding, tag = construct_minimal_padding_template(real, {}, eos_token_id=2)
assert tag["is_padding"] and not padding["response_mask"].any()

for padding_count in (0, 1, 3):
    mask = torch.nn.utils.rnn.pad_sequence(
        [real["response_mask"]] + [padding["response_mask"]] * padding_count,
        batch_first=True,
    )
    batch = DataProto.from_dict(tensors={
        "old_log_probs": torch.full(mask.shape, -2.0),
        "rollout_log_probs": torch.full(mask.shape, -1.0),
        "response_mask": mask,
    })
    _, metrics = compute_rollout_correction_and_add_to_batch(batch, {})
    print(padding_count, metrics["rollout_corr/log_ppl_abs_diff"],
          metrics["rollout_corr/training_ppl"])
```

Observed on unmodified source:

```text
0 1.0 7.389056205749512
1 0.5 4.194528102874756
3 0.25 2.597264051437378
```

Affected metrics include training/rollout PPL and log-PPL, log-PPL difference and its absolute mean/extrema, PPL ratio, and sequence-level chi-squared estimates. Token-weighted KL/k3/chi-squared are unchanged in the finite-input controls. This report concerns diagnostic correctness, not a demonstrated change in gradients, rewards, or model quality.

### Related work / duplicate check

Searched open and closed issues/PRs for `compute_offpolicy_metrics`, `log_ppl_abs_diff`, `perplexity mask`, and padding/rollout metrics; no direct fix found.

- [#3984 discussion](https://github.com/verl-project/verl/pull/3984#issuecomment-3610544538): an entirely empty mask triggers an assertion, unlike this mixed valid/padding batch.
- [#5860](https://github.com/verl-project/verl/pull/5860): handles an all-empty batch in a different debug metric function.
- [#4510](https://github.com/verl-project/verl/pull/4510): defines loss/gradient aggregation denominators; this report does not propose changing them.
- [#7449](https://github.com/verl-project/verl/pull/7449): addresses padding sequence length for THD/CP, not off-policy metric reductions.

### Expected behavior

Adding all-masked synthetic padding must leave diagnostics over the valid sequences unchanged. All three cases above should report `log_ppl_abs_diff=1.0` and `training_ppl≈7.389056`.

A narrow fix is to use `response_mask.any(dim=-1)` when reducing sequence-level diagnostics inside `compute_offpolicy_metrics`. Preserve token-level reductions, the original/pre-rejection mask semantics, IS weights, losses, and the existing entirely-empty-batch error contract.

A local fix and regression tests are prepared: 22 focused CPU tests pass; including the existing rollout-correction tests, 52 pass. The same focused tests on unmodified source give 15 failures and 7 passing controls. Coverage includes upstream padding through the batch wrapper, signed extrema, unequal valid lengths, internal mask holes, training-only metrics, pre-rejection diagnostics, and unchanged input masks/IS weights. The existing suite emits one unrelated `PytestReturnNotNoneWarning`.

Would excluding all-masked rows from these diagnostic sequence reductions be the preferred fix location? I can submit the focused patch separately.

rrrxxx0510:
Fixes #7771

`_balance_batch` pads the batch up to a divisible size, and those padding rows have an all-zero `response_mask`. `compute_offpolicy_metrics` then averages per-sequence means over *all* rows, so every padding row silently contributes `ppl=1, log_ppl_diff=0, rho²=1` — the more padding rows you get, the more the rollout-correction metrics drift, even though nothing about the real tokens changed.

This just skips rows with no valid tokens when computing the per-sequence stats. Token-level metrics (kl, k3_kl, chi2_token) already only count valid tokens, so they're untouched.

Repro is CPU-only: with constant log probs, 3 padding rows moved `training_ppl` from 7.39 to 2.60 and `log_ppl_abs_diff` from 1.0 to 0.25. After the fix the metrics don't move no matter how many padding rows you add. Added a regression test; existing rollout_corr tests all pass (31).


rrrxxx0510:
Hi, do you have any thoughts on this one? If there are any issues, I'm happy to make changes.

rrrxxx0510:
嗨 @wuxibin89，这条修 #7771——全 padding 行会稀释 off-policy 指标，CPU 复现一行 padding 就把 training_ppl 从 7.39 拉到 2.60。改动不大，46 行 2 文件，加了回归测试 31 个 case 都过。

fork 的 CI 卡 action_required 跑不了，能帮忙批下 workflow 吗？有 review 意见我接着跟。

PR #7794, opened 2026-09-08
[trainer] fix: keep raw score for singleton groups in vectorized RLOO

### LINKED ISSUE #7793: [trainer] RLOO scalar and vectorized advantage disagree on singleton groups
The scalar and vectorized RLOO advantage implementations disagree on singleton groups (groups with a single sample), which can arise when filtering (e.g. overlong or dynamic filtering) drops all but one sample of a prompt group.

- `compute_rloo_outcome_advantage` (scalar): a singleton group keeps the raw score as the advantage (`id2mean = 0`, the update loop is skipped), so the surviving sample still gets a learning signal. This matches the GRPO singleton convention in `compute_grpo_outcome_advantage` (mean=0, std=1).
- `compute_rloo_vectorized_outcome_advantage` (vectorized): the final `* (c > 1)` zeroes the advantage, so the surviving sample contributes no gradient at all.

Both estimators are selected via `algorithm.adv_estimator` and are otherwise mathematically equivalent (I verified the leave-one-out algebra for multi-sample groups), so silently differing behavior on singletons looks unintended. Example with `n_samples_per_prompt=2` where one sample gets filtered:

```
group scores: [0.7]            # singleton after filtering
scalar:      advantage = 0.7   # trains on the surviving sample
vectorized:  advantage = 0.0   # sample contributes nothing
```

Suggested fix: pass the raw score through for singleton groups in the vectorized path (e.g. `adv = torch.where(c > 1, adv, scores)`), aligning it with the scalar implementation and the GRPO singleton convention. Happy to send a PR.

rrrxxx0510:
Fixes #7793

The scalar and vectorized RLOO implementations disagree on singleton groups, which can appear when filtering (e.g. overlong or dynamic filtering) drops all but one sample of a prompt group:

- scalar: a singleton keeps the raw score as the advantage, so the surviving sample still trains
- vectorized: the final `* (c > 1)` zeroed the advantage, so the surviving sample contributed no gradient at all

Both estimators are selected via `algorithm.adv_estimator` and are otherwise mathematically equivalent, so the silent divergence looks unintended. This passes the raw score through for singletons in the vectorized path, matching the scalar implementation and the GRPO singleton convention (mean=0, std=1).

Extended the equivalence test with singleton groups — it fails on the old vectorized behavior and passes after this change (24 tests pass).

### DIFF
--- a/tests/trainer/ppo/test_core_algos_on_cpu.py
+++ b/tests/trainer/ppo/test_core_algos_on_cpu.py
@@ -260,6 +260,27 @@ def test_rloo_and_vectorized_equivalence(batch_size: int, seq_len: int, num_grou
     assert torch.allclose(ret1, ret2, rtol=1e-5, atol=1e-6)
 
 
+def test_rloo_and_vectorized_equivalence_with_singleton_groups():
+    r"""Singleton groups (e.g. after filtering drops all but one sample) must keep
+    the raw score in both implementations (#7793)."""
+    token_level_rewards = torch.tensor([[0.7], [0.3], [-0.5], [0.9], [0.1]], dtype=torch.float32)
+    response_mask = torch.ones_like(token_level_rewards)
+    # two multi-sample groups and one singleton group
+    index = np.array(["p1", "p1", "p2", "p2", "p3"], dtype=object)
+
+    adv1, ret1 = compute_rloo_outcome_advantage(
+        token_level_rewards=token_level_rewards, response_mask=response_mask, index=index
+    )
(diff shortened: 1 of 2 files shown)


rrrxxx0510:
@wuxibin89 这条是 #7793：RLOO 标量版和向量化版在单例组上对不上——向量化把 advantage 清零，标量不清零，等于两条路径算的核心优势不一样。

加了个 torch.where 让向量化版在单例组保留原始分数，补了等价性测试（24 过；旧代码跑同一个测试挂，反过来证明确实是 bug）。CI 还卡 action_required，麻烦批下 workflow。

PR #7802, opened 2026-09-09
[sglang, rollout] fix: forward video features to SGLang

HeatherLiuzh:
### What does this PR do?

SGLang rollout silently drops video input, so VLMs answer video questions blind.

`AsyncSglangServer.generate()` never forwards `video_data` to the engine — the line is commented out
with a `TODO` and there is no warning. For a video task SGLang therefore receives `input_ids`
containing thousands of `<|video_pad|>` placeholder tokens but **no visual features at all**, and the
model produces fluent, well-structured, entirely hallucinated answers. vLLM does forward it, so the
two backends are not comparable on any video task today.

This PR routes the video features that verl **already computes and then discards** to SGLang through
its `format="processor_output"` channel, and picks the right payload per backend.

Measured on TinyLLaVA-Video-R1 / NextQA (5-way MC), Qwen3.5-35B-A3B, identical base weights,
identical 469-sample validation split, greedy decoding (`val_kwargs`: `temperature=0`, `n=1`,
`do_sample=False`):

| backend | val acc | KL(rollout‖actor) nats/token |
| --- | --- | --- |
| vLLM | 0.7868 | 0.003 – 0.006 |
| SGLang (before) | 0.4968 | 0.093 – 0.114 |
| SGLang (after) | 0.77186 | 0.013 - 0.016 |

0.4968 is what the model scores from language priors and option wording alone (random is 0.20).

Related: #2510 (same topic, idle since 2025-08-01, body starts with `DO NOT MERGE`) — see
[Relation to #2510](#relation-to-2510) below. cc @JustinTong0323

### Checklist Before Starting

  https://github.com/verl-project/verl/pulls?q=is%3Apr+is%3Aopen+sglang+video —
  found #2510 and #4513; the difference from #2510 is explained below.
- [x] Format the PR title as `[{modules}] {type}: {description}`:
  `[sglang, rollout] fix: forward video features to SGLang instead of silently dropping them`

### Test

Environment: verl main @ `89dad2d7`, SGLang 0.5.12, CUDA, Qwen3.5-35B-A3B (video VLM, MoE),
GRPO + LoRA, Megatron backend, dataset TinyLLaVA-Video-R1 / NextQA with `fps=1, max_frames=32`.

**1. End-to-end validation, SGLang before vs after** — `<fill in the measured acc>`

```bash
ROLLOUT_NAME=sglang bash <launch script> \
  trainer.val_only=True trainer.validation_data_dir=/tmp/valdump_sgl
```

**2. vLLM regression: inputs provably unchanged.** Ran validation before and after the patch and
compared per sample. Divergence between the two runs **never starts at the first character**
(0 of 463 differing samples), i.e. vLLM receives the same prompt as before. Had the patch altered
vLLM's inputs, the first-token distribution would differ and most samples would diverge at char 0.

**3. mRoPE unaffected (CPU-only).** Compared verl's `get_rope_index`
(`verl/models/transformers/qwen3_vl.py`) against SGLang's `MRotaryEmbedding.get_rope_index` on real
video samples with identical `input_ids` / `video_grid_thw`: position ids match element-wise on all
three rope axes for 2283 / 10762 / 1742-token prompts.

**4. Static checks.** Patch applies cleanly to main @ `89dad2d7`; all 6 touched files byte-compile;
an AST check confirms all 7 `build_initial_tokens` / `render_tokens_with_mm` implementations accept
the new keyword (see *Design & Code Changes*).

> ⚠️ Caveat for reviewers reproducing these numbers: greedy validation is **not** bit-reproducible
> here. Two runs of the same backend on the same weights gave only 6 / 469 byte-identical responses
> and ~12% of samples flipped correctness (acc moved 0.7910 → 0.7825), because continuous batching
> changes bf16 reduction order and flips near-tied argmax. Use
> `rollout.full_determinism=True` (+ `VLLM_BATCH_INVARIANT=1`) for tight A/B. The 29-point
> SGLang/vLLM gap is far outside this noise.

### API and Usage Example

No user-facing config change; video simply starts working with `rollout.name=sglang`.

Two internal signatures gain an optional keyword (backward compatible, defaulted to `None`):

```python
# verl/utils/tokenizer/continuous_token.py
def build_initial_tokens(self, messages, *, tools=None, images=None, videos=None, audios=None,
                         mm_inputs_out: dict[str, Any] | None = None) -> list[int]: ...

def render_tokens_with_mm(self, messages, images, *, videos=None, audios=None,
                          add_generation_prompt=True, tools=None,
                          mm_inputs_out: dict[str, Any] | None = None) -> list[int]: ...
```

Callers that want the video features pass a dict and get it filled in:

```python
mm_inputs = {}
prompt_ids = await self.ct_build_initial_tokens(messages, videos=videos, mm_inputs_out=mm_inputs)
# mm_inputs now holds pixel_values_videos, video_grid_thw, ...
```

### Design & Code Changes

Two independent gaps compose into the bug:

**1. Type mismatch.** SGLang's `video_data` accepts a path / URL / base64 string, a `VideoData`, or a
dict (`sglang/srt/managers/io_struct.py`: `VideoDataInputItem = Union[str, VideoData, Dict]`). verl
passes `list[tuple[torch.Tensor, dict]]` — already-decoded frames plus metadata. Uncommenting the
line alone fails inside `async_load_mm_data`; that is presumably why the `TODO` was never finished.

**2. verl computes the features, then throws them away.** In
`VLContinuousTokenMixin.render_tokens_with_mm`, `build_multimodal_processor_inputs` returns
`pixel_values_videos` and `video_grid_thw` alongside `input_ids`, but only `input_ids` is returned.
The base-class docstring says so explicitly: *"Pixel tensors are intentionally not returned here."*

SGLang supports exactly this case: a dict tagged `format="processor_output"` is passed through
untouched, bypassing `load_video` (`base_processor.py::_load_single_item`). Downstream,
`collect_mm_items_from_processor_output` claims `pixel_values_videos` / `video_grid_thw` /
`second_per_grid_ts` as `Modality.VIDEO`, and `processors/qwen_vl.py` reads `video_grid_thw` out of
the dict for mRoPE.

Changes (6 files):

- `verl/utils/tokenizer/continuous_token.py` — add `mm_inputs_out` to `render_tokens_with_mm` /
  `build_initial_tokens`; the VL mixin fills it with every processor field except `input_ids`.
- `verl/utils/tokenizer/deepseek.py` — `DeepSeekV4ContinuousTokenBuilder.build_initial_tokens`
  accepts and ignores it (text-only builder).
- `verl/experimental/agent_loop/agent_loop.py` — thread it through `ct_build_initial_tokens`.
- `verl/experimental/agent_loop/single_turn_agent_loop.py` — build
  `[{"format": "processor_output", "pixel_values_videos": ..., "video_grid_thw": ...}]`.
- `verl/workers/rollout/llm_server.py` — pick per backend: SGLang gets the processor features, vLLM
  keeps the raw frames. The extra kwarg is `pop`ped before dispatch because neither server signature
  accepts `**kwargs`. This mirrors the existing vLLM-only `priority` handling in the same function.
- `verl/workers/rollout/sglang_rollout/async_sglang_server.py` — send `video_data`.

All 7 implementations of the two methods are updated, including two overriding subclasses that would
otherwise raise `TypeError` once the caller passes the keyword:
`MiniMaxVLContinuousTokenBuilder.render_tokens_with_mm` (accepts and forwards to `super()`) and
`DeepSeekVL2ContinuousTokenBuilder.build_initial_tokens` (accepts and ignores; it renders through
`DeepseekVLV2Processor` and has no video path). Passing the keyword unconditionally is deliberate:
a missed builder fails loudly on first call instead of silently degrading to blind inference, which
is the exact failure mode this PR fixes.

**Why not let SGLang decode the video itself** (e.g. pass the file path): it would re-sample frames
with its own `preprocess_video`. Training-side sampling here is `fps=1, max_frames=32`; with no
`video_metadata` the processor falls back to `fps=24`, which I measured to change `video_grid_thw`
from `[[7, 30, 40]]` to `[[2, 30, 40]]`. That trades "no video" for "a different set of frames" and
desyncs rollout from the actor just as invisibly. Sharing one processor output keeps actor and
rollout identical on the vision path.

<a name="relation-to-2510"></a>
#### Relation to #2510

#2510 is open on the same topic but has been idle since 2025-08-01 and its description begins with
`DO NOT MERGE`. It changes one file, `sglang_rollout/sglang_rollout.py`, in the batch-level
`_batch_level_generate_sequences` path, forwarding the **raw** `multi_modal_data["video"]`.

Differences:

- It targets the batch-level path; the agent-loop `generate` path used on current main is still
  unfixed.
- It forwards raw frames, which do not match `VideoDataInputItem`, and does not address the
  frame-resampling divergence above.
- It does not separate vLLM from SGLang, so both backends receive the same payload.

@JustinTong0323 — is #2510 blocked on something specific (SGLang version, API shape)? Happy to rebase
this on your branch or close this in favour of yours, whichever you prefer.

- [ ] Add unit or end-to-end test(s) to [the CI workflow](https://github.com/verl-project/verl/tree/main/.github/workflows) to cover all the code. If not feasible, explain why: an end-to-end video rollout test needs a video VLM checkpoint and GPU; the mRoPE parity test described above is CPU-only and can be contributed if maintainers want it in CI.
- [ ] If your PR is related to the `recipe` submodule, update the submodule reference.

---

*AI assistance was used for this investigation and for drafting the patch and this description. I have
reviewed every changed line and can defend the change end-to-end.*


verl/experimental/agent_loop/single_turn_agent_loop.py
```diff
@@ -50,13 +50,31 @@ async def run(self, sampling_params: dict[str, Any], priority: int = 0, **kwargs
...
             videos=videos,
             audios=audios,
+            mm_inputs_out=mm_inputs,
         )
 
+        # SGLang cannot ingest raw frames: its ``video_data`` accepts a path/url/base64 or a dict.
+        # Hand it the processor features computed above; ``format="processor_output"`` makes
+        # BaseMultimodalProcessor pass the dict through untouched instead of re-decoding the video,
+        # which would also re-sample frames and desync the rollout from the actor.
+        mm_processor_output = None
```
wuxibin89: `ToolAgentLoop` should be adjusted as well, or can we put it in a common place?

HeatherLiuzh: Good point — ToolAgentLoop has the exact same shape (ct_build_initial_tokens → generate(video_data=...)), so I'll factor the payload construction into a shared helper on AgentLoopBase rather than duplicating it.

HeatherLiuzh:
**Addressed** — moved the payload construction into a shared helper instead of duplicating it, and covered ToolAgentLoop as well.

**Common place**: new `AgentLoopBase.build_sglang_video_payload(videos, mm_inputs)` in `agent_loop.py `now owns the format="processor_output" wrapping and returns None when there's no video, so each call site is a single line:
`mm_processor_output = self.build_sglang_video_payload(videos, mm_inputs)`
**Both loops updated**:

- SingleTurnAgentLoop: replaced the inline dict with the helper call.

- ToolAgentLoop: builds it in _handle_pending_state (where the prompt is tokenized), stashes it on AgentData.mm_processor_output, and passes it in _handle_generating_state — mirroring how it already threads video_data through those two methods.

**One caveat worth flagging for ToolAgentLoop:** it's multi-turn, and I send the initial prompt's video payload on every generate call, which matches how raw video_data was already being passed each turn. Tool-returned video is still explicitly unsupported (tool_agent_loop.py raises "Multimedia type 'video' is not currently supported"), so this only covers video present in the initial prompt. Someone more familiar with the tool loop's multi-turn semantics may want to sanity-check that.

wuxibin89 (dismissed):

verl/experimental/agent_loop/single_turn_agent_loop.py
```diff
@@ -68,6 +74,7 @@ async def run(self, sampling_params: dict[str, Any], priority: int = 0, **kwargs
                 image_data=images,
                 audio_data=audios,
                 video_data=videos,
+                mm_processor_output=mm_processor_output,
```
wuxibin89, line 77: OPD teacher model should be adjusted as well:
https://github.com/verl-project/verl/blob/main/verl/experimental/teacher_loop/teacher_manager.py#L128-L136

verl/experimental/agent_loop/agent_loop.py
```diff
@@ -356,6 +358,31 @@ async def ct_build_initial_tokens(
...
+        there is no video, so callers can hand the result straight to ``generate(...)``.
+
+        Shared by SingleTurnAgentLoop and ToolAgentLoop; vLLM keeps the raw frames and ignores this
+        (the per-backend split lives in ``LLMServerClient.generate``).
+        """
+        if not videos or "pixel_values_videos" not in mm_inputs:
+            return None
+        payload = {
+            "format": "processor_output",
+            "pixel_values_videos": mm_inputs["pixel_values_videos"],
```
wuxibin89, line 382: Is these keys: `pixel_values_videos / video_grid_thw / second_per_grid_ts` specific to Qwen series model?

verl/experimental/agent_loop/single_turn_agent_loop.py
HeatherLiuzh: Good catch. TeacherManager.compute_teacher_logprobs_single forwards raw video_data=multi_modal_data.get("videos") to client.generate, and since teacher_client is the same LLMServerClient, an SGLang teacher would drop the video exactly like the rollout path did.

verl/experimental/agent_loop/agent_loop.py
HeatherLiuzh: Yes — pixel_values_videos / video_grid_thw / second_per_grid_ts are the Qwen2-VL / 2.5-VL / 3-VL processor output names. They're not arbitrary though: they're exactly the keys SGLang's collect_mm_items_from_processor_output maps to Modality.VIDEO (base_processor.py ATTR_NAME_TO_MODALITY), so the producer and consumer sides agree. For a non-Qwen video VLM, both verl and SGLang would need that family's keys added.

On "what happens for a non-Qwen video model": build_sglang_video_payload returns None when the processor output has no pixel_values_videos, and LLMServerClient.generate now sets video_data = mm_processor_output for the SGLang backend unconditionally — so SGLang gets the dict for Qwen-VL and None otherwise. It never receives the raw frame tuples (list[tuple[Tensor, dict]]), which its video_data (path/url/base64/dict only) cannot parse.

I'change it,For an unsupported family, video is dropped (same as before this PR) rather than crashing the server. Making it actually work would require that family's video keys to be recognized on both sides — verl's processor output and SGLang's collect_mm_items_from_processor_output — which I'd suggest as a follow-up. vLLM is unaffected: it keeps the raw frames and never enters the SGLang branch.

verl/experimental/agent_loop/single_turn_agent_loop.py
HeatherLiuzh: Done. Threaded the payload through: AgentLoopOutput now carries mm_processor_output, both loops set it, _compute_teacher_logprobs forwards it via getattr(output, "mm_processor_output", None) (so any AgentLoopOutput without the field is a safe no-op), and TeacherManager.compute_teacher_logprobs_single passes it to client.generate. Since the teacher client is the same LLMServerClient, the per-backend split already handles it — SGLang teacher gets the dict, vLLM keeps raw frames.

HeatherLiuzh:
@wuxibin89 Common place for the video payload: extracted into AgentLoopBase.build_sglang_video_payload; both SingleTurnAgentLoop and ToolAgentLoop now call it instead of duplicating the dict.
OPD / teacher path: threaded the payload through AgentLoopOutput.mm_processor_output → _compute_teacher_logprobs (via getattr(output, "mm_processor_output", None), so any output without the field is a safe no-op) → TeacherManager.compute_teacher_logprobs_single → client.generate. Since the teacher client is the same LLMServerClient, the per-backend split handles it — SGLang teacher gets the dict, vLLM keeps raw frames.
Non-Qwen video models: LLMServerClient.generate now sets video_data = mm_processor_output unconditionally for the SGLang backend, so SGLang never receives raw frame tensors it can't parse. For a family whose processor produces no pixel_values_videos, the payload is None → video is dropped (same as before this PR) rather than crashing.
**The two red CI jobs are pre-existing, not from this PR**
Both failures are in code this PR doesn't touch (my changes are confined to tokenizer/ agent_loop/ rollout/ teacher_loop/):

test_preprocess_thd_engine_preserves_1d_zigzag_roll_alignment — verl/models/mcore/util.py::preprocess_thd_engine. This is a deterministic test, and that file just went through a land → revert → re-land cycle for THD-packed router replay right before current main (#7106 / #7786 / #7805). It reproduces on a clean main @ 9e0252ef checkout.

test_tool_agent (num_turns == 2 vs 4) — a stochastic tool-calling test on Qwen3-0.6B at temperature=1.0; num_turns==2 just means the small model didn't emit a tool call on that sample. This PR is inert for text-only inputs: prompt ids are byte-identical (the text builder ignores the new mm_inputs_out kwarg), build_sglang_video_payload returns None, and that None is popped in LLMServerClient.generate with no effect — generation completed with no exception, so the added kwargs were accepted. A re-run should clear it.
Verified locally: patch applies cleanly on main @ 9e0252ef, all 8 touched files compile, ruff check / ruff format pass. Could you re-review and re-run CI when you get a chance? Happy to rebase past #7805 once it settles if that's cleaner.

wuxibin89 (approved):

PR #7805, opened 2026-09-09
[megatron] fix: make R2 router replay model-scoped and opt-in

hbhflw2000:
### What does this PR do?

This PR reintroduces the R2 router replay correctness fix from #7106 after addressing the regression that caused it to be reverted by #7786.

The reverted implementation queried the process-global router registry even when router replay was disabled. Normal Megatron SFT and RL jobs therefore failed when the registry was empty.

This version makes router replay explicitly opt-in and scopes RECORD, REPLAY, and action transitions to the routers owned by the currently forwarded model.

### Design and Code Changes

- Guard the complete router replay lifecycle behind the configured replay mode.
- Treat an empty router registry as a no-op when replay is disabled.
- Preserve hard failures when R2 is active but routing data is missing or incomplete.
- Collect recorded routes from the currently forwarded model instead of scanning the process-global registry.
- Apply replay targets and action transitions to the forwarded model's routers.
- Support PP/VPP without shared mutable per-VP router caches.
- Reuse the existing THD preprocess/postprocess padding semantics.
- Reject R2 with `rollout_correction.bypass_mode=True`, since R2 requires the actor old-logprob forward.
- Resolve Megatron attributes from nested multimodal Hugging Face text configs.

The existing `model=None` positional helper path remains available for backward compatibility.

### Regression Addressed

The previous implementation caused replay-disabled Megatron jobs to fail with errors such as:

```text
RuntimeError: cannot map router replay registry to the current PP/VPP stage:
registry=0, ...
```

The corrected implementation does not inspect replay state on the default disabled path.

### Validation

- Focused router replay, THD, nested-config, and model-walk tests: **39 passed**
- Targeted pre-commit checks:
  - ruff
  - ruff-format
  - mypy
  - compileall
  - license and repository structure checks
  - generated trainer configs
- Local pruned replay-disabled fully-async E2E:
  - **1/1 actor update completed**
  - rollout parameter version advanced from `0` to `1`
- 4-node, 32-GPU Qwen3-Omni R2 PP2/VPP2 fully-async E2E:
  - R2 enabled on the actor workers
  - 48 replay layers discovered
  - **1/1 actor update completed**
  - post-update weight synchronization completed
  - rollout parameter version advanced from `0` to `1`
  - process exited successfully

### Relationship to Previous Work

- Supersedes the reverted implementation in #7106.
- Addresses the regression documented by #7786.
- Reuses the forwarded-model router addressing design already merged in #7340.

### Non-goals

- Changing R3 routing semantics.
- Changing Megatron-Core routing or all-to-all kernels.
- Changing rollout-server or weight-transfer behavior.

### DIFF
--- a/verl/utils/megatron/router_replay_utils.py
+++ b/verl/utils/megatron/router_replay_utils.py
@@ -784,8 +784,9 @@ class RouterReplayHelper:
         Return the list of RouterReplay instances corresponding to the current micro-batch and local
         (pp_rank, vp_stage) layer range.
 
-        When virtual pipeline (VPP) is enabled, routers for earlier local VP stages precede the
-        current stage in the process-global registry.
+        Accept either one PP rank's complete registry in local VPP order, or one complete
+        model's registry in global layer order. These positional layouts must not contain
+        stale routers or routers from other models; use the model-scoped helpers for those cases.
 
         Args:
             tf_config: Configuration object used to compute layer assignments.
@@ -797,22 +798,27 @@ class RouterReplayHelper:
         if not RouterReplay.router_instances:
             return []
 
-        vp_size = tf_config.virtual_pipeline_model_parallel_size
-        if vp_size is not None:
-            vp_rank = 0 if vp_rank is None else vp_rank
-            offset = sum(get_moe_num_layers_to_build(tf_config, stage) for stage in range(vp_rank))
-        else:
-            offset = 0
+        vp_size = tf_config.virtual_pipeline_model_parallel_size or 1
+        vp_rank = 0 if vp_rank is None else vp_rank
+        if not 0 <= vp_rank < vp_size:
+            raise ValueError(f"vp_rank={vp_rank} is outside the configured VPP size {vp_size}.")
 
-        num_layers_to_build = get_moe_num_layers_to_build(tf_config, vp_rank)
-        router_instances_list = RouterReplay.router_instances[offset : offset + num_layers_to_build]
-        if len(router_instances_list) != num_layers_to_build:
+        local_counts = [get_moe_num_layers_to_build(tf_config, stage) for stage in range(vp_size)]
+        local_total = sum(local_counts)
+        global_total = sum(is_moe_layer(tf_config, idx) for idx in range(tf_config.num_layers))
+        registry_size = len(RouterReplay.router_instances)
+        if registry_size == local_total:
+            offset = sum(local_counts[:vp_rank])
+        elif registry_size == global_total:
+            layer_start = get_current_rank_layer_info(tf_config, vp_rank)["start"]
+            offset = sum(is_moe_layer(tf_config, idx) for idx in range(layer_start))
+        else:
             raise RuntimeError(
-                "router replay registry does not cover the current PP/VPP stage: "
-                f"registry={len(RouterReplay.router_instances)}, offset={offset}, "
-                f"expected={num_layers_to_build}, actual={len(router_instances_list)}, vp_rank={vp_rank}"
+                "router replay registry does not cover a complete local PP/VPP or global model layout: "
+                f"registry={registry_size}, local_expected={local_total}, global_expected={global_total}, "
+                f"vp_rank={vp_rank}. Use model-scoped replay for non-positional registries."
             )
-        return router_instances_list
+        return RouterReplay.router_instances[offset : offset + local_counts[vp_rank]]
 
     @staticmethod
     def _get_action_router_list(tf_config, vp_rank=None, model=None):
--- a/verl/utils/megatron/router_replay_utils.py
+++ b/verl/utils/megatron/router_replay_utils.py
@@ -264,18 +264,45 @@ def merge_router_topk_indices(
         [1, dynamic_bs_all, layer_num, topk] to mini_layer_topk_idx_list.
     """
     with torch.no_grad():
+        if not input_ids.is_nested and attention_mask is None:
+            raise RuntimeError("router replay RECORD requires attention_mask for non-nested BSHD inputs.")
         vp_rank = 0 if vp_rank is None else vp_rank
         if model is not None:
-            router_instances_list = [router for _, router in iter_model_routers(model)]
+            model_routers = sorted(iter_model_routers(model), key=lambda item: item[0])
+            layer_numbers = [layer_number for layer_number, _ in model_routers]
+            if len(layer_numbers) != len(set(layer_numbers)):
+                raise RuntimeError(
+                    "router replay RECORD found duplicate layer numbers in the forwarded model: "
+                    f"layer_numbers={layer_numbers}, vp_rank={vp_rank}"
+                )
+            router_instances_list = [router for _, router in model_routers]
         else:
             router_instances_list = RouterReplayHelper.get_micro_batch_router_list(tf_config, vp_rank)
 
         if not router_instances_list:
-            raise RuntimeError(
-                "router replay RECORD found no routers in the forwarded model: "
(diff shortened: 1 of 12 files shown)


SamitHuang (commented):
should-fix

**1. `transformer_impl.py` ~L984-L999 — empty `mini_layer_topk_idx_list` vs `pp_gather`**

`record_r2_routes` is batch/config scoped, but recording is still skipped when the forwarded chunk has no MoE routers (`is_r2_record_action` is false for an empty `iter_model_routers` walk). On a PP rank that owns 0 MoE layers (embedding-only first stage, `moe_layer_freq` leaving a dense stage, etc.) this now raises before `pp_gather`.

`pp_gather` is a PP `all_gather` / `all_gather_object`. Every rank must enter it. Prefer: if this rank has no local MoE routers, still participate with a 0-layer tensor (or an explicit empty nested map) instead of raising. Keep the hard fail only when the rank *has* routers but captured nothing.

**2. `router_replay_utils.py` `merge_router_topk_indices` ~L268-L269 — stack order vs replay index**

```python
router_instances_list = [router for _, router in iter_model_routers(model)]
```

`layer_number` is dropped. The stacked layer dim is `modules()` walk order. `set_router_replay_data` then indexes by `layer_number - 1` / global MoE ordinal after PP gather. If walk order is not increasing `layer_number` (HybridModel extras, unexpected module registration), RECORD and REPLAY silently disagree. Collect `(layer_number, router)` and sort by `layer_number` before stacking.

## Non-blocking

**3. `get_hf_config_attr` / `_iter_hf_config_tree` — sibling fallback**

After `get_text_config()` misses, BFS can return the first sibling that has the attr (audio/code2wav vs text). The Omni test is protected by `get_text_config()`. If text lookup fails, fail (or restrict the walk to the text branch) rather than picking a conflicting multimodal sibling. `getattr(..., None)` also treats legitimate `0`/`False` as missing, which is probably fine for `hidden_size` / `rope_theta`.

**4. `get_micro_batch_router_list` empty-registry no-op is good; non-empty short slice is silent**

The `#7786` empty-registry crash is fixed. If the registry is non-empty but shorter than `offset + num_layers_to_build`, the fallback path still returns a short list with no mapping error. The `model=` path is what production uses now; consider restoring a length check on the positional helper, or deleting it if it is only for back-compat tests.

**5. `attention_mask` for the BSHD packed path**

Passing the real mask into `merge_router_topk_indices` / `set_router_replay_data` is correct for packed BSHD. `prepare_model_inputs` still does `batch.get("attention_mask", None)`. THD nested does not use it; non-nested + `attention_mask is None` will throw on `.shape`. Megatron does not currently reject `router_replay` without `use_remove_padding` (VeOmni does). Either assert nested/THD or require a mask on the BSHD path.

**6. Bypass-mode validator**

`_validate_router_replay_config` looking at `actor_config.engine.router_replay` is correct after `McoreActorConfig.__post_init__` aliases `engine = megatron`. Fully-async YAML defaults `bypass_mode: True`, so enabling R2 will fail closed at `validate_config` unless bypass is turned off — that matches the stated design. Worth a one-line note in the engine/router-replay docs so fully-async users are not surprised.

verl/workers/engine/megatron/transformer_impl.py
```diff
@@ -975,7 +981,9 @@ def forward_backward_batch(self, data: TensorDict, loss_function: Callable, forw
                     losses_reduced[0]["metrics"] = {}
                 losses_reduced[0]["metrics"].update(metrics)
 
-        if RouterReplayHelper.is_r2_record_action(self.tf_config):
+        if record_r2_routes:
+            if not self.mini_layer_topk_idx_list:
```
SamitHuang, line 986: **Should-fix:** `record_r2_routes` is batch/config-scoped, but recording is skipped when the forwarded chunk has no MoE routers (`is_r2_record_action` false for an empty `iter_model_routers` walk).

On a PP rank that owns 0 MoE layers (embedding-only first stage, `moe_layer_freq` leaving a dense stage, etc.) this raises before `pp_gather`. `pp_gather` is a PP collective — every rank must enter it.

Prefer: if this rank has no local MoE routers, still participate with a 0-layer tensor (or an explicit empty nested map) instead of raising. Keep the hard fail only when the rank *has* routers but captured nothing.

verl/utils/megatron/router_replay_utils.py
```diff
@@ -250,13 +256,41 @@ def merge_router_topk_indices(
...
 
     Returns:
         None: The function has side effects only; it appends a tensor of shape
         [1, dynamic_bs_all, layer_num, topk] to mini_layer_topk_idx_list.
     """
     with torch.no_grad():
-        router_instances_list = RouterReplayHelper.get_micro_batch_router_list(tf_config, vp_rank)
+        vp_rank = 0 if vp_rank is None else vp_rank
+        if model is not None:
+            router_instances_list = [router for _, router in iter_model_routers(model)]
```
SamitHuang: **Should-fix:** this drops `layer_number` and stacks in `modules()` walk order. `set_router_replay_data` later indexes by `layer_number - 1` / global MoE ordinal after PP gather.

If walk order is not increasing `layer_number` (HybridModel extras, unexpected module registration), RECORD and REPLAY silently disagree. Collect `(layer_number, router)` and sort by `layer_number` before stacking.

verl/utils/megatron_utils.py
```diff
@@ -251,31 +251,85 @@ def get_model(
...
+
+def get_hf_config_attr(hf_config: PretrainedConfig, attr_name: str) -> Any:
+    """Find an attribute in a nested Hugging Face model configuration."""
+    text_config = _get_hf_text_config(hf_config)
+    if text_config is not None:
+        value = getattr(text_config, attr_name, None)
+        if value is not None:
+            return value
+
+    for config in _iter_hf_config_tree(hf_config):
```
SamitHuang, line 298: **nit:** after `get_text_config()` misses, this BFS can return the first multimodal sibling that has the attr (audio/code2wav vs text). The Omni test is protected by `get_text_config()`. If text lookup fails, consider failing (or restricting the walk to the text branch) rather than picking a conflicting sibling.

Also `getattr(..., None)` treats legitimate `0`/`False` as missing — probably fine for `hidden_size`/`rope_theta`, but worth keeping in mind.

SamitHuang (commented):
Three inline notes:
1. `transformer_impl.py` — empty local MoE stage vs `pp_gather` collective
2. `router_replay_utils.py` — sort by `layer_number` before stacking RECORD routes
3. `megatron_utils.py` — avoid multimodal sibling attr fallback (nit)

Overall direction still LGTM after the #7786 regression fix.

hbhflw2000:
Thanks @SamitHuang ,  the problems addressed in 0fdc1f83 and f7b44d84:
- Preserve empty PP/VPP chunks and handle uneven route counts during gathering; missing routes on expected MoE stages still fail.
- Sort RECORD routers by layer number to match REPLAY ordering.
- Restrict nested config lookup to text wrappers, preserving 0/False.
- Validate both complete local and global legacy registry layouts, rejecting incomplete mappings.
- Add explicit BSHD attention-mask validation and document the R2 bypass-mode requirement.
The production path remains model-scoped, without reintroducing shared per-VP caches or mutable RECORD-padding state.
Validation: 87 focused tests, 8 real four-GPU PP/VPP cases, and full pre-commit passed on the latest head. At 0fdc1f83, replay-disabled four-GPU and R2 PP2/VPP2 32-GPU E2E both completed one update and weight synchronization. The final legacy-helper follow-up passed local regressions; the 32-GPU run was not repeated for it.

wuxibin89 (approved):

PR #7825, opened 2026-09-10
[ci] fix: point workflow path filters at files that still exist

linhongyu510:
### What does this PR do?

A `paths:` filter entry naming a file that no longer exists keeps parsing fine and GitHub reports nothing — the entry just silently stops matching. **49 such entries** had accumulated across **22 workflows**, all naming one of three files that were removed some time ago:

| removed file | current equivalent |
|---|---|
| `verl/trainer/fsdp_sft_trainer.py` | `verl/trainer/sft_trainer.py` |
| `verl/trainer/main_generation.py` | *(gone; the generation job it guarded no longer exists)* |
| `tests/rollout/async_rollout_utils.py` | *(gone; `tests/rollout/` was removed)* |

The failure is silent and points **the opposite way** depending on the sign of the entry, so this is two distinct bugs:

**1. Positive entries stopped firing — the three SFT workflows lost their entrypoint trigger.**

`e2e_sft_llm.yml`, `e2e_sft_llm_ascend.yml` and `e2e_sft_vlm.yml` listed exactly two entrypoints:

```yaml
      - "verl/trainer/fsdp_sft_trainer.py"
      - "verl/trainer/config/sft_trainer.yaml"
```

Both are gone. The trainer those jobs actually run is `verl/trainer/sft_trainer.py` — from `tests/special_e2e/sft/run_sft.sh`:

```bash
ENTRYPOINT=${ENTRYPOINT:-"-m verl.trainer.sft_trainer"}
```

and its Hydra config is `sft_trainer_engine.yaml`, not `sft_trainer.yaml`:

```python
@hydra.main(config_path="config", config_name="sft_trainer_engine", version_base=None)
```

So a PR touching only the SFT trainer did not list these three e2e workflows as required checks. Repointed both entries.

**2. Negative entries stopped excluding — 35 of them.**

`- "!verl/trainer/fsdp_sft_trainer.py"` appears in 35 places, and the intent is the reverse: skip expensive e2e/unit jobs when a change only touches the SFT trainer. Since the path no longer matches anything, the exclusion became a no-op and those jobs run more often than intended. Repointed to the current path.

**3. vllm/sgl entries pointed at removed jobs.**

`vllm.yml` / `vllm_ascend.yml` still triggered on `tests/special_e2e/generation`, `verl/trainer/main_generation.py` and `verl/trainer/config/generation.yaml` — none exist, and neither workflow runs a generation job any more. `sgl.yml` / `sgl_ascend.yml` triggered on `tests/rollout/*sglang*` and `tests/rollout/async_rollout_utils.py`, but `tests/rollout/` is gone and the sglang tests now live in `tests/workers/rollout/rollout_sglang`. Replaced these with the paths those workflows actually exercise (verified against each file's own `pytest` invocations).

### Test

Added `tests/special_sanity/check_workflow_path_filters.py`, registered in `.pre-commit-config.yaml`. `pre-commit.yml` has **no** path filter and runs on every PR, so unlike `sanity.yml` (which is gated on `**/*.py` and would skip a YAML-only change) this actually enforces.

The hook is deliberately narrow to avoid false positives:

- only **literal** entries — anything with `*`, `?`, `[` or `${{ }}` is a pattern and may legitimately match nothing today
- skips `recipe/`, a submodule whose working tree is empty until `git submodule update --init`, so checking it would fail based on checkout state rather than correctness
- an `ALLOW_LIST` (currently empty) for entries intentionally kept

Verified the hook reproduces the bug and confirms the fix:

```
$ git stash && python3 tests/special_sanity/check_workflow_path_filters.py | grep -c ' -> '
49                       # on the parent commit
$ git stash pop && python3 tests/special_sanity/check_workflow_path_filters.py ; echo $?
0                        # after this change
```

28 unit tests covering both signs, `paths-ignore`, globs/expressions, submodule paths, directory entries, unquoted and trailing-comment entries, block boundaries, `ALLOW_LIST`, and all three exit codes.

I mutation-tested the hook rather than trusting a green run — two deliberate breakages (ignoring `!`-prefixed entries; dropping the sequence-dedent check) each fail a specific test. That second mutation initially **survived**, which exposed a real flaw in my own parser: a `paths:` whose items sit at the key's own indent (valid YAML) was being skipped entirely. Fixed, and both tests now catch it.

Other checks:

- all 46 workflows still parse (`yaml.safe_load`)
- semantic diff of every changed workflow at the **parsed** level, confirming only the intended entries moved and nothing else shifted
- `pre-commit run --all-files`: all hooks pass

### API and Usage Example

No API change. The new hook runs automatically; to run it directly:

```bash
python3 tests/special_sanity/check_workflow_path_filters.py
```

### Design & Code Changes

- 22 workflows: repointed or removed 49 stale literal path-filter entries
- `tests/special_sanity/check_workflow_path_filters.py`: new guard
- `tests/special_sanity/test_check_workflow_path_filters.py`: 28 tests
- `.pre-commit-config.yaml`: register `check-workflow-path-filters`

No job logic, step, runner or trigger *semantics* changed — only which paths the existing filters name.

- [x] Search for similar PRs: [`workflow paths filter stale`](https://github.com/verl-project/verl/pulls?q=repo%3Averl-project%2Fverl+workflow+paths+filter+stale) and [`is:pr fsdp_sft_trainer`](https://github.com/verl-project/verl/pulls?q=repo%3Averl-project%2Fverl+is%3Apr+fsdp_sft_trainer) — nothing covering this.
- [x] PR title formatted as `[{modules}] {type}: {description}` (validated locally with `tests/special_sanity/check_pr_title.py`).
- [x] Apply pre-commit checks — all hooks pass.
- [ ] Add / Update the documentation — not applicable, no user-facing behaviour changes.
- [x] Add unit test(s) to cover the code — 28 tests, plus the hook itself runs in `pre-commit.yml` on every PR.
- [ ] Send a message in the `ci-request` channel — **I don't have access to the verl Slack workspace or the Feishu group, so I can't trigger CI myself.** Could a maintainer kick it off? Worth noting that most of the touched workflows are GPU e2e jobs whose own trigger filters this PR changes, so it's a useful one to actually run.

---

One judgement call worth flagging for review: for `vllm`/`sgl` I **replaced** the dead entries with the test paths those workflows actually run today, rather than just deleting them. Deleting would have been the minimal change, but it would leave those workflows with almost no positive triggers. If you'd prefer the strictly minimal diff there, I'm happy to drop the replacements.

Assembled with AI assistance; every command, count and hash above is from a local run, and the before/after hook output is the thing I'd ask a reviewer to check first.


linhongyu510:
Small correction to my own checklist note above, in case it's sending anyone the wrong way.

I'd written that I couldn't trigger CI because I lack Slack access. That's true but it isn't what's actually blocking this PR — **no checks have run at all**, including the ones that need neither a GPU nor a CI request:

```
PR #7825 head 1e3e530 -> check runs: 0
```

For comparison, recently merged #7799 has 5 check runs on its head — `pre-commit (3.12)`, `check-title`, `test` — all lightweight `ubuntu-latest` jobs. `pre-commit.yml` triggers on a bare `pull_request:` with no path filter and no environment gate, so it should have run here too.

The difference is that I'm a first-time contributor to this repo (`author_association: NONE`), so everything is held behind GitHub's **"Approve and run workflows"** button on the PR page. The `ci-request` message is a separate, *additional* step for the GPU e2e jobs.

So concretely, for whoever picks this up:

1. **"Approve and run workflows"** — this alone gets `pre-commit`, `check-title` and `test` running, and `pre-commit` is the meaningful one here since that's where the new hook is registered. It's also the cheap way to sanity-check this PR, because the hook failing or passing is most of the signal.
2. **`ci-request`** — only if you also want the GPU e2e jobs. Worth doing at some point given this PR edits those workflows' own trigger filters, but it isn't needed to validate the change itself.

Sorry for the noise — I should have checked the check-run list before attributing it to Slack access.

Issue #5273, opened 2026-02-10
labels are missing from verl/verl/models/mcore/model_forward_fused.py


sneng1997:
### System Info

----------Python Info----------
Version      : 3.12.12
Compiler     : GCC 11.4.0
Build        : ('main', 'Oct 10 2025 08:52:57')
Arch         : ('64bit', 'ELF')
------------Pip Info-----------
Version      : 26.0
Directory    : /usr/local/lib/python3.12/dist-packages/pip
vllm         : 0.12.1.dev0+g4fd9d6a85.d20260116
sglang       : not found.
ray          : 2.53.0
torch        : 2.9.1+cu129
----------verl Info-----------
Version      : 0.8.0.dev
Directory    : /verl/verl
Commit Hash  : e4915bd8859e75a8634ec48c20a0bbc781d66eb8
----------Platform Info----------
Platform     : Linux-5.10.223-212.873.amzn2.x86_64-x86_64-with-glibc2.35
system       : Linux
node         : ip-10-1-31-153.us-west-2.compute.internal
release      : 5.10.223-212.873.amzn2.x86_64
version      : #1 SMP Wed Aug 7 16:53:32 UTC 2024
----------Environment----------
CUDA Runtime : 12.9
CUDA Compiler : Cuda compilation tools, release 12.9, V12.9.86
----------System Info----------
CPU Memory      : 1999.97 GB
GPU Count       : 8
GPU 1   Type    : NVIDIA H100 80GB HBM3
GPU 1   Memory  : 79.65 GB
GPU 2   Type    : NVIDIA H100 80GB HBM3
GPU 2   Memory  : 79.65 GB
GPU 3   Type    : NVIDIA H100 80GB HBM3
GPU 3   Memory  : 79.65 GB
GPU 4   Type    : NVIDIA H100 80GB HBM3
GPU 4   Memory  : 79.65 GB
GPU 5   Type    : NVIDIA H100 80GB HBM3
GPU 5   Memory  : 79.65 GB
GPU 6   Type    : NVIDIA H100 80GB HBM3
GPU 6   Memory  : 79.65 GB
GPU 7   Type    : NVIDIA H100 80GB HBM3
GPU 7   Memory  : 79.65 GB
GPU 8   Type    : NVIDIA H100 80GB HBM3
GPU 8   Memory  : 79.65 GB

### Information

- [ ] The official example scripts
- [ ] My own modified scripts

### Tasks

[...]

PR #7826, opened 2026-09-10
[megatron] fix: keep fused log-prob kernels available with MTP on bshd

AliceChenyy:
> **Draft.** Opened early so a user hitting this OOM can test the branch. AI assistance
> was used (see disclosure at the end); the end-to-end 128k run is still in progress and
> the numbers below state exactly what has and has not been verified.

### What does this PR do?

Enabling MTP rollout disables fused kernels outright, so the Megatron training forward
falls back to materialising a full `[s, b, vocab]` logits tensor. For Qwen3.5
(`vocab_size=248320`) at a 128k sequence that is ~30 GB per rank, doubled by the
`transpose(0, 1).contiguous()` copy. Long-context runs OOM on 96 GB cards **regardless
of world size** — each rank still processes one sequence, so adding GPUs does not help.

Three things combine to force the fallback:

1. **`mtp.enable` gates the fused path even when `mtp.enable_train=False`.**
   `patch_engine_mtp` is applied on the master switch alone (`transformer_impl.py:608`).
   Its `_postprocess` replacement accepts `output_processor` / `output_processor_context`
   in the signature but never invokes them, dropping the Megatron Core 0.18 contract that
   `GPTModel._postprocess` honours at `gpt_model.py:690`. With `mtp_num_layers` unset the
   patch does no MTP work at all, yet still costs the fused path.
2. **The fused entry point is thd-only.** `fused_forward_model_engine` calls
   `preprocess_thd_engine` unconditionally. Qwen3.5's GDN linear attention has no THD
   kernel in Megatron-LM, so the model is forced onto bshd
   (see the note in `examples/grpo_trainer/run_qwen3_5_35b_megatron.sh`).
3. **`use_remove_padding` therefore gates fused kernels off** — although
   `linear_cross_entropy` itself does `hidden.view(-1, hidden.shape[-1])` and is
   layout-agnostic. The requirement belongs to the thd *entry point*, not to the kernel.

Rather than widening the thd path, this wires the hook through the existing bshd forward:

- `mtp_patch`: invoke `output_processor` before projecting logits, mirroring
  `GPTModel._postprocess`'s argument list.
- `model_forward`: the bshd branch accepts a processor context and maps the fused output
  onto the same `{log_probs, entropy}` keys the `logits_processor` branch produces.
- `model_forward_fused`: the context carries labels and the processor falls back to them.
- `transformer_impl`: gate on hook availability rather than on `mtp.enable`; the
  value-model exclusion is unchanged.

Labels deliberately ride on the context rather than `model(labels=...)`: that argument
also drives the MTP patch's drafter branch, which must stay dormant when
`enable_train=False` — waking it hits
`MultiTokenPredictionLayer._checkpointed_forward(padding_mask=...)`, which the Megatron
build here does not accept.

### Not a duplicate

- **#7686 (`response_only_lm_head`)** targets the same memory pressure by a different
  mechanism: it reduces the *number of rows* projected via `loss_mask`, still producing
  `O(active_rows × vocab)` logits, and works on the unfused path. This PR removes the
  logits tensor entirely (`O(block)`, independent of sequence length) by unblocking the
  fused path. They are complementary — #7686 helps most when padding dominates, this PR
  also helps when generations fill the window.
- **#4539** patches `_postprocess` to add MTP *training* modes. It is a draft and does not
  address the output-processor contract; if it lands first this change reduces to keeping
  the hook call in the new structure.
- **#5332 / #7499** handle `labels=None` in the fused forward (#5273). Related but
  distinct: they fix a crash in the thd path, this PR is about the gate and the bshd path.

### Test

Numerical parity was checked on identical tensors rather than by comparing training runs
— rollout samples at `temperature=1.0`, so two runs see different data and the metric
spread is dominated by data variance, not by the computation path.

```
python3 fused_parity_test.py --seq {2048,8192,32768} --vocab 248320 --hidden 2048
```

| seq | logprob max_rel | entropy max_rel | fused peak | naive peak |
|---|---|---|---|---|
| 2,048 | 3.950e-05 | 6.142e-07 | 0.962 GB | 7.62 GB |
| 8,192 | 3.792e-05 | 6.142e-07 | 1.005 GB | 27.5 GB |
| 32,768 | 3.950e-05 | 6.910e-07 | 1.177 GB | 107.2 GB |

Relative error is ~200x below bf16 machine epsilon (7.8e-3) and does not grow with
sequence length — consistent with accumulation-order differences only; the fused kernel
accumulates in fp32. Fused peak memory is essentially the weight tensor itself
(248320 × 2048 × 2 B = 0.95 GB); the logits blocks never leave SRAM.

End-to-end on 8x H20-3e, Qwen3.5-35B-A3B, TP=2 EP=8, GRPO on DAPO-Math:

- 8k response, 2 steps, `use_fused_kernels=True`: completes, no gate warning.
- **128k response: run in progress; this PR will not be marked ready until it is reported.**

### Caveats

- Only the bshd path is touched; thd behaviour is unchanged.
- `mtp.enable_train=True` is untested here.
- Verified against Megatron Core 0.18.0; the gate falls back with a warning when the hook
  is absent.

### AI assistance disclosure

Investigation, patch and tests were produced with AI assistance (Claude). The failure
analysis above was derived from reading the code paths and confirmed by the measurements
shown. Reviewers should treat the 128k end-to-end claim as pending.

### DIFF
--- a/verl/models/mcore/model_forward.py
+++ b/verl/models/mcore/model_forward.py
@@ -278,8 +278,19 @@ def gptmodel_forward_model_engine(
     cp_layout: str = "zigzag",
     router_padding_mask: torch.Tensor | None = None,
     mtp_loss_normalization_factor: float | None = None,
+    fused_output_processor_context=None,
 ):
-    """Default forward pass for GPT models with optional sequence packing."""
+    """Default forward pass for GPT models with optional sequence packing.
+
+    ``fused_output_processor_context`` opts the bshd path into Megatron Core
+    0.18's output-processor hook, so log-probs and entropy are computed straight
+    from hidden states.  Without it the model returns a ``[b, s, vocab]`` logits
+    tensor that the caller then reduces -- for a 248k vocab at 128k sequence that
+    single tensor is ~30 GB per rank (doubled again when entropy forces a clone),
+    which is what makes long-context training OOM.  Only bshd needs this: the thd
+    path already has ``fused_forward_model_engine``, but sequence packing is
+    unavailable for models whose attention (e.g. Qwen3.5 GDN) has no THD kernel.
+    """
 
     assert data_format in ["thd", "bshd"], "data_format must be 'thd' or 'bshd'"
     pre_process = unwrap_model(model).pre_process
@@ -450,13 +461,52 @@ def gptmodel_forward_model_engine(
         else:
             attention_mask = attention_mask_bshd
 
+        use_fused_output_processor = fused_output_processor_context is not None and post_process
+        if use_fused_output_processor:
+            from verl.models.mcore.model_forward_fused import fused_output_processor
+
+            # The hook needs labels, but they must NOT go in as ``model_kwargs["labels"]``:
+            # GPTModel forwards that straight into _postprocess, where the MTP patch
+            # gates its drafter forward on ``labels is not None``.  With
+            # mtp.enable_train=False that branch is meant to stay dormant, and waking
+            # it hits MultiTokenPredictionLayer._checkpointed_forward(padding_mask=...),
+            # which older Megatron builds do not accept.  Carry them on the context
+            # instead; mtp_patch reads them from there when invoking the processor.
+            fused_output_processor_context.labels = preprocess_bshd_engine(
+                logits_processor_args["label"],
+                pre_process=True,
+                need_roll=True,
+                use_fp8_padding=use_fp8_padding,
+                forced_max_seqlen=forced_max_seqlen,
+            )[0].contiguous()
+            model_kwargs["output_processor"] = fused_output_processor
+            model_kwargs["output_processor_context"] = fused_output_processor_context
+
         output_orig = model(
             input_ids=input_ids_bshd,
             attention_mask=attention_mask,
             position_ids=None if vision_model else position_ids_bshd,
             **model_kwargs,
         )
-        if post_process and logits_processor is not None:
+        if use_fused_output_processor:
+            # output_orig is a CausalLMOutputForPPO; mirror the key names the
+            # logits_processor branch produces so downstream code is unchanged.
+            # linear_cross_entropy returns flat (num_tokens,) tensors, while the
+            # bshd postprocess asserts output.shape[:2] == attention_mask.shape,
+            # so restore the [b, s] layout the rest of the path expects.
+            bshd_shape = input_ids_bshd.shape
+
+            def _to_bshd(t):
+                return t.view(bshd_shape) if t.dim() == 1 else t
+
+            output_dict = {"log_probs": _to_bshd(output_orig.log_probs)}
+            if output_orig.entropy is not None:
+                output_dict["entropy"] = _to_bshd(output_orig.entropy)
+            output = {
+                k: postprocess_bshd_engine(v, attention_mask_bshd, post_process=post_process)
+                for k, v in output_dict.items()
+            }
+        elif post_process and logits_processor is not None:
             args = {
                 k: preprocess_bshd_engine(
                     v,
--- a/verl/models/mcore/model_forward_fused.py
+++ b/verl/models/mcore/model_forward_fused.py
@@ -17,7 +17,7 @@
 import inspect
 from collections import OrderedDict
 from dataclasses import dataclass
-from typing import Optional
+from typing import Any, Optional
 
 import megatron.core as mcore
 import torch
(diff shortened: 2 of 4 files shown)


AliceChenyy:
### Update: end-to-end memory measurement, and a correction to the numbers above

**Correction first.** The 91x figure in the table above is a *kernel-isolated* comparison
(`linear_cross_entropy` vs a naive `matmul → log_softmax → gather` reference on identical
tensors). It overstates what this PR delivers in a real run: the naive reference also
materialises fp32 `log_softmax` and `probs` copies that the production path does not.
The kernel numbers stand as evidence for numerical parity, but they are not the patch's
end-to-end benefit. The measurement below is.

**End-to-end, same machine, same config, only `use_fused_kernels` differs**
(8x H20-3e, Qwen3.5-35B-A3B, TP=2 EP=8, GRPO on DAPO-Math, 1k prompt + 8k response,
`ppo_micro_batch_size_per_gpu=1`, from `perf/max_memory_allocated_gb`):

| | baseline | with this PR | delta |
|---|---|---|---|
| step 1 | 42.315 GB | 40.257 GB | |
| step 2 | 42.383 GB | 40.257 GB | |
| **mean** | **42.349 GB** | **40.257 GB** | **-2.09 GB** |

Predicted saving for that shape is one logits tensor after vocab sharding:
`(1024 + 8192) x 248320 x 2 B / TP(2) = 2.13 GB`. Measured 2.09 GB — **98% of theory**,
so the patch removes essentially exactly the tensor it targets.

Two details worth noting: peak memory is *identical* across steps with the patch
(40.25712585449219 both times) but drifts on baseline, consistent with the peak no longer
being set by a sequence-dependent logits allocation. And this comparison is valid even
though the arms sample different rollouts — bshd pads to the batch's longest sequence, so
both arms pad to the same configured ceiling here; unlike the loss metrics, memory is not
confounded by sampling variance.

**Projected saving** (same formula, TP=2, `vocab=248320`):

| response length | per-GPU sequence | expected saving |
|---|---|---|
| 8k | 9,216 | 2.13 GB (**measured 2.09**) |
| 32k | 33,792 | 7.8 GB |
| 64k | 66,560 | 15.4 GB |
| 128k | 132,096 | **30.6 GB** |

This is deliberately the conservative model — it assumes one logits tensor is eliminated.
The current path also does `logits.transpose(0, 1).contiguous()`, so if both copies are
live at peak the real saving is up to 2x that. The 8k measurement is consistent with one
copy, so that is what the table reports.

**Still outstanding:** the 128k end-to-end run. Direct generation at that length is slow,
so it is being re-run with the sequence length supplied as a ~120k-token prompt instead —
the training-side logits cost depends on total sequence length, not on how the tokens were
produced. This PR stays in draft until that lands.

AliceChenyy:
### Long-context end-to-end: baseline OOMs, patched run completes

Same machine, same config, same data — `use_fused_kernels` is the only variable.
8x H20-3e (140 GB), Qwen3.5-35B-A3B, TP=2 EP=8, GRPO, `ppo_micro_batch_size_per_gpu=1`.

Generating 128k tokens per sample is slow, so the sequence length is supplied as a
~120k-token prompt with a short response instead. The training-side logits cost depends on
total sequence length, not on how the tokens were produced, and bshd pads to the batch's
longest sequence either way — so this exercises the same allocation.

| | baseline | with this PR |
|---|---|---|
| outcome | **CUDA OOM (x5)** | **step completes** |
| failing frame | `transformer_impl.py:1224` `_lm_head_logits_processor` | — |
| attempted allocation | **55.64 – 55.68 GiB** | never made |
| `perf/max_memory_allocated_gb` | — (crashed) | **77.74 GB** |
| `perf/total_num_tokens` | — | 962,244 |

```
torch.OutOfMemoryError: CUDA out of memory. Tried to allocate 55.68 GiB.
GPU 0 has a total capacity of 139.80 GiB of which 52.16 GiB is free.
  File "verl/workers/engine/megatron/transformer_impl.py", line 1224,
    in _lm_head_logits_processor
```

**The allocation size confirms the mechanism.** At ~120,300 tokens with vocab 248320 in
bf16, one logits tensor after TP=2 vocab sharding is `120300 x 248320 x 2 / 2 = 27.8 GB`.
The attempted 55.65 GiB is **2.00x** that — the second copy being `logits_bak =
logits.clone()` in `_lm_head_logits_processor`, taken when entropy is requested.

**This revises the projection I posted earlier upward.** That table assumed one logits
tensor was eliminated and put 128k at ~30 GB; the 8k measurement was consistent with one
copy, but at this length both copies are clearly live at peak. The realistic saving at
128k is therefore **~56 GB per rank**, not ~30 GB.

Worth noting for anyone hitting this on smaller cards: two 27.8 GB copies plus the rest of
the step means the unpatched path needs well over 96 GB at this sequence length, and it
was already tight at 32k. Shortening `max_response_length` moderately does not necessarily
clear it.

With this, the draft's outstanding item is closed. Remaining caveats are unchanged: only
the bshd path is touched, `mtp.enable_train=True` is untested, and verification is against
Megatron Core 0.18.0.

PR #7843, opened 2026-09-11
Add DeepSeek v4.1 Sample-Level Style ReplayBuffer

attack204:
Currently, it is quite difficult to reproduce the phenomenon that the batch-level scheduler mentioned in the paper causes training jitter. I am trying to increase the batch_size to reproduce it

Model: DPSK-V4-0731
Task: SWE-RL, response_length = 30k, nums_turns = 30, GRPO.n=8
ENV: 4 * 8 H20

- BS = 32, GRPO.n = 8

I still cannot reproduce the oscillation mentioned in the dpsk v4.1 paper

**Based on my observation of the inference kernel, I believe that the rollout batch_size for dpsk v4.1 should be greater than or equal to 256. Currently, I do not have sufficient resources to run such a large-scale experiment. Therefore, if you happen to see this PR, you can continue to increase the batch size to try to reproduce the results from the dpsk paper**

image

- BS = 8, GRPO.n = 8

Purple: Sample-Level Dispatcher
Green: Batch-Level Dispatcher

image


Begunner:
We might need a more compatible implementation to support the three trainers at the same time, like a background thread.

PR #7849, opened 2026-09-11
[single_controller] feat: discover TPU resource in Ray

askhat-g:
### What does this PR do?

Makes Ray resource discovery and worker placement ask the active platform instead of
assuming NVIDIA, and fixes two places that assume a shape Ray does not guarantee.

This is the in-tree half of out-of-tree platform support. `PlatformTPU` itself lives in
[verl-hardware-plugin#18](https://github.com/verl-project/verl-hardware-plugin/pull/18)
(merged); until the call sites here land, that plugin is inert because verl never consults it.

### Checklist Before Starting

- [x] Search for similar PRs: https://github.com/verl-project/verl/pulls?q=is%3Apr+single_controller+platform+resource

### Test

`tests/single_controller/test_device_agnostic_resources_on_cpu.py` (new, 14 cases) covers both
bug fixes and the colocation cap with a stub platform. Runs on CPU, no accelerator and no Ray
cluster required. Eight of the fourteen fail against current `main`; the other six are
regression guards for existing GPU/NPU behaviour and pass either way.

```
pytest tests/single_controller/test_device_agnostic_resources_on_cpu.py   # 14 passed
ruff check verl/ && ruff format --check verl/                             # clean
python tests/special_sanity/check_license.py --directories verl tests     # exit 0
python tests/special_sanity/check_device_api_usage.py --directory verl    # exit 0
```

Broader `tests/single_controller/` + `tests/plugin/` failure set is byte-identical before and
after this change. Cross-checked against the merged plugin with `VERL_PLATFORM=tpu`: the
colocation cap applies and the TPU branch resolves.

### API and Usage Example

No user-facing API change. Two new `PlatformBase` hooks, both defaulting to current behaviour:

```python
class PlatformBase:
    def supports_colocated_worker_groups(self) -> bool:
        return True   # False caps colocation at one WorkerGroup

    def ray_local_rank_override(self) -> Optional[str]:
        return None   # None keeps Ray's get_accelerator_ids() path
```

### Design & Code Changes

**New hooks**

- `supports_colocated_worker_groups()` — applied in both `RayResourcePool` and
  `ResourcePoolManager` so the bundle CPU count matches the pool actually created. Left at 3
  for an exclusive device, the placement group requests 3 CPUs per chip and never schedules.
- `ray_local_rank_override()` — for platforms whose devices Ray does not enumerate via
  `get_accelerator_ids()`.

`get_worker_env_vars` was previously a third hook. Per review it is no longer on
`PlatformBase`: `RayWorkerGroup` calls it behind an explicit `device_name == "tpu"` branch, so
the only platform needing placement-group-derived mesh environment is the only one carrying it.

**Deferred, not silently dropped**

- `is_tpu_available` in `verl/utils/device.py` — unshippable out-of-tree. `verl/__init__.py`
  imports `verl.utils.device` at line 33, freezing the module-level flag; entry-point plugins
  load at line 74, so `torch_tpu` is imported strictly after the flag is evaluated and the flag
  is always `False`, including on a real TPU host.
- The TPU early return in `set_numa_affinity` — belongs behind a vendor-neutral
  `supports_numa_affinity()` hook paired with a plugin change. The NVML failure it avoided is
  already caught and non-fatal.

**Docs:** `docs/hardware/multi_chip_support.rst` moves TPU from the built-in list to the
verl-hardware-plugin list.

- [x] Apply pre-commit checks — `pre-commit` not available in this environment; `ruff check`
      and `ruff format --check` run directly and clean, plus the `tests/special_sanity/` hooks.
- [x] Add / Update the documentation.
- [x] Add unit test(s) — `tests/single_controller/test_device_agnostic_resources_on_cpu.py`,
      CPU-only so it runs in existing CI.

### DIFF
--- a/verl/plugin/platform/platform_base.py
+++ b/verl/plugin/platform/platform_base.py
@@ -230,6 +230,37 @@ class PlatformBase(abc.ABC):
             return {"num_gpus": num_gpus}
         return {"resources": {resource_name: num_gpus}}
 
+    def supports_colocated_worker_groups(self) -> bool:
+        """Whether several WorkerGroups can share one accelerator.
+
+        ``False`` where a device is claimed exclusively by a single process, which caps
+        colocation at one WorkerGroup.
+        """
+        return True
+
+    def get_worker_env_vars(
+        self,
+        resource_pool,
+        rank: int,
+        world_size: int,
+        local_rank: int,
+        local_world_size: int,
+        name_prefix: str,
+        device_name: str,
+    ) -> dict[str, str]:
+        """Extra env vars for a Ray worker, called once per worker as the WorkerGroup is built.
+
+        Overridden by platforms that need topology or mesh information from the placement groups.
+        """
+        return {}
+
+    def ray_local_rank_override(self) -> Optional[str]:
+        """This worker's local rank, or ``None`` to use Ray's accelerator IDs.
+
+        For platforms whose devices Ray does not enumerate via ``get_accelerator_ids()``.
+        """
+        return None
+
     # ------------------------------------------------------------------
     # IPC support
     # ------------------------------------------------------------------
--- a/verl/plugin/platform/platform_tpu.py
+++ b/verl/plugin/platform/platform_tpu.py
@@ -1,0 +1,433 @@
+# Copyright (c) 2026 BAAI. All rights reserved.
+# Copyright (c) 2026 Google LLC. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Google TPU platform implementation.
+
+**Experimental: TPU support is under development and cannot run training end to end yet.**
+This module registers the platform and provides the device abstraction plus the Ray resource
+and worker-environment plumbing. The training engine, the generation engine and the
+weight-transfer path are still to come.
+
+``torch_tpu`` reuses the NVIDIA device API surface, so most of ``PlatformCUDA`` works
+unchanged. This class subclasses it and overrides device-specific environment configuration,
+resource options, and memory management proxies.
+"""
+
+import logging
+import os
+from typing import Any, Optional
+
+import ray
+import torch
+
+from .platform_cuda import PlatformCUDA
+from .platform_manager import PlatformRegistry
+
+logger = logging.getLogger(__file__)
+logger.setLevel(os.getenv("VERL_LOGGING_LEVEL", "WARN"))
+
+
+def _ensure_torch_tpu() -> bool:
+    """Try to import torch_tpu, which registers the ``tpu_dist`` distributed backend.
+
+    Returns True if the TPU runtime is usable after the attempt.
+    """
+    if hasattr(torch, "tpu"):
+        return True
+    try:
+        import torch_tpu  # noqa: F401
+
+        return hasattr(torch, "tpu")
+    except Exception as e:
+        logger.debug("The current machine has no torch.tpu, because: %s", e)
+    return False
+
+
+_ensure_torch_tpu()  # Attempt at module load time so availability checks are faster later
+
+# Base port for the TPU distributed slice builder mesh. Each local rank takes ``base + local_rank``.
+TPU_PROCESS_BASE_PORT = 8471
+
+# TPU chip HBM capacities in bytes
+HBM_BYTES_TPU_V5P = 95 * 1024 * 1024 * 1024  # 95 GB
+HBM_BYTES_TPU_V6E = 32 * 1024 * 1024 * 1024  # 32 GB
+HBM_BYTES_TPU_V7X = 192 * 1024 * 1024 * 1024  # 192 GB
+
+TPU_HBM_BYTES_MAP = {
+    "v5p": HBM_BYTES_TPU_V5P,
+    "v6e": HBM_BYTES_TPU_V6E,
+    "v7x": HBM_BYTES_TPU_V7X,
+}
+
+# Fallback HBM capacity when the chip generation cannot be determined.
+HBM_BYTES_TPU_DEFAULT = HBM_BYTES_TPU_V6E
+
+# TPU default 3D mesh topology mappings by pod type or total chips
+TPU_TOPOLOGY_MAP = {
+    "v6e-32": "4,8,1",
+    "v6e-8": "2,4,1",
+    "v6e-4": "2,2,1",
+    32: "4,8,1",
+    8: "2,4,1",
+    4: "2,2,1",
+}
+
+
+def get_tpu_chip_hbm_bytes() -> int:
+    """Detects the TPU chip generation from Ray node labels or environment variables and returns its HBM capacity."""
+    tpu_type = ""
+
+    # Query Ray cluster node labels for TPU resource type
+    try:
+        if ray.is_initialized():
+            tpu_nodes = [node for node in ray.nodes() if "TPU" in node.get("Resources", {}) and node.get("Alive")]
+            if tpu_nodes:
+                labels = tpu_nodes[0].get("Labels", {})
+                tpu_type = (labels.get("ray.io/accelerator-type") or labels.get("ray.io/tpu-pod-type") or "").lower()
+    except Exception as e:
+        logger.warning(f"Unable to query Ray node labels for TPU chip type: {e}")
+
+    # Fallback to environment variables
+    if not tpu_type:
+        tpu_type = (
+            os.environ.get("TPU_ACCELERATOR_TYPE")
+            or os.environ.get("ACCELERATOR_TYPE")
+            or os.environ.get("TPU_TYPE")
+            or ""
+        ).lower()
+
+    for chip_gen, hbm_bytes in TPU_HBM_BYTES_MAP.items():
+        if chip_gen in tpu_type:
+            return hbm_bytes
+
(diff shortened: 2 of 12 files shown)


verl/plugin/platform/platform_tpu.py
```diff
@@ -0,0 +1,433 @@
+# Copyright (c) 2026 BAAI. All rights reserved.
```
vlasenkoalexey: I'm not sure what Copyright (c) 2026 BAAI. All rights reserved. here means, previous new files were attributed to ByDance. 

Should all new added files be attributed to Google?

verl/plugin/platform/platform_tpu.py
```diff
@@ -0,0 +1,433 @@
...
+logger = logging.getLogger(__file__)
+logger.setLevel(os.getenv("VERL_LOGGING_LEVEL", "WARN"))
+
+
+def _ensure_torch_tpu() -> bool:
+    """Try to import torch_tpu, which registers the ``tpu_dist`` distributed backend.
+
+    Returns True if the TPU runtime is usable after the attempt.
+    """
+    if hasattr(torch, "tpu"):
```
vlasenkoalexey, line 45: should it be if not hasattr ... return false?

verl/plugin/platform/platform_tpu.py
```diff
@@ -0,0 +1,433 @@
...
+        return True
+    try:
+        import torch_tpu  # noqa: F401
+
+        return hasattr(torch, "tpu")
```
vlasenkoalexey, line 50: should this be torch.tpu.is_available() instead or you want to import torch_tpu explicitly to trigger distributed init?

verl/plugin/platform/platform_tpu.py
```diff
@@ -0,0 +1,433 @@
...
+            "CLOUD_TPU_TASK_ID": str(rank // local_world_size),
+            "TPU_WORKER_HOSTNAMES": ",".join(unique_hostnames),
+            "TPU_VISIBLE_CHIPS": str(local_rank),
+        }
+
+        # Apply TPU topology and host bounds based on TPU pod type or world size
+        tpu_nodes = [node for node in ray.nodes() if "TPU" in node.get("Resources", {}) and node.get("Alive")]
+        tpu_type = tpu_nodes[0].get("Labels", {}).get("ray.io/tpu-pod-type", "") if tpu_nodes else ""
+
+        topo = TPU_TOPOLOGY_MAP.get(tpu_type, TPU_TOPOLOGY_MAP.get(world_size, "1,1,1"))
```
vlasenkoalexey, line 393: can we read it from TORCH_TPU_TOPOLOGY (it will be different from TPU_TOPOLOGY, but I think it is guaranteed to be set for distributed runs, and should be the right one for environment to use

verl/plugin/platform/platform_tpu.py
askhat-g: Ah, it should be Google. This slipped the check. Thanks for pointing out

wuxibin89:
@askhat-g We prefer to support verl TPU integration as out-of-tree in [verl-project/verl-hardware-plugin](https://github.com/verl-project/verl-hardware-plugin), once it matures we can move it as in-tree.

askhat-g:
> @askhat-g We prefer to support verl TPU integration as out-of-tree in [verl-project/verl-hardware-plugin](https://github.com/verl-project/verl-hardware-plugin), once it matures we can move it as in-tree.

Thank you @wuxibin89 , started a new PR here https://github.com/verl-project/verl-hardware-plugin/pull/18

wuxibin89:
> > @askhat-g We prefer to support verl TPU integration as out-of-tree in [verl-project/verl-hardware-plugin](https://github.com/verl-project/verl-hardware-plugin), once it matures we can move it as in-tree.
> 
> Thank you @wuxibin89 , started a new PR here [verl-project/verl-hardware-plugin#18](https://github.com/verl-project/verl-hardware-plugin/pull/18)

@askhat-g https://github.com/verl-project/verl-hardware-plugin/pull/18 merged, please strip `platform_tpu.py` and keep single-controller changes in this PR.

verl/plugin/platform/platform_base.py
```diff
@@ -230,6 +230,37 @@ def ray_resource_options(self, num_gpus: float) -> dict[str, Any]:
...
 
+    def supports_colocated_worker_groups(self) -> bool:
+        """Whether several WorkerGroups can share one accelerator.
+
+        ``False`` where a device is claimed exclusively by a single process, which caps
+        colocation at one WorkerGroup.
+        """
+        return True
+
+    def get_worker_env_vars(
```
jialei777: should we keep this method only for tpu platform? It is actually a bit confusing for platforms other than TPU

verl/plugin/platform/platform_tpu.py
```diff
@@ -0,0 +1,432 @@
...
+
+# TPU chip HBM capacities in bytes
+HBM_BYTES_TPU_V5P = 95 * 1024 * 1024 * 1024  # 95 GB
+HBM_BYTES_TPU_V6E = 32 * 1024 * 1024 * 1024  # 32 GB
+HBM_BYTES_TPU_V7X = 192 * 1024 * 1024 * 1024  # 192 GB
+
+TPU_HBM_BYTES_MAP = {
+    "v5p": HBM_BYTES_TPU_V5P,
+    "v6e": HBM_BYTES_TPU_V6E,
+    "v7x": HBM_BYTES_TPU_V7X,
```
jialei777: just be clear: we have only tested v6e right? maybe have a comment mentioning that.

verl/plugin/platform/platform_tpu.py
jialei777: +1 

please check https://github.com/google-pytorch/torchtitan/blob/main/torchtitan/experiments/tpu/cloud_distributed_utils.py for env var setup

verl/single_controller/base/worker.py
```diff
@@ -276,8 +277,10 @@ def _setup_env_cuda_visible_devices(self):
             # RAY_EXPERIMENTAL_NOSET_*_VISIBLE_DEVICES is set,
             # so we need to set local rank when the flag is set.
             device_name = get_resource_name()
-            local_rank = ray.get_runtime_context().get_accelerator_ids()[device_name][0]
-            os.environ["LOCAL_RANK"] = local_rank
+            local_rank = get_platform().ray_local_rank_override()
```
jialei777: This is actually i a bit confusing to read. Can we add a condition of get_platform().device_name == "TPU" and also add a TODO on have ray fix it? cc: @andrewsykim

verl/single_controller/ray/base.py
```diff
@@ -120,6 +122,10 @@ def __init__(
         detached=False,
         accelerator_type: Optional[str] = None,
     ) -> None:
+        # Where a device is claimed exclusively, asking to colocate more than one WorkerGroup
```
jialei777: be explicit that this is the case for TPU in the comment?

verl/single_controller/base/worker.py
askhat-g: Yes, this happens because ranks seem to be treated different in the case of TPU. I wanted to avoid single out the tpu case(if platform == 'TPU') and tried generalize it by overriding the local rank for the tpu case. Let me take it offline with Andrew and discuss if there are better solutions to this

verl/plugin/platform/platform_tpu.py
askhat-g: Yeah but we will support this eventually. I will add the comment

verl/plugin/platform/platform_tpu.py
askhat-g: The latter is true, it is not about chip availability

verl/plugin/platform/platform_tpu.py
askhat-g: Hm, if it is not imported then we try to import it below, that would be a return prematurely

verl/plugin/platform/platform_tpu.py
askhat-g: Good to know, let me follow up on this, thank you for pointing that out

verl/single_controller/base/worker.py
andrewsykim: @ryanaoleary do you know if https://github.com/ray-project/ray/pull/65723 fixes this or is this a separate issue?

andrewsykim: @askhat-g did you run this with `RAY_EXPERIMENTAL_NOSET_TPU_VISIBLE_CHIPS=1`?

andrewsykim: I don't think you need to do this, I suspect it's because we were previously testing with RAY_EXPERIMENTAL_NOSET_TPU_VISIBLE_CHIPS on TPU v6e. Can you try running without `RAY_EXPERIMENTAL_NOSET_TPU_VISIBLE_CHIPS`?

askhat-g: It still didn't work but I have removed it from this PR. We can align internally if it is needed

jialei777 (approved):

vlasenkoalexey (approved):

verl/single_controller/base/worker.py
ryanaoleary: talked offline with @andrewsykim and believe it's a separate issue. The linked Ray PR fixes an issue with dual-device TPUs (tpu7x) by changing `get_current_process_visible_accelerator_ids()`  to expand physical chip indices (0..3 from GKE's injected `TPU_VISIBLE_CHIP`S) into logical device IDs (0..7). Previously it would cap the `TPU` Ray resource on the node with the length of the accelerator Ids, which returned the chip IDs. To handle dual-device TPUs you now set `RAY_TPU_RESOURCE_PER_CHIP=2 ` to ensure the translation occurs correctly without breaking default behavior.

I think running without `RAY_EXPERIMENTAL_NOSET_TPU_VISIBLE_CHIPS=1` fails here because when it's not set Ray rewrites `TPU_HOST_BOUNDS` based on the logical resources it sees. The verl actors each request `resources={"TPU": 1}`, so it overwrites `TPU_HOST_BOUNDS` to `1,1,1` (https://github.com/ray-project/ray/blob/8e779d8152f15ec7d91a1a8fc8b6f10a3d3d839d/python/ray/_private/accelerators/tpu.py#L847), overwriting the host bounds set by PlatformTPU.

wuxibin89 (approved):

PR #7806, opened 2026-09-09
[algo, trainer, rollout, vllm] feat: add reverse KL top-k with fixed-token prefill scoring

YichenZW:
### What does this PR do?

This PR adds `reverse_kl_topk` for on-policy distillation using the rollout student's top-k token support.

Instead of transferring full-vocabulary teacher log probabilities, the implementation:

- Collects response-aligned student top-k token IDs during rollout.
- Builds a request-wise union of those token IDs.
- Requests teacher prefill log probabilities only for that fixed token set through vLLM.
- Gathers the returned scores back into the per-token student top-k support.
- Aligns the sparse teacher outputs with the causal source rows consumed by the training loss.
- Computes truncated reverse KL on the student top-k support in the FSDP and VeOmni training paths.
- Supports optional chunking when computing student top-k log probabilities to reduce peak memory usage.

This PR supersedes #7590 and addresses its reviewer feedback by replacing the full-vocabulary teacher payload with fixed-token prefill scoring.

Related to #6676. This PR implements the non-renormalized truncated reverse-KL variant and therefore does not close the renormalized objective proposed in that issue.

> [!IMPORTANT]
> This PR temporarily depends on vLLM PR https://github.com/vllm-project/vllm/pull/54335.
>
> The current implementation was validated against vLLM commit `3f2b51e348190bb3382962a013860516db25bce8`. Until that PR is merged and the corresponding API is available in verl's supported vLLM environment, tests and experiments must use a vLLM wheel built from that exact commit.
>
> Stock vLLM currently does not expose the required fixed-token prefill scoring fields. The verl vLLM adapter performs an explicit feature check and fails with a clear error when the API is unavailable.
>
> Before this PR is merged, the implementation should be synchronized with the final API merged by vLLM PR #54335, and this dependency note should be replaced with the minimum supported vLLM commit or release.

### Checklist Before Starting

- [x] Search for similar PRs. Query: https://github.com/verl-project/verl/pulls?q=is%3Apr+reverse+kl+distillation+topk
- [x] Format the PR title as `[{modules}] {type}: {description}`.

### Test

Local-only CPU validation tests were run against the clean branch.

The configured pre-commit hooks passed for all files changed by this commit, including Ruff, Ruff format, mypy, configuration generation checks, documentation checks, license checks, device API checks, repository structure checks, and Python compilation.

End-to-end GPU validation used:

- Qwen3-0.6B student.
- Qwen3-1.7B teacher.
- GSM8K.
[...]

PR #7733, opened 2026-09-04
[megatron] feat: add memory-efficient chunked top-k KL computation

HSYZhang:
### What does this PR do?

Adds an opt-in, memory-efficient chunked path for Megatron vocab-parallel top-k forward KL. It extends the FSDP chunking approach from #6593 to tensor-parallel vocabulary shards, avoiding full fp32 `[N, V_local]` log-softmax/probability buffers while preserving the existing outputs and gradients.

With chunked loss implemented, small student models (e.g., 4B or smaller) can reduce GPU memory via an FSDP-like chunked loss approach, eliminating the need to increase TP to lower peak memory usage. This prevents GPU idleness and significantly improves utilization. We previously used a 4B student model, but the trainer OOM'd during loss computation. The original approach required TP=4 to avoid OOM; with chunked loss enabled, we can now run at TP=1 without OOM, greatly improving GPU utilization and reducing idle time.

### Test

Results (b=2, s=96, V=2048, K=32; d_* denotes the maximum absolute error):

| Config | Comparison | d_loss | d_grad | d_smass | d_ocount | Verdict |
|---|---|---|---|---|---|---|
| fp32, clamp=-10 | baseline vs **reference** | 8.9e-8 | **1.2e-9** | — | — | ✅ |
| fp32, clamp=-10 | chunked(cs=7/64/4096) vs baseline | ≤1.2e-7 | ≤1.3e-9 | 1.9e-8 | 0 | ✅ |
| fp32, clamp=None | baseline vs reference | 8.9e-8 | 1.2e-9 | — | — | ✅ |
| fp32, clamp=None | chunked vs baseline | ≤1.2e-7 | ≤1.3e-9 | 1.9e-8 | 0 | ✅ |
| bf16, clamp=-10 | chunked vs baseline | ≤8.9e-8 | **1.9e-6** | 1.9e-8 | 0 | ✅ |

- Under fp32, chunked and baseline are nearly bit-for-bit identical (same floating-point operators and reduction order when TP=1);
- overlap_count is exactly identical (diff=0);
- For bf16 (storing bf16 logits and recomputing p in backward), the additional rounding is only ~1.9e-6;
- chunk_size=7 (many odd-sized chunks + a residual chunk) produces results consistent with 4096 → chunk boundary handling is correct.

Actual Qwen3.5-4B distillation dimensions: V=248320, K=128, TP=1, N=4096 tokens, clamp=-10.
Individually compared the maximum absolute/relative differences across 5 forward outputs + gradients.

[...]

PR #7590, opened 2026-08-28
[trainer] feat: add reverse KL top-k distillation loss

YichenZW:
### What does this PR do?

Adds a reverse-KL top-k distillation loss for on-policy distillation. The teacher provides full-vocab prompt log-probs, while the student loss is evaluated on the student top-k support.

### Checklist Before Starting

- [x] Search for similar PRs. Query: https://github.com/verl-project/verl/pulls?q=reverse+kl+distillation+topk
- [x] Format the PR title as `[{modules}] {type}: {description}`

### Test

- `git diff --check upstream/main...HEAD`
- `PR_TITLE='[trainer] feat: add reverse KL top-k distillation loss' uv run --frozen --all-packages --extra cpu python tests/special_sanity/check_pr_title.py`
- `uv run --frozen --all-packages --extra cpu pytest -q tests/special_sanity/check_device_api_usage.py tests/workers/test_megatron_distillation_only_on_cpu.py tests/workers/test_distillation_topk_symmetry_on_cpu.py` (9 passed)
- Standalone Open-Verl smoke setup was prepared using `examples/on_policy_distillation_trainer/run_qwen3_0.6b_opd_veomni.sh` with a 1-step Qwen3 0.6B/1.7B reverse-KL top-k run.

### API and Usage Example

`distillation.distillation_loss.loss_mode=reverse_kl_topk` with `distillation.distillation_loss.use_policy_gradient=False`.

### Design & Code Changes

- Adds reverse-KL top-k loss metadata and FSDP implementation.
- Requests full-vocab teacher prompt log-probs when reverse-KL top-k needs teacher probabilities on the student top-k support.
- Wires distillation top-k metadata through PPO trainer paths and VeOmni transformer implementation.
- Documents the reverse-KL top-k mode.

### Checklist Before Submitting

- [x] Read the Contribute Guide.
- [x] Apply pre-commit checks.
- [x] Add / Update documentation.
- [x] Add unit or end-to-end test(s), or explain why not feasible.
- [ ] Once ready for CI, request CI in the project channel if required.

Issue #7239, opened 2026-08-03
[RPC] Full-Vocabulary Multi-Teacher On-Policy Distillation


Moocharr:
### 1. Motivation

verl's OPD losses split into two families :

- `forward_kl_topk` (`use_topk=True`): approximates forward KL using the teacher's top-k log-probs (vLLM `prompt_logprobs=k`). Truncated on large vocabs → tail-mass bias, and needs `clamp_min(0)` since top-k isn't normalized.

- Estimators `kl/k1/k3/...` (`use_estimator=True`): single-sample reverse-KL over the student's sampled token only. Unbiased but mode-seeking only — no mode-covering forward KL.

Neither materializes the full vocabulary. We add `forward_kl_fullvocab`: exact forward KL over the full vocab, no truncation bias, for mode-covering supervision where tail mass matters.

### 2. Design

**Loss.** $\mathcal{L}_t = \sum_v \mathrm{softmax}(T_t)_v [\log\pi_T(v) - \log\pi_S(v)]$, fp32, normalized → non-negative (drop `clamp_min(0)`). Supervised backprop, `use_policy_gradient=False`, teacher under `stop_grad`.

**Registration.** Add `use_fullvocab` to `DistillationLossSettings`; relax `__post_init__` from "one of two" to "one of three". Reuse `loss_max_clamp`/`log_prob_min_clamp`; rename `use_chunked_topk`→`use_chunked_fullvocab` (alias old name).

### 3.  Contribution

We plan to submit a PR to complete this feature.

Monaroj:
In the current verl implementation, the teacher model has its own dedicated resource pool (separate GPU). I'm wondering how the full-vocabulary logits will be transferred from the teacher to the student during distillation.

Is the plan to:

Use some kind of inter-process communication (e.g., shared memory, NCCL, or TCP) to pass the logits between the teacher and student pools?
Or refactor the architecture to co-locate the teacher and student (e.g., placing them on the same GPU nodes to enable cuda ipc transfer)?

Any clarification on the intended design would be greatly appreciated.

PR #7855, opened 2026-09-13
[algo, fsdp, megatron, veomni] feat: add tail-aware top-k KL for on-policy distillation

PengyuLi7:
### What does this PR do?

Adds a `forward_kl_topk_tail` loss mode for on-policy distillation.

The existing `forward_kl_topk` objective retains only the teacher's top-k terms:

```
L_truncated = Σ_{i∈T} p_i (log p_i − log q_i)
```

Because `T` is truncated, the retained masses `P = Σ_{i∈T} p_i` and `Q = Σ_{i∈T} q_i` are both below 1. The quantity is therefore not a KL and goes negative whenever `Q > P`; the current implementation clamps it with `clamp_min(0)`, which silently discards the gradient carried by those tokens. In our runs the per-token minimum was exactly `0.0` (clamping active) in **415 / 467 training steps (89 %)**.

This PR treats every token outside `T` as one aggregate category:

```
p_tail = 1 − P,  q_tail = 1 − Q
L_tail = Σ_{i∈T} p_i log(p_i / q_i) + p_tail log(p_tail / q_tail)
```

which is the exact KL between the two `K+1`-category distributions: non-negative by construction (no clamp), a lower bound on the full forward KL, and it reuses the existing teacher top-k payload, so teacher compute, schema and network traffic are unchanged.

**The claim of this PR is training efficiency**: at an equal training-step budget, `forward_kl_topk_tail` reaches a higher GSM8K accuracy than `forward_kl_topk`; equivalently, it reaches a given accuracy in fewer steps. It is not a claim about a higher final ceiling — the two curves approach the same asymptote and the gap narrows with longer training, which is the expected shape of faster convergence.

Related: #7239 / #7375 (full-vocabulary OPD), #6676 / #7590 / #7806 (reverse-KL on the student top-k support), #7733 / #7809 (memory/perf work on the existing top-k path). None of them change the truncated forward-KL objective in the way this PR does.

### Checklist Before Starting

- [x] Search for similar PRs. Query links:
  - https://github.com/verl-project/verl/pulls?q=is%3Apr+is%3Aopen+distillation+kl
  - https://github.com/verl-project/verl/pulls?q=is%3Apr+is%3Aopen+topk
  - https://github.com/verl-project/verl/issues?q=is%3Aissue+is%3Aopen+distillation+kl

### Test

Unit tests (CPU), on Linux with `torch 2.11.0+cu129`:

```bash
python -m pytest tests/workers/test_distillation_topk_symmetry_on_cpu.py \
  tests/workers/test_chunked_topk_log_probs_on_cpu.py \
  tests/test_base_config_on_cpu.py tests/utils/test_config_on_cpu.py -q
# 41 passed
```

The new coverage asserts: non-negativity; equality against an independent coarse-distribution oracle; autograd gradient equality; analytic sharded-gradient equality with cross-shard teacher IDs; the lower-bound property across K; equality with the full KL when K spans the vocabulary; regular and chunked FSDP paths; invalid clamp/epsilon configuration; and the fused-mass fallback / fail-closed behaviour.

Megatron tensor-parallel parity (2 GPUs):

```bash
torchrun --standalone --nnodes=1 --nproc-per-node=2 tests/utils/test_special_megatron_kl_loss_tp.py
# [PASS] VP KL divergence correctness verified for test case 0..3
# [PASS] VP tail-aware KL correctness verified for test case 0..3
```

Loss, `student_mass`, `teacher_mass` and `tail_loss` match the FSDP reference, and every vocabulary-shard gradient matches FSDP within `atol=rtol=1e-4`; existing truncated top-k cases do not regress.

Synthetic full-KL benchmark (`examples/on_policy_distillation_trainer/benchmark_tail_kl.py`) — gradient cosine similarity to the true full-KL gradient, and the frequency of negative (clamped) truncated losses:

| K | truncated | tail-aware | negative truncated loss frequency |
|---|---|---|---|
| 8 | 0.303 | 0.455 | 6.25 % → 0 % |
| 16 | 0.327 | 0.561 | 0 % |
| 32 | 0.393 | 0.653 | 0 % |
| 64 | 0.488 | 0.793 | 0 % |

Scalar error relative to the full KL is *not* uniformly better (it is slightly worse), which is expected: tail-aware KL is a lower bound, not a better point estimate. We do not claim otherwise.

Real-model efficiency measurement — student Qwen3-1.7B (FSDP + colocated vLLM rollout) and teacher Qwen3-32B (vLLM TP=2) on 3×RTX A6000, GSM8K, batch 16, identical seed/data/batch/optimizer for both arms, validation on the full 1319-example test set (standard error ≈ ±1.3 pp). The only difference between arms is `loss_mode`; both arms run with `log_prob_min_clamp=null`.

Experiment A (same run, same step grid):

| steps | `forward_kl_topk` | `forward_kl_topk_tail` |
|---|---|---|
| 175 | 0.3177 | **0.3472** |
| 350 | 0.3457 | **0.3616** |

At 175 steps the tail-aware arm already matches what the baseline reaches at 350 steps (0.3472 ≥ 0.3457) — i.e. **≤ 50 % of the training steps for the same accuracy**.

Experiment B (independent seed, earliest phase):

| steps | `forward_kl_topk` | `forward_kl_topk_tail` |
|---|---|---|
| 50 | 0.3412 | **0.4246** |

The leading margin shrinks with training (50 steps +8.34 pp, 175 steps +2.96 pp, 350 steps +1.59 pp), consistent with two curves sharing an asymptote but converging at different rates.

Cost is unchanged: 142.6 vs 143.0 tokens/s, 40.58 vs 40.61 GB peak HBM, and `grad_norm` of the same order, so the efficiency difference is not bought with extra compute. Mechanistically, `loss_min` is exactly `0.0` (clamped) in 415/467 steps for the baseline versus `≈ −1e−7` (finite-precision noise, never clamped) for the tail-aware arm, while the top-k overlap ratio is essentially identical (0.7153 vs 0.7119), i.e. the difference does not come from different token selection.

Caveats: both arms are single-seed, and the measured run-to-run noise for the same arm at the same step is ≈3.3 pp, so Experiment B's margin is ≈2.5× the noise while Experiment A's is of the same order. We report three step budgets rather than only the most favourable one, and multi-seed error bars are in progress.

### API and Usage Example

No API change; it is a new value for an existing config field.

```bash
distillation.distillation_loss.loss_mode=forward_kl_topk_tail
distillation.distillation_loss.topk=16
distillation.distillation_loss.log_prob_min_clamp=null   # required by this mode
distillation.distillation_loss.use_policy_gradient=false  # intended setting
```

`tail_mass_eps` (new, default `1e-6`) is the numerical guard on `q_tail`; it must lie strictly between 0 and 1.

### Design & Code Changes

- `verl/trainer/distillation/tail_kl.py` (new): stable tail-bucket value computation in FP32, plus an analytic, Megatron-independent logit-gradient helper. For `j` inside the teacher top-k `dL/dz_j = q_j − p_j`; for `j` outside, `dL/dz_j = q_j (P − Q) / (1 − Q)`, which every vocabulary shard can apply locally without gathering logits. The helper also matches the derivative of the `q_tail.clamp_min(tail_mass_eps)` guard when it is active.
- `verl/trainer/distillation/losses.py`: registers the mode with `use_topk=True` (so the teacher path requests top-k log-probs as before), dispatches to the FSDP/VeOmni and Megatron backends, keeps `clamp_min(0)` only for the legacy mode, and adds `head_loss`, `tail_loss`, `teacher_tail_mass`, `student_tail_mass` metrics. If a fused engine returns a detached `student_mass` it fails closed instead of silently training with an incomplete gradient.
- `verl/trainer/distillation/fsdp/losses.py`: refactors the existing top-k path without changing legacy behaviour, adds `compute_forward_kl_topk_tail`, and supports both the regular and `use_chunked_topk=True` paths.
- `verl/trainer/distillation/megatron/losses.py`: extends `_VocabParallelKLDivergence` with `include_tail` / `tail_mass_eps`, all-reduces the student top-k mass across TP ranks, and uses the shard-local analytic gradient in backward.
- `verl/workers/config/distillation.py` + config YAMLs: `tail_mass_eps`, validation rules (strictly between 0 and 1; `forward_kl_topk_tail` requires `log_prob_min_clamp=null`), and a warning when combined with `use_policy_gradient=true`. Per-entry log-prob clamping is rejected for this mode because it would change the selected top-k mass and invalidate the coarse-grained KL.
- Tests, benchmark, docs and generated configs as listed in the diff.

**VeOmni backend.** Two notes, neither of which is a defect in VeOmni:

- *Fused-kernel masses — our earlier claim was wrong.* Our first measurement with `model_engine=veomni` + `use_fused_kernels=True` showed `teacher_mass ≈ 2.11e-4`, which we described as the fused kernel's mass outputs being degenerate. Instrumenting the payload showed instead that every teacher top-k log-prob was exactly `-log V` — the *teacher payload* was uniform — and `2.11e-4 = K / V` is the correct mass of that payload. Serving the same teacher directly through vLLM returns a real distribution, and re-running the example with the default `distillation.teacher_models.teacher_model.*` entry instead of the named `+distillation.teacher_models.gsm8k.*` entry delivers a real payload (`Σ exp = 0.0377`). The artefact was in our own configuration, not in the fused kernel.
- *Design limitation (documented).* `chunk_topk_distill` returns `student_mass` / `teacher_mass` with `requires_grad=False` by design, so the tail bucket cannot receive gradient through the fully-fused path. `forward_kl_topk_tail` therefore requires `use_fused_kernels=False` on the VeOmni backend; the fail-closed guard rejects the fused combination rather than silently training with an incomplete gradient.

Two environment-level blockers were also worked around for the measurement: veomni 0.1.11's generated Qwen3 patch passes `cache_position` to `transformers.masking_utils.create_causal_mask`, which the pinned transformers 5.9.0 no longer accepts; and single-GPU VeOmni runs need `actor_rollout_ref.actor.veomni.init_device=cuda` because `fsdp_enabled` is false when `fsdp_size == 1` while the default is `meta`.

- [x] Apply pre-commit checks: `pre-commit run --all-files` (ruff, ruff-format, mypy 1.17.0, generated-config verification, docstring/license/device-API/DataProto/structure/example-naming/uv-gpu-only checks and `compileall` with warnings-as-errors all pass on Linux).
- [x] Add / Update the documentation (`docs/algo/opd.md`, `examples/on_policy_distillation_trainer/README.md`).
- [ ] Add unit or end-to-end test(s) to the CI workflow. The new CPU tests extend the existing `tests/workers/test_distillation_topk_symmetry_on_cpu.py` and the Megatron TP test covers the GPU path; wiring them into `.github/workflows` may need a maintainer's guidance on where they best fit.
- [ ] CI request sent in the `ci-request` channel.

**AI assistance disclosure:** the implementation, tests and experiment harness were produced with AI assistance (Claude Code). The submitting human reviews every changed line and ran the tests and experiments reported above.


PengyuLi7:
Design discussion is tracked in the RFC issue #7856 (motivation, the tail-bucket objective, and why this is complementary to the full-vocabulary route in #7239/#7375). Feedback there is welcome before review.

Fengqi0822:
This is a really good idea!

wuxibin89:
cc @JacobHelwig

JacobHelwig:
> Known limitation — VeOmni backend. With model_engine=veomni and use_fused_kernels=True the fused kernel's top-k mass outputs are degenerate on our host (teacher_mass ≈ 2.11e-4, effectively constant per token, versus ≈0.99 and varying in the FSDP path). The same values appear with the pre-existing forward_kl_topk mode, so this is a property of the fused path and not of this change. 

Does this mean that VeOmni top-k forward KL is broken @PengyuLi7 ?

docs/algo/opd.md
```diff
@@ -587,6 +645,18 @@ These metrics are logged for top-$k$ loss modes such as `forward_kl_topk`.
 - `actor/distillation/teacher_mass_min` / `actor/distillation/teacher_mass_max`  
   Minimum and maximum teacher mass on the teacher top-$k$ tokens within the batch.
 
+- `actor/distillation/tail_loss`
+  Mean contribution of the aggregate non-top-k bucket. Emitted by
+  `forward_kl_topk_tail`.
+
+- `actor/distillation/head_loss`
```
JacobHelwig: top-k loss would be a better name I think

examples/on_policy_distillation_trainer/benchmark_tail_kl.py
JacobHelwig, line 1: Is this developmental code that we should check in?

examples/on_policy_distillation_trainer/README.md
```diff
@@ -19,6 +19,27 @@ the single-teacher scripts. The MOPD script exposes per-teacher overrides.
...
+```
+
+To run an existing trainer script with the tail-aware objective, disable its
+legacy per-entry log-probability clamp explicitly:
+
+```bash
+DISTILLATION_LOSS_MODE=forward_kl_topk_tail USE_POLICY_GRADIENT=False \
+  bash examples/on_policy_distillation_trainer/run_qwen3_8b_fsdp.sh \
+  distillation.distillation_loss.log_prob_min_clamp=null
+```
```
JacobHelwig: Same question, do we expect users to use this?

docs/algo/opd.md
```diff
@@ -410,6 +426,32 @@ $$
 
 The reason GKD OPD is implemented only over the teacher top-$k$ logits is because current inference servers return log-probabilities for the sampled token and the teacher top-$k$ tokens, but do not support gathering log-probabilities at arbitrary token IDs. Therefore, the implementation supports teacher-top-$k$ forward KL, but not student-top-$k$ reverse KL.
 
+The truncated objective above is not itself a KL between normalized
+distributions: both the teacher and student top-$k$ masses can be smaller than
+one, so an individual token loss can be negative. The legacy
```
JacobHelwig: Why is it legacy?

docs/algo/opd.md
```diff
@@ -410,6 +426,32 @@ $$
...
+`forward_kl_topk` mode clamps those negative per-token values to zero for
+backward compatibility.
```
JacobHelwig: Backward compatibility with what?

docs/algo/opd.md
```diff
@@ -317,6 +322,16 @@ Lower clamp on log probabilities used inside divergence computations, to
...
 
+`forward_kl_topk_tail` requires this option to be `null`. Per-entry clamping
+changes the probability mass of the selected tokens and would invalidate the
+coarse-grained KL definition.
+
+### `distillation.distillation_loss.tail_mass_eps` (float)
+
+Minimum student probability mass used for the aggregate non-top-k bucket in
+`forward_kl_topk_tail`. Default: `1e-6`. This only guards `log(0)` when the
+student top-k mass rounds to one in finite precision.
```
JacobHelwig: this guard could be made better and more analogous to `log_prob_min_clamp` by clamping both the student and teacher tail bucket difference. Otherwise, we have the case where:

Student tail mass: ~1
Teacher tail mass: 1e-100
KL: 1 * log(1 / 1e-100) > 200

We could even re-use the existing log prob min clamp argument for this versus adding a new argument, or name them/document them better so that the relationship is more clear

docs/algo/opd.md
```diff
@@ -420,7 +462,23 @@ distillation:
       use_policy_gradient: false
 ```
 
-Do not use `forward_kl_topk` with `use_policy_gradient=true`. The top-$k$ loss contains distributional information for many teacher-preferred tokens, but a policy-gradient update only acts through the sampled token:
+To use the tail-aware coarse-grained objective without changing teacher
+communication:
```
JacobHelwig: can we clarify what "teacher communication" means? eg "...without adding any overhead to the teacher logprob computation"

verl/trainer/distillation/fsdp/losses.py
```diff
@@ -72,26 +73,48 @@ def kl_divergence(log_q: torch.Tensor, log_p: torch.Tensor) -> torch.Tensor:
...
+        mass, and the tail-bucket contribution.
+    """
+    log_p = log_p.float()
+    log_q = log_q.float()
+    p = log_p.exp()
+    q = log_q.exp()
+
+    teacher_mass = p.sum(dim=-1)
+    student_mass = q.sum(dim=-1)
+    head_loss = (p * (log_p - log_q)).sum(dim=-1)
```
JacobHelwig: topk_loss is a better name, "tail of probability distribution" returns results on google but "head of probability distribution" doesn't

verl/trainer/distillation/fsdp/losses.py
```diff
@@ -72,26 +73,48 @@ def kl_divergence(log_q: torch.Tensor, log_p: torch.Tensor) -> torch.Tensor:
...
-    Args:
-        student_logits: (bsz, seqlen/sp_size, vocab_size).
-        teacher_topk_log_probs: (bsz, seqlen, topk).
-        teacher_topk_ids: (bsz, seqlen, topk).
-        data_format: "thd" or "bshd", models not support THD format, e.g GPT-OSS, Qwen3.5
-
-    Returns:
-    - distillation_losses: (bsz, seqlen/sp_size)
-    - student_mass: (bsz, seqlen/sp_size)
-    - teacher_mass: (bsz, seqlen/sp_size)
```
JacobHelwig: Why do we want to drop args and return values from doc string?

verl/trainer/distillation/fsdp/losses.py
```diff
@@ -122,28 +145,88 @@ def compute_forward_kl_topk(
...
+
+
+def compute_forward_kl_topk_tail(
+    student_logits: torch.Tensor,
+    teacher_topk_log_probs: torch.Tensor,
+    teacher_topk_ids: torch.Tensor,
+    config: DistillationConfig,
+    data_format: str,
+) -> dict[str, torch.Tensor]:
+    """Compute coarse-grained forward KL with one bucket for non-top-k tokens."""
```
JacobHelwig: lets match doc string format to the non-tail version, with args and returns

verl/trainer/distillation/megatron/losses.py
```diff
@@ -241,46 +280,56 @@ def backward(
...
+            outside_scale = target_active_mass
+            grad_input = vp_source_probs * outside_scale.unsqueeze(-1)  # [b, s, vocab_shard]
 
         topk = target_topk_indices.size(-1)
         grad_input_2d = grad_input.view(-1, grad_input.size(-1))
-        target_topk_probs_flat = target_topk_probs.view(-1, topk)  # (b*s, topk)
         target_topk_indices_flat = target_topk_indices.view(-1, topk)  # (b*s, topk)
 
-        # Subtract q_j for active entries (i.e., j in A), accumulating repeats via scatter_add_.
+        # Correct the teacher-top-k entries, accumulating repeats via scatter_add_.
```
JacobHelwig: Why change this comment?

verl/trainer/distillation/megatron/losses.py
```diff
@@ -241,46 +280,56 @@ def backward(
...
-        student_logits: (bsz, seqlen/cp_size, vocab_size/tp_size).
-        teacher_topk_log_probs: (bsz, seqlen, topk).
-        teacher_topk_ids: (bsz, seqlen, topk).
-        data_format: "thd" or "bshd", models not support THD format, e.g GPT-OSS, Qwen3.5
-
-    Returns:
-    - distillation_losses: (bsz, seqlen/cp_size)
-    - student_mass: (bsz, seqlen/cp_size)
-    - teacher_mass: (bsz, seqlen/cp_size)
-    """
```
JacobHelwig: Lets keep the doc string for the public versions

verl/trainer/distillation/losses.py
```diff
@@ -134,19 +135,30 @@ def compute_topk_loss(
...
+    loss_mode = distillation_config.distillation_loss.loss_mode
     match config.strategy:
         # VeOmni uses FSDP2 internally, so its loss computation is identical to FSDP.
         case "fsdp" | "veomni":
             import verl.trainer.distillation.fsdp.losses as fsdp_losses
 
-            distillation_loss_fn = fsdp_losses.compute_forward_kl_topk
+            loss_functions = {
+                "forward_kl_topk": fsdp_losses.compute_forward_kl_topk,
+                "forward_kl_topk_tail": fsdp_losses.compute_forward_kl_topk_tail,
```
JacobHelwig: we could make these partials with use_tail set accordingly to avoid adding new private functions in fsdp/megatron files

verl/trainer/distillation/losses.py
```diff
@@ -321,9 +333,29 @@ def compute_forward_kl_topk(
     teacher_mass = no_padding_2_padding(model_output["teacher_mass"], data)
     overlap_count = model_output.get("overlap_count")
     overlap_token_advantage = model_output.get("overlap_token_advantage")
+    tail_loss = model_output.get("tail_loss")
+    loss_mode = getattr(distillation_config.distillation_loss, "loss_mode", "forward_kl_topk")
```
JacobHelwig: why getattr? loss_mode is a required arg

verl/trainer/distillation/losses.py
```diff
@@ -321,9 +333,29 @@ def compute_forward_kl_topk(
...
     if overlap_count is not None and overlap_token_advantage is not None:
         overlap_count = no_padding_2_padding(overlap_count, data)
         overlap_token_advantage = no_padding_2_padding(overlap_token_advantage, data)
+    if tail_loss is not None:
+        tail_loss = no_padding_2_padding(tail_loss, data)
+    elif loss_mode == "forward_kl_topk_tail":
+        # Fused engines may compute the existing top-k head in-kernel and expose
```
JacobHelwig: "top-k head" --> top-k probabilities/mass

verl/trainer/distillation/losses.py
```diff
@@ -321,9 +333,29 @@ def compute_forward_kl_topk(
...
     overlap_count = model_output.get("overlap_count")
     overlap_token_advantage = model_output.get("overlap_token_advantage")
+    tail_loss = model_output.get("tail_loss")
+    loss_mode = getattr(distillation_config.distillation_loss, "loss_mode", "forward_kl_topk")
     if overlap_count is not None and overlap_token_advantage is not None:
         overlap_count = no_padding_2_padding(overlap_count, data)
         overlap_token_advantage = no_padding_2_padding(overlap_token_advantage, data)
+    if tail_loss is not None:
+        tail_loss = no_padding_2_padding(tail_loss, data)
+    elif loss_mode == "forward_kl_topk_tail":
```
JacobHelwig, line 344: when do we hit this path? that is, tail_loss is not returned but the loss mode is topk_tail?

verl/trainer/distillation/losses.py
```diff
@@ -361,8 +393,22 @@ def compute_forward_kl_topk(
...
     }
 
-    # Due to use of top-k, student and teacher distributions don't sum to 1 -> divergences can be negative.
-    distillation_losses = distillation_losses.clamp_min(0.0)
+    if tail_loss is not None:
+        valid_tail_loss = tail_loss[response_mask_bool]
+        valid_head_loss = (distillation_losses - tail_loss)[response_mask_bool]
+        distillation_metrics.update(
+            {
+                "distillation/head_loss": valid_head_loss.mean().item(),
```
JacobHelwig: topk_loss

verl/trainer/distillation/losses.py
```diff
@@ -361,8 +393,22 @@ def compute_forward_kl_topk(
...
+                "distillation/tail_loss": valid_tail_loss.mean().item(),
+                "distillation/teacher_tail_mass": (1.0 - teacher_mass).clamp_min(0.0).mean().item(),
+                "distillation/student_tail_mass": (1.0 - student_mass).clamp_min(0.0).mean().item(),
+            }
+        )
+
+    # The legacy truncated objective can be negative because its top-k masses
```
JacobHelwig: legacy?

verl/trainer/distillation/tail_kl.py
```diff
@@ -0,0 +1,64 @@
...
+    """Return the KL contribution of one aggregate non-top-k bucket.
+
+    The teacher top-k IDs define a shared coarse-graining of the vocabulary:
+    each selected token is its own category and every remaining token belongs
+    to one tail category. ``tail_mass_eps`` only guards the student tail's
+    logarithm; a zero teacher tail contributes zero by continuity.
+    """
+    teacher_tail = (1.0 - teacher_topk_mass.float()).clamp(min=0.0, max=1.0)
+    student_tail = (1.0 - student_topk_mass.float()).clamp(min=tail_mass_eps, max=1.0)
+    teacher_tail_for_log = teacher_tail.clamp_min(torch.finfo(teacher_tail.dtype).tiny)
```
JacobHelwig, line 38: why do we not clamp this to the tail mass eps as well?

docs/algo/opd.md
PengyuLi7: Done — renamed the metric to `distillation/topk_loss` (docs + code + the metric listing). I kept the `loss_mode` value itself as `forward_kl_topk_tail`, since "topk" for the retained part and "tail" for the aggregate bucket follows the same convention — happy to rename the mode too if you'd prefer something else.

examples/on_policy_distillation_trainer/benchmark_tail_kl.py
PengyuLi7: Agreed — removed. It was written as evidence for the RFC; the numbers it produces (gradient cosine similarity vs. a full-vocabulary KL oracle) now live in the description/issue instead of as a maintained script.

examples/on_policy_distillation_trainer/README.md
PengyuLi7: Agreed — removed the usage example. The README change is now just the new value in the `loss_mode` enum list.

docs/algo/opd.md
PengyuLi7: Fixed — dropped "legacy" throughout. The sentence now reads "the existing `forward_kl_topk`" and states what the clamp actually costs: it keeps the scalar loss well behaved but discards the gradient of every token whose student top-k mass exceeds the teacher's.

docs/algo/opd.md
PengyuLi7: Fixed — dropped "legacy" throughout. The sentence now reads "the existing `forward_kl_topk`" and states what the clamp actually costs: it keeps the scalar loss well behaved but discards the gradient of every token whose student top-k mass exceeds the teacher's.

docs/algo/opd.md
PengyuLi7: Reworded to: *"...without any additional overhead in the teacher log-probability computation (the teacher request and payload are unchanged)"*.

verl/trainer/distillation/fsdp/losses.py
PengyuLi7: Renamed — `head_loss` → `topk_loss` (metric key and the local variable).

verl/trainer/distillation/fsdp/losses.py
PengyuLi7: Restored — the shared implementations have their Args/Returns back, and the public entry points keep full docstrings.

verl/trainer/distillation/fsdp/losses.py
PengyuLi7: Done — after the refactor below there is a single public entry point per backend, documented in the same Args/Returns format as before, with the new `include_tail` argument described.

verl/trainer/distillation/megatron/losses.py
PengyuLi7: Reverted — my rewording was gratuitous; the original comment is back.

verl/trainer/distillation/megatron/losses.py
PengyuLi7: Restored — the shared implementations have their Args/Returns back, and the public entry points keep full docstrings.

verl/trainer/distillation/losses.py
PengyuLi7: Done — the dispatch now uses `partial(compute_forward_kl_topk, include_tail=True)`, both backends keep a single implementation, and `compute_forward_kl_topk_tail` is gone (the tests build the same partial).

verl/trainer/distillation/losses.py
PengyuLi7: Fixed — direct attribute access. One CPU test was relying on the implicit default, so it now sets `loss_mode` explicitly.

verl/trainer/distillation/losses.py
PengyuLi7: Reworded to "top-k probabilities/mass".

verl/trainer/distillation/losses.py
PengyuLi7: Good question — that branch is easy to misread, and you're right that it is the one place where the mode and the payload disagree. It is for **fused engines**: the non-fused FSDP and Megatron implementations return `tail_loss` themselves and take the `if tail_loss is not None` branch above, whereas a fused engine that computes the existing top-k objective in-kernel (e.g. VeOmni's `chunk_topk_distill`) returns only the head loss plus `student_mass` / `teacher_mass`, so we add the tail bucket here from those masses — no new teacher request, no `[T, V]` logits.

The guard immediately below covers the case where such an engine returns a *detached* `student_mass`: the tail bucket would then receive no gradient, and we would rather raise than silently train with a partially-detached loss.

I'm happy to restructure this if you find it too implicit — either gating on an explicit engine capability flag, or having fused engines report an explicit `tail_loss`/sentinel instead of relying on `tail_loss is None`. Whichever you think reads better, I'll follow.

verl/trainer/distillation/losses.py
PengyuLi7: Renamed — `head_loss` → `topk_loss` (metric key and the local variable).

verl/trainer/distillation/losses.py
PengyuLi7: Fixed — dropped "legacy" throughout. The sentence now reads "the existing `forward_kl_topk`" and states what the clamp actually costs: it keeps the scalar loss well behaved but discards the gradient of every token whose student top-k mass exceeds the teacher's.

verl/trainer/distillation/tail_kl.py
PengyuLi7: Because the two tails play different roles, and clamping the teacher side to `tail_mass_eps` may be incorrect:

- **Student tail** sits inside a logarithm (`log(q_tail)`) as the denominator. `q_tail = 0` sends both the loss and its gradient to infinity, so `tail_mass_eps` is exactly the bound we want there.
- **Teacher tail** is only a *multiplicative weight*: as `p_tail → 0` the whole term vanishes by continuity (`x log x → 0`). Clamping it to `tail_mass_eps` would inject a spurious `eps · log(eps / q_tail)` contribution for every token whose teacher top-k mass is exactly 1, i.e. it would make the loss depend on the guard rather than on the distributions.

This is not hypothetical: on the FSDP path we measure `actor/distillation/teacher_mass_max` of `1.0000001 … 1.0000005` (fp32 rounding of a top-k mass that is exactly 1), so `p_tail = 1 - mass` is `0` after the `clamp(min=0)`, and the spurious term would be added on those tokens.

So `p_tail` is left unclamped and only the *input of its logarithm* is floored at the dtype minimum, which keeps `log` finite without changing the product (an exact-zero teacher tail still contributes exactly zero). I've expanded the in-code comment to spell this out.

PengyuLi7:
> Does this mean that VeOmni top-k forward KL is broken @PengyuLi7 ?

No — and our paragraph was wrong to imply it. Thank you for pushing on this. We have since localised it, and it is on our side.

**What we actually measured.** In the run we reported, the teacher's top-k log-probs were uniform: every entry was exactly `-log V`, so `Σ_k exp(log p_k) = K / V` — that is, the `2.1099e-4` we reported as "degenerate" is precisely the mass of the payload the loss was handed. The fused kernel computed it correctly; in the same run `student_mass` spans `4e-17 … 0.99999`, as it should.

**Where the uniform payload came from.** Two checks, both pointing away from VeOmni:

1. Serving the same Qwen3-1.7B directly through vLLM with `prompt_logprobs=32` returns a real distribution (`':'` at -1.50, `' is'` at -2.38, `'’s'` at -2.63, …), so the model and the log-prob request path are fine.
2. Re-running the VeOmni example **unchanged except for the teacher entry** — the named `+distillation.teacher_models.gsm8k.*` block replaced with the default `distillation.teacher_models.teacher_model.*` entry — delivers a real payload: first eight log-probs `[-4.40, -5.90, -6.46, -6.59, -6.77, -6.77, -6.96, -6.99]` and `Σ exp = 0.0377` instead of `K / V`.

So the degenerate payload was an artefact of how we had configured the teacher in that run (a named-teacher entry, alongside the three local workarounds for transformers 5.9.0 / FA2 / single-GPU init) — not the fused kernel, and not VeOmni.

One VeOmni limitation **does** apply, and it is a documented design choice: `chunk_topk_distill` returns `student_mass` / `teacher_mass` with `requires_grad=False` (they are reported as metrics; the kernel carries its own analytic backward for the truncated objective). The tail bucket needs the student-mass gradient, so `forward_kl_topk_tail` must run with `use_fused_kernels=False` on that backend; the fail-closed guard rejects the fused combination instead of silently training with an incomplete gradient.

I've rewritten that section of the description accordingly.

PR #7570, opened 2026-08-26
[rollout, vllm, ckpt] feat: add Laminar rollout integration hooks

tongyx361:
### What does this PR do?

Expose the verl APIs that an external Laminar async-RL controller needs to place rollout replicas, observe KV-cache pressure, interrupt generation, and reuse shared KV entries within a model version. Provide opt-in local staging for Megatron HF checkpoint saves on filesystems that cannot handle safetensors >= 0.8.0 allocation calls; direct writes remain the default.

### Checklist Before Starting

- [x] Search for similar PRs: [Laminar](https://github.com/verl-project/verl/pulls?q=is%3Apr+is%3Aopen+Laminar), [shared KV](https://github.com/verl-project/verl/pulls?q=is%3Apr+is%3Aopen+shared+KV), [safetensors staging](https://github.com/verl-project/verl/pulls?q=is%3Apr+is%3Aopen+safetensors+staging), [VLLM_PORT](https://github.com/verl-project/verl/pulls?q=is%3Apr+is%3Aopen+VLLM_PORT).
- [x] Use the `[{modules}] {type}: {description}` title format.
- [x] Duplicate-work check: #7859 already isolates colocated vLLM internal ports. This PR now reuses the same `get_ranked_port_base` helper, Ray-job namespace, and `port_namespace` injection instead of a second fixed-constant allocator. The remaining Laminar hooks (snapshot, abort, shared KV, checkpoint staging) are not in #7859.

The contribution is the Laminar integration surface. Its placement, observation, and generation-control APIs are separate from load-balancing and prefix-warmup policies; the native `cache_salt` parameter is a shared primitive that those policies can reuse.

### Test

CPU tests covering the new helpers and review fixes were added under existing `*_on_cpu.py` files. Suites were run with Python 3.12 and runtime-image dependencies, with this checkout first on `PYTHONPATH`:

```bash
uv run --no-project --python /usr/bin/python python -m pytest -q \
  tests/utils/ckpt/test_megatron_checkpoint_manager_on_cpu.py \
  tests/utils/test_net_utils_on_cpu.py \
  tests/workers/rollout/test_vllm_pd_disaggregation_on_cpu.py \
  tests/workers/rollout/rollout_vllm/test_submitter_gate_on_cpu.py
```

**112 passed** (10.44s), including the new port-allocator, abort/snapshot, and staging-path cases.

[...]

PR #7859, opened 2026-09-14
[rollout, vllm] fix: isolate internal ports across colocated engines

leooop-al:
# [rollout, vllm] fix: isolate internal ports across colocated engines

## What does this PR do?

Prevent concurrent vLLM engines on the same node from selecting the same
internal TCP port during startup.

vLLM uses `VLLM_PORT` as the starting point for its internal port scan. During
multiprocessing startup, a candidate port can be probed and released before a
worker binds its TCPStore. Engines that start concurrently from the same value
can therefore race and fail with `EADDRINUSE`, surfaced as an engine-core
initialization failure.

## Failure symptom

The issue occurs during high-concurrency deployment when several rollout or
teacher engine actors start on the same node. One actor can successfully probe
a free port, release the probe socket, and then lose the race before its
TCPStore binds. The affected engine fails before serving any requests, which
aborts rollout or teacher-manager initialization.

The user-visible error is typically reduced by vLLM to:

```text
RuntimeError: Engine core initialization failed
```

The underlying worker log reports the TCP bind collision:

```text
torch.distributed.DistNetworkError: ... EADDRINUSE ...
The server socket has failed to listen on a local network address.
The address is already in use.
```

The supplied deployment log demonstrates the triggering topology: many vLLM
and teacher workers initialize concurrently on the same nodes. It does not
contain the final exception lines above, so this section describes the
observed failure chain rather than quoting that file verbatim.

This PR:

- adds a stable, bounded port-start allocator based on a namespace and rank;
- namespaces each vLLM engine by Ray job ID and its unique server actor name;
- separates replica and node ranks within each namespace;
- covers standard rollout, reward, teacher, and PD prefill/decode servers; and
- keeps vLLM's existing fallback scan when the selected start port is occupied.

The change intentionally does not retry every generic
`Engine core initialization failed` error, because that message also covers
non-port failures such as invalid configuration or model initialization.

Related open PR: #7570. That PR assigns a fixed `VLLM_PORT` start from
`replica_rank` and `node_rank`. This change addresses the remaining collision
cases by including the Ray job and exact server identity, which separates
concurrent jobs, same-rank rollout/reward/teacher actors, and multiple PD
prefill/decode actors.

## Checklist Before Starting

- [x] Searched [open VLLM_PORT PRs](https://github.com/verl-project/verl/pulls?q=is%3Apr+is%3Aopen+VLLM_PORT).
- [x] Compared with #7570 and documented the material difference above.
- [x] Title follows `[{modules}] {type}: {description}`.

## Test

```bash
.venv-python312/bin/python -m pytest -q tests/utils/test_net_utils_on_cpu.py
# 9 passed

PATH="$PWD/.venv/bin:$PATH" pre-commit run --files \
  verl/utils/net_utils.py \
  verl/workers/rollout/vllm_rollout/vllm_async_server.py \
  verl/workers/rollout/vllm_rollout/vllm_pd_replica.py \
  tests/utils/test_net_utils_on_cpu.py
# All hooks passed
```

The unit tests cover deterministic allocation, namespace/rank separation,
range wraparound, and invalid arguments. A full multi-node vLLM deployment was
not run locally.

## API and Usage Example

No user-facing configuration is required. Each vLLM server actor derives its
internal port-search start automatically.

## Design & Code Changes

- `verl/utils/net_utils.py`
  - Add `get_ranked_port_base()`.
  - Hash the namespace with BLAKE2s and offset each rank by a fixed stride
    within the configured TCP port range.
- `verl/workers/rollout/vllm_rollout/vllm_async_server.py`
  - Combine the Ray job ID with the server actor namespace.
  - Flatten replica/node ranks and export the result through `VLLM_PORT`
    before vLLM creates its engine.
  - Pass unique actor names for rollout, reward, and teacher servers.
- `verl/workers/rollout/vllm_rollout/vllm_pd_replica.py`
  - Pass each prefill/decode actor name as its port namespace.
- `tests/utils/test_net_utils_on_cpu.py`
  - Add GPU-free coverage for the port allocator.

## Checklist Before Submitting

- [x] Relevant pre-commit hooks pass.
- [ ] Full repository-wide pre-commit and CI.
- [ ] Full multi-node vLLM deployment validation.
- [x] No documentation update is required because there is no public API or
  configuration change.
- [x] CPU unit tests were added under the existing `*_on_cpu.py` workflow.
- [x] Recipe submodule is not affected.

## AI assistance disclosure

AI assistance was used to analyze the failure mode, implement the change,
write tests, and prepare this PR description. The human submitter must review
every changed line and validate the deployment behavior before merge.


wuxibin89:
I think @aoshen02 is trying to fix port conflict in vllm side.

aoshen02:
Hi, most of the port conflict problem should have been solve via https://github.com/vllm-project/vllm/issues/51275, could you please upgrade to nightly version and recheck? Let me know if still problem exists.

leooop-al:
Thanks for the information. To clarify, this issue was reproduced on Ascend NPUs, using vLLM 0.23 together with vllm-
  ascend 0.23.

  The failure happens when multiple rollout and teacher vLLM engines are initialized concurrently on the same NPU node,
  resulting in an internal TCPStore port collision (`EADDRINUSE` / `Address already in use`).

  Could you confirm whether the relevant fixes tracked in vLLM #51275 are already included in vLLM 0.23 and supported by
  vllm-ascend 0.23?

  If the fix is only available in a newer nightly, could you recommend a compatible pair of vLLM nightly and vllm-ascend
  nightly versions or commits for verification? In the Ascend environment, upgrading vLLM alone may not be compatible with
  vllm-ascend. @aoshen02

aoshen02:
> Thanks for the information. To clarify, this issue was reproduced on Ascend NPUs, using vLLM 0.23 together with vllm- ascend 0.23.
> 
> The failure happens when multiple rollout and teacher vLLM engines are initialized concurrently on the same NPU node, resulting in an internal TCPStore port collision (`EADDRINUSE` / `Address already in use`).
> 
> Could you confirm whether the relevant fixes tracked in vLLM #51275 are already included in vLLM 0.23 and supported by vllm-ascend 0.23?
> 
> If the fix is only available in a newer nightly, could you recommend a compatible pair of vLLM nightly and vllm-ascend nightly versions or commits for verification? In the Ascend environment, upgrading vLLM alone may not be compatible with vllm-ascend. @aoshen02

These are all python code. What about just trying do patching? No compilation needed.

PR #7879, opened 2026-09-16
[vllm, rollout] fix: preserve explicit false for delayed-default boolean flags

zhangyimi:
### Summary

Preserve explicit `False` when converting vLLM engine configuration into CLI
arguments for plain `bool` fields whose default is `None` or `True`.

For example, a field declared as `enable_flashinfer_autotune: bool = None`
is a plain `bool` annotation with a delayed default.
The existing helper only recognizes `bool | None`, so an explicit `False`
is omitted instead of becoming `--no-enable_flashinfer_autotune`.
vLLM then receives the default rather than the user's explicit choice.

The fix depends on the field declaration, not on a particular vLLM release.
This does not change the default autotuning policy and has no NVFP4 dependency.
False-default bools, bool/string unions and unknown flags retain their existing
behavior.

### Tests

Tested commit: `57f9d435`, with vLLM 0.26.0 (the validation environment,
not a minimum-version requirement). Other versions were not runtime-tested.

```bash
python -m pytest -q tests/workers/rollout/test_vllm_cli_args_on_cpu.py
```

**28 passed, no skips**, including real vLLM parser round trips for True/False
and underscore/hyphen spellings, plus synthetic None/True/False defaults.
Scoped Ruff 0.12.2 check/format and `git diff --check` passed.
Full pre-commit and upstream CI have not been run.

### Related work

Searched for [matching CLI fixes](https://github.com/verl-project/verl/pulls?q=is%3Apr+is%3Aopen+optional_bool_vllm_args)
and `enable_flashinfer_autotune`; no other matching fix was found.

### AI assistance

Developed and tested with OpenAI Codex assistance. The submitting author
reviewed the changes and approved submission.


HollowMan6 (approved):
LGTM

PR #7880, opened 2026-09-16
[rollout, perf] fix: preserve uint8 expert routes in agent storage

zhangyimi:
### Summary

Preserve backend-provided uint8 expert IDs in rollout result storage, padding
and transport instead of unconditionally widening them to int16.
This halves the routing tensor's storage bytes for those inputs, not total
model memory.

Expert IDs are integer indices, independent of weight/activation precision.
For example, [vLLM 0.26 selects uint8 for models with at most 256 experts](https://github.com/vllm-project/vllm/blob/v0.26.0/vllm/model_executor/layers/fused_moe/routed_experts_capturer.py#L248-L264).
BF16 MoE rollouts can therefore benefit too; this is not NVFP4-specific.

- Preserve uint8 only when the NumPy/PyTorch input is already uint8.
- Keep the existing int16 fallback for signed/wider inputs and Python lists.
- Preserve IDs 128–255 without signed-int8 overflow.
- Leave padding, replay masks and final Megatron int64 targets unchanged.

The production change is confined to agent-loop result packing (+12/-2 lines).
This shared rollout layer is used by ordinary single-turn generation as well
as tool/multi-turn agents; the change does not alter optimization or model math.

### Tests

Tested commit: `d21555ee`. **15 passed, no skips**, including both CUDA BF16
replay cases. Coverage includes dtype/value preservation, high expert IDs,
signed/wider fallbacks, DataProto operations, pickle/NumPy transport and
equivalent Megatron replay targets.

```bash
python -m pytest -q tests/experimental/agent_loop/test_route_storage_on_cpu.py \
  tests/experimental/agent_loop/test_route_storage_replay.py
```

Test environment: GB200, PyTorch 2.11.0, vLLM 0.26.0, Megatron-Core 0.18.0,
Transformer Engine 2.18.0. Scoped Ruff 0.12.2 check/format and
`git diff --check` passed. Full pre-commit and upstream CI have not been run.

### Fixed-workload benchmark

The included benchmark measures host-side agent postprocessing, concat/select
and pickle round trips. It excludes generation, reward, Ray networking and model
computation; **these are not end-to-end training speedups**.

```bash
OMP_NUM_THREADS=1 python -m tests.experimental.agent_loop.benchmark_route_storage \
  --batch-size 32 --repeats 10
OMP_NUM_THREADS=1 python -m tests.experimental.agent_loop.benchmark_route_storage \
  --batch-size 192 --repeats 10
```

Single CPU thread on a GB200 host; identical synthetic expert IDs and sequence
lengths, two warmup pairs and ten measured pairs with alternating order.
Padded shape: `[batch, 21504, 48, 8]`; actual prompt length 256 and repeating
response lengths 128/512/1024/2048.

| Batch | int16 median | uint8 median | Pipeline speedup | Route storage |
|---|---:|---:|---:|---:|
| 32 | 0.9298 s | 0.6902 s | 1.347x | 504 → 252 MiB |
| 192 | 5.7732 s | 4.0635 s | 1.421x | 3024 → 1512 MiB |

### Related work

Searched [routed_experts](https://github.com/verl-project/verl/pulls?q=is%3Apr+is%3Aopen+routed_experts)
and uint8 changes. #4517 adds AgentData fields, #5506 fixes sequence
truncation/alignment, and #7318 aligns checkpoint bucket offsets; none preserves
the backend expert-ID storage dtype as this change does.

### AI assistance

Developed and tested with OpenAI Codex assistance. The submitting author
reviewed the changes and approved submission.


verl/experimental/agent_loop/agent_loop.py
```diff
@@ -128,7 +138,7 @@ def as_dict(self) -> dict[str, Any]:
 
         routed_experts = output.pop("routed_experts", None)
         if routed_experts is not None:
-            routed_experts = torch.tensor(routed_experts, dtype=torch.int16)
+            routed_experts = torch.tensor(routed_experts, dtype=_route_storage_dtype(routed_experts))
```
wuxibin89, line 141: rollout backends other than vllm may use different dtype for routed_experts, int16 is a safe dtype for all rollout backends. Moreover, since most MoE models have more experts than 256(e.g GLM-5.3-Flash 288, DeepSeek-V4.1-Flash 384, Kimi-K3 896), int8 of range 0-255 is not enough.

PR #7885, opened 2026-09-16
[vllm, rollout, trainer] fix: reject gate-parked requests when a hybrid replica leaves rotation

### LINKED ISSUE #7865: separate_async: requests routed to a hybrid replica after `abort_all_requests` closes the submission gate are parked until validation, silently stranding prompt groups
## Summary

In V1 separate_async, the hybrid vLLM replicas (engines colocated in the trainer processes) stay in the rollout load balancer from init until the first `switch_to_trainer`, which runs in on_step_begin of step 1. on_train_begin dispatches the warmup batch before that, so part of it is routed to hybrid servers. `switch_to_trainer` then calls `abort_all_requests`, which closes the per-server submission gate introduced in #7511 by @EricMarcus-ai . A request that reaches a hybrid server after the gate closes is neither rejected nor aborted: it waits on `_resume_event` until the next `resume_generation`, which for a hybrid replica with `hybrid_rollout.enable_switch=false` only happens at validation. It never reaches the engine, so the drain and the abort do not see it, and the agent loop never receives a response or a retry.

## Where

- Gate: `verl/workers/rollout/vllm_rollout/vllm_async_server.py`, `vLLMHttpServer.generate` (parking loop on self._submission_paused / await self._resume_event.wait()), `abort_all_requests` (sets _submission_paused = True, then pauses the engine), `resume_generation` (the only place that reopens the gate). Introduced by #7511 .
- Trainer side: `verl/trainer/ppo/v1/trainer_separate_async.py`, on_train_begin (dispatches warmup with hybrid replicas still in the balancer), `on_step_begin` (first _timed_switch_to_trainer), `switch_to_trainer` and `remove_replicas_from_balancer`.

### Information

- [ ] My own modified scripts

### Tasks

### Reproduction

Any `trainer.v1.trainer_mode=separate_async` run with hybrid replicas, `hybrid_rollout.enable_switch=false`, `num_warmup_batches >= 1`, and a validation interval larger than a few steps. Watch `timing_s/gen` stay high while `training/off_policy` shows missing groups; the parked requests appear in the vLLM server's debug log as "parking request ... until weight sync completes" and complete only when validation calls `resume_generation`.

### Expected behavior

A request that arrives at a server whose gate is closed for a trainer switch should fail fast with a re-tryable error so the agent loop re-routes it, or be counted and cancelled by `abort_all_requests` like an in-flight request. Independently, removal from the load balancer should complete before the gate closes, so no request can be routed to a server that is about to stop accepting.

EricMarcus-ai:
Might fix #7865, please see issue for context.

### What does this PR do?

### Checklist Before Starting

### Test

### API and Usage Example

```python
```

### Design & Code Changes

> [!IMPORTANT]


Begunner:
LGTM

verl/trainer/ppo/v1/trainer_separate_async.py
```diff
@@ -378,6 +378,8 @@ def switch_to_trainer(self):
         """Stop routing to Hybrid, abort partial requests, and return its GPU memory to training."""
         self.remove_replicas_from_balancer()
         self.checkpoint_manager.abort_replicas()
+        # Requests routed here before the removal landed would otherwise wait until validation.
+        self.checkpoint_manager.reject_parked_replicas()
```
wuxibin89, line 382: This is a common place for all rollout backend(e.g vllm, sglang, trtllm, etc), so it's not a good idea to do special case for vllm.

wuxibin89 (changes requested):

verl/trainer/ppo/v1/trainer_separate_async.py
EricMarcus-ai: Understand your concern. So, I also checked shortly the other engines, and the situation is as follows:

-  TRT-LLM already has its own gate in verl. abort_all_requests() clears _generation_allowed; generate() waits on that event; only resume_generation() sets it again. A late request can therefore wait on a withdrawn replica just like the original vLLM implementation. 
-  SGLang waits internally. Verl calls its tokenizer manager’s pause operation. In SGLang, that sets pause=True, and generation waits on a condition until continue_generation() clears it.
     
I.e. the issue probably exists for them as well, I have not run them fully in the investigation though.
The current PR is a bit ‘in between worlds’; either we do some vLLM-only fix, or we also make the methods here do something for other engines.

If we wanted vllm only: the option I can come up with is entirely inside the vLLM server by returning stop_reason="aborted" whenever the submission gate is closed.  That removes the new trainer/replica/checkpoint-manager calls. However, then the gate also stays closed during ordinary weight synchronization and throughout the training phase in colocate_async. Unfinished rollout requests would then retry once per second for the duration of those pauses, each involves balancer acquire/release calls and a server call. That does not sound so great either, haven’t benchmarked but adds a lot of idle ‘spinning’. Any preference / thoughts @wuxibin89 ?

wuxibin89:
@EricMarcus-ai There's another DP+resume fix https://github.com/verl-project/verl/pull/7846, is it relate to this PR?

wuxibin89:
@EricMarcus-ai Submit another PR to fix it: https://github.com/verl-project/verl/pull/7912, add a additional flag reject_request=True to abort_replicas, making replicas reject and return immediately instead of parking them.

PR #7864, opened 2026-09-14
[rollout, perf] fix: avoid full GC before weight resume

lbaolin:
### What does this PR do?

Our production RL training observed a throughput regression after upgrading verl v0.8.0 to v0.9.0, with colocated weight sync especially affected. Investigation found that v0.9.0 added `aggressive_empty_cache(force_sync=True)` immediately before rollout weights are resumed; controlled A/B tests show its full Python GC and retry materially slow weight sync without a measured GPU-memory benefit.

This PR replaces only that pre-resume call with `get_torch_device().empty_cache()`. Plain `empty_cache()` is sufficient at this boundary to return unused allocator blocks. The existing aggressive-cleanup call remains after weight transfer/offload, before KV-cache resume, so this removes redundant critical-path GC rather than eliminating aggressive cleanup.

Fixes #7848.

AI assistance was used for investigation, implementation, and drafting. I have reviewed the code changes myself.

### Checklist Before Starting

- [x] Searched open PRs for [issue #7848](https://github.com/verl-project/verl/pulls?q=is%3Apr+is%3Aopen+7848) and [`aggressive_empty_cache`](https://github.com/verl-project/verl/pulls?q=is%3Apr+is%3Aopen+aggressive_empty_cache). PR #7780 configures GC at the separate sender-cleanup boundary and does not overlap this change.
- [x] Formatted the title according to the repository convention.

### Test

Native V1 tests on 8×H200 with verl v0.9.0:

| Qwen3 workload | Aggressive | Plain | Improvement |
| --- | ---: | ---: | ---: |
| 8B dense | 4.867 s | 4.207 s | 13.6% |
| 8B parameter offload | 3.731 s | 3.031 s | 18.8% |
| 30B-A3B MoE | 8.204 s | 7.086 s | 13.6% |

The first `empty_cache()` released 7.66–40.55 GiB/rank. GC released 0 CUDA bytes, and its retry released no additional cache. All six runs completed without OOM and showed no meaningful peak-memory difference.

The complete pre-commit suite passed in a clean checkout, including Ruff, Ruff format, mypy, generated-config validation, documentation checks, device-API validation, naming checks, and compile-all. `git diff --check` also passed. No unit test is added because a CPU mock would only assert the selected cleanup call; the regression is latency and GPU-memory behavior covered by the controlled GPU A/B tests.

### API and Usage Example

No API or configuration change.

### Design & Code Changes

[...]

PR #7896, opened 2026-09-17
[rollout, perf] fix: backport weight-refit GC removal to v0.9.0

doudouyzgf:
### What does this PR do?

Backport #7864 and the substantive change from #7873 to release/v0.9.0.

- Replace the aggressive cleanup before rollout weight resume with a plain allocator cache release.
- Replace the remaining aggressive cleanup after actor offload with synchronization followed by allocator cache release.
- Remove full Python GC from the corresponding engine load and bucketed-transfer cleanup paths.
- Defer the final receiver ACK until its CUDA IPC mapping is released, preserving prompt bucket reclamation without relying on GC latency.

The changes cherry-picked cleanly onto the current release branch. The second, unrelated fix-ci commit from #7873 was intentionally excluded because its Megatron test expectation is specific to main and fails against the v0.9 implementation.

### Test

```bash
python -m pytest -q \
  -k "not distributed_accumulation_matches_per_micro_batch_sync"
```

Result: 21 passed, 1 deselected.

Ruff passed for every changed Python file, and git diff --check passed.

The two-rank Gloo test was deselected because the local macOS environment cannot bind its process group. Bucketed-transfer tests could not run locally because this environment does not provide vLLM; the CUDA IPC behavior and end-to-end weight-refit path retain the validation from the original upstream changes.


wuxibin89:
We will release 0.9.1 in next few days.

doudouyzgf:
> We will release 0.9.1 in next few days.

Thanks. Will 0.9.1 be cut from release/v0.9.0 and include #7864 and #7873? If so, could we merge this backport now?

We need the fix for an active rollout and can pin the merge commit immediately, then switch to the 0.9.1 tag once published. This would let us avoid waiting several days without introducing any additional divergence from the upcoming patch release.

wuxibin89:
@doudouyzgf 0.9.1 will be cut from main.

doudouyzgf:
> @doudouyzgf 0.9.1 will be cut from main.

Thanks for clarifying. In that case, we would strongly prefer this backport. We are currently based on release/v0.9.0, while moving to a 0.9.1 cut from main would bring 100+ unrelated commits and materially expand the integration risk for this urgent regression fix. 

Could we merge it so we can pin the release-branch commit now? We can evaluate the broader 0.9.1 upgrade separately.

doudouyzgf:
Thanks @wuxibin89 !

PR #7371, opened 2026-08-12
[hardware, fsdp, vllm] feat: add Intel GPU (XPU) support

kahlun:
### What does this PR do?

Adds **Intel GPU (PyTorch XPU)** as a hardware backend for verl, enabling end-to-end **GRPO, PPO, and SFT** training with **FSDP/FSDP2** actors and **vLLM** colocated rollout. Intel GPU is integrated through the existing platform-abstraction layer (the same path used by CUDA / Ascend NPU), so no algorithm or trainer code is forked — device dispatch is resolved generically via the platform registry and `get_device_name()`.

Validated end-to-end on Intel Arc Pro B-series (Battlemage) hardware, single-GPU through multi-GPU.

### Checklist Before Starting

- [x] Search for similar PRs. No existing PR adds Intel GPU / `torch.xpu` support.
- [x] Format the PR title as `[{modules}] {type}: {description}`: `[hardware, fsdp, vllm] feat: add Intel GPU (XPU) support`.

### Test

CI cannot cover this change: **Intel GPU hardware is not available in GitHub Actions.** Validated manually with Qwen2.5-0.5B-Instruct on GSM8K. Repro scripts are added under `tests/special_intel_gpu/`.

| Workload | Topology | Result |
|----------|----------|--------|
| GRPO | 1 GPU | ✅ pass (step metrics; rollout↔actor pearson ≈ 0.999) |
| GRPO | 2 GPU | ✅ pass |
| GRPO | 4 GPU | ✅ pass |
| PPO  | 2 GPU (pass `NUM_GPUS=2`; script default is 4) | ✅ pass |
| SFT  | 2 GPU (pass `NUM_GPUS=2`; script default is 4) | ✅ pass |

Commands:

```bash
NUM_GPUS=2 bash tests/special_intel_gpu/run_grpo_intel_gpu.sh
NUM_GPUS=2 bash tests/special_intel_gpu/run_ppo_intel_gpu.sh
NUM_GPUS=2 bash tests/special_intel_gpu/run_sft_intel_gpu.sh
bash tests/special_intel_gpu/run_grpo_colocate_rm_intel_gpu.sh
bash tests/special_intel_gpu/run_standalone_gen_intel_gpu.sh
```

### API and Usage Example

No public API or CLI changes. Intel GPU is selected automatically when `torch.xpu` is available; existing recipes run unmodified with `trainer.device=xpu`.

```bash
[...]

PR #3020, opened 2025-08-12
[fsdp] fix: set _set_allocator_settings to True to avoid fsdp2 oom

chenhaiq:
### What does this PR do?

Enable expandable_segments to avoid the increasing memory fragmentation caused by temporary variables during the training process of fsdp2, which may trigger probabilistic out-of-memory (OOM) errors.

Since both sglang and vllm can not work with expandable_segments:True, it has to be turn off during rollout.

### Checklist Before Starting

- [x] Search for similar PRs. Paste at least one query link here: https://github.com/volcengine/verl/pulls?q=is%3Apr+is%3Aopen+fsdp2+
- [x] Format the PR title as `[{modules}] {type}: {description}` (This will be checked by the CI)
  - `{modules}` include `fsdp`, `megatron`, `sglang`, `vllm`, `rollout`, `trainer`, `ci`, `training_utils`, `recipe`, `hardware`, `deployment`, `ray`, `worker`, `single_controller`, `misc`, `perf`, `model`, `algo`, `env`, `tool`, `ckpt`, `doc`, `data`
  - If this PR involves multiple modules, separate them with `,` like `[megatron, fsdp, doc]`
  - `{type}` is in `feat`, `fix`, `refactor`, `chore`, `test`
  - If this PR breaks any API (CLI arguments, config, function signature, etc.), add `[BREAKING]` to the beginning of the title.
  - Example: `[BREAKING][fsdp, megatron] feat: dynamic batching`

### Test

Without this fix, memory reserved could be very high after compute_log_prob or update_actor.
```
(WorkerDict pid=339320) [2025-08-11 17:43:01] dp actor After compute_log_prob, memory allocated (GB): 5.53, memory reserved (GB): 73.59, device memory used/total (GB): 77.47/79.15
```

With this fix, it stays low during training.
```
(WorkerDict pid=396879) [2025-08-12 07:39:42] dp actor After compute_log_prob, memory allocated (GB): 4.95, memory reserved (GB): 14.20, device memory used/total (GB): 17.72/79.15
```

### API and Usage Example

> Demonstrate how the API changes if any, and provide usage example(s) if possible.

```python
# Add code snippet or script demonstrating how to use this
```

### Design & Code Changes

> Demonstrate the high-level design if this PR is complex, and list the specific changes.

### Checklist Before Submitting

> [!IMPORTANT]
> Please check all the following items before requesting a review, otherwise the reviewer might deprioritize this PR for review.

[...]

PR #7917, opened 2026-09-18
[hardware, util, fsdp, profile] enable intel XPU to Verl with plugin mechanism with extra General API abstraction

kahlun:
### What does this PR do?

Extends `PlatformBase` (`verl/plugin/platform/platform_base.py`) with three optional hooks so a hardware backend that can't reuse CUDA's exact code paths plugs in cleanly, instead of the rest of verl growing more `if device == "..."` branches. These were surfaced while bringing up an Intel XPU backend (companion PR: #7371) and all default to today's behavior, so CUDA/ROCm/NPU are unaffected. It will a depdency from these PR for the XPU plugin. 

## The verl-hardware-plugin 

1.   the core code and some docs to properly show intel xpu use with the this PR 7917 for the new API

-    link https://github.com/verl-project/verl-hardware-plugin/pull/22

2.   the dockerfile usage of intel xpu of hardware plugin with verl

-    link https://github.com/verl-project/verl-hardware-plugin/pull/23

   ``` after this 7917 merged, will adjust the pr <23> docker image 's verl git clone to the official version ```

---

1. **Reduce-op fallback — `is_reduce_avg_supported()`.** Some collective backends only implement `ReduceOp.SUM`, not `AVG`. `verl.utils.device.all_reduce_avg()` is a drop-in replacement for `dist.all_reduce(..., op=ReduceOp.AVG)` that checks this hook and falls back to `SUM` + a manual divide by world size when `AVG` isn't supported. `sft_trainer.py`, `engine_workers.py`, and `profiler/performance.py::reduce_timing()` now go through it instead of hardcoding `ReduceOp.AVG`. `fsdp_utils.py::apply_fsdp2()` additionally calls `model.set_force_sum_reduction_for_comms(True)` on every `fully_shard()`'d module — the root **and** each individually-wrapped layer, since each has independent FSDP state — when the platform reports `AVG` unsupported, and raises a clear `RuntimeError` if the installed torch predates 2.8 (when that API was added).
2. **Attention fallback — `attention_utils_module()`.** Lets a platform point `verl/utils/attention_utils.py` at its own flash-attn-equivalent module (a dotted path exposing `index_first_axis`/`pad_input`/`rearrange`/`unpad_input`), checked after the NPU path and before falling back to the `flash_attn` pip package or the pure-PyTorch implementation. Also moved `torch_functional.py`'s two remaining hardcoded `flash_attn` / `get_device_name() == "cuda"/"npu"` branches onto this same dispatcher.
3. **Profiler — `profiler_markers()` / `dist_profiler_cls()`.** `profiler_markers()` lets a platform supply its own `(mark_start_range, mark_end_range, mark_annotate, marked_timer)` tuple, picked up by `profiler/__init__.py` when neither the `nvtx` package nor NPU's built-in markers apply. `dist_profiler_cls(tool)` lets a platform register a `DistProfiler` subclass under its own `profiler.tool` name; wired into `DistProfiler.__init__` (`profile.py`) and into the three rollout HTTP servers (`vllm_async_server.py`, sglang's `async_sglang_server.py`, `trtllm_async_server.py`) so a plugin-supplied tool name isn't rejected there either.

No public API changes — every new `PlatformBase` method is an optional override with a safe default (`is_reduce_avg_supported() -> True`, the other three `-> None`).

### Test

```bash
pre-commit run ruff ruff-format --files verl/plugin/platform/platform_base.py \
  verl/utils/device.py verl/utils/attention_utils.py verl/utils/fsdp_utils.py \
  verl/utils/profiler/__init__.py verl/utils/profiler/performance.py \
  verl/utils/profiler/profile.py verl/utils/torch_functional.py \
  verl/trainer/sft_trainer.py verl/workers/engine_workers.py \
  verl/workers/rollout/sglang_rollout/async_sglang_server.py \
  verl/workers/rollout/trtllm_rollout/trtllm_async_server.py \
  verl/workers/rollout/vllm_rollout/vllm_async_server.py

PYTHONPATH=. pytest tests/plugin/test_platform_abstraction.py -q
```

This PR doesn't add or modify any test files. `tests/plugin/test_platform_abstraction.py` covers the pre-existing `PlatformBase` surface but doesn't yet exercise the four hooks added here — that's a gap to close in a follow-up, ideally alongside #7371's `PlatformXPU`. The reduce-op fallback and the XPU attention/profiler paths were validated end-to-end on real Intel Arc B-series hardware as part of #7371's GRPO/PPO/SFT runs; re-validated in this PR for fsdp2.


verl/plugin/platform/platform_base.py
```diff
@@ -166,6 +166,30 @@ def visible_devices_envvar(self) -> str:
...
+
+        Some collective backends only implement ``SUM`` for
+        ``all_reduce``/``reduce_scatter``. Callers that want an averaged
+        reduction should check this first and fall back to a ``SUM``
+        all-reduce followed by a manual divide by world size when ``False``
+        (see :func:`verl.utils.device.all_reduce_avg`). FSDP2's
+        ``set_force_sum_reduction_for_comms`` should also be enabled when this
+        is ``False``, if the platform supports FSDP2.
+        """
+        return True
```
wuxibin89, line 180: If XPU doesn't support `ReduceOp.AVG`, then how do you do gradient reduce_scatter and all_reduce in HSDP?
https://github.com/pytorch/pytorch/blob/v2.13.0/torch/distributed/fsdp/_fully_shard/_fsdp_collectives.py#L846

kahlun: On coverage — the fix is validated on FSDP2 with a single-mesh setup (fsdp_size=-1): 6 runs with it and 6 without, back-to-back on the same node, clean pass every time with it and the same failure every time without. What that doesn't cover is the 2D mesh, so the all-reduce leg is from reading the source rather than measured. I'd rather keep this PR to the platform hooks themselves, since they're device-neutral and every default preserves current CUDA behavior. The XPU-side validation is a separate change, and I'll send it as its own PR: a 2×2 mesh run, plus a CPU/gloo test asserting SUM on both a 1D and a 2D mesh so it's covered in CI without needing XPU hardware.

wuxibin89: From https://github.com/intel/torch-xpu-ops/issues/3020, I think XCCL do support `ReduceOp.AVG` but seems to have double-division bug? If this bug still exists, it may also affect FSDP gradient reduce_scatter and all_reduce in 1/2D mesh, so disable it in verl side alone is not enough. Please check whether this bug has been fixed.

kahlun: Hi, pytorch xpu ops 3020 fixed. It is fixed, when using pytorch2.13 will not encounter average divison bug.
the Average that i mention not support is more of the AVG Reduction in Reduce (Scale-up & Scale-out)

this bug will be fix in oneccl2022.2 = pytorch2.15.

but for the FSDP gradient reduce_scatter and all_reduce in 1/2D mesh question or concern, i will put in the comment

On HSDP coverage: you're right that it wasn't covered, and running it surfaced a real bug — but it's not a Ray bug, and I want to be precise about what's proven versus what isn't.

Root cause: worker.py:273 only pins a device with set_device() when a RAY_EXPERIMENTAL_NOSET_* env var happens to be set. In the default config it never pins at all — it silently assumes Ray's visibility mask has already relabeled devices so index 0 is correct. That assumption doesn't hold on XPU (a separate torch-xpu issue discards the mask after lazy init), so every rank lands on the same physical GPU. Ray itself isn't at fault here — I checked ray-project/ray#64440, it's correct.

I have a fix on a branch (not yet in this PR): pin unconditionally, and resolve the index against what the process can actually see rather than trusting the mask blindly. Verified on real 4x B60 hardware against verl's own create_device_mesh() / collective path — 1 of 4 distinct GPUs before the fix, 4 of 4 after, 0 AVG failures across both 1-D and 2-D meshes.

What I haven't yet gotten to complete: a full real GRPO training run under this PR's exact colocated-RM + FSDP2 + transfer_queue configuration hit a separate, unrelated stall — verl's shared-memory weight-transfer fallback (BucketedWeightSender, used because XPU doesn't support CUDA-style IPC) deterministically stalls partway through sending weights into the rollout engine. Reproduced twice, same byte offset both times, independent of oneCCL env settings, so it isn't a flaky resource issue. Separate bug, still open, will report on its own.

So: the device-pinning fix is verified correct at the mechanism level — real hardware, real verl code path, not a mock. A full end-to-end HSDP training run under this specific config is still blocked by that second, unrelated issue.
I will change the PR description:
narrow the claim from "xccl does NOT support ReduceOp.AVG" to the accurate version — oneCCL implements AVG on its SYCL-kernel path but not its scheduler path, path selection is internal and not capability-aware, so a framework can't guarantee an AVG request will be serviceable at runtime. Forcing SUM is the only way to guarantee it for pre oneccl 2022.2

kahlun:
the full description for the question on, quite lengthy or detail
``` text
From https://github.com/intel/torch-xpu-ops/issues/3020, 
I think XCCL do support ReduceOp.AVG but seems to have double-division bug?
If this bug still exists, it may also affect FSDP gradient reduce_scatter and all_reduce in 1/2D mesh, 
so disable it in verl side alone is not enough. Please check whether this bug has been fixed.
```

Checked #3020 — it's real, and it's fixed on the stack I'm running. It's AVG double-division on
XCCL's small-message protocol (the non-monotonic size pattern in that issue). The fix shipped in
oneAPI DLE 2026.0 rather than in a torch release; `torch.version.xpu` reports `20260000` on my
image, which is how I can tell the measurements below are on the fixed build. Stack: 4× Arc Pro
B60, `torch 2.13.0+xpu`, oneCCL 2022.1.

So I measured your case instead of reasoning from the source. 4 ranks under `torchrun`, one rank
per physical GPU (verified by device uuid), real `fully_shard`. I wrapped
`_get_gradient_divide_factors` to record what FSDP2 actually requested, and checked gradients
against a reference that uses no AVG anywhere — each rank's local gradient, `ReduceOp.SUM` over
gloo on CPU, divided by world size by hand:

| mesh | ops FSDP2 requested `(pre, post, RS, AR)` | grad ÷ reference |
|---|---|---|
| 1-D `(4,)`, default | `None, None, AVG, AVG` | 1.000000 |
| 1-D `(4,)`, force_sum | `None, 4.0, SUM, SUM` | 1.000000 |
| 2-D `(2,2)`, default | `None, None, AVG, AVG` | 1.000000 |
| 2-D `(2,2)`, force_sum | `None, 4.0, SUM, SUM` | 1.000000 |

fp32; a surviving double division would show as `1/4 = 0.250000`. The AVG and force-SUM arms are
bitwise identical (`max|diff| = 0.000e+00`) — expected, since world size 4 is a power of two, so
oneCCL's AVG and an explicit `post_factor=4.0` are the same arithmetic. Raw
`all_reduce`/`reduce_scatter` AVG on the 2-D mesh's own size-2 subgroups is also correct (1.5
against a 3.0 SUM control). So no double division on either leg on either mesh — your reading was
right and my PR description is too broad.

There is a second, distinct failure, and it isn't a wrong number — it's a hard abort:
`coll_param.cpp:491 validate: EXCEPTION: average operation is not supported for the scheduler
path`. oneCCL has two execution paths, and AVG is implemented on the SYCL-kernel path but not on
the scheduler path (confirmed with Intel's oneCCL team). `ccl_coll_param::validate()` throws
unconditionally when the collective is allreduce/reduce_scatter/reduce and the reduction is avg —
no size threshold and no partial-correctness window, which is how I know it isn't #3020. The
caller doesn't pick the path: you request a reduction and oneCCL selects internally, and that
selection isn't capability-aware, so an AVG request the API accepted can be dispatched to a path
that cannot service it — surfacing mid-collective inside a training step. I'm not claiming the
path choice is wrong; it may be right for the topology. What's missing is capability-awareness in
the dispatch, and I'm following up with Intel to ask for that, or for AVG emulated as
SUM-then-scale on that path.

On "disabling it in verl alone is not enough" — agreed on the principle, and that's why the change
isn't scoped to verl's own AVG call sites. `is_reduce_avg_supported()` also gates the FSDP2 side,
at `fsdp_utils.py:615` in this PR: when it's `False`, `apply_fsdp2` calls
`set_force_sum_reduction_for_comms(True)` on the root *and* every `fully_shard()`'d layer, since
that API doesn't recurse into children. That's the same knob FSDP2 consults for its internal
reduce_scatter/all_reduce, so one flag covers both verl's `all_reduce_avg()` helper and FSDP's own
collectives — the `SUM, SUM` rows above are that block firing. Worth being explicit about what the
flag can and can't express: it's static, while oneCCL's path selection happens at runtime, so
`False` is the only safe answer a capability query can give when the path isn't predictable. It's
also a backend property rather than a device one, and it lives in the out-of-tree XPU platform
class — so when oneCCL's dispatch becomes capability-aware, flipping it back is a one-line change
there with nothing to revert in verl. The test is written and passing, not a follow-up:
`tests/utils/test_fsdp2_force_sum_reduction_on_cpu.py`, 4 gloo ranks on CPU, parametrised over a
1-D `(4,)` mesh and a 2-D `(2,2)` `("ddp", "fsdp")` mesh, calling the real `apply_fsdp2`; it
asserts both legs come back `ReduceOp.SUM` and that gradients match the default AVG path at
`rtol=atol=1e-6`. Both parametrisations pass.

What I'll change in the PR: narrow the claim from "xccl does NOT support ReduceOp.AVG" to the
accurate version — oneCCL implements AVG on its SYCL-kernel path but not its scheduler path, path
selection is internal and not capability-aware, so a framework can't guarantee an AVG request will
be serviceable at runtime. Forcing SUM is the only way to guarantee it, and per the table it's
numerically exact.

On size coverage, since #3020 was size-dependent: separately from FSDP I ran raw `all_reduce` and
`reduce_scatter` AVG at numel 0, 1, 8 and 1024 — the 0-dim scalar and size-1 cases that issue
lists as broken — and all return 1.5 against a 3.0 SUM control on two ranks. Double division
would give 0.75. So the small-message region the issue identifies as the risky one is correct on
this stack, and the FSDP runs above cover the gradient path itself up to 65536 elements including
a padded head.



---
if want the run log, i can put in gist if requested

PR #7519, opened 2026-08-22
[megatron, rollout] feat: MXFP8 training and SGLang/vLLM rollout

wengeezhang:
### What does this PR do?

Enables the **MXFP8 train + rollout combo (Megatron training, SGLang and vLLM rollout)** on Blackwell, addressing #5972 end-to-end with a single design goal: **train-inference consistency by construction**.

Three parts:

**1. Training (`fp8_recipe: "mxfp8"`)** — FP8 training with the `blockwise` recipe (Hopper) has been supported since #4223 via `override_transformer_config.fp8` on the Megatron-Bridge path. The only functional gap for mxfp8 was sequence padding: TE's MXFP8 quantizer requires **both** tensor dims divisible by 32 (`mxfp8_tensor.py` asserts `shape[-1] % 32 == 0 and prod(shape[:-1]) % 32 == 0`). The hidden dim is model-defined and naturally 32-aligned, but the packed THD token dim is not — the wgrad columnwise layout places 32-element scale blocks **along the token dim**, so packed sequences must be padded to 32-token boundaries. The existing FP8 padding (`lcm(16, align)`, designed for blockwise 1x128) is insufficient.

**2. SGLang rollout (`rollout.quantization: mxfp8`)** — SGLang is launched in MXFP8 mode against the bf16 checkpoint (via `quantization_config` override, no offline conversion — same pattern as the existing blockwise fp8 rollout), and weight sync quantizes bf16 actor weights on the fly.

**3. vLLM rollout (`rollout.quantization: mxfp8`)** — vLLM is launched with a MiniMax-style `quantization_config` that vLLM normalizes to its native `ModelOptMxFp8Config` (weight `fp8_e4m3fn` + `uint8` UE8M0 `weight_scale`, block `[1, 32]`). Refits reuse verl's existing FP8 pristine-layout record/stage/reprocess machinery — the MXFP8 kernel post-processing (flashinfer scale swizzle, Marlin repack, emulation dequant-to-bf16) is covered by registering two additional patchers, since the rewritten param names (`weight`/`weight_scale`, `w13_*`/`w2_*`) were already in `_FP8_REFIT_PARAM_NAMES`. vLLM's Marlin/emulation backends also allow **serving MXFP8 weights on pre-Blackwell GPUs (SM80+)**.

[...]

PR #7918, opened 2026-09-18
[rollout] fix: add opt-in exact MXFP8 exclusion matcher

wengeezhang:
## What

Adds an **opt-in, instance-local** exclusion matcher for verl-generated ModelOpt MXFP8 configs: `verl/utils/vllm/mxfp8_exclusion_patch.py` (`apply_mxfp8_exclusion_patch`), a CPU regression suite, and a doc describing the integration contract. The helper is **standalone and intentionally not wired into rollout startup**, so it changes no existing behavior on `main`.

## Why (the bug it targets)

For FP8/MXFP8 rollout, verl injects a router exclusion `model.layers.N.mlp.gate` for **every** decoder layer (to keep MoE routers in high precision), including dense models that have no router. vLLM 0.24's `ModelOptQuantConfigBase.is_layer_excluded()` keeps a legacy **substring** fallback (`exclude_module in prefix`), so `model.layers.N.mlp.gate` also matches the dense SwiGLU `model.layers.N.mlp.gate_up_proj`. The whole merged projection is then built as `UnquantizedLinearMethod` (BF16) while training runs it in fp8 — a real train/rollout precision mismatch on `gate_proj`/`up_proj` across every layer (e.g. 28 layers × 2 = 56 HF params on Qwen3-1.7B). `qkv`/`o`/`down` don't contain the substring, so they stay quantized. The upstream `ModelOpt >= 0.39.0` comment is about on-disk checkpoint configs and is **not** a runtime version guard; the substring branch runs unconditionally.

This was surfaced by the MXFP8 quantized-layer audit's engine-recheck on a B200 run.

image

## Design / safety

- Replaces `is_layer_excluded` on **one explicitly opted-in config instance** (marked `_verl_exact_mxfp8_exclusions: true` in the raw HF config) — it does **not** touch ModelOpt classes, `get_quant_method`, checkpoint-loaded configs, or other quant backends.
- Drops only the legacy substring fallback: keeps exact matching + fusion-aware `is_layer_skipped` (all-or-none shard validation) + case-sensitive `fnmatch` globs.
- Fails **closed**: an invalid/mismatched opt-in raises rather than silently leaving the bug active; repeated installation is a no-op.

## Not a duplicate

Searched open PRs on `verl-project/verl` for `mxfp8 exclusion`, `ignored_layers gate`, `gate_up_proj quant`, `is_layer_excluded`. The only related open PR is #7519 (MXFP8 training + SGLang/vLLM rollout), which introduces the rollout quantization but does **not** address this exclusion-substring collision. No other open PR covers it.

## Tests

- `python -m pytest tests/utils/test_mxfp8_exclusion_patch_on_cpu.py -q` → **23 passed** (dense 28-layer/56-param collision, router preservation, instance isolation, checkpoint compatibility, explicit opt-in, invalid-input fail-closed, fused shards, mixed exact/glob, no implicit parent/suffix matching, idempotence).
- `ruff@0.12.2 check` on the new files → clean.
- The CPU suite uses lightweight vLLM doubles and also exercises functions extracted from vLLM 0.24's real matcher. **GPU integration remains to be validated** on the MXFP8 feature branch (the helper must be wired into each model-building worker before it fixes a live run — see the doc's integration + B200 validation contract).

## AI assistance

This change was developed with AI assistance (OpenAI Codex authored the patch; Claude verified the root cause against vLLM 0.24 source and ran the CPU tests/ruff reported above). The human submitter owns and has reviewed the change.

## Hardware validation (B200)

The exclusion matcher this PR adds was exercised end-to-end on an NVIDIA B200, integrated into the MXFP8 rollout feature branch — which additionally (a) marks its verl-generated mxfp8 `quantization_config` with `_verl_exact_mxfp8_exclusions: true` to activate the matcher, and (b) registers the ModelOpt MXFP8 refit patches after the exclusion patchers. On `main` this PR is behavior-neutral until a caller marks a config (nothing on `main` builds an mxfp8 rollout), so this validates the matcher's logic rather than a `main` code path.

Setup: Qwen3-1.7B, `quantization=mxfp8`, vLLM 0.24 (`verlai/verl:vllm024.dev2`), TP1, 2 training steps. `hidden=2048`, `intermediate=6144`, so the fused `gate_up_proj` is `[12288, 2048]`.

Result — the quantized-layer audit at the first post-step weight sync (`TRAIN_EXIT=0`):
- **train vs configured rule:** `quantize the same layers (28 of 28 decoder layers, 311 parameters checked)`
- **engine re-check (live dtypes read back from vLLM):** `the rollout engine's live parameters agree with the training side (311 parameters re-checked against the engine itself)` — **0 disagreements**.

Without the matcher (same setup): the engine re-check reported **56 disagreements** — every `mlp.gate_proj` / `mlp.up_proj` across all 28 layers — because vLLM's ModelOpt legacy substring match excluded the fused `gate_up_proj` via the injected router name `mlp.gate`, so the rollout served gate/up in BF16 while training ran them fp8. With the matcher active, `gate_up_proj` is quantized, the weight sync loads the fused parameter correctly (no shape mismatch), and train = configured rule = engine live dtype.

Integrator note: when merging into a branch whose `build_fp8_method_patchers` appends more patchers after the exclusion `extend(...)` (e.g. ModelOpt MXFP8 refit patches), add them before the final `return` — an intermediate `return` right after the extend silently skips them.


wuxibin89:
> vLLM 0.24's ModelOptQuantConfigBase.is_layer_excluded() keeps a legacy substring fallback (exclude_module in prefix), so model.layers.N.mlp.gate also matches the dense SwiGLU model.layers.N.mlp.gate_up_proj. The whole merged projection is then built as UnquantizedLinearMethod (BF16) while training runs it in fp8 — a real train/rollout precision mismatch on gate_proj/up_proj across every layer

This bug seems to be specific to dense model, while nowadays most models are sparse MoE and user almost never enable low precision training on dense model, I prefer not to fix it.

wengeezhang:
> > vLLM 0.24's ModelOptQuantConfigBase.is_layer_excluded() keeps a legacy substring fallback (exclude_module in prefix), so model.layers.N.mlp.gate also matches the dense SwiGLU model.layers.N.mlp.gate_up_proj. The whole merged projection is then built as UnquantizedLinearMethod (BF16) while training runs it in fp8 — a real train/rollout precision mismatch on gate_proj/up_proj across every layer
> 
> This bug seems to be specific to dense model, while nowadays most models are sparse MoE and user almost never enable low precision training on dense model, I prefer not to fix it.

dense ffn blocks still exist in some fused moe models. Small precision change in dense blocks will result in differrent moe router top-k.

```
class TransformerBlock(nn.Module):
    def __init__(self, block_id: int = 0, n_heads: int = 8, n_kv_heads: int = 8, dim: int = 512, dim_ff: int = 2048):
        super().__init__()
        self.rms_norm_attn = RMSNorm(dim, eps=1e-5)
        self.attention = Attention(n_heads, n_kv_heads, dim)
        self.rms_norm_ffn = RMSNorm(dim, eps=1e-5)
        # todo: 1. 入参传哪些；2. moe_cfg 的初始化位置放在这里是否合适。
        moe_cfg = MoEConfig(dim)
        self.ffn = FFN(dim, dim_ff) if block_id < n_dense_layers else MoEFFN(moe_cfg)
```

wuxibin89:
For QAT+modelopt, the ignore_patterns is read from `qat.ignore_patterns`:
https://github.com/verl-project/verl/blob/main/verl/workers/rollout/vllm_rollout/vllm_async_server.py#L1217-L1240

https://verl.readthedocs.io/en/latest/low_precision/nvfp4_qat.html#fsdp-backend, the ignore_patterns has regex pattern `re:.*mlp.gate$` which should match `model.layers.N.mlp.gate_up_proj`
```bash
actor_rollout_ref:
  actor:
    fsdp_config:
      qat:
        enable: true
        mode: "w4a16"
        group_size: 16
        ignore_patterns:
          - "lm_head"
          - "embed_tokens"
          - "re:.*mlp.gate$"
        quantization_config_path: "recipe/qat/config/nvfp4_w4a16.json"
```

PR #7919, opened 2026-09-18
fix: move oversized CPU weights to IPC device before bucketed transfer

ccran:
## Summary

Fix bucketed weight transfer when an oversized tensor is on CPU.

The regular bucket path copies each tensor into the accelerator communication buffer before sending it, which implicitly normalizes the tensor device. The oversized direct-send path bypassed that copy and called `reduce_tensor(weight)` directly.

When LoRA/base-weight collection yields CPU tensors, an oversized tensor therefore produced a CPU reduction handle. The receiver still rebuilt every direct-send handle through the CUDA IPC path and rewrote `args[6]` as the device id, which is only valid for CUDA IPC reduction metadata. CPU reduction metadata has a different argument layout, so this can fail with an `IndexError` during weight sync.

## Fix

- Move oversized direct-send tensors to the communication buffer device before calling `reduce_tensor()`.
- Synchronize after the device copy so the IPC handle is created only after the transfer completes.
- Keep the migrated tensor alive until the receiver ACK returns.

## Test

Add a regression test that sends a 4 MB CPU tensor through a 1 MB CUDA IPC bucket. This specifically exercises the oversized direct-send path with a CPU source tensor, which the existing large-weight test did not cover.

## Why this matters for LoRA

The LoRA/base-weight collection path can materialize tensors on CPU, especially during initial base-weight synchronization. Small tensors worked because the regular bucket copy moved them to the accelerator buffer; oversized tensors skipped that device migration and failed in the direct-send path.


ccran:
Closing this submission so I can resubmit it using the current verl PR title/description conventions and AI-assisted contribution requirements. The replacement will contain the same code change with corrected PR metadata.

PR #7926, opened 2026-09-19
[trainer, rollout] fix: honor V1 standalone rollout memory budget

### LINKED ISSUE #7795: [Question][trainer] V1 separate_async shares GPU memory utilization between hybrid and standalone rollout
### Question

In V1 `trainer.v1.trainer_mode=separate_async`, should hybrid and standalone rollout replicas have independently configurable GPU memory utilization?

I encountered an OOM **during hybrid rollout initialization on the training GPUs** when setting:

```yaml
trainer:
  v1:
    trainer_mode: separate_async

actor_rollout_ref:
  rollout:
    gpu_memory_utilization: 0.9
```

This is the relevant configuration fragment, not a complete reproducer. The intention was to use a high memory fraction on the dedicated rollout GPUs, but the same value appears to apply to the hybrid replicas on the training GPUs as well.

### Source investigation

I checked main at `0d3f56a8980a55bfbbf1214ea54fa1b32ca1405c`. This is a source inspection of that revision, **not a GPU reproduction on that revision**. The full failing-run environment and traceback are not included in this initial question.

The initialization path appears to be:

1. [PPOTrainerSeparateAsync._setup()](https://github.com/verl-project/verl/blob/0d3f56a8980a55bfbbf1214ea54fa1b32ca1405c/verl/trainer/ppo/v1/trainer_separate_async.py#L126-L152) calls `super()._setup()`, then creates the standalone manager with the same `self.config`.
2. [The base trainer](https://github.com/verl-project/verl/blob/0d3f56a8980a55bfbbf1214ea54fa1b32ca1405c/verl/trainer/ppo/v1/trainer_base.py#L352-L368) unconditionally creates the hybrid manager using `actor_rollout_wg`. Sleeping the replicas happens only after initialization succeeds.
3. [LLMServerManager](https://github.com/verl-project/verl/blob/0d3f56a8980a55bfbbf1214ea54fa1b32ca1405c/verl/workers/rollout/llm_server.py#L373-L375) reads `config.actor_rollout_ref.rollout` for both managers.
4. `trainer.v1.separate_async.hybrid_rollout.enable_switch=false` controls runtime switching; it does not appear to skip hybrid initialization.

There is already a `standalone_gpu_memory_utilization` field, but its runtime override appears to be implemented only in [experimental fully_async_rollouter.py](https://github.com/verl-project/verl/blob/0d3f56a8980a55bfbbf1214ea54fa1b32ca1405c/verl/experimental/fully_async_policy/fully_async_rollouter.py#L104-L129), not the V1 separate-async path.

### Expected configuration / clarification

Ideally, V1 separate async could use a configuration such as:

```yaml
actor_rollout_ref:
  rollout:
    gpu_memory_utilization: 0.5             # hybrid replicas
    standalone_gpu_memory_utilization: 0.9 # dedicated rollout replicas
```

These values illustrate independent budgets; they are not claimed to be safe for every model. A null standalone value could retain the current fallback behavior.

Could maintainers confirm:

- Is the shared memory fraction in V1 separate async intentional, or is wiring up the existing standalone override missing?
- Is there an existing supported way to configure the two independently?
- If this is a gap, would applying the standalone override to a copied config when creating the V1 standalone manager be an appropriate fix?

### Related work / duplicate search

I searched issues and PRs for `standalone_gpu_memory_utilization`, `separate_async`, and hybrid/standalone memory issues and did not find a direct V1 fix.

- #6556 introduced separate memory knobs for experimental fully async.
- #7373 added hybrid switching in V1 separate async and addressed standalone weight-update memory handling.
- #7430 proposes exposing missing rollout YAML fields, including the standalone knob, but does not appear to wire its runtime behavior into V1.

AI assistance was used for source inspection and drafting this issue.

zupengwang:
### What does this PR do?

V1 `separate_async` currently ignores `standalone_gpu_memory_utilization` when creating the standalone rollout manager. Apply the existing override to a copied config so dedicated rollout GPUs can use a different budget while hybrid replicas retain their own setting.

Fixes #7795. This follows the runtime semantics introduced for experimental fully async in #6556. Missing or explicit `null` values retain the fallback to `gpu_memory_utilization`.

### Checklist Before Starting

- [x] Search for similar PRs: [configuration field](https://github.com/verl-project/verl/pulls?q=is%3Apr+standalone_gpu_memory_utilization), [issue reference](https://github.com/verl-project/verl/pulls?q=is%3Apr+7795+in%3Abody).
- [x] Use the required module/type title format.

Open #7430 adds a broader set of missing rollout YAML keys but does not implement this V1 runtime fix. Following [maintainer feedback](https://github.com/verl-project/verl/pull/7926#discussion_r4059130980), this PR also declares this one field in `rollout.yaml` and updates the four generated trainer configs. Other configuration-parity work remains outside this PR.

### Design & Code Changes

- Copy the full config before standalone manager creation, preserving OmegaConf interpolation context and leaving the trainer/hybrid config unchanged.
- Declare `standalone_gpu_memory_utilization: ${oc.select:actor_rollout_ref.rollout.gpu_memory_utilization}` in the default rollout YAML and regenerate the DP, Megatron, VeOmni, and TorchTitan reference configs.
- Cover default inheritance, explicit `null`, and ordinary Hydra overrides across all four model-engine configurations, with hybrid replicas enabled and disabled.
- Remove the unsupported `flush=True` argument from `logger.info` in standalone-only setup; the regression tests exposed the existing TypeError on that path.
- Update the V1 documentation to use an ordinary override without `++`.

### API and Usage Example

```bash
actor_rollout_ref.rollout.gpu_memory_utilization=0.5 \
actor_rollout_ref.rollout.standalone_gpu_memory_utilization=0.85
```

Choose fractions suitable for each GPU pool. The standalone value follows `gpu_memory_utilization` by default. An explicit `null` also falls back to that value. The override applies when `actor_rollout_ref.hybrid_engine=False` too.

### Test

Executed on the remote Linux host with Python 3.12.14 and `PYTHONPATH` unset:

```bash
env -u PYTHONPATH CUDA_VISIBLE_DEVICES= TORCH_COMPILE_DISABLE=1 TORCHINDUCTOR_DISABLE=1 \
  ../.venv/bin/python -m pytest -s -x --asyncio-mode=auto \
  tests/trainer/ppo/v1/test_separate_async_step_switch_on_cpu.py \
  tests/trainer/ppo/v1/test_trainer_base_on_cpu.py \
  tests/trainer/ppo/v1/test_v1_async_warmup_on_cpu.py \
  tests/trainer/ppo/v1/test_v1_async_alignment_on_cpu.py \
  tests/trainer/ppo/v1/test_rollout_profiling_on_cpu.py
```

- **107 passed**: 67 target tests and 40 adjacent tests.
- The 24 Hydra composition/setup cases all failed before adding the YAML field and pass afterward. They exercise DP, Megatron, VeOmni, and TorchTitan configuration composition, default inheritance, explicit `null`/numeric overrides, and hybrid enable/disable.
- The original runtime regression had 11 failures / 4 passes before the production fix.
- All four generated configs were regenerated with `scripts/generate_trainer_config.sh`; a second run passed without further changes.
- All 14 equivalent local pre-commit hooks passed with Ruff 0.12.2 and mypy 1.17.0; generated-config verification was included. The original isolated hook installation again exceeded a 60-second limit before hooks started, so this does not claim the original isolated pre-commit run passed.

The setup tests invoke the actual V1 `_setup` while mocking expensive base setup, LLM server creation, and checkpoint manager construction. They test configuration behavior, not execution of the four training engines. Existing CPU CI collects this `*_on_cpu.py` file.

The remote tests use the prepared environment, not the complete frozen CI environment: an earlier `uv run --frozen --all-packages --extra cpu` bootstrap timed out during downloads. A previous broader V1 directory run reached 36 passes before Ray/TransferQueue fixture initialization failed to create worker threads. No real multi-GPU hybrid/standalone vLLM/SGLang end-to-end job or performance measurement was run.

- [x] Read the contribution guide and agent instructions.
- [x] Update the V1 async trainer documentation and generated configs.
- [x] Add tests collected by the existing CPU CI workflow.
- [x] Human submitter confirmed review of the initial change and explicitly requested this review follow-up.
- [ ] Human submitter personally ran relevant tests; the results above were executed with AI assistance.
- Recipe submodule: not applicable.

AI assistance was used for investigation, implementation, documentation, and test execution. The submitting contributor remains accountable for the contribution.


Begunner (dismissed):

verl/trainer/ppo/v1/trainer_separate_async.py
```diff
@@ -151,8 +152,15 @@ def _setup(self):
         # initialize standalone rollout
         # TODO: make initialization parallel with super().init()
         hybrid_num_replicas = len(self.llm_server_manager.rollout_replicas)
+        # Standalone replicas have dedicated GPUs; keep their memory budget
+        # separate from the hybrid replicas that share GPUs with training.
+        standalone_config = deepcopy(self.config)
+        standalone_rollout_config = standalone_config.actor_rollout_ref.rollout
+        standalone_memory = standalone_rollout_config.get("standalone_gpu_memory_utilization")
```
wuxibin89, line 159: Please add an explicitly config field in `verl/trainer/config/rollout/rollout.yaml`
```yaml
gpu_memory_utilization: 0.5
standalone_gpu_memory_utilization: ${oc.select: actor_rollout_ref.rollout.gpu_memory_utilization}
```

zupengwang: Addressed in 7166f7564dff354ca63b598643daa35b9cb6e267. Added `standalone_gpu_memory_utilization` to `rollout.yaml` with the requested `${oc.select:actor_rollout_ref.rollout.gpu_memory_utilization}` default and regenerated all four trainer reference configs.

wuxibin89 (approved):

PR #7929, opened 2026-09-19
[utils] fix: accept bracketed IPv6 addresses in get_free_port

Redemption-ZTX:
## What

`get_free_port()` picks the socket family with `is_valid_ipv6_address(address)`. `ipaddress.IPv6Address` rejects the bracketed form, so `[::1]` is classified as IPv4 and the bind fails:

```python
get_free_port("[::1]")
# socket.gaierror: [Errno 8] nodename nor servname provided, or not known
```

The fix strips the brackets before choosing the family and binding.

## Why this is reachable

The bracketed form is how `ray.util.get_node_ip_address()` returns IPv6 addresses, and it is the notation used in URLs. Two call sites already work around this by hand:

```python
router_ip = ray.util.get_node_ip_address().strip("[]")   # naive_router.py:55
router_ip = ray.util.get_node_ip_address().strip("[]")   # inner_sglang_router.py:37
```

The other `get_free_port` callers pass their address straight through, so they are exposed:

- `workers/rollout/sglang_rollout/async_sglang_server.py:225`
- `workers/rollout/sglang_rollout/sglang_pd_replica.py`
- `workers/rollout/vllm_rollout/vllm_pd_replica.py`, `vllm_async_server.py`
- `checkpoint_engine/nccl_checkpoint_engine.py`, `nixl_checkpoint_engine.py`, `hccl_checkpoint_engine.py`

Moving the normalisation into `get_free_port` covers all of them at once.

## Scope

`is_valid_ipv6_address` is deliberately **not** changed. Other call sites use it to decide whether to *add* brackets when formatting a URL host (`f"[{addr}]:{port}" if is_valid_ipv6_address(addr) ...`), so making it accept brackets would produce `[::1]` → `[[::1]]`.

## Tests

Adds `tests/utils/test_net_utils_on_cpu.py` — this module had no test coverage at all. Covers IPv4/IPv6 classification, the `with_alive_sock` reservation path, and both the plain (`::1`) and bracketed (`[::1]`) forms, asserting the reserved socket lands on `AF_INET6`.

```
python -m pytest tests/utils/test_net_utils_on_cpu.py -q
```

- 8 passed with the fix
- Verified the tests catch the bug: reverting the change makes the two bracketed cases fail with the `gaierror` above
- `ruff check` and `ruff format --diff` clean

The IPv6 cases skip automatically when the host has no IPv6 loopback. Runs on CPU only.


verl/utils/net_utils.py
```diff
@@ -79,11 +79,20 @@ def get_free_port(address: str, with_alive_sock: bool = False) -> tuple[int, soc
     responsible for closing the socket before the port is actually bound
     by the target service (e.g. NCCL, uvicorn).
     """
-    family = socket.AF_INET6 if is_valid_ipv6_address(address) else socket.AF_INET
+    # Callers pass node addresses that may arrive in the bracketed form used by
+    # URLs and by `ray.util.get_node_ip_address()` for IPv6 (e.g. `[::1]`).
```
wuxibin89, line 83: Node ip address `[]` has been stripped in:
https://github.com/verl-project/verl/blob/main/verl/workers/rollout/vllm_rollout/vllm_async_server.py#L179

Redemption-ZTX:
Thanks for the pointer, and for explaining rather than just closing it.

You're right. I went back through the call sites and every one of them already normalises with `.strip("[]")` — including the checkpoint engines, which I had assumed were exposed — so the bracketed form never actually reaches `get_free_port`. My premise was wrong: I read the call line and didn't trace where `self._server_address` / `self.ip` get assigned. Closing it was the correct call.

Appreciated the review.

PR #7964, opened 2026-09-20
[algo] fix: keep rloo_vectorized advantages in the reward dtype

linhongyu510:
### What does this PR do?

`torch.bincount(weights=...)` accumulates in **float64** regardless of the weights' dtype, so the group sums in `compute_rloo_vectorized_outcome_advantage` promote the whole expression. A bf16 or fp16 reward tensor comes back as float64 advantages, while `compute_rloo_outcome_advantage` preserves the caller's dtype.

Reproduced on current `main` with the real functions, using **no singleton groups** so this is isolated from the `#7793` discussion:

```
index = ["p0","p0","p1","p1"]          # every group has 2 samples

input torch.float32    scalar->torch.float32    vectorized->torch.float32
input torch.bfloat16   scalar->torch.bfloat16   vectorized->torch.float64   <-- mismatch
input torch.float16    scalar->torch.float16    vectorized->torch.float64   <-- mismatch
```

The fix casts the group sums back to `scores.dtype`.

Both estimators are selectable via `algorithm.adv_estimator` and documented side by side in `docs/examples/config.rst`, so switching `rloo` → `rloo_vectorized` shouldn't change the dtype of what lands in `batch["advantages"]` downstream.

### Relationship to the existing singleton-group work

**This does not overlap with #7793 / #7581 / #7794.** I found the singleton disagreement independently while auditing the two RLOO paths, then searched and found it was already reported and fixed three times over — so I dropped my version of that fix rather than add a fourth. #7581 and #7794 look right to me, and `torch.where(c > 1, loo, scores)` matches the `mean=0` convention that `group_mean_std` already implements for GRPO singletons.

What none of them changes is the `bincount` dtype. To be sure this is additive rather than conflicting, I applied `#7794`'s `torch.where` on top of this branch and re-ran the file: **26 passed**. The two changes touch adjacent lines, so whichever lands first may leave a trivial textual conflict — happy to rebase on top of either.

### Test

```
$ pytest tests/trainer/ppo/test_core_algos_on_cpu.py -q
26 passed

$ ruff check … && ruff format --check …
All checks passed!  /  2 files already formatted
```

New test: `test_rloo_vectorized_preserves_reward_dtype`, parametrized over fp32/bf16/fp16. It asserts both that the vectorized output matches the requested dtype *and* that it matches the scalar path, and deliberately uses two-sample groups so it pins the dtype contract on its own.

I mutation-tested rather than trusting a green run — removing the `.to(scores.dtype)` fails exactly the bf16 and fp16 cases while fp32 still passes:

```
FAILED test_rloo_vectorized_preserves_reward_dtype[dtype1]   # bfloat16
FAILED test_rloo_vectorized_preserves_reward_dtype[dtype2]   # float16
2 failed, 1 passed
```

Also worth noting the existing `test_rloo_and_vectorized_equivalence` cannot catch this: it builds `token_level_rewards` as `torch.randn(..., dtype=torch.float32)`, and fp32 is the one dtype that is unaffected.

### API and Usage Example

No API change. Same values, same shapes; only the output dtype is corrected.

### Design & Code Changes

- `verl/trainer/ppo/core_algos.py`: cast the `bincount` group sums back to `scores.dtype` in `compute_rloo_vectorized_outcome_advantage`, with a comment on why.
- `tests/trainer/ppo/test_core_algos_on_cpu.py`: add the parametrized dtype test.

- [x] Apply pre-commit checks — `ruff` / `ruff-format` clean on both files.
- [ ] Add / Update the documentation — not applicable; no user-facing config or behaviour change beyond the dtype correction.
- [x] Add unit test(s) to cover the code — 1 parametrized test (3 cases), mutation-verified.
- [ ] Send a message in the `ci-request` channel — not needed: the covering test is `tests/trainer/ppo/test_core_algos_on_cpu.py`, picked up by the CPU unit-test job. No GPU path is touched.
- [ ] Not related to the `recipe` submodule.

---

Assembled with AI assistance. Every dtype table and test result above is from a local run against current `main` (`f542ae2`); the mutation result is the part I'd ask a reviewer to check first.


alanhuangyoo:
Heads up that #7581 rewrites this same line for a different reason (a group of one has no leave-one-out baseline, so it has to keep `scores` rather than zero), so whichever lands second will conflict. Together they come out as:

```python
group_sum = torch.bincount(inv, weights=scores)[inv].to(scores.dtype)
loo = (c * scores - group_sum) / (c - 1).clamp_min(1)
adv = torch.where(c > 1, loo, scores)
```

Happy to rebase mine onto yours if yours goes in first.

linhongyu510:
Thanks for the heads up — agreed they're complementary and should end up as your combined line. I checked both directions:

- Your `torch.where(c > 1, loo, scores)` is the right singleton behaviour; the current `* (c > 1)` **zeroes** a group-of-one instead of keeping its score. Modelling both against the scalar `compute_rloo_outcome_advantage`, with a singleton group present: scalar `[-1, 1, 5]`, your combined line `[-1, 1, 5]`, the old line `[-1, 1, 0]`. So #7581 is a correctness fix, not just a refactor.
- Your combined snippet keeps the `.to(scores.dtype)` on `group_sum`, so the bf16/fp16 → float64 promotion this PR fixes stays fixed. Without the cast a bf16 reward tensor still comes back float64.

No preference on order. If #7581 lands first I'll rebase this to just add `.to(scores.dtype)` onto your `group_sum` line and keep only the dtype test. If this lands first, your `torch.where` drops cleanly on top. Either way the end state is the single line you wrote — happy to be the one that rebases.

PR #6361, opened 2026-05-15
[trainer] mask kl reward nans

haoyang9804:
# [trainer] mask invalid KL values before KL-in-reward reduction

## How This Bug Is Triggered

Enable KL-in-reward:

```text
algorithm.use_kl_in_reward=true
```

and process a batch where a non-trainable response position has invalid logprob values:

```text
response_mask: [1, 0]
old_log_probs: [0.0, NaN]
ref_log_prob:  [0.0, NaN]
```

Masked-out positions can come from padding, tool observations, rejected rollout-correction tokens, or backend outputs that do not define logprobs for non-policy tokens.

## Symptom

There is no local crash at the masking line. The masked-out NaN survives and can later appear as NaN `token_level_rewards` or NaN `actor/reward_kl_penalty`.

The focused regression test builds a two-token batch with a NaN on the masked token. Before the fix, `NaN * 0` keeps the NaN. After the fix, rewards stay finite and the KL metric is `0.0`.

## Training Signal Validation

I also validated this with a real rollout-to-update path using Qwen2.5-0.5B-Instruct and GSM8K prompts. The validation script loads real prompts, generates real model responses, computes response logprobs with the model, applies a narrow controlled perturbation at the KL reward boundary, then compares buggy and fixed-equivalent reward/advantage behavior through one LM-head SGD update.

The recipe, runner, and relevant output content used for validation are included
below. Local machine paths are parameterized with environment variables.

Recipe content:

```json
{
  "name": "verl-qwen25-05b-gsm8k-f0c9-f61d-20260515-rerun",
  "target_bugs": [
    "BUG-F0C9A4E8D11B",
    "BUG-F61D0D8A9C2E"
  ],
  "script": "RL-Sentinel/scripts/verl_training_attack_validation.py",
  "runner": "${OUTPUT_DIR}/run_validation.sh",
  "model_path": "${MODEL_PATH}",
  "dataset_path": "${DATASET_PATH}",
  "device": "cuda:0",
  "seed": 20260515,
  "prompt_count": 2,
  "num_return_sequences": 2,
  "max_new_tokens": 24,
  "validation_path": [
    "load real GSM8K prompts",
    "generate real Qwen2.5-0.5B responses",
    "compute response logprobs with the model",
    "apply narrow controlled perturbations at the target reward/mask boundary",
    "compare buggy and fixed-equivalent reward/advantage behavior",
    "run one LM-head SGD update and record loss, grad_norm, and delta_norm"
  ],
  "outputs": {
[...]

PR #7970, opened 2026-09-21
[algo, trainer] fix: exclude aborted rollouts from the reported KL penalty

linhongyu510:
### What does this PR do?

`apply_kl_penalty` averages the per-sequence KL over **every** row of the batch:

```python
current_kl = masked_mean(kld, mask=response_mask, axis=-1)
current_kl = torch.mean(current_kl, dim=0).item()
```

Aborted rollouts have an all-zero response mask. `masked_mean` divides by `mask.sum() + 1e-8`, so those rows come out as exactly `0` rather than `NaN`, and the outer `torch.mean` then counts them as *zero KL*. The reported value sits below the true KL in proportion to the abort rate.

**This is not only a metric.** It is passed straight into `AdaptiveKLController.update`, so the controller reacts to the diluted number:

```
true per-sequence KL = 0.400, target_kl = 0.400, aborted = 2/8

  current_kl as computed : 0.3000      <-- 25% shortfall
  current_kl excl aborted: 0.4000

  after 200 updates (as computed ): beta = 0.193701
  after 200 updates (excl aborted): beta = 0.200000
```

The true KL is sitting *exactly on target*, so `proportional_error` should be 0 and beta should not move. Instead beta is driven down — weakening the KL penalty precisely when aborts are frequent, which is when drift is most likely.

**Aborted rollouts are a tracked condition everywhere else in this codebase**, which is what convinced me this call site is an oversight rather than a deliberate choice:

| site | how it handles fully-masked rows |
|---|---|
| `metric_utils.py:467` | `aborted_mask = (response_length == 0)`; excluded from score/reward metrics, and warns when *all* samples aborted |
| `core_algos.py:1187` / `:1201` | `seq_mask = (torch.sum(loss_mask, dim=-1) > 0)` — comment reads *"exclude fully masked sequences"* |
| `rollout_corr_helper.py:268` | `seq_valid_mask = response_mask.sum(dim=-1) > 0` |
| `apply_kl_penalty` | **no exclusion** |

The fix reports the mean over sequences with at least one valid token, and `0.0` when the whole batch aborted, so the controller is never handed a `NaN`.

### Test

New file `tests/trainer/ppo/test_kl_penalty_aborted_on_cpu.py` (5 tests). The `_on_cpu.py` suffix matches `cpu_unit_tests.yml`'s `tests/**/test_*_on_cpu.py` pattern, so CI picks it up.

- `test_aborted_rollouts_do_not_dilute_reported_kl` — 2 aborted of 8 must still report `0.4`.
- `test_reported_kl_matches_batch_without_the_aborted_rows` — the same 6 real rollouts report the same KL whether or not 3 aborted rows ride along. This is the invariant that matters: the number must not depend on how many aborts happen to be in the batch.
- `test_all_aborted_batch_reports_zero_and_does_not_crash` — the degenerate case stays finite.
- `test_aborted_rollouts_do_not_push_the_adaptive_coefficient_down` — drives the real controller 50 steps and asserts beta holds, plus a contrast case showing the diluted `0.3` would push it down.
- `test_token_level_rewards_still_exclude_masked_positions` — guards the surrounding behaviour I did *not* change.

```
$ pytest tests/trainer/ppo/test_kl_penalty_aborted_on_cpu.py \
         tests/trainer/ppo/test_core_algos_on_cpu.py \
         tests/trainer/ppo/test_metric_utils_on_cpu.py -q
66 passed

$ ruff check … && ruff format --check …
All checks passed!  /  2 files already formatted
```

Mutation-tested rather than trusting a green run — restoring the original all-rows mean fails exactly the 3 tests that assert the KL value, while the 2 that cover unrelated behaviour stay green:

```
FAILED test_aborted_rollouts_do_not_dilute_reported_kl
FAILED test_reported_kl_matches_batch_without_the_aborted_rows
FAILED test_aborted_rollouts_do_not_push_the_adaptive_coefficient_down
3 failed, 2 passed
```

One note on my own test: I first asserted `controller.value == 0.2` exactly and it failed at `0.20000000011920882` — float32 rounding in the KL computation leaves a ~1e-10 drift over 50 steps. Relaxed to `pytest.approx(rel=1e-6)`, which still fails the mutation by a wide margin.

### API and Usage Example

No API change. `actor/reward_kl_penalty` will read slightly higher in runs that have aborted rollouts, because it now reports the KL of the sequences that actually produced tokens. Runs with no aborts are bit-identical.

### Design & Code Changes

- `verl/trainer/ppo/ray_trainer.py`: in `apply_kl_penalty`, average the per-sequence KL over sequences with at least one valid token; fall back to `0.0` when every rollout aborted.
- `tests/trainer/ppo/test_kl_penalty_aborted_on_cpu.py`: new, 5 tests.

- [x] Search for similar PRs: [`is:pr apply_kl_penalty current_kl aborted`](https://github.com/verl-project/verl/pulls?q=repo%3Averl-project%2Fverl+is%3Apr+apply_kl_penalty+current_kl+aborted) (0 results) and [`is:pr reward_kl_penalty masked sequences`](https://github.com/verl-project/verl/pulls?q=repo%3Averl-project%2Fverl+is%3Apr+reward_kl_penalty+masked+sequences) — only #6361 (closed, unmerged), which changed the adjacent `kld = kld * response_mask` line for NaN masking and left the `current_kl` averaging untouched. No open issue covers this.
- [x] Format the PR title as `[{modules}] {type}: {description}` — validated with `tests/special_sanity/check_pr_title.py`.
- [x] Apply pre-commit checks — `ruff` / `ruff-format` clean on both files.
- [ ] Add / Update the documentation — not applicable; no config surface changed.
- [x] Add unit test(s) to cover the code — 5 tests, mutation-verified.
- [ ] Send a message in the `ci-request` channel — not needed: the covering test is a CPU test picked up by `cpu_unit_tests.yml`. No GPU path is touched.
- [ ] Not related to the `recipe` submodule.

---

Two things I'd flag rather than decide unilaterally:

1. **Is the metric or the controller input the thing to fix?** I changed both at once, since they read from the same variable. If you'd rather keep `actor/reward_kl_penalty` comparable with historical runs, the alternative is to feed the corrected value to the controller and report the old one — I think that's worse (two different "KL"s in one function) but it is a real choice.
2. I have **not** run a full training job. The controller trajectory above comes from driving the real `AdaptiveKLController` with the real `apply_kl_penalty` output, not from a GPU run, so the reward-curve impact is unmeasured.

Assembled with AI assistance; every number and table above is from a local run against current `main` (`ff93c77`), and the mutation result is the part I'd ask a reviewer to check first.


wuxibin89:
Why aborted rollouts have an all-zero response mask?

linhongyu510:
Good question — let me trace it, because the whole PR rests on this being true.

**An aborted rollout produces zero response tokens, and `response_mask` is built by masking over exactly those tokens.**

1. When a request is aborted mid-generation (e.g. `pause_generation(abort)` in partial rollout, or a length/timeout abort), the rollout server returns an **empty** token list. In vLLM (`vllm_async_server.py:706`):
   ```python
   if not final_res.outputs:
       return TokenOutput(token_ids=[], log_probs=None, ..., stop_reason="aborted")
   ```
   The trtllm and sglang servers have the same empty-output-on-abort path.

2. That empty response is right-padded to `response_length`, so the response block is all padding, and `response_mask` is the response-side attention mask over it — all zeros. In the agent-loop path (`agent_loop.py:779`) it's `response_mask = response_mask_output["input_ids"] * response_output["attention_mask"]`; in the classic path `compute_response_mask` (`ray_trainer.py:135`) is `attention_mask[:, -response_length:]`. Both give an all-zero row when nothing was generated.

3. This is the **same condition `metric_utils` already keys on**. `_compute_response_info` derives `response_length = response_mask.sum(-1)` (`metric_utils.py:87`), and `compute_data_metrics` then does `aborted_mask = (response_length == 0)` (`:467`) and excludes those rows from the score / reward / length means — even warning `"All samples are aborted"` when the whole batch is empty. So "all-zero response mask" and "aborted rollout" are the same set of rows, by verl's own definition.

`apply_kl_penalty` is the one place that reduces over the batch **without** applying that exclusion:
```python
current_kl = masked_mean(kld, mask=response_mask, axis=-1)  # -> 0 for an all-zero row (÷ mask.sum()+1e-8)
current_kl = torch.mean(current_kl, dim=0).item()           # counts that 0 as "zero KL"
```
so the reported KL — which feeds `AdaptiveKLController.update` — is pulled toward 0 in proportion to the abort rate, and the controller lowers β when it shouldn't.

If in your setup aborted rows are dropped before `apply_kl_penalty` (some configs filter them earlier), then this is a no-op rather than a fix — the `torch.where`/valid-mask guard just makes the two cases agree, matching what `agg_loss` and `rollout_corr_helper` already do. Happy to narrow the PR description to that framing if you'd prefer, or to gate it behind the same `response_length == 0` definition `metric_utils` uses so it's visibly the same condition.

PR #7136, opened 2026-07-24
[rollout, vllm] fix: FP8 rollout weight resync on vLLM 0.20.x via native layerwise reload lifecycle

KeitaW:
### What does this PR do?

Fixes FP8 rollout weight resync (`actor_rollout_ref.rollout.quantization=fp8`) on vLLM 0.20.x by driving vLLM's native layerwise reload lifecycle around verl's bucketed IPC weight sync.

Related to #5683 (a failure of the same FP8-resync family on a newer pin). This PR does NOT fix the exact environment reported there (verl 0.7.1 + vLLM 0.17.0): the patch is deliberately a no-op below vLLM 0.20, we did not reproduce the 0.17 failure, and a collaborator report in that issue indicates Qwen3 MoE FP8 already works on vLLM 0.17 with newer verl. This PR should not auto-close #5683.

Complementary to the merged #5599, which routes the STANDARD (non-FP8) weight sync through vLLM's `reload_weights`; that path is gated `not is_fp8_model(...)`, so FP8 models still take the bucketed `load_quanted_weights` path this PR fixes (details under Design & Code Changes).

### Checklist Before Starting

- [x] Search for similar PRs. Query links:
  - https://github.com/verl-project/verl/pulls?q=is%3Apr+is%3Aopen+5683+in%3Abody (0 results)
  - https://github.com/verl-project/verl/pulls?q=is%3Apr+is%3Aopen+fp8+weight+sync (reviewed; none touch the vLLM >= 0.20 kernel-format reload lifecycle for FP8 resync — see "Why this is not duplicating an existing PR" below)
- [x] Format the PR title as `[{modules}] {type}: {description}`: `[rollout, vllm] fix: FP8 rollout weight resync on vLLM 0.20.x via native layerwise reload lifecycle` (modules `rollout`, `vllm`; type `fix`; no API/config/signature break, so no `[BREAKING]`)

### Test

**Mechanism validation (verl v0.8.0 tag 7aed6b2 x vLLM 0.20.2 x CUDA 13.0, Qwen3-235B-A22B (MoE), GRPO with mixed code+math RLVR data and sandbox-executed code rewards, BF16 actor + FP8 rollout, B200 nodes):**

1. Crash reproduction (before fix): first weight resync crashes with the `ValueError` under Design & Code Changes below; 13 hits of the signature in the failing run's log.
[...]

Issue #6847, opened 2026-06-25
[Bug] FlashInfer FP16 MoE produces corrupted rollouts with Qwen3-Coder-30B-A3B in multi-node colocate-async


HanlinDu:
### Machines

- 2 nodes
- 8 x NVIDIA H20 per node
- 96 GB GPU memory per GPU (`97871 MiB`)
- Compute capability: 9.0
- NVIDIA driver: `535.161.08`
- Driver-reported CUDA compatibility: `12.9`

### Container image

The test ran in a private container from volcengine images, and the inner url is:
`vemlp-cn-beijing.cr.volces.com/preset-images/verl:v0.7.1_uni-agent`.

This image is built from `hub.docker.com/layers/verlai/verl/vllm017.dev3`, 
and can be reproduced with following configurations:

- Ubuntu `22.04.5 LTS`
- Image build serial: `20250714`
- CUDA toolkit/runtime: `12.9.1`
- cuDNN: `9.16.0.29`
- NCCL: `2.27.3`
- Python: `3.12.13`
- PyTorch: `2.11.0+cu129`
- vLLM: `0.20.2+cu129`
- FlashInfer: `0.6.8.post1`
- FlashAttention: `2.8.3`
- Triton: `3.6.0`
- Ray: `2.54.1`
- Transformers: `5.12.1`
- Megatron-Core: `0.16.0`
- mbridge: `0.15.1`

### Model

`Qwen3-Coder-30B-A3B-Instruct`

### Source revisions

Uni-Agent:

```text
Repository: https://github.com/HanlinDu/uni-agent (forked from https://github.com/verl-project/uni-agent)
Commit: 0754facf6537aee0a929ec4df3fa59f2495ed038
Subject: [core] fix: replace per-run log sinks with single dispatch sink to fix leaks and HDFS stalls (#63)
```

VERL:

```text
Repository: https://github.com/volcengine/verl
Commit: 8a694930275061f52ebd538c906ef8819af56dbd
Version description: v0.8.0-86-g8a694930
Subject: [fsdp, model] feat: per-unit LoRA summon, FSDP1/2 compatibility, and strip-modules support (#6512)
```

Both worktrees were dirty. Relevant local VERL changes were:

1. Run `process_weights_after_loading()` under
   `set_current_vllm_config(self.model_runner.vllm_config)` after colocated
   weight loading.
2. Disable the multi-node vLLM CLI arguments in `vllm_async_server.py` for this
   colocated deployment.
[...]

PR #7987, opened 2026-09-22
[rollout, vllm] fix: run the standard weight sync through vLLM's layerwise reload lifecycle

### LINKED ISSUE #7978: [rollout, vllm] bf16 MoE weight sync breaks on Blackwell with vLLM 0.24's TRT-LLM MoE backend (4-D w13/w2 layout → shard_dim ValueError at first sync)
### System Info

Observed on 1×NVIDIA B200 (SM100, 183 GB, driver 595.91.07, CUDA 13.2 host), image `verlai/verl:vllm024.dev2`:

- verl: `feat/megatron-mxfp8-training` @ `79caec3c` (fork of PR #7519); the standard weight-sync path in question (`verl/workers/rollout/vllm_rollout/utils.py::update_weights_from_ipc` → `_update_weights` → `model.load_weights`) is identical on `main` @ `fa8f14f8`
- vLLM 0.24.0, FlashInfer 0.6.12, torch 2.11+cu130, transformers 5.5.3, Megatron-Core 0.18, Megatron-Bridge 0.5.2
- Python 3.12, Ubuntu 22.04 container

### Information

- [x] My own modified scripts

### Tasks

### Reproduction

**Setting.** GRPO on gsm8k with the Megatron engine (`use_mbridge=True`) and vLLM async rollout, **bf16 rollout (no `rollout.quantization`)**, TP=1, EP=1, on a Blackwell GPU. Model: a random-init 2-layer / 8-expert Qwen3-MoE (`Qwen3MoeForCausalLM`, hidden 512, `moe_intermediate_size` 256, top-k 2, saved in the per-expert `experts.N.{gate,up,down}_proj.weight` layout). The failure does not depend on the weights; it depends only on the MoE backend vLLM picks, so any Qwen3-MoE-style checkpoint on SM100 should hit it (see "Why" below — not verified on a full-size model).

```bash
python3 -m verl.trainer.main_ppo --config-name=ppo_megatron_trainer.yaml \
  actor_rollout_ref.model.path=<qwen3-moe checkpoint> \
  actor_rollout_ref.actor.megatron.use_mbridge=True \
  actor_rollout_ref.actor.megatron.tensor_model_parallel_size=1 \
  actor_rollout_ref.rollout.name=vllm actor_rollout_ref.rollout.mode=async \
  ++actor_rollout_ref.actor.megatron.override_transformer_config.moe_grouped_gemm=True \
  ++actor_rollout_ref.actor.megatron.override_transformer_config.moe_token_dispatcher_type=alltoall \
  trainer.n_gpus_per_node=1 trainer.nnodes=1 trainer.total_training_steps=4
# no rollout.quantization  -> bf16 rollout
```

**Failure.** The very first weight sync (`trainer.init` → `update_weights`) crashes inside vLLM's fused-MoE weight loader:

```text
  File "verl/workers/rollout/vllm_rollout/utils.py", line 334, in on_bucket_received
  File "verl/workers/rollout/vllm_rollout/utils.py", line 419, in _update_weights        # model.load_weights(param_updates)
  File "vllm/model_executor/models/qwen3_moe.py", line 627, in load_weights
  File "vllm/model_executor/layers/fused_moe/routed_experts.py", line 852, in weight_loader
  File "vllm/model_executor/layers/fused_moe/routed_experts.py", line 350, in _load_model_weight_or_group_weight_scale
  File "vllm/model_executor/layers/fused_moe/routed_experts.py", line 484, in _load_w13
  File "vllm/model_executor/layers/fused_moe/routed_experts.py", line 403, in _get_hidden_dim
ValueError: shard_dim=0 is not a valid data dimension for a 3D tensor (expected 1 or 2)
```

**Why (verified against the vLLM 0.24.0 source that ships in the image).**

1. For an *unquantized* fused MoE on CUDA, vLLM 0.24 auto-selects the backend in priority order `FLASHINFER_TRTLLM, FLASHINFER_CUTLASS, TRITON, BATCHED_TRITON` (`fused_moe/oracle/unquantized.py::_get_priority_backends`). The two FlashInfer entries are moved to the back only on compute-capability family 9.0 (Hopper), so on SM100 the TRT-LLM bf16 backend wins.
2. Its `process_weights_after_loading` → `_setup_kernel` → `convert_to_unquantized_kernel_format(FLASHINFER_TRTLLM)` → `convert_moe_weights_to_flashinfer_trtllm_block_layout` (`quantization/utils/flashinfer_utils.py`) rewrites `w13_weight` / `w2_weight` into the **4-D BlockMajorK layout** `[E, K/128, N, 128]` and re-registers them with `replace_parameter`.
3. verl's standard sync path feeds `model.load_weights` the exporter's per-expert 2-D tensors (`experts.N.gate_proj.weight`, …). `RoutedExperts.weight_loader` does `expert_data = param.data[expert_id]` — on the 4-D parameter that is a 3-D slice — and `_load_w13` is called with `shard_dim=0`, which `_get_hidden_dim` rejects. The load fails before verl's end-of-sync `process_weights_after_loading(model, model_config, device)` (utils.py, standard `else` branch) ever runs.

In short: verl loads checkpoint-layout weights straight into parameters that vLLM has already converted to a kernel layout, and this backend's kernel layout is not even the same rank. The FP8 paths do not hit this because they go through verl's pristine-layout stage/reprocess cycle (`vllm_fp8_utils.py`); the bf16 path has no such cycle.

**Workaround (verified on the same pod).** Pin a backend that keeps the checkpoint layout:

```yaml
actor_rollout_ref.rollout.engine_kwargs.vllm.moe_backend: triton
```

With that, the same 4-step run completes (`TRAIN_EXIT=0`, `rollout_corr/kl` ≈ 0.0019 on the random model, no NaN).

**Related.** #6847 (closed) is a different failure of the same FlashInfer-MoE-plus-weight-update family on vLLM 0.20.2 (stale kernel tensor addresses, fixed upstream by vLLM's `prefer_copy` on weight updates). #7136 (closed) explored driving vLLM's native layerwise reload lifecycle for the FP8 resync path.

### Expected behavior

A bf16 MoE rollout should survive weight sync on Blackwell with vLLM's default backend, exactly as it does on Hopper (where the Triton backend keeps the checkpoint layout).

Two ways to get there; I'd like maintainers' preference before opening a PR:

- **A. Use vLLM's own reload lifecycle for the standard path.** vLLM 0.24 ships `initialize_layerwise_reload(model)` → `model.load_weights(...)` → `finalize_layerwise_processing(model, model_config)` (`model_executor/model_loader/reload/layerwise.py`), which restores checkpoint-layout parameters from recorded metadata before loading and re-runs `process_weights_after_loading` afterwards (including clearing `_already_called_process_weights_after_loading`). `GPUModelRunner.reload_weights(weights_iterator)` is exactly this sequence, and vLLM's loader already calls `record_metadata_for_reloading(model)` at the initial load (`model_loader/utils.py`), so no extra setup is needed. Wrapping verl's bucketed receive in it would make the bf16 path layout-agnostic for every backend vLLM adds.
- **B. Extend verl's pristine-layout staging cycle to `UnquantizedFusedMoEMethod`.** The same `record / stage / load / reprocess / fold-into-live-storage` machinery that `vllm_fp8_utils.py` already runs for `Fp8MoEMethod` and `ModelOptMxFp8FusedMoE`, applied to the bf16 MoE method. Smaller surface, but verl keeps owning the layout bookkeeping.

Either way the fix needs a Blackwell run to validate (the failing backend is only selected there). Happy to do the PR once the direction is agreed.

---
*AI assistance was used for the source-level diagnosis and for drafting this report (Claude, with Codex); every claim above was checked against the vLLM 0.24.0 source and reproduced on hardware by the author.*

wengeezhang:
### What does this PR do?

Fixes #7978: the **standard (non-quantized) vLLM weight sync now runs inside vLLM's own layerwise reload lifecycle**, so it no longer depends on the MoE kernel keeping the expert weights in checkpoint layout.

**Root cause (verified against vLLM 0.24.0 and 0.29.0 source).** `update_weights_from_ipc` feeds the bucketed checkpoint-format tensors straight into the live parameters via `model.load_weights` and runs `process_weights_after_loading` once at the end. That holds only while the kernel layout equals the checkpoint layout up to a same-shape in-place transform (Triton MoE, FlashInfer CUTLASS). On SM100 vLLM auto-selects the FlashInfer TRT-LLM bf16 MoE backend (`fused_moe/oracle/unquantized.py::_get_priority_backends` puts it first on non-Hopper CUDA), whose `convert_moe_weights_to_flashinfer_trtllm_block_layout` turns `w13_weight` / `w2_weight` into 4-D BlockMajorK tensors. `RoutedExperts.weight_loader` then indexes `param.data[expert_id]` (3-D) and `_load_w13(shard_dim=0)` rejects it at the first sync:

```
ValueError: shard_dim=0 is not a valid data dimension for a 3D tensor (expected 1 or 2)
```

The conversion is one-way (it cannot be re-applied to its own 4-D output), so the end-of-sync `process_weights_after_loading` could not repair it either. Observed 2026-09-21 on 1×B200 (random-init Qwen3-MoE, TP1, vLLM 0.24.0); any bf16 MoE RL rollout on Blackwell with the default backend hits it.

**Fix.** Use the lifecycle vLLM's own `GPUModelRunner.reload_weights` wraps around `load_weights` (`vllm/model_executor/model_loader/reload/layerwise.py`):

1. `initialize_layerwise_reload(model)` before the first bucket — every layer's parameters are rebuilt in checkpoint layout on the meta device from the metadata `record_metadata_for_reloading` stored at `initialize_model`, and the weight loaders are wrapped to buffer incoming tensors per layer.
2. `model.load_weights(bucket)` per bucket, unchanged — when a layer's last weight arrives vLLM materializes it, loads the buffered tensors, re-runs that layer's `process_weights_after_loading` (clearing `_already_called_process_weights_after_loading`) and copies the kernel-layout result into the original storage, so CUDA-graph addresses stay valid.
3. `finalize_layerwise_processing(model, model_config)` after the last bucket — attention scales, padded weights and layers the stream did not cover — instead of the bare `process_weights_after_loading`.

Two verl-specific details: param tensors are **cloned** before `load_weights` (the layerwise loaders hold them past the bucket callback while `BucketedWeightReceiver` reuses its IPC buffer per bucket, so views would be overwritten), and **buffer updates are deferred** until after finalize (buffers are on the meta device in between).

Fallbacks keep the previous behaviour: `VERL_VLLM_LAYERWISE_RELOAD=0`, a vLLM without `model_executor.model_loader.reload` (older than 0.24), and LoRA syncs (LoRA-wrapped layers are created after vLLM records its reload metadata, so the lifecycle must not touch them). QAT, ModelOpt-QAT and FP8 syncs are untouched (they have their own preparation paths).

Workaround until this lands: `+actor_rollout_ref.rollout.engine_kwargs.vllm.moe_backend=triton` (verified on B200).

### Checklist Before Starting

- [x] Similar PRs: https://github.com/verl-project/verl/pulls?q=is%3Apr+is%3Aopen+layerwise+reload (0), https://github.com/verl-project/verl/pulls?q=is%3Apr+is%3Aopen+initialize_layerwise_reload (0), https://github.com/verl-project/verl/pulls?q=is%3Apr+is%3Aopen+7978+in%3Abody (0). Closed #7136 explored the same lifecycle for the *FP8* resync on vLLM 0.20; this targets the standard path and is import-gated. #6847 (closed) is a different failure of the FlashInfer-MoE-plus-weight-update family (stale kernel tensor addresses on 0.20.2, fixed upstream by `prefer_copy`).
- [x] Title format `[rollout, vllm] fix: ...`

### Test

```bash
python -m pytest tests/utils/test_vllm_weight_name_normalization_on_cpu.py -q -k "standard_sync or lora_base_sync"
# 4 passed
python -m pytest tests/utils/test_vllm_weight_name_normalization_on_cpu.py -q
# 23 passed; the 6 failures are all pytest.importorskip("vllm") tests (tied-alias + bucket tests) that need the vllm CI venv - they fail identically on main in a cpu-only checkout
ruff 0.12.2 check / format on the two changed files: clean
```

New CPU tests (stub the receiver and vLLM's reload module; no vLLM needed):
- `test_standard_sync_runs_vllm_layerwise_reload_around_the_buckets` — order `initialize → load(bucket 1) → load(bucket 2) → finalize`, no bare `process_weights_after_loading`, the tensors handed to `load_weights` are copies rather than views into the reused bucket.
- `test_standard_sync_falls_back_when_the_lifecycle_is_disabled` (`VERL_VLLM_LAYERWISE_RELOAD=0`).
- `test_standard_sync_falls_back_on_vllm_without_the_reload_module`.
- `test_lora_base_sync_keeps_loading_into_the_live_parameters`.

**Hardware: pending.** The failing backend is only selected on Blackwell. Planned this week on B200: (a) a 4-step random-init Qwen3-MoE bf16 run with vLLM's default (TRT-LLM) backend, single GPU; (b) the bf16 arm of a Qwen3-30B-A3B TP2/EP2 run on 4×B200. I will post the results here; until then the `moe_backend=triton` workaround stands and this can be treated as awaiting validation.

### API and Usage Example

No API change. New opt-out env var `VERL_VLLM_LAYERWISE_RELOAD=0` (default on when vLLM provides the lifecycle).

### Design & Code Changes

- `verl/workers/rollout/vllm_rollout/utils.py`
  - `_vllm_layerwise_reload_api()`: env gate + import probe, returns `(initialize_layerwise_reload, finalize_layerwise_processing)` or `None`.
  - `update_weights_from_ipc`: step 1 resets the per-sync state and, on the standard branch with `peft_config is None`, calls `initialize_layerwise_reload` for every synced model; step 3 gains an `elif self._layerwise_reload is not None` branch that calls `finalize_layerwise_processing` per model and then applies the deferred buffer updates.
  - `_update_weights`: while the lifecycle is active, clones the param tensors before `load_weights` and defers `buffer_updates` (cloned) instead of applying them per bucket. Rebased on top of #7948's `drop_tied_alias_updates`.
- `tests/utils/test_vllm_weight_name_normalization_on_cpu.py`: the four tests above.

Things to confirm on hardware: transient memory per layer (materialized checkpoint copy + kernel copy, ~2× one MoE layer) plus clones ≤ one bucket; vLLM logs `"<Layer>: Failed to load weights"` for layers whose tensors the stream never carries (e.g. rotary caches) — expected, worth confirming it is only those.

- [x] Read the Contribute Guide
- [x] ruff check/format on the changed files (full `pre-commit run --all-files` not run locally)
- [x] Documentation: behaviour change is internal to the sync path; the env opt-out is documented in the code comment. Happy to add a line to the rollout docs if preferred.
- [x] Unit tests added (CPU)
- [ ] ci-request — after the hardware run above

AI assistance was used for the diagnosis (vLLM source reading) and for drafting the code, tests and this description (Claude); the author reviewed every changed line and ran the tests above.


verl/workers/rollout/vllm_rollout/utils.py
```diff
@@ -403,6 +447,11 @@ def _update_weights(
                     f"FP8 weights loaded (async), loaded_params: {len(loaded_params)}, loaded_buffers: {loaded_buffers}"
                 )
             else:
+                layerwise = getattr(self, "_layerwise_reload", None) is not None
+                if layerwise:
+                    # The layerwise loaders hold these tensors until a layer's last weight arrives, which can
+                    # be in a later bucket, while the receiver reuses its IPC buffer per bucket: keep copies.
+                    param_updates = [(name, tensor.clone()) for name, tensor in param_updates]
```
wuxibin89, line 454: Layerwise reload accumulate tensors until all weights in single layer are collected. For large MoE model, this may cause OOM.

wengeezhang: OK，i'have switched to in-place staging, and tested it in a b200 hardware. the peak extra per sync is 6.0 MB (5/5 syncs), vs. 460.1 MB with the layerwise version (all 3 syncs).

verl/utils/vllm/vllm_moe_refit_utils.py
```diff
@@ -0,0 +1,184 @@
+# Copyright 2026 Bytedance Ltd. and/or its affiliates
```
wuxibin89: Rename `vllm_moe_refit_utils.py` to `vllm_unquant_utils.py`? As we already have a `vllm_quant_utils.py` for FP8/FP4.

wuxibin89 (approved):

